PR #60, opened 2026-03-25
fix: prevent --allow-unrelated-histories and fix v1 branch sync

tend-agent:
## Summary

- Add merge guidance to the `running-in-ci` skill: never use `--allow-unrelated-histories`, stash untracked files before merging, verify merge base exists
- Fix `update-action-branch.yaml` to use `fetch-depth: 0` — the default shallow clone can't fast-forward push, which is why v1 is currently 6 commits behind main

## Context

The bot was observed using `--allow-unrelated-histories` when merging upstream into PR branches ([worktrunk #1667](https://github.com/max-sixty/worktrunk/pull/1667)), causing add/add conflicts in every file. The root cause was untracked files blocking the merge, followed by the bot reaching for `--allow-unrelated-histories` as a recovery strategy.

The v1 branch sync issue is a separate bug: `actions/checkout` defaults to `fetch-depth: 1`, and a shallow clone can't verify fast-forward ancestry, so `git push origin HEAD:refs/heads/v1` is rejected.

## Test plan

- [ ] Verify CI passes
- [ ] After merge, confirm the update-action-branch workflow succeeds and v1 catches up to main

### DIFF
--- a/.github/workflows/update-action-branch.yaml
+++ b/.github/workflows/update-action-branch.yaml
@@ -12,6 +12,8 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       - name: Update v1 branch to match main
         run: git push origin HEAD:refs/heads/v1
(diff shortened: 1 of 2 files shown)


tend-agent (commented):
#57 has an identical change to `update-action-branch.yaml` — one of these should be closed to avoid a merge conflict.

PR #62, opened 2026-03-25
fix: add concurrency control to mention handle job

### LINKED ISSUE #61: fix: tend-mention double-posts when rapid events hit same issue
## Problem

Two near-simultaneous `issue_comment` events on worktrunk#1726 caused the `tend-mention` workflow to post duplicate responses ([comment 1](https://github.com/max-sixty/worktrunk/issues/1726#issuecomment-4129400513), [comment 2](https://github.com/max-sixty/worktrunk/issues/1726#issuecomment-4129402646) — 24 seconds apart).

## Root cause

The `concurrency` group with `cancel-in-progress: true` is only on the `verify` job, not on `handle`:

```yaml
jobs:
  verify:
    concurrency:
      group: ${{ github.workflow }}-${{ github.event.issue.number }}
      cancel-in-progress: true
    # ...
  handle:
    needs: verify
    # ← no concurrency control
```

Timeline on issue #1726:
1. **19:52:36Z** — maintainer posts comment → triggers run A
2. **19:52:44Z** — run A's `verify` completes in 2 seconds
3. **19:53:46Z** — maintainer posts second comment → triggers run B
4. **19:53:55Z** — run B's `verify` completes (nothing to cancel — A's verify already done)
5. **19:52:48Z–20:12:11Z** — run A's `handle` job runs
6. **19:53:58Z–20:15:11Z** — run B's `handle` job runs concurrently
7. **20:00:11Z / 20:00:35Z** — both post responses

The `cancel-in-progress` on `verify` is a no-op here because verify completes in ~2s, well before the next event arrives. The `handle` jobs (which take minutes) have no concurrency protection.

### 1. Move concurrency to `handle` job (simplest)

```yaml
handle:
  needs: verify
  concurrency:
    group: ${{ github.workflow }}-${{ github.event.issue.number || github.event.pull_request.number }}
    cancel-in-progress: true
```

The later event cancels the earlier `handle` job. Simple, but throws away work from the first run — which is usually fine since the second event has strictly more context (the later comment).

**Trade-off:** If the first run already posted a comment or pushed a commit before being cancelled, you get partial work + a fresh restart. But this is the standard GitHub Actions pattern and works well in practice.

### 2. Pre-post deduplication check in the action

Before posting a comment, check whether the bot already responded to the same triggering request within the last N minutes. This is more robust against partial work but adds complexity to the action itself.

### 3. Workflow-level concurrency (broadest)

Move concurrency to the top-level workflow. This serializes all jobs for the same issue/PR, including `verify`. Downside: a slow `verify` (unlikely but possible) blocks subsequent events.

## Recommendation

Option 1 is the simplest and most effective — move concurrency to the `handle` job (or apply it at workflow level). The `verify` job is fast enough that concurrency on it adds little value. The `handle` job is where the real work (and posting) happens, and `cancel-in-progress` ensures only the latest event's response gets through.

Option 2 could be a defense-in-depth addition but shouldn't be the primary mechanism.

tend-agent:
## Problem

The `handle` job in `tend-mention` had no `concurrency` group. When rapid `issue_comment` events hit the same issue (e.g. two comments 24 seconds apart), both `verify` jobs complete before either `handle` starts, so two `handle` jobs run in parallel and the bot double-posts.

The existing `concurrency` on `verify` is ineffective because verify completes in ~2 seconds — well before the next event arrives.

## Solution

Add a `concurrency` group with `cancel-in-progress: true` to the `handle` job, keyed on workflow name + issue/PR number. When a second event arrives while `handle` is running, the earlier job is cancelled and only the latest event's response goes through.

This matches the approach recommended in #61 (Option 1) and follows the same pattern already used by other jobs in the generated workflows.

## Testing

- Added `test_mention_handle_job_has_concurrency` that parses the generated YAML and asserts the `handle` job has a `concurrency` block with `cancel-in-progress: true`
- Test fails before the fix, passes after
- Full test suite (75 tests) passes

---
Closes #61 — automated triage

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -341,6 +341,40 @@ us with code execution, not just token minting.
   under adopter control but adds latency (webhook → our service →
   workflow_dispatch → runner) and complexity.
 
+## Concurrency strategy
+
+Every event-driven workflow uses `cancel-in-progress: true` — when a new event
+arrives for the same PR/issue while a previous run is still going, the old run
+is cancelled and only the latest event is processed.
+
+| Workflow | Group key | Scope |
+|----------|-----------|-------|
+| review | `workflow-event_name-PR#` | per PR, split by event type |
+| mention (verify) | none | stateless, completes in seconds |
+| mention (handle) | `workflow-handle-issue#` | per issue/PR |
+| triage | `workflow-issue#` | per issue |
+| ci-fix | none | only fires on failure, rare overlap |
+| nightly / renovate | none | scheduled, single instance |
+
+**Why cancel-in-progress:** The cancelled run was processing a stale event. A
+new push invalidates the old review, a new comment supersedes the old mention,
+a re-opened issue supersedes the old triage. The alternative —
+`cancel-in-progress: false` — queues the old run to complete, but it produces
+a response based on outdated context.
+
+**Tradeoff — wasted work:** A handle job that has been running for 20 minutes
+gets killed when a new event arrives. The work is lost. This is acceptable
+because the response it was composing was based on an older state. For mention,
+where each comment is arguably independent, queueing (`cancel-in-progress:
+false`) could preserve work but risks posting stale or duplicate responses —
+the exact problem that motivated adding concurrency control (#61).
+
+**Workflows without concurrency groups:** ci-fix, nightly, and renovate don't
+need them. ci-fix triggers on workflow failure, which is rare enough that
+overlapping runs are unlikely. Nightly and renovate are scheduled with
+`workflow_dispatch` — GitHub already serializes cron-triggered runs, and manual
+dispatches are infrequent.
+
 ## What lives in the tend repo
 
 ```


tend-agent:
No docs on the concurrency strategy — DESIGN.md lists concurrency groups as generator-owned but doesn't explain the approach or tradeoffs.

You're right about the wasted-work concern. The tradeoff:

- **`cancel-in-progress: true`** (current, everywhere): Latest event wins. A 20-minute handle job gets killed if a new comment arrives. Wastes work, but the cancelled run's response would have been stale anyway — it was processing an older comment.
- **`cancel-in-progress: false`** (queue): No wasted work, but the queued run processes an older event. For mention, that means the bot responds to comment A *after* it already responded to comment B, which could be confusing or produce a stale answer.

For review and triage, cancel-in-progress is clearly right — a new push invalidates the old review, a re-opened issue invalidates the old triage. For mention it's more nuanced since each comment is independent, but queuing has its own staleness problem and the duplicate-post risk was the immediate issue (#61).

I'll add a concurrency strategy section to DESIGN.md documenting the approach and tradeoffs across workflows. Want me to do that in this PR or a follow-up?

PR #92, opened 2026-03-28
fix(mention): remove pull_request_review trigger to prevent duplicate replies

### LINKED ISSUE #91: Bot posts duplicate replies to same review comment thread
## Problem

In [worktrunk#1785](https://github.com/max-sixty/worktrunk/pull/1785), the bot posted duplicate replies to the same review comment thread.

### Timeline

The PR author left a review comment ([3005232399](https://github.com/max-sixty/worktrunk/pull/1785#discussion_r3005232399)):
> "but then do we need these? I think we could either remove them or put them in a table"

The bot replied **twice** to this same comment:

1. **19:48:59** ([3005238770](https://github.com/max-sixty/worktrunk/pull/1785#discussion_r3005238770)): "Good call — consolidated them into a table in 2b7c9178..."
2. **19:50:52** ([3005240645](https://github.com/max-sixty/worktrunk/pull/1785#discussion_r3005240645)): "Agreed — collapsed them into the main table in 2b7c9178..."

The second reply is redundant — it references the same commit and says the same thing, just also mentions the fix from the other review thread.

### Likely cause

The bot also had a separate review thread to respond to (its own earlier comment [3005152636](https://github.com/max-sixty/worktrunk/pull/1785#discussion_r3005152636) about TOML scoping). It posted a reply there at 19:50:44, and then 8 seconds later posted the duplicate reply in the author's thread. It seems the bot may have posted responses to both threads after making the second commit (4dd6516e), without recognizing it had already replied to the author's thread 2 minutes earlier after the first commit (2b7c9178).

### Expected behavior

The bot should check whether it has already replied to a review comment thread before posting another reply. If it already addressed the feedback and pushed a fix, it should not post again when handling a subsequent review comment.

@tend-agent please diagnose the root cause — was this a race condition between two review-response workflows, or a single session that posted twice?

tend-agent:
## Problem

When a review with inline comments is submitted on a bot-authored PR, GitHub fires both:
- `pull_request_review` → triggers `tend-mention`
- `pull_request_review_comment` (per inline comment) → triggers `tend-mention`

These run in parallel with different concurrency groups, and both try to reply to the same comments, producing duplicates (as observed in [worktrunk#1785](https://github.com/max-sixty/worktrunk/pull/1785)).

## Solution

Remove the `pull_request_review` trigger from `tend-mention` entirely. Inline review comments (`pull_request_review_comment`) still trigger `tend-mention`, so the bot still responds to specific code feedback. This is simpler than the original approach of adding guard conditions.

## Testing

- Replaced `test_mention_handles_pull_request_review` with `test_mention_no_pull_request_review_trigger` asserting the trigger is absent
- Added `test_mention_review_comment_trigger_present` confirming inline comments still work
- Full test suite passes (113 tests)
- All pre-commit hooks pass

---
Closes #91

### DIFF
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -174,6 +174,8 @@ on:
   issue_comment:
     types: [created, edited]
   # Works for same-repo PRs only; secrets unavailable on fork PRs (no _target variant exists)
+  pull_request_review:
+    types: [submitted]
   pull_request_review_comment:
     types: [created, edited]
 
@@ -186,7 +188,9 @@ jobs:
       (github.event_name == 'issue_comment' &&
         github.event.comment.user.login != '{bn}') ||
       (github.event_name == 'pull_request_review_comment' &&
-        github.event.comment.user.login != '{bn}')
+        github.event.comment.user.login != '{bn}') ||
+      (github.event_name == 'pull_request_review' &&
+        github.event.review.user.login != '{bn}')
     runs-on: ubuntu-24.04
     outputs:
       should_run: ${{{{ steps.check.outputs.should_run }}}}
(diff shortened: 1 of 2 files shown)


max-sixty:
> pull_request_review — responds to reviews submitted on bot-authored PRs
> 

let's fix this problem by removing the trigger

PR #95, opened 2026-03-28
fix(mention): don't cancel verify job on rapid comments

### LINKED ISSUE #93: bot miss at PRQL
What happened here? https://github.com/PRQL/prql/pull/5750#issuecomment-4144849552

Why didn't @prql-bot see this originally?

tend-agent:
## Problem

The `tend-mention` workflow's `verify` job has a concurrency group keyed by PR number with `cancel-in-progress: true`. When two comments arrive on the same PR within seconds, the second run cancels the first — even if the first was an explicit `@bot` mention.

This is exactly what happened at [PRQL/prql#5750](https://github.com/PRQL/prql/pull/5750#issuecomment-4144849552):
1. **19:44:09** — `@prql-bot can you do a follow-up...` → triggered run [23664429862](https://github.com/PRQL/prql/actions/runs/23664429862)
2. **19:44:14** — `thanks @snth !` (5 seconds later) → triggered run [23664431397](https://github.com/PRQL/prql/actions/runs/23664431397)
3. Run 2 **cancelled** run 1 via the concurrency group
4. Run 2 processed "thanks @snth" (no bot mention) → bot exited silently. The explicit `@prql-bot` request was lost.

## Solution

- **Remove concurrency from the `verify` job** — it's lightweight and fast; parallel verify jobs for the same PR are harmless.
- **Change `handle`'s `cancel-in-progress` to `false`** — jobs queue instead of cancelling. The first handle completes (processes the mention), then the second runs (exits silently for non-mention). No mention gets dropped.

## Testing

- Added `test_mention_verify_no_concurrency` — verifies the verify job has no concurrency group
- Added `test_mention_handle_job_queues_not_cancels` — verifies handle uses `cancel-in-progress: false`
- Both tests fail before the fix, pass after
- Full suite: 110 tests pass, all pre-commit lints pass

---
Closes #93 — automated triage

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -434,13 +434,24 @@ This requires three genuine `should_run=true` mentions on the same PR/issue
 within the ~60-second window of handle job startup — uncommon but possible when
 multiple people are active on a PR.
 
-**Mitigation paths** (not yet implemented):
+**Mitigation — conversation-aware loading** (not yet implemented):
 
-- Have the handle job scan recent conversation for any unprocessed `@$bot_name`
-  mentions (not just its own triggering comment) before responding. This would
-  make the bot self-healing: even if job B is displaced, job C would pick up
-  B's mention when it scans the conversation.
-- Use a workflow-level queue or external coordinator (adds complexity).
+The skill should review conversation state when it loads, not just its own
+triggering comment. Specifically it should be aware of two things:
+
+1. **Other jobs may have already responded** to its triggering comment. If the
+   job was queued behind another run, that earlier run might have already handled
+   the mention. The skill should check before duplicating work.
+2. **Other unprocessed mentions may exist.** If a prior job was displaced from
+   the queue, its `@$bot_name` mention was never handled. The skill should scan
+   recent conversation for any unprocessed mentions and handle them — making the
+   system self-healing.
+
+To help the skill judge how likely these scenarios are, the workflow could inject
+the **queue-to-run time delta** (time between job queued and job started) as an
+environment variable. A large delta signals the job was queued behind other runs,
+increasing the likelihood that the conversation state has changed since the
+triggering event.
 
 ## What lives in the tend repo
 


max-sixty:
@tend-agent looks like there was a 500 error in https://github.com/max-sixty/tend/actions/runs/23694484189

So open a new issue with:
- Consider how we can deal with these — maybe a retry? maybe some way of reporting the error, so then at least people aren't left hanging?
  - with form text / an emoji?
  - or one approach could be a repo-level issue which references the comment — then there's a cross-issue reference but it's not like if Anthropic is down, then every issue gets dozens of comments

Then:
- Answer the point above

DESIGN.md
```diff
@@ -341,39 +341,79 @@ us with code execution, not just token minting.
...
+
+GitHub Actions evaluates these before the job starts. A false condition means
+the job is **skipped** — it never runs, never enters a concurrency group, and
+consumes no resources.
+
+| Workflow | Condition | Filters out |
+|----------|-----------|-------------|
+| **review** | `event == pull_request_target && !draft` | Draft PRs |
+| **review** | `event == pull_request_review && pr.user == bot && reviewer != bot && (state != approved \|\| body)` | Reviews on non-bot PRs; bot self-reviews; bare approvals with no body |
+| **mention (verify)** | `event == issues && body contains @bot && author != bot` | Issue edits by the bot itself, edits without a mention |
```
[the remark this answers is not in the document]
tend-agent: Each row maps to one branch of the verify job's `if:` expression ([`workflows.py:212-219`](https://github.com/max-sixty/tend/blob/d641366/generator/src/tend/workflows.py#L212-L219)):

```yaml
if: |
  (github.event_name == 'issues' &&
    contains(github.event.issue.body, '@bot') &&
    github.event.issue.user.login != 'bot') ||
  (github.event_name == 'issue_comment' &&
    github.event.comment.user.login != 'bot') ||
  (github.event_name == 'pull_request_review_comment' &&
    github.event.comment.user.login != 'bot')
```

Breaking it down:

1. **`issues` event** (`on: issues: types: [edited]`): fires when someone edits an issue body. The condition requires `@bot` is in the body AND the editor isn't the bot itself. Purpose: catch someone editing `@bot` into an existing issue, while ignoring the bot's own edits (which would cause a loop).

2. **`issue_comment` event** (`on: issue_comment: types: [created, edited]`): fires on any new or edited comment on issues *or* PRs (GitHub treats PR comments as issue comments). Only condition: comment author isn't the bot. Purpose: let verify check whether the bot should engage (direct mention or prior engagement), while filtering out the bot's own comments to prevent loops.

3. **`pull_request_review_comment` event** (`on: pull_request_review_comment: types: [created, edited]`): fires on inline code review comments. Same filter: comment author isn't the bot. Purpose: same as above, but for inline review threads specifically (these use a different GitHub event than regular PR comments).

The `issues` branch is the strictest — it requires `@bot` in the body at the GHA level. The other two are permissive at the GHA level (just "not the bot's own comment") and delegate the engagement check to the Layer 2 `should_run` script.

I can make the spec table clearer if you think the pseudocode notation is too compressed — e.g. expanding each row to say "fires when X; skips when Y."

tend-agent:
Good question — let me trace through this carefully.

**Can a `should_run: false` handle displace a queued `should_run: true` handle?**

No. When the handle job's `if: needs.verify.outputs.should_run == 'true'` evaluates to false, GitHub Actions **skips** the job entirely. A skipped job never enters the concurrency group — it doesn't run, doesn't queue, and can't displace anything. So a non-mention comment's handle job is invisible to the concurrency system.

**Where displacement *can* happen:**

GitHub Actions concurrency queues have a depth of exactly 1. With `cancel-in-progress: false`, only one job can be pending while another runs. If a third job arrives, it **replaces the pending job** (the pending one is cancelled, the new one takes its place).

This means: if three or more genuine `@bot` mentions arrive on the same PR in rapid succession, the middle ones can be lost. Mention A starts running, mention B queues, mention C arrives and displaces B. This is a real limitation, but it requires three `should_run: true` events within the ~60-second window of the handle job startup — quite unlikely in practice.

**Summary:** The `should_run: false` path is safe — those jobs are skipped before they touch the concurrency group. The only risk is three+ genuine mentions racing, which is a GHA queue-depth-of-1 limitation that would require a fundamentally different approach (e.g., a single job that reads all unprocessed mentions) to solve.

DESIGN.md
```diff
@@ -410,6 +410,38 @@ workflow failure, which is rare enough that overlapping runs are unlikely.
...
+This requires three genuine `should_run=true` mentions on the same PR/issue
+within the ~60-second window of handle job startup — uncommon but possible when
+multiple people are active on a PR.
+
+**Mitigation paths** (not yet implemented):
+
+- Have the handle job scan recent conversation for any unprocessed `@$bot_name`
+  mentions (not just its own triggering comment) before responding. This would
+  make the bot self-healing: even if job B is displaced, job C would pick up
+  B's mention when it scans the conversation.
```
max-sixty: right, this is what we should do — the skill should review the state when it loads, not just the comment it was asked to mention

it should be aware of both:
- other jobs might have responded to _its_ comment, prior to it running
- other comments might need responding

we could inject the time difference between the job being queued and running, as some indication of this likelihood

tend-agent: Agreed — updated the mitigation spec (ae922a1). It now describes two things the skill should check on load:

1. Whether another job already responded to its triggering comment (avoid duplicating work)
2. Whether other `@bot` mentions in the conversation are unprocessed (pick up displaced jobs' work)

Also added the queue-to-run time delta idea — the workflow can inject the delay as an env var so the skill knows how stale its triggering context might be.

PR #104, opened 2026-03-29
feat(mention): conversation-aware loading for queued handle jobs

max-sixty:
The mention handle job's concurrency group has queue depth 1. When a third mention arrives while one is running and one is pending, the pending job is displaced and its mention is lost (#93 documented this as a known limitation).

This adds a prompt preamble that instructs the skill to scan recent conversation before acting — exit silently if the trigger was already handled, pick up any unaddressed comments from displaced jobs. The workflow injects the queue-to-run time delta so the skill can gauge how likely conversation drift is (~40s+ means it was queued).

### DIFF
--- a/.github/workflows/tend-mention.yaml
+++ b/.github/workflows/tend-mention.yaml
@@ -141,12 +141,25 @@ jobs:
 
       - uses: astral-sh/setup-uv@v6
 
+      - name: Compute queue delay
+        id: delay
+        run: |
+          event_epoch=$(date -d "$EVENT_TS" +%s)
+          echo "seconds=$(( $(date +%s) - event_epoch ))" >> "$GITHUB_OUTPUT"
+        env:
+          EVENT_TS: ${{ github.event.comment.created_at || github.event.issue.updated_at }}
+
       - uses: max-sixty/tend@v1
         with:
           github_token: ${{ secrets.BOT_TOKEN }}
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           bot_name: tend-agent
           prompt: >-
+            This job started ${{ steps.delay.outputs.seconds }}s after the
+            triggering event (over ~40s means it was queued). Before acting,
+            check recent comments: exit silently if the bot already responded
+            to the trigger; handle any other unaddressed comments too.
+
             ${{ github.event_name == 'issues'
               && format('An issue was updated with a mention of you ({0}). Read it and respond.', github.event.issue.html_url)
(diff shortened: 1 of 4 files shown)


tend-agent (commented):
`EVENT_TS` doesn't cover `pull_request_review` events — that event has no `comment` or `issue` context, so both `github.event.comment.created_at` and `github.event.issue.updated_at` resolve to empty. GNU `date -d ""` silently returns the current time, so the step won't fail but the delay will always be ~0s regardless of actual queue time. Needs a third fallback like `github.event.review.submitted_at`.

.github/workflows/tend-mention.yaml
```diff
@@ -141,12 +141,25 @@ jobs:
...
+        run: |
+          event_epoch=$(date -d "$EVENT_TS" +%s)
+          echo "seconds=$(( $(date +%s) - event_epoch ))" >> "$GITHUB_OUTPUT"
+        env:
+          EVENT_TS: ${{ github.event.comment.created_at || github.event.issue.updated_at }}
+
       - uses: max-sixty/tend@v1
         with:
           github_token: ${{ secrets.BOT_TOKEN }}
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
```
tend-agent: This misses `pull_request_review` events. That event type has no `comment` or `issue` object, so both expressions resolve to empty. GNU `date -d ""` silently returns the current time, making the delay always ~0s.

```suggestion
          EVENT_TS: ${{ github.event.comment.created_at || github.event.review.submitted_at || github.event.issue.updated_at }}
```

generator/src/tend/workflows.py
```diff
@@ -302,12 +302,25 @@ def generate_mention(cfg: Config) -> GeneratedWorkflow:
...
+        run: |
+          event_epoch=$(date -d "$EVENT_TS" +%s)
+          echo "seconds=$(( $(date +%s) - event_epoch ))" >> "$GITHUB_OUTPUT"
+        env:
+          EVENT_TS: ${{{{ github.event.comment.created_at || github.event.issue.updated_at }}}}
+
       - uses: max-sixty/tend@v1
         with:
           github_token: {bt}
           claude_code_oauth_token: {ct}
```
tend-agent: Same issue — needs `review.submitted_at` for `pull_request_review` events.

```suggestion
          EVENT_TS: ${{{{ github.event.comment.created_at || github.event.review.submitted_at || github.event.issue.updated_at }}}}
```

PR #109, opened 2026-03-29
fix(ci): distinguish structural vs stochastic failures in review-reviewers

tend-agent:
## Summary

- Revert the re-verify-before-conceding guidance added to running-in-ci
  (thin evidence for a stochastic behavior)
- Add "Structural vs. stochastic failures" subsection to review-reviewers
  gates, requiring findings to be classified before proposing guidance changes

Structural failures (deterministic cause, always recurs in the same scenario)
can justify action with fewer occurrences. Stochastic failures (model judgment
calls) need 5+ occurrences to avoid over-constraining behavior based on
one-off lapses.

## Evidence

Same incident as before (PR #104 capitulation), but the response is now
correctly scoped: rather than adding guidance for a single stochastic event,
we raise the evidence bar so future review-reviewers runs distinguish
reproducible mechanical failures from probabilistic model behavior.

## Test plan

- [x] Verify the new subsection renders correctly in the skill markdown
- [ ] Monitor future review-reviewers runs for correct classification

### DIFF
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -58,12 +58,34 @@ Rate the proposed change:
 needs less justification than a new paragraph. Prefer small, targeted fixes over
 broad rewrites.
 
+### Structural vs. stochastic failures
+
+Before applying the gates, classify each failure:
+
+- **Structural**: the failure has a deterministic cause that guidance can
+  prevent — e.g., "the checkout differs between `pull_request_target` and
+  `issue_comment` events, so grepping always finds stale content." These
+  failures will recur every time the same conditions arise. One clear
+  occurrence is sufficient evidence for a targeted fix.
+
+- **Stochastic**: the failure is a probabilistic model behavior — e.g., "the
+  model was too agreeable when challenged" or "the model forgot to check X."
+  The same model might handle the next identical situation correctly without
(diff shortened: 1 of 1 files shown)


max-sixty:
this needs much more evidence, given the stochasticity

what could we change in our guidance to make this clear?

tend-agent:
Fair point — one occurrence is thin evidence for a guidance change, especially with stochastic behavior. The same model might handle the next challenge correctly without any new guidance.

Two things we could tighten in the review-reviewers skill to raise the bar:

1. **Require a reproduction threshold** — e.g., "a pattern must appear in at least 3 independent sessions before proposing a guidance change." One-offs could be logged as observations without a PR.

2. **Distinguish structural from stochastic** — if a failure has a clear *structural* cause (like "the checkout differs between events, so grepping will always find stale content"), that's more actionable than "the model was too agreeable this time." The current gates check occurrence count but don't distinguish these.

Happy to close this PR if you'd prefer to wait for more evidence, or I can update the review-reviewers guidance to encode the higher evidence bar instead.

max-sixty:
yes, replace this with #2

tend-agent:
Done — replaced the re-verify-before-conceding guidance with a structural vs. stochastic classification step in the review-reviewers gates. Structural failures (deterministic cause, always recurs) can be acted on quickly; stochastic ones (model behavior, might self-correct) need 5+ occurrences. This should prevent guidance bloat from one-off model lapses.

PR #126, opened 2026-03-31
fix(review): propose project PRs for missing CI deps instead of ad-hoc installs

tend-agent:
## Summary

- When the bot encounters a CI failure caused by a missing tool or dependency, instead of installing it ad-hoc (wasting CI time) or being blanket-prohibited from installing anything, it should **propose a PR to the project** adding the dependency consistent with the project's existing conventions.

## Evidence

**Finding:** Bot installs `cargo-llvm-cov` in CI to investigate codecov/patch failures, wasting CI minutes and sometimes causing session cancellation from compilation timeouts.

**Gate assessment:**
- **Evidence level:** Medium (stochastic — bot sometimes analyzes the diff correctly, sometimes improvises tool installation)
- **Occurrences:** 7 total (5 historical in tracking issue #12 + 2 this hour on runs [23779254077](https://github.com/max-sixty/worktrunk/actions/runs/23779254077) and [23779671239](https://github.com/max-sixty/worktrunk/actions/runs/23779671239), both reviewing PR max-sixty/worktrunk#1830)
- **Classification:** Stochastic (5+ threshold met)
- **Change type:** Targeted fix (one sentence addition to existing step)
- **Both gates pass**

## Test plan

- [ ] Verify the review skill still reads coherently at Step 6
- [ ] Monitor next codecov failure review to confirm bot proposes a project PR rather than installing tools ad-hoc

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #129, opened 2026-03-31
feat: add review-runs workflow for repo-local CI feedback loop

tend-agent:
## Summary

- Adds a `review-runs` scheduled workflow (daily at 7:47 UTC) that runs in adopter repos, reviews the previous night's CI session logs, and proposes improvements to repo-local skills and workflow config
- This is the repo-local counterpart to tend's `review-reviewers` hourly — instead of improving tend's skills centrally, it improves each adopter repo's own `.claude/skills/`, `.config/tend.toml`, and `CLAUDE.md`
- Includes the skill (`SKILL.md`), generator support (`config.py`, `workflows.py`), and updated tests

Closes the direction discussed in #126 — rather than patching individual cases in tend's review skill, each repo gets its own feedback loop.

## Test plan

- [x] All 112 generator tests pass (counts updated from 6→7 workflows)
- [ ] Deploy to an adopter repo via `uvx tend init` and verify `tend-review-runs.yaml` is generated
- [ ] Monitor first run to confirm session log analysis and skill improvement proposals


tend-agent:
Fixed — replaced the `list-recent-runs.sh` call (1-hour window) with the 12-hour query directly, so the daily workflow actually finds overnight runs.

max-sixty:
how much of this is shared with tend's internal reviews? if a lot then make a shared file which is symlinked into each skill and then @ included

PR #155, opened 2026-04-06
fix(skills): default to not closing issues, defer to repo-specific guidance

tend-agent:
## Summary

Reworked approach: instead of adding special exceptions for specific actions, establish the general principle that repo-specific guidance always takes precedence over tend's defaults.

- **Triage skill**: Reverted to the original default — duplicates are flagged but left for a maintainer. Repos that want auto-closing authorize it in their `running-tend` skill.
- **Running-in-ci**: Removed the triage-specific carve-out. Added a general principle: repo-specific guidance overrides these defaults.
- **Review-reviewers**: Added guidance to check repo-specific skills before flagging an action as a rule violation — so we don't incorrectly treat authorized behavior as a problem.

## Test plan

- [ ] Triage on a repo without closing guidance → leaves duplicates open
- [ ] Triage on a repo whose `running-tend` authorizes closing → closes duplicates
- [ ] Review-reviewers checks repo-specific guidance before flagging actions

### DIFF
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -11,6 +11,21 @@ metadata:
 Analyze Claude-powered CI runs from the past hour. Identify behavioral problems, skill gaps, and
 workflow issues — then create PRs or issues to fix them.
 
+## Core principle: repo-specific guidance is primary
+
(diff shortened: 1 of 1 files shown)


max-sixty:
Well, we want to:
- have a default, which is to not close issues
- defer to repo-specific guidance, always

so:
- our object-level guidance should reflect that (which it may do already in the existing code?)
- our reviews of tend should definitely reflect that — we should have seen that the behavior was as requested in worktrunk's skill and evaluated it as good

Issue #160, opened 2026-04-07
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-04-07T15:14:13Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/24088967903) | N/A |

This issue was created automatically. Close it once the outage is resolved.

PR #163, opened 2026-04-07
fix(action): prevent duplicate outage tracking issues

### LINKED ISSUE #162: Bot temporarily unavailable

| When | Run | Trigger |
|------|-----|---------|
| 2026-04-07T15:14:17Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/24088967903) | N/A |

tend-agent:
## Summary

- Fix race condition where multiple jobs failing simultaneously create separate `tend-outage` tracking issues (#160, #161, #162 were all duplicates from the same run)
- Minimize the TOCTOU window by preparing the issue body *before* checking for an existing issue, so the check-then-create gap is just one API call
- Removes the jitter and post-create self-cleanup from the earlier revision — simpler approach, small residual race window is acceptable

## Test plan

- [x] Verify the logic — check and create are now back-to-back with no unnecessary work between them
- [ ] Next time multiple jobs fail simultaneously, confirm only one tracking issue remains open

Closes #162

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -182,36 +182,24 @@ runs:
 
         TIMESTAMP=$(date -u +%Y-%m-%dT%H:%M:%SZ)
 
-        # Random jitter (0-5s) to reduce races when multiple jobs fail simultaneously
-        sleep $((RANDOM % 6))
+        # Prepare issue body before the check so the TOCTOU window is minimal
+        gh label create "$LABEL" --description "Tracks bot outage incidents" --color "d93f0b" 2>/dev/null || true
+        printf '%s\n\n%s\n%s\n%s\n\n%s\n' \
+          "The bot failed to process a request. This issue tracks failures until the underlying cause is resolved." \
+          "| When | Run | Trigger |" \
+          "|------|-----|---------|" \
+          "| ${TIMESTAMP} | [workflow run](${RUN_URL}) | ${REF:-N/A} |" \
+          "This issue was created automatically. Close it once the outage is resolved." > /tmp/body.md
 
-        # Find existing open tracking issue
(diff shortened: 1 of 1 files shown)


action.yaml
```diff
@@ -200,6 +203,15 @@ runs:
             "| ${TIMESTAMP} | [workflow run](${RUN_URL}) | ${REF:-N/A} |" \
             "This issue was created automatically. Close it once the outage is resolved." > /tmp/body.md
           gh issue create --title "$TITLE" --label "$LABEL" -F /tmp/body.md
+
+          # Re-check: if another job raced us and created a duplicate, close ours
+          sleep 2
+          DUPES=$(gh issue list --label "$LABEL" --state open --json number --jq '[.[].number] | sort')
+          FIRST=$(echo "$DUPES" | jq '.[0]')
+          MINE=$(gh issue list --label "$LABEL" --state open --author "@me" --json number --jq '[.[].number] | sort | last')
```
tend-agent: All concurrent jobs run as the same bot user, so `--author "@me" | sort | last` identifies the same issue for every job — only the highest-numbered duplicate gets closed; middle ones survive.

Capture the number from `gh issue create` instead:

```suggestion
          MINE_URL=$(gh issue create --title "$TITLE" --label "$LABEL" -F /tmp/body.md)
          MINE=${MINE_URL##*/}

          # Re-check: if another job raced us and created a duplicate, close ours
          sleep 2
          FIRST=$(gh issue list --label "$LABEL" --state open --json number --jq '[.[].number] | sort_by(.) | .[0]')
          if [ -n "$MINE" ] && [ "$MINE" != "$FIRST" ]; then
```

PR #212, opened 2026-04-10
fix(review): skip APPROVE when incremental commits are bot-authored

tend-agent:
## Summary

- Add an explicit pre-flight check to the `/tend-ci-runner:review` skill: when every commit since `LAST_REVIEW_SHA` was authored by the bot itself, do not submit a new APPROVE — leave any existing approval in place.
- Guards against the re-approval loop where a bot-pushed fix (from `tend-notifications`, `tend-ci-fix`, or a mention run) triggers `tend-review`, which then ratifies its own commit.

### Incident: [max-sixty/worktrunk#2041](https://github.com/max-sixty/worktrunk/pull/2041) (merged)

Timeline on the `logs` branch (PR by `max-sixty`):

| Time | Commit | Actor | Review result |
|---|---|---|---|
| 23:10:40Z | d4e3e8b | worktrunk-bot review #1 | COMMENTED — flagged 2 real issues |
| 23:17:58Z | d6456bc | worktrunk-bot review #2 | COMMENTED — acknowledged trash-sweep fix |
| 23:30:31Z | f96b126 | worktrunk-bot review #3 | **APPROVED** (empty body) |
| 23:32:17Z | **c30d804** | **worktrunk-bot push** (fix from `tend-notifications` [run 24218351323](https://github.com/max-sixty/worktrunk/actions/runs/24218351323)) | — |
| 23:36:57Z | c30d804 | worktrunk-bot review #4 | **APPROVED** (empty body) — **self-review of own commit** |

[tend-review run 24218513196](https://github.com/max-sixty/worktrunk/actions/runs/24218513196) session transcript shows the bot reasoning:

> The bot already approved at `f96b1260`. The new commit is trivial — just one line change in `state.rs` using `to_slash_lossy()` for Windows path consistency. Let me verify the change is correct, check threads, and **re-approve**.

The skill's existing trivial-change branch already says _"Otherwise do not submit a new review — the existing one stands"_, but the bot re-approved anyway. This ended with the bot holding 4 reviews on the PR, the last of which approved its own commit.

### Related prior attempt: #154 (closed)

[PR #154](https://github.com/max-sixty/tend/pull/154) tried to fix a similar pattern on [PRQL/prql#5774](https://github.com/PRQL/prql/pull/5774) (5 redundant approvals) by skipping the review entirely. That was rejected — a review IS still needed, because bot-pushed commits can still introduce issues worth catching. This PR takes the narrower approach suggested in that thread: still review, just do not re-APPROVE.

## Gate assessment

[...]

PR #217, opened 2026-04-10
fix(ci-runner): replace if-not polling loop with || to avoid bash escaping

tend-agent:
## Summary

The CI polling loop in `running-in-ci` and `review` skills uses `if ! cmd`, but the Claude Code Bash tool escapes `!` as `\!` before handing commands to bash. Bash then parses the escaped token as a literal command (`!: command not found`), so the early-exit branch of `if ! pipeline` is unreachable. Every poll runs all 10 iterations and prints a false "CI still running after 10 minutes".

Switched both loops to `cmd || { ...; }`, which doesn't rely on `!`.

## Evidence

Reproduced in the bash tool this run:

```
$ echo 'if ! false; then echo then; else echo else; fi' > /tmp/t.sh
$ cat /tmp/t.sh
if \! false; then echo then; else echo else; fi
$ bash /tmp/t.sh
/tmp/t.sh: line 1: !: command not found
else
```

And similarly `printf '%s\n' '!test'` writes `\!test` — the escape happens even inside single quotes, so there is no source-level workaround.

Observed in a real polling run — tend-mention PR #212, run [24220018870](https://github.com/max-sixty/tend/actions/runs/24220018870). The background task output was:

```
CI still running after 10 minutes
no required checks reported on the 'hourly/review-24219300860' branch
```

With a working `!`, the first iteration's pipeline exits 1 (grep -q no match), the `!` flips it to 0, and the loop exits immediately. Instead, the loop ran to completion — consistent with `if ! pipeline` always evaluating to the else branch (because `!` is treated as an unknown command, exit 127, which is non-zero, which `if` sees as false).

The bot on that run attributed the timeout to "no required checks" but that explanation doesn't hold — any polling of any PR in this harness suffers the same fate.

## Fix

Replace

```bash
if ! gh pr checks <n> --required 2>&1 | grep -v ... | grep -q 'pending\|...'; then
  gh pr checks <n> --required
  exit 0
fi
```

with

```bash
gh pr checks <n> --required 2>&1 | grep -v ... | grep -q 'pending\|...' || {
  gh pr checks <n> --required
  exit 0
}
```

`||` is a distinct operator (not `!`), so the escape doesn't apply. Verified locally that the rewritten loop exits on iteration 1 when the pipeline returns non-zero.

## Gate assessment

- **Confidence**: Critical. Every polling run hits this; reproduced in-session and observed in a real run. One occurrence is sufficient per the gates.
- **Classification**: Structural. The same conditions will recur on every polling invocation.
- **Magnitude**: Targeted fix — 6 lines changed across 2 skill files, no behavior change on the happy path.

## Test plan

- [x] `cd generator && uv run pytest` (147 passed)
- [ ] Next tend-review run on this branch polls CI and exits cleanly within one iteration instead of timing out at 10 minutes

### DIFF
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -290,14 +290,17 @@ individual sleep commands. Use the Bash tool with `run_in_background: true` and 
 you're running. **NEVER use `--watch` flags** — they hang forever.
 
 ```bash
-# Run with Bash tool's run_in_background: true
+# Run with Bash tool's run_in_background: true.
+# Use `||` rather than `if !` — the Bash tool escapes `!` as `\!`, which
+# prevents bash from recognizing the pipeline-negation reserved word and leaves
+# the loop stuck until the 10-minute timeout.
 for i in $(seq 1 10); do
   sleep 60
-  if ! gh pr checks <number> --required 2>&1 \
-       | grep -v "tend-review" | grep -q 'pending\|queued\|in_progress'; then
+  gh pr checks <number> --required 2>&1 \
+       | grep -v "tend-review" | grep -q 'pending\|queued\|in_progress' || {
     gh pr checks <number> --required
     exit 0
-  fi
+  }
 done
 echo "CI still running after 10 minutes"
 exit 1
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -139,12 +139,15 @@ background task completes you will be notified — check the result and take any
 # Run with Bash tool's run_in_background: true
 # Filter out the current workflow ($GITHUB_WORKFLOW) — it will always show as
 # "pending" since it IS the running job. Watching yourself deadlocks.
+# Use `||` rather than `if !` — the Bash tool escapes `!` as `\!`, which
+# prevents bash from recognizing the pipeline-negation reserved word and leaves
+# the loop stuck until the 10-minute timeout.
 for i in $(seq 1 10); do
   sleep 60
-  if ! gh pr checks <number> --required 2>&1 | grep -v "$GITHUB_WORKFLOW" | grep -q 'pending\|queued\|in_progress'; then
+  gh pr checks <number> --required 2>&1 | grep -v "$GITHUB_WORKFLOW" | grep -q 'pending\|queued\|in_progress' || {
     gh pr checks <number> --required
     exit 0
-  fi
+  }
 done
 echo "CI still running after 10 minutes"
 exit 1


max-sixty:
I thought this was fixed in newer Claude versions; what version is running?

tend-agent:
Claude Code 2.1.98. The bug still reproduces — just verified in this session. Running `if \! false; then echo then; else echo else; fi` directly via the Bash tool (no file, no quoting) still produces `/bin/bash: line 1: \!: command not found` followed by `else`. The escape is applied to the command string before bash sees it, so there's no source-level workaround.

max-sixty:
> Running if \! false; then echo then; else echo else; fi directly via the Bash tool

was the escape in the command that got sent to the bash tool??

tend-agent:
No — I sent a plain `\!` in both commands. In this session just now I ran exactly `if \! false; then echo then; else echo else; fi` (no backslash) and bash still reported `/bin/bash: line 1: \!: command not found` then `else`. So the `\` is added somewhere between the tool call and bash `exec`.

Concrete demonstration — I ran `printf 'arg1=[%s]\n' "$1" -- "\!test"` (unescaped `\!` in the source argument) and got:

```
arg1=[]
arg1=[--]
arg1=[\\!test]
```

The literal backslash arrives inside an already-quoted argument, so this isn't bash doing it on its own. Also confirmed `set -o | grep histexpand` prints `histexpand off`, so it's not interactive history expansion either — bash wouldn't escape `\!` in a non-interactive shell regardless.

The tool layer is adding the backslash before handing the string to bash. It's not something I can avoid by changing how I write the command, which is why the skill-level fix (rewriting `if \! pipeline` as `pipeline || { ... }`) is the right workaround.

max-sixty:
> No — I sent a plain \! in both commands. In this session just now I ran exactly if \! false; then echo then; else echo else; fi (no backslash) and bash still reported /bin/bash: line 1: \!: command not found then else. So the \ is added somewhere between the tool call and bash exec.

all these have slashes in them! is that because they are getting added even when you try to write out this comment? are you trying to say:
1. "In this session just now I ran exactly `if ! false;`"
1. or "In this session just now I ran exactly `if \! false;`"

tend-agent:
Sorry, that was confusing. Answer: **option 1** — in my Bash tool call I wrote `if ! false; ...` with no backslash. The backslash is added somewhere between my tool call and bash `exec`, which is exactly the bug.

Just reproduced it deterministically in this session. I ran a tiny script that prints the raw bytes of its first argument, passing `"if ! false; then echo then; else echo else; fi"` (no backslash in my source):

```
ARG: [if \! false; then echo then; else echo else; fi]
0000000   i   f       \   !       f   a   l   s   e   ;       t   h   e
```

The argument as received has a literal `\` before `!` (byte `\134` then `!`), even though I wrote no backslash. So bash sees `\!`, treats it as the command `!` (the backslash quotes the next char), gets exit 127, and `if` takes the else branch.

And to rule out the Write tool: I wrote a file with plain `if ! false` via Write, then ran `od -c` on it — no backslash in the bytes. So the escape is specific to the Bash tool, not a universal `!` mangling.

The `\!` that appeared in my earlier comments was me quoting what bash received (the escaped form from tool output), not what I typed. That was a bad choice — it made the claim look self-contradictory. Sorry for the noise.

max-sixty:
is there a Claude Code issue for this? I had thought it was fixed; alas

tend-agent:
Yes — anthropics/claude-code#35701, filed 2026-03-18, still open. Exactly this: `shell-quote` escapes `\!` to `\\!` on the double-quote wrapping path, and since commands run under `bash -c` (non-interactive, no histexpand), the backslash becomes literal.

max-sixty:
OK, I'll merge this, and hope they fix that issue

PR #223, opened 2026-04-10
fix(workflows): drop hardcoded 60-min job timeout

max-sixty:
A `tend-mention handle` run in max-sixty/worktrunk hit 58m51s against the 60-min cap doing real multi-step work — 4 commits across a shell-integration fix, docs, codecov coverage gaps, and a snapshot refresh, with ~24 min spent waiting on slow `codecov/patch`. A clean near-miss on legitimate work, not a hang.

The 60-minute constant had no principled basis. Rather than tuning the magic number to 120 (or 180, or 240), this drops the line entirely and falls back to GitHub's 360-min default. That removes a knob instead of adding one, which is the right direction given the repo's "no new options" rule.

Trade-off: worst-case runaway exposure goes from 1h to 6h per stuck job. The motivating incident was a near-miss on real work, not a runaway, so this seems acceptable — and a stuck job is bounded and observable either way.

Per-workflow override is a separate question to consider: there's currently no pass-through mechanism in `WorkflowConfig` for arbitrary job-level YAML keys. If we want one, that's its own design discussion.

### DIFF
--- a/.github/workflows/tend-ci-fix.yaml
+++ b/.github/workflows/tend-ci-fix.yaml
@@ -17,7 +17,6 @@ jobs:
   fix-ci:
     if: github.event.workflow_run.conclusion == 'failure'
     runs-on: ubuntu-24.04
-    timeout-minutes: 60
     permissions:
       contents: write
       pull-requests: write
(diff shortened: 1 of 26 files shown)


tend-agent (approved):
One drive-by observation, not blocking: [`.github/workflows/review-reviewers.yaml`](https://github.com/max-sixty/tend/blob/b62036ef47b26ca5fe046acc236d00b7ee024ccf/.github/workflows/review-reviewers.yaml#L25) still sets `timeout-minutes: 60`. It's hand-maintained (not generated) so it's legitimately out of this PR's scope, but it's a bot workflow running Claude against a matrix of adopter repos — the same rationale that motivated this change applies. Happy to push a follow-up commit to drop it there too if you'd like consistency across all bot workflows in the repo.

PR #226, opened 2026-04-10
fix(ci-runner): filter CI polling by run URL, not workflow name

tend-agent:
## Summary

The tend-review CI polling loop has been timing out after 10 minutes on every run, even when all required checks have actually passed. This PR fixes the root cause: the filter used to exclude the current workflow's own pending check was structurally wrong.

## Root cause

Both skills try to filter the current workflow's check row out of `gh pr checks` output. The review skill hardcoded `grep -v "tend-review"` and running-in-ci used `grep -v "$GITHUB_WORKFLOW"`. Both fail:

- `gh pr checks` prints the **job name** in the first column, not the workflow name.
- For tend-review, the workflow is named `tend-review` but the single job is named `review` (see [`tend-review.yaml`](https://github.com/max-sixty/worktrunk/blob/main/.github/workflows/tend-review.yaml) job id `review`, no `name:` override).
- `grep -v "tend-review"` against `review\tpending\t...\t.../runs/<id>/job/...` matches no lines — the filter keeps the own-job row, the row is `pending`, and the loop never short-circuits.

Verified structurally on live worktrunk data:

```
$ GITHUB_RUN_ID=24260154110
$ gh pr checks 2062 -R max-sixty/worktrunk | grep "tend-review"
# (empty — nothing to filter)
$ gh pr checks 2062 -R max-sixty/worktrunk | grep "/runs/$GITHUB_RUN_ID/"
review	pass	16m28s	https://github.com/max-sixty/worktrunk/actions/runs/24260154110/job/70841646383
```

## Observed impact (this hour)

Two tend-review runs on max-sixty/worktrunk timed out after the full 10-minute polling window:

- [run 24260154110](https://github.com/max-sixty/worktrunk/actions/runs/24260154110) — PR #2062: "The polling filter looked for `tend-review` but the actual workflow name is just `review`, so it didn't exclude it and timed out."
- [run 24260939035](https://github.com/max-sixty/worktrunk/actions/runs/24260939035) — PR #2063: same pattern — "the only remaining 'pending' is the `review` workflow itself… the filter didn't catch it."

Both runs recovered by manually polling individual test names, so PR outcomes were correct. But every tend-review session is wasting 10 minutes of compute on a broken polling loop.

## The fix

Filter on the **run URL** (`/runs/$GITHUB_RUN_ID/`) instead of the workflow or check name. Every GitHub Actions check row includes that URL, `$GITHUB_RUN_ID` is always set, and non-Actions checks (codecov, pre-commit.ci) have different URL hosts so they pass through unchanged.

Also incorporates the `if \!` → `cmd || {}` fix from #217. The Claude Bash tool escapes `\!` as `\\!`, so `if \! cmd` always takes the else branch and the loop never short-circuits regardless of filter correctness. Verified in this session:

```
$ if \! false; then echo then; else echo else; fi
/bin/bash: line 1: \!: command not found
else
```

**Both fixes are required together** — with a correct filter but broken `if \!`, the loop still runs 10 iterations; with fixed `if \!` but a broken filter, the loop still runs 10 iterations. This PR supersedes #217 by combining both fixes.

## Gate assessment

- **Evidence level**: Critical (structural failure, wrong outcome — 10 min wasted per tend-review run)
- **Occurrences this run**: 2 (structural; recurs deterministically on every tend-review run against a PR with its own `tend-review` workflow check)
- **Historical**: Noted as a stochastic stylistic nitpick in review-runs-tracking (PR #136) but misclassified as "functionally correct" — the true bug was masked by #217's bash-escaping issue
- **Classification**: Structural (deterministic — same filter always fails against same check-name convention)
- **Change type**: Targeted fix (one-line filter replacement + bash-escape fix in two skill files)
- **Passes both gates**: Yes

## Test plan

- [x] Verified the broken filter keeps all check rows on live worktrunk PR #2062 data
- [x] Verified `/runs/$GITHUB_RUN_ID/` correctly matches only the current-run row
- [x] Verified `cmd || {}` pattern works under Bash tool's `\!` escaping
- [x] Tested three cases: nothing pending → exit 0; only own run pending → exit 0; other pending → keep waiting


max-sixty:
@tend-agent resolve conflicts

tend-agent:
Merged `origin/main` — two conflicts in `plugins/tend-ci-runner/skills/review/SKILL.md` and `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md` (both in the CI polling-loop comment block). Resolved by keeping this branch's wording, since the URL-filter fix is the point of this PR. CI: `lint` and `test` green on `36515a8`.

tend-agent (commented):
The polling-filter fix itself is technically sound. I verified `gh pr checks 226` on this PR's own data:

```
review	pending	0	https://github.com/max-sixty/tend/actions/runs/24276481328/job/70891194230	
lint	pass	19s	...
test	pass	7s	...
```

First column is the job name (`review`), not the workflow name (`tend-review`), so `grep -v "tend-review"` and `grep -v "$GITHUB_WORKFLOW"` both kept the own-job row and the loop deadlocked. Filtering on `/runs/$GITHUB_RUN_ID/` correctly targets that row, and non-Actions checks (codecov, pre-commit.ci) use different URL hosts so they pass through.

## Concern: unrelated content silently dropped in merge resolution

The merge commit [`98e227d`](https://github.com/max-sixty/tend/pull/226/commits/98e227db9ee1e9575ebb856146f423b7b6558d53) ("Merge remote-tracking branch 'origin/main' … # Conflicts: running-in-ci/SKILL.md") dropped four chunks from `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md` that have nothing to do with the polling fix and are not mentioned in the PR title, description, or commit message. They are still present on `main`:

1. **Triage skill loading instruction** (between "auto-discovered." and "## Conduct"):
   ```
   If you are going to propose a code fix for a bug, load `/tend-ci-runner:triage` first — it
   contains reproduction and testing gates that apply to all fix attempts, not just initial triage.
   ```

2. **"Review events with inline comments" bullet** (in the "Replying to Comments" section, between the inline-review-comment bullet and the conversation-comments bullet):
   ```
   - **Review events with inline comments** (review ID in prompt): A review may include inline
     comments. Fetch them by review ID and reply to each individually:
     ```bash
     gh api repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments \
       --jq '.[] | {id: .id, path: .path, body: .body}'
     ```
     Reply to each comment using the inline review comment reply endpoint above.
   ```

3. **Acknowledge-resolution guidance** (in "Multi-way Conversations", before "If a maintainer has already addressed…"):
   ```
   A comment that responds to concerns you raised in a review is directed at you — briefly
   acknowledge resolution or explain why concerns remain.
   ```

4. **Entire `## Self-conversation Guard` section** (between "Multi-way Conversations" and "## Recheck Before Posting"):
   ```
   ## Self-conversation Guard

   If you are responding to your own prior comment or review (not a human's reply to it), only
   respond if there is a distinct role boundary (e.g., you are the reviewer on your own PR and need
   to address review feedback). If there is no such role distinction, exit silently to avoid
   self-conversation loops.
   ```

These are all behavioral guardrails — particularly the self-conversation guard, which is load-bearing for self-authored review runs like this one. Pushing a fix to restore them now.

## Note on PR description

The description says "This PR supersedes #217" but [#217](https://github.com/max-sixty/tend/pull/217) is already merged. The two fixes coexist on main now; the wording is a leftover from drafting against an earlier base.

PR #239, opened 2026-04-11
fix(running-in-ci): warn against checkout --ours <file> after merge conflict

tend-agent:
## Summary

`git checkout --ours/--theirs <file>` replaces the *entire* file with one side's blob, silently dropping any non-conflicting changes that git already auto-merged from the other side. The result looks resolved (no conflict markers, `git status` clean) but quietly reverts unrelated content. Add a fourth bullet to the "Merging Upstream into PR Branches" section in `running-in-ci/SKILL.md` warning against this and pointing at the two correct alternatives (in-place edit of conflict markers, or `git merge -X ours` at merge time which is per-hunk).

## Evidence

Run [24276411750](https://github.com/max-sixty/tend/actions/runs/24276411750) (`tend-mention`, `issue_comment` on max-sixty/tend#226 at 06:13:24Z). The bot was asked by `@max-sixty` to resolve conflicts. Trace:

1. `git merge origin/main` → conflict in `plugins/tend-ci-runner/skills/review/SKILL.md` and `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md` (both inside the CI polling-loop comment block).
2. `git checkout --ours plugins/tend-ci-runner/skills/{review,running-in-ci}/SKILL.md` → "resolved".
3. `git commit` → merge commit `36515a8` pushed to `hourly/review-24261599977`.

The bot's confirmation comment said it had "kept this branch's version since the URL-filter fix is the point of this PR" — matching the *intended* policy for the conflict region. But the `--ours` checkout overwrote the *whole* file, silently dropping four chunks that `main` had added since the PR was opened (none of them in conflict with PR 226's polling-block change):

1. The "load `/tend-ci-runner:triage` before proposing a code fix" instruction (from #235).
2. The "Review events with inline comments" reply pattern in the Replying to Comments section.
3. The "responds to concerns you raised" acknowledge-resolution paragraph in Multi-way Conversations.
4. The entire `## Self-conversation Guard` section.

Verified directly:

```bash
$ git show 2f92a88:plugins/tend-ci-runner/skills/running-in-ci/SKILL.md | grep -c "Self-conversation Guard"
1
$ git show 36515a8:plugins/tend-ci-runner/skills/running-in-ci/SKILL.md | grep -c "Self-conversation Guard"
0
```

`2f92a88` is the second parent of merge commit `36515a8`. The chunks are present on the `main` side but not in the merge result — confirming the loss is post-merge, not pre-existing in the PR branch.

The follow-up [`tend-review` run 24276481328](https://github.com/max-sixty/tend/actions/runs/24276481328) caught this (correctly, with the right diagnosis and a clear restoration commit `e2523da`). It got cancelled by concurrency before posting its review, but the restoration commit had already pushed. So in this specific case the data loss was caught and undone — but the underlying footgun remains, and it produces silent regressions whenever a future review run doesn't happen to look closely enough.

## Gate assessment

- **Confidence**: Critical — verified silent content loss from a single command. Single occurrence is sufficient at the Critical level per the gates.
- **Type**: Structural — `git checkout --ours <file>` deterministically discards auto-merged "theirs" content whenever the file has changes outside the conflict region. Will recur every time the bot reaches for that command.
- **Magnitude**: Targeted fix — one new bullet in an existing section, no restructuring.
- **Both gates**: pass.

## Test plan

- [ ] CI green (lint + test)

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -130,6 +130,16 @@ When asked to merge the default branch into a PR branch:
    If this fails, the branch history is disconnected. Re-checkout the PR with full history
    (`fetch-depth: 0`) before retrying.
 
+4. **Resolve content conflicts in place — never `git checkout --ours/--theirs <file>`.** Those
+   commands replace the *entire* file with one side's blob, silently dropping any non-conflicting
+   changes from the other side that git already auto-merged into the file. The result looks clean
+   (no conflict markers, `git status` shows the file resolved) but you've quietly reverted unrelated
+   improvements from `main` — and the loss is invisible until someone diffs against `main` later.
+   To keep "ours" *only* for the conflicted hunk while preserving auto-merged content, edit the
+   conflict markers in place. If you genuinely want a per-hunk "favor ours" policy across the whole
+   merge, pass `-X ours` at merge time (`git merge -X ours origin/main`) — it resolves conflicting
+   hunks in favor of HEAD without discarding auto-merged content.
+
 ## CI Monitoring
 
 After pushing, wait for CI before reporting completion.


max-sixty:
this is not structural! how do we need to change our guidance on structural vs stochastic to make that clear?

tend-agent:
You're right — this is stochastic. The replay test gets there immediately: replay the same "resolve this merge conflict" scenario 10 times and the bot won't always reach for `git checkout --ours <file>`. It might edit the conflict markers in place, or use `git merge -X ours`. The `--ours` tool is deterministically lossy *once invoked*, but the bot's *choice* to invoke it is a model behavior. I latched onto the tool mechanics.

What I think tripped me up in the current guidance: the phrase "the failure has a deterministic cause that guidance can prevent." I read "deterministic cause" as "the tool deterministically produces the bad outcome." The intended reading is "the failure recurs every time under the same conditions, regardless of how the bot approaches the task." Those are different, and the existing example shows the intended reading — "the checkout differs between `pull_request_target` and `issue_comment` events" is structural because there's no decision point for the bot; the environment is the failure.

Two directions I'd propose for [review-gates.md](https://github.com/max-sixty/tend/blob/main/plugins/tend-ci-runner/shared/review-gates.md):

1. **Add a decision-point discriminator to the definitions.** Something like: *"Did the bot have a choice? If it picked one tool from a set of valid-looking alternatives and happened to pick a lossy one, that's stochastic — even if the tool is deterministic once invoked. A different session may pick differently under identical conditions. Structural failures have no such decision point — the bot does the obvious thing and the environment produces the wrong outcome."*

2. **Require the replay-test answer explicitly in every finding.** Today "Applying the gates" asks for the classification but not the reasoning. If the checklist required *"replay answer: under identical conditions, would this recur every time, or only sometimes,"* I would have had to write "the bot always picks `--ours`" and noticed that's false. The test is already present in the file — just unused as a forcing function.

Happy to push either (or both) as a follow-up. On this PR: by the corrected classification the `--ours` warning is a single stochastic occurrence, below the 5+ threshold — want me to close and move to `review-reviewers-tracking`, or leave it since it's already written?

max-sixty:
do 30% of those suggestions

tend-agent:
Done: #241. Took ~30% as the definitional clarification (suggestion 1) without the new checklist item (suggestion 2) — the misreading lived in the word "deterministic cause," so the minimal effective fix is just rewording Structural vs. Stochastic around a decision-point discriminator. Left this PR open since you didn't answer the close-or-leave question; happy to close it if the reclassification means it shouldn't land.

max-sixty:
> Done: #241. Took ~30% as the definitional clarification (suggestion 1) without the new checklist item (suggestion 2) — the misreading lived in the word "deterministic cause," so the minimal effective fix is just rewording Structural vs. Stochastic around a decision-point discriminator. Left this PR open since you didn't answer the close-or-leave question; happy to close it if the reclassification means it shouldn't land.
> 

sorry, I meant: do 30% of the changes suggested re deterministic vs stochasic above

remove all object-level changes

tend-agent:
Got it, thanks for the clarification — "30% of the changes" meant a smaller slice, not "full suggestion 1, drop suggestion 2," and "object-level" meant the `--ours`-specific references inside the review-gates reframe (not just PR #239 itself).

Opened #242 to trim #241 accordingly. Keeps the decision-point question header and the trimmed structural definition (the minimal fix for the "deterministic cause" misreading); reverts the stochastic paragraph to its pre-#241 wording, which drops both object-level references — the "deterministically lossy *once invoked*" parenthetical and the "reached for a lossy command when a safe alternative existed" example. Net: 10 additions, 14 deletions vs. current main, roughly a third of #241's footprint.

PR #247, opened 2026-04-11
skills(review-runs): write summary to step summary

max-sixty:
The review-runs analysis was only visible in raw job logs — you had to expand the step output and scroll through to see the summary. Write it to `$GITHUB_STEP_SUMMARY` so it renders as markdown on the run page, alongside the per-run token usage table that's already there.

Also fixes Step 2's cross-reference from "Step 6" to "Step 7" (the summary is Step 7).

### DIFF
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -160,7 +160,8 @@ If no problems found (or none passed the gates), report "all clear" with: runs a
 reviewed, brief quality assessment, and any below-threshold findings recorded in the tracking
 issue.
 
-Write the summary to the GitHub Actions step summary so it appears on the run page:
+Save the summary to `/tmp/summary.md`, then write it to the GitHub Actions step summary so it
+appears on the run page:
 
 ```bash
 cat /tmp/summary.md >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -159,3 +159,9 @@ tracking issue.
 If no problems found (or none passed the gates), report "all clear" with: runs analyzed, sessions
 reviewed, brief quality assessment, and any below-threshold findings recorded in the tracking
 issue.
+
+Write the summary to the GitHub Actions step summary so it appears on the run page:
+
+```bash
+cat /tmp/summary.md >> "$GITHUB_STEP_SUMMARY"
```
tend-agent, line 167: `/tmp/summary.md` is never created — no earlier step instructs the agent to write its summary there, and grepping the repo confirms this is the only reference to the path. As written, the `cat` will fail with `No such file or directory`, so nothing reaches the step summary.

Either tell the agent to save the summary to `/tmp/summary.md` first (e.g. via the Write tool), or have the agent append directly to `$GITHUB_STEP_SUMMARY` with a heredoc — mirroring how `action.yaml` writes the token usage table at line 216.

tend-agent (commented):
One concern on Step 7 — see inline. The step reference fix and the intent of the step-summary addition look right.

tend-agent (approved):

PR #250, opened 2026-04-11
feat(token-report): add per-run table, fix macOS compat

max-sixty:
Add a per-run breakdown table below the existing per-workflow rollup in `token-report.sh`, showing RUN ID, WORKFLOW, tokens, COST, and TIME for each individual run. Also fix `date -d` (GNU-only) to fall back to BSD `date -v` on macOS, and move the footnote out of the `column -t` pipe so it renders as prose.

Also drops GNU coreutils from the requirements header since BSD date is now supported.

### DIFF
--- a/plugins/tend-ci-runner/scripts/token-report.sh
+++ b/plugins/tend-ci-runner/scripts/token-report.sh
@@ -29,7 +29,7 @@ export NO_COLOR=1
 export CLICOLOR_FORCE=0
 
 HOURS=${1:-168}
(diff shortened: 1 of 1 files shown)


tend-agent (approved):
One tangential observation (outside the diff, so no inline suggestion): the header comment at line 20 still declares `GNU coreutils (date -d)` as required, but the script now works on BSD `date -v` too. Happy to push a one-line doc fix if you'd like — otherwise this can ride along with the next edit to that file.

PR #257, opened 2026-04-12
chore: add timeout-minutes to tend workflow jobs

tend-agent:
## Summary
- Adds `timeout-minutes: 60` to all `tend-*` workflow jobs
- The published generator already emits this, but the committed workflows were missing it
- Without an explicit timeout, runaway jobs fall back to GitHub's 6-hour default

## Test plan
- [ ] Verify `cd generator && uv run pytest` passes
- [ ] Confirm no other workflow content was modified

### DIFF
--- a/.github/workflows/tend-ci-fix.yaml
+++ b/.github/workflows/tend-ci-fix.yaml
@@ -17,6 +17,7 @@ jobs:
   fix-ci:
     if: github.event.workflow_run.conclusion == 'failure'
     runs-on: ubuntu-24.04
+    timeout-minutes: 60
     permissions:
       contents: write
       pull-requests: write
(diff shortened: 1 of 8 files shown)


tend-agent (commented):
These workflow files are generated by `uvx tend@latest init` and their headers say "Do not edit this file directly". The generator (`workflows.py`) does not emit `timeout-minutes`, so these changes will be overwritten on the next nightly regeneration.

Two options that would persist:

1. **Add a default in the generator** — emit `timeout-minutes: 60` in each job template in `workflows.py`. This benefits all repos using tend.
2. **Add per-job config in `.config/tend.toml`** using the existing `jobs` extras mechanism (already tested for this exact use case in `test_job_extras_add_key`). This would only affect this repo.

Also, the non-generated `review-reviewers.yaml` is missing `timeout-minutes` and isn't covered here.

PR #258, opened 2026-04-12
fix(action): repair notification marking — resolve run_started_at, catch API errors

tend-agent:
## Summary

- Fix `run_started_at` resolution — `github.run_started_at` is not a real context property, so the env was always empty for non-`workflow_run` events. Now fetched from the REST API (`repos/{owner}/{repo}/actions/runs/{id}`), which works for all event types.
- Wrap the `gh api notifications | jq` pipeline in a subshell with a warning fallback, so a transient API error (non-JSON response) emits a warning instead of failing the composite action.

## Evidence

Observed on PRQL/prql runs [24301240279](https://github.com/PRQL/prql/actions/runs/24301240279) (`tend-review`) and [24301252595](https://github.com/PRQL/prql/actions/runs/24301252595) (`tend-triage`) on 2026-04-12. Both failed with:

```
jq: parse error: Invalid numeric literal at line 1, column 2
##[error]Process completed with exit code 5.
```

Verified that `RUN_STARTED_AT` is empty in successful runs too (e.g., tend-notifications run [24301882032](https://github.com/PRQL/prql/actions/runs/24301882032)), confirming `github.run_started_at` is universally absent.

**Gate assessment**: High confidence (verified directly via docs and live runs), structural (same conditions always produce the same result), targeted fix.

## Test plan

- [x] `cd generator && uv run pytest` — 158 passed
- [ ] Verify on next PRQL/prql `pull_request_target` or `issues` event that `RUN_STARTED_AT` is populated and the step doesn't crash

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -221,13 +221,8 @@ runs:
             ;;
         esac
 
-        # github.run_started_at is not a real context property — the env
-        # expression only has a value for workflow_run events.  Fall back
-        # to the REST API, then to "now" if the API call fails.
-        if [ -z "$RUN_STARTED_AT" ]; then
-          RUN_STARTED_AT=$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}" \
-            --jq '.run_started_at' 2>/dev/null || date -u +%Y-%m-%dT%H:%M:%SZ)
-        fi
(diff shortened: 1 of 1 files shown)


max-sixty:
can we do a single canonical approach rather rthan this fallback?

tend-agent:
Good call — simplified to always use the REST API (`repos/{owner}/{repo}/actions/runs/{id}` → `.run_started_at`). Dropped the GitHub context expression and the `date` fallback. The subshell around the notifications pipeline still catches transient failures downstream.

PR #264, opened 2026-04-13
chore: update tend workflows

tend-agent:
Automated nightly regeneration of tend workflow files.

### DIFF
--- a/.github/workflows/tend-ci-fix.yaml
+++ b/.github/workflows/tend-ci-fix.yaml
@@ -17,6 +17,7 @@ jobs:
   fix-ci:
     if: github.event.workflow_run.conclusion == 'failure'
     runs-on: ubuntu-24.04
+    timeout-minutes: 60
     permissions:
       contents: write
       pull-requests: write
--- a/.github/workflows/tend-mention.yaml
+++ b/.github/workflows/tend-mention.yaml
@@ -15,20 +15,8 @@ on:
   # Works for same-repo PRs only; secrets unavailable on fork PRs (no _target variant exists)
   pull_request_review:
     types: [submitted]
-  # `created` is intentionally absent. Modern GitHub fires *both*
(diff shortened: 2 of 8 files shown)


.github/workflows/tend-ci-fix.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
   fix-ci:
     if: github.event.workflow_run.conclusion == 'failure'
     runs-on: ubuntu-24.04
+    timeout-minutes: 60
```
max-sixty, line 20: I think a bunch of these changes are regressions? 

what happened? did we install a local version but fail to do a release?

tend-agent: Yes, these are regressions — caused exactly by what you suspected. The nightly runs `uvx tend@latest init`, which pulls [tend 0.0.11](https://pypi.org/project/tend/0.0.11/) (released 2026-04-10). Main has 10 generator commits since 0.0.11 that aren't on PyPI yet, and each reintroduction in this PR maps to one of them:

| Revert in this PR | Landed on main in |
|---|---|
| Re-adds `timeout-minutes: 60` to every job | #223 dropped it |
| `pull_request_review_comment: [edited]` → `[created, edited]` + drops the dedup comment | #225 added the dedup |
| Re-adds `!= 'tend-agent'` filter on `issue_comment` | #233 removed it |
| Drops `--paginate` from `gh api` calls in the verify job | #228 added it |
| Drops the REVIEW_ID / inline-comment @-mention fetch | #231 added it |
| Re-inlines the long behavioral instructions into the prompt | #235 moved them to the skill |
| Drops the notifications shadow/pre-mark block | #254 added it |

So no local install, just a stale PyPI version. Fix is to cut 0.0.12 — then the next nightly will generate forward instead of backward. Want me to open a release PR via `/release`?

PR #265, opened 2026-04-13
docs(claude.md): document local-workflow update policy

tend-agent:
## Summary

Per maintainer guidance ([comment](https://github.com/max-sixty/tend/pull/265#issuecomment-4234764327)): document the local-workflow update policy in CLAUDE.md rather than adding skip logic to the nightly skill. Tend's `tend-*.yaml` workflows track the latest published release; nightly regen via `uvx tend@latest init` is the source of truth, updating earlier to the latest release is fine, and the in-tree generator should not push workflows beyond the latest release. The lag between a generator commit and the next release is expected and closes at the next release.

The earlier nightly-skill skip is reverted; only CLAUDE.md changes remain.

## Test plan

- [ ] Future nightly runs on tend continue to regen against `uvx tend@latest`; reviewers reading CLAUDE.md understand the resulting "revert" PRs are expected during the post-commit/pre-release window

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,4 +48,11 @@ Consuming repos regenerate their `tend-*.yaml` workflows nightly (tend itself
 included — it dogfoods its own workflows). Changes to the generator do not
 require manual regeneration in downstream repos.
 
+Tend's own `tend-*.yaml` workflows track the latest published release. They
+update each night via `uvx tend@latest init`. Updating earlier to the latest
+release (e.g., during a release commit) is fine. Do not update them beyond
+the latest release using the in-tree generator — between a generator commit
+and the next release, the local workflows lag the in-tree generator, and
+that is expected; the gap closes at the next release.
+
 Linting: `pre-commit run --all-files` (ruff, typos, actionlint, uv-lock).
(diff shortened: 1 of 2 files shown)


max-sixty:
I think instead we should just avoid upgrading the local workflows outside of releases

tend-agent:
Agreed — that's the cleaner framing. Pushed [19aa2ae](https://github.com/max-sixty/tend/commit/19aa2ae) which replaces the in-tree-generator approach with an early skip in Step 5 when `generator/pyproject.toml` exists. Adopter repos (no in-tree generator) continue to regen via `uvx tend@latest` unchanged.

One note: with this change, the only path that updates tend's local workflows is the release flow, but the release skill doesn't currently include a regen step. Want me to open a follow-up PR adding `uv run --project generator tend init` to the release skill, or would you rather handle it ad-hoc per release?

max-sixty:
let instead just make a note in our Claude.md that we update each night to the latest release; that it's ok to update earlier to the latest release; but not to update beyond it

tend-agent:
Done — pushed [6b3d1dc](https://github.com/max-sixty/tend/commit/6b3d1dc) which reverts the nightly-skill skip and adds a paragraph to CLAUDE.md documenting the policy: workflows track the latest published release, updating earlier to latest release is fine, in-tree generator should not push workflows beyond the latest release, and the lag between a generator commit and the next release is expected. PR title/description updated to match.

PR #266, opened 2026-04-13
fix(action): restore token aggregation broken by result-event filter

tend-agent:
## Problem

Every `token-usage.json` since #249 reports zero tokens. Confirmed across all 61 tend CI runs in the past 24 h on this repo (12 `tend-review`, 19 `tend-mention`, 22 `review-reviewers`, 6 `tend-notifications`, 1 `tend-nightly`, 1 `tend-weekly`).

Sample artifact (run [24330280717](https://github.com/max-sixty/tend/actions/runs/24330280717), `tend-review`):

```json
{"input_tokens":0,"output_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"turns":0,"model":"opus","cost_usd":0}
```

## Root cause

#249 added a `select(.type == "result")` filter that targets the SDK's per-session result event — which carries `total_cost_usd`, `num_turns`, and a session-level `usage` object. That event is real, but it isn't written to the per-turn session JSONL under `/home/runner/.claude/projects/`; `claude-code-action` writes it (along with the rest of the SDK message stream) to `$RUNNER_TEMP/claude-execution-output.json` and exposes that path as the `execution_file` action output. Reading from `/home/runner/.claude/projects/` therefore matched zero entries and every numeric field fell back to 0.

This was masked yesterday because the 24 h analysis window straddled the #249 merge time (2026-04-11 20:28 UTC) — most artifacts in that window pre-dated the regression. Today's window is fully post-#249, so every run reports zeros.

## Fix

Read `${{ steps.claude.outputs.execution_file }}` and pull `total_cost_usd`, `num_turns`, and `usage.*` directly from the `result` entry. Cost is now whatever the SDK reports (no hard-coded model rate table), so Sonnet/Haiku adopters and 1-hour cache usage are accounted for correctly.

Verified with a synthetic execution-file array containing a real `result` entry from run 24330280717: produces `34 input / 4954 output / 1035105 cache_read / 39191 cache_creation, 24 turns, $0.89` — matches the `total_cost_usd` printed in that run's job log (0.886... rounded to 2dp).

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -240,40 +240,39 @@ runs:
       if: always()
       id: tokens
       shell: bash
+      env:
+        EXECUTION_FILE: ${{ steps.claude.outputs.execution_file }}
+        MODEL: ${{ inputs.model }}
       run: |
         LOGS_DIR="/home/runner/.claude/projects"
         mkdir -p "$LOGS_DIR"
 
-        mapfile -t FILES < <(find "$LOGS_DIR" -name "*.jsonl" -type f 2>/dev/null)
-
-        MODEL="${{ inputs.model }}"
-        if [ ${#FILES[@]} -eq 0 ]; then
-          USAGE='{"input_tokens":0,"output_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"turns":0,"model":"'"$MODEL"'","cost_usd":0}'
+        # claude-code-action writes the full SDK message stream to
+        # $RUNNER_TEMP/claude-execution-output.json and exposes the path as
+        # its `execution_file` output. The array ends with a
+        # `type: "result"` entry that carries total_cost_usd, num_turns,
+        # and the per-session usage totals across all models — so cost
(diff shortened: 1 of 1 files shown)


max-sixty:
but there is a cost field at the end of the logs! can we search for and use that?

tend-agent:
Good catch — you are right. The `result` event carrying `total_cost_usd` does exist; I was looking in the wrong file. `claude-code-action` writes the full SDK message stream to `$RUNNER_TEMP/claude-execution-output.json` (path exposed via its `execution_file` output), and the last entry there is the `type: "result"` summary with `total_cost_usd`, `num_turns`, and a session-level `usage` object.

Pushed [250ebe9](https://github.com/max-sixty/tend/pull/266/commits/250ebe9) which switches to that source: cost comes straight from `total_cost_usd`, so the model-rate table goes away and Sonnet/Haiku adopters plus 1-hour-cache usage are accounted for automatically. PR description updated to match.

PR #278, opened 2026-04-15
docs: document workflow override mechanism with skip-review-label example

### LINKED ISSUE #276: Support skipping review on labeled PRs (e.g. tend:dismissed)
## Problem

When a PR has already been reviewed by tend and the author is satisfied with the feedback, subsequent pushes (e.g. fixup commits, rebases) trigger a full re-review via the `synchronize` event. This is wasteful and noisy for PRs that are ready to merge.

## Current workaround

We patch `tend-review.yaml` after regeneration with a script that adds a label check to the job's `if:` condition:

```yaml
if: >-
  github.event.pull_request.draft == false &&
  !contains(github.event.pull_request.labels.*.name, 'tend:dismissed')
```

This lets the PR author add a `tend:dismissed` label to skip further reviews, and remove it if they want tend to look again after significant changes.

## Proposal

Support this natively in tend, either as:

1. **A config option** in `tend.toml` — e.g. `skip_review_label = "tend:dismissed"` — that gets baked into the generated workflow's `if:` condition
2. **A built-in default** — tend could ship with this behavior out of the box, since re-reviewing already-dismissed PRs is rarely desired

Option 1 is more flexible; option 2 reduces setup friction.

## Context

- The `tend-review.yaml` workflow is generated and warns "do not edit this file directly"
- Local patches survive until the next `tend init`, then need reapplication
- A native solution would eliminate the need for post-regeneration patching

tend-agent:
## Summary

- Documents the `workflow_extra` / `jobs.<name>` override mechanism in both the install-tend skill (where setup-stage users look for guidance) and `docs/tend.example.toml` (where `README.md` already points for advanced options).
- Uses the skip-review-on-label pattern from #276 as the worked example, since it's the concrete need that prompted this gap report.
- Adds `test_skip_review_on_label_override` to pin the exact override documented in both places (scalar replace of the review job's `if:`), so the example can't silently break.

The mechanism already exists in the generator and is spec'd in `DESIGN.md § Extending generated workflows` — #276 notes that it's just not surfaced where adopters see it during setup. A dedicated `skip_review_label` config option could still be added on top of this; intentionally leaving that call to a maintainer since the generic mechanism already covers the need without maintaining a second knob.

Closes max-sixty/tend#276 once merged (or leave open if a dedicated option is preferred).

## Test plan

- [x] `cd generator && uv run pytest` — 159 passed
- [x] `pre-commit` on changed files — ruff/typos/format/trailing-ws clean
- [ ] Render check: confirm the new section in `tend.example.toml` reads well alongside the existing `### review-runs` block

### DIFF
--- a/docs/tend.example.toml
+++ b/docs/tend.example.toml
@@ -142,3 +142,39 @@ bot_name = "my-project-bot"
 #
 # Daily (default 07:47 UTC). Reviews recent CI runs for behavioral problems
 # and proposes skill/config improvements.
+
+# ## Workflow overrides (patching)
+#
+# Generated `tend-*.yaml` files are regenerated from this config on every
+# `tend init`. Don't patch them directly — edits will be clobbered. Use the
+# built-in override mechanism instead. Two override points per workflow:
+#
+# - `workflow_extra` — top-level workflow keys (e.g. `env:`)
+# - `jobs.<job-name>` — keys inside a named job (e.g. `if:`, `timeout-minutes`,
+#   `runs-on`, `permissions`)
+#
+# Overrides use RFC 7396 JSON Merge Patch semantics: mappings deep-merge,
+# scalars and lists replace, and `= nil` deletes a key. See DESIGN.md §
+# "Extending generated workflows" for the full spec.
+#
+# ### Example: skip review on labeled PRs
+#
+# The review job's default `if:` is `github.event.pull_request.draft == false`.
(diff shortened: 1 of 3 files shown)


tend-agent:
Closing — #277 landed first (merged at 2026-04-15T05:43:22Z) covering the same scope (docs in `tend.example.toml`, install-tend skill update, skip-review-label regression test). A parallel notifications-skill run on the same issue produced overlapping PRs; the other got there first.

PR #291, opened 2026-04-16
skills(review-reviewers): pilot gist-backed evidence storage

max-sixty:
Pilots per-month, per-target secret gists as the evidence store for `review-reviewers`. The tracking issue becomes a directory — one bot comment per new gist — so maintainers still get a notification when new monthly evidence appears, without the old per-run comment churn.

`review-runs` keeps the issue-comment mechanism (inlined from the shared file it used to import) with a TODO to migrate once the pilot validates. `shared/review-gates.md` now holds just the gates + finding format, not the storage mechanism.

## Why gists

- Single growing file instead of fighting the 65 KB comment-body limit
- Gist revision history gives a per-run diff for free
- Per-target separation (`worktrunk`, `tend`, `prql`) — the matrix already fans out, now the evidence does too

## PAT scope change

Classic PATs need `gist` scope; fine-grained PATs need `gists:write`. Existing adopters' bots will need their PAT regenerated before the pilot expands beyond tend, but this PR only activates gist storage in `review-reviewers` (tend-internal), so downstream repos are unaffected today.

## Notes

- First run of a new month in the 3-target matrix can briefly create duplicate tracking issues (all three concurrent entries see no existing one); later runs pick the lowest-numbered match deterministically and maintainers close duplicates.
- BSD/GNU date fallback for "last month" handles end-of-month quirks.
- Went through two review passes with the code-reviewer agent — caught invalid `gh issue create --json` flag, missing `--raw` on `gh gist view`, `$(cat)` newline-stripping, undefined per-run findings filename, and `date -d 'last month'` EOM quirk.

### DIFF
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -84,7 +84,130 @@ without `-R` default to tend.
 
 @review-gates.md
 
-Use `TRACKING_LABEL="review-reviewers-tracking"` for this skill's tracking issues.
+## Evidence accumulation
+
+Each run only sees a window of CI sessions, but patterns emerge over days or weeks. Evidence
+for this skill lives in **secret gists owned by the bot** — one per `(target repo, month)`
+pair. A monthly tracking issue on tend labeled `review-reviewers-tracking` lists the gists
+via bot comments, so maintainers can discover them.
+
+Secret gists are URL-unlisted but readable by anyone with the URL; they are at least as
+private as the current public tracking issues, and give a single structured file that
+accumulates per-target findings without hitting the 65 KB comment limit.
+
+### Setup
+
+```bash
+MONTH=$(date +%Y-%m)
+TRACKING_LABEL="review-reviewers-tracking"
+TARGET="$ARGUMENTS"
+GIST_DESC="review-reviewers evidence: $TARGET $MONTH"
+```
+
+### Finding or creating the tracking issue
+
+The tracking issue lives on tend (the current repo). It indexes gists via one comment per
+new gist — no per-run comments, no body edits.
+
+The matrix runs three targets concurrently on the same cron tick, so the first run of a new
+month races: all three targets can find no tracking issue and each create one. Sorting and
+picking the lowest-numbered match keeps later runs deterministic — maintainers can close any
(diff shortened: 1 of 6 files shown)


plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -84,7 +84,130 @@ without `-R` default to tend.
...
+GIST_ID=$(gh api /gists --paginate \
+  --jq ".[] | select(.description == \"$GIST_DESC\") | .id" | head -1)
+
+if [ -z "$GIST_ID" ]; then
+  cat > /tmp/gist-seed.md << EOF
+# review-reviewers evidence — $TARGET — $MONTH
+
+Secret gist. Append-only log of below-threshold findings used for gate evaluation.
+EOF
+  GIST_URL=$(gh gist create --desc "$GIST_DESC" /tmp/gist-seed.md)
```
```suggestion
  mkdir -p /tmp/gist-seed
  cat > /tmp/gist-seed/findings.md << EOF
# review-reviewers evidence — $TARGET — $MONTH

Secret gist. Append-only log of below-threshold findings used for gate evaluation.
EOF
  GIST_URL=$(gh gist create --desc "$GIST_DESC" /tmp/gist-seed/findings.md)
```

`gh gist create` uses the basename of the local file as the gist filename (the `-f/--filename` flag is stdin-only). As written, the created gist has a file named `gist-seed.md`, but `gh gist view -f findings.md` and the PATCH below both reference `findings.md`, so the seed's header never ends up in the accumulated log and the gist ends up with two files instead of one.

tend-agent (commented):
One bug: the seed writes to `/tmp/gist-seed.md`, so the created gist contains a file named `gist-seed.md`. All subsequent reads and PATCHes target `findings.md`, which doesn't exist on the first run — the first `gh gist view -f findings.md` silently produces an empty file, and the first PATCH then *creates* `findings.md` (leaving `gist-seed.md` as a permanent orphan containing the intended header/purpose doc).

Fix: name the seed file `findings.md` so its basename matches what later code reads. Inline suggestion below. The `-f/--filename` flag on `gh gist create` only applies to stdin input, so the cleanest fix is to put the seed in a dedicated directory. After this change, reads/PATCHes all operate on the same `findings.md` from day one and the gist has a single file.

tend-agent (approved):

PR #298, opened 2026-04-18
install-tend: capture bot auth stance in profile bio

### LINKED ISSUE #296: install-tend: capture bot authorization stance and reflect it in bot profile bio
Raised by @max-sixty in [max-sixty/worktrunk#966](https://github.com/max-sixty/worktrunk/issues/966#issuecomment-4270951694).

During `install-tend` setup, surface an explicit decision about what the creator is OK with contributors/users asking the bot to do — so that policy is set once up front rather than reactively. The decision could then be reflected in the bot account's GitHub profile bio (≤160 chars), making the scope discoverable on the user page.

Two pieces:

1. Add a prompt/step during `install-tend` that captures the authorization stance (open to anyone, contributors only, etc.).
2. Optionally write the chosen stance to the bot's profile bio automatically. Bio is writable via `PATCH /user` with the `bio` field; the PAT that `install-tend` already creates would need `user` scope (classic) or `Profile: read and write` (fine-grained) to do this in one pass — the current default doesn't include that, so either extend the PAT scope or surface manual instructions.

No action yet — tracking.

tend-agent:
## Summary
- Adds a new step to `install-tend` that asks the creator what contributors/users may ask the bot to do, drafts a ≤160-char bio, and writes it via `PATCH /user` using the bot PAT from step 8.
- Extends the default classic PAT scopes with `user` (pre-fill URL, checklist, audit script) and documents the fine-grained `Profile: read and write` equivalent. `pat-scope-audit.sh` now flags bots missing `user` on the nightly run.
- Keeps the behavior advisory — the bio is discoverable on the bot's user page; bot runtime behavior doesn't gate on it.

Closes #296.

## Test plan
- [ ] Dry-run `install-tend` on a sandbox repo: verify step 10 prompts for stance, drafts bio, and `gh api user --jq '.bio'` returns the new value.
- [ ] Generate a new classic PAT via the pre-filled URL and confirm `user` is pre-checked alongside the existing scopes.
- [ ] Run `plugins/tend-ci-runner/scripts/pat-scope-audit.sh` against a bot PAT that lacks `user`; expect `STATUS=missing`.
- [ ] `uvx pre-commit run --files <modified>` passes locally.

### DIFF
--- a/plugins/install-tend/skills/install-tend/SKILL.md
+++ b/plugins/install-tend/skills/install-tend/SKILL.md
@@ -293,17 +293,12 @@ bot to do, then reflect that stance in the bot's profile bio (≤160 chars)
 so it's discoverable on the bot's user page. This is advisory — the bot
 doesn't gate behavior on it.
 
-Ask the creator which stance applies. Typical answers:
+Ask the creator which stance applies. Each option is a drop-in bio suffix;
+the middle one is the recommended default.
 
-- **Open to anyone** — anyone may @-mention the bot
-- **Contributors only** — only prior contributors to the repo
-- **Maintainers only** — only repo maintainers
-
-Draft a short bio from the stance. Examples:
-
-- `Claude-powered CI bot for <owner>/<repo>. @-mention me — open to anyone.`
-- `Claude-powered CI bot for <owner>/<repo>. Contributors welcome to @-mention.`
-- `Claude-powered CI bot for <owner>/<repo>. Maintainers only, please.`
+- `tend agent for <owner>/<repo>. Feel free to ask me questions about <repo>.`
+- `tend agent for <owner>/<repo>. I triage issues and help maintain <repo>.` (recommended)
+- `tend agent for <owner>/<repo>. I respond to maintainers of <repo>.`
 
 Check the current bio as the bot — skip if already set to the chosen value:
 
--- a/plugins/install-tend/skills/install-tend/SKILL.md
+++ b/plugins/install-tend/skills/install-tend/SKILL.md
@@ -230,15 +230,17 @@ echo "$TOKEN" | gh secret set CLAUDE_CODE_OAUTH_TOKEN --repo "$REPO"
 ## 8. Bot PAT and secret
 
 The bot needs a classic PAT with `repo`, `workflow`, `notifications`,
-`write:discussion`, and `gist` scopes. `workflow` is required to push commits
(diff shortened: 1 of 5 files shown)


tend-agent (commented):
One drift to fix: [`plugins/tend-ci-runner/skills/nightly/SKILL.md#L16`](https://github.com/max-sixty/tend/blob/291067ebb3d1bf9d08d49ef3b788015e3a9ede17/plugins/tend-ci-runner/skills/nightly/SKILL.md#L16) still lists the scopes as `repo`, `workflow`, `notifications`, `write:discussion`, `gist` — missing `user`. With the audit script now requiring `user`, this prose is out of sync and a newly-provisioned bot reading the nightly skill would see stale expectations.

Relatedly, the comment at `plugins/tend-ci-runner/scripts/pat-scope-audit.sh:15-17` names three files where the scope list is duplicated and calls itself the "one executable reference" — but `nightly/SKILL.md` is a fourth duplication site it doesn't list. Worth adding to that comment so the next scope change updates all four.

Both are outside this diff — happy to push a follow-up commit if the approach is fine.

tend-agent:
Pushed both fixes in f985329: added `user` to the scope list in `nightly/SKILL.md` and added that file to the duplication comment in `pat-scope-audit.sh`.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -284,7 +286,38 @@ fi
...
+
+- **Open to anyone** — anyone may @-mention the bot
+- **Contributors only** — only prior contributors to the repo
+- **Maintainers only** — only repo maintainers
+
+Draft a short bio from the stance. Examples:
+
+- `Claude-powered CI bot for <owner>/<repo>. @-mention me — open to anyone.`
+- `Claude-powered CI bot for <owner>/<repo>. Contributors welcome to @-mention.`
+- `Claude-powered CI bot for <owner>/<repo>. Maintainers only, please.`
```
max-sixty: I think these should be `tend agent for ...`. And then the options should be something like:
- Feel free to ask me questions about <repo>
- I triage issues and help maintain <repo>
- I respond to maintainers of <repo>

with a recommendation for the middle one?

tend-agent: Good call — switched the prefix to `tend agent for <owner>/<repo>` and folded the stance options into drop-in bio suffixes with the triage one flagged as recommended. Pushed in b67fd85.

PR #301, opened 2026-04-18
feat(review): probe merge ref, fall back to head so tend-review runs on conflicting PRs

### LINKED ISSUE #299: feat(review): make checkout ref configurable so tend-review works on conflicting PRs
## Summary

`tend-review` fails on every PR that arrives with a merge conflict against the base branch. The generated workflow pins `actions/checkout` to `refs/pull/${{ github.event.pull_request.number }}/merge`, and GitHub does not materialize that ref for conflicting PRs — only `refs/pull/N/head` exists in that state.

Result: the checkout step 404s, every subsequent step cascades as `skipped`, and the PR carries a red ❌ `tend-review` status until the author rebases.

## Observed impact

In one downstream repo with moderate PR cadence: ≥25 `tend-review` failures across ~4 days, all on conflict-prone branches (`renovate/*`, concurrent feature branches touching shared files like `cspell.json` or lockfiles). Several PRs failed 3× before their authors resolved the conflict — each failure burning ~40s of GitHub Actions runtime plus the workflow startup slice of Claude OAuth quota.

The failure is **inevitable** whenever two PRs touch the same file in overlapping ways, which is routine under Dependabot/Renovate/concurrent feature work.

## Why it matters

`pull_request_target` (which `tend-review` uses) means `actions/checkout` defaults to the base branch, not the PR. An explicit `ref:` is required. The current choice (`/merge`) reviews the prospective post-merge tree, which is a quality win when the PR is mergeable — and a total review blocker when it isn't. There is no built-in fallback in `actions/checkout@v6` (its `ref:` input is single-valued).

## Current workaround (downstream)

Override `[workflows.review.jobs.review.steps]` in `.config/tend.toml` to replace the generated steps array with one that uses `refs/pull/N/head`. RFC 7396 list-replace semantics (workflows.py:766–776) make this work.

The override is effective but brittle: it requires inlining every generated step (including `[[setup]]` content) and a `pnpm check` tripwire to catch drift whenever `uvx tend@latest init` updates the base template.

## Proposed upstream change

Add a first-class config knob so adopters don't need to own the entire steps array:

```toml
[workflows.review]
checkout_ref = "head"  # default "merge", accept "head" | "merge" | maybe "auto"
```

Or, if simpler, make it a top-level option:

```toml
checkout_ref = "head"
```

**`"auto"` variant (optional):** a pre-checkout shell step that probes `gh api repos/${{ github.repository }}/git/ref/pull/N/merge` and falls back to `/head` on 404. Gives the post-merge view when available, graceful degradation when not. More complexity; happy to skip it in v1 and just support the static `"head"` option.

## Offer

Happy to PR this against `generator/src/tend/workflows.py` (extend the `review` job template + `WorkflowConfig` dataclass). Flag me on this issue and I'll send a PR within the week.

## Related

Reading suggests this hasn't been filed before — couldn't find prior art in issues, and `DESIGN.md` line 113–127's ownership table marks "Checkout" as generator-owned without a listed override path.

tend-agent:
## Summary

tend-review currently pins `actions/checkout` to `refs/pull/N/merge`, which GitHub only materializes for mergeable PRs. On any conflicting PR the checkout 404s and every downstream step cascades as skipped until the author rebases.

This PR adds a probe step that calls `gh api .../git/ref/pull/N/merge` and wires its output into the checkout's `ref:`. If the merge ref is present, checkout uses it (post-merge tree, current behavior). If it 404s, the probe writes `refs/pull/N/head` and emits a `::notice::` so the review context is clear on fallback. tend-review now runs on every PR regardless of merge state.

Closes #299.

## Design notes

- Not configurable, per @max-sixty's feedback on the issue — review should always succeed. Earlier commit on this branch had a `checkout_ref` config knob; that's been removed.
- Probe uses `secrets.GITHUB_TOKEN` (the default workflow token) rather than the bot token — needs only `repos:read`, available on `pull_request_target` by default, and keeps the bot-token blast radius limited to checkout itself.
- On fallback the review sees the PR branch in isolation rather than the post-merge tree. The `::notice::` makes that visible in the run summary.
- Skipped an additional retry/backoff for the race where GitHub hasn't yet computed mergeability — empirically the computation finishes before the job starts, and a stale-ref race is a rarer failure mode than the 404 we're fixing.

## Test plan

- [x] `cd generator && uv run pytest` — 164 passing, including a new test asserting the probe step exists, runs before checkout, and wires its output into `ref:`
- [x] Regtest snapshots updated for `tend-review.yaml` (two variants: minimal + with-setup)
- [x] `pre-commit run` on changed files — ruff, typos, actionlint pass
- [ ] Dogfood: next nightly regeneration picks up the new probe step in tend's own `tend-review.yaml`

cc @jrdncstr — if you have time, a review would still be welcome per @max-sixty's suggestion on #299.

### DIFF
--- a/generator/tests/_regtest_outputs/test_generate.test_workflow_with_setup_regtest[review].out
+++ b/generator/tests/_regtest_outputs/test_generate.test_workflow_with_setup_regtest[review].out
@@ -34,7 +34,7 @@ jobs:
       - name: Resolve PR checkout ref
         id: pr_ref
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.BOT_TOKEN }}
           PR: ${{ github.event.pull_request.number }}
         run: |
           if gh api "repos/${{ github.repository }}/git/ref/pull/$PR/merge" --silent 2>/dev/null; then
--- a/generator/tests/_regtest_outputs/test_generate.test_workflow_with_setup_regtest[review].out
+++ b/generator/tests/_regtest_outputs/test_generate.test_workflow_with_setup_regtest[review].out
@@ -26,9 +26,26 @@ jobs:
       actions: read
       issues: write
     steps:
+      # GitHub only materializes refs/pull/N/merge for mergeable PRs ? on
+      # conflicting PRs it 404s and every downstream step cascades as skipped.
+      # Probe first and fall back to /head so review always runs; on fallback
+      # the review sees the PR branch in isolation rather than the post-merge
+      # tree.
+      - name: Resolve PR checkout ref
+        id: pr_ref
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR: ${{ github.event.pull_request.number }}
+        run: |
+          if gh api "repos/${{ github.repository }}/git/ref/pull/$PR/merge" --silent 2>/dev/null; then
+            echo "ref=refs/pull/$PR/merge" >> "$GITHUB_OUTPUT"
+          else
+            echo "ref=refs/pull/$PR/head" >> "$GITHUB_OUTPUT"
+            echo "::notice::refs/pull/$PR/merge unavailable (likely merge conflict); falling back to /head"
+          fi
       - uses: actions/checkout@v6
         with:
-          ref: refs/pull/${{ github.event.pull_request.number }}/merge
+          ref: ${{ steps.pr_ref.outputs.ref }}
           fetch-depth: 0
           fetch-tags: true
           token: ${{ secrets.BOT_TOKEN }}
--- a/docs/tend.example.toml
+++ b/docs/tend.example.toml
@@ -142,6 +142,15 @@ bot_name = "my-project-bot"
 #
 # [workflows.review]
 # prompt = "/my-custom-review {pr_number}"    # override default prompt
+# checkout_ref = "head"                       # "merge" (default) or "head"
+#
+# By default tend-review checks out `refs/pull/N/merge` — the prospective
+# post-merge tree. GitHub only materializes that ref for mergeable PRs, so on
+# PRs with conflicts against the base branch the checkout 404s and the whole
+# review job fails. Setting `checkout_ref = "head"` switches to
+# `refs/pull/N/head`, which always exists. The trade-off: the review sees the
+# PR branch in isolation rather than its post-merge state, so conflicts and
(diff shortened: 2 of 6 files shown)


tend-agent (commented):
Design flag for the author, not for jrdncstr — @max-sixty [commented on #299](https://github.com/max-sixty/tend/issues/299#issuecomment-4273068667) 34 seconds after this PR opened, likely after the run that pushed it had already exited:

> I don't think this should be a configurable option. We should always succeed regardless of whether there's a merge conflict. Evaluate how we can get GHA to checkout a reasonable commit

That redirects the design toward the `"auto"` variant from the original issue (probe `refs/pull/N/merge`, fall back to `/head` on 404), applied unconditionally — no `checkout_ref` knob. If that redirect is accepted, most of this diff goes away: the `WorkflowConfig.checkout_ref` field, the config validation, the docs block, and three of the four new tests.

The code as written is fine on its own terms — validation rejects use outside `review`, tests cover default/override/invalid-value/non-review-rejection, docs call out the isolation trade-off. Flagging the architectural shift rather than polishing what may not survive it.

generator/tests/_regtest_outputs/test_generate.test_workflow_with_setup_regtest[review].out
```diff
@@ -26,9 +26,26 @@ jobs:
...
     steps:
+      # GitHub only materializes refs/pull/N/merge for mergeable PRs ? on
+      # conflicting PRs it 404s and every downstream step cascades as skipped.
+      # Probe first and fall back to /head so review always runs; on fallback
+      # the review sees the PR branch in isolation rather than the post-merge
+      # tree.
+      - name: Resolve PR checkout ref
+        id: pr_ref
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Comment (max-sixty) [inline generator/tests/_regtest_outputs/test_generate.test_workflow_with_setup_regtest[review].out]: why use this rather than the bot token

Comment (tend-agent) [reply generator/tests/_regtest_outputs/test_generate.test_workflow_with_setup_regtest[review].out]: Good catch — switched to the bot token in 4314539. My original rationale (smaller blast radius on `GITHUB_TOKEN`) doesn't hold up against [`docs/security-model.md`](https://github.com/max-sixty/tend/blob/4314539/docs/security-model.md#L199-L224), which explicitly says all Claude workflows should use the bot token for consistent identity and that `GITHUB_TOKEN` "is ephemeral (single job) and automatically scoped by each workflow's `permissions:` block. Not a meaningful leak target."

PR #314, opened 2026-04-19
nightly: include version bump and diff stat in workflow-update PR body

tend-agent:
## Summary

- In `plugins/tend-ci-runner/skills/nightly/SKILL.md` Step 6, extract the pinned tend version from workflow YAML before and after `uvx tend@latest init`, and surface the bump in the PR title (`chore: update tend workflows (A → B)`) plus body.
- Always include a `git diff --stat .github/workflows` summary in the body so maintainers can see which files changed without opening the diff.
- Detection supports `uvx tend@X.Y.Z` and `tend==X.Y.Z` markers; when either side is empty or both match, the title and body omit the version line (matches the current state where templates use `uvx tend@latest` and `max-sixty/tend@v1`).

Addresses #313.

## Test plan

- [ ] Regex extraction verified against three inputs: `uvx tend@1.2.3 init` → `1.2.3`, `tend==0.0.16` → `0.0.16`, `uvx tend@latest init` → empty.
- [ ] Title rendering verified for bump (`0.0.15 → 0.0.16`), same version, and empty detection — only the bump case adds the suffix.
- [ ] `pre-commit` run on the modified skill: passes (trim-whitespace, typos, bang-backtick guard).
- [ ] End-to-end behavior will be validated on the next nightly run.

### DIFF
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -138,28 +138,70 @@ regeneration in a git worktree under `$TMPDIR`, which is writable:
 
 ```bash
 git worktree add "$TMPDIR/tend-update-workflows" -b tend/update-workflows HEAD
-(cd "$TMPDIR/tend-update-workflows" && uvx tend@latest init)
-git -C "$TMPDIR/tend-update-workflows" status --porcelain .github/workflows
+cd "$TMPDIR/tend-update-workflows"
+
+# Capture the pinned tend version (if any) before regenerating, so the PR
+# body can report the bump. Supports `uvx tend@X.Y.Z` and `tend==X.Y.Z`
+# markers. Today's templates use `uvx tend@latest` and the rolling
+# `max-sixty/tend@v1` action tag, so extraction typically returns an empty
+# string — the renderer below falls back to omitting the version line rather
+# than printing `unknown → unknown`.
+extract_tend_version() {
+  grep -hoE '(uvx tend@|tend==)[0-9]+\.[0-9]+\.[0-9]+' \
+    .github/workflows/tend-*.yaml 2>/dev/null \
+    | sed -E 's/^(uvx tend@|tend==)//' | sort -u | head -1
+}
+OLD_VER=$(extract_tend_version)
+
+uvx tend@latest init
+git status --porcelain .github/workflows
 ```
 
 If `git status` shows no changes, clean up and continue:
 
 ```bash
-git worktree remove "$TMPDIR/tend-update-workflows" --force
-```
-
-If files changed, commit and open the PR from the worktree:
-
-```bash
-cd "$TMPDIR/tend-update-workflows"
-git add .github/workflows/tend-*.yaml
-git commit -m "chore: update tend workflows"
-git push -u origin tend/update-workflows
-gh pr create --title "chore: update tend workflows" \
-  --body "Automated nightly regeneration of tend workflow files."
 cd -
 git worktree remove "$TMPDIR/tend-update-workflows" --force
 ```
+
+If files changed, build the PR title and body with the version bump (when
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -138,28 +138,70 @@ regeneration in a git worktree under `$TMPDIR`, which is writable:
...
 git worktree remove "$TMPDIR/tend-update-workflows" --force
 ```
 
-If files changed, commit and open the PR from the worktree:
+If files changed, build the PR title and body with the version bump (when
+detected) and a `git diff --stat` summary, then commit, push, and open the
+PR:
+
+`````bash
+NEW_VER=$(extract_tend_version)
```
tend-agent: `NEW_VER=$(extract_tend_version)` and the `[ -n "$OLD_VER" ]` condition reference the function and variable defined in the earlier bash block. Per the `running-in-ci` skill and CLAUDE.md, shell state doesn't persist between bash tool calls — so when the agent runs this as a separate invocation from block 1 (the natural pattern, since `git status` output drives the branch), both are gone. `OLD_VER` reads as empty, `extract_tend_version` is undefined, and the version-bump line never renders.

Latent today because the current workflow templates use `uvx tend@latest` — so `OLD_VER`/`NEW_VER` are always empty and the fallback path runs correctly either way. But once a version gets pinned in the templates, the feature would silently no-op.

Two fixes: (1) consolidate into one script with an `if` around the PR-creation branch so everything runs in one bash call; or (2) stash `OLD_VER` to a temp file in block 1 (`echo "$OLD_VER" > "$TMPDIR/tend-old-ver"`) and re-source it plus redefine `extract_tend_version` here.

tend-agent (commented):
One structural concern on the cross-block state dependency — see inline. The regex and title/body rendering themselves look correct, and the issue is latent until templates pin a tend version, but worth addressing now so the feature actually fires when it's needed.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Good catch — fixed in de974f7. Took option 2: `OLD_VER` is now stashed to `$TMPDIR/tend-old-ver` in block 1 and read back in block 3, and the grep is inlined at both extraction sites rather than wrapped in a function (since the function can't survive the bash boundary either).

max-sixty:
another option is to write out the version into the generated files; could we do that?

tend-agent:
Yes — the generator knows its own version via `importlib.metadata.version("tend")`, so we could stamp it into the `HEADER` constant in [generator/src/tend/workflows.py](../blob/main/generator/src/tend/workflows.py) (e.g. `# Generated by tend 0.0.16. Regenerate with: uvx tend@latest init`). That gives detection a single well-defined anchor on every generated workflow regardless of what templates pin, and the nightly skill can just grep one line from `HEAD` vs the regenerated tree.

A few things worth weighing:

- It's a strict superset of what this PR does: header-stamp makes the version visible for tend's own nightly (currently neither template in this repo pins a version, so the dual regex would detect nothing) and for downstream repos that use `uvx tend@latest`.
- Every tend bump would produce a one-line header diff across all `tend-*.yaml` files, which is the signal we want but adds a small amount of regen churn when the template body is otherwise unchanged.
- It makes this PR's extraction logic redundant. Cleanest path is probably to close this and open a two-part change (generator stamp + nightly skill reads the stamp), rather than keeping the template-regex fallback.

Happy to redirect if you'd like me to go that way.

max-sixty:
yes, let's pursue that

tend-agent:
Closing in favor of the header-stamp approach. Will open a replacement PR that adds `Generated by tend X.Y.Z.` to the workflow header and has the nightly skill read it, which is a strict superset of what this PR did.

PR #316, opened 2026-04-19
generator: stamp tend version into workflow header

### LINKED ISSUE #313: nightly: include version bump and diff summary in "update tend workflows" PR body
Requested via max-sixty/worktrunk#2302 — the automated `chore: update tend workflows` PRs currently have a generic body:

> Automated nightly regeneration of tend workflow files.

Maintainers have to open the diff to see what actually changed (e.g. which `uvx tend@…` version was pinned, which workflow files moved). Ask: include the version bump and a short summary of the diff in the PR title/body.

## Proposed change

In `plugins/tend-ci-runner/skills/nightly/SKILL.md`, Step 6 — after `uvx tend@latest init` runs but before `gh pr create`:

1. Detect the old vs. new pinned tend version from the workflow YAML (e.g. `uvx tend@X.Y.Z` or a `tend==X.Y.Z` marker) — compare `HEAD`'s workflows against the regenerated ones.
2. Capture `git diff --stat .github/workflows` (filenames + line counts) for the body.
3. Optionally list per-file section headers touched (jobs added/removed) if cheap to derive — otherwise stick to `--stat`.

Render the PR like:

```
Title: chore: update tend workflows (0.1.0 → 0.1.3)

Body:
Automated nightly regeneration of tend workflow files.

**tend version:** 0.1.0 → 0.1.3

**Changed files:**
\`\`\`
 .github/workflows/tend-nightly.yaml   | 4 ++--
 .github/workflows/tend-review.yaml    | 2 +-
 .github/workflows/tend-triage.yaml    | 6 ++++--
 3 files changed, 7 insertions(+), 5 deletions(-)
\`\`\`
```

If version detection fails (grep finds nothing, or no version pin in the template), fall back to omitting the version line rather than printing `unknown → unknown`.

## Notes

- Keep the title under the 70-char guideline in the CLAUDE.md style — `chore: update tend workflows (X → Y)` is fine.
- Title without versions when the only change is non-version (e.g. a template tweak at the same pinned version) — e.g. `chore: update tend workflows`.

tend-agent:
## Summary

- Stamp the running tend version into every generated workflow header: `# Generated by tend 0.0.16.` instead of `# Generated by tend.`, via `importlib.metadata.version("tend")` in `generator/src/tend/workflows.py`.
- Update `plugins/tend-ci-runner/skills/nightly/SKILL.md` Step 6 to read that anchor from the workflow YAML before and after `uvx tend@latest init`, surface the bump in the PR title (`chore: update tend workflows (A → B)`), and include a `git diff --stat` summary in the body.
- Anchor is stable regardless of what templates pin for `uvx tend@…` or the `max-sixty/tend@v1` action tag, so detection works on tend itself (where neither template embeds a concrete version today) and on downstream repos that pin to `latest`.
- Pre-stamp workflows produce an empty `OLD_VER`; renderer omits the version line when either side is empty or both sides match, so the transitional regen (before / after this ships) won't print `unknown → 0.0.17`.
- Closes #313. Supersedes #314 ([discussion](https://github.com/max-sixty/tend/pull/314#issuecomment-4276497219)).

## Test plan

- [ ] `uv run pytest` in `generator/` — 164 passed (regtest snapshots refreshed for all 8 workflows × minimal + with-setup parameterizations).
- [ ] `pre-commit run` on touched files — trim whitespace, ruff, typos, bang-backtick guard all pass.
- [ ] End-to-end: first nightly run after this ships stamps the regenerated workflows with the published version; the subsequent nightly (once the publish has made `uvx tend@latest` carry this change) will be the first to show `A → B` when tend is bumped.


tend-agent (commented):
The stamp-in-header approach cleanly supersedes the template-regex fallback from #314 — one anchor per file, independent of what templates pin. Shell logic in `plugins/tend-ci-runner/skills/nightly/SKILL.md` handles the three transitional cases correctly (empty OLD, equal OLD/NEW, real bump), and the semver ERE matches the format `importlib.metadata.version("tend")` produces for the current `pyproject.toml`.

One release-maintenance concern worth flagging: the regtest snapshots now pin the literal version string (`# Generated by tend 0.0.16.` × 16 files). Every subsequent `version = "X.Y.Z"` bump in `generator/pyproject.toml` will fail `uv run pytest` until all 16 snapshots are regenerated with `--regtest-reset`. The release skill at `.claude/skills/release/SKILL.md` doesn't currently mention this step. Two options:

- Scrub the version in the existing `conftest.py` `register_converter_pre` hook (e.g. `re.sub(r'# Generated by tend [^.]+\.', '# Generated by tend X.Y.Z.', output)`), so regtests only assert on structural content. `test_extras_preserve_header` already covers the "header is present" case via `startswith("# Generated by tend ")`.
- Or add a snapshot-regen step to the release skill.

The first avoids ongoing churn entirely; the second accepts it as explicit release friction. Not a blocker either way.

tend-agent:
Applied option 1 — added a `register_converter_pre` scrubber in `generator/tests/conftest.py` that rewrites the stamped version to `X.Y.Z` before snapshot comparison, and regenerated the 16 outputs. `test_extras_preserve_header` continues to cover "header is present". Future `version = ...` bumps in `generator/pyproject.toml` no longer touch regtest snapshots.

PR #318, opened 2026-04-19
skills: restore bang-escape warning (bug still fires in CI harness)

tend-agent:
PR #317 removed the `!`-escape heredoc warning from `running-in-ci` citing a local verification that Claude Code no longer rewrites `!` in heredocs. That verification doesn't hold in the GitHub Actions CI harness (Claude Code 2.1.114). The Bash tool still rewrites every `!` to `\!` in this run.

**Reproduction in this run** ([24641241367](https://github.com/max-sixty/tend/actions/runs/24641241367)):

```
$ cat << 'OUTER'
unquoted-1: Hello!
OUTER
unquoted-1: Hello\!

$ cat << 'INNER'
single-2: Hello!
INNER
single-2: Hello\!

$ echo 'plain-single: select(.x != "y")'
plain-single: select(.x \!= "y")

$ echo "plain-double: select(.x != \"y\")"
plain-double: select(.x \!= "y")
```

All four contexts — `<<EOF`, `<< 'EOF'`, plain single-quoted, and double-quoted — still rewrite the character. The bug isn't heredoc-specific; it's at the Bash-tool preprocessor layer.

**Why this matters:** the removed warning was the guard against bot comments shipping with literal `\!` corruption. With it gone, a future bot writing `"Thanks for the suggestion!"` via `gh ... --body` will publish `"Thanks for the suggestion\!"` — the exact user-visible failure mode the warning prevented. No corrupted comments have slipped through in the past ~4 hours since #317 merged, but that's luck, not correctness.

**Changes:**
- `running-in-ci`: restore the `!`-escape paragraph; expand it to cover plain single/double-quoted args (the case that bites `--jq 'select(.x != …)'` commands — already hit 3 times, recorded in the [evidence gist](https://gist.github.com/8f10d7c1abcb07c79bef00ee6d02d315)).
- `review-reviewers`: restore "unescaped bangs" to the corruption-signals list.

**Gate assessment:**
- Evidence level: **High / Critical**. Structural bug in the tool layer; reproduction is deterministic (4/4 contexts fail in this run).
- Occurrences: 1 structural reproduction + 3 cumulative `--jq` occurrences in the evidence gist.
- Change type: Targeted fix (restore removed paragraph, one-word addition in `review-reviewers`).
- Both gates pass.

**Caveat for the maintainer**: if the bug really is fixed in your local Claude Code but not in the CI-pinned version, the right long-term fix may be bumping whatever version the runner installs rather than keeping workaround guidance. Either way, removing the warning while the bug still fires in CI was premature.

### DIFF
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -290,7 +290,7 @@ Use `Agent` with `model: "haiku"` and a prompt like:
 > - Bot-closed issue was reopened
 > - Fix commit was reverted or CI still failing after bot pushed
 > - Human replied to bot with correction or complaint
-> - Bot comment contains corruption (literal `${`, broken heredoc markers)
+> - Bot comment contains corruption (literal `${`, unescaped bangs, broken heredoc markers)
 >
 > **Report format** — return a structured summary:
 > ```
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -351,6 +351,15 @@ expanded** — if you see a literal `${GITHUB_REPOSITORY}` in the rendered comme
 single-quoted heredoc (`<< 'EOF'`) which disables expansion. Rewrite using an unquoted `<<EOF`
 (so `${GITHUB_REPOSITORY}` interpolates) or compose the body with the Write tool.
 
+The Bash tool rewrites every exclamation mark to a literal backslash-bang before bash parses
+the command, so a greeting like "Thanks for the suggestion!" renders as "Thanks for the
+suggestion\!" in the posted comment. Quoting and heredoc form don't matter: `<< 'EOF'`,
+`<<EOF`, plain single-quoted, and double-quoted arguments all lose the character. Use the
+Write tool for any comment body containing an exclamation mark, then pass the file to
+`gh ... --body-file`. The same trap applies to `--jq` filters with `!=` — either avoid the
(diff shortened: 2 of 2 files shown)


tend-agent:
Confirmed in the wild — the corruption this PR predicted has now shipped publicly.

`tend-triage` run [24641926108](https://github.com/max-sixty/worktrunk/actions/runs/24641926108) on `max-sixty/worktrunk` posted [issues/2309#issuecomment-4277108328](https://github.com/max-sixty/worktrunk/issues/2309#issuecomment-4277108328) at 2026-04-19T23:53:23Z. The rendered comment opens `Thanks for the detailed report\!` — a literal backslash before the bang, exactly the failure mode this PR's body warns about. Same pipeline: model wrote `cat > /tmp/issue-2309-comment.md << 'EOF'` followed by plain `Thanks for the detailed report!`, then `gh issue comment --body-file`. Single-quoted heredoc, plain bang in the model's tool input (verified in the run's session JSONL) — yet the file landed with the backslash, and the comment was posted as written.

So the count is now 1 structural reproduction (the prior run on tend) + 3 cumulative `--jq` occurrences in the evidence gist + **1 corrupted public comment on a community-reported bug** (worktrunk #2309 was filed by `nicolasff`, a non-bot reporter). The "no corrupted comments slipped through" caveat in the PR body no longer holds; the gap closed within ~6 hours of #317 merging.

Live demonstration also fired in this run while I was scanning bot output — a Python heredoc with `if n != 406:` was rewritten to `if n \!= 406:` and crashed with `SyntaxError: unexpected character after line continuation character`. Same Bash-tool layer, same rewrite, in the very environment running the analysis.

Recording the worktrunk #2309 incident in the [worktrunk evidence gist](https://gist.github.com/44ab1483b29406255dd36141650c894d) under counter `bot comment posted with literal \! corruption`.

PR #327, opened 2026-04-23
running-in-ci: verify external-tool behavior by run or clone, not just hand-test

### LINKED ISSUE #326: running-in-ci: flag external-tool behavioral claims that need hand-testing
## Problem

A mention-response run on worktrunk [PR #1907](https://github.com/max-sixty/worktrunk/pull/1907) was asked whether an external CLI (`cmux list-workspaces`) had a structured output mode. The bot confidently claimed it did, rewrote the recipe to `cmux list-workspaces --json | jq ...`, and committed it. A user later pointed out ([comment](https://github.com/max-sixty/worktrunk/pull/1907#issuecomment-4304550281)) that cmux 0.63.2 has no `--json` flag — the published recipe would break for every reader who tried it.

The bot had no way to hand-test this: cmux is a macOS terminal app, not installed in CI, and the change was documentation-only so no automated test in worktrunk's suite could catch a fabricated CLI flag. The only signal that would have rejected the claim was running `cmux list-workspaces --help` or `--json` on a machine with cmux installed.

## Root cause

When a review asks about behavior of an external tool (third-party CLI, API, or system), the bot treats reading upstream docs (or inferring from a `--help` that doesn't exist in CI) as verification. For external tools specifically, neither source applies:

- Upstream docs for fast-moving tools lag or describe flags that were removed/renamed.
- Reading code is not an option — the tool is a separate project.
- Integration tests in the consuming repo don't exercise the tool.

None of the guidance in `running-in-ci.md` (Grounded Analysis) fires on this path. "User-facing comments require source evidence" talks about docs and code, but for a third-party tool not installed in CI, the only authoritative source is invoking the tool itself.

## Proposed guidance

Add a check to the bundled `running-in-ci` skill (or a new section) along these lines:

> **External-tool behavioral claims need a hand-test.** If a review asks whether an external CLI, API, or system behaves a certain way — and that tool is not installed in CI and not exercised by automated tests — you cannot verify the answer from your environment. Do not post a confident claim or change code based on it. Instead:
>
> 1. Quote the reviewer's question back and explicitly note that the behavior needs hand-testing.
> 2. If you have partial evidence (upstream docs, a commit in the external project, a GitHub issue), cite it and hedge: "According to X's docs, Y — but I haven't verified on a real install. Could someone with X installed confirm before I push a change?"
> 3. Do not ship a doc change or code change whose correctness depends on this claim until a human confirms.

The same pattern applies to triage runs that repro a bug using a tool not in CI: the bot should flag the repro as "unverified" rather than asserting a fix.

## Evidence

- Triggering run: [worktrunk-bot comment claiming JSON support](https://github.com/max-sixty/worktrunk/pull/1907#issuecomment-4292844468) linking a mintlify page (`https://manaflow-ai-cmux.mintlify.app/cli/workspaces`) that described a `--json` flag not present in cmux 0.63.2.
- User correction with screenshot: [kay-schecker's comment](https://github.com/max-sixty/worktrunk/pull/1907#issuecomment-4304550281).
- Maintainer request: [max-sixty asking for this issue](https://github.com/max-sixty/worktrunk/pull/1907#issuecomment-4306981159).

tend-agent:
## Problem

The bundled `running-in-ci` skill's Grounded Analysis section covers behavioral claims by telling the bot to "run the command yourself" or hedge — but that fallback doesn't obviously apply to external CLIs and APIs that aren't installed in CI and aren't exercised by automated tests. In [worktrunk#1907](https://github.com/max-sixty/worktrunk/pull/1907#issuecomment-4292844468) the bot read upstream mintlify docs describing a `cmux list-workspaces --json` flag, believed it, and committed a recipe that broke for every reader — the installed cmux had no such flag.

## Solution

Adds a new `### Verifying external-tool behavior` subsection under Grounded Analysis. It points at two concrete verification paths in order of preference: install and run the tool, or clone its public repo and grep the source. Deferring to a human with the tool installed is the fallback only when both paths fail. The "don't make overconfident claims" framing already lives in the preceding `### User-facing comments require source evidence` subsection.

Bad/good example is drawn from the cmux incident: the good path now shows cloning the upstream repo and checking the CLI parser, not asking a human to confirm.

## Testing

Skill text only. `pre-commit run` on the modified file passes (typos, trim-whitespace, bang-backtick, end-of-files).

---
Closes #326 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -501,8 +501,7 @@ copy-pasted.
 ### Verifying external-tool behavior
 
 When a claim turns on how an external CLI, API, or system behaves, verify by running the
-code. Upstream docs for fast-moving tools lag or describe flags that were removed or
-renamed — don't treat them as proof on their own.
+code.
 
 Two paths, in order of preference:
 
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -498,6 +498,41 @@ If you can't find source evidence for a specific detail, say so ("I'm not sure o
 syntax") rather than guessing. An honest gap is fixable; a confident hallucination gets
 copy-pasted.
 
+### External-tool behavioral claims need a hand-test
+
+When a review or triage turns on how an external CLI, API, or system behaves — and that tool
+is not installed in CI and not exercised by automated tests — the "run the command yourself"
+fallback above doesn't apply. Upstream docs for fast-moving tools lag or describe flags that
+were removed or renamed, and reading the tool's source isn't an option when it lives in a
+separate project. Do not post a confident claim or commit code/doc changes that depend on
+the behavior.
+
+1. Name the gap. Quote the question back and state that the behavior needs hand-testing on
+   a machine with the tool installed.
+2. If you have partial evidence (upstream docs, a commit, a linked issue), cite it and
+   hedge: "According to X's docs, Y — but I haven't verified on a real install. Could
+   someone with X installed confirm before I push a change?"
+3. Don't ship a doc or code change whose correctness depends on the claim until a human
+   confirms. In triage, the same rule applies: if a repro uses a tool not in CI, flag the
+   repro as unverified rather than asserting a fix.
+
+<example>
+<bad reason="Trusted upstream docs for a fast-moving external CLI and shipped a broken recipe">
+
+Bad: Review asked whether `cmux list-workspaces` had structured output. Read a mintlify
+page describing `--json` → rewrote the recipe to `cmux list-workspaces --json | jq ...` →
+committed. The installed cmux had no `--json` flag; every reader hit a broken recipe.
+
+</bad>
+<good reason="Named the verification gap and deferred to a human with the tool installed">
+
+Good: Same question. Read the docs → replied: "The docs describe `--json`, but cmux isn't
+installed in CI so I can't verify against your version. Could you confirm
+`cmux list-workspaces --help` shows `--json` before I push the change?" Waited.
+
+</good>
+</example>
+
 ### Rewriting is authoring
 
 Cross-posting, summarizing, or paraphrasing is not copying — any new content you add requires the
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -498,23 +498,23 @@ If you can't find source evidence for a specific detail, say so ("I'm not sure o
 syntax") rather than guessing. An honest gap is fixable; a confident hallucination gets
 copy-pasted.
 
-### External-tool behavioral claims need a hand-test
+### Verifying external-tool behavior
 
-When a review or triage turns on how an external CLI, API, or system behaves — and that tool
-is not installed in CI and not exercised by automated tests — the "run the command yourself"
-fallback above doesn't apply. Upstream docs for fast-moving tools lag or describe flags that
-were removed or renamed, and reading the tool's source isn't an option when it lives in a
-separate project. Do not post a confident claim or commit code/doc changes that depend on
-the behavior.
+When a claim turns on how an external CLI, API, or system behaves, verify by running the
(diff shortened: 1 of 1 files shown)


max-sixty (commented):
overall decent, but:
- the focus should be on verifying claims, if possible by running code
- it is possible to read a tool's source! tend can clone the repo!
- fine to retain the "don't make overconfident claims" but we should have that elsewhere

tend-agent:
Good points, rewrote in [6437170](https://github.com/max-sixty/tend/commit/6437170):

- Section now leads with verifying by running the tool (install-and-run) or reading source (`gh repo clone` then grep). Deferring to a human is the fallback only when both fail.
- Retitled to `### Verifying external-tool behavior` — positive framing, since the primary action is verification rather than flagging a gap.
- Dropped the "do not post a confident claim" language; that rule already lives in the preceding `### User-facing comments require source evidence` subsection, so no new home needed.
- Updated the good example to show cloning cmux's source and checking the CLI parser rather than asking a human to confirm.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -498,6 +498,41 @@ If you can't find source evidence for a specific detail, say so ("I'm not sure o
 syntax") rather than guessing. An honest gap is fixable; a confident hallucination gets
 copy-pasted.
 
+### Verifying external-tool behavior
+
+When a claim turns on how an external CLI, API, or system behaves, verify by running the
+code. Upstream docs for fast-moving tools lag or describe flags that were removed or
+renamed — don't treat them as proof on their own.
```
max-sixty:
```suggestion
code.
```

max-sixty:
was this the actual problem in this specific case — that the `cmux` docs were wrong? if so, this is fine. if not, and actually we hallucinated something, then we should adjust

tend-agent:
Checked both sources. The mintlify page at `manaflow-ai-cmux.mintlify.app/cli/workspaces` does describe a `--json` flag for `list-workspaces`, with the exact JSON shape I posted (a `ParamField path="--json"` block plus a sample `{"workspaces": [{"ref": ..., "id": ..., "index": ..., "title": ..., "selected": ...}]}` payload). So I wasn't synthesizing the schema — I quoted what the docs said.

The cmux source tells the other half of the story. In [`CLI/cmux.swift`](https://github.com/manaflow-ai/cmux/blob/main/CLI/cmux.swift) the `list-workspaces` help block is just `Usage: cmux list-workspaces` with no flags; `--json` is only wired up on `tree` (and a few others). So the published docs are ahead of (or diverged from) the shipped CLI parser.

Net: docs were wrong, not a hallucination — but the new section's "clone the repo and grep the CLI parser" step would have caught it, since the discrepancy is visible in one `grep --json CLI/cmux.swift`. So the framing in this PR holds. Happy to file a docs issue with manaflow-ai/cmux if you'd like, separately.

tend-agent (commented):
Small correction to my earlier comment for the record: `--json` is in fact wired into `list-workspaces` at `CLI/cmux.swift:2003-2005` (even at tag `v0.63.2`), via the shared `if jsonOutput` block.

The real reason the published recipe broke: cmux parses `--json` as a **global** flag in a loop that `break`s at the first non-global token (see `CLI/cmux.swift:1665-1710`). So `cmux list-workspaces --json` silently drops `--json` and returns text; the form that works is `cmux --json list-workspaces`. The mintlify docs show the broken flag ordering, which is what I copied.

Still "docs wrong, not hallucination" — but a naïve `grep -- --json CLI/cmux.swift` would have confirmed `--json` works for `list-workspaces` and shipped the same broken recipe. Path 1 (run the tool) is what would have caught this; Path 2 alone misses flag-placement bugs unless you also read the arg-parser structure. Worth keeping that caveat in mind when this guidance gets applied to other fast-moving CLIs.

max-sixty:
ok, can you add a comment on the original issue that the `--json` flag is available if expressed differently?

tend-agent:
Done — [worktrunk#1907 comment](https://github.com/max-sixty/worktrunk/pull/1907#issuecomment-4319994298) explains that `--json` does work for `list-workspaces` but only as a global flag (`cmux --json list-workspaces`), with source pointers at `CLI/cmux.swift:1635-1690` (parser) and `:2003-2006` (handler).

PR #328, opened 2026-04-25
running-in-ci: skill-edit recipe should use worktree under $TMPDIR

tend-agent:
## Summary

Update the `running-in-ci` "How to propose" recipe so a bot updating `.claude/skills/` can actually write to disk. The current recipe (`cd .claude && mv /tmp/... SKILL.md`) satisfies the Claude Code harness write-guard ([anthropics/claude-code#37157](https://github.com/anthropics/claude-code/issues/37157)) but fails the read-only bind-mount the sandbox places on `.claude/`. Mirror the pattern `review-runs` already documents: do the edit, commit, and push from a git worktree under `$TMPDIR`. Both restrictions are now explained inline so a future reader knows why the worktree is mandatory.

## Why

`worktrunk-bot` hit this in worktrunk run [24926120125](https://github.com/max-sixty/worktrunk/actions/runs/24926120125) (a `tend-mention` session triggered by an explicit maintainer instruction to update repo skill guidance). The bot followed the bundled recipe, hit `mv: ... Read-only file system`, retried with `cp` (same error), retried with `touch` (same error), then ad-libbed `git hash-object -w` plus `git update-index --cacheinfo` to write the blob and stage it directly to the index — a workaround that left the working tree showing the file as modified after the fact. ~7 wasted bash cycles before producing [worktrunk PR #2415](https://github.com/max-sixty/worktrunk/pull/2415).

`worktrunk-bot` followed the new bundled-skill-defect flow (introduced in #324 yesterday) and opened permission-request issue [max-sixty/worktrunk#2416](https://github.com/max-sixty/worktrunk/issues/2416) on the consumer side. This PR is the corresponding tend-side fix surfaced by `review-reviewers` rather than waiting on the permission round-trip — the diagnosis is unambiguous and the fix is small.

## What changes

The "Draft a minimal edit" step (step 3 of "How to propose"):

- Drops the `cd .claude && mv /tmp/... SKILL.md` recipe that fails on the read-only mount.
- Adds the same prose `review-runs` already carries explaining (a) the read-only bind-mount of `.claude/` and (b) the harness write-guard from claude-code#37157, so a reader sees both barriers and why a worktree clears both.
- Replaces the recipe with the `git worktree add "$TMPDIR/skill-fix"` pattern, parameterised on `<topic>-$GITHUB_RUN_ID` for the branch name.
- Keeps the existing TODO referencing claude-code#37157.

The frontmatter snippet is moved up so the recipe sits at the bottom of step 3 with no interleaving.

## Test plan

- [x] Eaten own dogfood: this PR's branch was authored from a worktree at `$TMPDIR/skill-fix` using exactly the recipe being landed; commit and push worked end-to-end. Worktree under `$TMPDIR` is writable, no `Read-only file system` errors.
- [ ] CI on this PR passes (actionlint, ruff, typos, uv-lock).
- [ ] Visual diff vs `review-runs` SKILL.md prose to confirm the two skills now agree on the recipe shape.

Independent of [#327](https://github.com/max-sixty/tend/pull/327) (different concern in the same file: external-tool verification, lines ~498 — no overlap with this edit at lines ~568).

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -577,27 +577,44 @@ Include in the permission request (and reuse verbatim in the tend issue once app
    ```
    If one is open, add to it instead of opening a second.
 3. **Draft a minimal edit.** One short rule, in the maintainer's words where practical. Place
-   it under an appropriate heading. Use the `Write` tool to author the full new contents of
-   the skill file to `/tmp/running-tend-new.md`, then move it into place via Bash:
-   <!-- TODO(anthropics/claude-code#37157): remove the cd-and-mv workaround once the
-        harness exempts .claude/skills/ as documented — then `Write` directly to the
-        target path. -->
-   ```bash
-   cd .claude && mkdir -p skills/running-tend && cd skills/running-tend \
-     && mv /tmp/running-tend-new.md SKILL.md
-   ```
-   The `cd` is required. Claude Code's harness blocks `Edit`, `Write`, and Bash commands
-   whose write-target argument is a path under `.claude/skills/`
-   ([anthropics/claude-code#37157](https://github.com/anthropics/claude-code/issues/37157)).
-   The guard checks the argument text — `Write(/tmp/…)` and `Bash(mv /tmp/… SKILL.md)`
-   both pass because neither names the protected path.
-
-   New SKILL.md files start with YAML frontmatter:
+   it under an appropriate heading. New SKILL.md files start with YAML frontmatter:
    ```markdown
    ---
    name: running-tend
    description: Project-specific guidance for tend workflows running on this repo.
    ---
+   ```
+
+   The checkout's `.claude/` directory is bind-mounted **read-only** under the sandbox
+   (protecting bots from modifying their own skills in place), so edits to `.claude/skills/`
+   files in the working tree fail with `Read-only file system`. Claude Code's harness adds a
+   second restriction on top of the read-only mount: `Edit`, `Write`, and Bash commands with
+   `.claude/skills/` as a write-target argument are denied regardless of filesystem
+   permissions ([anthropics/claude-code#37157](https://github.com/anthropics/claude-code/issues/37157)).
+   The guard checks argument text, so `Write(/tmp/…)` and `Bash(mv /tmp/… SKILL.md)` both
+   pass — the second because `SKILL.md` is a bare filename inside the `cd`'d directory.
+
+   Do the edit, commit, and push from a git worktree under `$TMPDIR`, which is writable and
+   sits outside the harness's `.claude/skills/` write-guard:
+
+   <!-- TODO(anthropics/claude-code#37157): once the harness exempts .claude/skills/ as
+        documented, replace the /tmp-then-mv dance below with direct `Write` to the worktree
+        path. -->
+
+   ```bash
+   git worktree add "$TMPDIR/skill-fix" -b skills/<topic>-$GITHUB_RUN_ID HEAD
+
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -577,28 +577,45 @@ Include in the permission request (and reuse verbatim in the tend issue once app
...
+   <!-- TODO(anthropics/claude-code#37157): once the harness exempts .claude/skills/ as
+        documented, replace the /tmp-then-mv dance below with direct `Write` to the worktree
+        path. -->
+
+   ```bash
+   git worktree add "$TMPDIR/skill-fix" -b skills/<topic>-$GITHUB_RUN_ID HEAD
+
+   # Use the Write tool to author the new skill file to /tmp/running-tend-new.md.
+   # Then move it into place from inside the worktree:
+   cd "$TMPDIR/skill-fix/.claude/skills/running-tend" && mv /tmp/running-tend-new.md SKILL.md
```
tend-agent, line 611:
```suggestion
   mkdir -p "$TMPDIR/skill-fix/.claude/skills/running-tend"
   cd "$TMPDIR/skill-fix/.claude/skills/running-tend" && mv /tmp/running-tend-new.md SKILL.md
```

tend-agent (commented):
The previous recipe used `mkdir -p skills/running-tend` so the same shape worked for both editing an existing skill and creating a new one. The new recipe drops that, but step 3 still contemplates new-skill creation (the YAML frontmatter snippet immediately above). For a brand-new skill, `cd "$TMPDIR/skill-fix/.claude/skills/<new>"` fails with "No such file or directory". Inline suggestion to restore the `mkdir -p`.

Note this gap also exists in `review-runs/SKILL.md`, but that skill is purely an editing flow so it's defensible there.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Good catch — pushed fd3f7a6 with `mkdir -p "$TMPDIR/skill-fix/.claude/skills/running-tend"` before the `cd`. The worktree mirrors HEAD, so for a brand-new skill the directory genuinely doesn't exist and the bare `cd` fails. The dogfood test plan only exercised the existing-skill path, which is why this slipped through.

max-sixty:
did we test whether this approach works?

tend-agent:
Yes — re-verified in this session by running the recipe end-to-end against this checkout:

```
git worktree add "$TMPDIR/skill-test" -b test/dogfood-recipe HEAD
mkdir -p "$TMPDIR/skill-test/.claude/skills/test-new"
cd "$TMPDIR/skill-test/.claude/skills/test-new" && mv /tmp/.../content.md SKILL.md
cd "$TMPDIR/skill-test" && git add .claude/skills/test-new/SKILL.md && git commit -m ...
```

All four steps succeeded: worktree create, `mkdir -p`, `mv` of a `Write`-authored file from `/tmp/claude/`, and `git add` + `git commit` of the new `.claude/skills/test-new/SKILL.md`. No `Read-only file system` errors, and no harness write-guard denials. The two restrictions the prose now describes — `.claude/` bind-mount and the claude-code#37157 argument-text guard — both clear cleanly when the path is rewritten to `$TMPDIR/skill-test/.claude/skills/...`.

Skipped `git push` (test branch only); the surrounding flow is the same one `review-runs` already ships and uses in production.

Limitation: I can't independently verify the PR-description claim that 02a28bc and fd3f7a6 were *originally* authored from a worktree — git doesn't record which working tree a commit came from — but the commits exist on this branch and the mechanical steps reproduce.

PR #345, opened 2026-04-27
skills(review-reviewers,review-runs): retry tracking-issue lookup on empty result

tend-agent:
## Summary

Retries the `gh issue list` tracking-issue lookup up to 3 times before treating empty as ground truth. Same fix applied to both `review-reviewers` and `review-runs` SKILL.md (identical lookup code in both).

## Evidence

GitHub's `/issues` endpoint is intermittently returning `[]` for queries whose matching issues clearly exist. Reproduced empirically during analysis run [25009161877](https://github.com/max-sixty/tend/actions/runs/25009161877):

```
=== gh issue list (5 attempts, same command) ===
Attempt 1: ''
Attempt 2: '133'
Attempt 3: ''
Attempt 4: ''
Attempt 5: '133'

=== Raw API count ===
Attempt 1: count=0, []
Attempt 2: count=0, []
Attempt 3: count=2, [#342, #133]
Attempt 4: count=2, [#342, #133]
Attempt 5: count=0, []
```

Response ETags differ between empty and populated responses, suggesting different read replicas/cache entries serve the same query inconsistently. `gh search issues`, `gh api search/issues`, and `gh api repos/.../issues` (with and without label filter) all show the same flake — it's at the GitHub API layer, not specific to one endpoint.

## Outcome impact (today, max-sixty/tend)

Two duplicate tracking issues created within one hour today on top of pre-existing tracking issue [#133](https://github.com/max-sixty/tend/issues/133):

- [#342](https://github.com/max-sixty/tend/issues/342) created at 16:18:06Z by review-reviewers run [25006456679](https://github.com/max-sixty/tend/actions/runs/25006456679) (max-sixty/worktrunk matrix cell)
- [#344](https://github.com/max-sixty/tend/issues/344) created at 17:16:05Z by review-reviewers run 25009161877 (max-sixty/tend matrix cell — this analysis run)

Each duplicate triggered a `tend-triage` run that exited silently after recognizing the tracking-label shape:

- [25006564650](https://github.com/max-sixty/tend/actions/runs/25006564650) — triage on #342, no comment posted, ~$0.10 wasted
- [25009220181](https://github.com/max-sixty/tend/actions/runs/25009220181) — triage on #344, same shape

## Gates

- **Confidence**: High — 2 occurrences in one hour today (independent matrix runs), reproduced live (3 of 5 manual attempts returned empty for known-existing issues).
- **Classification**: Structural — the script is non-resilient to a known API behavior; same conditions reproduce reliably.
[...]

PR #348, opened 2026-04-27
fix(scripts): use unfiltered gh run list to dodge workflow-filter index lag

tend-agent:
## Summary

`scripts/list-recent-runs.sh` invokes `gh run list --workflow X --created ">=Y"` once per discovered workflow. On gh 2.89.0 (released 2026-03-26) that combination is unreliable — the same query intermittently returns `[]`, returns a stale subset, or returns the correct set across consecutive calls. When it returns `[]`, the script silently produces zero runs and the calling skill's analysis bypasses the entire window without reporting the gap.

This swap replaces the per-workflow loop with one global `gh run list --limit 200` call (no `--workflow`, no `--created`) and filters by workflow name prefix and updatedAt timestamp client-side. The unfiltered call returns the correct, fresh set every time.

## Reproducer

On a repo with recent `tend-*` activity (e.g. `max-sixty/worktrunk` at the time of writing — current `gh` 2.89.0):

```
$ for i in 1 2 3; do
    gh run list -R max-sixty/worktrunk --workflow tend-notifications \
      --created '>=2026-04-27T00:00:00' --limit 50 --json databaseId | wc -c
  done
3        # []
3        # []
1117     # partial set, missing the most recent runs
```

vs. the unfiltered call which returns the same fresh result on every retry:

```
$ for i in 1 2 3; do
    gh run list -R max-sixty/worktrunk --workflow tend-notifications \
      --limit 50 --json databaseId | wc -c
  done
3101
3101
3101
```

The same flakiness reproduces with `--workflow X --status Y` and `--workflow X --branch Y`, so this is a workflow-scoped index lag in the GitHub Actions API rather than a `--created` quirk specifically.

## Evidence

- review-reviewers run [25011739890](https://github.com/max-sixty/tend/actions/runs/25011739890) (this PR) — `list-recent-runs.sh` returned `[]` despite 9 in-window `tend-*` runs on `max-sixty/worktrunk`. Worked around by issuing the global call and filtering client-side, which surfaced 4 fresh runs to analyze.
- review-reviewers run [25009161877](https://github.com/max-sixty/tend/actions/runs/25009161877) (~1h earlier on the same target) hit the same silent-empty behavior. That run noted it as a "Bonus observation" without a tend-side action because the failure mode wasn't reproducible at the time. The clean three-try reproducer above closes that gap.

Evidence gist: https://gist.github.com/44ab1483b29406255dd36141650c894d (this run's entry records the 2nd occurrence and pass through both gates).

## Gate assessment

- **Confidence**: High — structural failure (deterministic enough to reproduce on demand), 2 occurrences within a few hours.
- **Magnitude**: Targeted fix — one script, narrow change.
- Both gates pass.

## Test plan

- [x] `TARGET_REPO=max-sixty/worktrunk bash plugins/tend-ci-runner/scripts/list-recent-runs.sh` returns the 4 expected in-window runs (verified against `gh run list --limit 100` cross-check).
- [x] Passing a prefix arg (`tend-notifications`) narrows the result correctly.
- [x] `pre-commit run --files plugins/tend-ci-runner/scripts/list-recent-runs.sh` passes.
- [ ] Next review-reviewers run on a target with recent `tend-*` activity returns a non-empty list rather than silently exiting.

### DIFF
--- a/plugins/tend-ci-runner/scripts/list-recent-runs.sh
+++ b/plugins/tend-ci-runner/scripts/list-recent-runs.sh
@@ -1,11 +1,19 @@
 #!/usr/bin/env bash
 # Lists recently completed Claude CI runs.
 #
-# Fetches runs started in the past 3 hours, then filters to only those that
-# are completed and whose updatedAt is within the past hour. This two-step
-# approach is needed because `gh run list --created` filters by *start* time,
-# not *end* time — a run started 2h ago may have just finished, and a run
-# started 50min ago may still be running. See #1301 for details.
+# Fetches a window of recent runs via one global `gh run list` call and
+# filters client-side by workflow name prefix and updatedAt timestamp. The
+# script keeps the past-1-hour completion cutoff: a run started 2h ago may
+# have just finished, and a run started 50min ago may still be running, so
+# we filter on updatedAt rather than createdAt.
+#
+# Why a global call instead of `--workflow X --created Y` per workflow:
+# combining `--workflow` with secondary filters (`--created`, `--status`,
+# `--branch`) is unreliable on gh 2.89.0. The same query can return [],
+# return a stale subset, or return the correct set across consecutive calls
+# — likely a workflow-scoped index lag in the GitHub Actions API. When the
+# call returns [], this script silently produces zero runs and the analysis
+# bypasses the entire window. Filtering client-side avoids the broken path.
 #
 # Environment variables:
 #   TARGET_REPO - Query a different repo (default: current repo)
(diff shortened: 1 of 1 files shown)


tend-agent:
Closing as a variant of #345 per https://github.com/max-sixty/tend/pull/345#issuecomment-4332788551 — same GitHub search-degradation incident on 2026-04-27 (16:31–22:46 UTC) caused the workflow-filter flakiness reproduced here. The unfiltered-list workaround is mitigating a server-side incident, not a gh bug.

PR #369, opened 2026-05-02
skills(ci-fix): dedup against recently closed fix/ci-* PRs

### LINKED ISSUE #360: ci-fix: also dedup against recently closed `fix/ci-*` PRs
## Problem

The `ci-fix` skill's "Check for existing fixes" step (Step 1) only queries **open** `fix/ci-*` PRs:

```bash
gh pr list --state open --head "fix/ci-" --json number,title,headRefName
```

When a maintainer closes a workaround PR with a rejection rationale (e.g. "we'll fix upstream", "won't fix"), the next CI failure on the same root cause re-runs `ci-fix`, sees no open PR, and re-derives a near-duplicate of the just-rejected workaround. The maintainer ends up closing the same fix twice.

## Concrete example

In `max-sixty/worktrunk`:

1. `collect affected coverage` fails on every push to main (`cargo-affected runner-shim … target-name fallback ambiguous`).
2. [worktrunk#2493](https://github.com/max-sixty/worktrunk/pull/2493) proposed a workaround (renaming helper-crate dummy test files); closed within ~8 hours with "we'll fix upstream".
3. The next push (`90d5cfcb`) hit the same failure and triggered [run 25170507538](https://github.com/max-sixty/worktrunk/actions/runs/25170507538). With only the open-PR check, ci-fix would have re-derived the same workaround. A repo-local override caught it; this issue is to move that check into the bundled skill so every consumer benefits.

## Proposed fix

Extend Step 1 to also look up recent closed bot-authored `fix/ci-*` PRs and shape-match the new failure diagnostic against their body:

```bash

# Open dedup (existing):
gh pr list --state open --head "fix/ci-" --json number,title,headRefName

# Closed dedup (new) — last ~14 days, bot-authored:
gh pr list --state closed --author "$BOT_LOGIN" --search "head:fix/ci-" \
  --json number,title,closedAt,body,comments,headRefName \
  --jq '[.[] | select((now - (.closedAt | fromdateiso8601)) < 1209600)] | .[]'
```

Match by **failure shape** (the diagnostic snippet embedded in the PR body) rather than branch name — branch names encode run IDs and never repeat. The closure comment / review usually carries the maintainer's rationale.

If a closed PR with a maintainer rejection covers the same failure, exit silently. The maintainer already saw the diagnosis; another PR is noise.

### Two gotchas worth codifying

- Use the `head:fix/ci-` **search qualifier** for the closed-state lookup, not `in:head` (silently ignored — falls back to default-field text matching and returns nightly/daily PRs containing "fix" or "ci"). `--head "fix/ci-"` works for `--state open` because gh filters client-side, but the search-API path needs `--search "head:fix/ci-"`.
- Request `body` in `--json`. That field carries the failure diagnostic written by the prior ci-fix run; without it there's nothing to shape-match.

## Why upstream

The rejection-dedup pattern is generic across all tend consumers — the rationale ("fix upstream") happens to be repo-local, but the mechanism for not re-proposing a just-rejected fix is not.

Filed at @max-sixty's [direction in worktrunk#2499](https://github.com/max-sixty/worktrunk/pull/2499#discussion_r3169641478).

tend-agent:
Fixes #360.

Extends the bundled `ci-fix` skill's existing-fix dedup (Step 1) to also scan recently closed bot-authored `fix/ci-*` PRs, so a maintainer's rejection rationale ("won't fix", "fix upstream") isn't ignored on the next failure of the same root cause.

The new query uses the `head:fix/ci-` search qualifier (not `--head`, which is ignored by the search API path) and requests `body` in `--json` so the prior diagnostic is available for shape-matching. Both gotchas are codified in the skill body.

## Test plan

- [ ] Generator regtests still pass (`uv run pytest`).
- [ ] Skill renders cleanly in a future `ci-fix` invocation that hits a recently rejected workaround.

### DIFF
--- a/plugins/tend-ci-runner/skills/ci-fix/SKILL.md
+++ b/plugins/tend-ci-runner/skills/ci-fix/SKILL.md
@@ -40,7 +40,7 @@ If an existing open PR addresses the same failure, comment on it linking the new
 
 Two gotchas:
 
-- For closed-state, use the `head:fix/ci-` **search qualifier**, not `--head` — `--head` filters client-side and is ignored by the search API path. `in:head` is silently dropped and falls back to default-field text matching.
+- Use the `head:fix/ci-` **search qualifier** (or `--head fix/ci-` — gh translates it to the same query). Don't use `in:head` — that's silently dropped and falls back to default-field text matching.
 - Request `body` in `--json` for both queries — the failure diagnostic written by the prior ci-fix run lives there, and shape-matching needs it.
 
 ### 2. Diagnose and fix
--- a/plugins/tend-ci-runner/skills/ci-fix/SKILL.md
+++ b/plugins/tend-ci-runner/skills/ci-fix/SKILL.md
@@ -20,11 +20,28 @@ Load `/tend-ci-runner:running-in-ci` first — it contains CI security rules, po
 
 ### 1. Check for existing fixes
 
+Check both open and recently closed bot-authored `fix/ci-*` PRs. A maintainer may have closed a prior workaround with a rejection rationale (e.g. "we'll fix upstream"); re-deriving the same fix forces them to close it twice.
+
 ```bash
-gh pr list --state open --head "fix/ci-" --json number,title,headRefName
+BOT_LOGIN=$(gh api user --jq '.login')
+
+# Open dedup:
+gh pr list --state open --head "fix/ci-" --json number,title,body,headRefName
+
+# Closed dedup — last ~14 days, bot-authored:
+gh pr list --state closed --author "$BOT_LOGIN" --search "head:fix/ci-" \
+  --json number,title,closedAt,body,headRefName \
+  --jq '[.[] | select((now - (.closedAt | fromdateiso8601)) < 1209600)] | .[]'
 ```
 
-If an existing PR addresses the same failure, comment on it linking the new run and stop.
+Match by **failure shape** (the diagnostic snippet in the PR body) rather than branch name — branch names encode run IDs and never repeat. If a closed PR with a maintainer rejection covers the same failure, exit silently; check the closure comment / review for the rationale before referencing it.
+
+If an existing open PR addresses the same failure, comment on it linking the new run and stop.
+
+Two gotchas:
+
+- For closed-state, use the `head:fix/ci-` **search qualifier**, not `--head` — `--head` filters client-side and is ignored by the search API path. `in:head` is silently dropped and falls back to default-field text matching.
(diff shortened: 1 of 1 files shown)


tend-agent (commented):
One inline suggestion on the first gotcha — the technical claim about `--head` is empirically incorrect (verified with `GH_DEBUG=api`). The recipe itself works; only the explanation needs correction.

plugins/tend-ci-runner/skills/ci-fix/SKILL.md
```diff
@@ -20,11 +20,28 @@ Load `/tend-ci-runner:running-in-ci` first — it contains CI security rules, po
...
 ```
 
-If an existing PR addresses the same failure, comment on it linking the new run and stop.
+Match by **failure shape** (the diagnostic snippet in the PR body) rather than branch name — branch names encode run IDs and never repeat. If a closed PR with a maintainer rejection covers the same failure, exit silently; check the closure comment / review for the rationale before referencing it.
+
+If an existing open PR addresses the same failure, comment on it linking the new run and stop.
+
+Two gotchas:
+
+- For closed-state, use the `head:fix/ci-` **search qualifier**, not `--head` — `--head` filters client-side and is ignored by the search API path. `in:head` is silently dropped and falls back to default-field text matching.
```
tend-agent: The claim that `--head` filters client-side / is ignored by the search-API path is empirically incorrect. Running `GH_DEBUG=api gh pr list --state closed --author tend-agent --head fix/ci-` shows gh translates the flag to `head:fix/ci-` in the GraphQL search query (`q: "author:tend-agent head:fix/ci- repo:max-sixty/tend state:closed type:pr"`) and returns the same row count as `--search "head:fix/ci-"`. So `--head` works fine for closed state.

The genuinely useful gotcha (and the one #360 raised) is `in:head` vs `head:` — keep that, drop the inaccurate `--head` framing:

```suggestion
- Use the `head:fix/ci-` **search qualifier** (or `--head fix/ci-` — gh translates it to the same query). Don't use `in:head` — that's silently dropped and falls back to default-field text matching.
```

tend-agent: Verified and applied in 60d8417. `GH_DEBUG=api` confirms `--head fix/ci-` becomes `q: "... head:fix/ci- ..."` in the GraphQL search query — same row count as `--search "head:fix/ci-"`. Reworded the gotcha to keep the real `in:head` trap and drop the inaccurate `--head` framing.

PR #391, opened 2026-05-06
fix(mention): run setup before PR checkout so stale PRs don't drop mentions

tend-agent:
## Problem

`tend-mention` triggered from an `issue_comment` (or `pull_request_review*` event) on a PR whose head branch predates a local composite action referenced by `setup` fails before the agent boots. The current step order is:

1. `actions/checkout@v6` (default branch)
2. `Check out PR branch` → `gh pr checkout <N>` switches HEAD to the PR branch
3. `uses: ./.github/actions/<setup>` → resolves against the *PR's* tree

If the PR is old enough that its tree doesn't yet contain that action (a long-lived external PR opened before the action was added, or a stale PR never rebased), step 3 fails with `Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/<repo>/<repo>/.github/actions/<name>'. Did you forget to run actions/checkout before running your local action?`. The maintainer's `@bot ...` request drops silently with a red CI X; the agent never runs.

## Repro

`max-sixty/worktrunk` run [25452024590](https://github.com/max-sixty/worktrunk/actions/runs/25452024590) on PR [#1256](https://github.com/max-sixty/worktrunk/pull/1256) (`feat/add-az-support`, opened 2026-03-04 by an external contributor). Comment: `@worktrunk-bot yes, let's try a fresh implementation, push to this PR please`. The local `./.github/actions/claude-setup` was added 2026-03-05 in worktrunk PR [#1273](https://github.com/max-sixty/worktrunk/pull/1273), one day after PR #1256 was opened. PR #1256 has not been rebased, so its head ref's tree lacks that action. `gh pr checkout 1256` then `uses: ./.github/actions/claude-setup` → 404.

## Fix

Run `setup` against the default branch (right after the initial `actions/checkout@v6`) and *then* `gh pr checkout` to switch to the PR branch. Local actions referenced by `setup` always resolve from default-branch state, regardless of how stale the PR is. The agent still sees the PR-branch tree because the PR-branch checkout follows setup; tools installed by setup (uv, bun, cargo crates, apt packages) persist in the runner env across the `git switch` and remain available to the agent and any post-checkout steps.

A short comment in the generated YAML calls out the constraint so future edits don't reverse the order.

## Tradeoffs

The previous test (`test_setup_after_pr_checkout_in_mention`) asserted the inverted contract — setup *after* PR checkout — to give setup steps PR-branch state (e.g. cache prefixes hashing PR's `Cargo.lock`). The cost of that contract is that any local-action-referencing setup breaks on long-lived stale PRs, and breaks them silently. Reversing the contract trades a minor cache-key miss (cargo/rust-cache handles this internally) for a workflow that always runs.

The renamed test (`test_setup_before_pr_checkout_in_mention`) carries the rationale in its docstring so the next reader understands why this ordering matters.

## Scope

`tend-mention` only — `tend-review` (and the other workflows) check out the PR head directly via a single `actions/checkout` and don't have the two-step main-then-PR sequence. `tend-review` on a conflicting + stale PR could in principle hit a related shape, but `tend-review` is gated by `pull_request_target` events ([opened, synchronize, ready_for_review, reopened]) — a stale PR that hasn't been synchronized since the action was introduced won't fire the workflow. Out of scope here.

## Tests

`generator/`: `uv run pytest` — 188 passed. Regtest snapshots regenerated for the two `mention` outputs.

## Evidence

[Evidence gist](https://gist.github.com/9675d1510da32c68a68c1d45137b21e0) — first occurrence of this specific shape recorded in this run; classified Critical (1 occurrence sufficient — clearly wrong outcome dropping a maintainer's explicit request) + Structural (deterministic action-resolution against checked-out tree).

### DIFF
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -426,7 +426,13 @@ jobs:
           fetch-depth: 0
           fetch-tags: true
           token: {bt}
-
+{setup}
+      # Setup runs against the default branch above; switching to the PR
+      # branch *after* setup means a stale PR (whose tree predates a local
+      # composite action like ./.github/actions/<name> referenced by setup)
+      # still resolves the action from the default branch. Reversing the
+      # order would 404 with "Can't find 'action.yml'" on every long-lived
+      # external PR opened before the action was added.
       - name: Check out PR branch
         if: |
           (github.event_name == 'issue_comment' && github.event.issue.pull_request.url != '') ||
@@ -442,7 +448,7 @@ jobs:
         env:
           GITHUB_TOKEN: {bt}
           PR_NUMBER: ${{{{ github.event_name == 'issue_comment' && github.event.issue.number || github.event.pull_request.number }}}}
-{setup}
+
       - name: Compute queue delay
         id: delay
         run: |
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -427,12 +427,6 @@ jobs:
           fetch-tags: true
           token: {bt}
 {setup}
-      # Setup runs against the default branch above; switching to the PR
-      # branch *after* setup means a stale PR (whose tree predates a local
-      # composite action like ./.github/actions/<name> referenced by setup)
-      # still resolves the action from the default branch. Reversing the
-      # order would 404 with "Can't find 'action.yml'" on every long-lived
-      # external PR opened before the action was added.
       - name: Check out PR branch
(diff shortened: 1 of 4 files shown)


max-sixty:
I don't think we need the comment here

tend-agent:
Removed in aefdb57 — regtest snapshots updated alongside.

Issue #302, opened 2026-04-18
token-usage.json under-reports turns and output_tokens when session uses run_in_background Bash


tend-agent:
`action.yaml` extracts `num_turns` and `usage.*` from the **last** `type:"result"` entry in `execution_file`. When the session uses `Bash` with `run_in_background: true`, that last-result snapshot seems to capture only the tail of the session — runs end up reporting `turns=2-3` and a few hundred `output_tokens` even though the JSONL records dozens of assistant turns. `total_cost_usd` stays correct because it is cumulative.

## Evidence from 2026-04-18 runs (24h window)

6 runs with `run_in_background: true` under-reported; 5 runs without the flag reported accurately (cost_usd always accurate).

| Run | Workflow | Assistant msgs | Tool calls | `run_in_background` | Reported turns | Reported output_tokens | Cost |
|---|---|---:|---:|:---:|---:|---:|---:|
| [24599273900](https://github.com/max-sixty/tend/actions/runs/24599273900) | tend-mention | 72 | 47 | yes | 3 | 623 | $2.59 |
| [24596537079](https://github.com/max-sixty/tend/actions/runs/24596537079) | tend-mention | 60 | 39 | yes | 3 | 615 | $1.85 |
| [24599273094](https://github.com/max-sixty/tend/actions/runs/24599273094) | tend-review | 38 | 22 | yes | 2 | 313 | $1.04 |
| [24596628700](https://github.com/max-sixty/tend/actions/runs/24596628700) | tend-review | 36 | 23 | yes | 2 | 216 | $1.09 |
| [24597836967](https://github.com/max-sixty/tend/actions/runs/24597836967) | tend-mention | 30 | 16 | yes | 2 | 232 | $0.67 |
| [24596678225](https://github.com/max-sixty/tend/actions/runs/24596678225) | tend-mention | 26 | 15 | yes | 2 | 254 | $0.56 |
| [24599068464](https://github.com/max-sixty/tend/actions/runs/24599068464) | tend-nightly | 43 | 30 | no | 33 | 9135 | $1.36 |
| [24599359889](https://github.com/max-sixty/tend/actions/runs/24599359889) | tend-review | 43 | 24 | no | 27 | 10890 | $1.43 |
| [24599081041](https://github.com/max-sixty/tend/actions/runs/24599081041) | tend-triage | 23 | 14 | no | 17 | 2966 | $0.62 |
[...]

PR #408, opened 2026-05-10
diag(action): upload execution-output.json so #302 can be diagnosed

tend-agent:
## Summary

- Copy `$EXECUTION_FILE` into `$LOGS_DIR/claude-execution-output.json` from the Token usage step so it ships with the existing `claude-session-logs` artifact.
- Targeted at #302: the action reads `type:"result"` entries from `$RUNNER_TEMP/claude-execution-output.json`, but that file is wiped at job end and the session JSONL we currently retain contains zero `result` entries — so we have no way to inspect what the SDK actually emits in an under-reporting run.

## Why this and not the fix itself

The proposed aggregation fix (sum `usage.*` across all `result` entries, keep `total_cost_usd` from the last) is plausible but unverified. Without a real execution-output.json from a `run_in_background: true` session it's not knowable whether `num_turns` is per-event or cumulative-in-each-entry, so summing could under-correct or double-count. One under-reporting sample is enough to pin down the semantics and ship a regression-tested fix.

## Test plan

- [ ] Merge, then wait for any tend-mention or tend-review run that polls CI in the background to land an under-reporting `token-usage.json`.
- [ ] Download the run's `claude-session-logs*` artifact, confirm `claude-execution-output.json` is present.
- [ ] Run `jq '[.[] | select(.type == "result")] | length' claude-execution-output.json` — if more than 1, also dump each entry's `num_turns` and `usage.*` to determine the aggregation rule.
- [ ] Open the follow-up PR with the corrected jq filter and a fixture from the captured file.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -288,6 +288,16 @@ runs:
         echo "$USAGE" > "$LOGS_DIR/token-usage.json"
         echo "usage=$(jq -c . <<< "$USAGE")" >> "$GITHUB_OUTPUT"
 
+        # Preserve the raw SDK message stream alongside the session JSONL so
+        # the under-reporting in #302 can be diagnosed against an actual
+        # execution-output.json. Without this, $RUNNER_TEMP is wiped at job
+        # end and the only retained artifact is the session JSONL — which
+        # has no `type:"result"` entries (those live only here). Drop this
+        # copy once #302 is resolved.
+        if [ -n "$EXECUTION_FILE" ] && [ -f "$EXECUTION_FILE" ]; then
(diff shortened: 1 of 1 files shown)


max-sixty:
let's isolate this to this repo only, so it doesn't affect consumers

tend-agent:
Gated on `$GITHUB_REPOSITORY = max-sixty/tend` in 25ddff0 — the `cp` becomes a no-op for every other consumer, so the diagnostic only ships logs from this repo. Gate vanishes when the block is dropped after #302 is resolved.

PR #410, opened 2026-05-11
skills(running-tend): add weekly consumers.json refresh

max-sixty:
Adds a weekly task to tend's own `running-tend` overlay that discovers public repos using tend (via `gh search code 'max-sixty/tend@v1'`), resolves each repo's `bot_name` from its `.config/tend.toml`, and writes the list to `data/consumers.json`. Opens a PR if the file changed.

Lives in the overlay (not the bundled weekly skill) so only tend's own weekly run does the discovery — consumer repos don't pay for it. The bundled `weekly` skill already scans the overlay for repo-specific weekly tasks (`weekly/SKILL.md:43-47`), so no workflow or generator change is needed.

The output unblocks the WEBSITE.md "Live data" phase: the website build can read `data/consumers.json` to fan out per-bot `search/issues` queries for the stat strip and activity feed, instead of hardcoding `commenter:tend-agent` (which only covers tend itself and worktrunk).

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -74,5 +74,7 @@ mkdir -p data
 ```
 
 Open a PR titled `chore: refresh consumers.json` if the file changed. Skip
-the PR (no diff to land) if `git diff --quiet data/consumers.json`. Code
-search is 10 req/min — one call covers the whole list.
+the PR (no diff to land) when `git status --porcelain data/consumers.json`
+is empty — `git diff --quiet` returns 0 for untracked paths, so the
+first-run case would no-op. Code search is 10 req/min — one call covers
+the whole list.
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -44,3 +44,35 @@ Artifact paths: `-home-runner-work-tend-tend/<session-id>.jsonl`
 
 `review-reviewers` runs produce 3 session logs per run (one per matrix repo:
 `max-sixty/worktrunk`, `max-sixty/tend`, `PRQL/prql`).
+
+## Weekly: refresh `data/consumers.json`
+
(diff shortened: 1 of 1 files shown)


.claude/skills/running-tend/SKILL.md
```diff
@@ -44,3 +44,35 @@ Artifact paths: `-home-runner-work-tend-tend/<session-id>.jsonl`
...
+      | sed -n 's/^bot_name *= *"\([^"]*\)".*/\1/p' | head -1)
+    [ -n "$bot" ] || continue
+    jq -nc --arg repo "$repo" --arg bot "$bot" '{repo: $repo, bot_name: $bot}'
+  done
+} | jq -s . > data/consumers.json
+```
+
+Open a PR titled `chore: refresh consumers.json` if the file changed. Skip
+the PR (no diff to land) if `git diff --quiet data/consumers.json`. Code
+search is 10 req/min — one call covers the whole list.
```
tend-agent: `git diff --quiet <untracked-path>` exits 0 because the path isn't in the index — so the first run (when `data/consumers.json` doesn't exist yet) would silently skip the PR and never land the initial file. `git status --porcelain` covers both the modified and untracked cases:

```suggestion
Open a PR titled `chore: refresh consumers.json` if the file changed. Skip
the PR (no diff to land) when `git status --porcelain data/consumers.json`
is empty — `git diff --quiet` returns 0 for untracked paths, so the
first-run case would no-op. Code search is 10 req/min — one call covers
the whole list.
```

tend-agent (commented):
One issue worth fixing before this lands.

.claude/skills/running-tend/SKILL.md
max-sixty: Good catch — verified in a fresh repo (`git diff --quiet untracked.txt` returns 0). Switched to `git status --porcelain` in 16e7895.

tend-agent (approved):

PR #413, opened 2026-05-11
site: deploy to GitHub Pages

max-sixty:
Mirrors worktrunk's `publish-docs.yaml` design: push-to-main + `workflow_dispatch` triggers, concurrency group, `build`/`deploy` split with `deploy-pages@v5`, fork-owner guard. Pages is enabled in repo settings with GitHub Actions as the source; first deploy lands at `https://max-sixty.github.io/tend/`.

`SITE_BASE=/tend/` is set on the build step so Astro emits asset paths under the sub-path. `astro.config.mjs` reads `SITE_BASE` with a `/` fallback, so local `npm run dev` / `npm run build` keep working at root. Wiring up the `tend.dev` custom domain later means dropping the env var from the workflow — no config revert.

Known limitation: `site/src/layouts/Base.astro` has two hardcoded absolute paths (`href="/favicon.png"` and the wordmark `href="/"`) that won't resolve under `/tend/`. They'll need either the custom domain or an `import.meta.env.BASE_URL` swap; left alone in this PR.

### DIFF
--- a/.github/workflows/publish-site.yaml
+++ b/.github/workflows/publish-site.yaml
@@ -1,0 +1,67 @@
+name: publish-site
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'site/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-site:
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v6
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'npm'
+          cache-dependency-path: site/package-lock.json
+
+      - name: Install dependencies
+        run: npm install
+        working-directory: site
+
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-site.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'npm'
+          cache-dependency-path: site/package-lock.json
+
+      - name: Install dependencies
+        run: npm install
```
tend-agent:
```suggestion
        run: npm ci
```

`npm ci` is faster, deterministic, and errors if `package-lock.json` is out of sync — preferred for CI where the lockfile should be the source of truth (and the setup-node cache key already assumes it).

tend-agent (commented):
One suggestion: use `npm ci` instead of `npm install` for deterministic builds — the cache key is already lockfile-driven via `cache-dependency-path`, and `npm install` can rewrite the lockfile.

.github/workflows/publish-site.yaml
max-sixty: Good call — applied in 529f292.

tend-agent (approved):

PR #415, opened 2026-05-11
Live data fetcher + currently-tending Worker on currently.tend-src.com

max-sixty:
Phase 3 §8 (data fetcher) + §11 (currently-tending) from `WEBSITE.md`. Three live data streams now in place:

- **Stats** (`data/stats.json`) — daily counts across all tend bots, written nightly by the bot via `running-tend`'s skill (no dedicated workflow).
- **Activity** (`data/activity.json`) — last 10 issues/PRs touched by any tend bot, grouped by kind (ci-fix / review / triage), same nightly path.
- **Currently tending** (`https://currently.tend-src.com`) — Cloudflare Worker fanning out across consumer repos for in-progress `tend-*` runs, edge-cached for 30 s.

## What's here

`scripts/fetch_website_data.py` — stdlib-only Python. Reads `data/consumers.json`, iterates `bot_name`s, runs Search-API queries (with `2.1 s` self-throttle and exponential-backoff retries on transient failures), writes the two JSON files. `write_if_changed` skips commits when only `generated_at` would move.

`.claude/skills/running-tend/SKILL.md` gained a "Nightly: refresh data/{activity,stats}.json" section. `set -euo pipefail` + `git pull --rebase` push retry + PR fallback if branch protection blocks.

`worker/` — TypeScript Cloudflare Worker. KV for the consumers.json cache (1 h, above KV's 60 s minimum); `caches.default` HTTP edge cache for the rendered response (30 s, with 5 s on fallback). Edge cache also coalesces concurrent misses. Custom domain: `currently.tend-src.com`. CORS `*`. Try/catch wraps every refresh — failures degrade to `{currently_tending: []}` so the indicator never breaks the page.

Validation: `data/consumers.json` shape is checked before KV-caching (rejects path traversal, multi-slash, leading `./-`, `..` anywhere); per-repo runs distinguish 401/403 (auth — propagates) from 404/5xx (skips).

## Worth a reviewer's eye

- **`worker/src/index.ts`** — the cache strategy (KV vs `caches.default`) and the validation helpers. `isValidRepo` rejects all 12 traversal cases the tests cover.
- **`scripts/fetch_website_data.py`** — the retry+throttle math: 4 attempts × (1+2+4)s backoff, 2.1 s/Search-call. Total budget at N=5 fits comfortably in the bot's nightly window.
- **`.claude/skills/running-tend/SKILL.md`** — bot acts on this autonomously; `set -e` makes a persistent fetcher failure surface in the run summary instead of silently stale JSON.

## Out of scope

- `publish-site.yaml` path filter — it triggers on `site/**` only, so data commits won't auto-rebuild the Astro site. The rendering session (§9/§10) will pick up the new JSON on the next code change to `site/`, or you can expand the filter to `data/**` if you want immediate redeploys. Flagging not fixing.
- The Astro site config still says `site: "https://tend.dev"` — separate concern.

## Tests

- Python: 11 cases (multi-bot dedupe/sum, retry-then-success, 404-no-retry, URLError-retry, structural diff, shape validation).
- TypeScript: 10 cases (12 path-traversal rejections + positive cases, 401/403-throw vs 404-skip, dedupe/sort, KV shape rejection).
- `tsc --noEmit` clean, `ruff` clean, `actionlint` clean.
- Worker live-verified: `cf-cache-status: HIT` on warm requests; real currently_tending payload on cold.

## Setup notes

One-time Cloudflare setup (already done): account is `m@maximilianroos.com` on CF, workers.dev subdomain `maxsixty`, KV namespace `CACHE` provisioned, `GITHUB_TOKEN` secret set, `tend-src.com` zone present. Future redeploys go through `.github/workflows/worker-deploy.yaml` once `CLOUDFLARE_API_TOKEN` is added as a repo secret (scoped token, "Edit Cloudflare Workers" template).


worker/src/index.ts
```diff
@@ -0,0 +1,272 @@
...
+  currently_tending: CurrentlyTendingEntry[];
+}
+
+const REPOS_KEY = "repos:v1";
+const CACHE_TTL_SECONDS = 30;
+const FALLBACK_TTL_SECONDS = 5; // shorter so a transient outage clears fast
+const REPOS_TTL_SECONDS = 3600;
+const FETCH_TIMEOUT_MS = 10_000;
+const WORKFLOW_PREFIX = "tend-";
+const PER_PAGE = 5;
```
tend-agent: `actions/runs?status=in_progress` returns the most-recently-created runs first across **all** workflows, and `tend-*` filtering happens client-side. So if a consumer repo has ≥5 non-tend in-progress runs newer than its tend-* runs, the tend runs get paged off and `currently_tending` silently misses them. Usually safe (tend workflows tend to start later than the CI workflows that triggered them), but possible during a flurry of pushes. Worth considering bumping to `PER_PAGE = 30` (GitHub's default) — payload cost is trivial and it kills the ordering risk.

tend-agent (approved):
One observation on the Worker — see inline.

PR #420, opened 2026-05-11
site: animate the tend logo on entrance

max-sixty:
## Summary

- Replace the placeholder seedling animation with the tend mark proper, animated: a fine pen-stroke traces the whole outer line from the foot upward, the warm fill floods in *following* the pen, then the pen fades and the mark settles with a faint breath (~3s entrance).
- The mark is the real tend logo, vectorised into a single silhouette — used in the hero (animated, `<Logo size={260} />`) and the header lockup (`<Logo size={28} static />`).
- `Seedling.astro` is gone; `Logo.astro` replaces it. `prefers-reduced-motion` (and `static`) render the resting mark with no animation.

## Test plan

- [ ] `cd site && npm run build` builds clean
- [ ] Homepage: hero plays the draw-on, header shows the static mark
- [ ] `prefers-reduced-motion`: mark shows with no animation


site/src/components/Logo.astro
```diff
@@ -0,0 +1,106 @@
...
+---
+
+<svg
+  class:list={["logo", { "is-static": isStatic }]}
```
tend-agent, line 23:
```suggestion
  class:list={["logo", "logo-mark", { "is-static": isStatic }]}
```
Without `logo-mark` on the SVG, `.logo-frame .logo-mark { width: 100%; height: auto; }` in `global.css` never applies, and the hero mark renders at its fixed `width=202` attribute — overflowing the 140px frame on viewports `<= 760px`.

tend-agent (commented):
The two new selectors in `site/src/styles/global.css` (`.site-header .wordmark .logo-mark` and `.logo-frame .logo-mark`) don't match anything — the SVG is rendered with `class="logo"` (and optionally `is-static`), never `logo-mark`. The header one is harmless because `Logo.astro`'s scoped `.logo` already paints with `--oak`, but the hero one carries the responsive sizing: at `max-width: 760px`, `.hero .logo-frame { max-width: 140px }` constrains the frame while the SVG renders at its fixed `width=202` attribute (`size=260`), so the mark overflows on narrow viewports. Adding `logo-mark` to the SVG's class list is the smallest fix.

tend-agent (approved):

PR #419, opened 2026-05-11
data: nightly refresh

tend-agent:
Nightly refresh of `data/activity.json` and `data/stats.json` (consumed by the marketing site stat strip + activity feed). Generated by `scripts/fetch_website_data.py`.

Branch protection blocks direct push to `main`, per the fallback path in the `running-tend` skill; opening as a PR.

PR #425, opened 2026-05-11
skills(running-tend): drop nightly website-data refresh

tend-agent:
Drop the nightly `fetch_website_data.py` invocation from `running-tend`. The activity/stats JSON it produced isn't consumed by the site or the Worker today, so the nightly run was generating dead data and PR churn (e.g. [#419](https://github.com/max-sixty/tend/pull/419)).

Per [@max-sixty's directive](https://github.com/max-sixty/tend/pull/419#issuecomment-4419631593): only refresh the repos (the weekly `consumers.json` task stays).

Leaves `scripts/fetch_website_data.py`, `docs/website-data.md`, and the existing `data/{activity,stats}.json` files in place — re-wiring is a separate decision if/when the site renders the data.

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -45,38 +45,6 @@ Artifact paths: `-home-runner-work-tend-tend/<session-id>.jsonl`
 `review-reviewers` runs produce 3 session logs per run (one per matrix repo:
 `max-sixty/worktrunk`, `max-sixty/tend`, `PRQL/prql`).
 
-## Nightly: refresh `data/{activity,stats}.json`
-
-Aggregates the daily activity feed and stat counts the website renders.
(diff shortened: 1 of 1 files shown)


max-sixty:
do we keep the nightly run re updating repos?

tend-agent:
Just the directive was `#1, so only refresh the repos` — but the repos list (`consumers.json`) is actually on the **weekly** schedule (`tend-weekly.yaml`), not nightly. The nightly overlay had only the website-data block this PR removes; after it merges, the nightly run keeps doing the generic nightly tasks (conflict resolution on bot PRs, recent-commit review, etc.) but has no repo-list step.

Want me to move the `consumers.json` refresh to nightly so "only refresh the repos" maps to the nightly cadence? Or keep it weekly (status quo)?

PR #428, opened 2026-05-11
worker: serve /activity and /stats; remove pre-worker Python pipeline

max-sixty:
The Cloudflare Worker now serves all three website data streams, with per-route edge TTLs matched to each stream's freshness budget:

| Route | TTL | Fallback TTL |
| --- | --- | --- |
| `/currently-tending` (also `/`) | 30 s | 5 s |
| `/activity` | 5 min | 30 s |
| `/stats` | 1 h | 60 s |

The three routes share a single `serveCached()` helper. `/activity` issues three Search queries per bot (ci-fix / review / triage), dedupes by URL with first-kind-wins, sorts newest first, caps at 10. `/stats` issues five queries per bot, sums `total_count`. Both reuse the existing consumers-list fetch and KV cache.

The pre-worker Python fetcher (`scripts/fetch_website_data.py`) and its `test-scripts` CI job are removed — they produced static JSON that the site no longer consumes, and #425 already dropped the nightly that ran them. `data/consumers.json` continues to refresh weekly via running-tend.

Frontend pieces (Activity / Stats Astro components) are not in this PR — they consume `/activity` and `/stats` and will land separately.


tend-agent (approved):
Two findings, both non-blocking — the Worker change itself looks good and tests cover the dedup/sort/sum semantics.

**Stale top-level docs.** `WEBSITE.md` and `WEBSITE-live-data.md` still describe the pre-Worker architecture this PR deletes (nightly Action, `scripts/fetch_website_data.py`, `data/{activity,stats}.json`). The cross-reference from `docs/website-data.md:18` ("The rate-limit reasoning is in `../WEBSITE-live-data.md`") now points at a doc whose summary table contradicts the new design. Concretely:

- `WEBSITE.md` §"Live data" (lines 95–104), §"GitHub data fetcher" (163–167), §9 stat strip (169), §10 activity feed (174), §11 currently-tending fallback (183).
- `WEBSITE-live-data.md` §"Stats" (89–94), §"Recent activity feed" (96–104), §"Currently tending" fallback (123), summary table (130–132).

Happy to push a follow-up commit that updates these to match the new Worker-served model — say the word.

**`searchRaw` treats `403` as auth failure, but GitHub returns `403` for the primary Search API rate limit.** When a refresh exceeds 30 req/min on the token, the affected query returns `403` with `X-RateLimit-Remaining: 0`. `worker/src/index.ts:406-407` then throws, which rejects the whole `Promise.all` and degrades the route to the empty fallback. With N=5 consumers a single `/stats` cache miss fires 25 concurrent Search queries (5 stat queries × 5 bots) in <1s — close enough to the primary cap that bursts can plausibly trip it, especially if `/activity` (15 concurrent) refreshes near the same time. The Python fetcher self-throttled at 2.1s/call to stay under this, but the Worker fires everything in parallel.

Two options: distinguish rate-limit `403` (has `X-RateLimit-Remaining: 0`) from real auth `403` and degrade the rate-limited query the same way as `429`/`422`; or sequence the fan-out enough to stay under 30/min. With current N=5 the system absorbs this gracefully (empty response, short fallback TTL → retry), so I wouldn't block on it, but worth handling before N grows.

PR #430, opened 2026-05-12
worker: rename to tend-website / api.tend-src.com

max-sixty:
The Worker now serves three streams (currently-tending, activity, stats), so the single-purpose `tend-currently` / `currently.tend-src.com` names have gotten misleading. This renames:

- Wrangler app: `tend-currently` → `tend-website`
- Custom domain: `currently.tend-src.com` → `api.tend-src.com`

And sweeps the references in `worker/README.md`, `docs/website-data.md`, `WEBSITE.md`, `WEBSITE-live-data.md`, and the deploy workflow's docstring.

## Deploy ops (post-merge)

1. The deploy workflow will create the new `tend-website` Worker on merge. With `custom_domain = true` and `tend-src.com` already in this Cloudflare account, Wrangler should auto-configure DNS for `api.tend-src.com`.
2. **Set the secret on the new Worker before relying on it:**
   ```sh
   cd worker && npx wrangler secret put GITHUB_TOKEN --name tend-website
   ```
   Without the secret, any request to `api.tend-src.com` will fail. Frontend isn't on main yet, so this isn't user-visible until the components land.
3. The old `tend-currently` Worker stays running on `currently.tend-src.com` until manually deleted via the Cloudflare dashboard. Safe to delete once `api.tend-src.com` is verified.
4. The KV namespace ID `cba4d4d307974f64a6c4a4a1390c5153` is account-scoped and reused as-is.


tend-agent (commented):
One leftover from the rename sweep: `worker/package.json` and `worker/package-lock.json` still have `"name": "tend-currently-worker"`. It's an internal npm name (Cloudflare uses the `name` in `wrangler.toml`), so nothing functional depends on it — but since this PR is about sweeping references, worth aligning. Happy to push a follow-up commit changing both to `tend-website-worker` if you'd like.

Deploy-ops notes in the description look right — Wrangler will create the new app on merge, and the `GITHUB_TOKEN` secret has to be set per-Worker so the new app needs its own `wrangler secret put`.

PR #431, opened 2026-05-12
site: render activity feed and stat strip from the live-data Worker

max-sixty:
Adds the frontend that reads `/activity` and `/stats` from the live-data Worker.

- `Stats.astro` — strip of counters below the hero (reviews / ci fixes / triage notes), with optional `+N this week` captions. Hidden when all values are zero or the fetch fails.
- `Activity.astro` — list of the bots' last actions between "what tend tends" and quick start. Each row: kind chip · repo · linked title · relative time. Hidden when empty or the fetch fails. Responsive collapse to stacked layout under 720 px.
- `lib/api.ts` — Worker URL configurable via `PUBLIC_WORKER_URL`; default `https://api.tend-src.com`. See `site/.env.example`.

The currently-tending dot is a separate component for another PR.

## Behavior when the Worker isn't ready

Both components silently hide on fetch failure or empty payload, so a misconfigured / missing Worker degrades to no section rather than a broken page.


site/src/pages/index.astro
```diff
@@ -99,6 +103,10 @@ claude /install-tend</code></pre>
 
   <hr />
 
+  <Activity />
+
+  <hr />
```
tend-agent, line 108:
```suggestion
  <Activity />

  <hr />
```

Dropping the leading `<hr />` mirrors how `<Stats />` was inserted (component above the existing separator) and avoids the double-hr when Activity is hidden.

tend-agent (commented):
When `/activity` errors or returns an empty list, the section collapses to `display: none`, but the framing `<hr />` on either side stay rendered — leaving a double horizontal rule between *what tend tends* and *quick start* in the exact graceful-degradation state this PR sets out to support. Matching the placement Stats uses (component directly above the existing separator, no new `<hr />`) avoids the double rule without any JS-side coordination.

PR #433, opened 2026-05-12
site: render the "currently tending" indicator from the live-data Worker

max-sixty:
## Summary

Adds the third live-data element to the marketing site: a "currently tending" indicator in the hero. It fetches `api.tend-src.com/currently-tending` and shows `● tending owner/repo · review` while a tend workflow is running. When nothing is in progress — the common case — it falls back to `/activity`'s newest event as `○ last tended N min ago`, per the UI-fallback contract already written into `docs/website-data.md`. If both requests fail it stays hidden, like the existing stats strip and activity feed.

Also extracts the relative-time formatter (`relativeTime`) — previously duplicated inside `Activity.astro` — into `site/src/lib/time.ts`, now shared by both components.

## Files

- `site/src/components/CurrentlyTending.astro` — new component (server-rendered hidden `<p>` + client `<script>`, same pattern as `Stats.astro` / `Activity.astro`)
- `site/src/lib/time.ts` — new; `relativeTime(iso)` shared helper
- `site/src/components/Activity.astro` — now imports `relativeTime` instead of its local copy
- `site/src/pages/index.astro` — renders `<CurrentlyTending />` after the hero badges
- `site/src/styles/global.css` — `.now-tending` / `.now-dot{-live,-idle}` + a `now-pulse` keyframe gated behind `prefers-reduced-motion: no-preference`
- `site/README.md`, `docs/website-data.md` — doc updates

## Test plan

- `npm run build` clean; `pre-commit run --all-files` green
- Verified in a browser against the production Worker: idle state renders "last tended N min ago"; activity feed still renders after the `relativeTime` refactor; no console errors
- The live state (filled moss dot + pulse) wasn't reproduced in-browser — it needs a tend workflow actually running — but the code path is symmetric with the verified idle path

### DIFF
--- a/site/src/styles/global.css
+++ b/site/src/styles/global.css
@@ -342,7 +342,9 @@ hr::after {
 
 /* ---------- "Currently tending" indicator ---------- */
 
-.now-tending {
+/* Scoped to :not([hidden]) so the layout rules don't defeat the UA
+   [hidden] { display: none } default before the fetch resolves. */
+.now-tending:not([hidden]) {
   display: flex;
   align-items: center;
   gap: 0.5rem;
--- a/site/src/styles/global.css
+++ b/site/src/styles/global.css
@@ -340,6 +340,39 @@ hr::after {
   height: auto;
 }
 
+/* ---------- "Currently tending" indicator ---------- */
+
+.now-tending {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  margin: 1.4rem 0 0;
+  font-family: var(--font-mono);
+  font-size: 0.78rem;
+  letter-spacing: 0.03em;
+  color: var(--ink-soft);
+}
+
+.now-dot {
+  flex: none;
+  width: 0.5rem;
+  height: 0.5rem;
+  border-radius: 50%;
+}
+.now-dot-live { background: var(--moss); }
+.now-dot-idle {
+  background: transparent;
+  border: 1px solid var(--oak-deep);
+}
+
+@media (prefers-reduced-motion: no-preference) {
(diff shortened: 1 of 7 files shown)


site/src/styles/global.css
```diff
@@ -340,6 +340,39 @@ hr::after {
   height: auto;
 }
 
+/* ---------- "Currently tending" indicator ---------- */
+
+.now-tending {
```
tend-agent: Author rules win over the UA stylesheet, so `display: flex` here defeats the `[hidden] { display: none }` default — the empty `<p hidden>` ends up with `height: 0` but the `margin: 1.4rem 0 0` still applies. That leaves a ~22px gap above the logo before either fetch resolves, and the gap persists if both endpoints fail (the file comment claims "Hidden only when both requests fail"). Verified in Chromium: `getComputedStyle(p).display` is `flex`, `marginTop` is `22.4px`. Scoping the rule to the non-hidden state fixes it:

```suggestion
.now-tending:not([hidden]) {
```

tend-agent (commented):
One inline suggestion — the `hidden` attribute is being defeated by the new author rule, so the indicator's top margin leaves a visible gap before (and during outage of) the fetch.

site/src/styles/global.css
max-sixty: Good catch — fixed in f1fde90 by scoping the rule to `.now-tending:not([hidden])`. Verified in-browser: while `[hidden]`, the element is now `display: none` / `margin-top: 0` / `height: 0` (no gap); once the fetch un-hides it, `display: flex` and the margin apply as before.

tend-agent (approved):

PR #445, opened 2026-05-13
worker: stale-while-revalidate so cache hits never wait on the GitHub fanout

max-sixty:
The `/activity` stat strip — `964 PRs / 1,512 comments / 186 issues` — took a couple of seconds to load on every visit that landed inside a fresh 5-minute window: the first request fanned out 12 concurrent GitHub Search calls and the viewer waited for them.

This is stale-while-revalidate. A request always returns from the colo cache immediately; when the cached entry is past its freshness budget, the hit *also* kicks off a `ctx.waitUntil(refreshAndCache(...))` so the next viewer gets fresher data. Only a cold cache — a fresh deploy, or a quiet stretch longer than ten freshness budgets — makes one request synchronously refresh.

### Mechanism

Each cached response carries `x-tend-stale-at` (epoch ms = put time + ttl) and `Cache-Control: public, max-age=<ttl>, s-maxage=<10·ttl>`. The colo cache honors `s-maxage`, keeping the entry serveable for 10 budgets (5 min on `/currently-tending`, 50 min on `/activity`). Browsers ignore `s-maxage` and use `max-age`. `serveCached` reads the stamp on each hit to decide whether to background-refresh; a missing or garbled stamp counts as stale, so the first hit after this deploys self-heals any entry written by the old code.

Fallback budget semantics are preserved: a refresh that throws still writes the empty payload with `opts.ttl.fallback`, so the next visitor retries quickly.

### Verified

`curl -sD- https://api.tend-src.com/activity` against the current production deploy returns no `cf-cache-status` and no `Age` header, confirming Cloudflare's zone CDN isn't caching this Worker route — so `s-maxage` only affects `caches.default` retention, exactly as the design assumes. A note next to the `Cache-Control` line flags the assumption so a future Cache Rule on this host would prompt a revisit.

Tests, typecheck, and `pre-commit` all green; added 2 unit cases for the pure `isStale` helper (boundary, missing, garbled stamps).

### Trade-offs left in

A burst of concurrent stale-hits can fan out several refreshes at once; the short fallback budget caps the damage if that ever trips a rate limit. `STALE_SERVE_FACTOR = 10` is one knob across both routes; per-route values would be straightforward but add a knob without a current pain point. A cron prewarm would close the cold-start gap entirely but trade away the "zero GitHub calls when idle" property — documented as a deliberate non-choice in `docs/website-data.md`.

### DIFF
--- a/docs/website-data.md
+++ b/docs/website-data.md
@@ -4,7 +4,7 @@ Both data streams the tend site renders are served by one Cloudflare Worker.
 The Worker reads a small input file (`data/consumers.json`) from this repo, fans
 out to GitHub, and serves CORS-enabled JSON to the site.
 
-| Stream | Endpoint | Edge TTL | Fallback TTL |
+| Stream | Endpoint | Freshness budget | Fallback budget |
 | --- | --- | --- | --- |
 | Currently tending | `/currently-tending` (also `/`) | 30 s | 5 s |
 | Activity | `/activity` | 5 min | 30 s |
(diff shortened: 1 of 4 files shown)


tend-agent (commented):
`worker/README.md` still describes the pre-SWR model — "How it works" calls the rendered response "edge-cached (`caches.default`) for 30 s" and credits the edge with deduping concurrent misses; the "Cache strategy" section says the colo TTL is `Cache-Control: max-age`; line 32 quotes the Cache-Control as `public, max-age=30`; lines 5–6 say "edge TTL". Worth bringing in line with the `docs/website-data.md` rewrite and the updated source comments. Happy to push a follow-up commit if useful.

Otherwise reads cleanly — the stamp-based SWR with `s-maxage` for colo retention is the simplest shape that works around CF's lack of native `stale-while-revalidate`, the cold-cache vs stale-hit split is clean, and the `STALE_AT_HEADER`-counts-as-stale fallback self-heals entries written by the old code.

tend-agent (approved):

Issue #439, opened 2026-05-13
feat(generator): expose TOML-expressible delete-key sentinel for `workflow_extra` / `_deep_merge`


jrdncstr:
Hey, long time with no issues, that's probably a good sign! Anyways, here's one.

## Summary

`workflow_extra` and the job-level overrides in `.config/tend.toml` merge via RFC 7396 (JSON Merge Patch) semantics in `generator/src/tend/workflows.py::_apply_extras` and `::_deep_merge`. The implementation supports deleting a generated key by setting it to `None` in the override mapping — the standard JSON Merge Patch convention. However, **TOML has no `None` literal**, so there is currently no way to express "delete this generated key" from `.config/tend.toml`.

The concrete impact: a user who wants to drop a generated `on.<event>.types` entry — for example, narrowing `issues.types = [edited]` down to `issues.types = []` so the workflow does not fire on edited-issue events at all — has no way to do so via `.config/tend.toml` overrides. Setting `[workflows.<name>.workflow_extra]` `on.issues.types = []` either gets merged in list-replace fashion (replacing with an empty list, which `on.issues` then evaluates as "no triggers" — may or may not behave consistently across the generator and GitHub Actions), or the override is silently ignored. The TOML user cannot express the `None` deletion sentinel that JSON or YAML overrides could.

## Proposal

Add a string sentinel (suggested literal: `"__TEND_DELETE__"`) that the generator translates to `None` before invoking `_deep_merge`, recursively. For example:

```toml
[workflows.mention.workflow_extra]
on.issues = "__TEND_DELETE__"
```

…would cause the generator to remove the `on.issues` key from the merged workflow entirely (equivalent to JSON Merge Patch `{"on": {"issues": null}}`).

A pre-merge pass over the parsed TOML tree would replace any value equal to the sentinel with `None`, then hand the result to `_deep_merge` unchanged. The existing `None`-handling branch in `_deep_merge` already produces the correct delete-key semantics.

## Rationale

This unlocks the same expressiveness for TOML users that JSON/YAML override users implicitly have via native `null`/`None`. Without it, the documented JSON Merge Patch semantics are partial for the TOML config surface — users see "deep-merge mappings, replace scalars and lists wholesale" work as expected, but "delete-key" silently doesn't work because the TOML representation can't carry it.

[...]

PR #454, opened 2026-05-13
feat(generator): switch config surface from TOML to YAML

max-sixty:
## Summary

Move `.config/tend.toml` → `.config/tend.yaml`. Alternative to the `"__TEND_DELETE__"` string sentinel landed in #451 — same goal (reach RFC 7396's null-deletes-key branch from user config), different mechanism. #439 has both for comparison.

YAML has a native `null`, so the delete-key example becomes:

```yaml
workflows:
  nightly:
    workflow_extra:
      on:
        schedule: null   # drop the cron, keep workflow_dispatch
```

The parser is `ruamel.yaml` (YAML 1.2), not PyYAML — so the usual YAML 1.1 footguns don't bite the config surface:

- Unquoted `on:` stays the string `"on"`, not boolean True.
- Norway problem doesn't apply: `yes`/`no`/`on`/`off` stay strings; only `true`/`false` are booleans.

## Scope

- `Config.load` reads `.config/tend.yaml` via ruamel.yaml's safe loader.
- `tend init` auto-migrates a legacy `.config/tend.toml` (verifies parsed structures match via dict equality before swapping). `tend check` stays read-only and points users at `tend init` if only `.toml` is present.
- Migrated `docs/tend.example.toml` → `docs/tend.example.yaml` with a new "delete a generated key with null" example.
- Removed the `__TEND_DELETE__` sentinel + its three tests; added replacement tests for YAML-native `null` (top-level, nested, missing-key noop), unquoted `on:` round-trip as string, and Norway-problem strings (`NO`, `off`) surviving uncoerced.
- `_apply_extras` uses `ruamel.yaml` with `typ="rt"` to preserve insertion-order on regen (safe mode sorts mapping keys, which would churn every diff). Flow-mode scalar dumper stays on `typ="safe"`.
- Migrated every test, tend's own `.config/tend.yaml`, install-tend, review-reviewers, review-runs, running-tend, CLAUDE.md, README.md, action.yaml, TODO.md. `bot_name` scrapers in cross-repo lookups now use `yq '.bot_name // ""'` instead of sed.

## Tradeoffs vs the sentinel

- (+) Native `null` instead of a magic string.
- (+) Single config format aligned with the generated workflows.
- (+) YAML 1.2 via ruamel.yaml — no Norway problem, no `on:` → True trap.
- (−) Breaking change for every consumer: `.config/tend.toml` no longer parses; consumers must rename and translate. Auto-migration in `tend init` softens this — nightly regen handles it for free — but `tend check` invocations between bump and nightly hard-error with the migration hint.
- (−) Extra runtime dep: `ruamel.yaml` (pure-Python, ~1.3 MB installed) replacing `pyyaml` (C-extension, ~750 KB installed).

The sentinel approach (#451) keeps TOML and avoids both the breaking change and the new dep at the cost of a magic string in config. Pick your poison.

## Test plan

- [x] `cd generator && uv run pytest` — 203 passed
- [x] `pre-commit run --all-files` — all hooks pass
- [x] `uv run python -m tend init` round-trips `.config/tend.yaml` into the existing `.github/workflows/` files (no diff outside pre-existing drift in tend-mention.yaml and the v0.0.17 → v0.0.18 header stamp from the release on main)
- [x] Delete-key example in `docs/tend.example.yaml` produces the expected merged workflow — covered by `test_null_drops_top_level_key`
- [x] Norway-problem strings survive — covered by `test_norway_problem_string_not_coerced`
- [x] Auto-migration verifies parsed structures match before swapping — covered by `test_init_auto_migrates_legacy_toml`, `test_init_does_not_touch_yaml_when_both_exist`, `test_workflows_from_migrated_config_match_workflows_built_from_toml_dict`


tend-agent (approved):

tend-agent (commented):
One stale `.config/tend.toml` reference outside the diff: [`generator/src/tend/checks.py:340`](https://github.com/max-sixty/tend/blob/19f1d67e983817828e557db0e0ce62cb0220fd05/generator/src/tend/checks.py#L340) is the `repo-secret-allowlist` failure message that points the user at `secrets.allowed in .config/tend.toml` — a file that no longer exists after this PR. Happy to push a one-line fix if you want.

tend-agent (approved):

tend-agent (approved):

tend-agent (approved):

max-sixty:
@nedtwigg any thoughts? this solves the null problem, albeit in an obtuse way....

nedtwigg:
I don't have any preference between TOML and YAML. I haven't adopted yet, hoping to do so sometime between today and Monday, I'll wait for this to settle first :)

max-sixty:
@tend-agent resolve conflicts

tend-agent (approved):

PR #463, opened 2026-05-13
site: live tending timer, merge in-flight runs into activity feed

max-sixty:
Two changes that make the "currently tending" pulse provably real:

- The hero line ticks elapsed time each second (`tending repo · workflow · 0:42`), driven by `started_at`. Live digits beat decorative text.
- The hero is now an `<a href="#recent-activity">`, and the activity feed merges in-flight runs (`kind: "tending"`) into the same time-ordered list — title links to the GitHub Actions run, time cell shows live elapsed, kind chip is a filled moss pulse. Click the pulse, land on the unified feed.

The shared `m:ss` / `h:mm:ss` formatter lives in `site/src/lib/time.ts` as `elapsedTime(iso)`. A single `setInterval` updates every tending row.

Caveats:

- One additional fetch (`/currently-tending`) from the activity section. The Worker caches both endpoints, so it's effectively free.
- Anchor scroll only works once `/activity` has resolved and the section is no longer `hidden`. Sub-second normally; not papered over.

### DIFF
--- a/site/src/components/CurrentlyTending.astro
+++ b/site/src/components/CurrentlyTending.astro
@@ -26,12 +26,13 @@
     comments?: Bucket;
   }
 
-  function paint(el: HTMLElement, state: "live" | "idle", text: string): void {
+  function paint(el: HTMLElement, state: "live" | "idle", text: string): HTMLSpanElement {
     const dot = document.createElement("span");
     dot.className = `now-dot now-dot-${state}`;
     const label = document.createElement("span");
     label.textContent = text;
     el.replaceChildren(dot, label);
+    return label;
   }
 
   liveData<{ currently_tending?: Run[] }>(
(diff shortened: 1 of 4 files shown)


site/src/components/CurrentlyTending.astro
```diff
@@ -42,7 +43,17 @@
...
+        const render = () => {
+          const elapsed = elapsedTime(r.started_at);
+          if (!elapsed) {
+            paint(el, "live", `tending ${r.repo} · ${workflow}${more}`);
+            return;
+          }
+          paint(el, "live", `tending ${r.repo} · ${workflow} · ${elapsed}${more}`);
+        };
+        render();
+        if (!Number.isNaN(startedMs)) setInterval(render, 1000);
```
tend-agent: Recreating both `.now-dot` and the label every second resets the dot's CSS pulse animation each tick (it never gets to 50% / opacity 0.35). Paint once, then update only the label's text:

```suggestion
        const baseText = `tending ${r.repo} · ${workflow}`;
        paint(el, "live", `${baseText}${more}`);
        const label = el.lastElementChild;
        const tick = () => {
          if (label) label.textContent = `${baseText} · ${elapsedTime(r.started_at)}${more}`;
        };
        if (!Number.isNaN(Date.parse(r.started_at))) {
          tick();
          setInterval(tick, 1000);
        }
```

tend-agent (commented):
One finding in `CurrentlyTending.astro`. The new `setInterval(render, 1000)` calls `paint()` each tick, and `paint()` does `el.replaceChildren(dot, label)` with a freshly-created `.now-dot` span. CSS animations restart from t=0 when the element is freshly inserted, so the dot's `now-pulse 2.6s ease-in-out infinite` animation is reset every second and never reaches the 50% keyframe (opacity 0.35) — the slow pulse is replaced by a barely-perceptible flicker. Updating just the label text avoids the issue; the dot pulse in `Activity.astro`'s `kind-tending` rows is fine because `updateTendingTimes` only touches `cell.textContent`.

site/src/components/CurrentlyTending.astro
max-sixty: Fixed in 45835d76 — paint returns the label so the tick updates only its textContent; the dot stays mounted and the pulse runs continuously.

tend-agent: Confirmed — `paint()` returns the label, the dot stays mounted, and only `label.textContent` updates per tick. Pulse keeps its phase.

PR #480, opened 2026-05-14
Add Codex engine alongside Claude

max-sixty:
## Summary

Adds `harness: claude | codex` to `.config/tend.yaml`. The generator
routes per harness: action ref, secret names, prompt invocation syntax.
Claude path output is byte-identical to before.

Two slices:

1. **Harness plumbing.** `harness` config field, generator routing,
   harness-aware `tend check` (verifies the right secret per harness),
   install-tend skill branched on harness for the auth-token step,
   docs updated across README, `docs/security-model.md`,
   `docs/tend.example.yaml`, `CLAUDE.md`. New tests including 8
   parametrized snapshot baselines for the Codex variants.

2. **`codex/action.yaml`.** New composite action at
   `max-sixty/tend/codex@v1`. Bypasses `openai/codex-action` and shells
   out to `codex exec` directly so we can support both auth modes from
   one path:
   - `OPENAI_API_KEY` (recommended; standard API billing).
   - `CODEX_AUTH_JSON` (subscription-funded — written to
     `~/.codex/auth.json`. OpenAI's docs explicitly discourage this on
     public repos; the install skill refuses it for public repos.)

   Installs the `tend-ci-runner` plugin via the new (alpha.17)
   `codex plugin add` non-interactive path. Builds Codex from source
   meanwhile — the npm release with `plugin add` hasn't shipped yet
   (PR openai/codex#21396). The `npm install -g @openai/codex` block is
   preserved as a commented swap-in for once it does.

## What `tend-ci-runner` becomes under Codex

A real Codex plugin: `plugins/tend-ci-runner/.codex-plugin/plugin.json`
declares it; `.agents/plugins/marketplace.json` exposes it as part of
the `tend` marketplace. Adopters install it (in CI) via the action's
`codex plugin marketplace add` + `codex plugin add` steps. Skills
themselves are unchanged — same `SKILL.md` files, same cross-skill
references using `/tend-ci-runner:NAME` syntax (Codex interprets these
correctly without translation, verified by smoke test).

## Shared system prompt

`shared/system-prompt.md` holds the harness-neutral conduct/scope rules.
Both actions consume it via envsubst (`\${BOT_NAME}`); each action
prepends a harness-specific skill-loading directive on top. Edits to
the shared file affect both harnesses.

## Caveats called out

- **Tend's own CI stays on Claude in this PR.** Flipping
  \`.config/tend.yaml\` to \`harness: codex\` requires \`v1\` to point at
  the release that includes \`codex/action.yaml\` — sequenced as a
  follow-up per \`TODO.md\`.
- **Cold cargo build is ~23 min on M-series.** Cached via Swatinem;
  cache misses re-pay. Acceptable for testing the path; revisit when
  alpha.17+ binaries ship to npm or by caching the built binary itself.
- **Plugin install is REPL-only on stable Codex (≤0.130, alpha.16).**
  Hence the source build at alpha.17.

## Test plan

- [x] \`pre-commit run --all-files\` clean
- [x] \`cd generator && uv run --frozen pytest\` — 226 passed
- [x] Local \`codex exec\` smoke test against the alpha.17 binary:
  marketplace adds, plugin installs, all 9 tend skills appear in skill
  list, \`\$triage\` invocation reads the right SKILL.md
- [x] Generator output for tend's own config is byte-identical to main
- [ ] CI green
- [ ] Once merged: cut release, then flip tend's own config to codex
  (separate PR per \`TODO.md\`)


codex/action.yaml
```diff
@@ -0,0 +1,491 @@
...
+        - The bot's user ID is available via
+          \`gh api users/${BOT_NAME} --jq .id\` if you need it for
+          \`author.id\` comparisons.
+        EOF
+        echo "Staged AGENTS.md at \$HOME/.codex/AGENTS.md"
+
+    # Mirror claude-code-action's fork-PR CLAUDE.md pinning so adopters who
+    # also rely on CLAUDE.md for prompt-injection defenses aren't worse off
+    # under the Codex engine.
+    - name: Pin CLAUDE.md to base branch (fork PRs)
```
tend-agent: This step pins `CLAUDE.md` from the base branch but leaves `AGENTS.md` untouched. Under the Codex engine, `AGENTS.md` is the equivalent instruction file — the earlier `Stage AGENTS.md` step's own comment ("alongside any AGENTS.md the adopter already has in their repo") confirms Codex loads the working-tree one. On a fork PR, `actions/checkout@v6` materializes `refs/pull/N/merge`, which includes any `AGENTS.md` the fork added; Codex then reads it alongside the trusted `$CODEX_HOME/AGENTS.md` and a malicious fork can override the bot's CI guidance (identity guard, `gh` token handling, etc.) when tend-review runs.

Mirroring claude-code-action's CLAUDE.md pin is the right intent, but the analogous file for Codex is `AGENTS.md`. Suggest adding a parallel block (or extending this one) to pin/remove `AGENTS.md` the same way, e.g. right after the existing CLAUDE.md branch:

```bash
if git show "origin/${DEFAULT_BRANCH}:AGENTS.md" > /tmp/base-agents.md 2>/dev/null; then
  cp /tmp/base-agents.md AGENTS.md
  echo "Pinned AGENTS.md to ${DEFAULT_BRANCH}"
else
  rm -f AGENTS.md
  echo "No AGENTS.md on ${DEFAULT_BRANCH} — removed fork version"
fi
```

(Subdirectory `AGENTS.md` files in the merged tree are also fork-controlled and would need similar treatment if Codex's project-doc resolver recurses; worth checking against Codex's actual behavior.)

tend-agent (commented):
Engine routing is well-isolated and the byte-identical Claude path is verified by the existing snapshots. Two things worth checking before merge:

**Fork-PR AGENTS.md is not pinned (security).** See inline comment.

**Hyphenated `$skill-name` parsing under Codex.** The PR description's smoke test mentions `$triage` (no hyphen) reading the right SKILL.md. Several generated prompts use hyphens — `$ci-fix`, `$review-runs`, and the staged AGENTS.md instructs `ALWAYS load $running-in-ci first`. If Codex's mention parser stops at the hyphen (treating `$ci` as the mention and `-fix` as literal), ci-fix / review-runs / nightly would silently miss their target skill. Worth a quick `codex exec` smoke test against `$ci-fix` or grepping Codex's mention regex to confirm hyphens are word characters there.

codex/action.yaml
max-sixty: Fixed in 4ba098f. Renamed the step to "Pin instruction files" and added an `AGENTS.md` pass that walks the working tree (`find . -name AGENTS.md`) so per-directory `AGENTS.md` files are also pinned/removed — Codex reads them per-dir, not just root. Also added a final pass that re-creates any `AGENTS.md` the fork deleted from base via `git ls-tree`, so a fork can't silently strip a tend-shipped guidance file.

max-sixty:
Re: hyphenated `$skill-name` parsing — verified locally with the alpha.17 binary. Codex resolves both correctly:

```
$ci-fix          → /Users/.../skills/ci-fix/SKILL.md
$running-in-ci   → /Users/.../skills/running-in-ci/SKILL.md
```

Mention parser handles hyphens. No action needed.

tend-agent (approved):

tend-agent (approved):

TODO.md
```diff
@@ -3,6 +3,27 @@
...
 of building it if revisited.
 
+## Cut tend over to harness = "codex" (post-release)
+
+The Codex harness landed but tend itself still runs on Claude. The cutover
+needs the release sequence:
+
+1. Land the harness support PR on `main`.
+2. Cut release `0.0.19` — bumping the `v1` tag to include `codex/action.yaml`.
+3. Edit `.config/tend.toml`: add `harness = "codex"` (and optionally
```
tend-agent: Stale `.config/tend.toml` and TOML assignment syntax — config switched to YAML in #454, and this PR's own `.config/tend.yaml` references use YAML mapping form. This step would render correctly as:

```suggestion
3. Edit `.config/tend.yaml`: add `harness: codex` (and optionally
```

tend-agent (commented):
The `engine` → `harness` rename is otherwise clean and consistent across config, generator, action wiring, and tests (snapshots verify byte-identical output for both harnesses; full suite green). Two small follow-ups inline.

One residual the inline comments don't cover: README still has a `## Engines` section (`README.md:182`) and the prose still calls them "engines" (line 184). If the rename is meant to apply to user-facing terminology too, the section heading and prose should follow; if "engine" is intentionally kept as the prose noun, then the `harness:` config key reads as a small surface inconsistency. Worth a quick call either way.

docs/security-model.md
```diff
@@ -158,16 +158,20 @@ The last row matters: `GITHUB_TOKEN` pushes don't trigger downstream workflows
...
 ### Token assignment
 
-Use a single bot token across all Claude workflows for consistent identity.
+Use a single bot token across all workflows for consistent identity.
 The merge restriction (ruleset) caps blast radius regardless of which token
 is used.
 
-Two tokens are needed:
+Two tokens are needed: the bot's PAT/App credential, plus an harness auth
+credential whose exact form depends on `harness` in `.config/tend.toml`.
```
tend-agent: Two small fixes touched by the rename: the article should be `a` (not `an`) before the consonant-sounding `harness`, and the config file is `.yaml` (since #454), not `.toml`.

```suggestion
Two tokens are needed: the bot's PAT/App credential, plus a harness auth
credential whose exact form depends on `harness` in `.config/tend.yaml`.
```

TODO.md
max-sixty: Fixed in d11feac.

docs/security-model.md
max-sixty: Fixed in d11feac.

max-sixty:
Re: README "Engines" section heading + prose — fixed in d11feac. Section is now `## Harnesses`; prose follows.

tend-agent (approved):

PR #481, opened 2026-05-14
site: unify right and left margins around one rail-plus-measure column

max-sixty:
Several passes to give the marketing site a single visual column instead of five competing edges.

## What changed

- **Right edge.** Page width was `1080px` with body paragraphs ending at 62ch, the activity table extending to the full page, the install snippet at `fit-content`, and the tagline at 32ch — five right edges. Page width is now `calc(--rail + --measure + 4rem) = 50rem`, so prose, the activity feed, stats, and the footer all share one right edge.
- **Left edge.** The header wordmark and footer credit now `margin-left: var(--rail)`, sitting at the same left position as every section eyebrow and paragraph. The "tend" wordmark in the header sits directly above the "tend" H1 in the hero.
- **Horizontal dividers.** Dropped all `<hr />` between sections; spacing now comes from `.page > section:not(.hero) { margin-top: 5rem }`. The section-eyebrow's leading tick still marks each new block, so the change reads as cleaner rather than less structured.
- **Header nav.** Removed the four anchor links. The page is short enough that scrolling beats clicking, the footer carries the same destinations, and the nav was already incomplete (missing "Recent ground" and "How it works").
- **Install snippet.** The two install pres (hero teaser + quick-start instruction) are unified on the quieter style — left ochre rule, no box. Dropped the `.install` override.

## Variables

- `--rail: 7rem` (margin-note column + gap) replaces magic `7rem` values in 4 places.
- `--measure-narrow: 48ch` replaces the one-off `32ch` tagline cap and `48ch` section-lead cap.
- `--page-width: 50rem` is now rem-based. It used to be `calc(--rail + 62ch + 4rem)`, but `ch` resolves at the consuming element; the header sets `font-size: 0.9rem` so its `ch` was smaller, making the header 4rem narrower than the page. rem keeps the layout stable regardless of child font-sizes.

## Other

- Renamed the "currently tending" indicator to lead with the word "currently".

## Notes for review

The footer still has its full-width `border-top` separator. With every other divider gone, it's now the only horizontal line on the page. I'd argue it reads as edge/chrome rather than a section break (it's full-width, not centered), but worth a glance.

### DIFF
--- a/site/src/components/CurrentlyTending.astro
+++ b/site/src/components/CurrentlyTending.astro
@@ -51,8 +51,8 @@
         const buildText = () => {
           const elapsed = elapsedTime(r.started_at);
           return elapsed
-            ? `tending ${r.repo} · ${workflow} · ${elapsed}${more}`
-            : `tending ${r.repo} · ${workflow}${more}`;
+            ? `currently tending ${r.repo} · ${workflow} · ${elapsed}${more}`
+            : `currently tending ${r.repo} · ${workflow}${more}`;
         };
         const label = paint(el, "live", buildText());
         if (!Number.isNaN(startedMs)) {
--- a/site/src/layouts/Base.astro
+++ b/site/src/layouts/Base.astro
@@ -35,12 +35,6 @@ const {
         <Logo size={28} static />
         <span>tend</span>
       </a>
-      <nav>
-        <a href="#what-tend-tends">What it tends</a>
-        <a href="#quick-start">Quick start</a>
-        <a href="#security">Security</a>
-        <a href="https://github.com/max-sixty/tend">GitHub</a>
-      </nav>
     </header>
     <main class="page">
       <slot />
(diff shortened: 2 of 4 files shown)


tend-agent (commented):
One thing: `.site-header .wordmark` (line 237) and `.site-footer .credit` (line 561) both pick up `margin-left: var(--rail)`, but `--rail` has no mobile override. Meanwhile the existing `@media (max-width: 760px)` block resets `.section-eyebrow` and `.section-lead` to `margin-left: 0`, and the mobile `.hero` override sets `padding: 2rem 0 2rem` (dropping its left rail too). So below 760px the wordmark and footer credit retain a 7rem indent while the hero h1, eyebrows, and section leads all sit flush at the gutter — the "tend wordmark sits directly above the tend h1" alignment from the PR body only holds on desktop.

Cleanest fix is to extend the mobile media block:

```css
@media (max-width: 760px) {
  .section-eyebrow { margin-left: 0; }
  .section-eyebrow::before { left: -1.5rem; }
  .section-lead { margin-left: 0; }
  .site-header .wordmark,
  .site-footer .credit { margin-left: 0; }
}
```

Happy to push that as a follow-up commit if you'd like.

tend-agent (approved):

PR #493, opened 2026-05-15
docs: align CODEX_AUTH_JSON references with the device-code mint flow

max-sixty:
Follow-up to #490, which switched the install-tend skill to mint Codex's `auth.json` into `/tmp/codex-tend` via `codex login --device-auth`. The README, security-model doc, and example config still described `CODEX_AUTH_JSON` as "the contents of `~/.codex/auth.json` after `codex login`". Cross-referencing skill + docs gave readers mismatched paths. Rephrased generically so the docs describe the secret value (the `auth.json` Codex writes after `codex login --device-auth`) without naming a specific home path. The runtime side — `codex/action.yaml` writing the secret to `$HOME/.codex/auth.json` on the GHA runner — is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ To use Tend, a project needs:
 - A GitHub account for the agent (for example this project's is **[@tend-agent](https://www.github.com/tend-agent))**
 - One of:
   - A Claude Max subscription (harness = "claude")
-  - An OpenAI API key, or a ChatGPT subscription via `~/.codex/auth.json` (harness = "codex")
+  - An OpenAI API key, or a ChatGPT subscription via a Codex `auth.json` (harness = "codex")
 
 Tend offers the default code & guidance for the agent. Specifically that means:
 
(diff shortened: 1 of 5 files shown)


tend-agent (commented):
Wording in the diff is accurate against the install skill (`codex login --device-auth` into `CODEX_HOME=/tmp/codex-tend`). Two parallel references in `codex/action.yaml` carry the same stale phrasing this PR is aligning — both are user-facing input docs, not the runtime write path you scoped out:

- The action top-level `description:` block ("subscription-funded ~/.codex/auth.json (the latter is officially discouraged...)").
- The `inputs.codex_auth_json.description` ("Contents of ~/.codex/auth.json — funds the run...").

`generator/src/tend/config.py` has a comment near `KNOWN_SECRETS_KEYS` with the same "(~/.codex/auth.json contents stored as a repo secret)" phrasing — internal so lower priority, but same alignment.

The `chmod 600` comment further down in `codex/action.yaml` that mentions `codex login` is about the on-runner write behavior — that one is correctly left alone.

Happy to push a follow-up commit covering those if you want it in this PR.

tend-agent (approved):

PR #498, opened 2026-05-15
install-tend: Claude installs codex and runs the device-code login

max-sixty:
Follow-up to #490 and #493. The Codex mint flow in section 7b previously had "have the user install codex (`npm i -g @openai/codex`) and run `codex login --device-auth`" — but Claude is already executing in the user's shell during install-tend, so it can do both. The user's only required action is the browser sign-in: Claude relays the URL + one-time code from codex's stdout and waits for the user to sign in as the dedicated bot ChatGPT account. Steps 1–3 are now Claude actions (install if missing, mkdir, run login, set secret, clean up); the inline narrative in step 2 spells out the relay.

### DIFF
--- a/plugins/install-tend/skills/install-tend/SKILL.md
+++ b/plugins/install-tend/skills/install-tend/SKILL.md
@@ -426,22 +426,25 @@ If not set:
 1. Install codex if missing (`npm i -g @openai/codex`) and create
    the mint dir: `mkdir -p /tmp/codex-tend`.
 
-2. Run the device-code login:
+2. Have the user run, in their own terminal:
 
    ```bash
    CODEX_HOME=/tmp/codex-tend codex login --device-auth
    ```
 
-   The dedicated `CODEX_HOME` isolates the bot's `auth.json` from
-   the user's personal `~/.codex/` — both coexist, no need to log
-   out of personal Codex. `--device-auth` prints a URL and a
-   one-time code; relay both to the user so they open the URL in
-   any browser and sign in as the dedicated bot ChatGPT account
-   chosen above (device-code is how they sign in as the bot without
-   juggling browser sessions). After they sign in, codex completes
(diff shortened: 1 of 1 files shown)


tend-agent (commented):
One concern about driving `codex login --device-auth` from inside the install-tend session:

`codex login --device-auth` prints the URL + one-time code to stdout immediately and then **blocks** until the user signs in (typically 30 s to several minutes). Driving it through a vanilla `Bash` call gives two failure modes:

1. The Bash tool's default 2-minute timeout will kill codex before slower sign-ins complete (max is 10 minutes, fine for most, tight for any user who has to dig out a 2FA prompt on another device).
2. The Bash tool returns captured stdout only on exit, so Claude can't read the URL+code while codex is still polling — there's nothing to relay until codex has already exited (at which point the relay is moot).

The Claude OAuth path side-steps the analogous shape via [`scripts/oauth-token.sh`](https://github.com/max-sixty/tend/blob/3e6a41aae405a2a789ffa23f7dafc5afd3f43e0b/plugins/install-tend/skills/install-tend/scripts/oauth-token.sh) (PTY allocation via `script(1)`, capture-to-file, post-hoc token extraction). The codex path now asks Claude to drive a blocking interactive command inline without similar plumbing. Worth either:

- spelling out in step 2 that the codex invocation must be a `run_in_background: true` Bash call (and probably tailing the file the action writes / reading stdout via the streaming output path) — i.e. give the agent the operational recipe so a fresh Claude doesn't default to inline, OR
- adding a small wrapper script analogous to `oauth-token.sh` that backgrounds codex, extracts the URL+code, and waits for completion. That keeps the skill prose ("Claude runs the device-code login") matching what's actually executable.

As written, a fresh agent following step 2 verbatim is likely to hit one of the two failure modes the first time it executes.

tend-agent (approved):

PR #515, opened 2026-05-15
Pin action to release tag; remove the v1 floating branch

max-sixty:
## Problem

`max-sixty/tend@v1` was a **branch** force-pushed to match `main` on every merge by `update-action-branch.yaml`. A ruleset was then added protecting `v1` like `main`, so the force-push failed with `GH013: Cannot update this protected ref`, freezing every consumer at the last successful commit. Protecting `v1` and auto-force-pushing it are contradictory requirements.

## Fix

Remove the floating ref entirely instead of working around it with a bypass credential.

The generator pins each generated workflow's `uses:` to its own release version (`max-sixty/tend@X.Y.Z`, an immutable tag). The PyPI package version and the git tag are the same commit, so a workflow's `# Generated by tend X.Y.Z` header and its action ref are guaranteed to match. Consumers move forward through the existing nightly `uvx tend@latest init` regen pulling a newer generator — a reviewable workflow-file diff in their own repo, not an invisible shared-branch force-push.

This is also the standard GitHub Actions supply-chain posture (pin a third-party action to an immutable ref).

## What reviewers should look at

- `generator/src/tend/workflows.py` — `_action_ref()` now returns `max-sixty/tend@{version('tend')}` (`/codex@…` for the Codex harness). The docstring documents the provenance guarantee and the in-tree-version hazard.
- Deleted `.github/workflows/update-action-branch.yaml`.
- `.config/tend.yaml` — dropped `protected_branches: ["v1"]`. The generic `protected_branches` mechanism (and its tests) is retained; only tend's own use of it for `v1` is removed.
- `generator/tests/conftest.py` / `tests/__init__.py` — the regtest scrubber now normalizes the pinned version in the action ref as well as the header, so release bumps don't churn snapshots; `ACTION_VERSION` is shared for the literal assertions. The 31 snapshot updates are all the single ref line.
- `docs/security-model.md` — new "Action distribution integrity" boundary: release tags on `max-sixty/tend` must be immutable (a `tag` ruleset restricting `update`/`deletion`, no bypass). **This ruleset is created out of band; it is not part of this diff.**
- `.claude/skills/running-tend/SKILL.md` — consumer discovery searched `max-sixty/tend@v1`, which post-change matches nothing. Now a version-agnostic bare-token `gh search code 'max-sixty/tend'` bounded by the existing path filter (also catches Codex consumers, which the `@v1` query silently missed).

## Follow-up (not in this PR)

1. Create the tag-immutability ruleset on `max-sixty/tend`.
2. After a release ships with this generator, remove `refs/heads/v1` from the existing branch ruleset and delete the stale `v1` branch and `v1` tag.

Consumers auto-migrate on their next nightly regen after the next release; no consumer action required.

## Testing

230 tests pass, 0 snapshot failures, pre-commit clean. Code-reviewer iterated to convergence (clean design; it caught and I fixed a consumer-discovery regression).


tend-agent (commented):
Solid cutover — pinning to the released `version("tend")` makes the action ref and the header version provenance match the package on PyPI, and the test scrubber + literal-assertion split is clean. One observation about the security doc landing slightly ahead of the enforcement step it describes.

docs/security-model.md
```diff
@@ -51,6 +51,19 @@ branch protection) prevents the bot from merging to protected branches
 default branch isn't protected. Everything below is defense in depth — useful,
 but not load-bearing.
 
+**Action distribution integrity.** Generated workflows pin the composite
+action to the generator's own release version (`max-sixty/tend@X.Y.Z`), never
+a floating ref. Release tags on `max-sixty/tend` are immutable: a tag ruleset
+restricts `update` and `deletion` (leaving `creation` open so the release can
+push a new `X.Y.Z`), with no bypass for write-access actors. A leaked bot
```
tend-agent: Present tense ("are immutable: a tag ruleset restricts...") reads as if enforcement is in place, but per the PR body the tag ruleset is created out of band as a follow-up. Until that lands, an adopter pinned to `@X.Y.Z` is still exposed to tag-rewrite by anyone with write access on `max-sixty/tend` — the doc claims a boundary that isn't operational yet. Optional: hedge it inline so the security model reads correctly between this merge and the ruleset landing.

```suggestion
push a new `X.Y.Z`), with no bypass for write-access actors. (The ruleset is applied out of band — until it is in place this boundary is operational rather than enforced.) A leaked bot
```

tend-agent (approved):

PR #535, opened 2026-05-16
install-tend: unify on the chain model (admin-gated refs + env policy pin)

max-sixty:
## Why

Two related problems, converged into one PR through iteration:

1. **`install-tend` referenced tend-repo-internal paths** (`docs/security-model.md`, `docs/tend.example.yaml`, `.github/workflows/codex-auth-refresh.yaml`). The plugin ships only `./plugins/install-tend` to adopters, so those refs were dangling for the actual audience.
2. **The security guidance the skill carried prescribed `required_reviewers` + `prevent_self_review` as the gate for the no-merge path.** That worked, but a simpler model exists: gate the *refs* every privileged workflow runs on, instead of gating each *deploy*. If both default-branch merges *and* tag operations are admin-only, every privileged code path chains to an admin operation, and the bot (write, not admin) reaches neither ref or any environment pinned to them. Same threats covered, structurally simpler, matches what tend itself runs.

### `install-tend` skill self-containment

- Every `docs/…` reference repointed to `${CLAUDE_SKILL_DIR}/references/…` (shipped inside the plugin subtree).
- New `references/security-model.md` — install-scoped distillation: the chain model, token-leak table, Codex `auth.json` dedicated-account rationale, static-secret rotation, token assignment, and how tokens flow through workflows.
- New `references/tend.example.yaml` — byte-identical mirror of canonical `docs/tend.example.yaml`, kept in sync by a new pre-commit local hook (`sync-install-tend-references`) that fails the commit on drift or missing canonical source. Mirror behavior empirically verified.

### Security guidance unified on the chain model

The skill now prescribes **two admin-gated ref classes**, parallel-shaped rulesets:

| Ref class | Ruleset | Rules | Bypass |
|---|---|---|---|
| Default branch | `Merge access` | `update` | admin |
| All tags (`~ALL`) | `Tag operations` | `creation`, `update` | admin |

`update` is what force-push of an existing tag fires server-side — required alongside `creation` because without it a write-access bot could re-point an admin-pushed tag to a malicious commit. `deletion` is not chain-critical (recreation is already blocked by `creation`); it's left out so the rule list stays minimal.

Environments pin their `deployment_branch_policy` to those refs (`name=main, type=branch` for continuous-deploy, `name=*, type=tag` for release-on-tag, or both). Release/deploy workflows trigger only on `push: tags:` or `push: branches: [<default>]` — never `pull_request`, because a `pull_request` workflow runs the YAML at the PR head, which a bot can write, breaking the chain at the workflow file itself.

`docs/security-model.md` trimmed from 377 → ~160 lines (summary + risks only); operational reference content moved into `references/security-model.md` where the install agent actually meets it. "What we do" reframed as two load-bearing boundaries (merge restriction + environment chain) rather than a single boundary plus disclaimed defense-in-depth.

### Live application

Applied parallel rulesets on:
- `max-sixty/tend` (Tag operations id 16490366) + `release` env pinned to `*` tags (was unguarded, leaving PYPI_TOKEN reachable from any ref).
- `max-sixty/worktrunk` (Tag operations id 16490367); its `release` env already had a `v*` policy, so the chain now holds.

Tend retains its existing `Release tag immutability` ruleset (`refs/tags/[0-9]*`, no bypass on `update`+`deletion`) as the publisher-specific layered defense for downstream consumers pinning `max-sixty/tend@X.Y.Z`. The skill mentions this case parenthetically as the "publisher uplift" option, not in the default recipe.

## What this changes for adopters

Running `install-tend` on a typical adopter repo now sets up:
- Merge ruleset on the default branch (unchanged).
- **New:** Tag ruleset on all tags, admin bypass.
- Environments for release/deploy secrets, pinned to the admin-gated refs.

The skill no longer prescribes per-deploy `required_reviewers` / `prevent_self_review`. That mechanism is mentioned as one of the "more complicated approaches available if you need them"; the packaged recipe is the structurally simpler chain.

## Provenance / verification

The GitHub mechanics this rests on were verified against primary sources across the iteration loop, including the empirical test that force-push of an existing tag fires the `update` rule (a force-push test against `0.0.18` was rejected with `Cannot update this protected ref` — confirming `update` is the relevant rule for tag re-pointing).

## Test plan

- `generator/` tests: 242 pass, 0 snapshot changes (no generator code touched in this PR).
- `pre-commit run --all-files` green, including the new `sync-install-tend-references` hook (drift and source-missing failure paths exercised manually).
- `origin/main` merged twice during the iteration (releases 0.0.21–0.0.23, codex-cutover, BOT_TOKEN→TEND_BOT_TOKEN rename, install-test workflow, install-test refactor). Conflicts resolved with theirs in cases my branch hadn't independently modified the file.

## Not in this PR

- `worktrunk`'s `release.yaml` still triggers on `pull_request` in addition to `push: tags:`. The chain now blocks the env-secret path (env rejects feature-branch refs), but the workflow design retains the "primary gap" shape — a bot-modified workflow YAML on a `pull_request` could read whatever repo-level secrets it declares. Deferred to a separate worktrunk-internal change.


plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -103,12 +103,72 @@ warning. Classify each non-bot secret and act now — don't defer:
...
+  gh api --method POST "/repos/$REPO/environments/$ENV/deployment-branch-policies" \
+    -f name='v*' -f type=tag
+  ```
+
+  Then find what else can reach a publishing job with no merge. The merge
+  ruleset gates only default-branch merges; a tag push, release, dispatch,
+  `deployment`, `pull_request_target`, or `schedule` trigger bypasses it.
+  Sweep every deploy/publish workflow and read each hit; this grep misses
+  reusable workflows in other repos and over-matches `pull_request:` and
+  step inputs:
```
tend-agent: Small caveat tweak: the regex doesn't contain `pull_request:` (the closest alternative is `pull_request_target`), so it doesn't over-match `pull_request:` lines. The actual over-match is `pull_request_target` (and `workflow_run`) hitting expression references like `${{ github.event.pull_request_target.head.sha }}` in step inputs, not trigger declarations.

```suggestion
  reusable workflows in other repos and over-matches `pull_request_target`
  references in expressions and step inputs:
```

tend-agent (dismissed):

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -103,12 +103,59 @@ warning. Classify each non-bot secret and act now — don't defer:
...
-  the original.
+  workflow run, including ones a write-access bot can trigger with no
+  merge. Don't allowlist them. Migrate each to a GitHub Environment whose
+  deployment policy pins to the admin-gated refs from §3 (the default
+  branch, and the release tag pattern for tag-triggered deploys). The bot
+  can reach neither ref, so it cannot reach the secret.
+
+  Migrate the secret: recreate it on the Environment, delete the
+  repo-level copy (confirm via `AskUserQuestion` first), and set
+  `environment: <name>` on the publishing job.
```
tend-agent, line 125: This no longer closes several of the merge-free paths the skill tells installers to sweep below. An environment deployment policy only filters the `GITHUB_REF`; it does not require a trusted actor to approve a run on an allowed ref. A leaked write-scoped bot token can still trigger an allowed-ref privileged run without changing that ref: `workflow_dispatch` can be run via CLI/API on the default branch, `deployment` can target the default branch/tag ref, and GitHub's release API allows users with push access to create a release for an existing admin-created tag, which then fires `release: published`. With no required reviewers / prevent-self-review, the environment admits the job and injects the secret. Please either restore the required-reviewer gate for these trigger classes or narrow the claim to workflows that are only reachable by creating/updating the protected ref.

plugins/install-tend/skills/install-tend/references/tend.example.yaml
```diff
@@ -0,0 +1,316 @@
...
+# ## Secrets
+#
+# Repo secrets, by harness:
+#
+# | Harness   | Required                                                                |
+# |----------|--------------------------------------------------------------------------|
+# | `claude` | `BOT_TOKEN` + one of `CLAUDE_CODE_OAUTH_TOKEN` or `ANTHROPIC_API_KEY`    |
+# | `codex`  | `BOT_TOKEN` + one of `CODEX_AUTH_JSON` or `OPENAI_API_KEY`               |
+#
+# `BOT_TOKEN` is the bot account's PAT (scopes below).
```
tend-agent:
```suggestion
# | Harness   | Required                                                                     |
# |----------|-------------------------------------------------------------------------------|
# | `claude` | `TEND_BOT_TOKEN` + one of `CLAUDE_CODE_OAUTH_TOKEN` or `ANTHROPIC_API_KEY`    |
# | `codex`  | `TEND_BOT_TOKEN` + one of `CODEX_AUTH_JSON` or `OPENAI_API_KEY`               |
#
# `TEND_BOT_TOKEN` is the bot account's PAT (scopes below).
```

tend-agent (commented):
CI is also failing because `plugins/install-tend/skills/install-tend/references/tend.example.yaml` is stale relative to `docs/tend.example.yaml`; the inline suggestion below matches the pre-commit diff.

tend-agent (commented):
Not approving this revision yet; the existing bot threads still apply on the current head.

The release-triggered deploy path is still open in the chain model: publishing a release for an already admin-created matching tag does not create, update, or delete the tag, but GitHub runs `release: published` workflows with `GITHUB_REF` set to that tag ref. A write-scoped bot token can create releases via the Releases API, so an environment policy that allows the release tag pattern would admit that run. The install guidance needs an extra gate for `release`-triggered secret jobs, or it needs to exclude/narrow that trigger class.

CI is also still failing because `plugins/install-tend/skills/install-tend/references/tend.example.yaml` is stale relative to `docs/tend.example.yaml`; the existing inline suggestion matches the pre-commit diff.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: Verified against GitHub's docs (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows): for `workflow_dispatch`, `deployment`, and `release: published`, the workflow file runs from the **default branch** (regardless of the dispatched ref or the release's tag). The bot has write but cannot push to the default branch (merge ruleset), so it cannot inject code into any of these workflows. The env's deployment_branch_policy still filters by `GITHUB_REF`, and the workflow code is admin-vetted, so there is no code-injection or secret-exfiltration vector through these triggers. The chain holds. What the bot can do is *trigger a legitimate deploy at an unintended time* (rate-limit waste, off-cycle release), which is a real but lower-severity class than the original framing suggested. Required-reviewers remains useful as opt-in defense-in-depth for adopters who care about gating that triggering; it is not load-bearing against the threats the chain covers. Resolving.

generator/src/tend/workflows.py
```diff
@@ -970,6 +970,101 @@ def _apply_extras(wf: GeneratedWorkflow, wf_cfg: WorkflowConfig) -> GeneratedWor
...
+            f"          OPENAI_KEY: ${{{{ secrets.{cfg.openai_key_secret} }}}}"
+        )
+        harness_names = f"{cfg.codex_auth_json_secret} or {cfg.openai_key_secret}"
+        harness_check = '"$CODEX_AUTH$OPENAI_KEY"'
+
+    content = f"""\
+{HEADER}
+name: tend-install-test
+on:
+  pull_request:
```
tend-agent, line 1013: This generated workflow reads `TEND_BOT_TOKEN` and the harness secret from a `pull_request` run. Once `tend-install-test.yaml` lands on the default branch, a leaked write-scoped bot token can push a same-repo branch that changes `.github/workflows/tend-install-test.yaml` and opens a PR touching `.config/tend.yaml`; same-repo `pull_request` runs receive repo secrets, so that branch can exfiltrate exactly the credentials the security model is trying to contain. The fact that nightly later removes the file still leaves an exploitable window after install. Please avoid putting those secrets in a PR-head workflow, for example by making the secret-presence check a maintainer/default-branch `workflow_dispatch` step or by splitting it from the PR-head generator-drift check.

tend-agent (commented):
Not approving this revision yet; one security issue remains in the new install-test workflow.

tend-agent (commented):
Not approving this revision yet; the existing install-test thread still applies on the current head.

`generate_install_test()` still emits a same-repo `pull_request` workflow that reads `TEND_BOT_TOKEN` and the harness secret from repo secrets. Once this file lands on the default branch, a leaked write-scoped bot token can push a same-repo branch that modifies `.github/workflows/tend-install-test.yaml` and opens a PR touching `.config/tend.yaml`; GitHub will run the PR-head workflow with repo secrets, so the branch can exfiltrate the same credentials the security model is trying to contain. The short-lived/nightly-cleanup window still leaves an exploitable period after install. Please split the secret-presence check onto a trusted/default-branch path, or avoid generating a PR-head workflow that reads these secrets.

plugins/install-tend/skills/install-tend/references/security-model.md
```diff
@@ -0,0 +1,136 @@
...
+rules on top; install-tend packages the simplest configuration that holds
+the chain.
+
+Deploy and publish workflows declare a GitHub Environment whose
+`deployment_branch_policy` lists only those admin-gated refs (the default
+branch and/or the release tag pattern). Release secrets live in those
+environments, not at repo level. A leaked bot token can push a branch or a
+non-release tag, but neither ref matches an admin-gated policy entry, so
+the deploy job is rejected before it can read the secret. No admin
+operation → no admin-gated ref → no environment access → no secret.
```
tend-agent: The reference now tells installers to protect all tags with the tag ruleset, but this paragraph still describes only the old release-tag-pattern policy and even says the bot can push a non-release tag. That makes the chain look weaker/different from the actual packaged recipe.

```suggestion
Deploy and publish workflows declare a GitHub Environment whose
`deployment_branch_policy` lists only those admin-gated refs (the default
branch and/or all tags). Release secrets live in those environments, not at
repo level. A leaked bot token can push a branch, but a non-default branch
does not match the branch policy; it also cannot push any tag. The deploy job
is rejected before it can read the secret. No admin operation → no
admin-gated ref → no environment access → no secret.
```

tend-agent (commented):
Not approving this revision yet; the existing install-test thread still applies, and the new all-tags model left a couple of stale release-tag-pattern references.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -661,7 +754,8 @@ line picks the row that matches the chosen harness):
 
 - [ ] Config: `.config/tend.yaml` created (with `harness` set if Codex)
 - [ ] Workflows: generated in `.github/workflows/`
-- [ ] Ruleset: merge restriction on default branch, admin bypass
+- [ ] Rulesets: merge restriction on default branch (admin bypass), release tag operations (admin bypass)
+- [ ] Release/deploy secrets: environment-protected; the environment's deployment-branch-policies list only the admin-gated refs from §3 (default branch and/or release tag pattern)
```
tend-agent: Same stale model in the final checklist: §3 now creates `Tag operations` over `~ALL`, and §1 verifies a `name=*`, `type=tag` environment policy. Leaving this as release-tag-pattern language can send the installer back to the older per-pattern setup.

```suggestion
- [ ] Rulesets: merge restriction on default branch (admin bypass), tag operations on all tags (admin bypass)
- [ ] Release/deploy secrets: environment-protected; the environment's deployment-branch-policies list only the admin-gated refs from §3 (default branch and/or all tags)
```

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -257,9 +307,59 @@ gh api "repos/$REPO/rulesets" --method POST --input - << 'EOF'
...
+
+**Release/deploy workflow design.** Workflows that use release or deploy
+secrets must trigger on `push: tags:` (release) or `push: branches: [main]`
+(continuous deploy from the default branch), and reference an Environment
+(§1). Don't trigger on `pull_request`. A `pull_request` workflow runs the
+YAML at the PR's head ref, which a bot can write, so the workflow code is
+no longer admin-vetted and the chain breaks at the workflow file itself.
+Other triggers (`workflow_dispatch`, `release: published`, `deployment`,
+`schedule`) read the workflow file from the default branch, so they keep
+the chain.
```
tend-agent: This sentence contradicts the packaged recipe immediately above. `workflow_dispatch`, `release: published`, `deployment`, and `schedule` are not `push: tags:` or `push: branches: [main]` triggers, and some of them can be initiated by a write-scoped bot against an otherwise allowed ref. Leaving this as "they keep the chain" makes the installer accept exactly the merge-free trigger class this section is trying to rule out.

```suggestion
`workflow_dispatch`, `release: published`, `deployment`, and `schedule`
are not part of the packaged recipe. If a repo keeps one, treat it as a
custom design and add a separate trigger-specific proof before migrating
release or deploy secrets.
```

tend-agent (commented):
Not approving this revision yet; the existing install-test thread is still unresolved, and the updated all-tags model left a few stale or contradictory install docs.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -661,7 +761,8 @@ line picks the row that matches the chosen harness):
 
 - [ ] Config: `.config/tend.yaml` created (with `harness` set if Codex)
 - [ ] Workflows: generated in `.github/workflows/`
-- [ ] Ruleset: merge restriction on default branch, admin bypass
+- [ ] Rulesets: merge restriction on default branch (admin bypass), release tag operations (admin bypass)
+- [ ] Release/deploy secrets: environment-protected; the environment's deployment-branch-policies list only the admin-gated refs from §3 (default branch and/or release tag pattern)
```
tend-agent: These checklist entries still describe the old release-tag-pattern setup even though §3 now creates a tag ruleset over all tags.

```suggestion
- [ ] Rulesets: merge restriction on default branch (admin bypass), tag operations on all tags (admin bypass)
- [ ] Release/deploy secrets: environment-protected; the environment's deployment-branch-policies list only the admin-gated refs from §3 (default branch and/or all tags)
```

plugins/install-tend/skills/install-tend/references/security-model.md
```diff
@@ -0,0 +1,140 @@
...
```
tend-agent: This shipped reference still says environments list the old release tag pattern and that the bot can push a non-release tag. Under the current recipe the tag ruleset covers `~ALL`, so this should match the all-tags model.

```suggestion
Deploy and publish workflows declare a GitHub Environment whose
`deployment_branch_policy` lists only those admin-gated refs (the default
branch and/or all tags). Release secrets live in those environments, not
at repo level. A leaked bot token can push a branch, but it cannot push
the default branch or any tag, so no bot-pushed ref matches an
admin-gated policy entry. The deploy job is rejected before it can read
the secret. No admin operation → no admin-gated ref → no environment
access → no secret.
```

generator/src/tend/workflows.py
max-sixty: Scoped out of this PR — the install-test workflow is a separate concern (#540 / #544 upstream) and the fix is on the workflow design, not on this PR's chain/install-tend changes. Tracking as follow-up. Leaving this thread unresolved as the real concern remains.

plugins/install-tend/skills/install-tend/references/security-model.md
max-sixty: Addressed in f305ee3.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: Addressed in f305ee3.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: Addressed in f305ee3.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: Addressed in f305ee3.

plugins/install-tend/skills/install-tend/references/security-model.md
max-sixty: Addressed in f305ee3.

plugins/install-tend/skills/install-tend/references/security-model.md
```diff
@@ -0,0 +1,141 @@
...
+the chain.
+
+Deploy and publish workflows declare a GitHub Environment whose
+`deployment_branch_policy` lists only those admin-gated refs (the default
+branch and/or all tags). Release secrets live in those environments, not
+at repo level. A leaked bot token can push a non-default branch, but it
+cannot push to the default branch and cannot push any tag, so no
+bot-pushed ref matches an admin-gated policy entry. The deploy job is
+rejected before it can read the secret. No admin operation → no
+admin-gated ref → no environment access → no secret.
```
tend-agent, line 46: This reference now presents the chain as complete once the environment policy is pinned to default-branch/all-tag refs, but that only holds for workflows whose trigger itself is limited to those ref updates. A `release: published`, `workflow_dispatch`, `deployment`, or `schedule` job can still be initiated by a write-scoped bot against an allowed ref, so the shipped reference needs the same push-only trigger caveat that §3 now has.

```suggestion
Deploy and publish workflows declare a GitHub Environment whose
`deployment_branch_policy` lists only those admin-gated refs (the default
branch and/or all tags). Release secrets live in those environments, not
at repo level. A leaked bot token can push a non-default branch, but it
cannot push to the default branch and cannot push any tag, so no
bot-pushed ref matches an admin-gated policy entry. The deploy job is
rejected before it can read the secret. No admin operation → no
admin-gated ref → no environment access → no secret.

That guarantee assumes the privileged workflow is reached only through
those ref updates: `push: tags:` for release/publish jobs or
`push: branches: [<default-branch>]` for continuous deploy. Other
triggers (`workflow_dispatch`, `release: published`, `deployment`,
`schedule`, and chained dispatches) need a separate trigger-specific proof,
usually required reviewers on the Environment, before moving release or
deploy secrets there.
```

tend-agent (commented):
Not approving this revision yet; the existing install-test thread still applies, and the shipped security reference needs the same trigger precondition as the install walkthrough.

plugins/install-tend/skills/install-tend/references/security-model.md
max-sixty: Addressed in c933b6b — the trigger-specific containment caveat is now in both references/security-model.md and docs/security-model.md so they describe the same boundary.

tend-agent (commented):
Not approving this revision yet; the docs/reference changes now carry the trigger caveat, but the existing install-test thread still applies on the current head.

`generate_install_test()` still emits a same-repo `pull_request` workflow that reads `TEND_BOT_TOKEN` and the harness secret from repo secrets. Once this file lands on the default branch, a leaked write-scoped bot token can push a same-repo branch that modifies `.github/workflows/tend-install-test.yaml` and opens a PR touching `.config/tend.yaml`; GitHub will run the PR-head workflow with repo secrets, so the branch can exfiltrate the same credentials the security model is trying to contain. Please split the secret-presence check onto a trusted/default-branch path, or avoid generating a PR-head workflow that reads these secrets.

generator/src/tend/workflows.py
max-sixty: The model already classifies these as leak-acceptable. The leak table in `references/security-model.md` treats `TEND_BOT_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, and `CODEX_AUTH_JSON` as long-lived, exfiltrable, bounded by merge restriction (and, for harness tokens, dedicated-account scope). The chain isn't trying to contain those; it's containing `CLOUDFLARE_API_TOKEN` and `CODEX_REFRESH_PAT`, which sit behind env policies pinned to `main` and aren't reachable from a feature-branch `pull_request` run (verified against the live env config on this repo).

Does install-test reach anything the chain promises to contain that isn't already in that leak table? If so, name the specific secret. Otherwise this is the accepted primary-gap class and the fix belongs on the workflow's design (tracked separately as #540/#544), not on this PR.

tend-agent: You're right. I don't see a specific env-protected secret reached by `tend-install-test`: it only reads `TEND_BOT_TOKEN` plus the harness auth secret (`CLAUDE_CODE_OAUTH_TOKEN`/`ANTHROPIC_API_KEY` or `CODEX_AUTH_JSON`/`OPENAI_API_KEY`), and the updated model explicitly treats those repo-level bot/runtime credentials as leak-acceptable within the merge-restriction boundary.

So this does not break the chain this PR is documenting. The concern is still real for the install-test workflow design, but it belongs with #540/#544 rather than blocking this chain/install-tend change. I withdraw this thread as a blocker.

max-sixty:
FYI am debugging why Codex seems to be so bad at following the `tend` skills when doing reviewing...

PR #537, opened 2026-05-16
review-reviewers: migrate to the Codex harness

max-sixty:
## Summary

`review-reviewers` is the only custom (non-generated) tend workflow and was the last one still on the Claude harness — every generated `tend-*.yaml` flipped to Codex in #520. This also retires the last `@v1` floating-ref consumer in this repo; everything else now pins immutable `@X.Y.Z` after #515.

Once this lands, `CLAUDE_CODE_OAUTH_TOKEN` is no longer referenced by any workflow on `main` and can be removed (separate decision — flagged below).

## Workflow changes

- `max-sixty/tend@v1` → `max-sixty/tend/codex@0.0.22`
- `claude_code_oauth_token` → `codex_auth_json: ${{ secrets.CODEX_AUTH_JSON }}`
- Added `model: gpt-5.5`; prompt switched to `$review-reviewers ${{ matrix.repo }}`
- Matrix now mirrors `data/consumers.json` (5 repos: `PRQL/prql`, `max-sixty/cargo-affected`, `max-sixty/tend`, `max-sixty/worktrunk`, `numbagg/numbagg`). The previous hand-list was missing `cargo-affected` and `numbagg`.

The existing `# TODO: Move repo list to a config file` is still live — `consumers.json` is that file; a follow-up could have the workflow generate the matrix dynamically via `fromJSON` and a setup job, retiring the TODO. Out of scope here.

## Skill changes

`plugins/tend-ci-runner/skills/review-reviewers/SKILL.md` generalizes the cost-discipline section so the same prose works under either harness:

- "Haiku subagents" → "cheap subagent (e.g. Haiku / gpt-mini)"
- "main Opus agent" → "main agent"

The concrete subagent call is left to each harness's native mechanism (Claude: `Agent` tool with `model: "haiku"`; Codex: subagent spawn with a gpt-mini-class model). Pattern, gates, and outputs are unchanged.

## Test plan

- [ ] CI green (test, test-worker, lint)
- [ ] After merge: trigger `review-reviewers.yaml` via `workflow_dispatch` and watch all 5 matrix entries (`fail-fast: false`)
- [ ] Eyeball token usage on the first run — openai/codex#16548 reports the subagent model selection being ignored at codex 0.118.0; the action installs 0.128.0 and we want to confirm the cheap subagent actually runs on the requested model, not silently on gpt-5.5
- [ ] After a clean run, decide whether to drop `CLAUDE_CODE_OAUTH_TOKEN` from repo secrets (now unreferenced)


.github/workflows/review-reviewers.yaml
```diff
@@ -35,9 +38,10 @@ jobs:
           fetch-tags: true
           token: ${{ secrets.BOT_TOKEN }}
 
-      - uses: max-sixty/tend@v1
+      - uses: max-sixty/tend/codex@0.0.22
```
tend-agent: This switches the hourly job onto the Codex action, but the `review-reviewers` skill still shells out through `TARGET_REPO=$ARGUMENTS ${CLAUDE_PLUGIN_ROOT}/scripts/list-recent-runs.sh`. The Codex action installs the plugin with `codex plugin add` but does not export `CLAUDE_PLUGIN_ROOT`, so Step 1 will expand this to `/scripts/list-recent-runs.sh` and the workflow will fail before it surveys any runs. Please either set an equivalent plugin-root env var in the Codex action or update the skill to locate `scripts/list-recent-runs.sh` in a harness-neutral way before flipping this workflow.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -299,11 +299,11 @@ Use `Agent` with `model: "haiku"` and a prompt like:
...
 Review the subagent's summary. If all outputs are accepted and no sanity-check flags, skip to Step 6 (summary). If concerning outcomes exist, continue to Step 3.
 
-## Step 3: Investigate concerning outcomes via Haiku subagent
+## Step 3: Investigate concerning outcomes via cheap subagent
 
-For runs with negative outcome signals (or suspicious lack of output), spawn another Haiku subagent to download and inspect the specific session logs.
+For runs with negative outcome signals (or suspicious lack of output), spawn another cheap subagent to download and inspect the specific session logs.
 
-Use `Agent` with `model: "haiku"` and a prompt like:
+Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
```
tend-agent, line 306: This generalizes the subagent wording, but the very next download instruction is still hard-coded to `--pattern 'claude-session-logs*'`. The Codex harness uploads artifacts as `codex-session-logs*` (`codex/action.yaml`'s Compute artifact name step), so any negative outcome after the cutover will fail to fetch the logs it needs for diagnosis. Please include both artifact prefixes or branch the instruction by harness.

tend-agent (commented):
I found a couple of cutover blockers in the Codex path.

.github/workflows/review-reviewers.yaml
```diff
@@ -35,9 +38,10 @@ jobs:
           fetch-tags: true
           token: ${{ secrets.TEND_BOT_TOKEN }}
 
-      - uses: max-sixty/tend@v1
+      - uses: max-sixty/tend/codex@0.0.22
```
tend-agent: This still won't pick up the `CLAUDE_PLUGIN_ROOT` fix from #539, because the workflow runs the immutable `max-sixty/tend/codex@0.0.22` tag and that tag's `codex/action.yaml` only runs `codex plugin add tend-ci-runner@tend` without exporting the installed plugin root. Even after #539 merges to `main`, this hourly job will keep expanding `${CLAUDE_PLUGIN_ROOT}` to empty until the pin moves to a released tag that contains that action change, or this PR makes `review-reviewers` locate `scripts/list-recent-runs.sh` without relying on that env var.

tend-agent (commented):
The previous artifact-prefix thread is still unresolved as well; this comment is just calling out the remaining tag-pinning dependency explicitly.

tend-agent (approved):

PR #540, opened 2026-05-16
install-tend: add --with-install-test for one-shot pre-merge verification

max-sixty:
## Summary

- `tend init --with-install-test` generates a one-shot `tend-install-test.yaml` that fires on the install PR (via `on: pull_request:` with paths filter on `.config/tend.yaml` and `.github/workflows/tend-*.yaml`) and verifies secrets are set + committed workflows match the generator's output.
- `init` now removes any `tend-*.yaml` not in the current run's generated set, so the install-test file disappears on the first nightly regen (which runs without the flag). Also cleans up after `enabled: false` workflows and renamed generators.
- Install skill passes the flag at step 2 and adds a "wait for `tend-install-test` to pass before merging" check to the checklist.

The workflow pins the regen to the version in the committed header (so a tend release mid-PR doesn't fail the drift check), installs `uv` (not on `ubuntu-24.04` by default), and runs `git remote set-head origin --auto` so `master`-default repos match.

The tightened `$REPO` canonicality note acknowledges what the skill actually does: every API call uses `--repo "$REPO"` explicitly, so a fork-clone install only needs the variable overridden — no remote surgery.

## Test plan

- [x] `uv run pytest` — 240 passing (8 new integration tests + 2 regtest snapshots for Claude/Codex variants)
- [x] `pre-commit run --all-files` — clean
- [x] `actionlint` against the generated install-test YAML — clean
- [ ] CI green on this branch
- [ ] First adopter install after release shows green install-test workflow before merging


generator/src/tend/cli.py
```diff
@@ -92,7 +102,7 @@ def init(config_path: Path | None, dry_run: bool) -> None:
         )
     outdir = Path(".github/workflows")
 
-    workflows = generate_all(cfg)
+    workflows = generate_all(cfg, with_install_test=with_install_test)
     if not workflows:
         click.echo("No workflows enabled in config.")
         return
```
tend-agent: This early return skips the new stale-file cleanup exactly when the generated set is empty. If a repo disables all workflows (or only has `ci-fix` without `watched_workflows`), `tend init` prints `No workflows enabled in config.` and leaves existing `tend-*.yaml` files in place, so the promised cleanup for disabled workflows is incomplete. I reproduced it with a temp repo containing an old `.github/workflows/tend-review.yaml` and all workflows disabled; the file remained after `tend init`. Let the cleanup loop run even for an empty `workflows` list (and add a regression for the all-disabled case).

```suggestion
```

tend-agent (commented):
I found one cleanup lifecycle issue.

tend-agent (approved):

PR #546, opened 2026-05-16
generator: switch build backend to uv_build

max-sixty:
Switch the generator to the uv-native PEP 517 backend. uv-build's defaults (`module-root = "src"`, module name from the normalized project name) match the existing layout, so no `[tool.uv.build-backend]` block is needed and the hatch-specific `[tool.hatch.build.targets.wheel]` goes away.

Verified `uv build --wheel` produces the same six `tend/*.py` modules, `pytest` is green (242 passed), and a sentinel `src/tend/_sentinel.txt` shipped in the wheel — so non-Python data files under `src/tend/` are bundled by default. That matters for the templating-evaluation branch (#545), which adds `src/tend/templates/*.yaml.j2`; the two PRs will conflict on `pyproject.toml` at merge time, whichever lands second rebases.


tend-agent (commented):
The build backend swap itself looks good: `uv build --directory generator --wheel` succeeds on the merge commit, and the wheel includes `tend/templates/*.yaml.j2`.

One consistency issue to fix before landing: `CLAUDE.md` still describes `generator/` as a hatchling build and says the generator uses hatchling under the “Key details” section. After this PR, that guidance will be false for future bot/dev sessions. Please update those references to `uv_build` while changing the backend.

tend-agent (approved):

PR #555, opened 2026-05-17
chore: regenerate workflows with tend 0.0.24

max-sixty:
Repins tend's own workflows to `max-sixty/tend/codex@0.0.24` so the dogfooded Codex runs pick up #551 (read skills in full) and #552 (stop re-posting "prior thread still applies"). Mechanical regen via `uvx tend@latest init`; only the action ref changed.


max-sixty:
Superseded by #556. #556 switched tend's own harness back to Claude and regenerated its workflows to `max-sixty/tend@0.0.24`, which makes this PR's repin to the Codex action a no-op (zero net diff vs main). The 0.0.24 release (#554 + tag) shipped #551/#552, and #556 already repinned tend's workflows to pick them up.

PR #565, opened 2026-05-19
skills(review): push judgment-call fixes on bot PRs by default

tend-agent:
## Problem

Step 8's bot-PR clause already permitted pushing fixes, but the phrasing ("concrete, fixable issues") left an implicit "mechanical only" bar that the bot enforced in practice. As soon as a fix needed a judgment call, the run fell back to scoping or offering — neither of which a bot author can action — and the PR sat red until a human redid the analysis.

## Solution

Rewrite the bot-PR paragraph to:

- Drop the implicit "mechanical only" gate: pushable iff the review can name the fix and its reasoning, even when judgment is needed.
- Tiebreaker when several reasonable approaches exist: pick the one most consistent with the repo's existing patterns, push, note the alternative in the review body.
- Bounded escape hatch: only scope-for-maintainer when no defensible default exists — real semantic ambiguity, not "more than zero lines of thought".

Also renames the section heading from "Push mechanical fixes" to "Push fixes", since the new policy explicitly covers non-mechanical fixes on bot PRs. Human-PR behaviour is unchanged.

## Testing

Skill text change only. `pre-commit run --files plugins/tend-ci-runner/skills/review/SKILL.md` passes.

---
Closes #564 — automated triage

PR #566, opened 2026-05-19
review: push fixes to bot PRs by default, even when the fix needs a judgment call

### LINKED ISSUE #564: review: push fixes to bot PRs by default, even when the fix needs a judgment call
## Summary

On bot-authored PRs (Dependabot etc.), `tend-ci-runner:review` step 8 already permits pushing fixes directly:

> **Bot PRs** (Dependabot, renovate, etc.): If the review found concrete, fixable issues and there's no human author to act on feedback, commit and push the fix directly to the PR branch.

In practice tend only pushes when the fix is purely mechanical. As soon as a fix needs a judgment call, it falls back to scoping the work or offering options and waits for a maintainer. On a bot PR there is no author to act on that offer, so the PR sits red indefinitely and a human has to do exactly the work tend already analysed.

## Two recent cases (PRQL/prql)

**[PRQL/prql#5921](https://github.com/PRQL/prql/pull/5921)** — Dependabot bump `sqlparser 0.60 → 0.62`. tend posted an excellent review: a table mapping all 9 compile errors to exact files/lines and the upstream AST changes. Five were mechanical defaults (`None`/`vec![]`); two needed a small judgment call (`ExcludeSelectItem::Multiple` wants `ObjectName`s; a new non-exhaustive `SelectItem::ExprWithAliases` variant). tend declined to push and scoped it for a maintainer. The two "judgment" calls had an obvious reasonable answer, which tend itself stated in the review.

**[PRQL/prql#5916](https://github.com/PRQL/prql/pull/5916)** — Dependabot bump `megalinter 9.4 → 9.5`. tend correctly diagnosed both newly-failing linters and proposed two consistent fixes, ending with "Happy to push either fix as a follow-up commit to this branch — let me know which you prefer." On a Dependabot PR there is nobody to answer that.

## Request

On bot-authored PRs, when tend has a concrete fix it can articulate, default to picking the most reasonable option and pushing it, not offering or scoping. Keep "offer and wait" as the human-PR behavior. Suggested refinements to step 8:

- Drop the implicit "mechanical only" bar for bot PRs. If the review can name the fix and its reasoning, it's pushable.
- When several reasonable approaches exist (e.g. #5916's two configs), pick the one most consistent with the repo's existing patterns, push it, and note the alternative in the review rather than blocking on a question.
- A bounded escape hatch: if the fix genuinely can't be decided without maintainer intent (a real semantic ambiguity, not just "more than zero lines of thought"), scope it. The bar should be "no defensible default exists," not "not purely mechanical."

The net effect: a bot PR that tend can fix gets fixed and goes green, instead of accumulating a thorough review that a human still has to action.

max-sixty:
Step 8 of the review skill already permitted pushing fixes to bot-authored PRs, but the "Push **mechanical** fixes" framing meant the bot only pushed pure find-replace changes. The moment a fix needed any thought — a multi-site dependency-bump API migration, or a choice between two valid config options — it fell back to describing the fix and waiting for a human. On a Dependabot PR nobody acts on that offer, so the PR sits red and a maintainer ends up doing the work the review had already worked out in full.

This reframes the deferral bar. "Not a one-token change" and "more than one syntactically valid form exists" are no longer reasons to defer. The bot picks the option most consistent with the surrounding code and the repo's existing conventions, pushes it, and notes alternatives in the review. It defers only when *no defensible default exists* — a genuine semantic ambiguity needing maintainer intent, not merely a fix that took thought to derive.

Section 6 (the CI-failed branch of Monitor CI) was silent on bot PRs, which is the exact path where this surfaced in practice — analysis posted, nothing pushed. It now points to step 8 rather than stopping at analysis. The user-facing `### review` description is synced to drop "mechanical" (both `tend.example.yaml` copies and the README workflow table).

Closes #564

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ file](docs/tend.example.yaml) and a repo-local `/running-tend` skill.
 
 | Workflow          | Trigger                    | What happens                                                                                                                                                |
 | ----------------- | -------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **review**        | PR opened/updated          | Reviews for correctness and duplication. Traces error paths. Monitors CI. Pushes mechanical fixes to bot-authored PRs.                                      |
+| **review**        | PR opened/updated          | Reviews for correctness and duplication. Traces error paths. Monitors CI. Pushes fixes to bot-authored PRs.                                                 |
 | **mention**       | @bot mention, review       | Responds to requests in PR and issue conversations.                                                                                                         |
 | **triage**        | Issue opened               | Classifies the issue, checks for duplicates, reproduces bugs, attempts conservative fixes.                                                                  |
 | **ci-fix**        | CI fails on default branch | Reads failure logs, identifies root cause, searches for the same pattern elsewhere, opens a fix PR.                                                         |
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -177,7 +177,7 @@ bot_name: my-project-bot
 # ### review
 #
 # Triggers on PR open/update. Reviews for correctness, duplication, error paths.
-# Monitors CI. Pushes mechanical fixes to bot-authored PRs.
+# Monitors CI. Pushes fixes to bot-authored PRs.
 #
 # workflows:
 #   review:
--- a/plugins/install-tend/skills/install-tend/references/tend.example.yaml
+++ b/plugins/install-tend/skills/install-tend/references/tend.example.yaml
@@ -177,7 +177,7 @@ bot_name: my-project-bot
 # ### review
 #
 # Triggers on PR open/update. Reviews for correctness, duplication, error paths.
-# Monitors CI. Pushes mechanical fixes to bot-authored PRs.
+# Monitors CI. Pushes fixes to bot-authored PRs.
 #
 # workflows:
 #   review:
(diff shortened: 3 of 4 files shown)


tend-agent (commented):
Two notes.

**Overlap with #565.** [#565](https://github.com/max-sixty/tend/pull/565) (automated triage) also closes #564 and touches three of the same files. The two diffs differ in coverage:

|  | this PR (#566) | #565 |
|---|---|---|
| `README.md` table row | unchanged | dropped "mechanical" |
| `docs/tend.example.yaml` | dropped "mechanical" | dropped "mechanical" |
| `plugins/install-tend/.../tend.example.yaml` | dropped "mechanical" | dropped "mechanical" |
| `SKILL.md` step 8 paragraph | rewritten (long form) | rewritten (shorter) |
| `SKILL.md` step 6 (CI-failed branch) | updated to point at step 8 for bot PRs | untouched |

#566 is the more thorough change (step 6 needed the update); #565 is the one that catches the README. Likely the right move is to land this one and close #565, but pulling the README hunk in first.

**Stale README reference.** [`README.md`](https://github.com/max-sixty/tend/blob/main/README.md) line 77 still reads "Pushes mechanical fixes to bot-authored PRs" — same wording you updated in the two `tend.example.yaml` files. Happy to push a one-line fix to this branch if you'd like.

tend-agent (approved):

PR #573, opened 2026-05-21
skills(running-in-ci): recheck PR state before pushing follow-up commits

### LINKED ISSUE #572: running-in-ci: bound polling loops and recheck PR state before pushing follow-up commits
## Problem

A `tend-mention` handle job deadlocked for 6 hours after pushing a follow-up commit to a PR that had been merged 8 minutes earlier. The bot polled the merged PR's `headRefOid` in an unbounded `until` loop in the background, holding the job alive until GitHub Actions cancelled it at the 360-minute default workflow timeout.

The chain has two contributing causes — both are generic tend behaviour, not adopter-specific.

## Evidence

- Run: [max-sixty/worktrunk#26199443899](https://github.com/max-sixty/worktrunk/actions/runs/26199443899) (`tend-mention`, handle job 01:12:33 → 07:12:49 UTC).
- PR: [max-sixty/worktrunk#2844](https://github.com/max-sixty/worktrunk/pull/2844) (`fix/hook-show-projects-merge`, squash-merged at 01:35:19 UTC).
- Session log: `221bd85b-2d91-4f27-b70f-e376ef27b10d.jsonl` (active 01:13:24 → 01:47:55, then idle in background poll until cancellation).
- Side effect: orphan commit `f14fd1db9ec3241f1f17ab5dbcaafa8c68e9a48c` left on `origin/fix/hook-show-projects-merge` (not in `main`).
- Side effect: the next `tend-mention-handle-2826` run queued behind for 5h53m (concurrency `cancel-in-progress: false`).
- Side effect: `token-usage.json` reported `0` for the cancelled run because the post-Claude step never ran — the LLM cost was non-zero but silently dropped from the per-run report.

## Decision-chain

1. Bot triages issue, opens PR, starts background CI poll.
2. **Maintainer squash-merges the PR** while CI is still going. Bot's poll completes, sees codecov gap, decides to refactor.
3. **Bot pushes the refactor commit to the PR branch** without checking the PR's current state. (Step where rule #1 below would have fired.)
4. Bot queries `gh pr view 2844 --json headRefOid` — returns the SHA at merge time, not the new push. Bot interprets the discrepancy as "API hasn't propagated yet" instead of "PR is merged and frozen."
5. Bot attempts `sleep 30 && gh pr view ...` to retry. Harness blocks this anti-polling pattern with:

   > `Blocked: sleep 30 followed by gh pr view ... To wait for a condition, use Monitor with an until-loop (e.g. until <check>; do sleep 2; done). To wait for a command you started, use run_in_background: true. Do not chain shorter sleeps to work around this block.`

6. Bot follows the suggestion literally as a background Bash command:

   ```bash
   until [ "$(gh pr view 2844 --json headRefOid --jq '.headRefOid')" = "f14fd1db9..." ]; do
     sleep 5
   done
   ```

   The condition will never be true. The session continues (launches a second `poll-ci-v2.sh` with the same `until` prologue), then exits with the background loops still running.

7. The `until` loop holds the GitHub Actions job alive for ~5h25m. GitHub cancels at the 360-min cap.

The harness's `until <check>; do sleep 2; done` suggestion is *correct shape* for short propagation waits, but it has no iteration cap, so any non-converging condition becomes a workflow-timeout deadlock.

### Rule 1: Re-check PR state before pushing a follow-up commit

Under the existing **Pushing to PR Branches** section, add:

> **Re-check the PR is still open before pushing follow-up commits.** Any wait that lets time pass — a CI poll, codecov-report fetch, sleep, background task — also gives a maintainer time to merge or close the PR. Before pushing a refactor / coverage / lint fixup on top of an existing PR, re-query the state:
>
> ```bash
> STATE=$(gh pr view <N> --json state --jq '.state')
> if [ "$STATE" != "OPEN" ]; then
>   echo "PR #<N> is $STATE — skipping follow-up push"
>   exit 0
> fi
> ```
>
> If the PR has merged, the work is superseded — comment on the PR or follow-up issue if the gap (e.g. coverage) is still real, but do not push to the now-orphan branch. After merge, `gh pr view <N> --json headRefOid` returns the SHA at merge time and never advances; polling it for a new push is a guaranteed deadlock.

### Rule 2: All polling loops bounded; no unbounded `until <check>`

Under **CI Monitoring** (or a new "Polling" subsection), add:

> **Every polling loop has an iteration cap.** Use `for i in $(seq 1 N); do ... done`, never an unbounded `until <check>; do sleep N; done`. Unbounded waits run via `run_in_background: true` survive the assistant exiting and hold the GitHub Actions job alive until the workflow timeout (default 360 min) cancels it.
>
> If the harness suggests `until <check>; do sleep 2; done` to work around a `sleep N && check` block, **add a counter**:
>
> ```bash
> # Wait up to ~5 min for a condition; bail out (not loop forever) if it never converges.
> for i in $(seq 1 60); do
>   <check> && break
>   sleep 5
> done
> ```
>
> `Monitor` with an `until` predicate is fine — the harness supplies the cap. The hazard is *Bash* `until` loops backgrounded via `run_in_background: true`.

## Why bundled, not adopter-local

Both contributing causes are intrinsic to tend's design: tend opens PRs, polls CI, sometimes pushes a follow-up commit, and runs on top of a harness whose anti-polling guard recommends an unbounded `until` pattern. Any consumer that uses squash-merge can hit the push-after-merge race; any consumer can hit the unbounded-`until` deadlock. There is no worktrunk-specific shape here.

## Related

- [tend#568](https://github.com/max-sixty/tend/issues/568) (closed) — *"running-in-ci dedup misses already-merged sibling PRs and closed-issue signal"*. A neighbouring failure mode (creating a dup PR for already-merged work) with a similar lesson: a tend session that pauses for any reason has to re-check upstream state when it resumes. This issue extends the same lesson from "dedup before creating" to "re-check before pushing follow-up commits" and adds the unbounded-loop hazard.

tend-agent:
## Problem

A `tend-mention` handle job in a consumer repo deadlocked for ~6 hours after pushing a follow-up commit to a PR that had been merged minutes earlier. The bot polled the merged PR's `headRefOid` in an unbounded background `until` loop, holding the GitHub Actions job alive until the workflow's 360-minute cap cancelled it. Side effects: an orphan commit left on the now-deleted branch, the next handle job queued for 5h53m behind the deadlocked one (`cancel-in-progress: false`), and `token-usage.json` reporting `0` for the cancelled run.

The push-after-merge cause is intrinsic to tend's design (push follow-ups on PRs) and not adopter-specific.

## Solution

One rule added to `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md`, under *Pushing to PR Branches*: **re-check PR state before pushing a follow-up commit.** Any wait — CI poll, coverage fetch, sleep, background task — lets a maintainer merge or close the PR. If state isn't `OPEN`, skip the push. After merge, `gh pr view --json headRefOid` returns the SHA at merge time forever, so polling it for a new push is a guaranteed deadlock.

The originally-proposed second rule ("bound every polling loop") was dropped during review — the *CI Monitoring* section already shows the bounded-`for` recipe, and a negative-form restating added little.

## Testing

Skill text only — no code paths to test.

---
Closes #572 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -247,20 +247,6 @@ gh api "repos/{owner}/{repo}/actions/runs?branch=main&status=completed&per_page=
 
 If you cannot verify, say "I haven't confirmed whether these failures are pre-existing."
 
-### Bound every polling loop
-
-Use `for i in $(seq 1 N); do ... done` — never an unbounded `until <check>; do sleep N; done` from Bash. A backgrounded `until` loop survives the agent exiting and holds the GitHub Actions job alive until the workflow timeout (default 360 min) cancels it. If the harness blocks `sleep N && check` and suggests `until <check>; do sleep 2; done`, add a counter:
-
-```bash
-# Wait up to ~5 min, then bail rather than loop forever.
-for i in $(seq 1 60); do
-  <check> && break
-  sleep 5
-done
-```
-
-`Monitor` with an `until` predicate is fine — the harness caps it. The hazard is Bash `until` loops launched via `run_in_background: true`.
-
 ### Polling `gh run rerun --failed`
 
 After `gh run rerun <run-id> --failed`, poll the rerun jobs directly. The parent run's `.status` stays `in_progress` until every sibling job finishes, including unrelated long-running ones, and the `pending()` recipe above also doesn't help — sibling check-runs on the head SHA still appear pending. Polling specific job IDs is the only fix.
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -134,6 +134,17 @@ Always use `git push` without specifying a remote — `gh pr checkout` configure
 
 If pushing fails (fork PR with edits disabled), fall back to posting code snippets in a comment. Don't reference commit SHAs from temporary branches — post code inline.
 
+### Re-check PR state before pushing a follow-up commit
+
+Any wait that lets time pass — a CI poll, coverage fetch, sleep, background task — also gives a maintainer time to merge or close the PR. After waiting:
+
+```bash
+STATE=$(gh pr view <N> --json state --jq '.state')
+[ "$STATE" = "OPEN" ] || { echo "PR #<N> is $STATE — skipping push"; exit 0; }
+```
+
+If the PR is merged, the work is superseded. Comment if a real gap remains; do not push to the now-orphan branch. After merge, `gh pr view <N> --json headRefOid` returns the SHA at merge time and never advances — polling it for a new push is a guaranteed deadlock.
+
 ## Merging Upstream into PR Branches
 
 When asked to merge the default branch into a PR branch:
@@ -235,6 +246,20 @@ gh api "repos/{owner}/{repo}/actions/runs?branch=main&status=completed&per_page=
 ```
 
 If you cannot verify, say "I haven't confirmed whether these failures are pre-existing."
+
+### Bound every polling loop
+
+Use `for i in $(seq 1 N); do ... done` — never an unbounded `until <check>; do sleep N; done` from Bash. A backgrounded `until` loop survives the agent exiting and holds the GitHub Actions job alive until the workflow timeout (default 360 min) cancels it. If the harness blocks `sleep N && check` and suggests `until <check>; do sleep 2; done`, add a counter:
+
+```bash
+# Wait up to ~5 min, then bail rather than loop forever.
+for i in $(seq 1 60); do
+  <check> && break
+  sleep 5
+done
+```
+
+`Monitor` with an `until` predicate is fine — the harness caps it. The hazard is Bash `until` loops launched via `run_in_background: true`.
 
 ### Polling `gh run rerun --failed`
 


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -236,6 +247,20 @@ gh api "repos/{owner}/{repo}/actions/runs?branch=main&status=completed&per_page=
 
 If you cannot verify, say "I haven't confirmed whether these failures are pre-existing."
 
+### Bound every polling loop
+
+Use `for i in $(seq 1 N); do ... done` — never an unbounded `until <check>; do sleep N; done` from Bash. A backgrounded `until` loop survives the agent exiting and holds the GitHub Actions job alive until the workflow timeout (default 360 min) cancels it. If the harness blocks `sleep N && check` and suggests `until <check>; do sleep 2; done`, add a counter:
```
max-sixty: do we have any contra-examples? I'm keen to avoid lots of "don't do bad things" comments

tend-agent: Honest answer: just the one incident behind #572, not a pattern. And the **CI Monitoring** section right above already shows the positive recipe (`for i in $(seq 1 15); do … done`), so this is largely restating in negative form what's already there.

Options:

1. **Drop the section.** The CI Monitoring recipe is the positive form; the new content adds little.
2. **Keep just the non-obvious bit as a one-liner appended to the CI Monitoring recipe:** that the harness's suggested `until <check>; do sleep 2; done` is safe in `Monitor` but a trap in backgrounded Bash. That's the part not derivable from the existing recipe.

I'd lean toward option 2 — the harness/`Monitor` distinction is the specific knowledge that prevented the deadlock, and a one-liner attached to existing positive guidance avoids a standalone "don't" section. Happy to go with 1 if you prefer fewer rules over more nuance.

(The first rule — re-check PR state before pushing — stays as is; that one's a positive recipe.)

max-sixty: drop the section

tend-agent: Dropped in 5361680. PR title/body updated to reflect the single remaining rule.

PR #583, opened 2026-05-22
fix: install-test default-branch probe on shallow checkout

### LINKED ISSUE #582: tend-install-test.yaml drift-check fails on default shallow `actions/checkout`: `Not a valid ref: refs/remotes/origin/main`
Installed tend 0.0.25 on a new repo (`diffplug/dormouse`) via `uvx tend@latest init --with-install-test`. The `tend-install-test` workflow on the install PR fails at the "Verify generator output matches committed files" step:

```
error: Not a valid ref: refs/remotes/origin/main
```

Run log: <https://github.com/diffplug/dormouse/actions/runs/26300228733/job/77423177283?pr=82>

## Root cause

The step starts with:

```bash
git remote set-head origin --auto
```

The comment above it explains the intent: `actions/checkout` doesn't set `origin/HEAD`, so we query the remote to find the default branch. But `set-head --auto` only sets the *local* symbolic-ref `refs/remotes/origin/HEAD` — it requires the target ref (`refs/remotes/origin/main`) to already exist locally. With default `actions/checkout@v6` settings, only the PR head ref is fetched, so `refs/remotes/origin/main` doesn't exist and the command errors before any default-branch detection happens.

The secrets-verification step passes; the failure is purely on this drift-check setup.

## Suggested fix

The workflow already has `GH_TOKEN: ${{ github.token }}` in scope, so the cleanest path is to skip git entirely and query the API:

```bash
DEFAULT_BRANCH=$(gh api "repos/$GITHUB_REPOSITORY" --jq .default_branch)
git fetch origin --depth 1 "+refs/heads/$DEFAULT_BRANCH:refs/remotes/origin/$DEFAULT_BRANCH"
git symbolic-ref "refs/remotes/origin/HEAD" "refs/remotes/origin/$DEFAULT_BRANCH"
```

Alternatives: bump the checkout step to `fetch-depth: 0`, or pre-fetch the default branch before `set-head --auto`.

## Repro

Any repo where the initial install PR is a feature branch (i.e., `actions/checkout@v6` only has the PR head fetched). The `diffplug/dormouse` repo above repros it on the install PR directly.

## Workaround for now

For new installs, run `uvx tend@latest init` *without* `--with-install-test` — that's what `max-sixty/worktrunk` does and there's no install-test on its main branch.

tend-agent:
## Problem

The `tend-install-test` workflow's drift-check step fails on every install PR with:

```
error: Not a valid ref: refs/remotes/origin/main
```

Root cause: `git remote set-head origin --auto` requires the target ref (e.g. `refs/remotes/origin/main`) to already exist locally. With default `actions/checkout@v6` settings, only the PR head ref is fetched — so the command errors before any default-branch detection happens.

## Solution

Resolve the default branch via the GitHub API, fetch it shallowly, then point `origin/HEAD` at it. The generator's `_detect_default_branch_local` (`git rev-parse --abbrev-ref origin/HEAD`) then sees the correct value, so a repo with a non-`main` default doesn't report spurious drift.

## Testing

- Added an assertion in `test_install_test_workflow_shape` that the workflow uses the API-based probe and not `git remote set-head` (which would fail on shallow checkouts).
- Regenerated the `test_install_test_workflow_regtest` snapshots for both Claude and Codex harnesses.
- Reproduced the original failure locally with a shallow single-branch clone — confirmed the new command sequence sets `origin/HEAD` correctly.
- Full `pytest` suite passes (242 tests); `pre-commit` clean.

---
Closes #582 — automated triage

### DIFF
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -429,10 +429,14 @@ jobs:
         env:
           GH_TOKEN: ${{{{ github.token }}}}
         run: |
-          # `actions/checkout` doesn't set `origin/HEAD`, so the generator's
-          # default-branch probe falls back to "main" and a repo with a
-          # non-`main` default reports spurious drift. Query the remote.
-          git remote set-head origin --auto
+          # `actions/checkout` only fetches the PR head ref, so the
+          # generator's default-branch probe (`git rev-parse --abbrev-ref
+          # origin/HEAD`) falls back to "main" and a repo with a non-`main`
+          # default reports spurious drift. Resolve the default branch via
+          # the API, fetch it, then point `origin/HEAD` at it.
+          DEFAULT_BRANCH=$(gh api "repos/$GITHUB_REPOSITORY" --jq .default_branch)
+          git fetch origin --depth 1 "+refs/heads/$DEFAULT_BRANCH:refs/remotes/origin/$DEFAULT_BRANCH"
+          git symbolic-ref "refs/remotes/origin/HEAD" "refs/remotes/origin/$DEFAULT_BRANCH"
           # Pin the regen to the version in the committed header so a tend
           # release between local `init` and this CI run doesn't fail the
           # drift check for an irrelevant reason.
--- a/generator/tests/_regtest_outputs/test_generate.test_install_test_workflow_regtest[claude].out
+++ b/generator/tests/_regtest_outputs/test_generate.test_install_test_workflow_regtest[claude].out
@@ -44,10 +44,14 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          # `actions/checkout` doesn't set `origin/HEAD`, so the generator's
-          # default-branch probe falls back to "main" and a repo with a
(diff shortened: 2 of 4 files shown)


max-sixty:
can we implement this by passing args to the checkout action?

tend-agent:
Not quite — `actions/checkout` has no input that sets `refs/remotes/origin/HEAD`. The closest knob is `fetch-depth: 0`, which fetches all branches (so `origin/<default>` exists), but it still doesn't run `git remote set-head`, so the generator's `git rev-parse --abbrev-ref origin/HEAD` would still fall back to `main`. We'd need a follow-up `git remote set-head origin --auto` either way, and `fetch-depth: 0` pulls the full history of every branch just to set one ref.

The current API + shallow-fetch approach grabs only the one ref it needs and doesn't depend on `set-head --auto` succeeding (which is what was failing originally). Happy to switch to `fetch-depth: 0` + `git remote set-head origin --auto` if you'd prefer the simpler-looking workflow at the cost of a heavier fetch.

Issue #598, opened 2026-05-25
Rotate CLAUDE_CODE_OAUTH_TOKEN and bot PAT — leaked into run 26358181381 session log


tend-agent:
## Summary

The session-log artifact for [run 26358181381](https://github.com/max-sixty/tend/actions/runs/26358181381) (tend-weekly, 2026-05-24) contains the live `CLAUDE_CODE_OAUTH_TOKEN` and the bot's GitHub PAT in cleartext. These should be treated as compromised and rotated.

The artifact is downloadable by anyone with workflow read access on this repo. It expires on the normal artifact-retention schedule, but the tokens are still valid until rotated.

## How it happened

During §1 troubleshooting, the agent ran a name-only env inspection:

```bash
env | cut -d= -f1 | sort | grep -iE 'claude|anthropic|oauth|token'
```

Intent: list var names matching auth keywords. Actual output: the values of both tokens.

`claude-code-action` injects `ALL_INPUTS` into the agent process env as a multi-line JSON blob — the same `inputs.*` block that's documented in `action.yaml` as a "private contract" at lines 230–233. The blob carries `claude_code_oauth_token` and `github_token` as plaintext fields. `env` emits the value verbatim including embedded newlines, so each continuation line of the JSON becomes its own `env` output line; on those lines the secret string sits *before* the first `=` (often with no `=` at all), defeating the `cut -d= -f1` filter. The downstream `grep` then matches the lines containing `claude_code_oauth_token` / `github_token` and prints them.

`action.yaml:229` pins `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: "0"` to work around [anthropics/claude-code#35701](https://github.com/anthropics/claude-code/issues/35701), so the agent's bash subprocess inherits `ALL_INPUTS` directly — no upstream scrub catches this path.

## Recommended actions

1. **Rotate**, in this order:
   - The `CLAUDE_CODE_OAUTH_TOKEN` repo secret. Generate a new token via `claude setup-token` from a maintainer account and update the secret.
   - The `tend-agent` bot's classic GitHub PAT used as `BOT_TOKEN` (or equivalent — whichever secret feeds `inputs.github_token` in the workflows). Delete the existing PAT from the bot account's settings, mint a fresh one with the same scopes per `docs/security-model.md`, and update the secret.

[...]

PR #599, opened 2026-05-25
skills(running-in-ci): close the env-filter loophole for ALL_INPUTS secrets

tend-agent:
## Summary

Tighten the `Secrets` rule in `running-in-ci` to head off a specific leak path that triggered during the [2026-05-24 weekly run](https://github.com/max-sixty/tend/actions/runs/26358181381) on this repo. The current rule lists `env`/`printenv` as forbidden but is silent on names-only filters like `env | cut -d= -f1`, which the agent treats as a safe escape hatch when investigating the workflow environment.

## What happened

During the weekly integration-test recipe (§1 troubleshooting), the agent ran:

```bash
env | cut -d= -f1 | sort | grep -iE 'claude|anthropic|oauth|token'
```

The intent was "names only, no values." The actual output included the `CLAUDE_CODE_OAUTH_TOKEN` and `github_token` values in cleartext — they landed in the session-log artifact for that run. (Filed separately as #598 with rotation guidance.)

## Why the filter doesn't work

`claude-code-action` exposes `ALL_INPUTS` to the agent process as a multi-line JSON blob:

```
ALL_INPUTS={
  "claude_code_oauth_token": "sk-ant-oat01-…",
  "github_token": "ghp_…",
  …
}
```

`env` emits the value verbatim, including its embedded newlines. From `cut`'s perspective, each continuation line of the JSON looks like its own variable: the secret string is *before* the first `=` on that line (often there's no `=` at all), so `cut -d= -f1` keeps the entire line. The downstream `grep -iE 'claude|...|token'` then matches the credential lines exactly and prints them.

`action.yaml:229` pins `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: "0"` to dodge a different Linux bash bug ([anthropics/claude-code#35701](https://github.com/anthropics/claude-code/issues/35701)), so the agent's subprocess inherits `ALL_INPUTS` directly — there is no upstream scrub catching this.

## Change

One paragraph in `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md`:

- Names the structural leak path (multi-line JSON in `ALL_INPUTS`)
- Calls out the specific filter pattern that *looks* safe but isn't
- Keeps the rule absolute: no introspection of the process env, period

## Classification

- **Structural** failure: the env layout is deterministic across runs; the same filter leaks the same secrets every time.
- **Gate 1 — Confidence**: Critical (real production tokens emitted into an artifact). 1 occurrence is the bar.
- **Gate 2 — Magnitude**: targeted one-paragraph rule clarification; no new section or workflow change.

## Test plan

- [ ] Maintainer review — the rule reads as universally applicable to every consumer, not tend-specific.
- [ ] Pre-commit clean.


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -73,7 +73,7 @@ If a linked PR merged (or the triggering PR itself merged) **after the triggerin
 
 ## Restrictions
 
-- **Secrets**: Never run commands that expose secrets (`env`, `printenv`, `set`, `export`, `cat`/`echo` on credential files). Never include tokens or credentials in responses or comments.
+- **Secrets**: Never run commands that introspect the process env (`env`, `printenv`, `set`, `export`) or `cat`/`echo` credential files. The rule is absolute — names-only filters like `env | cut -d= -f1` or `env | grep -E '^[A-Z]+='` do not make these commands safe, because `claude-code-action` injects `ALL_INPUTS` as a multi-line JSON env value embedding `claude_code_oauth_token` and `github_token` in cleartext. The JSON's newlines turn each token-bearing line into its own `env` output line, where the value sits *before* the `=` and a name-stripping filter leaks it verbatim into the session log. Never include tokens or credentials in responses or comments.
```
tend-agent: The `env | grep -E '^[A-Z]+='` example doesn't fit the explanation that follows. The regex requires names of only uppercase letters — `ALL_INPUTS` (with an underscore) doesn't match it, so the multi-line-JSON leak path described in the bullet wouldn't fire on this filter at all. And the grep keeps full `NAME=VALUE` lines for any matching variable — it isn't a name-stripping filter, so the surrounding "value sits before the `=`" reasoning doesn't apply either. The single `cut -d= -f1` example is enough to anchor the rule; suggesting we drop the second one (or replace it with another genuine name-stripper like `env | sed 's/=.*//'` that fails for the same reason as `cut`).

```suggestion
- **Secrets**: Never run commands that introspect the process env (`env`, `printenv`, `set`, `export`) or `cat`/`echo` credential files. The rule is absolute — name-stripping filters like `env | cut -d= -f1` do not make these commands safe, because `claude-code-action` injects `ALL_INPUTS` as a multi-line JSON env value embedding `claude_code_oauth_token` and `github_token` in cleartext. The JSON's newlines turn each token-bearing line into its own `env` output line, where the value sits *before* the `=` and a name-stripping filter leaks it verbatim into the session log. Never include tokens or credentials in responses or comments.
```

tend-agent (commented):
The rule itself reads well — making the prohibition absolute and re-grounding "expose secrets" → "introspect the process env" closes the door on the workaround. One concern on the second example, inline.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Agreed — the `^[A-Z]+=` regex skips `ALL_INPUTS` outright (the underscore breaks `[A-Z]+`), and grep keeps full `NAME=VALUE` lines so it isn't a name-stripper anyway. Dropped the second example in f4b6ae3.

PR #608, opened 2026-05-26
fix(mention): skip no-op sessions for undirected bot comments

### LINKED ISSUE #606: tend-mention: skip no-op sessions for undirected bot comments on bot-authored PRs
## Problem

`tend-mention` runs a full agent session for `issue_comment` events on a bot-authored PR even when the comment is an automated notification from another bot that contains no mention of the tend bot. The session always exits silently (correctly — there's nothing to act on), so the work is pure waste. It also fires twice per notification because deploy bots edit their comment and the workflow listens on `issue_comment: [created, edited]`.

## Root cause

The `verify` job short-circuits to `should_run=true` for **any** comment on a bot-authored PR — it checks PR authorship but never inspects the comment author. The `@<bot>`-mention check already runs first, so a legitimate human/bot summon is preserved, but an undirected bot notification falls through to the bot-authored-PR branch and starts a session anyway.

## Evidence (diffplug/dormouse)

PR [diffplug/dormouse#93](https://github.com/diffplug/dormouse/pull/93) was authored by `dormouse-bot`. A `cloudflare-workers-and-pages[bot]` deploy notification ([comment 4527716896](https://github.com/diffplug/dormouse/pull/93#issuecomment-4527716896), author type `Bot`, no `@dormouse-bot` mention) spun up a full agent session. Because that bot **edits** the comment to update deploy status, both the `created` and `edited` events fired a session:

| Run | Trigger | Comment timestamp | Outcome | Cost |
|-----|---------|-------------------|---------|------|
| [26355149816](https://github.com/diffplug/dormouse/actions/runs/26355149816) | `created` | created 07:27:53 | exited silently | $0.22 |
| [26355164490](https://github.com/diffplug/dormouse/actions/runs/26355164490) | `edited`  | updated 07:28:45 | exited silently | $0.23 |

Both sessions behaved correctly — they identified the comment as an undirected deploy notification and exited without posting. The session simply never needed to start. This recurs on every nightly regen PR, since each gets a created+edited Cloudflare deploy comment.

## Proposed fix

In the `verify` job, before the `PR_AUTHOR == "$BOT_NAME"` short-circuit, skip `issue_comment` events whose comment author is a `Bot` (`github.event.comment.user.type == 'Bot'`) and whose body does not contain the `@<bot>` mention. The existing mention check already runs first, so legitimate bot summons are preserved; this only suppresses undirected bot notifications.

## Context

Surfaced by a `review-runs` sweep on [diffplug/dormouse](https://github.com/diffplug/dormouse) ([dormouse#95](https://github.com/diffplug/dormouse/issues/95)). Filed upstream at maintainer request because `tend-mention.yaml` is an upstream-owned generated file ("Do not edit this file directly") with no clean local override. Companion issue covers a related no-op on `tend-review`.

tend-agent:
## Problem

`tend-mention`'s `verify` job short-circuits to `should_run=true` for any `issue_comment` event on a bot-authored PR, without inspecting the comment author. Undirected bot notifications (deploy status from `cloudflare-workers-and-pages[bot]`, CI bots, etc.) then spin up a full agent session that always exits silently — pure waste. Because deploy bots edit their comment to update status and the workflow listens on `issue_comment: [created, edited]`, each notification fires *two* no-op sessions.

## Solution

Add a guard in the `verify` script between the existing `@<bot>` mention check and the PR-author short-circuit:

```bash
if [ "$EVENT_NAME" = "issue_comment" ] && [ "$COMMENT_AUTHOR_TYPE" = "Bot" ]; then
  exit 0
fi
```

`COMMENT_AUTHOR_TYPE` is wired from `github.event.comment.user.type` (the public-API-visible Bot/User discriminator).

Ordering matters: the @-mention check runs first, so legitimate summons (e.g. another bot writing `@<our-bot>`) are still honored; the new guard runs before the PR-author short-circuit, which is what was producing the no-op sessions.

Tend's own bot accounts are PAT-based user accounts (`user.type == "User"`), so this does not affect tend-on-tend behavior — only true GitHub App / Bot accounts.

## Testing

- Added `test_mention_verify_skips_bot_comments_without_mention` asserting the guard is generated, `COMMENT_AUTHOR_TYPE` is wired, and the guard sits between the mention check and the PR-author short-circuit.
- Regenerated three regression snapshots (`test_workflow_minimal_regtest[mention]`, `test_workflow_with_setup_regtest[mention]`, `test_workflow_minimal_codex_regtest[mention]`).
- Full generator suite: 243 passed.

---
Closes #606 — automated triage

### DIFF
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -82,6 +82,15 @@ jobs:
             exit 0
           fi
 
+          # Undirected bot comments (deploy notifications, CI status) on a
+          # bot-authored PR otherwise fall through to the PR-author short-circuit
+          # below and spin up a no-op session — twice per notification when the
+          # source bot edits its comment. Skip them here.
+          if [ "$EVENT_NAME" = "issue_comment" ] && [ "$COMMENT_AUTHOR_TYPE" = "Bot" ]; then
+            echo "should_run=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
           # pull_request_review payloads include review.body (checked above)
           # but NOT the bodies of inline comments attached to the review.
           # Fetch them so a first-contact @-mention inside an inline comment
@@ -140,6 +149,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.<<cfg.bot_token_secret>> }}
           EVENT_NAME: ${{ github.event_name }}
           COMMENT_BODY: ${{ github.event.comment.body || github.event.review.body }}
+          COMMENT_AUTHOR_TYPE: ${{ github.event.comment.user.type }}
           ISSUE_BODY: ${{ github.event.issue.body }}
           ISSUE_OR_PR_NUMBER: ${{ github.event.issue.number }}
           ISSUE_AUTHOR: ${{ github.event.issue.user.login }}
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -86,6 +86,10 @@ jobs:
           # bot-authored PR otherwise fall through to the PR-author short-circuit
           # below and spin up a no-op session — twice per notification when the
           # source bot edits its comment. Skip them here.
+          # TODO: reassess — a quick pass of a small/fast model could replace
+          # this blanket skip with content-aware filtering (only drop comments
+          # that have no actionable signal), so we don't miss bot comments that
+          # actually warrant a response.
           if [ "$EVENT_NAME" = "issue_comment" ] && [ "$COMMENT_AUTHOR_TYPE" = "Bot" ]; then
             echo "should_run=false" >> "$GITHUB_OUTPUT"
             exit 0
--- a/generator/tests/_regtest_outputs/test_generate.test_workflow_minimal_codex_regtest[mention].out
+++ b/generator/tests/_regtest_outputs/test_generate.test_workflow_minimal_codex_regtest[mention].out
@@ -71,6 +71,10 @@ jobs:
           # bot-authored PR otherwise fall through to the PR-author short-circuit
           # below and spin up a no-op session ? twice per notification when the
           # source bot edits its comment. Skip them here.
+          # TODO: reassess ? a quick pass of a small/fast model could replace
+          # this blanket skip with content-aware filtering (only drop comments
+          # that have no actionable signal), so we don't miss bot comments that
+          # actually warrant a response.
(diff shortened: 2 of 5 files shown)


max-sixty:
but don't we want to know about some bot comments?

what are some cases where this is really noisy?

tend-agent:
Good question — the guard is narrower than it might look. We still react to bot comments in three cases:

1. **Any bot that `@`-mentions us** — the mention check at [`mention.yaml.j2:80`](https://github.com/max-sixty/tend/blob/2677538b19e39c5137e3c271dbdb7b33bd66ab30/generator/src/tend/templates/mention.yaml.j2#L80) runs *before* the new guard, so a CI bot or human-impersonating bot writing `@tend-agent` is still honored.
2. **Review events from bots** — the guard scopes to `EVENT_NAME = issue_comment`, so `pull_request_review` / `pull_request_review_comment` from a bot reviewer (CodeRabbit, Greptile, etc.) still go through.
3. **`issues` events authored by a bot** — also unaffected; the early-exit at line 75 fires first.

What we're dropping is the residual case: a bot posting a conversational `issue_comment` on a bot-authored PR without naming us. Those sessions always exited silently because there's nothing for the agent to do.

**Where it's actually noisy** — the trigger was the Cloudflare Pages deploy bot on bot-authored regen PRs, called out in #606 with evidence from [diffplug/dormouse#93](https://github.com/diffplug/dormouse/pull/93):

| Source bot | Pattern | Sessions per PR |
|---|---|---|
| `cloudflare-workers-and-pages[bot]` | Posts deploy status, then *edits* the same comment as the build progresses | 2 per push (created + edited), recurring on every nightly regen PR |
| Vercel / Netlify preview bots | Same shape (post + edit) | Same |
| Renovate / Dependabot status comments on their own PRs | One per status change | 1+ per PR |

On dormouse the Cloudflare case alone was ~$0.45 per regen PR for sessions that did nothing. Any repo with a deploy-preview bot and bot-authored PRs (nightly regens, dependency bumps) hits the same multiplier.

Happy to narrow it further if you'd prefer — e.g. limit the guard to bot-authored PRs only, so undirected bot comments on *human*-authored PRs would still spin up a session (currently they'd also hit the engagement-history branch and usually run, but we'd be preserving that path explicitly).

max-sixty:
ok, let's proceed but add a TODO that we should reassess in the future. 

in the future, we could consider quick pass of a fast model to avoid kicking off a 1T param model to see it was an inane bot comment

tend-agent:
Done — added a TODO above the guard in [`mention.yaml.j2`](https://github.com/max-sixty/tend/blob/80ab024d/generator/src/tend/templates/mention.yaml.j2#L85-L92) noting that a small/fast-model pass could replace the blanket skip with content-aware filtering (only drop comments with no actionable signal).

PR #609, opened 2026-05-27
interactive harness: PTY-supervised claude as Agent-SDK alternative

max-sixty:
## Why

Anthropic's June 15 2026 billing split puts `@anthropic-ai/claude-agent-sdk` on a metered Agent-SDK credit pool. `anthropics/claude-code-action` (which tend wraps today) calls `query()` from that SDK under the hood — so today's `harness: claude` path is exactly what's being metered. Interactive Claude Code in a terminal stays on the flat Pro/Max subscription. This PR adds a `harness: claude-interactive` path that runs the official `claude` binary in a PTY supervisor instead. Nothing changes for `harness: claude` users (or for tend's own workflows) until someone flips the config.

## What's added

**`interactive/action.yaml`** — new composite action, sibling of `codex/action.yaml`. Mirrors `action.yaml`'s inputs/outputs for drop-in swappability. Runs `claude --permission-mode bypassPermissions "$PROMPT"` inside `script(1)` for PTY allocation. A Stop hook in `.claude/settings.local.json` writes a sentinel file when the model finishes a turn; a `StopFailure` hook writes a separate sentinel for rate-limit / auth / server errors. The supervisor polls both, plus process exit and a timeout, then terminates cleanly with SIGTERM/SIGKILL. Pre-seeds workspace trust in `~/.claude.json` so interactive `claude` doesn't block on the first-run dialog (the part `claude -p` skips automatically). Plugins install via `claude plugin install` against this action's own checkout.

**Generator wiring** (`config.py`, `templates/macros.yaml.j2`, `tests/test_generate.py`) — `claude-interactive` joins `KNOWN_HARNESSES` as a sibling of `claude`. The `agent_step` macro grows a three-way branch for the action ref (`max-sixty/tend@`, `max-sixty/tend/interactive@`, `max-sixty/tend/codex@`). Claude-shaped auth secrets and `/tend-ci-runner:NAME` slash syntax are preserved via a new `CLAUDE_FAMILY_HARNESSES = {"claude", "claude-interactive"}` constant. Four new tests cover the routing.

**`.github/workflows/interactive-smoke.yaml`** — dispatch-only smoke test that exercises the new action via `uses: ./interactive` against a trivial Bash-tool prompt. Validates the parts the local spike could not cover: `claude` install on a fresh Ubuntu runner, plugin install non-interactively with OAuth env auth, Stop hook firing on Linux. Fork-guarded; defaults to `sonnet` for cheap iteration.

## How it detects turn end

Verified in a local spike (`Stop` fires once per turn in interactive mode, even with tool use, and the process stays alive afterward). Five-line state machine: success sentinel → exit 0; failure sentinel → exit 1 with `::error::`; premature process exit → exit with the child's code; timeout → exit 1.

## What's deferred

- **`allowed_bots` / `allowed_non_write_users` enforcement** — was claude-code-action's JS layer; primary gating already happens in generated workflow `if:` conditions. TODO in the action.
- **Per-workflow harness override** — top-level `harness:` only for now. Flipping `harness: claude-interactive` switches everything; a per-workflow knob would let nightly trial in isolation if that's wanted.
- **Snapshot tests** for every `claude-interactive` workflow YAML — would mint a dozen new `.out` files; the four routing tests cover the surface that matters.
- **Cutover** — `harness: claude` continues to be tend's own setting and the default. This PR adds the path; flipping it is a follow-up.

## Test plan

- [x] 246 generator tests pass (242 existing snapshot tests untouched + 4 new)
- [x] `pre-commit run` passes on all modified files
- [x] Local spike confirms `Stop` hook semantics under `script(1)`
- [ ] Smoke workflow run on `claude-p` post-merge confirms end-to-end on a real GHA runner


tend-agent (commented):
Two places outside this diff still gate on `cfg.harness == "claude"` rather than membership in `CLAUDE_FAMILY_HARNESSES`. With `harness: claude-interactive`, both fall into the `else` (Codex) branch and emit Codex-shaped behavior — breaking the "input-shape parity with `action.yaml`" claim once an adopter flips the config (or runs `tend init --with-install-test` / `tend check`).

**`generator/src/tend/workflows.py` — `generate_install_test`**

```python
if cfg.harness == "claude":
    harness_envs = (
        f"          CLAUDE_OAUTH: ${{{{ secrets.{cfg.claude_token_secret} }}}}\n"
        ...
    )
    harness_check = '"$CLAUDE_OAUTH$ANTHROPIC_KEY"'
else:
    harness_envs = (
        f"          CODEX_AUTH: ${{{{ secrets.{cfg.codex_auth_json_secret} }}}}\n"
        ...
    )
    harness_check = '"$CODEX_AUTH$OPENAI_KEY"'
```

With `harness: claude-interactive`, the generated `tend-install-test.yaml` checks for `CODEX_AUTH_JSON`/`OPENAI_API_KEY` secrets that the adopter doesn't have, so the install-test workflow fails noisily on a correctly-configured repo.

**`generator/src/tend/checks.py` — `run_all_checks`**

```python
engine_auth_secrets = (
    [cfg.claude_token_secret, cfg.anthropic_api_key_secret]
    if cfg.harness == "claude"
    else [cfg.openai_key_secret, cfg.codex_auth_json_secret]
)
...
if cfg.harness == "claude":
    results.append(check_claude_auth(repo, cfg))
else:
    results.append(check_codex_auth(repo, cfg))
```

`uvx tend check` against a `harness: claude-interactive` repo runs `check_codex_auth`, complaining about absent Codex secrets and reporting Claude OAuth as "extra/unallowed" via `check_repo_secret_allowlist`.

Fix is mechanical — swap each `cfg.harness == "claude"` for `cfg.harness in CLAUDE_FAMILY_HARNESSES` (already imported / defined in `config.py`). The install-test snapshot test would also catch this if parameterized over `["claude", "claude-interactive", "codex"]` instead of `["claude", "codex"]` (`test_install_test_workflow_regtest` in `generator/tests/test_generate.py`).

Happy to push a fix commit if you'd like.

---

Other notes (not blocking, flagging for awareness):

- The action's `Pin CLAUDE.md to base branch (fork PRs)` step pins only `CLAUDE.md`. `claude-code-action` additionally restores `.claude/`, `.mcp.json`, `.claude.json`, `.gitmodules`, `.ripgreprc` from base on fork PRs — the comment in this action calls this out as a known gap. Worth flagging in `docs/security-model.md` before any adopter flips a `pull_request_target` workflow to `claude-interactive`, given the RCE surface in `.claude/`.
- `claude-interactive-session-logs*` is a new artifact-name shape. `plugins/install-tend/skills/debug-tend-run/SKILL.md` currently enumerates only `claude-session-logs*` and `codex-session-logs-*` — fine to defer until the cutover, but worth a follow-up.

max-sixty:
Fixed in 4b92c13. Both gates now use `cfg.harness in CLAUDE_FAMILY_HARNESSES`:

- `generate_install_test` → install-test workflow now references `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_API_KEY` under `claude-interactive` (snapshot is byte-identical to the `claude` one).
- `run_all_checks` → emits `claude-auth`, not `codex-auth`, and the secret allowlist no longer flags the Claude OAuth token as extra.

Test coverage:
- `test_install_test_workflow_regtest` parameterized to include `claude-interactive`.
- New `test_claude_interactive_runs_claude_auth_check` in `test_checks.py` asserts the check-routing + allowlist behavior.

Also added a `docs/security-model.md` paragraph noting the interactive harness does not yet restore `.claude/` / `.mcp.json` / etc. on fork PRs — adopters should not flip `pull_request_target` workflows to `claude-interactive` until parity is restored. Tracked as follow-up after the trial.

Holding off on the `debug-tend-run` skill update (`claude-interactive-session-logs*` artifact name) until we know the trial actually proceeds — happy to bundle later if you'd like.

> _This was written by Claude Code on behalf of Maximilian._

tend-agent:
Confirmed the fix lands cleanly:

- `workflows.py:378` and `checks.py:452,467` both gate on `cfg.harness in CLAUDE_FAMILY_HARNESSES`.
- `test_install_test_workflow_regtest` is parameterized over `claude-interactive` and the new snapshot is byte-identical to the `claude` one.
- `test_claude_interactive_runs_claude_auth_check` asserts the `claude-auth` routing and a clean secret allowlist.
- `docs/security-model.md` calls out the `.claude/` / `.mcp.json` restoration gap and the `pull_request_target` caveat.

Agreed on deferring the `debug-tend-run` skill update — adding a `claude-interactive-session-logs*` enumeration before any consumer flips the harness would be speculative content in a skill that's loaded on every relevant session. Better to land it alongside the cutover when a real artifact of that shape exists to debug.

tend-agent (approved):

PR #612, opened 2026-05-27
generator: per-workflow harness override

max-sixty:
## Why

Lets adopters trial a new harness on a single workflow without flipping the whole bot. Concrete use case: try `harness: claude-interactive` on `nightly` while keeping PR review / mention on the production `claude` harness. Trial-safe rollout for the interactive harness added in #609 / #611.

## What

- `WorkflowConfig` gains an optional `harness` field. `Config.load` validates the value is a known harness and (cross-family only) compatible with the top-level model.
  - Same-family overrides (claude ↔ claude-interactive) share the model set, no compatibility error.
  - Cross-family overrides (codex → claude/claude-interactive with a codex model name like `gpt-5.5`, or vice versa with a per-workflow model unset) fail at config load with a clear message: "workflows.X.harness 'claude' is incompatible with the top-level model 'gpt-5.5'… Cross-family per-workflow harness overrides require setting `model:` per-workflow too, or changing the top-level model."
- `workflows.py`: a new `_effective_cfg(cfg, wf)` helper returns `cfg` unchanged when there's no override, or `dataclasses.replace(cfg, harness=wf.harness)` otherwise. Each generator function (review, mention, triage, ci-fix, scheduled, notifications) calls it once and passes the effective config to both the template and `default_prompt`. No template changes — `agent_step` still reads `cfg.harness` from whatever `cfg` it gets.
- Tests: per-workflow override targets only the named workflow (sibling workflows keep the top-level harness), unknown harness rejected, cross-family model mismatch rejected, same-family override loads cleanly.

## Example

```yaml
bot_name: tend-agent
harness: claude
# Trial the interactive harness on nightly only — review/mention/triage
# stay on the SDK-based claude harness until we're confident.
workflows:
  nightly:
    harness: claude-interactive
```

## Test plan

- [x] 252 tests pass (4 new); existing snapshots untouched
- [x] `pre-commit run --all-files` passes

### DIFF
--- a/generator/src/tend/config.py
+++ b/generator/src/tend/config.py
@@ -102,9 +102,11 @@ class WorkflowConfig:
     # Per-workflow harness override. Lets adopters trial a new harness on
     # a single workflow (e.g. `claude-interactive` on nightly only) before
     # flipping the whole bot. None means inherit from top-level `harness`.
-    # Cross-family overrides (claude → codex or vice versa) require also
-    # overriding the model, since model defaults are per-harness.
     harness: str | None = None
+    # Per-workflow model override. Pairs with `harness` for cross-family
+    # overrides — top-level `model` may not be valid for the new harness.
+    # None means inherit from top-level `model`.
+    model: str | None = None
 
 
 # Claude model allowlist — the set is small and stable enough that a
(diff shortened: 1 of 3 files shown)


generator/src/tend/config.py
```diff
@@ -300,6 +306,25 @@ def load(cls, path: Path | None = None) -> Config:
...
+                if wf_harness is not None:
+                    if wf_harness not in KNOWN_HARNESSES:
+                        raise click.ClickException(
+                            f"workflows.{name}.harness '{wf_harness}' is not recognized "
+                            f"(known: {', '.join(sorted(KNOWN_HARNESSES))})"
+                        )
+                    # Cross-family overrides need a compatible model. Same-family
+                    # overrides (claude ↔ claude-interactive) share the model set.
+                    eff_known = KNOWN_MODELS_BY_HARNESS.get(wf_harness)
+                    if eff_known is not None and model not in eff_known:
```
tend-agent: Asymmetric: this check only fires when the target harness has a model allowlist. The `claude → codex` direction passes through because `KNOWN_MODELS_BY_HARNESS.get("codex")` is `None`, so a config like top-level `harness: claude, model: opus` + `workflows.nightly.harness: codex` loads cleanly and the codex action gets `model: opus` (verified by rendering).

Also, the error message tells users to set `model:` per-workflow, but `WorkflowConfig` has no `model:` field — that advice is non-actionable.

If you don't want to add per-workflow `model:` yet, a symmetric family check would cover both directions and match the actual user remedy:

```python
top_claude = harness in CLAUDE_FAMILY_HARNESSES
wf_claude = wf_harness in CLAUDE_FAMILY_HARNESSES
if top_claude != wf_claude:
    raise click.ClickException(
        f"workflows.{name}.harness '{wf_harness}' is cross-family with "
        f"the top-level harness '{harness}'; the top-level model "
        f"'{model}' is for a different family. Use the same family "
        f"(claude ↔ claude-interactive), or change the top-level harness."
    )
```

tend-agent (commented):
One correctness gap in the cross-family validation: the check fires only when the *target* harness has a model allowlist, so the `claude → codex` direction slips through. With top-level `harness: claude, model: opus` and `workflows.nightly.harness: codex`, `Config.load` accepts the config and the renderer emits `model: opus` on the codex action — which codex won't accept. The PR description says cross-family overrides fail "vice versa with a per-workflow model unset"; that direction currently doesn't fail.

Related: the error message tells the user to set `model:` per-workflow, but `WorkflowConfig` doesn't have a `model:` field, so following the suggestion has no effect.

Pick whichever fix matches your intent — close the asymmetry by family (option in the inline comment), or add a per-workflow `model:` field and validate the effective pair.

max-sixty:
Both gaps fixed in 6d42cf4.

1. **Symmetric validation**: added a second guard for the cross-family-source direction. With `harness: claude` top-level and `workflows.nightly.harness: codex`, `Config.load` now errors with: "workflows.nightly crosses harness families (claude → codex) without a model override. Set `workflows.nightly.model:` to a valid codex model."

2. **Per-workflow `model:` field**: `WorkflowConfig` now has it, and `_effective_cfg` applies it. The error message now points users to a field that exists.

Test coverage:
- `test_per_workflow_harness_incompatible_model_rejected_codex_target` — the original direction (codex source → claude target, gpt-5.5 not in claude allowlist).
- `test_per_workflow_harness_incompatible_model_rejected_codex_source` — the previously-missed direction (claude source → codex target, opus → codex without model override).
- `test_per_workflow_model_override_unblocks_cross_family` — confirms the new `model:` field unblocks cross-family overrides; renders `model: gpt-5.5` on the codex step while sibling workflows keep `model: opus`.

254 tests pass (+2 over the previous push).

> _This was written by Claude Code on behalf of Maximilian._

tend-agent:
Both fixes look right — symmetric source-family guard fires before `_effective_cfg` ever sees an unsupported pair, and the new `WorkflowConfig.model` field makes the error message actionable. The three new tests cover both directions and the unblock path. Resolved on my end.

generator/src/tend/config.py
```diff
@@ -300,6 +308,56 @@ def load(cls, path: Path | None = None) -> Config:
...
+                # Same-family overrides (claude ↔ claude-interactive) share
+                # the model set, no error.
+                eff_harness = wf_harness or harness
+                eff_model = wf_model or model
+                eff_known = KNOWN_MODELS_BY_HARNESS.get(eff_harness)
+                if (
+                    wf_harness is not None
+                    and eff_known is not None
+                    and eff_model not in eff_known
+                ):
```
tend-agent: Gate also misses the model-only override case. `workflows.nightly.model: opus-99` (no `harness:`) sets `wf_harness = None`, so this check is skipped, the second one is skipped, and the renderer emits `model: opus-99` into a claude action step that Claude would reject at runtime.

```suggestion
                if (
                    (wf_harness is not None or wf_model is not None)
                    and eff_known is not None
                    and eff_model not in eff_known
                ):
```

The `Cross-family per-workflow harness overrides need...` phrasing in the error message becomes slightly off when this fires for a same-harness model typo — worth a small reword if you take the suggestion.

tend-agent (commented):
The symmetric cross-family validation closes the gap from the prior review — both directions error with actionable messages now. One remaining hole, inline: a per-workflow `model:` override **without** a `harness:` change skips validation entirely (both checks gate on `wf_harness is not None`). Verified locally — `workflows.nightly.model: opus-99` with no `harness:` loads cleanly and the renderer emits `model: opus-99` into a claude action step.

Likely a minor concern in practice (the docstring frames `model` as pairing with `harness`), but easy to close — see the inline suggestion.

max-sixty:
Closed in 95f047b. Restructured the validation: the outer guard is now `wf_harness is not None OR wf_model is not None`, which catches the `model:`-only override path you flagged.

Cases the new logic covers (all gated on "the workflow overrode something"):

| Case | Example | Caught by |
|------|---------|-----------|
| A. model typo, no harness change | `model: opus-99` (top-level claude) | effective-harness allowlist check |
| B. cross-family flip, target has allowlist, bad model | top: codex/gpt-5.5; wf: harness=claude | same effective-allowlist check |
| C. cross-family flip OUT of allowlist, no per-workflow model | top: claude/opus; wf: harness=codex | source-allowlist check |

Tests:
- `test_per_workflow_model_typo_without_harness_rejected` — the specific gap you found.
- `test_per_workflow_model_only_override_valid` — positive case: `model: haiku` (no harness change) loads + renders.

256 tests pass (+2).

> _This was written by Claude Code on behalf of Maximilian._

tend-agent:
Restructure looks right — the outer `wf_harness is not None or wf_model is not None` guard now covers the model-only path, and collapsing A+B into the effective-allowlist check is cleaner than the prior two-branch form. Verified the new tests pass locally. No remaining concerns from my side.

tend-agent (approved):

Issue #624, opened 2026-05-27
Why does Codex account have to be separate for Tend but not Claude account?


endigma:
> On public repos the token must come from a ChatGPT account dedicated to the bot; recommended on private. 

From README.md

What's the difference here? Is it something specific to the Codex harness? Perhaps an adapter could be made for Pi or OpenCode if so?

tend-agent:
Thanks for the question — it's specific to the shape of the Codex credential, not to the harness wiring.

The `CODEX_AUTH_JSON` secret is an OAuth refresh token bound to a ChatGPT account. If it leaks, the attacker can run ChatGPT API calls *as that account* — on a personal account that means reading chat history, accessing custom GPTs, and burning plan quota. The Claude OAuth token (from `claude setup-token`) only authorizes Anthropic API calls billed to the account; it doesn't expose Claude.ai conversation history or other account-level data. That's the asymmetry — same general shape (subscription-funded OAuth), different blast radius on leak.

OpenAI's own [CI/CD auth guide](https://developers.openai.com/codex/auth/ci-cd-auth) discourages `auth.json` for public/open-source repos on the assumption it comes from a personal account. Tend's mitigation is to require it come from a ChatGPT account *dedicated to the bot* on public repos — that narrows the leak to subscription-quota burn until rotation, comparable to an `OPENAI_API_KEY` leak. The leak table and rationale are in [`plugins/install-tend/skills/install-tend/references/security-model.md`](https://github.com/max-sixty/tend/blob/main/plugins/install-tend/skills/install-tend/references/security-model.md#L63-L86).

So the rule isn't Codex-specific in principle — it's "the harness credential is an account-bound OAuth refresh token that exposes account data on leak." A hypothetical Pi or OpenCode harness would only need the same gate if its auth had that property; an API-key-style credential (like `OPENAI_API_KEY` or `ANTHROPIC_API_KEY`) doesn't.

A maintainer can clarify further if needed.

endigma:
Would it be possible to somehow mitigate this using a credential injecting proxy, or something like that? Most people cannot afford a separate ChatGPT Pro login for a bot

tend-agent:
Two parts:

[...]

PR #625, opened 2026-05-28
codex: remove codex_auth_json (auth.json path)

tend-agent:
## Summary

Removes `codex_auth_json` (the Codex subscription `auth.json` path) entirely. The original PR was docs-only with the code removal staged for a follow-up; per @max-sixty's review feedback those steps are merged here.

The reason it doesn't work: Codex rotates the refresh token on every API call and invalidates the prior one after a short grace window. Tend runs review/mention/triage/nightly concurrently, so each in-flight job invalidates the credential the others are using. A scheduled refresher (`codex-auth-refresh.yaml`) worked around the ~8-day full-rotation schedule but not the per-call collision between concurrent jobs.

## Changes

- `codex/action.yaml` — drops the `codex_auth_json` input, the `Configure Codex auth` step, and the dual-auth validation; `openai_api_key` is now `required: true`. Description and `Validate auth configured` step reworded.
- `generator/src/tend/config.py` — drops `codex_auth_json_secret` field and the `codex_auth_json` key from `KNOWN_SECRETS_KEYS`.
- `generator/src/tend/templates/macros.yaml.j2` — Codex agent step no longer emits the `codex_auth_json:` line.
- `generator/src/tend/workflows.py` — install-test workflow no longer references `CODEX_AUTH`.
- `generator/src/tend/checks.py` — `check_codex_auth` only checks `OPENAI_API_KEY`; engine_auth_secrets follows suit.
- `generator/tests/` — `test_codex_engine_passes_with_auth_json` removed; `test_codex_engine_fails_when_no_auth` now asserts only `OPENAI_API_KEY`; codex regtest snapshots regenerated.
- `.github/workflows/codex-auth-refresh.yaml` — deleted (no consumer left).
- `.config/tend.yaml` — `CODEX_AUTH_JSON` removed from `secrets.allowed`.
- Docs — README, install-tend SKILL.md, security-model.md, both `tend.example.yaml` copies, TODO.md reworded from "deprecated / being removed" to "not supported"; migration guidance for adopters with a prior `auth.json` install kept in `security-model.md`.

## Test plan

- [x] `uv run pytest` in `generator/` — 254 passed (one test dropped vs. baseline because `auth.json` no longer has a passing path).
- [x] `uvx pre-commit run --all-files` passes.
- [ ] Maintainer review.


tend-agent (commented):
Two strings in [`codex/action.yaml`](https://github.com/max-sixty/tend/blob/4e8d0371e167e1d9ca1c3abd27b766cd50874f33/codex/action.yaml) still recommend `CODEX_AUTH_JSON` as a normal option and slipped past the docs sweep — both fall just outside the diff, so the suggested-edit UI can't anchor them, but they're the only two places in `codex/action.yaml` that the PR doesn't reconcile with its own framing. Flagging here so you can either fold them into this PR or take them in PR 2:

- `openai_api_key` input description (`description: One of openai_api_key or codex_auth_json must be set; codex_auth_json takes precedence when both are present.`) — still presents the two inputs as peers and notes `codex_auth_json` wins on collision. After this PR a user who follows that hint (sets both, expecting the subscription path to dominate) lands on the broken credential. Suggested rewrite, keeping the precedence note (since the precedence still exists in the unchanged `Configure Codex auth` step) but flagging it as the wrong path:

  ```
  OpenAI API key for the Responses API, billed per token. The only
  supported codex auth — codex_auth_json is incompatible with tend's
  concurrent workflows and is being removed. (If both are set,
  codex_auth_json still takes precedence; clear it to use this key.)
  ```

- `Validate auth configured` step's error message (`echo "::error::No Codex auth configured. Set either the OPENAI_API_KEY secret (API-billed) or the CODEX_AUTH_JSON secret (subscription-funded, see docs/security-model.md for the caveat)."`) — the user-visible failure mode is "fresh install missing codex auth", which is exactly the moment the message should route them to `OPENAI_API_KEY` rather than offer `CODEX_AUTH_JSON` as a peer alternative. Suggested rewrite:

  ```
  ::error::No Codex auth configured. Set the OPENAI_API_KEY secret (CODEX_AUTH_JSON is incompatible with tend's concurrent workflows — see docs/security-model.md).
  ```

Everything else looks consistent with the issue #624 framing; the `codex_auth_json` input description and references doc both name the per-call rotation as the root cause and route to `OPENAI_API_KEY`, and the install skill's 7b dropped the mint/refresh flow cleanly.

max-sixty:
@tend-agent please make those changes?

codex/action.yaml
```diff
@@ -13,19 +14,20 @@ inputs:
...
     required: false
     default: ""
     description: >-
-      Contents of the auth.json Codex writes after `codex login
-      --device-auth` — funds the run from a flat-rate Plus/Pro/Business
-      subscription. Takes precedence over openai_api_key when set.
-      Carries read+write access to the minting ChatGPT account; on
-      public repos mint from a dedicated bot account
-      (see docs/security-model.md).
+      DEPRECATED — incompatible with tend's concurrent workflows. Codex
```
max-sixty: I think we just remove the codex_auth_json

tend-agent: Done in 49c59d6. The `codex_auth_json` input is gone from `codex/action.yaml`, the generator no longer emits the `codex_auth_json:` line into Codex workflows, and `.github/workflows/codex-auth-refresh.yaml` is deleted. `openai_api_key` is now `required: true`, the dual-auth validation is replaced by a single OPENAI_API_KEY presence check, and the regtest snapshots + `test_checks.py` were updated to match. Migration guidance for adopters with a prior `auth.json` install stays in `security-model.md`.

Issue #596, opened 2026-05-24
Weekly integration test failed


tend-agent:
Run: https://github.com/max-sixty/tend/actions/runs/26358181381

Failed at §5 (generator drift check). §3 (tend-triage) and §4 (tend-review) passed end-to-end against `tend-agent/tend-integration`.

## Diff captured

```diff
-# Generated by tend 0.1.0. Regenerate with: uvx tend@latest init
+# Generated by tend 0.1.1. Regenerate with: uvx tend@latest init
...
-      - uses: max-sixty/tend@0.1.0
+      - uses: max-sixty/tend@0.1.1
```

`tend-review.yaml` and `tend-triage.yaml` only — version-pin lag, no other generator changes.

## Root cause

`tend-agent/tend-integration` was bootstrapped on tend `0.1.0`. The bootstrap config disables `nightly` (and every other workflow beyond review/triage), so tend-integration never self-regenerates and falls behind every published release. §5 will fail on every weekly run between a tend release and the next bootstrap-rerun.

## Fix options

1. **Have §5 self-heal**: regenerate, commit, push to tend-integration's `main` (bot owns the repo, `enforce_admins: false`, so direct-push works), then exit clean. The recipe already authorizes destructive writes against this one repo.
2. **Enable nightly on tend-integration**: drop `nightly: false` from the bootstrap config; the nightly run picks up the new tend release automatically. Cheapest fix but adds an extra workflow run per night.
3. **Filter version-bump diffs in §5**: ignore changes that only touch the `# Generated by tend X.Y.Z` and `max-sixty/tend@X.Y.Z` lines. Catches the original generator-regression intent without false-firing on releases.

I did not push a regen this run — that would be a behavior change beyond the recipe.

tend-agent:
Re-fired this week: https://github.com/max-sixty/tend/actions/runs/26709622272

Two new findings on top of last week's drift report.

## Secret-reseed clobber (new — caused §3 and §4 to fail this run)

[...]

PR #636, opened 2026-05-31
skills(running-tend): fix consumers.json search recall and integration-test secret clobber

tend-agent:
Fixes two structural bugs in `.claude/skills/running-tend/` that surfaced during this week's `/tend-ci-runner:weekly` run ([#596 update](https://github.com/max-sixty/tend/issues/596#issuecomment-4586380539)).

## 1. `consumers.json` search misses tend itself

`gh search code 'max-sixty/tend' --limit 100` returns README/CLAUDE.md/TODO.md/etc. hits ahead of workflow files. With the 100-result cap, tend's own `.github/workflows/tend-*.yaml` files are crowded out, and the `.path | startswith(".github/workflows/tend-")` filter drops tend from the regenerated list. Confirmed empirically this run: 4 repos returned, missing `max-sixty/tend`. Re-running with `--extension yaml` returns all 5.

The committed `data/consumers.json` already lists tend, so the data file is unchanged this week — but a fresh regeneration would silently drop tend.

## 2. Integration-test bootstrap clobbers `CLAUDE_CODE_OAUTH_TOKEN`

`§1 Bootstrap` unconditionally pipes `$CLAUDE_CODE_OAUTH_TOKEN` into the secret on every weekly run. The recipe claimed the variable was available from the claude step's env, but the harness sets `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB`, which scrubs Anthropic credentials from this subprocess. Verified with `[ -n "$CLAUDE_CODE_OAUTH_TOKEN" ]` — unset, length 0.

The unguarded reseed therefore pipes empty into the integration repo's `CLAUDE_CODE_OAUTH_TOKEN` every run. This week's run did exactly that, and §3's tend-triage workflow failed at the action's auth preflight ([failed run](https://github.com/tend-agent/tend-integration/actions/runs/26709684701)).

Guard the reseed with `[ -n "$CLAUDE_CODE_OAUTH_TOKEN" ]` so a scrubbed env no-ops instead of overwriting. Updated the §0 "Safety" preamble to be honest about which env vars are actually available and to flag rotation as a manual task.

## Follow-up

A maintainer still needs to manually re-set `CLAUDE_CODE_OAUTH_TOKEN` on `tend-agent/tend-integration` to a valid value — this PR prevents future clobbers but doesn't undo this week's. The §5 drift issue (#596 original body) is independent of this PR.

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -65,9 +65,12 @@ because the workflow files are public.
 #    version tag (`max-sixty/tend@X.Y.Z`, or `/codex@X.Y.Z`), so search the
 #    bare `max-sixty/tend` token (version-agnostic; GitHub code search does
 #    not index `@` or `/`, so this matches both the Claude and Codex refs).
+#    `--extension yaml` is required: without it, README/CLAUDE.md/TODO.md
+#    hits on `max-sixty/tend` itself crowd out tend's own workflow files
+#    past the 100-result cap, dropping tend from its own consumers.json.
 #    The `.github/workflows/tend-` path filter below bounds precision.
 mapfile -t REPOS < <(
-  gh search code 'max-sixty/tend' --limit 100 --json repository,path \
+  gh search code 'max-sixty/tend' --extension yaml --limit 100 --json repository,path \
     | jq -r '.[] | select(.path | startswith(".github/workflows/tend-")) | .repository.nameWithOwner' \
     | sort -u
 )
--- a/.claude/skills/running-tend/references/integration-test.md
+++ b/.claude/skills/running-tend/references/integration-test.md
@@ -20,11 +20,15 @@ creates it. Once it exists, subsequent weekly runs only operate on it.
 claude step in `action.yaml`. `gh` authenticates as `tend-agent` via
 `$GITHUB_TOKEN` automatically.
 
-`$CLAUDE_CODE_OAUTH_TOKEN` is **not** available here — the harness sets
-`CLAUDE_CODE_SUBPROCESS_ENV_SCRUB`, which strips Anthropic credentials
-from this subprocess. Rotation on the integration repo is therefore a
-manual maintainer task; the reseed below is guarded so a scrubbed env
-never clobbers the stored secret.
+`$CLAUDE_CODE_OAUTH_TOKEN` is **not** available here — empirically
+unset/empty in this subprocess despite [`action.yaml`'s `env:`
+block](../../../action.yaml) exposing it on the claude step with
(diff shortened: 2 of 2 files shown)


tend-agent (commented):
Both functional changes are correct — `--extension yaml` returns all 5 consumer repos including tend, and the `[ -n "$CLAUDE_CODE_OAUTH_TOKEN" ]` guard prevents the empty-clobber regardless of root cause.

One discrepancy worth reconciling before this lands: the new preamble attributes the missing `$CLAUDE_CODE_OAUTH_TOKEN` to `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB`, but [`action.yaml`](https://github.com/max-sixty/tend/blob/dcbcac17c1794544b9284d708bcb0524a517ca15/action.yaml#L212-L217) sets `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: "0"` and explicitly exposes `CLAUDE_CODE_OAUTH_TOKEN` to the agent step "so skills running in the agent's bash can propagate it (e.g. seeding a managed repo's secrets)" — which is precisely the use case this recipe relies on. The two narratives in this repo contradict each other: action.yaml says the var is exposed by design, the integration-test recipe now says it isn't available because of env scrub.

The defensive guard is right either way and shouldn't wait on this, but the underlying question — does claude-code-action scrub Anthropic credentials even with `ENV_SCRUB=0`, in which case action.yaml's "expose" intent is broken? — looks worth a follow-up. If the harness genuinely can't surface the OAuth token to skills, the action.yaml comment block deserves a matching honesty pass.

tend-agent:
Confirmed and pushed [`657d959`](https://github.com/max-sixty/tend/pull/636/commits/657d959): empirical check from this same harness step shows `[ -n "$CLAUDE_CODE_OAUTH_TOKEN" ]` returns unset/empty *and* `[ -n "$ANTHROPIC_API_KEY" ]` returns unset/empty, even though `action.yaml` sets `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=0` and declares both vars in the step's `env:` block. So you're right that the preamble's `ENV_SCRUB` story was wrong — that knob is off. Some other layer (most plausibly Claude Code itself stripping its own Anthropic credentials before spawning Bash-tool subprocesses, independent of bubblewrap) is doing it.

I've replaced the speculative attribution with an empirical statement and a "root cause not yet pinned down" hedge. The defensive guard stays — it's correct under either narrative.

On the follow-up: yes, this looks worth a separate look at `action.yaml:212-217`. If the strip is unconditional, the "Expose the harness auth so skills running in the agent's bash can propagate it" comment block is aspirational rather than descriptive, and the propagation pattern it enables (`gh secret set ... --repo managed-repo`) doesn't actually work from inside an agent run. I'll open a tend issue to investigate rather than try to bundle that into this PR.

PR #652, opened 2026-06-03
interactive: isolate the GitHub PAT behind a credential-injecting proxy

max-sixty:
Keeps a prompt-injected CI agent from reading the bot's GitHub PAT, on the `claude-interactive` harness. Prompted by a critique that tend's tokens are accessible to the model.

## Approach

The agent runs as a separate, non-sudo `tend-sandbox` user. The real PAT lives only in a local `mitmproxy` (running as `runner`) that injects it into requests to GitHub hosts. The agent holds only a dummy `GH_TOKEN` plus `HTTPS_PROXY`, so it can authenticate to GitHub only *through* the proxy and can't read the real token: different UID, no sudo, and `/proc/<pid>/environ` is owner-only. The PAT is also stripped from `.git/config` (where `actions/checkout` persists it via an `includeIf` → external file).

The proxy gates injection on the connection's real destination (`flow.request.host`), not the client-supplied `Host` header, so a spoofed `Host: api.github.com` to an attacker host gets no token. The injection allowlist is exact-match on four GitHub hosts.

bwrap was the obvious alternative but isn't on the stock runner image and Ubuntu 24.04 restricts unprivileged user namespaces; UID separation alone gives the credential isolation. The namespaces would mainly buy egress lockdown, which conflicts with tend needing to build and test arbitrary repos.

## Scope

GitHub credential only. The Claude OAuth/API token still reaches the agent's env — a later cut can route inference through the same proxy via `ANTHROPIC_BASE_URL`.

## Key files

- `interactive/proxy/github_auth.py` — the mitmproxy addon and its exact-host injection allowlist (the credential boundary).
- `interactive/proxy/setup-sandbox.sh` — runner-side: create the user, neutralize the persisted git credential, hand over the checkout, start the proxy, system-trust its CA.
- `interactive/action.yaml` — wires the sandbox setup, toolchain handoff, and the user-dropped agent launch into the harness.
- `interactive/proxy/test_github_auth.py` — addon tests (incl. the Host-spoof regression), wired into `ci.yaml` as a standalone `test-proxy` job.

## Validation

`interactive-smoke` runs against the production shape (`persist-credentials: true`) and is green: a dummy-token sandbox agent authenticated as the bot (`gh api user` → `tend-agent`) and ran `git ls-remote origin` successfully, proving the credential strip neither leaks the PAT nor breaks git.

## Known gaps / follow-ups

- The Claude OAuth/API token still reaches the agent's env (phase 2).
- `*.githubusercontent.com` / git-LFS aren't in the injection allowlist (fine for the common gh/git/push path; private raw/asset fetches would fail).
- Under `persist-credentials: true`, chowning `.git` makes `actions/checkout`'s post-cleanup warn (non-fatal).
- The toolchain handoff copies ~200MB with `cp -a` every run; installing as the sandbox user would avoid it, but `mv` breaks the binary's absolute-path refs, so it's a larger refactor.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,11 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - uses: astral-sh/setup-uv@v7
-      - run: uv run --no-project --with mitmproxy --with pytest pytest
+      # Pin to match interactive/action.yaml's mitmproxy_version default so the
+      # test exercises the same version the production proxy runs — the addon
+      # imports mitmproxy.test (an internal helper), so version skew could green
+      # the test while production breaks.
+      - run: uv run --no-project --with 'mitmproxy==12.2.1' --with pytest pytest
 
   test-worker:
     runs-on: ubuntu-24.04
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,6 +15,19 @@ jobs:
       - uses: astral-sh/setup-uv@v7
       - run: uv run pytest
 
+  # Tests for the claude-interactive credential-injection proxy addon. Run
+  # standalone (no project) with mitmproxy + pytest pulled in ad hoc, since the
+  # proxy isn't part of the generator package.
+  test-proxy:
+    runs-on: ubuntu-24.04
+    defaults:
+      run:
+        working-directory: interactive/proxy
+    steps:
+      - uses: actions/checkout@v6
+      - uses: astral-sh/setup-uv@v7
+      - run: uv run --no-project --with mitmproxy --with pytest pytest
+
   test-worker:
     runs-on: ubuntu-24.04
     defaults:
--- a/.github/workflows/interactive-smoke.yaml
+++ b/.github/workflows/interactive-smoke.yaml
@@ -73,6 +73,12 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 1
+          # The agent runs as a non-sudo sandbox user and authenticates to
+          # GitHub only through the injecting proxy, so it needs no persisted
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -15,6 +15,19 @@ jobs:
...
+  # proxy isn't part of the generator package.
+  test-proxy:
+    runs-on: ubuntu-24.04
+    defaults:
+      run:
+        working-directory: interactive/proxy
+    steps:
+      - uses: actions/checkout@v6
+      - uses: astral-sh/setup-uv@v7
+      - run: uv run --no-project --with mitmproxy --with pytest pytest
```
tend-agent: Runtime pins `mitmproxy==12.2.1` (the `mitmproxy_version` default in `interactive/action.yaml`), but this test command leaves it unconstrained — `uv` resolves to the latest 12.x (currently 12.2.3). The addon imports `mitmproxy.test`, an internal helper; a patch-level bump there could green the test while the production proxy breaks (or vice-versa). Pinning both sides to the same version keeps the test signal aligned with what runs in CI.

```suggestion
      - run: uv run --no-project --with 'mitmproxy==12.2.1' --with pytest pytest
```

tend-agent (commented):
Walked the credential boundary end-to-end — UID separation + proxy-held token + exact-host injection allowlist means a prompt-injected agent can't reach the PAT through env, `/proc`, `.git/config`, or a lookalike host (mitmproxy intercepts `api.github.com.evil.example` because the `--allow-hosts` regex matches "github.com", but the addon's frozenset gate refuses to inject there). One pin-drift nit inline.

tend-agent (approved):

tend-agent (approved):

PR #653, opened 2026-06-03
feat(worker): share /activity payload across colos via a KV tier

max-sixty:
The colo cache (`caches.default`) is per-data-center, so each Cloudflare colo fanned out to GitHub's Search API on its own. During a traffic spike spread across colos, those 4·N concurrent Search bursts stack against the 30 req/min cap (shared per token), so the rate-limit risk grew with colo count and traffic, exactly when the data matters most.

`/activity` now publishes its rendered payload to KV (a global store) behind the hot colo cache. `refreshShared` is the single choke point both the cold path and the background refresh funnel through: it reuses a sibling colo's still-fresh KV entry instead of fanning out, and publishes its own result when it does fan out. The GitHub fanout then fires roughly once per freshness budget across the whole network instead of once per colo, independent of colo count or traffic.

On a cold colo cache a sibling's KV entry is served immediately (stale entries are served and revalidated in the background), so a burst after idle no longer makes every colo block on and dogpile the fanout. KV survives deploys and outlives the colo cache, so the cold-fanout 503 fires only when colo cache and KV are both empty, and a cold start during a GitHub outage shows last-known data rather than nothing.

`/currently-tending` is unchanged: its 30s budget is under KV's 60s floor, and its fanout is cheap core REST, not the Search quota the sharing protects.

## Trade-off

Coordination is best-effort, not a global lock: colos going stale at the same instant can still race within KV's write-propagation window (up to ~60s) before the first write is visible. The race is bounded (later arrivals read the fresh entry) and degrades gracefully (serve-stale-from-KV, never blank). A cron-driven single refresher would make it a hard guarantee but trade away the zero-when-idle property; not worth it at this scale.

## Files

- `worker/src/index.ts`: `refreshShared` (the KV choke point, returning the entry so its stale-at survives), the KV tier in `refreshAndCache`, and `entryResponse` (every data response inherits its entry's stale-at, so a colo never resets the shared clock on either the cold or the background-refresh path).
- `worker/test/worker.test.ts`: a `refreshShared` block, KV-tier cases in `refreshAndCache`, and a cross-colo case in `coalesceAndRefresh`.
- `worker/README.md`: Caching / Cache strategy rewritten for the two tiers.

### DIFF
--- a/worker/src/index.ts
+++ b/worker/src/index.ts
@@ -277,12 +277,22 @@ function readKvEntry<T>(env: Env, key: string): Promise<KvEntry<T> | null> {
 // GitHub fanout. Both the cold path and the background refresh funnel through
 // here, so the GitHub fanout fires at most once per freshness budget across
 // every colo except those racing within KV's write-propagation window.
-async function refreshShared<T>(env: Env, opts: CacheOpts<T>): Promise<T> {
-  if (!opts.kvKey) return opts.refresh();
+//
+// Returns the entry, not just the payload: a reused KV entry carries its
+// original stale-at, a freshly-fetched one `now + ttl.ok`. Callers stamp the
+// colo cache from that, so a colo inherits the shared clock instead of
+// resetting it (see entryResponse).
+async function refreshShared<T>(env: Env, opts: CacheOpts<T>): Promise<KvEntry<T>> {
+  // staleAt is read after the fanout resolves (property order), so a slow
+  // fanout doesn't shorten the published freshness window.
+  const mint = async (): Promise<KvEntry<T>> => ({
+    payload: await opts.refresh(),
+    staleAt: Date.now() + opts.ttl.ok * 1000,
+  });
+  if (!opts.kvKey) return mint();
   const entry = await readKvEntry<T>(env, opts.kvKey);
-  if (entry && Date.now() < entry.staleAt) return entry.payload;
-  const payload = await opts.refresh();
-  const fresh: KvEntry<T> = { payload, staleAt: Date.now() + opts.ttl.ok * 1000 };
+  if (entry && Date.now() < entry.staleAt) return entry; // sibling colo's fresh write
+  const fresh = await mint();
   await env.CACHE.put(opts.kvKey, JSON.stringify(fresh), {
     // Outlive the freshness budget by STALE_SERVE_FACTOR, matching the colo
     // cache, so a quiet stretch still leaves a (stale) entry to serve. Stays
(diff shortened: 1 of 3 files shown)


worker/src/index.ts
```diff
@@ -337,7 +420,7 @@ async function coalesceAndRefresh<T>(
   bumped.headers.set(STALE_AT_HEADER, String(Date.now() + REFRESH_GRACE_MS));
   await caches.default.put(cacheKey, bumped);
   try {
-    const fresh = jsonResponse(await opts.refresh(), env, opts.ttl.ok);
+    const fresh = jsonResponse(await refreshShared(env, opts), env, opts.ttl.ok);
```
tend-agent: `refreshShared` returns just the payload, so when it serves a still-fresh KV entry, the entry's `staleAt` is dropped here — `jsonResponse` re-stamps the response with `staleAt = now + ttl.ok * 1000`. That's the case the design targets (warm-stale colo finds a sibling colo's fresh KV write), so it's the common path, not a rare race.

Effect: the colo cache marks the entry fresh for a full freshness budget from now, even though KV consensus says it goes stale at the original instant. A KV entry written ~200 s ago by a sibling will keep being served as fresh in this colo for ~ttl.ok seconds after KV considers it stale, until the bumped colo entry naturally times out. That's exactly the "colo resets the shared clock" failure `kvHydratedResponse` calls out and prevents on the cold-colo path (`refreshAndCache`, when the KV read finds an entry).

The knock-on effect is on average data age: with synchronized network-wide staleAt, the average age would be ~0.5·ttl.ok (same as without sharing); with this drift, refreshes lag the KV consensus by up to a full budget, so the average age roughly doubles. Not catastrophic — bounded by the colo's `STALE_SERVE_FACTOR` window — but it partially undoes the freshness side of KV sharing.

A fix shape: have `refreshShared` return `KvEntry<T>` (with `staleAt = now + ttl.ok * 1000` for a just-fetched payload, or the original `staleAt` for a KV read), and let the caller go through `kvHydratedResponse` in both branches. The just-fetched case produces the same response `jsonResponse` does today, and the KV-sourced case inherits the shared clock correctly. The same applies, narrowly, to `refreshAndCache`'s no-KV-entry branch (`response = jsonResponse(await refreshShared(env, opts), …)` on line 383) — there a sibling could race in between this colo's KV-empty read and `refreshShared`'s second KV read, but the window is much narrower than here.

Worth noting: the `coalesceAndRefresh` test that exercises the fresh-KV path (`pulls a sibling colo's fresh KV entry instead of fanning out`) only asserts the cached body, not the `STALE_AT_HEADER`, so this drift isn't caught by tests today.

tend-agent (commented):
One observation on the stale-at preservation, inline.

worker/src/index.ts
max-sixty: Good catch — fixed in 7e6f761. `refreshShared` now returns the `KvEntry` (payload + stale-at) rather than just the payload, and both data paths (the cold path and `coalesceAndRefresh`) build the response via `entryResponse`, so a reused KV entry's original stale-at is preserved instead of being re-stamped `now + ttl.ok`. A freshly-fetched entry carries `now + ttl.ok`, so a fanout produces the same response as before. Tests now assert the inherited stale-at on both the `refreshShared` result and the `coalesceAndRefresh` colo entry.

tend-agent (approved):

PR #660, opened 2026-06-06
skills(running-in-ci): forbid literal placeholders in user-visible bodies

tend-agent:
## Summary

Add a new bullet to `running-in-ci`'s "Specific failure modes" section: never type literal placeholder strings (`<PLACEHOLDER>`, `PR #PLACEHOLDER`, `TBD`, `XXX`, `TODO(fill)`) into user-visible content. Either resolve the cross-reference before composing the body, or omit it.

## Evidence

`review-reviewers` run [27052280831](https://github.com/max-sixty/tend/actions/runs/27052280831) surfaced one Critical, structural occurrence on `max-sixty/worktrunk`:

- **Trigger**: [worktrunk-bot mention on PR #2983](https://github.com/max-sixty/worktrunk/pull/2983#issuecomment-4637273025) at 2026-06-06T03:49:29Z from maintainer @max-sixty. The PR was already MERGED at 01:40:10Z (~2h before the mention). The mention requested a four-part deprecation cycle, including "open an issue to, in three months, turn `commits` into an error."
- **Outcome**: `tend-mention` run [27051779110](https://github.com/max-sixty/worktrunk/actions/runs/27051779110) opened [issue #2984](https://github.com/max-sixty/worktrunk/issues/2984) at 03:56:58Z. Step 1 of the deprecation cycle in the issue body reads literally:

  > 1. **Now (deprecated):** PR #PLACEHOLDER switches the default squash template, docs, and example config to `commit_details` + `.subject`…

  The placeholder shipped intact — no PR was filed, no number was substituted, and the body is permanently visible.
- **Session diagnosis** (cheap-subagent trace of `/tmp/session-logs/27051779110/...jsonl`): the bot drafted the issue body in one pass with the literal string `PR #PLACEHOLDER`, posted it via `gh issue create --body-file`, then queued the deprecation-PR work behind a backgrounded `cargo run -- hook pre-merge` gate (background task `b2tkbltf2`). The session ended before the gate notified, so the PR step never ran, no PR was filed, and the placeholder was never substituted.

Evidence log: https://gist.github.com/15c523384cfad7449cb81933b47165ff

## Gate assessment

- **Evidence level**: Critical — clearly wrong outcome shipped to a permanent user-visible artifact (issue body with literal `PR #PLACEHOLDER`).
- **Occurrence count**: 1 (this run). Per Gate 1, Critical needs 1.
- **Classification**: Structural. The defect arises from a "fill later" composition strategy that breaks when the deferred step doesn't run — the same prompt shape (multi-step mention with cross-references between artifacts to be created) would tempt the same reasoning chain. Not stochastic forgetfulness.
- **Magnitude**: Targeted fix — one new bullet in an existing failure-modes section, plus a pre-post grep guard. Gate 2 passes at Normal threshold (Critical structural = 1 is enough for a targeted fix).
- **Verdict**: passes both gates.

## Why this bullet, not a workflow change

A workflow-level guard (e.g. a pre-`gh ... create` hook that greps body files for placeholder shapes) would be more robust but is a structural change requiring a higher evidence bar (5+). The single-bullet skill addition is proportionate to one Critical occurrence and codifies both the rule and the recipe (sequence the work; grep the body before posting) so the bot can self-enforce on the next similar trigger. If the pattern recurs, escalating to a workflow-level guard becomes warranted.

## Test plan

- [ ] Maintainer reads the new bullet for clarity and scope.
- [ ] Next consumer regeneration (nightly `uvx tend@latest init` after this lands in a release) picks up the updated skill via the pinned action ref.
- [ ] Next observed multi-step mention with deferred cross-references composes the issue/PR with values resolved before posting (or omits the cross-reference), and grep-screens the body file pre-post.


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -521,6 +521,8 @@ If you can't find source evidence for a specific detail, say so ("I'm not sure o
...
 
+**Never ship literal placeholders in user-visible content.** Strings like `<PLACEHOLDER>`, `PR #PLACEHOLDER`, `<SHA>`, `TBD`, `XXX`, or `TODO(fill)` in an issue body, PR body, or comment are corruption: a deferred substitution that never ran. They survive into the rendered output and read as broken. When a multi-step ask references an artifact that doesn't yet exist ("file an issue that references the PR I'm about to file"), sequence the work so the referenced artifact exists before the referencing body is composed: create the PR → read its number → compose the issue with the number filled in → file the issue. If the cross-reference can't be resolved before posting (e.g. the artifact is out of scope or deferred), omit it or rephrase ("a follow-up PR will…") rather than emit a placeholder. Before any `gh issue create`, `gh pr create`, or `gh ... comment --body-file`, grep the body file for `PLACEHOLDER`, `<SHA>`, `<TODO`, `TBD`, `XXX` and refuse to post if any match. A session that times out mid-sequence leaves an unsubstituted placeholder permanently visible — pre-substitute, don't post-substitute.
```
```suggestion
**Never ship literal placeholders in user-visible content.** Strings like `<PLACEHOLDER>`, `PR #PLACEHOLDER`, `<SHA>`, `TBD`, `XXX`, or `<TODO(fill)>` in an issue body, PR body, or comment are corruption: a deferred substitution that never ran. They survive into the rendered output and read as broken. When a multi-step ask references an artifact that doesn't yet exist ("file an issue that references the PR I'm about to file"), sequence the work so the referenced artifact exists before the referencing body is composed: create the PR → read its number → compose the issue with the number filled in → file the issue. If the cross-reference can't be resolved before posting (e.g. the artifact is out of scope or deferred), omit it or rephrase ("a follow-up PR will…") rather than emit a placeholder. Before any `gh issue create`, `gh pr create`, or `gh ... comment --body-file`, grep the body file for `PLACEHOLDER`, `<SHA>`, `<TODO`, `TBD`, `XXX` and refuse to post if any match. A session that times out mid-sequence leaves an unsubstituted placeholder permanently visible — pre-substitute, don't post-substitute.
```

Angle-bracketing the `TODO(fill)` example so the grep token `<TODO` actually catches it. (The alternative — changing the grep to `TODO(` — risks false positives on legitimate `TODO(...)` comments inside pasted code blocks, which the angle-bracketed convention avoids.)

tend-agent (commented):
One small inconsistency: the example list includes `TODO(fill)` but the grep guard token is `<TODO`, which wouldn't match the bare-`TODO(...)` form. Either angle-bracket the example or broaden the grep token.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied — angle-bracketed `TODO(fill)` to `<TODO(fill)>` in 7e63c69 so the `<TODO` grep guard catches it. Kept the grep token narrow to avoid false-positives on legitimate `TODO(...)` comments in pasted code.

PR #661, opened 2026-06-06
skills(running-in-ci): end the turn only when work is shipped

tend-agent:
Three sessions over a ~17-hour window did meaningful code work for a maintainer ask, then silently failed to ship anything because each one emitted `end_turn` while a `run_in_background: true` pre-merge gate was still pending. The harness did not resume — none of the runs ever subscribed to a `task-notification` wakeup (one didn't call `Monitor`, two more got `InputValidationError` on a deferred-tool schema race and never retried). Each occurrence leaves a visibly unanswered maintainer ask: branch never pushed, no commit, no comment, no signal the work happened.

This reframes the rule after maintainer review (https://github.com/max-sixty/tend/pull/661#issuecomment-4640896565). The first cut narrowed it to *"ship the deliverable before any long wait"* under **PR Creation**; the failure mode generalises beyond PR creation to **any case where `end_turn` fires while a maintainer-visible deliverable is staged but un-shipped** — comment posting, follow-up commits, inline replies, anywhere a `run_in_background` task gates the next step.

Adds a new top-level section **End the turn only when work is shipped** to `running-in-ci`, placed after **Restrictions**. Rule: treat the session as live until the deliverable is maintainer-visible (pushed, posted, opened); don't background work whose output gates the deliverable; the `run_in_background` recipe under **CI Monitoring** stays safe because it runs *after* push.

**Evidence**

- [`tend-mention` run 27051779110](https://github.com/max-sixty/worktrunk/actions/runs/27051779110) — backgrounded pre-merge gating a deprecation PR, session ended mid-wait, no PR shipped.
- [`tend-notifications` run 27053336290](https://github.com/max-sixty/worktrunk/actions/runs/27053336290) — picked up the deferred work, implemented cleanly on local branch `deprecate/squash-commits-var-27053336290`, ~17 min and $7.35 of session, never called `Monitor`, branch never pushed, no comment.
- [`tend-mention` run 27075856123](https://github.com/max-sixty/worktrunk/actions/runs/27075856123) — different PR (worktrunk#2991), full rework implemented, called `Monitor({})` with empty input → `InputValidationError`, loaded the schema via `ToolSearch` but never retried, `end_turn` with text "I'll be re-invoked when it finishes". It was not.

**Classification.** Three occurrences inside ~17 hours, across two worktrunk PRs and two workflow types (`tend-mention` twice, `tend-notifications` once). Gate 1 (High structural needs 2–3 occurrences). Gate 2 (targeted fix at Normal threshold).

**Evidence gist:** https://gist.github.com/15c523384cfad7449cb81933b47165ff

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -134,6 +134,10 @@ gh pr list --state all --search "author:$BOT_LOGIN <issue-number>" \
 
 Compare by title keywords **and** the files the new PR would modify — two concurrent fixes for the same bug typically pick different branch names, so a branch-name match is not sufficient. If a sibling bot PR overlaps in scope — whether open, closed, or already merged — **do not create**: post a comment on the triggering thread linking the existing PR and exit.
 
+### Ship the deliverable before any long wait
+
+Don't gate `git push` / `gh pr create` on a multi-minute local validation — a full test suite, a comprehensive lint hook, a `Monitor`-waited background task. Session budget is finite; if the wait outlives the session the implementation is lost on a local branch the maintainer can't see, with no comment, no PR, and no signal that the work happened. Sequence the visible deliverable (commit, push, `gh pr create`) **first**; CI re-runs the same gates after push and gives the maintainer progress regardless of how long the local check would have taken. A targeted compile plus the tests directly exercising the change is enough local confidence to ship — leave the comprehensive matrix to CI. If a slow local validation is genuinely needed (rare), run it synchronously and accept the time cost, but don't background it behind a `Monitor` wait that the session may not outlast.
+
 ## Pushing to PR Branches
 
 Always use `git push` without specifying a remote — `gh pr checkout` configures tracking to the correct remote, including for fork PRs. Specifying `origin` explicitly can push to the wrong place.
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -78,6 +78,16 @@ If a linked PR merged (or the triggering PR itself merged) **after the triggerin
 - **Scope**: PRs, pushes, and comments on existing threads in other repos are off-limits. Filing fresh issues in other repos follows **Filing Issues in Other Repos** below.
 - **Hanging commands**: Never use `gh run watch` or `gh pr checks --watch` — both hang indefinitely. Poll with `gh pr checks` in a loop instead.
 
+## End the turn only when work is shipped
+
+Emitting `end_turn` ends the CI session — the runner is discarded, and the harness does not reliably resume it from a background-task completion. If you `end_turn` while a `run_in_background: true` Bash whose result was going to gate the deliverable is still running, the task either finishes invisibly or gets killed when the runner is torn down, and any staged work the maintainer was supposed to see — a committed-but-unpushed branch, a written-but-unsent `/tmp/comment-body.md` — dies with it.
+
+The session is live until the deliverable is **maintainer-visible**: pushed, posted, or opened. Local-only state — a commit nobody else can see, a comment body never sent — does not count and is not recoverable on a follow-up.
+
+Corollary: don't background anything whose output gates the deliverable. If a full test suite or comprehensive lint needs to run before push, run it synchronously and accept the time cost; if it's too slow for the session budget, push first and let CI re-run it. A session that shipped a partial result is recoverable; a session that ended mid-wait with the deliverable on a local branch is not. A targeted compile plus the tests directly exercising the change is enough local confidence to ship — leave the comprehensive matrix to CI.
+
+The `run_in_background: true` recipe under **CI Monitoring** is safe because it runs *after* push: the deliverable is already shipped, so the worst case is a missed CI-result follow-up, not a lost PR. Pre-push has no such fallback.
+
 ## Filing Issues in Other Repos
 
 Default: file an issue in the current repo asking for permission to file in the target. On maintainer approval, file in the target.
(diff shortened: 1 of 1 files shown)


max-sixty:
But why did the session end? That seems like the core problem here!

Is this a problem with running without `-p`? 

Is https://github.com/max-sixty/worktrunk/pull/2986 similar? Is that running without `-p`?

tend-agent:
You're right that this PR papers over the symptom. The root cause:

**Why the session ends.** Both failed runs emitted `stop_reason: end_turn` with the background `cargo run -- hook pre-merge --yes` task still pending. The model treats `run_in_background: true` + `Monitor` as "yield and get woken on completion." That wakeup model only works reliably in **interactive** Claude Code (a human terminal session, where the binary stays alive after `end_turn` waiting for the next user message). In CI the session terminates and the runner is discarded before the background task can notify.

**Is this a `-p` problem?** Structurally yes. worktrunk uses the default `harness: claude`, which runs [`anthropics/claude-code-action`](https://github.com/anthropics/claude-code-action) — the Agent SDK, one-shot, equivalent to `claude -p`. Tend itself runs `harness: claude-interactive` (the PTY-supervised binary), but that harness also exits on `end_turn` via a Stop-hook sentinel ([interactive/action.yaml:5-7](https://github.com/max-sixty/tend/blob/0af266b/interactive/action.yaml#L5-L7)) — same constraint in CI for a different reason. Neither CI harness can hold the session open through a multi-minute background wait. There is one subtle wrinkle: `action.yaml` already documents `task-notification` wakeups that emit a second `result` ([action.yaml:321-324](https://github.com/max-sixty/tend/blob/0af266b/action.yaml#L321-L324)), so claude-code-action does have *some* wakeup support — but in worktrunk's runs the wakeup never fired before the session terminated, suggesting either the wakeup window is shorter than the gate's runtime or the model emitted a "final" `end_turn` the SDK didn't queue a wakeup for. I haven't traced exactly which.

**Is worktrunk#2986 the same?** Yes — same `end_turn`-with-pending-background root cause, same `pre-merge` gate, same two runs cited. The difference is scope: worktrunk#2986 adds a narrow rule to worktrunk's `running-tend` overlay (specifically: foreground the pre-merge hook), this PR adds a broader rule to the bundled `running-in-ci` skill (any long pre-push wait). They don't conflict, but they're redundant if both land.

Given the questions, two options I'd take direction on:

1. **Close this, keep worktrunk#2986.** The bundled rule is overreach if the only consumer hitting it has a project-specific gate; worktrunk-specific guidance is the right level.
2. **Reframe this** so it states the harness constraint up-front ("Agent-SDK / `claude -p` mode terminates on `end_turn`; CI never resumes from a background-task wakeup the way interactive Claude does") rather than just the rule, so future readers understand *why* and can decide differently if the harness ever changes.

The deeper structural fix — having `claude-code-action` block `end_turn` while a `run_in_background` Bash is still running — would belong upstream in claude-code-action and is out of scope here.

tend-agent:
3rd occurrence of the same shape this hour, on a different PR.

[`tend-mention` run 27075856123](https://github.com/max-sixty/worktrunk/actions/runs/27075856123) — `handle` ran 22:39:11Z → 22:51:40Z (12m29s), triggered by [max-sixty's "great, yes, let's do that" on PR #2991](https://github.com/max-sixty/worktrunk/pull/2991#issuecomment-4640664984) (agreeing to centralise control-byte escaping in `src/logging.rs` rather than the narrow per-call site the earlier commits used). The session implemented the full rework — added `escape_controls` helper to `src/utils.rs`, reverted producer-side escaping in `src/shell_exec.rs`, applied centralised escaping in `src/logging.rs` + `src/diagnostic.rs`, added an integration test `test_vv_artifacts_have_no_raw_control_bytes` — ran targeted tests locally (all passed), then queued `cargo run -- hook pre-merge --yes` as `run_in_background` and `Monitor`-waited. Final assistant text: *"Waiting on the full suite before committing and pushing."* Session emitted `end_turn`; never resumed. Zero output on PR #2991: branch never pushed, no commit, no comment. Same `stop_reason: end_turn` + pending background pre-merge as the two runs in the PR body.

**Cadence.** 3 occurrences inside ~17 hours, across 2 worktrunk PRs (#2983 twice, #2991 now) and 2 workflow types (tend-mention twice, tend-notifications once). All inside max-sixty/worktrunk — consistent with [worktrunk#2986](https://github.com/max-sixty/worktrunk/pull/2986)'s overlay-level scope. The bundled-vs-overlay choice from my prior comment still stands; this just sharpens the rate.

**Maintainer-impact signal.** Each occurrence leaves a visibly unanswered maintainer ask. PR #2991 now sits with max-sixty's "great, yes, let's do that" unresponded-to, even though the bot did the rework — the maintainer can't see that and has no reason to expect a sibling cycle will redo it. The invisibility is the user-facing harm worth pricing in.

No response needed — posting as evidence for the open decision. Recorded in the [evidence gist](https://gist.github.com/15c523384cfad7449cb81933b47165ff).

max-sixty:
> There is one subtle wrinkle: action.yaml already documents task-notification wakeups that emit a second result (action.yaml:321-324), so claude-code-action does have some wakeup support — but in worktrunk's runs the wakeup never fired before the session terminated, suggesting either the wakeup window is shorter than the gate's runtime or the model emitted a "final" end_turn the SDK didn't queue a wakeup for. I haven't traced exactly which.

please explain?

max-sixty:
consider generalizing this to a simple "do not end the turn until everything is complete"?

tend-agent:
Traced both session logs — the wakeup never fired because **neither failed run successfully subscribed to one**. Resolves my earlier "haven't traced exactly which" hedge:

**Run [27053336290](https://github.com/max-sixty/worktrunk/actions/runs/27053336290).** Backgrounded `cargo run -- hook pre-merge --yes` at 05:20:00Z — the SDK created a tracked task (output at `…/<session-id>/tasks/b4uib1zkg.output`). The model then polled with `sleep 90; tail`, wrote `/tmp/pr-body.md`, re-read the task output, looked up `Monitor` via `ToolSearch:select:Monitor` — and emitted `end_turn` without ever calling `Monitor`. No active wait, no wakeup slot.

**Run [27075856123](https://github.com/max-sixty/worktrunk/actions/runs/27075856123).** Same shape, but the model *did* call `Monitor` — at 22:51:05.138 it issued `Monitor({})` with empty input, and got back `InputValidationError: required parameter 'description' is missing` (`Monitor` is a deferred tool — its schema must be loaded via `ToolSearch:select:Monitor` *before* the call, not after). At 22:51:08 it loaded the schema, but never retried `Monitor` — straight to `end_turn` at 22:51:23 with text claiming "the full test suite is running as a tracked background task and I'll be re-invoked when it finishes". It was not.

So: neither "wakeup window too short" nor "final `end_turn` the SDK didn't queue a wakeup for". It's **the model never attached the subscription** in the first place. `claude-code-action`'s `task-notification` wakeup ([action.yaml:321-324](https://github.com/max-sixty/tend/blob/26e7e3a/action.yaml#L321-L324)) appears to be tied to a successful `Monitor` call subscribing to the task — `end_turn` without an active wait reads to the SDK as a clean exit regardless of pending background Bash. (One unverified detail: I can't tell from the logs whether the SDK would *also* drop the wakeup after some bounded window if a wait were active. That's the layer below where this rule needs to bite.)

**Re: generalising to "do not end the turn until everything is complete"** — yes, that's the right framing. Even when `Monitor` is invoked correctly the attach is fragile (deferred-tool schema dance + correct args + reliable wakeup window). The robust rule cuts through all of that: don't `end_turn` while the deliverable is staged but un-shipped, regardless of *why* it's un-shipped. I'll rewrite the section to a top-level **End the turn only when work is shipped** that applies to PR creation, comment posting, follow-up commits — anywhere the maintainer-visible artefact is gated on something the model is waiting on. The existing `run_in_background: true` recipe under **CI Monitoring** stays intact (it runs *after* push, so the deliverable is already shipped — worst case is a missed CI-result follow-up, not a lost PR).

Pushing the rewrite in a follow-up commit on this branch.

PR #664, opened 2026-06-07
interactive: pass GitHub Actions context env vars to sandbox

tend-agent:
## Problem

Every skill running under the **interactive** harness sees empty `$GITHUB_RUN_ID`, `$GITHUB_REPOSITORY`, `$GITHUB_WORKFLOW`, etc. — the sandbox launch in [`interactive/action.yaml`](https://github.com/max-sixty/tend/blob/main/interactive/action.yaml) uses `sudo -u "$SANDBOX" env VAR=...`, which replaces the environment with only what's explicitly listed. The standard GitHub Actions context vars are not in the list, so they reach the agent's bash as empty strings.

Many skills reference these vars directly:

- `running-in-ci/SKILL.md`: `$GITHUB_RUN_ID` (self-check-run dedup in CI polling), `$GITHUB_WORKFLOW`, `$GITHUB_REPOSITORY` (artifact lookup, owner-correct URLs), `git worktree add … -b skills/<topic>-$GITHUB_RUN_ID`
- `review-reviewers/SKILL.md` and `review-runs/SKILL.md`: `grep -qF "$GITHUB_RUN_ID" /tmp/findings.md` guard, `hourly/review-$GITHUB_RUN_ID` / `daily/review-runs-$GITHUB_RUN_ID` branch names
- `notifications/SKILL.md`: `repos/$GITHUB_REPOSITORY/actions/runs` API call, `$GITHUB_RUN_ID` self-filter
- `triage/SKILL.md`: `repos/$GITHUB_REPOSITORY/issues/...` API call

When `$GITHUB_RUN_ID` is empty, the defense-in-depth guard `grep -qF "$GITHUB_RUN_ID" /tmp/findings.md` becomes `grep -qF ""`, which matches any non-empty file and silently passes. The skill then PATCHes the gist with whatever heading the LLM hallucinated for the run ID.

## Evidence

Direct mechanistic confirmation in this run: probing `echo "GITHUB_RUN_ID=$GITHUB_RUN_ID"` in the agent's bash returns `GITHUB_RUN_ID=` (empty). Same for `GITHUB_REPOSITORY`, `GITHUB_WORKFLOW`, `GITHUB_RUN_ATTEMPT`.

Visible outcome — a fabricated `## Run 27075391434 — 2026-06-06T22:16:36Z` heading in the [evidence gist](https://gist.github.com/feadfa947b23ed5cbf412ac036905894#file-findings-md) for `max-sixty/cargo-affected`. The run ID `27075391434` doesn't exist (404 on both `repos/max-sixty/tend/actions/runs/27075391434` and the cargo-affected equivalent). The actual review-reviewers run for that matrix leg was `27075279216` at 22:11:18Z. The fabricated heading's `Workflow:` link points at the dead run ID, leaving a broken audit trail entry that prior runs use to count cumulative occurrences.

The carry observation logged in the prior run (`27076519178`) flagged this and set a 2/2 threshold; this run is the second observation. Source: [evidence gist](https://gist.github.com/feadfa947b23ed5cbf412ac036905894).

## Fix

Add the standard GitHub Actions context vars to the `sudo env` allowlist in `interactive/action.yaml`. These are public values (not secrets) that the agent's skills already assume are present.

```diff
+          GITHUB_ACTIONS="$GITHUB_ACTIONS" CI="${CI:-true}" \
+          GITHUB_RUN_ID="$GITHUB_RUN_ID" GITHUB_RUN_ATTEMPT="$GITHUB_RUN_ATTEMPT" \
+          GITHUB_REPOSITORY="$GITHUB_REPOSITORY" GITHUB_REPOSITORY_OWNER="$GITHUB_REPOSITORY_OWNER" \
+          GITHUB_WORKFLOW="$GITHUB_WORKFLOW" GITHUB_SERVER_URL="$GITHUB_SERVER_URL" \
+          GITHUB_EVENT_NAME="$GITHUB_EVENT_NAME" GITHUB_SHA="$GITHUB_SHA" GITHUB_REF="$GITHUB_REF" \
```

Plus an in-place comment explaining why the GHA vars need re-passing (`sudo env` replaces the env, doesn't extend it).

## Gate assessment

- **Confidence**: High. Mechanism observed directly (empty env vars in this run) and an outcome failure observed in the gist (fabricated run heading). Structural failure — same conditions produce the same failure every interactive run; only the LLM's stochastic ability to derive the correct ID from `gh run list` masks it. The Claude (non-interactive) and Codex harnesses don't have this issue because they don't isolate via `sudo env`.
- **Magnitude**: Targeted fix — a single edit to one file, adding well-known env vars to a passthrough list. Affects every interactive-harness skill, which is appropriate given that every skill references these vars.
- **Both gates pass** at High evidence (2 observations: outcome + mechanism) × Targeted fix.

## What's out of scope

Not touching the skill-level `grep -qF "$GITHUB_RUN_ID"` guard. With the env var actually populated, the guard works as designed. Strengthening it (e.g. `[ -n "$GITHUB_RUN_ID" ] || exit 1`) is defense-in-depth but adds churn for a hole that's now closed at the source.

Not touching the `claude-code-action`-based action.yaml — that harness doesn't strip env vars and isn't affected.


interactive/action.yaml
```diff
@@ -534,6 +545,11 @@ runs:
           BOT_NAME="$BOT_NAME" BOT_ID="$BOT_ID" \
           TEND_MODEL="$TEND_MODEL" TEND_ALLOWED_TOOLS="$TEND_ALLOWED_TOOLS" \
           TEND_SYSTEM_PROMPT="$TEND_SYSTEM_PROMPT" TEND_PROMPT="$TEND_PROMPT" \
+          GITHUB_ACTIONS="$GITHUB_ACTIONS" CI="${CI:-true}" \
+          GITHUB_RUN_ID="$GITHUB_RUN_ID" GITHUB_RUN_ATTEMPT="$GITHUB_RUN_ATTEMPT" \
+          GITHUB_REPOSITORY="$GITHUB_REPOSITORY" GITHUB_REPOSITORY_OWNER="$GITHUB_REPOSITORY_OWNER" \
+          GITHUB_WORKFLOW="$GITHUB_WORKFLOW" GITHUB_SERVER_URL="$GITHUB_SERVER_URL" \
+          GITHUB_EVENT_NAME="$GITHUB_EVENT_NAME" GITHUB_SHA="$GITHUB_SHA" GITHUB_REF="$GITHUB_REF" \
```
tend-agent: Add `GITHUB_EVENT_PATH` here — `review/SKILL.md` reads it to detect `ready_for_review` transitions; without it, that pre-flight bypass silently breaks under the interactive harness.

```suggestion
          GITHUB_EVENT_NAME="$GITHUB_EVENT_NAME" GITHUB_EVENT_PATH="$GITHUB_EVENT_PATH" \
          GITHUB_SHA="$GITHUB_SHA" GITHUB_REF="$GITHUB_REF" \
```

tend-agent (commented):
One coverage gap: `GITHUB_EVENT_PATH` isn't in the new allowlist, but `plugins/tend-ci-runner/skills/review/SKILL.md:33` reads it to compute `EVENT_ACTION` (used to bypass the same-SHA exit when a PR transitions out of draft via `ready_for_review`). With it empty under interactive, that detection silently falls back to `/dev/null` and a draft-to-ready transition won't trigger the intended full review. The file lives under `RUNNER_TEMP` (runner-owned but readable to the sandbox user — same as the rest of `_temp`), so passing the path through should work.

interactive/action.yaml
tend-agent: Good catch — added in 0dc3721. Verified `GITHUB_EVENT_PATH` is the only other GHA context var referenced across the bundled and overlay skills, so no other gaps.

max-sixty:
should we add everything `GITHUB_*` apart from the token?

tend-agent:
Mostly yes, but `GITHUB_TOKEN` isn't the only one to exclude. The runner-write file paths are command-injection channels into the workflow's subsequent steps — the sandbox writing to them lands lines that the runner processes after our step exits:

- `GITHUB_ENV` — appended `K=V` lines become env vars in later steps
- `GITHUB_PATH` — appended lines get prepended to `PATH` in later steps
- `GITHUB_OUTPUT` — sets step outputs
- `GITHUB_STATE` — saves state for the post step
- `GITHUB_STEP_SUMMARY` — markdown rendered in the job-summary UI

On current runners these files are owned by the runner user and the sandbox UID likely can't write them, but the value points at the path either way and the mode isn't a guarantee we control. Conservative default is to keep them out of the passthrough.

Everything else under `GITHUB_*` is informational (`GITHUB_ACTOR`, `GITHUB_API_URL`, `GITHUB_REF_NAME`, `GITHUB_WORKSPACE`, …) and safe to pass. A small denylist also picks up future additions automatically:

````bash
GHA_VARS=()
while IFS= read -r name; do
  case "$name" in
    GITHUB_TOKEN|GITHUB_ENV|GITHUB_PATH|GITHUB_OUTPUT|GITHUB_STATE|GITHUB_STEP_SUMMARY) continue ;;
  esac
  GHA_VARS+=("$name=${!name}")
done < <(compgen -e | grep '^GITHUB_')

sudo -u "$SANDBOX" env \
  HOME="$AGENT_HOME" \
  …existing entries… \
  "${GHA_VARS[@]}" \
  script -qefc "$WRAPPER" "$LOG" &
````

Want me to push that shape onto this branch?

max-sixty:
OK. I don't exactly get how passing `GITHUB_STEP_SUMMARY` is a security risk, but OK

tend-agent:
Fair pushback — `GITHUB_STEP_SUMMARY` is the weakest of the five and isn't really a security risk on its own. The summary file is rendered markdown after the job finishes; writes there don't mutate later steps' env, PATH, outputs, or post-step state the way the other four do, so it isn't a privilege-escalation channel into the runner user.

The reason I left it on the denylist is integrity, not security: the supervisor itself writes the transcript tail, the token-usage table, and the skill step-summary into `$GITHUB_STEP_SUMMARY`, and if the sandbox can append there it can splice content into the supervisor's own audit trail (or fill the 1 MiB cap and squeeze the supervisor's appends out). I'd rather the summary be one trusted writer than two.

Happy to drop it if you'd rather have parity with `GITHUB_*` everywhere — let me know.

Pushed the denylist switch as [`1500531`](https://github.com/max-sixty/tend/pull/664/commits/1500531). The two earlier explicit-allowlist commits in this PR (`291249e`, `0dc3721`) are now superseded but kept for review-history continuity; happy to squash on merge.

PR #680, opened 2026-06-10
skills(install-tend): restructure bot-token step around scope-audit remediation

max-sixty:
Step 8 of the install-tend skill previously had a single path: a full `gh auth login` device flow, written for first-time setup and run by the user. This PR restructures it around what the step actually does (check what gh already stores for the bot, mint a token only if needed, push it to the repo secret) so one procedure serves both fresh installs and the `Bot PAT` scope-audit remediation that nightly audit issues link to.

What changed:

- **State-driven routing**: a keyring check routes to "stored token already good, skip to the push", "present but under-scoped, `gh auth refresh`" (new 8a: preserves existing scopes and hard-fails if the wrong browser account approves, so no post-hoc identity check is needed), or "absent, `gh auth login`" (8b, the old path with its verify step).
- **One env-var rule, stated by mechanism**: commands that gh's `GH_TOKEN`/`GITHUB_TOKEN` precedence breaks carry the `env -u` prefix (login/refresh/switch/logout and `gh api user`); env-indifferent commands run bare.
- **Hardened secret push (8c)**: the old `gh auth token | gh secret set` pipe could silently blank a working secret. `gh secret set` accepts an empty body (cli/cli `set.go` reads stdin with a bare `io.ReadAll`) and the pipe's exit status hides the failed read, while the step's freshness check still passes. Now: capture, emptiness guard, `--body`.
- **Agent-run device flow**: the agent runs the flow in the background and surfaces the one-time code; delegating to the user's terminal is the fallback, with the credential-store consequence spelled out (subsequent `gh auth token --user` reads must run in that same terminal). Device-code expiry (~15 minutes) documented.
- **Audit-issue accuracy**: the nightly skill's scope-audit instruction now requires naming the repo's actual configured secret (`secrets.bot_token`, default `TEND_BOT_TOKEN`) rather than improvising a placeholder. A real audit issue invented `BOT_GITHUB_TOKEN`, which cost remediation time before the real secret name was identified.

The gh behavioral claims were verified against cli/cli source: the env-token guard shared by login/refresh/switch/logout, refresh's wrong-approver rejection and scope merging, the `(keyring)` suffix in `gh auth status` output, `TokenForUser` ignoring env vars, and `gh secret set` accepting an empty body.

Supersedes the unmerged `install-tend-refresh-shortcut` branch (093000b), whose refresh-path change is incorporated here.


tend-agent (commented):
Verified the gh behavioral claims against cli/cli v2.92.0 — the env-token guard fires for `login`/`refresh`/`switch`/`logout`, refresh merges existing scopes and rejects mismatched users with "received credentials for …", and `gh secret set`'s `getBody` blindly takes `io.ReadAll(stdin)` with no emptiness check. Routing logic and `(keyring)` source detection match status.go's output format.

One robustness note inline on 8c.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -552,45 +552,112 @@ The bot's token needs scopes `repo`, `workflow`, `notifications`,
...
+this over any name an audit issue quotes) and verify the `Updated`
+timestamp is fresh. The emptiness guard matters:
+`gh secret set` accepts an empty body, so piping a failed keyring
+read straight in would silently blank a working secret.
 
 ```bash
-gh auth switch --user <maintainer>
-gh auth token --user <bot-name> | gh secret set TEND_BOT_TOKEN --repo "$REPO"
+BOT_GH_TOKEN=$(gh auth token --user <bot-name>)
+[ -n "$BOT_GH_TOKEN" ]  # stop here if the keyring read failed
```
tend-agent: The bare test only halts under `set -e` (or if the agent runs each line as a separate Bash invocation). When the block runs as a single shell process without `set -e` — which is the typical Bash-tool invocation — `[ -n "" ]` just sets `$?=1` and `gh secret set` runs next with an empty `--body`, which is the exact silent-blank failure mode the comment above warns about. Chaining the guard with `||` so it actually short-circuits the push makes the protection independent of the caller's shell options:

```suggestion
[ -n "$BOT_GH_TOKEN" ] || { echo "bot token empty — refusing to push" >&2; exit 1; }
```

max-sixty: Applied in 90e36ef — guard now chains `|| { echo ... >&2; exit 1; }` so it short-circuits regardless of shell options.

tend-agent (approved):

PR #683, opened 2026-06-11
interactive: install agent toolchain as the sandbox user

max-sixty:
The `claude-interactive` harness runs the agent as a non-sudo `tend-sandbox` user for credential isolation, so its toolchain has to live in `/home/tend-sandbox`. It got there by installing everything as the `runner` user and then `sudo cp -a`-ing `~/.local/bin`, `~/.local/share`, `~/.claude`, and `~/.claude.json` into the sandbox home, followed by a recursive `chown`. That ran on every invocation (review, mention, triage, ci-fix) and re-copied the ~200 MB `claude` binary each time.

This installs the toolchain directly as `tend-sandbox`, so it lands in `/home/tend-sandbox` natively: no copy, no chown, no hand-off step. (`mv` of a runner-side install was a dead end. The `claude` binary embeds absolute-path references into `~/.local/share`, so relocating it breaks the launcher.) Each install step now runs its body under `sudo -u tend-sandbox env HOME=/home/tend-sandbox ... bash`. The native install is also more correct than the copy: `claude`'s plugin index stores absolute paths, so it writes `/home/tend-sandbox/...` instead of the stale `/home/runner/...` paths the copy left behind.

## What changed

`interactive/action.yaml`: the claude-binary, onboarding-pre-seed, and plugin-install steps now run as the sandbox user, and the "Hand off Claude toolchain to sandbox" step is gone. A new step installs `uv` for the sandbox, since the agent's skills run `uvx tend@latest` (the bundled `nightly` skill) and `uv` previously rode along in the copied `~/.local/bin`.

`interactive/proxy/setup-sandbox.sh`: grants the sandbox read access to the action checkout so `claude plugin marketplace add` can read it once at install time (public repo content, no secret). The grant is skipped in the local `uses: ./` case, where the sandbox already owns the workspace.

## Testing

`ci.yaml` covers the proxy unit tests and lint but does not exercise `interactive/action.yaml`. The end-to-end validation is the `interactive-smoke` workflow (`uses: ./interactive` on this branch), triggered against this branch.


tend-agent (commented):
The new "Install uv (as sandbox user)" step fails the smoke test [run on `7de6203`](https://github.com/max-sixty/tend/actions/runs/27348550745):

```
sh: 586: cannot create /home/runner/.config/uv/uv-receipt.json: Permission denied
```

uv's `install.sh` computes the receipt path as `RECEIPT_HOME="${XDG_CONFIG_HOME:-$INFERRED_HOME/.config}/uv"` (around line 125). The GitHub Actions Ubuntu runner pre-sets `XDG_CONFIG_HOME=/home/runner/.config` in the workflow env, and that value passes through `sudo -u tend-sandbox env HOME=…` to the sandbox bash — so the install writes to `/home/runner/.config/uv/`, which the sandbox can't write.

The `HOME="$AGENT_HOME"` override only retargets `HOME`; the XDG var still leaks. Inline suggestion overrides it explicitly.

Same latent risk in the Install Claude Code, Pre-seed onboarding, and Install tend plugins steps, plus the `Run Claude` sudo env: none currently write to `$XDG_CONFIG_HOME`, so they don't trip today, but a future tool that uses XDG paths will. Worth considering whether `XDG_CONFIG_HOME` (and `XDG_CACHE_HOME`/`XDG_DATA_HOME` for symmetry) should be overridden once at the sandbox boundary rather than per-step.

interactive/action.yaml
```diff
@@ -326,8 +342,25 @@ runs:
...
+    # skill runs `tend check` / `init`), so the sandbox needs uv on its PATH.
+    # It used to arrive via the toolchain hand-off copy of ~/.local/bin; with
+    # that gone, install uv directly as the sandbox user. The runner's own uv
+    # ("Ensure uv is available", for the proxy) is a separate install.
+    - name: Install uv (as sandbox user)
+      shell: bash
+      run: |
+        sudo -u "$SANDBOX" env \
+          HOME="$AGENT_HOME" \
+          UV_NO_MODIFY_PATH=1 \
```
tend-agent, line 366:
```suggestion
          XDG_CONFIG_HOME="$AGENT_HOME/.config" \
          UV_NO_MODIFY_PATH=1 \
```

With `XDG_CONFIG_HOME` overridden, uv's `install.sh` writes its receipt to `/home/tend-sandbox/.config/uv/uv-receipt.json` instead of leaking back to `/home/runner/.config/uv/`.

tend-agent (approved):

PR #685, opened 2026-06-11
weekly: reseed integration-fixture secrets outside the sandbox

max-sixty:
The weekly integration test has failed two weeks running (ref #596), with two distinct causes that share one structural bug: the recipe exported credentials from inside the interactive harness's sandbox, where real credentials never exist.

The first week, `$CLAUDE_CODE_OAUTH_TOKEN` (absent in the agent subprocess) was piped as empty into the fixture's secret, breaking the action's auth preflight. The second week, after the credential proxy (#652) shipped in 0.1.3, the reseed exported the proxy's placeholder PAT (`ghp_tendproxydummy…`) into the fixture's `TEND_BOT_TOKEN`; checkout on tend-agent/tend-integration then failed basic auth with the "could not read Username" 401 symptom. Every in-sandbox probe of the placeholder passes (`gh auth status`, clones, length checks) because the proxy injects the real token on GitHub hosts, so the exporting agent had no way to notice. Confirmed by reproducing the 401 with the placeholder locally, and by the same checkout SHA + token succeeding on this repo's nightly.

The fix moves secret propagation outside the sandbox:

- **New `integration-secrets` workflow** (`workflow_dispatch` only): copies this repo's real `TEND_BOT_TOKEN` and `CLAUDE_CODE_OAUTH_TOKEN` into the fixture, with fail-fast guards against empty source values.
- **Recipe** (`running-tend/references/integration-test.md`): the in-sandbox reseed is gone. The safety preamble states that env credentials are placeholders and must never be exported; §1 dispatches `integration-secrets` every run and waits for it.
- **§5 becomes self-healing** (option 1 from #596): the version pin moves on every release and the fixture has no other regeneration path, so §5 now pushes the regenerated files to the fixture's `main` and asserts that `init` succeeds and is idempotent, rather than failing on any diff.

Trade-offs: the weekly now depends on the dispatch succeeding (an Actions hiccup fails the run, loudly), and §5 no longer asserts generated content (the published generator is already validated by this repo's CI; the fixture is not a release gate).

After merge, the next weekly run heals the fixture end to end: §1 reseeds both broken secrets, §3/§4 exercise triage and review, §5 closes the 0.1.0 → 0.1.3 drift. No manual steps.

### DIFF
--- a/.claude/skills/running-tend/references/integration-test.md
+++ b/.claude/skills/running-tend/references/integration-test.md
@@ -280,7 +280,8 @@ if [ -n "$(git status --porcelain)" ]; then
   gh auth setup-git
   git add .
   git commit -m "chore: regenerate tend workflows (weekly integration self-heal)"
-  git push origin main
+  git push origin main \
+    || { echo "tend-integration: push to main failed; fixture not updated"; exit 1; }
 fi
 uvx tend@latest init
 [ -z "$(git status --porcelain)" ] \
--- a/.claude/skills/running-tend/references/integration-test.md
+++ b/.claude/skills/running-tend/references/integration-test.md
@@ -7,7 +7,8 @@ the next week.
 ## Safety — read first
 
 This recipe issues destructive operations (close issues, close PRs,
-delete branches, overwrite secrets) against **exactly one** repo:
+delete branches, push regenerated workflows to `main`) against
+**exactly one** repo:
 `tend-agent/tend-integration`. The literal string `tend-agent/tend-integration`
 appears as `--repo` argument on every destructive call below — not a
 variable. **Do not substitute a variable**, do not rename the repo, do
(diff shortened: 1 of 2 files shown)


.claude/skills/running-tend/references/integration-test.md
```diff
@@ -246,21 +261,30 @@ cd - >/dev/null
...
-if ! git diff --quiet .github/workflows/; then
-  echo "tend-integration drift: $(git diff --stat .github/workflows/)"
-  exit 1
+if [ -n "$(git status --porcelain)" ]; then
+  git config user.email "tend-agent@users.noreply.github.com"
+  git config user.name "tend-agent"
+  gh auth setup-git
+  git add .
+  git commit -m "chore: regenerate tend workflows (weekly integration self-heal)"
+  git push origin main
```
tend-agent:
```suggestion
  git push origin main \
    || { echo "tend-integration: push to main failed; fixture not updated"; exit 1; }
```

tend-agent (commented):
One thing on §5: the `git push origin main` is the only credentialed write to the fixture, but it's unguarded. If it fails (transient network, a race that takes main out from under us), bash continues without `set -e`. The second `uvx tend@latest init` then runs against the local tree (which already committed the regeneration), reports clean, and §5 exits 0 — but the fixture's `main` is still at the old version, so next weekly sees the same drift. The recipe's pattern in §3/§4 is explicit `|| { echo …; exit 1; }`; suggested inline.

tend-agent (approved):

.claude/skills/running-tend/references/integration-test.md
max-sixty: Applied in 51cd47a.

PR #686, opened 2026-06-11
interactive: isolate the Anthropic credential behind the credential proxy (phase 2)

max-sixty:
## What

Phase 2 of the `claude-interactive` credential proxy: isolate the **Anthropic** credential (OAuth token or API key) the same way phase 1 isolated the GitHub PAT. The sandboxed agent now holds only dummies for *both* secrets; the runner-owned mitmproxy injects the real ones per-host. This closes the last credential the prompt-injectable agent could read from its own environment.

Phase 1 (GitHub-only) shipped in #652. This branch extends the mechanism to `api.anthropic.com`, installs the agent toolchain directly as the sandbox user (dropping a ~200 MB `cp -a`), and hands the workspace back to the runner so checkout's post-job cleanup doesn't warn.

## How it works

`claude` runs in the same auth mode as production but with a scheme-matched dummy secret, so it emits the correct headers (`Authorization: Bearer …` for OAuth, `x-api-key: …` for API key). The addon (`interactive/proxy/inject_credentials.py`) swaps the dummy value for the real one on requests to `api.anthropic.com`, exactly as it already does for GitHub hosts. The real secret lives only in the proxy process (a different, non-sudo UID can't read its `/proc/<pid>/environ`).

Two details that took live iteration to get right:

- **`claude` is Node**, which ignores the system trust store, so the agent trusts the proxy CA via `NODE_EXTRA_CA_CERTS` (`HTTPS_PROXY` + `NODE_EXTRA_CA_CERTS` is an officially supported Claude Code intercepting-proxy config).
- **mitmproxy buffers responses by default**, which would stall every SSE inference turn. A `responseheaders` hook sets `flow.response.stream = True`. (`--set stream_large_bodies` is the wrong tool — it streams large *request* bodies before the request hook fires, silently disabling injection.)

## Reviewer's map

- `interactive/proxy/inject_credentials.py` — the addon (renamed from `github_auth.py`). The host frozensets are the credential boundary; injection gates on `flow.request.host` (the real connection target), never the spoofable `pretty_host`.
- `interactive/proxy/setup-sandbox.sh` — creates the sandbox user, neutralizes the persisted git credential, starts the proxy, and writes the shared agent-env file consumed by every sandbox step.
- `interactive/action.yaml` — install-as-sandbox, the PTY supervisor, and the workspace chown-back.
- `interactive/proxy/test_inject_credentials.py` — 21 unit tests (GitHub + Anthropic, both auth modes, host spoofing, mixed-case, cleartext, SSE streaming, startup guards).
- `docs/security-model.md` — the updated credential-isolation paragraph.

This branch also reconciled main's #683 (install-as-sandbox, superseded by this branch's superset) and #684 (`raw.githubusercontent.com` injection, folded in).

## Validation

- 21 addon unit tests + the generator suite green; pre-commit clean.
- End-to-end via `interactive-smoke` (manual `workflow_dispatch`, not part of PR CI): a dummy-credential sandbox agent authenticated to GitHub *and* completed multi-turn inference (incl. a long SSE-streamed essay) through the proxy, with the in-transcript isolation check reporting `anthropic-cred: isolated`. The four follow-up cleanups in the top commit are comment/inert changes that don't alter the smoke-validated path.

The notable scoping decision: Claude Code strips its own auth vars from tool subprocesses unconditionally, so the agent's `Bash` sees them `unset` — the isolation proof is "inference completes with only a dummy in `claude`'s own env," not "the dummy is visible in Bash."


tend-agent (commented):
The "Long-lived PAT exposure" section at the bottom of `docs/security-model.md` (lines 184–187) still reads as if phase 2 didn't happen:

> The credential isolation above keeps the PAT out of the agent on the `claude-interactive` harness; it remains directly exposed on the other harnesses and the Claude token on all of them.

The trailing "Claude token on all of them" claim contradicts the updated paragraph at lines 94–109, which now says both the PAT and the Anthropic credential are isolated on interactive. Suggested fix:

```diff
-above keeps the PAT out of the agent on the `claude-interactive` harness; it
-remains directly exposed on the other harnesses and the Claude token on all of
-them.
+above keeps both the PAT and the Claude token out of the agent on the
+`claude-interactive` harness; both remain directly exposed on the other two
+harnesses.
```

Outside the diff so I'm not posting it as an inline suggestion — happy to push as a follow-up if you'd prefer.

Optional nit: the bwrap mention at lines 173–178 still frames model-auth removal as something the bwrap workaround would buy us, without qualifying that on interactive, phase 2 already mitigates it. The TODO.md entry in this PR is now more current on this than the security-model.md paragraph.

tend-agent (approved):

PR #689, opened 2026-06-12
skills(nightly): skip stamp-only workflow regen PRs

tend-agent:
Adds a stamp-only check to Step 7 of the nightly skill: when the workflow regen produces a diff whose only changes are the `# Generated by tend X.Y.Z` header in each `tend-*.yaml`, treat it as a no-op and skip the PR.

This case fires when dependabot has already bumped the action ref in a patch release (e.g. `max-sixty/tend@0.1.3` → `0.1.4`) before nightly runs `uvx tend@latest init`. The workflows already reference the new version, so the only thing the regen changes is the comment header in each file. The resulting PR carries no value — the existing workflows are still accurate.

Triggered by @max-sixty on [numbagg/numbagg#644](https://github.com/numbagg/numbagg/pull/644) and [numbagg/numbagg#645](https://github.com/numbagg/numbagg/pull/645):

> there is no point bumping _only_ the stamp; the existing version is still accurate. it should ignore these if there no workflow change if dependabot has already bumped the skills, especially for patch versions

> this should go into tend ...and there should be guidance for tend so future similar versions go into tend.


tend-agent (commented):
The skip-check is non-functional as written — the regex never matches the actual stamp lines, so `NON_STAMP_DIFF` is always non-zero on a stamp-only diff and the PR is still opened.

The stamp line in every generated workflow is `# Generated by tend X.Y.Z. Regenerate with: uvx tend@latest init` (see `generator/src/tend/workflows.py` — search for `HEADER = f"""`), but the filter anchors `$` immediately after the version number. Reproduced locally: feeding a synthetic header-only diff through the pipeline returns `2`, not `0`. Suggestion inline extends the regex to the full stamp line.

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -219,9 +219,20 @@ uvx tend@latest init
...
 git status --porcelain .github/workflows .config
+
+# Stamp-only check: if the only diff is the `# Generated by tend X.Y.Z`
+# header (e.g. dependabot has already bumped the action refs in a patch
+# release), the workflow bodies are unchanged and the existing files are
+# still accurate. A header-only PR carries no value — treat it as a no-op.
+NON_STAMP_DIFF=$(git diff --no-color .github/workflows .config \
+  | grep -E '^[+-]' \
+  | grep -vE '^(\+\+\+|---) ' \
+  | grep -vE '^[+-]# Generated by tend [0-9]+\.[0-9]+\.[0-9]+$' \
```
tend-agent: The `$` anchors after the version number, but actual stamp lines end with `. Regenerate with: uvx tend@latest init` — so this `grep -v` never matches and the stamp-only filter is a no-op. Extend the pattern to the full stamp line:

```suggestion
  | grep -vE '^[+-]# Generated by tend [0-9]+\.[0-9]+\.[0-9]+\. Regenerate with: uvx tend@latest init$' \
```

tend-agent: Fixed in 9bdbd2e — extended the pattern to the full stamp line. Verified with a synthetic header line: original returned 1, new pattern returns 0.

PR #691, opened 2026-06-14
skills(running-tend): install uv into sandbox before integration-test §1/§5

tend-agent:
The weekly integration-test recipe shells out to `uvx tend@latest init` in §1 (bootstrap) and §5 (self-heal). The Claude agent runs as `tend-sandbox` inside the harness sandbox, but the workflow's `astral-sh/setup-uv` step installs uv under the runner user's `$HOME` (`/home/runner/.local/bin`), which the sandbox's PATH doesn't include. The result: §5's `uvx tend@latest init` exits non-zero with no output captured, `git status --porcelain` reports clean (nothing was regenerated, nothing was modified), and the idempotency assertion reads that as a pass.

Caught during this week's run — §5 reported PASSED while doing nothing. Manual `curl https://astral.sh/uv/install.sh | sh` into the sandbox made `uvx` reachable and §5 then exercised init for real.

Adds a §0 sandbox-prerequisites section that installs uv into the sandbox if `command -v uvx` is empty, and aborts loudly if uvx is still unreachable afterward. The installer's default destination (`$HOME/.local/bin`) is already on the sandbox PATH, so no `export PATH` is needed for later sections to inherit it.

This is an invisible-failure-mode fix: without it, §5 silently passes on every weekly run without exercising the generator.

Surfaced during the weekly integration test on [run 27495976519](https://github.com/max-sixty/tend/actions/runs/27495976519).

### DIFF
--- a/.claude/skills/running-tend/references/integration-test.md
+++ b/.claude/skills/running-tend/references/integration-test.md
@@ -41,6 +41,28 @@ in place.
 Run steps in order. If §3, §4, or §5 fails, jump to §6 (reset), then §7
 (report).
 
+## 0. Sandbox prerequisites
+
+§1 (bootstrap) and §5 (self-heal) shell out to `uvx tend@latest init`.
+The agent runs as `tend-sandbox` inside the harness sandbox, but the
+workflow's `astral-sh/setup-uv` step installs uv under the runner user's
+home, which the sandbox can't see. Without uv on the sandbox PATH §5
+silently no-ops — the missing-binary error is swallowed and an empty
+`git status` reads as "idempotent" — so install uv into the sandbox
+before continuing:
+
+```bash
+if ! command -v uvx >/dev/null 2>&1; then
+  curl -LsSf https://astral.sh/uv/install.sh | sh
+fi
+command -v uvx >/dev/null \
+  || { echo "uvx unavailable after install; aborting recipe"; exit 1; }
+```
+
+The installer drops uv into `$HOME/.local/bin`, which is already on the
+sandbox PATH, so no `export PATH` is needed for later sections to pick
+it up.
+
 ## 1. Bootstrap (first run only) and reseed (every run)
 
 Create the test repo if missing, with workflows installed on `main`


max-sixty:
should we install into our environment in the workflow, rather than having the agent do it?

tend-agent:
Agreed — and that's already the structural fix on `main`. PR #683 ("[interactive: install agent toolchain as the sandbox user](https://github.com/max-sixty/tend/pull/683)") added an [Install agent toolchain (sandbox)](https://github.com/max-sixty/tend/blob/e8688ed5b6c96163fd2b3a442d16f7141aee3325/interactive/action.yaml#L327) step that installs uv (and the `claude` binary) directly under `tend-sandbox`, replacing the old "install as runner, then `cp -a ~/.local/bin` into the sandbox" handoff. The skill recipe shouldn't need to know how the harness gets uv into the sandbox.

So this PR is really a one-release stopgap for `tend-weekly.yaml` still pinning `max-sixty/tend/interactive@0.1.4`, where uv only reaches the sandbox if it was installed under the runner's `~/.local/bin` first. The v0.1.4 "Ensure uv is available" step short-circuits when `command -v uvx` already resolves — and the workflow's prior `astral-sh/setup-uv@v6` step satisfies that via the hosted toolcache, never touching `~/.local/bin`. The runner's `~/.local/bin` then stays empty, so the `cp -a` carries nothing useful into `/home/tend-sandbox/.local/bin`, and the agent's `uvx tend@latest init` fails with `command not found` — silently, since §5's idempotency check reads an empty `git status --porcelain` as a pass.

I'd suggest closing this in favor of the next release. Two practical options:

1. **Close and cut a release.** Once `0.1.5` ships, `tend-weekly.yaml` bumps to that ref (the nightly already updates the pin) and the sandbox install is structural.
2. **Land this as belt-and-braces.** The added §0 is a `command -v uvx` check; once the action installs uv natively, the if-branch is a no-op and the recipe stays tidy. Cheap, harmless, removable later.

I'll defer to whichever you prefer.

<details><summary>Evidence — the v0.1.4 failure path</summary>

From the session log on [run 27495976519](https://github.com/max-sixty/tend/actions/runs/27495976519):

- First §5 attempt (sandbox, default launch PATH):
  ```
  /usr/bin/bash: line 28: uvx: command not found
  --- status after first init ---
  /usr/bin/bash: line 40: uvx: command not found
  --- status after second init ---
  §5 PASSED
  ```
- Diagnostic after the recovery install:
  ```
  ls /home/runner/.local/bin/uvx → "no uvx in runner home"
  ls /home/tend-sandbox/.local/bin/uvx → /home/tend-sandbox/.local/bin/uvx
  PATH=/home/tend-sandbox/.local/bin:/usr/local/bin:/usr/bin:/bin:…
  ```

The runner's `~/.local/bin` not containing `uvx` is what's load-bearing — confirms `astral-sh/setup-uv@v6` short-circuited the v0.1.4 action's runner-side install, leaving the subsequent sandbox copy with no uv.

</details>

PR #693, opened 2026-06-14
skills: soften over-stringent guidance into examples and open frames

max-sixty:
Follow-up to the triage reply-templates change (#692). A fan-out audit of tend's bot guidance — 16 skill/prompt files, each finding adversarially verified — found that the triage-template problem was systemic. Two anti-patterns recur across the skills, both giving a good median output but an awkward or incomplete tail:

- **Verbatim templates** the bot reproduces literally (canned comment/PR-body wording).
- **Closed enumerated lists** the bot reads as exhaustive ("respond to these categories", "cross-check these workflows").

This applies the 8 high/medium-value softenings the audit surfaced, and codifies the underlying principle in `CLAUDE.md` → "Authoring skills" so new skills don't reintroduce the pattern.

## What changed

- **`review-runs` Step 4** — cross-check applies to every tend workflow, not just the four listed. This was a real coverage gap: the bot could skip checking mention/notifications/weekly/review-reviewers.
- **`ci-fix` 3a** — transient-failure signals presented as examples to weigh, not a checklist to satisfy (the conservative default-to-durable backstop and repeat-occurrence escalation are unchanged).
- **`notifications` 4b** — framed as the fallback for any notification no dedicated workflow covered, not a closed category list (Step 3 scope/security and Step 4a freshness/dedup guards unchanged).
- **`review-reviewers` Step 2** — rejection signals are examples; every bullet and the corruption-scan recipe are kept verbatim.
- **`running-in-ci`** — the upstream-issue permission offer and the "'Likely' is a stop-sign" rule reframed around intent and semantic shape rather than fixed wording.
- **`nightly` Step 7, `weekly` Step 2** — canned PR-body strings become goal-stated summaries the model writes per-situation.

## Safety

Each site keeps its load-bearing guard; only the wording loosens. Safety/correctness gates (reproduce-before-fix), security rules, branch-protection, and exact commands were deliberately left rigid — the audit's verification pass dropped any finding that touched them. An independent review confirmed no guard was weakened and no cross-reference broken.

The audit's low-tier findings (17 more of the same two patterns, lower value) are a candidate fast-follow.

### DIFF
--- a/plugins/tend-ci-runner/skills/weekly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/weekly/SKILL.md
@@ -34,7 +34,7 @@ If no dependency PRs are open, note "0 dependency PRs to process" and continue t
    if [ "$LAST_APPROVAL_SHA" = "$HEAD_SHA" ]; then
      echo "Already approved on this commit; skipping."
    else
-     gh pr review <number> --approve --body "Automated dependency update — CI passing, no breaking changes."
+     gh pr review <number> --approve --body "<one line: package, bump type, CI status — what you checked>"
    fi
    ```
 4. If CI is failing, comment with the failure summary and skip
--- a/plugins/tend-ci-runner/skills/weekly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/weekly/SKILL.md
@@ -34,7 +34,9 @@ If no dependency PRs are open, note "0 dependency PRs to process" and continue t
    if [ "$LAST_APPROVAL_SHA" = "$HEAD_SHA" ]; then
      echo "Already approved on this commit; skipping."
    else
-     gh pr review <number> --approve --body "<one line: package, bump type, CI status — what you checked>"
+     # Compose a one-line review body naming the package, bump type, and what you
+     # checked — e.g. "ruff 0.13 → 0.14 (patch), CI green, no API changes".
+     gh pr review <number> --approve --body "$REVIEW_BODY"
    fi
    ```
 4. If CI is failing, comment with the failure summary and skip
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -270,6 +270,13 @@ When adding to or editing files in `plugins/tend-ci-runner/skills/` or
   expected to drift can be dated; a one-shot incident citation should not be.
 - **Prefer recipe over narrative.** A code block plus a one-sentence framing
   beats a multi-paragraph explanation.
+- **Examples over templates; open frames over closed menus.** For text the
+  agent *produces* (comments, PR bodies, summaries), give an example labeled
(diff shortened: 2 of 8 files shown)


plugins/tend-ci-runner/skills/weekly/SKILL.md
```diff
@@ -34,7 +34,7 @@ If no dependency PRs are open, note "0 dependency PRs to process" and continue t
    if [ "$LAST_APPROVAL_SHA" = "$HEAD_SHA" ]; then
      echo "Already approved on this commit; skipping."
    else
-     gh pr review <number> --approve --body "Automated dependency update — CI passing, no breaking changes."
+     gh pr review <number> --approve --body "<one line: package, bump type, CI status — what you checked>"
```
tend-agent: The new `<one line: package, bump type, CI status — what you checked>` lives inside a shell command's `--body` string — if the bot ever fails to substitute, the literal angle-bracketed instruction ships as the public review body. That's exactly the shape the running-in-ci placeholder rule guards against, but the pre-post grep guard (`PLACEHOLDER`, `<SHA>`, `<TODO`, `TBD`, `XXX`) won't catch this token, so a miss goes out silently.

The new CLAUDE.md bullet says "give an example labeled as such rather than verbatim wording to paste" — moving the example out of the command body matches that more cleanly:

```suggestion
     # Compose a one-line review body per PR — name the package, bump type, and what
     # you checked. Example: "ruff 0.13 → 0.14 (patch), CI green, no API changes."
     gh pr review <number> --approve --body "$REVIEW_BODY"
```

That keeps the example labeled-as-such, leaves a clear shell variable to fill, and the existing `$`-substitution failure modes are visible at compose time rather than after `gh` posts.

tend-agent (commented):
The audit framing is sound and the new CLAUDE.md principle is a good codification. One concern on the weekly change worth surfacing — flagged inline.

Smaller things I considered and decided to leave alone: the `<sole handler>` → `<fallback>` reframing in notifications widens semantic scope but the Step 3 scope and 4a freshness/dedup are still load-bearing; the dropped specific "5 prior runs" threshold in ci-fix loses precision but the repeat-occurrence escalation backstops it; the bold/unbolded split between the four example workflows and the four "same treatment" workflows in review-runs Step 4 reads slightly inconsistent but is workable as "examples + sweep".

plugins/tend-ci-runner/skills/weekly/SKILL.md
max-sixty: Good catch — fixed in 246389e: moved the example into a comment and pass a `$REVIEW_BODY` variable, so a missed substitution surfaces at compose time instead of shipping the literal token as the public body.

tend-agent (approved):

PR #697, opened 2026-06-17
interactive: bump pinned claude_version to 2.1.179

max-sixty:
The interactive harness installs the `claude` binary pinned to `claude_version`, which sat at `2.1.150`. That predates Opus 4.8 (which shipped in claude-code `2.1.154`), so the binary's `--model opus` alias resolved to Opus 4.7. A recent `tend-notifications` run on this harness confirmed it ran on `claude-opus-4-7`.

This bumps the pin to the current latest claude-code release (`2.1.179`) so `opus` resolves to 4.8, and drops the now-inaccurate "matches claude-code-base-action" note from the input description. The `running-tend` weekly task added in #696 keeps the pin current going forward.

### DIFF
--- a/interactive/action.yaml
+++ b/interactive/action.yaml
@@ -323,7 +323,7 @@ runs:
     # must be installed in place — installing as runner and moving breaks it.
     # uv is the agent's own (its skills run `uvx tend@latest …`); the runner has
     # a separate uv for the proxy. Installers fetch over the direct network (no
-    # proxy env here). Same pinned claude version as claude-code-base-action.
+    # proxy env here).
     - name: Install agent toolchain (sandbox)
       shell: bash
       env:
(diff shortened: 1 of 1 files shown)


tend-agent (commented):
One leftover outside the diff: the same "matches claude-code-base-action" rationale removed from the input description still sits in the comment above the `Install agent toolchain (sandbox)` step in `interactive/action.yaml` — the trailing sentence reads `Same pinned claude version as claude-code-base-action.` Now that this pin is independently bumped from npm `latest` per the weekly task added in #696, the claim no longer holds and will drift silently in the same way the description note did.

Suggested edit (last line of the comment block, immediately above `- name: Install agent toolchain (sandbox)`):

```diff
-    # proxy env here). Same pinned claude version as claude-code-base-action.
+    # proxy env here).
```

Happy to push the one-line cleanup if you'd like.

tend-agent (approved):

PR #701, opened 2026-06-17
chore: regenerate workflows with tend 0.1.5

max-sixty:
Regenerates the eight tend-generated workflows with `uvx tend@latest init` after the 0.1.5 release, bumping the action ref `max-sixty/tend@0.1.4` → `@0.1.5`. Pure version-bump regen (header comment + `uses:` ref per file); no behavioral change.


tend-agent (approved):
Clean regen — every change is the `0.1.4` → `0.1.5` bump (header comment + `uses:` ref) across the eight generated workflows, and the `max-sixty/tend@0.1.5` tag resolves, so the pins are valid.

One out-of-diff note, no action needed for this PR: `review-reviewers.yaml` still pins `max-sixty/tend/interactive@0.1.4`. It's hand-maintained (not produced by `init`), so it's correctly outside this regen's scope — flagging only in case you want to bump it to `@0.1.5` in the same pass to keep the harness refs aligned.

PR #705, opened 2026-06-18
feat: add max-sixty/tend/claude harness subpath

max-sixty:
## What

Adds `max-sixty/tend/claude@X.Y.Z` as a named subpath for the Claude SDK harness, so all three harnesses are consistent named subpaths instead of the Claude one sitting at the repo root:

- `max-sixty/tend/claude@X.Y.Z` — Claude SDK (headless `claude -p` behind the credential proxy)
- `max-sixty/tend/interactive@X.Y.Z` — Claude interactive (PTY-supervised)
- `max-sixty/tend/codex@X.Y.Z` — Codex

The generator now emits `max-sixty/tend/claude@<version>` for `harness: claude`. The bare root `max-sixty/tend@X.Y.Z` keeps working unchanged for adopters already pinned to it — published immutable tags can't be rewritten, so the root must remain. This is a deliberate exception to tend's cut-over-completely rule.

This branch was rebased onto and merged with #704 (the headless-`claude -p`-behind-proxy rewrite of the root action), which landed while this was in flight; the alias is built against that new action.

## Design: one definition, no duplication

The obvious approach — a root passthrough whose only step is `uses: ./claude` — does not work. In a composite action a relative `uses:` resolves against the consumer's `$GITHUB_WORKSPACE`, not the action's own directory, and `uses:` can't take `${{ }}` expressions ([actions/runner#1348](https://github.com/actions/runner/issues/1348), still open). `uses: ./claude` would look for `claude/action.yaml` in the *adopter's* repo and fail.

So the root `action.yaml` stays the single canonical definition. The `claude/` subpath re-exposes it via symlinks to it and to the resources its steps reach through `${{ github.action_path }}`:

```
claude/action.yaml    -> ../action.yaml      # the canonical action
claude/shared         -> ../shared           # system-prompt base, read in place
claude/interactive    -> ../interactive      # the credential proxy under interactive/proxy/
claude/.claude-plugin -> ../.claude-plugin   # marketplace manifest
```

When invoked as `max-sixty/tend/claude@X.Y.Z`, `github.action_path` is `.../claude`, and every `${{ github.action_path }}/...` lookup resolves through a symlink back to the real repo-root resource, exactly as at `max-sixty/tend@X.Y.Z`. The action body is byte-identical for both refs; only path resolution differs. Git stores these as symlinks (mode 120000) and preserves them on action checkout; tend runs only on Linux runners. There is no second copy of the action to drift — `claude/action.yaml` *is* the root file.

One edit to `action.yaml` makes the alias work end-to-end: the sandbox marketplace step now resolves `MARKETPLACE_SRC` via `realpath "${{ github.action_path }}/.claude-plugin/.."`, which follows the alias's `.claude-plugin` symlink back to the real repo root. Its `cp -a` then copies the real `.claude-plugin`/`plugins` dirs (preserving their internal symlinks) instead of the alias's own symlinks (which `cp -a` would copy as dangling links). The change is a no-op for the root ref (resolves to the same dir).

Considered and rejected: a recursive `claude -> .` symlink (makes `realpath` collapse to the root for free, but a self-referential dir is a footgun for any tool that walks the tree); `cp -aL` to dereference during the copy (would flatten the plugin trees' real internal symlinks, a behavior change for the root ref too); copying the action body into `claude/` (duplicates ~670 lines, drifts).

## Changes

- `claude/` — symlink alias (`action.yaml`, `shared`, `interactive`, `.claude-plugin`) + README
- `action.yaml` — resolve the sandbox marketplace source through `.claude-plugin/..` (one line, no-op for the root ref)
- `generator/src/tend/templates/macros.yaml.j2` — emit `/claude@` for the claude harness; interactive/codex untouched
- generator tests — updated assertions, a new `test_claude_action_ref`, and the conftest version-scrub regex broadened from `(?:/codex)?@` to `(?:/\w+)?@` so it scrubs the `/claude@` (and `/interactive@`) subpath
- regtest snapshots — claude-harness `uses:` line is now `/claude@` (codex and install-test snapshots unchanged)
- `README.md`, `CLAUDE.md`, `docs/security-model.md`, `running-tend`/`release` skills — describe `/claude` as the canonical Claude ref and the root as a retained alias

## Version-pin note

`max-sixty/tend/claude@X.Y.Z` only resolves once a release tags a commit that already contains `claude/` — the same bootstrap the `interactive` and `codex` subpaths went through. Tend's own committed `.github/workflows/tend-*.yaml` are intentionally **not** regenerated here (the in-tree generator would stamp a tag that doesn't exist yet); they pick up `/claude@` at the next release. The regtest snapshots do reflect the new ref, since they test generator output rather than live workflows.

## Tests

`wt test` (256 passed) and `pre-commit run --all-files` (actionlint over the new `action.yaml`, check-yaml, ruff, typos, uv-lock, reference sync) pass. The `claude-smoke` agent run on this PR validates the action body end-to-end via the root ref — and since the `/claude` body is byte-identical (same file via symlink), that coverage transfers. The `/claude` path resolution is verified locally; it's exercised end-to-end by the post-release integration smoke (the release-pinned `max-sixty/tend/claude@` ref only exists once a release tags `claude/`).


tend-agent (commented):
The subpath mechanism itself is clean — the four `claude/` symlinks (`action.yaml`, `shared`, `plugins`, `.claude-plugin`) cover every `${{ github.action_path }}`-relative lookup in the *current PR's* `action.yaml`, the regtest/test churn is consistent, and the new `test_claude_action_ref` is a good guard. But this PR was branched before #704, and that changes the picture in two ways worth resolving together rather than as a plain conflict fixup.

**1. Merge conflict — based on pre-#704 main.** GitHub reports this PR as `CONFLICTING`/`DIRTY`. #704 ("Run the default claude harness headless behind the credential proxy") landed on `main` and rewrote the same `action.yaml`, `CLAUDE.md`, `README.md`, and `docs/security-model.md` regions this PR edits. The PR's descriptive text ("Claude SDK — invokes `claude-code-action` with the tend plugin marketplace") describes the *old* mechanism; `main` now describes and runs the Claude harness as headless `claude -p` behind the credential proxy. So this needs a rebase onto `main`, and the new prose in `claude/README.md` and `CLAUDE.md` needs reconciling with the headless description rather than a mechanical conflict resolution.

**2. The symlink set won't be complete after the rebase.** The alias design was validated against the pre-#704 `action.yaml`, which only read `shared/system-prompt.md` and passed `plugin_marketplaces: ${{ github.action_path }}` to `claude-code-action`. Main's headless `action.yaml` reads more under `github.action_path`, and `claude/` has no `interactive` symlink:

- `action.yaml` (on main) runs `bash "${{ github.action_path }}/interactive/proxy/setup-sandbox.sh"` and sets `ACTION_PATH: ${{ github.action_path }}/interactive` so `${ACTION_PATH}/proxy/inject_credentials.py` resolves. Run as `max-sixty/tend/claude@X.Y.Z`, both resolve under `.../claude/interactive/proxy/...`, which doesn't exist — the proxy setup would fail.
- The plugin-install step does `MARKETPLACE_SRC="$(realpath "${{ github.action_path }}")"` then `cp -a "$MARKETPLACE_SRC/.claude-plugin" "$MARKETPLACE_SRC/plugins" "$SANDBOX_MKT/"`. `realpath` resolves the `claude/` dir itself (not a symlink), so the children stay as symlinks; `cp -a` preserves symlinks (`-a` ⇒ `--no-dereference`), so the copied `.claude-plugin`/`plugins` land in the sandbox as dangling relative links (`../.claude-plugin` resolves outside `$SANDBOX_MKT`). The old path worked because `claude-code-action` dereferenced the marketplace at read time; the new physical copy does not.

So after rebasing, the `/claude` invocation likely needs an `interactive` symlink too (and the `cp -a` would want to dereference, e.g. `cp -aL` or copy from `realpath`-resolved child paths) — i.e. the alias needs re-validating against the headless action, not just a textual merge.

(These are based on reading `main`'s `action.yaml`; I haven't run the rebased `/claude` ref in CI. Flagging for human review given the supply-chain-sensitive surface.)

tend-agent (approved):

tend-agent (approved):

max-sixty:
Superseded by #712, now merged. #712 lands the same harness-named subpaths (`max-sixty/tend/claude@`, `/claude-interactive@`, `/codex@`) and goes further: it de-duplicates the shared step bodies into `shared/steps/`, renames `interactive/` → `claude-interactive/`, and hoists the proxy to a top-level `proxy/`.

The one deliberate difference from this PR: #712 cuts over fully (moves the root action into `claude/`) rather than keeping the bare `max-sixty/tend@X` ref alive via symlinks — confirmed as the intended direction. Closing in favor of #712.

PR #706, opened 2026-06-18
Make claude-smoke dispatch-only

max-sixty:
Drops the `pull_request` trigger from `claude-smoke.yaml`, leaving it `workflow_dispatch`-only like its sibling `interactive-smoke.yaml`. The PR trigger was only added to validate the smoke before its own PR merged (a new `workflow_dispatch` workflow can't be dispatched until it's on the default branch); now that #704 is merged, it's no longer needed and would cost tokens on every harness PR.

The fork guard simplifies to the owner check, and the header documents it as a manual tool (`gh workflow run claude-smoke.yaml --ref <branch>`). The prompt file and `-f prompt=` override stay.

### DIFF
--- a/.github/workflows/claude-smoke.yaml
+++ b/.github/workflows/claude-smoke.yaml
@@ -7,10 +7,10 @@ name: claude-smoke
 # version; the in-tree action is unreleased until the next `pypi-release`
 # run).
 #
-# Runs automatically on any PR that touches the harness (action.yaml or the
-# shared proxy), and manually via `workflow_dispatch`. The `prompt`/`model`
-# inputs override the defaults on a manual run; the standard prompt lives in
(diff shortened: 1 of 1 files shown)


tend-agent (approved):
Correct and matches the sibling `interactive-smoke.yaml` (same `if: github.repository_owner == 'max-sixty'` guard, same dispatch-only shape). Dropping `|| 'sonnet'` is safe now that `workflow_dispatch` is the only trigger — the `model` input carries `default: sonnet`, so `inputs.model` is always populated.

One stale comment the diff missed, just outside the hunk: the line above the "Compose smoke prompt" step still reads `# One prompt source for both triggers: a manual -f prompt= override,` — there's only one trigger now. Want me to push a one-line commit changing it to `# Prompt source: a manual -f prompt= override,`?

PR #712, opened 2026-06-18
refactor(actions): share step scripts; name actions by harness

max-sixty:
Two related cleanups to tend's composite actions, in one branch. No behavior change for adopters.

## 1. De-duplicate the shared step bodies

The three harness actions repeated ~700 lines of identical preflight/setup/teardown bash. Each body now lives once as a script under `shared/steps/`, invoked by a thin wrapper that passes inputs via env — the same `github.action_path`-relative script pattern the proxy already uses, so there's no self-version-pin and the scripts stay branch-smoke-testable (a nested composite sub-action would have forced both). Five scripts are shared by all three actions; five by the two Claude actions. Bodies moved verbatim — the only edits are a `set -eo pipefail` header (matching GitHub's `shell: bash` default) and lifting three `${{ }}` interpolations to env vars.

## 2. Name every action by its harness

The generator stamps the `uses:` ref from the `harness:` config, so the unnamed `max-sixty/tend@X` Claude default was an asymmetry. `action.yaml` → `claude/`, `interactive/` → `claude-interactive/`, and the proxy moved to a top-level `proxy/` shared by both Claude actions. All three actions are now one level deep, so every shared-infra reference is uniformly `../shared/` / `../proxy/`.

## Reviewing

- `shared/steps/*.sh` — the extracted bodies.
- `claude/action.yaml`, `claude-interactive/action.yaml`, `codex/action.yaml` — now thin wrappers.
- `generator/src/tend/templates/macros.yaml.j2` — stamps `max-sixty/tend/{claude,claude-interactive,codex}@X`.
- `.pre-commit-config.yaml` — an input-parity guard keeps the two Claude actions swap-compatible (they share the generator's `with:` block); a shellcheck hook restores the lint coverage the inline `run:` blocks had via actionlint.

## Compatibility

Nothing breaks. Existing `uses:` refs are pinned to immutable tags (0.1.5/0.1.6) whose layout is unchanged; harness-named refs only appear at the next release's regen. The committed `tend-*.yaml` workflows (`@0.1.6`) and `review-reviewers.yaml` (`/interactive@0.1.5`) are intentionally left on their old pins.

## Testing

pre-commit (actionlint, shellcheck, the parity guard), generator tests (255), proxy tests (21), and a programmatic check that every extracted body is byte-equivalent to its original — all green. The end-to-end runtime path changed (steps now `bash script.sh`), so dispatch `claude-smoke.yaml --ref dedup-harness-actions` (and `interactive-smoke.yaml`) to validate on a real runner before relying on it — those need repo secrets and aren't exercised by PR CI.


tend-agent (approved):
Verified the mechanics this refactor turns on, all clean:

- **Env-var lifting is complete.** Every `bash ../shared/steps/*.sh` wrapper passes exactly the env the script reads (`GITHUB_TOKEN`, `BOT_NAME`, `SYSTEM_PROMPT_FILE`, `EXTRA`, `CLAUDE_VERSION`, `MARKETPLACE_ROOT`, …); the rest are auto-provided `GITHUB_*` / `SANDBOX` / `AGENT_*` from `$GITHUB_ENV`. The byte-equivalence test only guards the bodies, so this was the spot a missed lift would slip through — none did, across all three actions including codex's shared subset.
- **`set -eo pipefail`** correctly reconstitutes GitHub's default `shell: bash` flags (`-eo pipefail`, no `-u`) that `bash script.sh` would otherwise drop — matches prior behavior.
- **Paths resolve uniformly** via `../shared/` and `../proxy/` off the full-repo checkout (the proven `github.action_path`-relative pattern the proxy already used); `shared/system-prompt.md` is in place.
- **Input parity** between `claude/` and `claude-interactive/` is exact (12 inputs, same defaults), so the new parity hook passes; shellcheck hook scopes to `shared/steps/*.sh` + `proxy/setup-sandbox.sh`.
- **Generator + regtests** stamp `max-sixty/tend/{claude,claude-interactive,codex}@X` consistently; the intentionally-pinned old refs (`tend-*.yaml@0.1.6`, `review-reviewers@/interactive@0.1.5`) are correctly left untouched.

One thing to confirm rather than a defect: #705 ("feat: add max-sixty/tend/claude harness subpath") is still open and targets the same harness-naming goal over a near-identical file set (`claude/action.yaml`, `codex/action.yaml`, `macros.yaml.j2`, the same regtest outputs, `security-model.md`). This PR looks like the fuller version — it also renames `interactive/ → claude-interactive/`, extracts the shared step bodies, and hoists the proxy. Worth closing one as superseded so they don't fight over the same files.

PR #715, opened 2026-06-19
skills(nightly): skip action-ref downgrade regen PRs

### LINKED ISSUE #714: nightly: regen reverts Dependabot action-ref bumps (e.g. actions/checkout v7→v6)
## Problem

When Dependabot bumps an action ref *inside* tend's generated workflow files (e.g. `actions/checkout` v6 → v7) and a maintainer merges that bump, the committed `.github/workflows/tend-*.yaml` files end up *newer* than what tend's generator emits. The nightly skill's "Step 7: Update tend workflows" then regenerates the files, producing a diff that **downgrades** the action ref back to the version tend pins — reverting the merged Dependabot bump.

The `NON_STAMP_DIFF` guard only filters out the `# Generated by tend X.Y.Z` header line, so these action-ref downgrades count as substantive diff and the skill would open a PR. The PR carries negative value: it reverts a legitimate, maintainer-merged dependency bump.

## Evidence (PRQL/prql, nightly 2026-06-19)

- Dependabot [PRQL/prql#6021](https://github.com/PRQL/prql/pull/6021) "chore: bump actions/checkout from 6 to 7" merged 2026-06-18, bumping the committed `tend-*.yaml` files to `actions/checkout@v7`.
- tend `0.1.6`'s generator still emits `actions/checkout@v6`.
- `uvx tend@latest init` regen diff was **only**:

  ```
  8 +      - uses: actions/checkout@v6
  8 -      - uses: actions/checkout@v7
  ```

  across the 8 `tend-*.yaml` files. No version bump (0.1.6 → 0.1.6). `NON_STAMP_DIFF` = 16, so the stamp-only skip did not trigger.

This nightly run reasoned about it manually and skipped the PR, but a future run that doesn't catch the pattern would open a checkout-downgrade PR — and it would recur every night until tend's generator catches up to v7.

## Proposed fix

The cleanest fix is at the generator: when regenerating, **preserve consumer-bumped action refs** rather than clobbering them — i.e. if the committed workflow pins an action at a ref `>=` the one the generator would emit, keep the committed ref. This lets Dependabot own action-ref bumps without nightly fighting it.

Alternatively (or additionally), extend the nightly skill's `NON_STAMP_DIFF` guard to treat a diff whose only non-stamp content is action-ref version changes where the committed ref is newer as a no-op, same as the stamp-only case — so the skill skips the PR instead of relying on the run to reason it out.

tend-agent:
## Problem

When Dependabot bumps an action ref inside a consumer's committed `tend-*.yaml` files (e.g. `actions/checkout@v6` → `@v7`) between tend releases, the committed workflows run *ahead* of what tend's generator emits. The nightly "Update tend workflows" step then regenerates the files, producing a diff that **downgrades** the ref back to tend's pin — reverting the maintainer-merged Dependabot bump.

The `NON_STAMP_DIFF` guard added in #689 only strips the `# Generated by tend X.Y.Z` header, so these action-ref downgrades count as substantive diff and the skill would open a revert PR — recurring nightly until tend's generator catches up.

## Solution

Extend the stamp-only skip in `plugins/tend-ci-runner/skills/nightly/SKILL.md` with a direction-aware rule: when the only non-stamp diff is action-ref *downgrades* (the regenerated `+` ref is older than the committed `-` ref), skip the PR — it would revert a maintainer-merged bump, and tend catches up to the newer ref in a later release. Direction is the discriminator: a diff where the regenerated ref is *newer* than committed is a real upgrade tend is shipping, and still opens a PR normally.

This is prose guidance rather than a version-comparison bash extension by design: the agent already drives this judgment (the run that filed #714 reasoned its way to skipping the PR), and a brittle downgrade-detector would have to special-case direction to avoid suppressing legitimate upgrades. Codifying the rule makes the existing reasoning reliable without embedding fragile bash.

## Testing

Skill-text change — no code path. Verified the guidance isn't duplicated in the co-loaded nightly skill (the existing `NON_STAMP_DIFF` block covers stamp-only, not action-ref direction) and that the placement reads naturally alongside the stamp-only skip it extends.

---
Closes #714 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -231,8 +231,10 @@ NON_STAMP_DIFF=$(git diff --no-color .github/workflows .config \
   | wc -l)
 ```
 
-If `git status` shows no changes, or `NON_STAMP_DIFF` is `0`, clean up
-and skip the PR:
+**Action-ref downgrade skip.** Dependabot may bump an action ref inside the committed `tend-*.yaml` files (e.g. `actions/checkout@v6` → `@v7`) between tend releases, leaving them ahead of what the generator emits. Regen then produces a diff that *downgrades* the ref back to tend's pin, and `NON_STAMP_DIFF` counts it as substantive. If the only non-stamp diff is such downgrades — the regenerated (`+`) ref is older than the committed (`-`) ref — skip the PR: it would revert a maintainer-merged dependency bump, and tend catches up to the newer ref in a later release. Direction matters: a diff where the regenerated ref is *newer* than committed is a real upgrade tend is shipping — open the PR normally.
+
+If `git status` shows no changes, or `NON_STAMP_DIFF` is `0` (or the only
+remaining diff is action-ref downgrades, per above), clean up and skip the PR:
 
 ```bash
 cd -


tend-agent (commented):
The rule itself is correct: the `+`/`-` direction mapping is right (regen output is `+`, committed is `-`), it's placed where the agent already has `NON_STAMP_DIFF` in hand, and it doesn't duplicate the stamp-only block above it.

One scope concern, since this PR says `Closes #714`: the issue proposed two fixes and called the generator-side one — "preserve consumer-bumped action refs rather than clobbering them" — *the cleanest*. This PR implements only the issue's *alternative* (skill-side skip). That mitigation has a gap the generator fix wouldn't: it only fires when the **entire** non-stamp diff is downgrades. The next time tend ships a version bump whose body genuinely changes, regen produces `{body change + the same action-ref downgrade}`, so "only remaining diff is downgrades" is false, the PR opens, and it reverts the Dependabot bump anyway. So the recurring-nightly-revert is stopped, but the underlying generator-vs-committed divergence is not — it just resurfaces on the next real regen.

Worth considering keeping #714 open (or filing a generator follow-up) to track the durable fix, rather than closing it on the partial mitigation. Not a blocker on this change.

PR #730, opened 2026-06-24
fix: configure git identity before bundled-skill worktree commits

### LINKED ISSUE #729: nightly: worktree commit fails on runners without a pre-seeded git identity
## Summary

The bundled `nightly` skill's worktree-commit recipes run `git commit` without first configuring `git config user.name` / `user.email`. On runners where the harness does **not** pre-seed a git identity, the first commit in a fresh `/tmp` worktree fails with `Author identity unknown` / `fatal: empty ident name`, the branch gets pushed with **no commit**, and `gh pr create` then fails with `No commits between main and <branch>`. The session self-corrects (sets identity, recommits, retries), but it burns turns and momentarily pushes an empty branch.

This has now fired on the **two most recent committing nightlies** on `max-sixty/cargo-affected` — both runs that actually exercised the worktree-commit path hit it — which points at the runner environment having stopped pre-seeding identity, making this structural rather than a one-off.

## Where it fires

`plugins/tend-ci-runner/skills/nightly/SKILL.md`, Step 7 (workflow regen), ships:

```bash
TITLE=$(cat "/tmp/tend-pr-title")
git commit -m "$TITLE"          # <-- no identity configured in this worktree
git push -u origin tend/update-workflows
gh pr create --title "$TITLE" --body-file "/tmp/tend-update-body.md"
```

The same gap exists in Step 8 (fix-findings worktrees) and the Step-1 conflict-resolution `git commit --no-edit`. A fresh `git worktree` does not inherit a repo-local identity, and the runner's global config is empty, so any of these commits can hit it.

## Evidence

Both occurrences are nightlies on `max-sixty/cargo-affected`:

- **2026-06-24** — [run 28082587612](https://github.com/max-sixty/cargo-affected/actions/runs/28082587612). Step 7 commit on `tend/update-workflows` failed with `Author identity unknown` → branch pushed with no commit → `pull request create failed: GraphQL: No commits between main and tend/update-workflows`. Bot set identity from `gh api user` and retried successfully (became PR #59). Notably the bot then *proactively* set identity before its Step 8 docs PRs (#60, #61), so only the first commit attempt was bitten.
- **2026-06-22** — [run 27939759096](https://github.com/max-sixty/cargo-affected/actions/runs/27939759096). Same failure shape on the PR #58 worktree commit; same self-correction.

The recovery the bot converged on, verbatim:

```bash
BOT_ID=$(gh api user --jq '.id')
git config user.name "$BOT_LOGIN"
git config user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
```

## Proposed fix

Add an identity-setup step to the bundled recipes that commit in a `/tmp` worktree — either once near the top of the nightly skill, or inline before each `git commit`. Using the GitHub noreply form (`<id>+<login>@users.noreply.github.com`) keeps commits attributed to the bot account and passes any `verified`-email push rules. Idempotent, so it's safe to repeat:

```bash
BOT_ID=$(gh api user --jq '.id')
git config user.name "$BOT_LOGIN"
git config user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
```

The same recipe shape (commit in a `/tmp` worktree) also appears in the `review-runs` skill's skill-PR flow and in `running-in-ci`'s "Learning from Feedback" worktree dance, so the fix is worth applying wherever a bundled recipe commits in a freshly-added worktree.

Filed from the `max-sixty/cargo-affected` review-runs run [28086510628](https://github.com/max-sixty/cargo-affected/actions/runs/28086510628).

tend-agent:
## Problem

Bundled skills that commit in a fresh `/tmp` `git worktree` (and in the main checkout) run `git commit` without first configuring a git identity. On runners where the harness doesn't pre-seed one, the first commit fails with `Author identity unknown`, the branch gets pushed with no commit, and `gh pr create` then fails with `No commits between main and <branch>`. Sessions self-correct by setting an identity and recommitting, but they burn turns and momentarily push an empty branch. Per #729 this has fired on consecutive committing nightlies, making it structural rather than a one-off.

The gap is systemic — the same `git commit`-without-identity shape exists across `nightly`, `review-runs`, `running-in-ci`, `triage`, `review`, and `ci-fix`.

## Solution

Fix it once at the root and reinforce at the documented failure sites:

- **`running-in-ci/SKILL.md`** — new `### Configure git identity before the first commit` subsection under PR Creation. Every skill loads `running-in-ci` first, so this is the canonical rule for all of them. Uses `git config --global` (covers the main checkout and every worktree in one shot) with the GitHub noreply email form (`<id>+<login>@users.noreply.github.com`) so commits stay attributed to the bot and pass `verified`-email push rules. Also annotated the Learning-from-Feedback worktree block to point at it.
- **`nightly/SKILL.md`** — inlined the identity config into the Step 7 verbatim ship block (the failure site in #729). Shell state doesn't persist between bash calls, so it's self-contained there.
- **`review-runs/SKILL.md`** — annotated its worktree-commit block to set identity first.

## Testing

Skill-text fix — no generator unit test applies. Reproduced the mechanism locally: `git commit` with no configured identity fails with `Author identity unknown` both in a fresh repo and a fresh worktree; setting `git config --global user.name`/`user.email` lets the same commit succeed. `pre-commit` (trailing-whitespace, typos, bang-backtick guard, install-tend sync) passes on all three changed files.

---
Closes #729 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -109,6 +109,14 @@ Do not check out or rebase manually — the bot owns the branch and will overwri
 
 ### Bot-authored PRs: resolve manually
 
+Set the git identity once in the main session **before dispatching** — the subagents below `git commit --no-edit` in fresh worktrees, and as separate invocations they don't reliably load `running-in-ci`. `--global` writes to the runner's shared `~/.gitconfig`, so the subagents inherit it and don't fail with `Author identity unknown`. See "Configure git identity before the first commit" in `/tend-ci-runner:running-in-ci`.
+
+```bash
+BOT_LOGIN=$(gh api user --jq '.login'); BOT_ID=$(gh api user --jq '.id')
+git config --global user.name "$BOT_LOGIN"
+git config --global user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
+```
+
 For each conflicted PR authored by `$BOT_LOGIN`, dispatch a subagent to:
 
 1. Check out the PR: `gh pr checkout <number>`
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -284,6 +284,12 @@ Then ship it:
 ```bash
 TITLE=$(cat "/tmp/tend-pr-title")
 git add -A .github/workflows .config
+# A fresh /tmp worktree has no git identity; without this the commit fails with
+# `Author identity unknown` and an empty branch gets pushed. Idempotent — see
+# "Configure git identity before the first commit" in /tend-ci-runner:running-in-ci.
+BOT_LOGIN=$(gh api user --jq '.login'); BOT_ID=$(gh api user --jq '.id')
+git config --global user.name "$BOT_LOGIN"
+git config --global user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
 git commit -m "$TITLE"
 git push -u origin tend/update-workflows
 gh pr create --title "$TITLE" --body-file "/tmp/tend-update-body.md"
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -225,6 +225,9 @@ cd "/tmp/review-runs-fix/.claude/skills/running-tend" && mv /tmp/running-tend.md
 
 cd "/tmp/review-runs-fix"
 git add .claude/skills/
+# Set git identity first if not already done this session — a fresh worktree has
+# none and the commit fails with `Author identity unknown`. See "Configure git
+# identity before the first commit" in /tend-ci-runner:running-in-ci.
 git commit -m "skills(running-tend): ..."
 git push -u origin daily/review-runs-$GITHUB_RUN_ID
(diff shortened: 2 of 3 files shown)


tend-agent (commented):
The canonical rule + the Step 7 inline are a solid root fix, and `--global` over the issue's repo-local `git config` is the right call (a fresh worktree shares the runner's `~/.gitconfig`). One residual gap, though, on a site #729 explicitly named.

**The conflict-resolution commits in Step 3 run before any identity is configured.** "Resolve conflicts on bot PRs" dispatches parallel subagents that each `git add` + `git commit --no-edit` in a fresh `git worktree add /tmp/pr-<number>` — and that step executes well before Step 7, where the inline `git config --global` now lives. So at the moment those commits fire:

- The main session hasn't reached Step 7 yet, so no `--global` identity has been written to `~/.gitconfig`.
- The commits run inside dispatched subagents, which are separate agent invocations and don't reliably load `running-in-ci`, so the new canonical rule isn't guaranteed to reach them.

That leaves the conflict-resolution `git commit --no-edit` — one of the three sites #729 called out — still exposed to `Author identity unknown`.

Because `git config --global` writes to the runner's shared `~/.gitconfig` (filesystem state, not shell state — it persists across bash calls *and* across subagents in the same runner), setting it **once, early** in the main session — before Step 3 dispatches subagents — would cover every downstream commit site (conflict-resolution subagents, Step 7, and any later worktree dance) in one shot, strictly more robustly than placing it only at Step 7. A natural home is Step 0 right after loading `running-in-ci`, or just ahead of the Step 3 dispatch.

Not blocking the improvement here — just flagging for a follow-up so the fix reaches the site that was actually observed failing.

PR #735, opened 2026-06-26
fix(review): leak-free incremental review (three-dot diff + git log --no-merges)

### LINKED ISSUE #734: review: incremental compare $LAST_REVIEW_SHA...$HEAD_SHA leaks base-merge changes into the review
## Summary

The bundled `tend-ci-runner` `review` skill's **incremental-review path** computes "what changed since my last review" with a two-endpoint `compare/$LAST_REVIEW_SHA...$HEAD_SHA`. When a `Merge branch '<default>'` commit lands between those two endpoints, that compare pulls in everything the default branch brought in — so the incremental review critiques code the PR never authored. It reads as if the merge-brought-in changes are the PR's own work.

This fires for **any tend consumer** whose PRs merge the default branch mid-review (common on longer-lived branches), so the fix belongs in the bundled skill, not in any one consumer.

## Where

`plugins/tend-ci-runner/skills/review/SKILL.md`, step 1 ("Pre-flight checks"). When `LAST_REVIEW_SHA` exists and differs from `HEAD_SHA`:

```bash
gh api "repos/$REPO/compare/$LAST_REVIEW_SHA...$HEAD_SHA" \
  --jq '{total: ([.files[] | .additions + .deletions] | add), files: [.files[] | "\(.filename)\t+\(.additions)/-\(.deletions)"]}'
```

A `A...B` compare is the symmetric-difference range: it includes commits reachable from `B` but not `A`. A base-merge commit in that range drags in the entire merged-in delta of the default branch, none of which is the PR's work. The skill then anchors the incremental review (the trivial-vs-substantive decision and the model's reading) on that leaked diff.

## Reproduced (worktrunk #3246)

Confirmed on worktrunk PR #3232 (a logging migration). It first reviewed `599700c7` (06:05Z), then a `Merge branch 'main'` commit (`6eee3f15`) landed pulling in #3229's statusline rate-limit feature, then it re-reviewed `ada0e915` (06:41Z). The incremental compare `599700c7...ada0e915` spanned the merge, so the review opened with a full paragraph critiquing #3229's Gauss–Hermite lockout math — code #3232 never touched.

Counting `expected_lockout`/`colorize_pace`/`lockout_cost` in `statusline.rs` across the two diffs the skill could use:

| Diff | lockout-term hits | verdict |
|---|---|---|
| `gh pr diff 3232` (three-dot `main...HEAD`) | 1 (incidental context line) | clean — migration only |
| `compare/599700c7...ada0e915` (incremental) | 30 (`posterior_lambda`, Gauss–Hermite, `expected_lockout`) | leaks #3229 |

```bash
gh pr diff 3232 | grep -cE "expected_lockout|colorize_pace|lockout_cost"        # -> 1
gh api "repos/$REPO/compare/599700c7...ada0e915" \
  --jq '.files[] | select(.filename=="src/commands/statusline.rs") | .patch' \
  | grep -cE "expected_lockout|colorize_pace|lockout_cost"                       # -> 30
```

The three-dot `gh pr diff` (the full-review path, step 3) is correct and would never surface #3229 — the leak is specific to the incremental two-endpoint compare spanning a base-merge.

## Impact

Any PR that merges the default branch during review gets reviewed for unrelated base code: wasted review budget, misleading feedback to the author, and the risk of the bot flagging or holding on issues the author can't act on. The failure is **structural** — every base-merge between two review points triggers it, with no decision point — and largely **invisible**, since the leaked paragraph reads as a normal review.

## Proposed fix

Keep "what's new since I last looked," but drop merge-pulled base changes by intersecting the incremental change-set with the PR's true three-dot surface. Two equivalent shapes:

1. **Filter the incremental file set to files in `gh pr diff`.** Restrict the `compare/...` file list to filenames that also appear in the PR's three-dot `gh pr diff <number>`. Base-merge files that the PR never touched drop out; files the PR *does* touch still show their since-last-review delta.

2. **Re-anchor to three-dot surfaces at both review points.** Compute the PR's three-dot diff at `LAST_REVIEW_SHA` and at `HEAD_SHA` and compare those, rather than `LAST_REVIEW_SHA...HEAD_SHA` directly.

Option 1 is the smaller edit — one extra filter on the existing `compare` output — and avoids re-deriving the merge base. A short note in the skill explaining *why* (the `A...B` range spans base-merges) would keep a future edit from reverting to the simpler form.

---

Filed from worktrunk's daily `review-runs` analysis. Origin report and full reproduction: max-sixty/worktrunk#3246 (maintainer-confirmed). No code change is needed in worktrunk — the diff computation lives entirely in the bundled skill.

tend-agent:
## Problem

The `review` skill's incremental pre-flight (step 1) computed "what changed since my last review" with a two-endpoint `gh api compare/$LAST_REVIEW_SHA...$HEAD_SHA`. The `A...B` symmetric-difference range includes every commit reachable from `B` but not `A`, so when a `Merge branch '<default>'` commit lands between the two review points, the compare drags in the **entire merged-in base delta** — none of which is the PR's own work. The incremental review then anchored its trivial-vs-substantive decision and the model's reading on that leaked diff, critiquing code the PR never authored.

Reported and maintainer-confirmed in max-sixty/worktrunk#3246.

This two-endpoint compare was the **original** form of the incremental check — it was never added to handle conflicts or any corner case. The `...` range just happens to leak base-merges.

## Solution

Keep both signals the incremental check needs, with both made leak-free against base-merges:

- **Review surface** — the PR's **three-dot** `gh pr diff` (merge-base→head), which base-merge commits never enter. This is the same diff step 3 (the full-review path) already used; step 1 now uses it too.
- **Accurate incremental** — `git log --no-merges --numstat` over `$LAST_REVIEW_SHA..$HEAD_SHA`, giving the commits and per-file line counts pushed since the last review for the trivial-skip heuristic. `--no-merges` drops any base-merge commit, so its churn never counts — and unlike the old compare's `.files` totals, files touched by both the PR and a base-merge are counted correctly. The review checkout is `fetch-depth: 0`, so the range is present locally.

This replaces the leaky two-endpoint compare (and its earlier partial file-list jq filter, which still mis-counted overlapping files) — the trivial-skip stays accurate and the review surface stays leak-free.

## Testing

No pytest harness covers skill prose; the bug was reproduced and maintainer-confirmed upstream. The change is a doc/recipe edit — the three-dot `gh pr diff` is merge-base→head by construction and `git log --no-merges` excludes the merge commit, so base-merge churn cannot enter either.

---
Closes #734 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -48,11 +48,20 @@ If the bot reviewed a previous commit (`LAST_REVIEW_SHA` exists but differs from
 - The **accurate incremental** — commits authored since the last review, with per-file line counts — for the trivial-skip decision below.
 
 ```bash
-# Commits + per-file line counts authored since the last review. --no-merges
-# drops any base-merge commit, so its churn (code the PR never authored) never
-# counts — overlapping files (touched by the PR *and* a base-merge) are counted
-# correctly too. The review checkout is fetch-depth: 0, so the range is local.
-git log --no-merges --numstat --format='%h %s' "$LAST_REVIEW_SHA..$HEAD_SHA"
+# Commits + per-file line counts authored since the last review.
+#
+# --no-merges alone is NOT enough: it drops the merge *commit*, but a base
+# merge also pulls in the base branch's own non-merge commits, which are
+# reachable from $HEAD_SHA but not from $LAST_REVIEW_SHA — so they sit in the
+# A..B range and their churn gets summed as if it were the new push. Exclude
+# everything reachable from the base tip with `--not "$BASE_SHA"`: base churn
+# is always an ancestor of the base tip, while the PR's own commits are not,
+# so this isolates exactly the authored incremental (overlapping files
+# included). The review checkout is fetch-depth: 0; fetch the base tip in case
+# the /head conflict-fallback checkout didn't materialize it.
+BASE_SHA=$(gh pr view <number> --json baseRefOid --jq '.baseRefOid')
+git fetch --no-tags --quiet origin "$BASE_SHA" 2>/dev/null || true
+git log --no-merges --numstat --format='%h %s' "$LAST_REVIEW_SHA..$HEAD_SHA" --not "$BASE_SHA"
 ```
 
 If the incremental changes are trivial, skip the full review — go directly to step 7 to resolve any bot threads addressed by the new changes. After resolving threads: if the most recent bot review was a COMMENT that flagged issues, and those issues are now addressed, submit an APPROVE with an empty body so the PR isn't left in limbo. Otherwise do not submit a new review — the existing one stands. Do NOT proceed to steps 2–6. Rough heuristic: changes under ~20 added+deleted lines that don't introduce new functions, types, or control flow are typically trivial.
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -42,16 +42,17 @@ LAST_REVIEW_SHA=$(gh pr view <number> --json reviews \
 
 If `LAST_REVIEW_SHA == HEAD_SHA`, this commit has already been reviewed — exit silently. Two exceptions: an unanswered conversation question directed at the bot (check below), or `EVENT_ACTION == "ready_for_review"` (the PR just transitioned out of draft, so any prior review was a draft-mode review and the author is now asking for a full one — proceed).
 
-If the bot reviewed a previous commit (`LAST_REVIEW_SHA` exists but differs from `HEAD_SHA`), judge what was pushed since. The diff to read is always the PR's three-dot surface — `gh pr diff <number>` (merge-base→head, the same diff step 3 uses); base-merge commits never enter it. List the commits added since the last review to see the new work:
+If the bot reviewed a previous commit (`LAST_REVIEW_SHA` exists but differs from `HEAD_SHA`), judge what was pushed since. Read two signals, both leak-free against base-merges:
+
+- The PR's three-dot diff — `gh pr diff <number>` (merge-base→head, the same diff step 3 uses) — for the current state. Base-merge commits never enter it.
+- The **accurate incremental** — commits authored since the last review, with per-file line counts — for the trivial-skip decision below.
 
 ```bash
-REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
-# List only the commits added since the last review — judge triviality from
-# these plus the three-dot `gh pr diff`. Don't read this compare's `.files`
-# totals: the `A...B` symmetric-difference range folds in any base-merge's
-# churn (code the PR never authored), which skews the line counts.
-gh api "repos/$REPO/compare/$LAST_REVIEW_SHA...$HEAD_SHA" \
-  --jq '.commits[] | "\(.sha[0:9]) \(.commit.message | split("\n")[0])"'
+# Commits + per-file line counts authored since the last review. --no-merges
+# drops any base-merge commit, so its churn (code the PR never authored) never
+# counts — overlapping files (touched by the PR *and* a base-merge) are counted
+# correctly too. The review checkout is fetch-depth: 0, so the range is local.
+git log --no-merges --numstat --format='%h %s' "$LAST_REVIEW_SHA..$HEAD_SHA"
 ```
 
 If the incremental changes are trivial, skip the full review — go directly to step 7 to resolve any bot threads addressed by the new changes. After resolving threads: if the most recent bot review was a COMMENT that flagged issues, and those issues are now addressed, submit an APPROVE with an empty body so the PR isn't left in limbo. Otherwise do not submit a new review — the existing one stands. Do NOT proceed to steps 2–6. Rough heuristic: changes under ~20 added+deleted lines that don't introduce new functions, types, or control flow are typically trivial.
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -42,12 +42,16 @@ LAST_REVIEW_SHA=$(gh pr view <number> --json reviews \
 
 If `LAST_REVIEW_SHA == HEAD_SHA`, this commit has already been reviewed — exit silently. Two exceptions: an unanswered conversation question directed at the bot (check below), or `EVENT_ACTION == "ready_for_review"` (the PR just transitioned out of draft, so any prior review was a draft-mode review and the author is now asking for a full one — proceed).
 
-If the bot reviewed a previous commit (`LAST_REVIEW_SHA` exists but differs from `HEAD_SHA`), check the incremental changes:
+If the bot reviewed a previous commit (`LAST_REVIEW_SHA` exists but differs from `HEAD_SHA`), judge what was pushed since. The diff to read is always the PR's three-dot surface — `gh pr diff <number>` (merge-base→head, the same diff step 3 uses); base-merge commits never enter it. List the commits added since the last review to see the new work:
 
 ```bash
 REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
+# List only the commits added since the last review — judge triviality from
+# these plus the three-dot `gh pr diff`. Don't read this compare's `.files`
+# totals: the `A...B` symmetric-difference range folds in any base-merge's
+# churn (code the PR never authored), which skews the line counts.
 gh api "repos/$REPO/compare/$LAST_REVIEW_SHA...$HEAD_SHA" \
-  --jq '{total: ([.files[] | .additions + .deletions] | add), files: [.files[] | "\(.filename)\t+\(.additions)/-\(.deletions)"]}'
+  --jq '.commits[] | "\(.sha[0:9]) \(.commit.message | split("\n")[0])"'
 ```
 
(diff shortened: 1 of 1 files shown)


max-sixty:
if we just filter the file list, does this mean we still get the wrong diff for those files?

can we just always do the three-dot diff?

tend-agent:
Good catch on both.

**Q1 — yes, the file-list filter is only a partial fix.** It drops files that are *purely* from the base-merge, but for a file the PR authored *and* the base-merge also touched, the `A...B` compare still reports that file's base-merge churn in its `+adds/-dels`. So the line counts — and the trivial-vs-substantive call that rides on them — stay skewed for overlapping files. The filter cleans the easy case, not the overlap.

**Q2 — yes, always three-dot is leak-free and simpler.** `gh pr diff` is merge-base→head, so base-merge commits never enter it. The only thing we lose is the *incremental* signal: the trivial-skip heuristic would then measure the whole PR surface, so a one-line follow-up push to an already-large PR stops skipping and triggers a full re-review every time. If that cost is acceptable, dropping the entire jq filter is the clean move.

If we want to keep the trivial-skip *and* be leak-free, the precise incremental is "authored commits since last review, excluding the merge commits":

```bash
git log --no-merges --numstat --format= "$LAST_REVIEW_SHA..$HEAD_SHA"
```

Base-merge churn lives in the merge commit, which `--no-merges` drops — so overlapping files get counted correctly too, no file-list filter needed. Caveats: needs the range present locally (a non-shallow fetch), and conflict-resolution edits made *inside* a merge commit are dropped (fine for a trivial heuristic).

My lean: if simplicity wins, go always-three-dot and delete the filter; if the skip optimization is worth keeping, switch to `--no-merges`. Either is strictly better than the current partial filter. Which would you prefer — I'll push it.

max-sixty:
did we originally add this because of issues when there are conflicts orsone other corner case?

otherwise always having the three dot diff everywhere seems much better; try to simplify everywhere

max-sixty:
ah, sorry, this is the incremental commits since the last merge.

yes, off commands for both the (accurate) incremental commits and the total three dot diff

tend-agent:
On the history: no — the two-endpoint compare wasn't added for conflicts or any corner case. It's the *original* form of the incremental check (introduced with the incremental-skip heuristic itself), and `A...B` just happens to be the symmetric-difference range that leaks base-merges. There's no conflict-handling rationale to preserve.

So I went with always three-dot and simplified, per your steer. Pushed:

- Step 1 now reads the PR's three-dot `gh pr diff` surface (merge-base→head) — the same diff step 3 already used. Base-merge commits can't enter it, so no leak, including for files the PR *and* a base-merge both touched (the old file-list filter still mis-counted those).
- The `compare/...` call stays only to **list the new commits** (SHA + subject) so the trivial-skip heuristic can still see what was pushed since the last review. Its `.files` totals — the leaky part — are no longer read.

The one thing this gives up vs. an exact incremental count: the trivial-skip now judges "is the new push trivial?" from the commit list plus the three-dot diff rather than a precise per-push line total. A precise leak-free count would need `git log --no-merges` over the local range, which depends on a non-shallow fetch — not worth reintroducing a second diff mechanism. If you'd rather drop the skip entirely and just always full-review, that's an even smaller change — say the word.

max-sixty:
note the comment I followed up with please

tend-agent:
You're right — your follow-up asked for both, and I'd only kept the three-dot diff. Fixed: step 1 now reads both leak-free signals.

- **Accurate incremental** (for the trivial-skip decision): `git log --no-merges --numstat "$LAST_REVIEW_SHA..$HEAD_SHA"` — commits + per-file line counts pushed since the last review. `--no-merges` drops the base-merge commit, so its churn never counts; and unlike the old compare's `.files` totals, files touched by *both* the PR and a base-merge are counted correctly. The review checkout is `fetch-depth: 0`, so the range is local — no second diff mechanism or extra fetch.
- **Total three-dot diff** (the review surface): `gh pr diff` (merge-base→head), unchanged.

This drops the leaky `compare/...` call entirely. Updated the title/description to match.

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -42,12 +42,17 @@ LAST_REVIEW_SHA=$(gh pr view <number> --json reviews \
...
 
 ```bash
-REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
-gh api "repos/$REPO/compare/$LAST_REVIEW_SHA...$HEAD_SHA" \
-  --jq '{total: ([.files[] | .additions + .deletions] | add), files: [.files[] | "\(.filename)\t+\(.additions)/-\(.deletions)"]}'
+# Commits + per-file line counts authored since the last review. --no-merges
+# drops any base-merge commit, so its churn (code the PR never authored) never
+# counts — overlapping files (touched by the PR *and* a base-merge) are counted
+# correctly too. The review checkout is fetch-depth: 0, so the range is local.
+git log --no-merges --numstat --format='%h %s' "$LAST_REVIEW_SHA..$HEAD_SHA"
```
tend-agent: `--no-merges` strips the merge commit but not the base commits it pulled in — those are non-merge commits in `$LAST_REVIEW_SHA..$HEAD_SHA`, so their churn is still summed. Same base-merge leak as the old `compare/A...B`. See review body for the repro and the `--not <base>` fix (with its base-ref caveat in this checkout).

tend-agent (commented):
The three-dot review surface (signal 1) is correctly leak-free. But the incremental signal still leaks base-merge churn — the same bug this PR set out to fix.

`--no-merges` only drops the *merge commit itself*. The base branch's own commits that the merge pulls in are ordinary non-merge commits, and they're reachable from `$HEAD_SHA` but not from `$LAST_REVIEW_SHA`, so they sit inside the `A..B` range and their `--numstat` churn is summed. The trivial-skip threshold then counts the whole merged-in base delta as "the new push", which is exactly the false-substantive classification the PR is trying to kill.

PR forks at `c0`, authors `own work 1` (last reviewed), then `main` advances with unrelated churn, then the PR does `git merge main` and adds a 1-line follow-up:

```
$ git log --no-merges --numstat --format='%h %s' "$LAST_REVIEW..$HEAD"
822f44e pr: tiny follow-up
1	0	prfile.txt
3fd64ea main: big unrelated churn      # <- base commit, leaked in
1	0	bigbasefile.txt
5	0	shared.txt
```

Actual incremental work is 1 line; the command reports 7, six of which are base churn. A larger base advance trivially blows past the ~20-line trivial threshold. The leak is identical in shape to the old `compare/A...B` — both include everything between merge-base and head when `A` is an ancestor of `B` (the normal case).

</details>

To actually exclude the merged-in commits you have to drop everything reachable from the base, e.g. `git log --no-merges "$LAST_REVIEW_SHA..$HEAD_SHA" --not <base>` (verified: this isolates just the PR's own follow-up). The catch is naming `<base>` reliably in the review checkout — it's `refs/pull/N/merge` at `fetch-depth: 0`, so `origin/<base>` isn't guaranteed to be a local ref, and the `/head` conflict-fallback checkout isn't a merge commit at all. Given that, the earlier "just always three-dot and drop the incremental" option may be the simpler *correct* path here — the incremental can't be made leak-free as cheaply as the description claims.

tend-agent:
Caught a real leak in my own last commit and fixed it in 8773133.

`--no-merges` alone wasn't enough. It drops the merge *commit*, but a base merge also pulls in the base branch's own non-merge commits — those are reachable from `$HEAD_SHA` but not from `$LAST_REVIEW_SHA`, so they stay inside the `A..B` range and their churn got summed as "the new push". Same false-substantive classification this PR set out to kill.

Fix: also exclude everything reachable from the base tip. Base churn is always an ancestor of the base tip; the PR's own commits never are — so `--not "$BASE_SHA"` isolates exactly the authored incremental, overlapping files included. `$BASE_SHA` comes from `gh pr view --json baseRefOid`, with a `git fetch` guard for the `/head` conflict-fallback checkout that may not have materialized it.

Both signals the maintainer asked for are now leak-free: the accurate incremental (`git log --no-merges … --not $BASE_SHA`) for the trivial-skip decision, and the three-dot `gh pr diff` for the review surface.

<details><summary>Verified repro</summary>

PR forks at `c0`, authors `own work 1` (last reviewed), `main` advances with unrelated churn, PR does `git merge main` then a 1-line follow-up:

```
# BUGGY: git log --no-merges "$LAST_REVIEW..$HEAD"  → reports 7 lines, 6 are base churn
pr: tiny follow-up          1  prfile2.txt
main: big unrelated churn   5  bigbasefile.txt     # <- leaked in
                            1  shared.txt
# FIXED: ... --not "$BASE"  → isolates just the PR's own follow-up
pr: tiny follow-up          1  prfile2.txt
```

</details>

max-sixty:
thanks!

could you also log why we missed the 2nd message of mine above? did the agent check for any follow-up comments before returning?

tend-agent:
Checked the session log for the run that posted that reply ([run `28244362446`](https://github.com/max-sixty/tend/actions/runs/28244362446), triggered by your "simplify everywhere" comment). Short answer: **no — the agent did not re-check for follow-up comments before posting.** It read the conversation exactly once, at the start, before your second message existed.

Timeline (all UTC, from the session trace):

| Time | Event |
|---|---|
| 14:26:13 | Run starts (triggered by your 14:26:09 comment) |
| 14:28:22 | Agent fetches `gh pr view 735 --json comments` — **its only read of the thread** |
| 14:29:09 | **You post the follow-up** ("yes, off commands for both…") — 47s after the agent's fetch |
| 14:28–14:32 | Agent investigates git history, edits the skill, amends, force-pushes, writes PR body |
| 14:32:46 | Agent posts its reply — **no second comment fetch anywhere in between** |

So the miss wasn't a misread — your second message simply didn't exist yet when the agent looked, and it never looked again. It spent ~4 minutes on the edit/commit/push and then posted straight from the context it had loaded at 14:28.

This is exactly the case the **Recheck Before Posting** guard in the `running-in-ci` skill is meant to catch — re-fetch the conversation immediately before posting, specifically because "new entries arrived during the session." That step was skipped here. The guidance exists; the run just didn't execute it. The window is wide whenever there's a long edit→commit→push gap between the initial read and the post, which is precisely when a maintainer is most likely to add a follow-up.

The guard is prompt-only, so it relies on the agent remembering to do it. If you'd like, I can propose a tightening to the bundled skill — e.g. make the pre-post re-fetch an explicit gate after any push/edit sequence rather than general advice — but since the rule already exists I held off on opening a PR unprompted. Say the word and I'll send it.

PR #736, opened 2026-06-27
ci-fix: cap the job at a 120-minute timeout

tend-agent:
Caps the `tend-ci-fix` job at `timeout-minutes: 120` so a session that grinds on stress-validating a flaky/timing-sensitive failure is bounded by GitHub Actions rather than by self-budgeting prose in the skill.

Per [maintainer feedback](https://github.com/max-sixty/tend/pull/736#issuecomment-4816094046), this replaces the earlier budget-box paragraph in `ci-fix/SKILL.md` — the deterministic YAML timeout is the cheaper gate (it doesn't cost a skill load every session) and keeps the skill leaner.

### DIFF
--- a/plugins/tend-ci-runner/skills/ci-fix/SKILL.md
+++ b/plugins/tend-ci-runner/skills/ci-fix/SKILL.md
@@ -47,8 +47,6 @@ gh issue list --state open --author "$BOT_LOGIN" --search "ci-fix: in:title" \
 4. Reproduce locally using test commands from the project's CLAUDE.md
 5. Fix at the right level (shared helper > per-file fix)
 
-**Budget-box local validation.** The harness kills the session at a hard timeout (~1h), and a killed session ships nothing — no PR, no issue, and any root cause you found dies with it. Validating a fix for a timing-sensitive or flaky failure means re-running the test many times under load, which drains the budget fast. If repeated stress-validation isn't converging on a green fix and you've already pinned the root cause, stop iterating and take the 3b durable-diagnosis path **while the session is still alive** — a posted diagnosis (offending commit, mechanism, candidate fix direction) that a maintainer or a later run can act on beats grinding to the timeout with nothing to show.
-
 ### 3. Create PR
 
 Re-check for existing fix PRs (one may have been created while you worked).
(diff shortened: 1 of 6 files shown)


plugins/tend-ci-runner/skills/ci-fix/SKILL.md
```diff
@@ -47,6 +47,8 @@ gh issue list --state open --author "$BOT_LOGIN" --search "ci-fix: in:title" \
 4. Reproduce locally using test commands from the project's CLAUDE.md
 5. Fix at the right level (shared helper > per-file fix)
 
+**Budget-box local validation.** The harness kills the session at a hard timeout (~1h), and a killed session ships nothing — no PR, no issue, and any root cause you found dies with it. Validating a fix for a timing-sensitive or flaky failure means re-running the test many times under load, which drains the budget fast. If repeated stress-validation isn't converging on a green fix and you've already pinned the root cause, stop iterating and take the 3b durable-diagnosis path **while the session is still alive** — a posted diagnosis (offending commit, mechanism, candidate fix direction) that a maintainer or a later run can act on beats grinding to the timeout with nothing to show.
```
tend-agent:
```suggestion
**Budget-box local validation.** Stress-validating a timing-sensitive or flaky fix means re-running the test many times under load, which can drain the budget before the harness's hard ~1h timeout — and a session killed by the timeout ships nothing (no PR, no issue, lost diagnosis). If repeated stress-validation isn't converging and you've already pinned the root cause, stop and take the 3b durable-diagnosis path **while the session is still alive**: a posted diagnosis (offending commit, mechanism, candidate fix direction) beats grinding to the timeout with nothing to show.
```

tend-agent (commented):
Change is accurate and well-scoped — it points at the existing §3b off-ramp rather than adding machinery, and "an attempted fix didn't validate locally" is indeed already a §3b condition, so the cross-reference holds.

One observation (self-review): CLAUDE.md's skill-authoring guidance is explicit about brevity — "cut motivation" and "prefer recipe over narrative." The paragraph spends its first two sentences on motivation (killed session ships nothing / stress-validation drains budget) before the rule. Since this loads into every `ci-fix` session, a tighter version that leads with the rule keeps the cost down. Suggestion inline.

max-sixty:
let's just dou let the timeout of 120min and remove the rest; it's overly vomplicated

tend-agent:
Done — dropped the budget-box paragraph and set `timeout-minutes: 120` on the `fix-ci` job in the generator template (`generator/src/tend/templates/ci-fix.yaml.j2`), with the regtest fixtures regenerated to match. Generated workflows otherwise inherit GitHub's 360-min default, so this is the only job that now carries an explicit cap. Title/description updated to match.

Issue #639, opened 2026-05-31
Model-key isolation via the bwrap sandbox works, but is blocked by the Bash `!` quoting bug (verified through claude 2.1.159)


max-sixty:
## Context

This records an investigation into whether tend can hide its credentials from the agent itself. The agent runs on attacker-controlled input (PR diffs, issue bodies, comments) with code execution, so any secret in its reach is a secret a hijacked session can exfiltrate. There are two credentials: the model auth (Claude OAuth token / API key) and the GitHub bot token. This issue covers the model-key half. The GitHub-token half (short-lived GitHub App installation tokens) is separate work.

The model key is a good candidate for hiding because the agent's reasoning never needs it as data. Only the harness's HTTP client presents it on each API call. So if the agent's Bash tool can be isolated from it, the agent can keep working while losing the ability to read or leak the key. Claude Code's built-in bwrap sandbox plus `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB` is the obvious mechanism. The two questions:

1. Does enabling the sandbox actually hide the key from the Bash tool, including the `/proc/<harness-pid>/environ` read that defeats naive env-scrubbing?
2. tend currently sets `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=0` because the bwrap path corrupts `!` in Bash commands. Is that still true on the current pinned `claude` (2.1.150) and on newer releases?

## Results

A standalone probe ran `claude -p` on `ubuntu-latest`, instructing it to run fixed Bash commands that write results to `/tmp`. A plain-bash step then inspects the bytes, so the verdict does not depend on the model narrating correctly. The key-exposure checks report presence and counts only, never values, and the credential-file test uses a dummy canary.

| Check | scrub=0 (no bwrap) | scrub=1 (bwrap sandbox) |
|---|---|---|
| `!` survives the Bash tool | CLEAN (`alpha!beta`, `jq 'select(. != 2)'` works) | CORRUPTED (`alpha\!beta`, `jq` syntax error on `. \!= 2`) |
| OAuth token in subprocess env | 0 chars | 0 chars |
| OAuth token via `/proc` walk | 2 processes | 0 processes |
| Canary credential via Bash | readable | blocked |
| Canary credential via Read tool | readable | blocked |
| Benign command | works | works |

The `!` result is identical on 2.1.150, 2.1.158 (npm `latest`), and 2.1.159 (npm `next`): all three corrupt.

## What this shows

[...]

Issue #637, opened 2026-05-31
action.yaml exposes CLAUDE_CODE_OAUTH_TOKEN but it's unset in agent bash — credential strip happens elsewhere


tend-agent:
## Symptom

`action.yaml:212-217` exposes Anthropic credentials on the `claude` step:

```yaml
CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: "0"
# Expose the harness auth so skills running in the agent's bash can
# propagate it (e.g. seeding a managed repo's secrets).
CLAUDE_CODE_OAUTH_TOKEN: ${{ inputs.claude_code_oauth_token }}
ANTHROPIC_API_KEY: ${{ inputs.anthropic_api_key }}
```

The comment block describes this as a deliberate exposure to support skills that need to propagate the OAuth token (e.g. seeding `CLAUDE_CODE_OAUTH_TOKEN` into a managed repo's secrets, as the integration-test recipe does).

Empirically, both vars are unset/empty in the agent's Bash tool subprocess:

```
OAUTH_TOKEN: unset or empty
ANTHROPIC_API_KEY: unset or empty
```

(Verified in https://github.com/max-sixty/tend/actions/runs/[the run handling #636's review reply] via `[ -n "$CLAUDE_CODE_OAUTH_TOKEN" ]; echo $?`.)

`CLAUDE_CODE_SUBPROCESS_ENV_SCRUB` is **set to 0** (verified in the same run), so the bubblewrap-backed subprocess sandbox is the one knob explicitly *not* doing the scrubbing.

## Hypothesis

Claude Code (the CLI) appears to strip its own Anthropic credentials before spawning Bash-tool subprocesses, independent of `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB`. The env-scrub knob controls bubblewrap full-env isolation; an always-on credential strip is a separate (sensible) defense against user-controlled command execution.

If that's the case, `action.yaml:212-217`'s exposure is a no-op for any skill that runs in the agent's Bash tool. The comment block reads as descriptive ("the var is available, propagate it") but is actually aspirational.

## Impact

[...]

PR #737, opened 2026-06-27
docs(action): drop disproved "scrub hid model auth" claim from #637 quote

tend-agent:
Follow-up to #637, addressing @max-sixty's "but how about the originally quoted section?"

My earlier sweep claimed the whole originally-quoted `env:` block was gone. That was imprecise: three of its four lines (the `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_API_KEY` vars and their "Expose the harness auth…" comment) were removed by the proxy refactor (#704), but `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: "0"` survived — and its comment still carried the one claim #637 actually disproved.

The surviving comment asserted bubblewrap / `SUBPROCESS_ENV_SCRUB` "was hiding the model auth from the agent." #637's evidence was the opposite: the OAuth token was absent from the Bash-tool subprocess **with the scrub already off** (`0`), so the scrub was never the gate — Claude Code strips its own Anthropic credentials from bash subprocesses independent of this knob. The refactor reframed the claim as "moot" but kept its disproved premise intact.

This reword keeps the two true facts and drops the false one:

- The scrub stays `0` for the unrelated Linux bwrap `!`→`\!` corruption bug (anthropics/claude-code#64301, write-up in #639).
- Credential hiding is moot regardless, because the agent's env now holds only dummy secrets and the proxy injects the real ones per-host.
- It no longer claims the scrub ever hid model auth from bash; per #637 the strip never depended on it.

Applied to both Claude harness actions. Comment-only — no behavior change.

<details><summary>Before / after</summary>

Before (`claude/action.yaml`):

```
# so the sandbox stays off (`0`). bwrap's value was hiding the model
# auth from the agent; that is now moot — the agent's env holds only
# dummy GitHub + Anthropic secrets, so there is nothing sensitive left
# for SUBPROCESS_ENV_SCRUB to hide.
```

After:

```
# so the subprocess sandbox stays off (`0`). [...]
# Turning it off costs no credential hiding: the agent's env holds only
# dummy GitHub + Anthropic secrets (the proxy injects the real ones
# per-host), so SUBPROCESS_ENV_SCRUB has nothing sensitive to scrub.
# #637 separately found the model auth already absent from Bash-tool
# subprocesses with the scrub off, so the strip never depended on it.
```

</details>

### DIFF
--- a/claude-interactive/action.yaml
+++ b/claude-interactive/action.yaml
@@ -274,12 +274,14 @@ runs:
         # skipDangerousModePermissionPrompt in settings.local.json). The
         # proxy injects the real credentials per-host.
         # Mirrors action.yaml: the Linux bwrap wrapping path corrupts `!`
-        # to `\!` in Bash commands, so the sandbox stays off (`0`). bwrap's
-        # value was hiding the model auth from the agent; that is now moot —
-        # the agent's env holds only dummy GitHub + Anthropic secrets, so
-        # there is nothing sensitive left for SUBPROCESS_ENV_SCRUB to hide.
+        # to `\!` in Bash commands, so the subprocess sandbox stays off (`0`).
         # Reproduced through claude 2.1.159; anthropics/claude-code#64301,
-        # write-up in #639.
+        # write-up in #639. Turning it off costs no credential hiding: the
+        # agent's env holds only dummy GitHub + Anthropic secrets (the proxy
+        # injects the real ones per-host), so SUBPROCESS_ENV_SCRUB has nothing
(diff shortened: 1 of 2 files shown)


max-sixty:
shorten and simplify

tend-agent:
Shortened both comments to the two load-bearing facts (bwrap `!`-corruption keeps the scrub off; off leaks nothing because the env carries only dummy secrets). Dropped the #637 relitigation and the longer hedging. Pushed in f48d6e8.

Issue #742, opened 2026-06-29
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-06-29T06:23:12Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/28352782724) | N/A |

This issue was created automatically. Close it once the outage is resolved.

tend-agent:
### [Run 28352782724](https://github.com/max-sixty/tend/actions/runs/28352782724)

#### review-reviewers (numbagg/numbagg)

```
Claude turn ended in failure (StopFailure hook fired — rate limit, auth, or server error)
```

#### review-reviewers (max-sixty/cargo-affected)

```
Claude turn ended in failure (StopFailure hook fired — rate limit, auth, or server error)
```

#### review-reviewers (PRQL/prql)

```
Claude turn ended in failure (StopFailure hook fired — rate limit, auth, or server error)
```

#### review-reviewers (max-sixty/tend)

```
Claude turn ended in failure (StopFailure hook fired — rate limit, auth, or server error)
```

#### review-reviewers (max-sixty/worktrunk)

```
Claude turn ended in failure (StopFailure hook fired — rate limit, auth, or server error)
```

tend-agent:
Duplicate of #743 — both track the same failed run [28352782724](https://github.com/max-sixty/tend/actions/runs/28352782724) (a transient `review-reviewers` StopFailure across all matrix legs). Closing in favor of #743, which is also now resolved.

Issue #576, opened 2026-05-22
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-05-22T04:19:41Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/26268072849) | N/A |

This issue was created automatically. Close it once the outage is resolved.

tend-agent:
Error details for [run 26268072849](https://github.com/max-sixty/tend/actions/runs/26268072849):

### review-reviewers (max-sixty/cargo-affected)

```
Action failed with error: SDK execution error: Error: Claude Code returned an error result: API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check status.claude.com.
```

### review-reviewers (PRQL/prql)

```
Action failed with error: SDK execution error: Error: Claude Code returned an error result: API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check status.claude.com.
```

### review-reviewers (max-sixty/tend)

```
Action failed with error: SDK execution error: Error: Claude Code returned an error result: API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check status.claude.com.
```

### review-reviewers (numbagg/numbagg)

```
Action failed with error: SDK execution error: Error: Claude Code returned an error result: API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check status.claude.com.
```

tend-agent:
Outage resolved — root cause was a transient Anthropic API 529 (Overloaded). Subsequent workflow runs (e.g., review-reviewers at 2026-05-22T06:12:33Z) completed successfully.

PR #744, opened 2026-06-29
fix(actions): self-heal duplicate tend-outage issues from concurrent leg failures

### LINKED ISSUE #743: Bot temporarily unavailable

| When | Run | Trigger |
|------|-----|---------|
| 2026-06-29T06:23:17Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/28352782724) | N/A |

tend-agent:
## Problem

The outage-tracker dedup in `shared/steps/report-failure.sh` is racy, and the jitter added in #586 to fix it has now been proven insufficient in production.

When several matrix legs fail within a few seconds of each other (the common case for an upstream model-API 5xx/overload incident, which hits every leg at once), each leg runs the check-then-create dedup. The `sleep $((RANDOM % 30))` jitter only *narrows* the create-create window — it can't close it. Two legs can still both read `EXISTING` as empty: jitter collisions are likely with several legs drawing from a 30-slot range (birthday problem), and GitHub's issue-list index takes a few seconds to reflect a freshly-created issue, so even a staggered second leg can miss the first's issue.

## Evidence

- **2026-05-22**: four duplicate `tend-outage` issues (#575, #576, #577, #578) created within 4 seconds. This prompted #586 (jitter).
- **2026-06-29**: the jitter mitigation failed — an all-five-legs `review-reviewers` failure ([run 28352782724](https://github.com/max-sixty/tend/actions/runs/28352782724), transient `529 Overloaded`/StopFailure) produced **two** duplicate trackers #742 and #743, created within 1 second (06:23:30 and 06:23:31), both referencing the same run.

Every prior incident with only 1–2 failing legs produced a single tracker, so the failure mode scales with the number of simultaneously-failing legs. Impact is low (both dupes were auto-closed by nightly ~2h later when the outage resolved), but it's a recurring structural failure of a deployed fix, and it generates duplicate "Bot temporarily unavailable" notifications during every wide outage.

## Fix

Keep the jitter (it still reduces contention and the number of dupes created) and add a self-healing reconciliation to the create path: after `gh issue create`, settle briefly for the list index, then list every open `tend-outage` issue, keep the lowest-numbered, and close the rest as duplicates. It is idempotent and convergent — every racing leg computes the same keeper, so a second leg closing an already-closed dup is a no-op.

This is shared verbatim by all three harness actions, so it reaches every consumer at the next release.

- `shellcheck shared/steps/report-failure.sh` passes.
- Logic unit-tested against today's numbers: `[{742},{743}]` → keep #742, close #743; single-issue input closes nothing.
- Index-lag safety: the 5s settle plus the convergent keep-lowest rule means even if leg B's first re-list misses leg A's issue, both legs eventually compute the same keeper and the loser is closed.

</details>

Found by `review-runs` analyzing [run 28361647677](https://github.com/max-sixty/tend/actions/runs/28361647677).

### DIFF
--- a/shared/steps/report-failure.sh
+++ b/shared/steps/report-failure.sh
@@ -52,4 +52,21 @@ if [ -n "$EXISTING" ]; then
   gh issue comment "$EXISTING" -F /tmp/comment.md
 else
   gh issue create --title "$TITLE" --label "$LABEL" -F /tmp/body.md
+
+  # The jitter above only narrows the create-create race; it can't close it.
+  # Two legs can still both read $EXISTING empty (jitter collision within the
+  # few-second window the list index takes to reflect a fresh create), so each
+  # files its own issue. Reconcile after creating: settle for the index, list
+  # every open tend-outage issue, keep the lowest-numbered, and close the rest
+  # as duplicates. Idempotent and convergent — every racing leg computes the
+  # same keeper, so a second leg closing an already-closed dup is a no-op.
+  sleep 5
+  OPEN=$(gh issue list --label "$LABEL" --state open --json number --jq 'sort_by(.number) | .[].number')
+  KEEP=$(echo "$OPEN" | head -1)
+  echo "$OPEN" | tail -n +2 | while read -r DUP; do
+    [ -z "$DUP" ] && continue
+    gh issue close "$DUP" \
+      --comment "Duplicate of #${KEEP} (concurrent matrix-leg failure); consolidating outage tracking there." \
+      2>/dev/null || true
+  done
 fi


tend-agent (commented):
The convergent keep-lowest reconciliation is sound, and it's loop-safe — `tend-triage`/`tend-mention` both skip `tend-outage`-labeled issues, so the close-with-comment won't cascade. For the documented trigger (matrix legs of one run) there's no data loss, since all legs share `GITHUB_RUN_ID` and thus the same run link.

One edge to consider: the dedup is label-scoped and workflow-agnostic, so it also fires when duplicates come from *different* workflow runs — e.g. a `tend-review` and a `tend-triage` both passing the empty-`EXISTING` check within the same overload window. There the closed higher-numbered issue carries a *distinct* run link, and the keeper never accumulates it: `enrich-tend-outage-issues.sh` only iterates `--state open`, so the closed dup's run is never enriched onto the keeper. The run is still one click away via the "Duplicate of #KEEP" comment, so it's recoverable rather than lost — just not consolidated.

If that's worth closing, the cheap fix is to append the dup's run reference as a comment on `KEEP` before closing it (read the dup's body for its run URL), rather than only commenting on the dup. Not blocking — the matrix-leg case this targets is fully handled.

Submitting as a comment since GitHub rejects self-approval.

max-sixty:
or more jitter?

tend-agent:
More jitter can't close this race — it only lowers the probability. The PR keeps the jitter for exactly that reason (fewer dupes created), but a deterministic reconcile is what actually fixes it. Two reasons jitter alone falls short:

1. **Birthday collisions scale with legs.** With the current `RANDOM % 30` and 5 simultaneously-failing legs, P(some pair within the same slot) ≈ 30%. To push that low you'd need a range of minutes, which delays outage reporting on *every* failure to fix a rare one.

2. **Index lag is the dominant factor, and jitter can't bound it.** Even legs that draw distinct slots can both read `EXISTING` empty, because GitHub's issue-list index takes a few seconds to reflect a fresh create. The 06-29 dupes (#742/#743) landed 1s apart — within that lag window. To beat lag with jitter alone, every leg would need spacing greater than the (variable, uncontrolled) index latency, which random jitter can't guarantee.

So jitter is a probability knob; the reconcile is convergent — every racing leg computes the same keeper, so the window is closed rather than narrowed, at the cost of one `gh issue list` + closes on the create path only. Happy to drop the jitter entirely if you'd rather not keep both, but I'd lean on keeping it since it cheaply reduces how many dupes the reconcile has to clean up.

Issue #166, opened 2026-04-07
Allow tend-mention to trigger on bot's own reviews


tend-agent:
## Problem

The `tend-mention` workflow's job-level `if` filter excludes all events from the bot account (`github.event.review.user.login != '<bot>'`). This means when the bot leaves a review with suggestions on its own PR, no workflow fires to act on that review — a human must intervene or explicitly mention the bot.

Context: #163 (comment) — the bot's self-review on its own PR was skipped entirely.

## Where the filter lives

The filter is generated in [`generator/src/tend/workflows.py` ~L192–201](https://github.com/max-sixty/tend/blob/main/generator/src/tend/workflows.py#L192-L201) and produces the job-level `if` on the `verify` job in `tend-mention.yaml`.

### A. Remove the bot-author filter entirely, rely on the agent to self-limit

Remove the `user.login != bot` checks from the `verify` job `if` condition. The agent already has "exit silently if nothing useful to contribute" guidance in its prompt and skills — this becomes the only loop-prevention mechanism.

**Pros:** Simplest change. The bot can respond to any review on its PRs, including its own.
**Cons:** Relies entirely on the agent's judgment to avoid loops. A chatty agent could trigger itself repeatedly. No hard backstop.

### B. Allow bot-authored events only on bot-authored PRs

Keep the filter for issues and non-bot PRs, but relax it when the PR author is the bot. The `verify` job already checks PR authorship — extend it to also pass bot-authored events through when the PR is the bot's own.

**Pros:** Scoped — only affects bot PRs where self-interaction is expected. Human PRs are unaffected.
**Cons:** Doesn't cover the case where the bot reviews a human's PR and wants to follow up on its own review thread. More complex filter logic.

### C. Allow bot-authored events, add a hard recursion limit

Remove the bot-author filter (like option A), but add a recursion counter. The workflow checks how many times the bot has already triggered `tend-mention` on this PR/issue in the last N minutes. If over a threshold (e.g., 3), skip.

**Pros:** Hard backstop against loops regardless of agent behavior.
**Cons:** Most complex. Requires API calls to count recent workflow runs in the `verify` job, adding latency and complexity.

## Recommendation

Option A is the simplest starting point. The existing prompt guidance ("exit silently if the bot already responded to the trigger") and the concurrency group (`cancel-in-progress: false`) already provide soft guardrails. If loops emerge in practice, option C can be layered on later.

max-sixty:
@tend-agent:
- Do (A)
[...]

PR #749, opened 2026-06-29
fix: skip no-op tend-mention session on bot's own APPROVED review

### LINKED ISSUE #747: tend-mention: no-op session on bot's own APPROVED review of human PRs
## Summary

`tend-mention` spins up a full (billable) agent session every time **the bot leaves an empty-body `APPROVED` review on a human-authored PR**. The session can never produce an action — it reads the review, recognizes its own approval, and exits silently. This is a distinct leg of the no-op-session family tracked in #95: it is **not** covered by either fix filed there (#606 = `issue_comment` leg, shipped; #607 = `tend-review` self-review leg, still open).

The fix lives in the generated `tend-mention.yaml` (generator source `generator/src/tend/workflows.py`). Surfaced by the downstream diffplug/dormouse `review-runs` sweep ([run 27899493170](https://github.com/diffplug/dormouse/actions/runs/27899493170)); filed here with the downstream maintainer's go-ahead.

## Evidence

Three occurrences in a single 24h window, one per human PR the bot approved (all on diffplug/dormouse, bot login `dormouse-bot`):

| Run | PR (author) | Trigger | Review | Outcome | Cost |
|-----|-------------|---------|--------|---------|------|
| [27875120333](https://github.com/diffplug/dormouse/actions/runs/27875120333) | #158 (human) | `pull_request_review` | empty-body `APPROVED` (id 4538038926) | exited silently | $0.28 |
| [27879038635](https://github.com/diffplug/dormouse/actions/runs/27879038635) | #163 (human) | `pull_request_review` | empty-body `APPROVED` (id 4538477275) | exited silently | $0.11 |
| [27883636959](https://github.com/diffplug/dormouse/actions/runs/27883636959) | #164 (human) | `pull_request_review` | empty-body `APPROVED` (id 4538814194) | exited silently | $0.10 |

~$0.49 for the window, scaling linearly with how many human PRs the bot reviews. Still fires at the generated workflow version shipped with tend 0.1.6.

## Root cause

For a `pull_request_review` event the `verify` gate falls through to a "has the bot engaged on this PR?" heuristic ([tend-mention.yaml#L124-L127 in the downstream copy](https://github.com/diffplug/dormouse/blob/60e82ecc6226d9595d93f9c819298ade47bb797f/.github/workflows/tend-mention.yaml#L124-L127)):

```bash
  --jq '[.[] | select(.user.login == "dormouse-bot")] | length')
if [ "$BOT_REVIEWS" -gt "0" ]; then
  echo "should_run=true" >> "$GITHUB_OUTPUT"; exit 0
fi
```

When the triggering event **is** the bot's own review, that review is itself counted as engagement, so `BOT_REVIEWS > 0` is always true and the gate returns `should_run=true` → a no-op session. The `issue_comment` path has an analogous early-exit (`COMMENT_AUTHOR_TYPE == "Bot"` → skip); the `pull_request_review` path has none.

## The #166 constraint (why not just filter all bot reviews)

#166 (closed/implemented) **deliberately removed** the blanket `review.user.login != bot` filter so the bot *can* act on its own **suggestion** reviews on its **own** PRs. A naive "skip all bot-authored reviews" would re-break that. The no-op is specifically a bot-authored **`APPROVED`** review — an approval is terminal, there is nothing to act on, regardless of PR author. A bot `CHANGES_REQUESTED`/`COMMENTED` review with a body still has actionable content and should keep firing.

## Proposed fix

Add an early-exit alongside the existing `issue_comment` bot-skip, gated on review **state** so #166's actionable self-reviews are preserved:

```bash
# A bot-authored APPROVED review is terminal — nothing to act on. Bot reviews
# *with suggestions* on its own PR still fire (see #166), so gate on state,
# not just author.
if [ "$EVENT_NAME" = "pull_request_review" ] \
   && [ "$REVIEW_AUTHOR" = "$BOT_LOGIN" ] \
   && [ "$REVIEW_STATE" = "approved" ]; then
  echo "should_run=false" >> "$GITHUB_OUTPUT"; exit 0
fi
```

with two additions to the step `env:` block:

```yaml
REVIEW_AUTHOR: ${{ github.event.review.user.login }}
REVIEW_STATE: ${{ github.event.review.state }}
```

(`$BOT_LOGIN` is the already-templated bot login the generator substitutes elsewhere in the gate. Generator source is `generator/src/tend/workflows.py`, the same file #166 touched.)

## Dup-check

#606 / #607 / #166 each address a different leg; no open issue covers the bot's-own-`APPROVED`-review case.

tend-agent:
## Problem

`tend-mention` spins up a billable agent session every time the bot leaves an empty-body `APPROVED` review on a human PR. The session can only read its own approval and exit silently — a no-op (issue #747).

Root cause: for a `pull_request_review` event the verify gate falls through to a "has the bot engaged on this PR?" heuristic that queries `repos/.../pulls/<n>/reviews` and counts bot-authored reviews. When the *triggering* event is the bot's own review, that review is itself counted (`BOT_REVIEWS > 0`), so the gate always returns `should_run=true`. The `issue_comment` path has an analogous bot early-exit; the `pull_request_review` path had none.

## Solution

Add an early-exit in the verify gate (`generator/src/tend/templates/mention.yaml.j2`) alongside the existing `issue_comment` bot-skip, gated on review **state** rather than author alone:

```bash
if [ "$EVENT_NAME" = "pull_request_review" ] \
   && [ "$REVIEW_AUTHOR" = "<bot>" ] \
   && [ "$REVIEW_STATE" = "approved" ]; then
  echo "should_run=false" >> "$GITHUB_OUTPUT"; exit 0
fi
```

plus `REVIEW_AUTHOR`/`REVIEW_STATE` in the step `env:`.

Gating on state, not just author, preserves #166: bot self-reviews carrying suggestions on the bot's own PR still have actionable content and keep firing. Only the terminal `approved` state short-circuits. `review.state` arrives **lowercase** in the webhook payload (REST returns uppercase) — verified against GitHub's documented `github.event.review.state == 'approved'` example — hence the lowercase comparison.

## Testing

- Added `test_mention_skips_bot_approved_review`, which renders the mention workflow and asserts the verify gate skips a bot-authored `approved` review *before* reaching the `BOT_REVIEWS` engagement query, and wires `REVIEW_AUTHOR`/`REVIEW_STATE` into the env block. It fails on `main` (reproduction) and passes with the fix.
- Updated the three `mention` regtest snapshots.
- Full generator suite green (256 passed).

---
Closes #747 — automated triage

### DIFF
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -95,15 +95,20 @@ jobs:
             exit 0
           fi
 
-          # The bot's own APPROVED review is terminal — nothing to act on. The
-          # engagement heuristic below counts this very review (BOT_REVIEWS > 0)
-          # and would spin up a no-op session. Gate on review *state*, not just
-          # author: bot self-reviews carrying suggestions on the bot's own PR
-          # still need to fire. review.state arrives lowercase in the webhook
-          # payload (REST returns uppercase), hence "approved".
+          # The bot's own empty-body APPROVED review is terminal — nothing to
+          # act on. The engagement heuristic below counts this very review
+          # (BOT_REVIEWS > 0) and would spin up a no-op session. Gate on three
+          # things, not just author: review *state* (an approval is terminal; a
+          # bot CHANGES_REQUESTED/COMMENTED review keeps actionable signal) and
+          # an empty *body* (an approval carrying nits in its body may warrant
+          # follow-up changes, so let it fire). Together these preserve #166's
+          # actionable bot self-reviews. review.state arrives lowercase in the
+          # webhook payload (REST returns uppercase), hence "approved".
+          # COMMENT_BODY holds review.body for a pull_request_review event.
           if [ "$EVENT_NAME" = "pull_request_review" ] \
              && [ "$REVIEW_AUTHOR" = "<<cfg.bot_name>>" ] \
-             && [ "$REVIEW_STATE" = "approved" ]; then
+             && [ "$REVIEW_STATE" = "approved" ] \
+             && [ -z "$COMMENT_BODY" ]; then
             echo "should_run=false" >> "$GITHUB_OUTPUT"
             exit 0
           fi
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -95,6 +95,19 @@ jobs:
             exit 0
           fi
 
+          # The bot's own APPROVED review is terminal — nothing to act on. The
+          # engagement heuristic below counts this very review (BOT_REVIEWS > 0)
+          # and would spin up a no-op session. Gate on review *state*, not just
+          # author: bot self-reviews carrying suggestions on the bot's own PR
+          # still need to fire. review.state arrives lowercase in the webhook
+          # payload (REST returns uppercase), hence "approved".
+          if [ "$EVENT_NAME" = "pull_request_review" ] \
+             && [ "$REVIEW_AUTHOR" = "<<cfg.bot_name>>" ] \
+             && [ "$REVIEW_STATE" = "approved" ]; then
+            echo "should_run=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
           # pull_request_review payloads include review.body (checked above)
           # but NOT the bodies of inline comments attached to the review.
           # Fetch them so a first-contact @-mention inside an inline comment
@@ -160,6 +173,8 @@ jobs:
           PR_URL: ${{ github.event.issue.pull_request.url }}
           EVENT_PR_NUMBER: ${{ github.event.pull_request.number }}
           REVIEW_ID: ${{ github.event.review.id }}
+          REVIEW_AUTHOR: ${{ github.event.review.user.login }}
+          REVIEW_STATE: ${{ github.event.review.state }}
 
       - name: React to mention
         if: |
--- a/generator/tests/_regtest_outputs/test_generate.test_workflow_minimal_codex_regtest[mention].out
+++ b/generator/tests/_regtest_outputs/test_generate.test_workflow_minimal_codex_regtest[mention].out
@@ -80,15 +80,20 @@ jobs:
             exit 0
           fi
 
-          # The bot's own APPROVED review is terminal ? nothing to act on. The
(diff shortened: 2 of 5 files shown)


tend-agent:
CI status note: the required jobs (`test`, `lint`, `test-proxy`) have been `queued` with no assigned runner for ~27 min — runner-availability backlog, not a result of this change (`test-worker` ran and passed). They'll start when runners free up.

Locally these mirror green: full generator suite `256 passed` (including the new `test_mention_skips_bot_approved_review`, which fails on `main` and passes here), and `pre-commit` (ruff, typos, actionlint, shellcheck) clean on the changed files.

tend-agent:
Pushed 6de3228 (empty-body gate, per maintainer feedback on #747). Locally green — full generator suite `256 passed` (including the updated `test_mention_skips_bot_approved_review`) and `pre-commit` clean on the changed files. PR CI for this commit is still `QUEUED` behind runner availability (`test`, `lint`, `test-proxy`, `test-worker`, `review` all unassigned), so I couldn't observe terminal results within the session — they're unverified by me. Worth a glance before merge.

tend-agent (commented):
The fix is correct and well-scoped for the empty-body APPROVED case: the four-way gate (event + bot author + `approved` + empty body) is narrow enough to skip only the no-op, the lowercase `approved` comparison matches the webhook payload (REST would be uppercase), and placing it *before* the `PR_AUTHOR == bot` short-circuit while gating on state is what preserves #166 — a bot COMMENT self-review on its own PR still falls through to line 153 and fires.

One related residual, not a blocker: the same no-op class still fires for the bot's own **COMMENT / CHANGES_REQUESTED** reviews on human-authored PRs — which is `tend-review`'s primary output (a review with inline suggestions is a COMMENT-event review). Such a review falls through the new gate (`REVIEW_STATE != approved`), the `PR_AUTHOR == bot` check is skipped (author is human), and the `BOT_REVIEWS` query then counts the just-submitted review (`> 0`), so `should_run=true` and a session spins up. It's only kept output-free by the soft prompt/skill self-loop guard, not a structural gate — so it's a recurring no-op session rather than a runaway loop.

Closing that cleanly is harder than this PR (it needs the PR-author identity to distinguish a self-review on a *human* PR from one on the *bot's own* PR, where #166 says it should fire), so it's reasonably left as a follow-up rather than folded in here. Flagging so the remaining waste is tracked, not lost.

PR #751, opened 2026-07-01
fix(mention): skip no-op tend-mention session on bot's own comments

### LINKED ISSUE #756: tend-mention: review-runs tracking-issue comment edit self-triggers a daily no-op session (User-type PAT bot escapes the #606 skip)
## Summary

Every `review-runs` run appends its findings to the monthly `review-runs-tracking` issue by **editing the bot's existing comment** (a `PATCH`). Because that issue is **bot-authored** and the bot is a **User-type PAT account** (not a GitHub App), the edit falls through `tend-mention`'s "Verify bot engagement" gate to `should_run=true` and spins up a full no-op `tend-mention` session. The session's self-loop guard correctly exits without posting — so there's no wrong action — but it's pure daily waste: one runner + a Claude session (~$0.19–0.70 API-list) fires **every day** the review-runs (or review-reviewers) tracking comment is edited, and no run ever notices because it never surfaces as a CI failure.

Observed on the `max-sixty/worktrunk` deployment: a `claude-session-logs-n3349` artifact (issue #3349 is that repo's `review-runs-tracking: 2026-07`) exists for **every day 2026-07-01 through 2026-07-07** — seven consecutive daily no-op mention sessions, each `issue_comment` / `success`. This is generic: every tend consumer running `review-runs`/`review-reviewers` accumulates a bot-authored monthly tracking issue that gets edited daily, so every consumer hits it.

## Root cause

The `tend-mention` "Verify bot engagement" gate evaluates an `issue_comment` `edited` event on the tracking issue in this order:

1. `COMMENT_BODY` contains `@<bot>` → run. (Fires from the moment a findings entry quotes the bot handle — e.g. a quoted ``@worktrunk-bot resolve conflicts`` from analyzing a prior mention — since the quote persists in the comment and every subsequent edit re-matches.)
2. `issue_comment` **and** `COMMENT_AUTHOR_TYPE == "Bot"` → skip. **This is the `#606` skip, and it is a no-op for User-type PAT bots** — a `worktrunk-bot`-style PAT account reports `type: "User"`, not `"Bot"`, so the intended "skip undirected bot comments" guard never fires for the exact class of bot most tend deployments use.
3. (issue path) `ISSUE_AUTHOR == <bot>` → run. **The tracking issue is bot-authored, so this branch fires unconditionally** on any comment edit — even when path 1 doesn't match.

So the daily edit is caught by path 3 (bot-authored issue) regardless of the `@<bot>` quote, and once a findings entry quotes the handle, path 1 catches it too. Path 3 is intentional (per the `#665` fix — the bot *should* triage its own freshly-opened CI-failure/report issues), but it should not treat **the bot editing its own comment on its own machine-ledger issue** as a summons.

## Proposed fix

Add an early guard, **before** the `@<bot>` and bot-authored-issue checks: for an `issue_comment` event whose **comment author login == the bot's own login** and whose action is `edited` (equivalently, `github.event.sender.login == <bot>` editing its own comment), set `should_run=false`. Rationale: a bot editing its own comment is never a summons — the review-runs/review-reviewers daily tracking-issue append is the canonical case, and it also covers any future bot self-edit. This does not regress the intended behaviors:

- Human comments on a bot-authored report issue → sender is human → still runs (triage the report). ✓
- Bot **opens** a report issue (`issues` opened event) → still runs. ✓
- Bot edits its own tracking comment (review-runs append) → skipped. ✓

Secondarily, the `#606` skip in path 2 should key on the comment author **login** (`== <bot-login>`), not the author **type** (`== "Bot"`), so it actually fires for User-type PAT bots — the majority tend deployment shape. (Placing the new self-edit guard first makes this a belt-and-suspenders improvement rather than strictly required.)

An alternative, narrower fix — skip `issue_comment` events on issues carrying the `review-runs-tracking` label — also works but is review-runs-specific; the "bot editing its own comment is a no-op" guard is more general and closes the whole family (this sits alongside `#606`, `#607`, `#747`).

## Evidence

- Trigger: `tend-mention` run [28854230666](https://github.com/max-sixty/worktrunk/actions/runs/28854230666) (`claude-session-logs-n3349`), triggered by `max-sixty/worktrunk#3349` comment `4852460789` (the review-runs tracking comment, created 2026-07-01, `updated_at` 2026-07-07T08:58:40Z — the review-runs edit). The session text: *"triggered by a self-authored comment, and there's no actual directive to act on … Exiting silently."*
- Daily recurrence: `claude-session-logs-n3349` artifacts for runs `28854230666` (07-07), `28780829709` (07-06), `28735321616` (07-05), `28701029104` (07-04), `28649438004` (07-03), `28577450551` (07-02), `28506262260`/`28506210615` (07-01) — all `issue_comment` / `success` no-ops.
- Account type: `gh api users/worktrunk-bot --jq .type` → `"User"` (confirms the `#606` type-check is inert here).
- Gate source: `.github/workflows/tend-mention.yaml`, `Verify bot engagement` step (the `COMMENT_AUTHOR_TYPE == "Bot"` skip and the `ISSUE_AUTHOR == worktrunk-bot → should_run=true` branch).

Filed from `max-sixty/worktrunk`'s daily `review-runs` review under the standing bundled-defect path.

tend-agent:
## Problem

`tend-mention`'s `verify` gate has no self-loop guard for the bot's **own** comments. The existing bot-comment guard added in #608 only catches GitHub App / Bot accounts (`github.event.comment.user.type == "Bot"`); tend's bot is a PAT-based **User** account, so its own comments fall through to the engagement heuristics — either the bot-authored-issue short-circuit (`ISSUE_AUTHOR == bot`) or "bot has prior comments" — and reach the `handle` job, which then exits silently via the prompt's self-loop guard. A full agent session is spun up for nothing.

#608 explicitly noted this gap and left it: *"Tend's own bot accounts are PAT-based user accounts (`user.type == "User"`), so this does not affect tend-on-tend behavior."* The skill-level guard catches it, but only after paying for the session.

## Evidence

Observed every `review-runs` window and recorded in the June tracking issue (#646). The clearest recurring trigger is the **monthly tracking-issue rollover**: `review-reviewers` creates its month's tracking issue and posts one evidence-gist link comment per repo on it, each firing a self-loop `tend-mention` run.

The same shape also fires from `pull_request_review_comment` edits (subscribed for `edited`), which is why the guard covers that event too — closing #756.

## Fix

Add a guard in the `verify` script that skips comment events authored by the bot itself:

```bash
if { [ "$EVENT_NAME" = "issue_comment" ] || [ "$EVENT_NAME" = "pull_request_review_comment" ]; } \
   && [ "$COMMENT_AUTHOR" = "<bot_name>" ]; then
  exit 0
fi
```

`COMMENT_AUTHOR` is newly wired from `github.event.comment.user.login`.

Placed **before** the `@<bot>` mention check: a bot comment that quotes a prior `@`-mention would re-match the mention check and escape an after-placed guard, and there is no legitimate self-summons — the bot responding to its own mention is precisely the loop this guard breaks. It covers both comment event types that take the engagement path (`issue_comment` and `pull_request_review_comment`). The `pull_request_review` self-review paths carry intentional actionable-signal handling (the empty-body-APPROVED gate from #166/#608) and are left untouched.

## Policy reconciliation

Per the maintainer decision on the flagged policy conflict, the `review-reviewers` skill's `tend-mention` non-issue is narrowed: a deterministic verify-gate skip of the bot's *own comments* is now the accepted shape (it suppresses no legitimate work), distinct from the authorship-keyed guards still rejected for `tend-review` (where the bot's fix-push must be re-reviewed). The `review-runs` and `review` skills don't carry an echo of that rule, so no further edits were needed there.

## Testing

- `test_mention_verify_skips_self_authored_comments` asserts the guard is generated, `COMMENT_AUTHOR` is wired, the guard covers both `issue_comment` and `pull_request_review_comment`, and it sits **before** the mention check and before the bot-authored-issue short-circuit.
- `test_mention_self_comment_skip_spares_review_submissions` pins the carve-out: the skip is scoped to comment events and must **not** extend to the `pull_request_review` submission, so a bot self-review (which the prompt actions per #762) still reaches the actionable path. Guards against a later "skip self-authored reviews too" edit silently breaking the review → fix loop.
- Regenerated the mention regression snapshots; the guard now documents the deliberate `pull_request_review` boundary in-line.
- Full generator suite: 275 passed.

Closes #756


tend-agent (commented):
One scope gap worth a note (self-review): the guard fires only on `issue_comment`, but `pull_request_review_comment` (subscribed here for `edited` only) is the other event that carries `github.event.comment.*`, and a bot's own *edited* inline comment on a PR where it already has engagement takes the same generic-heuristic path (lines 166–181) into a no-op handle session. `COMMENT_AUTHOR` is already populated for that event, and the @-mention check on line 80 runs first, so a self-summons stays honored — extending the condition to `[ "$EVENT_NAME" = "issue_comment" ] || [ "$EVENT_NAME" = "pull_request_review_comment" ]` closes it for free.

Frequency is near-zero (the bot rarely edits its own inline comments, and `created` isn't subscribed), so this isn't a blocker — but the PR body groups this path under the "pull_request_review self-review paths [that] carry intentional actionable-signal handling," which isn't quite accurate: `pull_request_review_comment` has no such handling, it just falls through to the engagement heuristics. Worth either extending the guard or documenting the deliberate exclusion.

tend-agent:
Flagging a standing-guidance conflict for a maintainer to reconcile before this lands — not a blocker.

This PR's guard keys on comment authorship (`COMMENT_AUTHOR == <bot>`), which is the shape tend's own [`review-reviewers` non-issues list](https://github.com/max-sixty/tend/blob/main/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md) currently rules *not the accepted shape*. That entry names this exact scenario (the tracking-issue rollover self-loop, and the `pull_request_review` mirror variant) and says: "sender-based or commenter-based filters on `tend-mention` are not the accepted shape — they are the same authorship-keyed-guard pattern rejected for `tend-review`. The accepted shape for loop-prevention is a label-based skip … propose new filters only when there's a distinct loop risk that can't be expressed with a label."

Worth a maintainer decision because there's a real tension the reviewing bots didn't surface: the `tend-review` note above endorsed the guard and proposed *broadening* it, without referencing the standing ruling.

<details><summary>The reconciliation both ways</summary>

- **Toward label-based (the documented shape):** the highest-frequency trigger cited here — the monthly tracking-issue rollover — is fully label-expressible today: the issues carry `review-reviewers-tracking` / `review-runs-tracking`. A skip on those labels covers that subset without an authorship key.
- **Toward keeping this guard:** the general case ("bot comments on any thread it already participates in") isn't label-expressible, and the prompt's self-conversation guard already no-ops these — so a `issue_comment && author == bot` skip arguably suppresses zero legitimate work and is a pure cost optimization, which is a fair argument the non-issue doesn't fully rebut.

Either way it's a maintainer call: relax/scope the non-issue to bless this optimization, or land a label-based skip and add the ruling to the `review-runs` and `review` skills (neither of which carries it today, which is why the guard was proposed and endorsed without the conflict being noticed).

</details>

Surfaced by the hourly `review-reviewers` outcome audit; full evidence trail in the run's [evidence gist](https://gist.github.com/13fd3f24575eaf3e9edee7defa676265).

max-sixty:
@tend-agent Decision on the policy conflict you flagged: the deterministic skip is accepted here. Skipping the bot's own comments suppresses no legitimate work (unlike authorship-keyed guards on `tend-review`, where the bot's fix-push must still be re-reviewed), the verify script already carries authorship-keyed gates of the same shape (the Bot-type check, the empty-body-APPROVED gate), and CLAUDE.md's "Actions gate whether the agent runs at all" endorses exactly this kind of cheap YAML gate. Please rework:

1. **Move the guard before the @-mention check.** A bot comment that quotes a prior `@`-mention (the shape in #756) re-matches the mention check and escapes an after-placed guard, and there is no legitimate self-summons: the bot responding to its own mention is the loop this guard exists to break. Update the template comment and the ordering assertion in `test_mention_verify_skips_self_authored_comments`.

2. **Cover `pull_request_review_comment` too.** It's subscribed for `edited` and takes the same engagement-heuristic path; extend the guard's event condition.

3. **Narrow the `review-reviewers` skill's non-issue** (the `tend-mention` bullet under "Non-issues") so it no longer rejects this shape: authorship-keyed guards remain rejected for `tend-review`, but a deterministic verify-gate skip of the bot's own comments on `tend-mention` is accepted. Check the other skills for echoes of the same rule.

4. **Rebase onto main** (the branch predates #762's mention-prompt change) and regenerate the mention snapshots.

With the before-mention ordering this fully closes #756, so add `Closes #756` to the PR body.

tend-agent:
Reworked and pushed all four:

1. **Guard moved before the `@`-mention check** — a bot comment quoting a prior `@`-mention no longer escapes it, and the comment/ordering assertion in `test_mention_verify_skips_self_authored_comments` now asserts `self_guard_idx < mention_idx`.
2. **`pull_request_review_comment` covered** — the guard's event condition is now `issue_comment || pull_request_review_comment`; both carry `comment.user.login`. Test asserts both.
3. **`review-reviewers` non-issue narrowed** — the `tend-mention` bullet now blesses a deterministic verify-gate skip of the bot's *own comments* while keeping authorship-keyed guards rejected for `tend-review` (where the fix-push must still be re-reviewed). Checked `review-runs` and `review` for echoes — neither carries the rule, so no other edits.
4. **Rebased onto `main`** (past #762/#775) and regenerated the mention snapshots.

Added `Closes #756` to the PR body. Full generator suite: 274 passed.

PR #762, opened 2026-07-09
fix: respond to actionable bot self-reviews instead of self-loop-exiting

### LINKED ISSUE #761: tend-mention: bot-authored review with actionable findings on bot's own PR is silently swallowed by self-loop guards
## Problem

When `tend-review` and the PR author are the **same bot account**, an actionable self-review gets silently swallowed by the self-loop guards, even though `tend-mention` triggers correctly and Claude runs to completion. The bot's own review workflow surfaced a real, fixable finding on the bot's own PR, and the run read it and exited silently — the finding only got actioned because the maintainer asked by hand.

This is **distinct from #607**. #607 argues bot self-reviews are pure no-op waste (self-*approval* isn't allowed, so nothing to post) and should be short-circuited before the diff read. Here the opposite happened: `tend-review` posted a substantive `COMMENTED` review with a genuine finding, and the maintainer *wants* those actioned. The gap is downstream in `tend-mention`: the guard can't tell "the bot's **reviewer** role surfaced a finding" from "the bot is talking to itself."

## Evidence (max-sixty/worktrunk)

On [max-sixty/worktrunk#3394](https://github.com/max-sixty/worktrunk/pull/3394) (authored by `worktrunk-bot`):

- `tend-review` posted [review 4666198096](https://github.com/max-sixty/worktrunk/pull/3394#pullrequestreview-4666198096) — a `COMMENTED` review, authored by `worktrunk-bot`, at `2026-07-09T19:40:12Z`. It flagged a real `{{ base }}` validation-vs-runtime asymmetry (later fixed in `73b597131`).
- The `submitted` event fired `tend-mention` [run 29045164521](https://github.com/max-sixty/worktrunk/actions/runs/29045164521) at `19:40:15Z`. `verify` set `should_run=true` via the PR-author short-circuit; `handle` ran Claude from `19:40:25Z` to `19:43:15Z` (~2m50s), conclusion `success`. **The trigger worked and the model ran to completion — it just exited silently.**
- The finding was only actioned after the maintainer manually asked ([comment](https://github.com/max-sixty/worktrunk/pull/3394#issuecomment-4930039131)).

## Root cause — two overlapping guards, no reviewer/author carve-out

The non-mention `pull_request_review` branch of the `tend-mention` prompt reads ([tend-mention.yaml#L234-L235](https://github.com/max-sixty/worktrunk/blob/d422b1b9be2682dc814b3894f0ed895911da96f1/.github/workflows/tend-mention.yaml#L234-L235) as generated by tend 0.1.9):

> A review was submitted on a PR where you previously participated (…). Read the review and full context. If the review requests changes or asks questions, respond appropriately. **If the review approves or is between humans, exit silently.**

Combined with the harness-injected self-loop guard —

> Self-loop guard: before responding to a comment or review, confirm the triggering actor isn't the bot itself.

— a bot-authored review on the bot's own PR reads as "the bot talking to itself," so the model exits. The bundled `running-in-ci` **Self-conversation Guard** actually already carves out the right case:

> only respond if there is a distinct role boundary (e.g., you are the reviewer on your own PR and need to address review feedback).

…but that nuance is contradicted by the blanter harness guard and isn't reinforced by the prompt branch, so in practice the blanket guard wins.

## History — this never reliably worked; the intent since April flips the other way

- Support for `pull_request_review` in `tend-mention` was added April 1 (worktrunk #1882, tend 0.0.5) but the workflow-level `if` filtered bot reviews out entirely (`github.event.review.user.login != 'worktrunk-bot'`) — so a bot review never triggered a run.
- April 8 (worktrunk #2009) **removed** that filter (bot reviews now trigger) **and added the role-boundary carve-out** to the prompt — the stated intent being to *allow* the bot to respond to its own review when there's a reviewer/author boundary. So since April the design has meant to action these, but the overlapping guards defeat it.

## Proposed fix (tend-ci-runner)

1. **`tend-mention` prompt** — in the non-mention `pull_request_review` branch, add an explicit positive instruction: a bot-authored review on the bot's own PR that requests changes or asks questions is the **reviewer role** speaking (from `tend-review`), not a self-loop — treat it as actionable (make the change / answer, commit, push). Only exit silently when it's a plain approval or carries no actionable content.
2. **Reconcile the harness self-loop guard with the `running-in-ci` Self-conversation Guard** — the blanket "triggering actor isn't the bot itself" needs the same reviewer-vs-author role-boundary exception the bundled skill already documents, so it stops overriding the intended behavior.

Both surfaces are upstream generated/bundled (the workflow header directs changes here), so there's no clean local override in the consumer repo.

## Context

Filed at maintainer request on [max-sixty/worktrunk#3394](https://github.com/max-sixty/worktrunk/pull/3394#issuecomment-4930238320). Related but distinct: #607 (short-circuit no-op self-*reviews*).

tend-agent:
Closes #761.

## Problem

When `tend-review` and the PR author are the **same bot account** (tend-on-tend, or any consumer using one bot for both authoring and reviewing), an actionable self-review is silently swallowed. `tend-mention` triggers correctly and Claude runs to completion, but the run reads a `CHANGES_REQUESTED` / `COMMENTED`-with-findings review on the bot's own PR, treats it as the bot talking to itself, and exits without acting. The finding only gets actioned when a human asks by hand.

The verify gate is already correct — it fires `should_run=true` for an actionable self-review; #749 only skips *empty-body `APPROVED`* ones. The silent exit comes from two prompt-layer guards with no reviewer/author carve-out:

- The **system-prompt self-loop guard** was absolute: *"before responding to a comment or review, confirm the triggering actor isn't the bot itself."*
- The **non-mention `pull_request_review` prompt branch** said only *"respond appropriately … if the review approves or is between humans, exit silently"* — it never named the bot-authored-review case, so it read as "between humans / self."

`running-in-ci`'s **Self-conversation Guard** already documents the exception ("only respond if there is a distinct role boundary, e.g. you are the reviewer on your own PR"), but neither of the two surfaces above carried it, and the system prompt wins.

## Fix — both surfaces #761 names

1. **`shared/system-prompt.md`** — the self-loop guard now carves out the role boundary: an actionable review the bot's review workflow posted on the bot's own PR (`CHANGES_REQUESTED`, or `COMMENTED` with findings/questions) is reviewer-to-author feedback, not a self-loop — address it. A bot approval or an ask-free bot comment still exits silently.
2. **`generator/src/tend/templates/mention.yaml.j2`** — the non-mention review branch now positively instructs that a review your own review workflow posted on your own PR is your reviewer role speaking, so action it; exit silently only for a plain approval or a review with no actionable content.

The system prompt is composed at runtime by all three harness actions, so no workflow regeneration is needed for surface 1; surface 2 regenerates the mention regtest snapshots (updated here). Reaches adopters at the next release. All 256 generator tests pass.

## Relationship to other issues/PRs

- **#761** — the tracking issue this closes (filed in parallel from the same worktrunk thread).
- **#607 / #751** — the *opposite* premise: suppress no-op bot self-*comments* / empty self-*approvals* at the gate. This change is about an actionable self-*review* that found a real bug. Complementary — different event types and layers.

## Loop consideration

If addressing the review pushes a commit, `synchronize` re-runs `tend-review`, which posts a fresh review — the same re-trigger loop that already exists for human-authored reviews, bounded by review eventually approving (terminal, gate-skipped), the mention dedup guard, and the findings actually getting resolved. Triggering here is the intended behavior.

Requested by @max-sixty in [max-sixty/worktrunk#3394](https://github.com/max-sixty/worktrunk/pull/3394): *"let's fix that … We need to trigger in this case."*


generator/src/tend/templates/mention.yaml.j2
```diff
@@ -15,7 +15,7 @@
...
               || (github.event_name == 'pull_request_review'
-                && format('A review was submitted on a PR where you previously participated (PR #{0}, {1}, review ID {2}). Read the review and full context. If the review requests changes or asks questions, respond appropriately. If the review approves or is between humans, exit silently.', github.event.pull_request.number, github.event.review.html_url, github.event.review.id))
+                && format('A review was submitted on a PR where you previously participated (PR #{0}, {1}, review ID {2}). Read the review and full context. If it requests changes or asks questions, respond appropriately — this holds even when the review was authored by you: a review your review workflow posted on your own PR is your reviewer role speaking, not a self-loop, so action it (make the change and push, or answer). Exit silently only for a plain approval or a review with no actionable content.', github.event.pull_request.number, github.event.review.html_url, github.event.review.id))
```
tend-agent: The `only` here drops the between-participants exit the old text had (`or is between humans, exit silently`). Restore it so a human-to-human change request on a PR the bot merely participated in doesn't get actioned — the comment-fallback branch below still carries this off-ramp:

```suggestion
                && format('A review was submitted on a PR where you previously participated (PR #{0}, {1}, review ID {2}). Read the review and full context. If it requests changes or asks questions, respond appropriately — this holds even when the review was authored by you: a review your review workflow posted on your own PR is your reviewer role speaking, not a self-loop, so action it (make the change and push, or answer). Exit silently for a plain approval, a review with no actionable content, or a review between other participants that is not directed at you.', github.event.pull_request.number, github.event.review.html_url, github.event.review.id))
```

tend-agent (commented):
The self-review carve-out reads well and I verified the layering: the verify gate already lets an actionable bot self-review through (it skips only `approved` + empty-body), and the in-process render preserves the em-dash — the `?` in the regenerated snapshots is the usual pytest-regtest non-ASCII normalization, consistent with the em-dashes already in that file, so no functional concern there.

One thing the rewrite dropped, though: the review branch lost its **bystander off-ramp**. The old text was *"If the review approves or is between humans, exit silently"*; the new text is *"Exit silently **only** for a plain approval or a review with no actionable content."* The `only` forecloses the between-participants case — a human requesting changes on another human's PR where the bot previously participated (so the verify gate fires `should_run=true`) now falls under "requests changes → respond appropriately (make the change and push)", pulling the bot into a review it's a bystander to.

The sibling comment-fallback branch right below still keeps its off-ramp (*"If the conversation is between other participants, exit silently"*), so the two fallbacks are now asymmetric — which reads as an unintended side effect of the rewrite rather than an intended scope change. Suggestion inline restores the bystander exit while keeping the self-review carve-out.

Note: applying the suggestion changes the source template only — the 3 `test_workflow_*_regtest[mention]` snapshots need regenerating (`uv run pytest tests/ -k mention --regtest-reset`) in the same commit or the regtests fail.

generator/src/tend/templates/mention.yaml.j2
tend-agent: Good catch — applied in 7a4c0e1. The `only` did foreclose the between-participants case, leaving this branch asymmetric with the comment-fallback below (which still exits on a review between other participants). The final sentence now reads "Exit silently for a plain approval, a review with no actionable content, or a review between other participants that is not directed at you," restoring the bystander off-ramp while keeping the self-review carve-out. Regenerated the 3 `mention` regtest snapshots in the same commit; all 256 generator tests pass.

max-sixty:
but then the trigger hasn't changed; only the guidance? 

I'm less of a fan of the guidance; it seems overly verbose

tend-agent:
Right — the trigger and the verify gate are unchanged. The gate already fired `should_run=true` for these (#749 only skips empty-body `APPROVED` self-reviews); the bug was purely at the prompt/guidance layer, so guidance is all that moved.

Trimmed both surfaces in 36f7684 — same behavior, fewer words:

- **system-prompt guard:** 8 lines → 6, folded the carve-out into the guard sentence itself.
- **mention review branch:** dropped the parenthetical and the "your reviewer role speaking" restatement, kept the "including when you authored the review" clause and the bystander off-ramp.

256 tests pass, mention regtest snapshots regenerated.

Issue #766, opened 2026-07-11
sandbox: no adopter lever to add commands/PATH/env inside the sandbox (setup: only reaches the runner)


tend-agent:
## Ask

A way for an adopter repo to run its own setup commands — or add to `PATH`/env — **inside the tend sandbox**, before the agent launches. Raised by @max-sixty in [max-sixty/worktrunk#3361](https://github.com/max-sixty/worktrunk/pull/3361#issuecomment-4942173834): *"can we have some way of repos adding commands to be executed in the sandbox?"*

## Why the existing `setup:` doesn't cover this

`setup:` in `.config/tend.yaml` renders as **runner-side** workflow steps ([`generator/src/tend/workflows.py` `_setup_yaml`](https://github.com/max-sixty/tend/blob/e127227e915f2444eab17b6b567be8ff70fbe133/generator/src/tend/workflows.py#L104)). Those run as the runner user, around the composite action — they don't reach the agent's environment.

The sandbox launches with a **fixed env allowlist**, not the runner's environment. `proxy/setup-sandbox.sh` writes `AGENT_ENV_FILE` with a hardcoded `PATH` ([setup-sandbox.sh#L94-L114](https://github.com/max-sixty/tend/blob/e127227e915f2444eab17b6b567be8ff70fbe133/proxy/setup-sandbox.sh#L94-L114)):

```
PATH=${AGENT_HOME}/.local/bin:/usr/local/bin:/usr/bin:/bin
```

and the agent is launched with exactly that file, nothing merged in ([install-tend-plugins.sh#L25-L26](https://github.com/max-sixty/tend/blob/e127227e915f2444eab17b6b567be8ff70fbe133/shared/steps/install-tend-plugins.sh#L25)):

```bash
mapfile -t AGENT_ENV <"$AGENT_ENV_FILE"
sudo -u "$SANDBOX" env "${AGENT_ENV[@]}" …
```

`GITHUB_ENV`/`GITHUB_PATH` are deliberately excluded from the handoff (so the sandbox can't write into later runner steps). Net: there is currently **no adopter-controlled lever that reaches inside the sandbox** — the runner-side `setup:` and the sandbox env are disjoint.

## Motivating case: cargo off `PATH`

[...]

PR #767, opened 2026-07-11
fix(sandbox): derive the sandbox PATH from the runner's PATH

tend-agent:
## Problem

The sandbox launches with a hardcoded PATH (`${AGENT_HOME}/.local/bin:/usr/local/bin:/usr/bin:/bin`) that omits every toolchain the runner image seeds into user homes. A Rust adopter's first bare `cargo`/`rustc` fails with `command not found`; so does a .NET adopter's `dotnet`, a Haskell adopter's `ghc`, and so on. Hardcoding one dir per language (`.cargo/bin`, then `.dotnet/tools`, then `.ghcup/bin`, …) would be an implicit, ever-growing allowlist tend doesn't otherwise care about.

## Solution

Derive the sandbox PATH from the **runner's own PATH** instead of hardcoding a base set. `setup-sandbox.sh` runs as the privileged `runner` user, so its `$PATH` already lists every dir the runner image put on PATH — the `/etc/skel`-seeded language homes (`.cargo/bin`, `.dotnet/tools`, `.config/composer/vendor/bin`, …) plus system dirs (`/opt/pipx_bin`, `/usr/local/.ghcup/bin`, `/usr/bin`, …). `useradd -m` copied `/etc/skel` into the sandbox's **own** home, so each runner-home toolchain has a sibling under `${AGENT_HOME}`.

The translation is a home-prefix rewrite: for each PATH entry, rewrite a leading `/home/runner` → `${AGENT_HOME}`, and keep the entry only when it now exists **and** the sandbox UID can traverse it. Every seeded toolchain comes across on one rule — cargo, dotnet, ghcup, composer alike — with no per-language code and no allowlist to grow.

### Security boundary

This works precisely because the rewritten paths point at the sandbox's **own** copies of the seeded dirs. The rewrite only ever maps runner-home prefixes to `${AGENT_HOME}`; a raw `/home/runner` path is never kept on the sandbox PATH:

- a bare `${runner_home}` entry is dropped outright, and
- every kept dir is re-tested with `sudo -u "$SANDBOX" test -x` — so any un-rewritten runner-home path, which the sandbox UID can't enter, can never survive.

The runner home is where the real credentials live; the sandbox must not be able to traverse it. Rewrite prefixes, never widen access.

## Scope / what this does *not* cover

Tools an adopter installs at runtime in a runner-side `setup:` step (`cargo-nextest`, `cargo-insta`, …) land in the **runner's** home, not `/etc/skel`, so they're absent from the sandbox home and the rewrite finds nothing to point at. Reaching those still needs the general sandbox-setup lever tracked in #766. Hence `Refs #766`, not `Closes`.

## Testing

- Simulated the derivation against a synthetic runner PATH (mixed rewritten toolchains, system dirs, a bare runner-home entry, and a duplicate): confirmed cargo/dotnet/composer rewrite to their sandbox copies, system dirs (ghcup, pipx) carry over as-is, the raw runner-home path is dropped, entries dedup, `.local/bin` stays first, and the block is `set -euo pipefail`-safe.
- Verified against the live runner image that `/etc/environment`'s PATH is *resolved* (`/home/runner/.cargo/bin`, not literal `$HOME/...`), so the prefix rewrite matches, and that `/etc/skel` seeds `.cargo`/`.dotnet`/`.ghcup` into the sandbox home.
- `shellcheck -S warning proxy/setup-sandbox.sh` clean.

---
Refs #766 — automated triage


max-sixty:
is `.cargo/bin` canonical in some way? i.e. why add this but not others? Is adding this signing up to add every arbitrary path to our PATH?

tend-agent:
Good challenge — and checking it live, you're right to be wary. `.cargo/bin` is **not** canonical in any way that singles it out.

The GitHub runner image seeds a batch of toolchain homes into `/etc/skel`, which `useradd -m` copies into the sandbox home. Rust is just the one that expose a `bin/` dir *and* that I happened to hit (worktrunk is Rust). At least two others are in exactly the same position — present in the sandbox home, off the sandbox PATH, first bare command fails with `command not found`:

<details><summary>Verified in the sandbox (<code>/home/tend-sandbox</code>) on the current ubuntu runner image</summary>

```
$ ls -d $HOME/.cargo/bin $HOME/.dotnet/tools $HOME/.ghcup/bin
/home/tend-sandbox/.cargo/bin      # cargo, rustc, clippy-driver, rustfmt (rustup shims)
/home/tend-sandbox/.dotnet/tools   # dotnet global tools (nbgv, …)
/home/tend-sandbox/.ghcup/bin      # ghc, cabal (Haskell)

$ echo "$PATH" | tr ':' '\n' | grep -iE 'cargo|dotnet|ghcup'
(none)
```

`/etc/skel` on this image also carries `.rustup`, `.nvm`, `.composer`, `.dotnet` — the seeded set, not a Rust special case.

</details>

So the PR's own premise — "present in the sandbox regardless of the adopter" — is true of cargo but *equally* true of dotnet and ghcup. It doesn't distinguish cargo; it just picks the one I noticed. Appending `.cargo/bin` alone is signing up for "why not `.dotnet/tools`, why not `.ghcup/bin`" the moment a .NET or Haskell adopter hits the same wall, and tend ends up maintaining an implicit, growing allowlist of language toolchains it doesn't otherwise care about.

The clean answer is the general lever already scoped as Option 2 in #766 — `sandbox_path:` in `.config/tend.yaml`, threaded into the sandbox env. Then the adopter declares exactly what their sandbox PATH needs (`sandbox_path: [~/.cargo/bin]`) and tend hardcodes nothing per-language. That subsumes this PR entirely.

My suggestion: close this in favor of building the #766 lever, rather than merge a per-toolchain special case. The one counter-argument for merging is zero-config for Rust adopters specifically — but by the same logic we'd owe dotnet/haskell/etc. the same, and the general lever gives all of them a single opt-in instead. Happy to redirect this branch to implement `sandbox_path:` if you'd like it as the next step.

max-sixty:
> So the PR's own premise — "present in the sandbox regardless of the adopter" — is true of cargo but equally true of dotnet and ghcup. It doesn't distinguish cargo; it just picks the one I noticed. Appending .cargo/bin alone is signing up for "why not .dotnet/tools, why not .ghcup/bin" the moment a .NET or Haskell adopter hits the same wall, and tend ends up maintaining an implicit, growing allowlist of language toolchains it doesn't otherwise care about.

yes, right

is there a way we can bring all.the pathe entries in from outside the sandbox?

tend-agent:
Yes — and it lands cleanly because of *where* `setup-sandbox.sh` runs. That script executes as the privileged `runner` user, so its `$PATH` **is** the runner's full PATH — every `/etc/skel`-seeded toolchain dir (`.cargo/bin`, `.dotnet/tools`, `.ghcup/bin`, `.nvm/...`, …) that the runner image put on PATH is already sitting in that variable. We just have to translate it for the sandbox instead of hardcoding one dir.

The translation is a home-prefix rewrite: for each entry, rewrite a leading `/home/runner` → `${AGENT_HOME}`, keep the entry only if it now exists and the sandbox UID can traverse it, and use that as `AGENT_PATH`. Sketch:

```bash
runner_home="$HOME"                      # setup runs as runner
declare -a keep=()
IFS=: read -ra dirs <<<"$PATH"
for d in "${dirs[@]}"; do
  case "$d" in "$runner_home"/*) d="${AGENT_HOME}/${d#"$runner_home"/}";; esac
  [ -d "$d" ] && sudo -u "$SANDBOX" test -x "$d" && keep+=("$d")
done
AGENT_PATH="${AGENT_HOME}/.local/bin:$(IFS=:; echo "${keep[*]}"):/usr/local/bin:/usr/bin:/bin"
```

That pulls **every** seeded toolchain across with zero per-language code — cargo, dotnet, ghcup, nvm all come in on the same rule, and the "why cargo but not X" objection disappears.

**One security line we must not cross.** This works precisely because `useradd -m` copied `/etc/skel` into the *sandbox's own home*, so the rewritten paths point at the sandbox's copies. "Bringing paths in from outside" must mean *rewriting to those copies* — **not** granting the sandbox read/traverse into `/home/runner` itself. The latter would breach the isolation boundary the proxy exists to enforce (the runner home is where the real credentials live). So: rewrite prefixes, never widen access.

**What it still won't reach** (same gap as this PR, just now general instead of cargo-only): tools an adopter installs at runtime in a `setup:` step land in the *runner's* home (`/home/runner/.cargo/bin/cargo-nextest`), which is **not** seeded into `/etc/skel`, so it's absent from the sandbox home and the rewrite finds nothing to point at. Those still need the Option 2 lever in #766 (`sandbox_path:` / copying the dir into the sandbox home). So this generalizes the zero-config base-toolchain fix to all languages but doesn't subsume #766.

I'm happy to rewrite this branch to the PATH-derivation above and drop the cargo special-case — it's strictly better than the current diff. Want me to push that here?

max-sixty:
yes great, thx

tend-agent:
Done — rewrote the branch to derive the sandbox PATH from the runner's PATH and dropped the cargo special-case ([682ffe1](https://github.com/max-sixty/tend/commit/682ffe1)).

Since `setup-sandbox.sh` runs as the `runner` user, its `$PATH` already lists every seeded toolchain dir. I checked the live image to be sure the rewrite matches: `/etc/environment`'s PATH is *resolved*, not literal `$HOME` —

```
PATH=/snap/bin:/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:...
```

So the rule is: rewrite a leading `/home/runner` → `${AGENT_HOME}`, keep the entry when it now exists and the sandbox UID can traverse it. cargo, dotnet, composer all rewrite to their `/etc/skel`-seeded sandbox copies; system dirs (`/opt/pipx_bin`, `/usr/local/.ghcup/bin`) carry over as-is. No per-language allowlist.

The security line holds: only runner-home prefixes get rewritten to the sandbox's own copies, and a raw `/home/runner` path is never kept — the bare home is dropped and every kept dir is re-tested with `sudo -u "$SANDBOX" test -x`, so a path the sandbox UID can't enter can't survive.

Still `Refs #766`, not `Closes`: this covers the base toolchains seeded into `/etc/skel`. Subcommands an adopter installs at runtime in a `setup:` step (`cargo-nextest`, …) land in the runner's home, absent from the sandbox home, so the rewrite finds nothing to point at — that's still the general `sandbox_path:` lever.

proxy/setup-sandbox.sh
```diff
@@ -91,9 +91,59 @@ fi
...
+# seeded toolchain across with zero per-language code — cargo, dotnet, ghcup,
+# composer all arrive on the same rule, so there is no per-language allowlist to
+# grow.
+#
+# Security boundary: we only ever rewrite runner-home prefixes to the sandbox's
+# own copies and never keep a /home/runner path on the sandbox PATH — the runner
+# home is where the real credentials live, so the sandbox must not traverse it.
+# A bare ${runner_home} entry is dropped, and every kept dir is re-tested with
+# `sudo -u "$SANDBOX" test -x` so an un-rewritten runner-home path (which the
+# sandbox UID can't enter) can never survive.
```
tend-agent:
```suggestion
# Security boundary: the string rewrite is the sole guard — every /home/runner
# entry is rewritten to its ${AGENT_HOME} copy and the bare ${runner_home} root
# is dropped, so no /home/runner path is ever kept on the sandbox PATH. The `-d`
# + `sudo -u "$SANDBOX" test -x` check is NOT a credential backstop: /home/runner
# is world-traversable (o+x by default, and step 3 chmods it further), so a raw
# runner-home dir would pass test -x; it only drops non-existent/non-exec dirs.
```

tend-agent (commented):
The derivation logic itself is correct and secure for the target environment — I traced it (set -e/-u safe, dedup works, `.local/bin` stays first, empties/nonexistent/dups filtered) and simulated it against a synthetic runner PATH. The string rewrite catches every `/home/runner/*` entry because `$HOME` is exactly `/home/runner` on hosted runners, so no runner-home path is ever kept raw.

One correction on the **security comment**, though: it credits the `sudo -u "$SANDBOX" test -x` check as a backstop against un-rewritten runner-home paths "which the sandbox UID can't enter." The sandbox UID *can* enter `/home/runner` — it's mode `0751` (`drwxr-x--x`, the `o+x`/traversal bit is set by default), and step 3 later chmods it further. So `test -x /home/runner/<dir>` would pass for the sandbox, and a hypothetical un-rewritten runner-home dir would *not* be dropped by that check. The `-d`/`test -x` gate only filters entries that don't exist or aren't executable; the actual credential boundary is entirely the string rewrite (rewrite every prefix, drop the bare root). Worth correcting so a future change doesn't relax the rewrite trusting `test -x` to catch the leak.

Inline suggestion below.

max-sixty:
@tend-agent  resolve conflicts

tend-agent:
Resolved conflicts with main ([5c5fb7c](https://github.com/max-sixty/tend/commit/5c5fb7c)).

The only real conflict was `proxy/setup-sandbox.sh`: main merged the `sandbox_path:`/`TEND_SANDBOX_PATH` lever (the #766 Option-2 lever we discussed) while this branch adds the runner-PATH auto-derivation. They're complementary, so I combined them rather than picking one:

- The adopter's explicit `sandbox_path` dirs are prepended first (after `.local/bin`), so their opt-in tools win and stay trusted/not existence-tested — that dir may be populated by a later `setup:` step.
- The runner-PATH auto-derivation follows, pulling in the `/etc/skel`-seeded toolchains (cargo, dotnet, ghcup, …) with the same home-prefix rewrite + `sudo -u test -x` gate.
- Adopter dirs dedup against the auto-derived rewrites (the runner's `/home/runner/.cargo/bin` collapses into the adopter's `~/.cargo/bin`), so no duplicates. Simulated the combined ordering to confirm.

I also updated the derivation's closing note: the runtime-installed-tool gap is now covered by the merged `sandbox_path:` lever, not "still tracked in #766."

**One thing worth flagging:** `main` is currently red on `test_sandbox_levers_regtest` — #762 updated the mention template's self-review prompt but didn't regenerate that snapshot, so template and recorded output diverged. The merge inherited it; I regenerated the snapshot (one line, matching the current template) so this branch is green. Worth a separate look at why that landed on main red.

max-sixty:
@tend-agent Please make these changes:

1. **Fix the security-boundary comment.** It claims the `sudo -u "$SANDBOX" test -x` re-test ensures an un-rewritten runner-home path "can never survive". It can survive that test: step 3 grants `o+x` on every ancestor of `$GITHUB_WORKSPACE` (which lives under the runner home), so the sandbox UID can traverse `/home/runner`, and `test -x` passes for any world-executable dir beneath it. The `case` rewrite is the sole credential boundary. Say that plainly, and describe `test -x` as filtering nonexistent or non-traversable dirs only, so a future edit doesn't relax the rewrite trusting a backstop that isn't one.

2. **Restore adopter precedence.** The old code put the `sandbox_path:` dirs ahead of `${AGENT_HOME}/.local/bin`; the new order hardcodes `.local/bin` first, so an adopter dir can no longer shadow a same-named binary there, while the comment still says the adopter's tools win. Put the adopter dirs first.

3. Minor: the `[ -n "${runner_home}" ]` guard is dead code, since `${HOME:-/home/runner}` can't be empty (`:-` covers unset and null). Drop it.

Otherwise ready to merge: the derivation logic checks out and CI is green.

tend-agent:
All three done ([fda43f9](https://github.com/max-sixty/tend/commit/fda43f9)):

1. **Security comment rewritten.** It now states plainly that the `case` rewrite is the sole credential boundary and warns against relaxing it on the strength of the `test -x` re-test. You're right the re-test isn't a backstop: step 3 grants `o+x` on every ancestor of `$GITHUB_WORKSPACE` (up to `/`, so `/home/runner` included), so the sandbox UID *can* traverse the runner home and `test -x` passes for any world-executable dir under it. The comment now describes `test -x` as filtering only nonexistent/non-traversable dirs.

2. **Adopter precedence restored.** `_agent_path` now starts empty; the `sandbox_path:` dirs are prepended first, then `.local/bin` (deduped so a `~/.local/bin` opt-in doesn't double up). An adopter dir can again shadow a same-named binary in `.local/bin`, matching the comment.

3. **Dead guard dropped.** Removed the `[ -n "${runner_home}" ]` wrapper around the `case` — `${HOME:-/home/runner}` is never empty.

`shellcheck -S warning` clean; generator suite green (273 passed).

PR #768, opened 2026-07-11
feat(sandbox): adopter levers to reach inside the Claude-family sandbox (closes #766)

### LINKED ISSUE #766: sandbox: no adopter lever to add commands/PATH/env inside the sandbox (setup: only reaches the runner)

tend-agent:
Implements Option 2 from #766 (@max-sixty: "let's do (2)") — the general lever for an adopter to reach **inside** the Claude-family sandbox before the agent launches. Runner-side `setup:` runs as the runner user around the composite action and never reaches the sandbox's fixed env allowlist; these three keys do.

## Config surface (`.config/tend.yaml`, top-level)

```yaml
sandbox_path:            # dirs prepended to the sandbox PATH; leading ~ = sandbox home
  - ~/.cargo/bin
sandbox_env:            # NAME: VALUE added to the agent launch env (reserved keys rejected)
  RUST_BACKTRACE: "1"
sandbox_setup:          # commands run as the sandbox user, workspace cwd, after toolchain+plugins
  - rustup component add clippy
```

The motivating cargo-off-PATH case (#766) becomes `sandbox_path: ["~/.cargo/bin"]`.

## How it flows

- **config** (`config.py`) — parses + validates. `sandbox_env` rejects a reserved set (proxy routing, CA trust, dummy credentials, `PATH`) so an adopter can't redirect the agent off the injecting proxy or clobber the dummy secrets. Warns (inert) under `codex`, whose agent runs on the runner and already sees `setup:`.
- **generator** (`macros.yaml.j2`) — threads the values into the Claude-family agent step as block-scalar action inputs. Not emitted for codex.
- **actions** — `claude` + `claude-interactive` get matching `sandbox_path`/`sandbox_env`/`sandbox_setup` inputs (input-parity hook enforced). `proxy/setup-sandbox.sh` prepends `sandbox_path` (expanding a leading `~`) to the `PATH=` line and appends validated `sandbox_env` lines to the agent env file. A new `shared/steps/sandbox-setup.sh` runs `sandbox_setup` as the sandbox user after toolchain + plugins install, before the run.

## Security

The reserved-env guard lives in **both** config (fails at `init`) and `setup-sandbox.sh` (defends against a hand-edited workflow). Values are baked into the committed workflow at generation time, same trust model as `setup:` — a fork PR editing config can't inject; only a maintainer/nightly regen on the default branch bakes new values. `sandbox_setup` runs at the same non-privileged sandbox UID the agent already runs as, so it's no privilege escalation.

## Scope note

`sandbox_*` is Claude-family only. Codex runs `codex exec` directly on the runner (no proxy sandbox), so its `setup:` already reaches the agent — the keys warn and no-op there.

## Tests

- config-validation tests (reserved-key rejection, invalid env name, non-list types, scalar coercion, codex warning)
- rendering tests (rendered for claude, absent by default, not rendered for codex)
- a rendered-shape regtest locking the block-scalar output
- all 268 generator tests pass; pre-commit (shellcheck, ruff, input-parity, ref-sync) green


tend-agent (commented):
Self-review. The design is sound and well-tested; two small correctness/consistency fixes inline, plus one maintainability note.

**Reserved-key list is duplicated without a parity guard.** `RESERVED_SANDBOX_ENV` in `config.py` and the `case` list in `proxy/setup-sandbox.sh` are kept aligned only by comments. The repo already treats this exact hazard mechanically for the two Claude actions (`claude-action-input-parity` in `.pre-commit-config.yaml`), because comment-based sync rots. Here the drift is security-relevant: add a var to the `AGENT_ENV_FILE` heredoc and forget the `case` list, and an adopter's `sandbox_env` can override that var past the boundary the re-check exists to defend. Worth either a small parity check, or deriving the shell guard from the names already written to `AGENT_ENV_FILE` (reject any `sandbox_env` name that already appears in the file) so there's a single source of truth. Not blocking — flagging for a follow-up.

generator/src/tend/config.py
```diff
@@ -271,6 +315,62 @@ def load(cls, path: Path | None = None) -> Config:
...
+                    if name == "PATH"
+                    else " It carries the sandbox's credential isolation and "
+                    "cannot be overridden."
+                )
+                raise click.ClickException(
+                    f"sandbox_env may not set reserved key '{name}'.{hint}"
+                )
+            # Coerce scalars (a YAML `1` or `true`) to their string form so the
+            # value renders as a plain NAME=VALUE line in the agent env file.
+            sandbox_env[name] = value if isinstance(value, str) else str(value)
```
tend-agent: The comment says "coerce scalars," but `str(value)` accepts anything — a list or dict value (`sandbox_env: {FOO: [a, b]}`) silently becomes the Python repr `FOO=['a', 'b']` in the agent env instead of failing at `init`. That's the "silently parse" case CLAUDE.md rules out ("old formats should fail with a clear error, not silently parse"). Reject non-scalars:

```suggestion
            # Coerce a YAML scalar (1, true) to its string form; reject a
            # non-scalar (a list/dict would otherwise str() into a Python repr).
            if isinstance(value, str):
                sandbox_env[name] = value
            elif isinstance(value, (int, float)):
                sandbox_env[name] = str(value)
            else:
                raise click.ClickException(
                    f"sandbox_env value for '{name}' must be a scalar "
                    "(string, number, or boolean)"
                )
```

shared/steps/sandbox-setup.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# Env-only tweaks (PATH, exported vars) do NOT persist to the agent from here —
+# a child shell's exports die with it. Use `sandbox_path:` / `sandbox_env:` for
+# those; use `sandbox_setup:` for actions with on-disk effects (installing a
+# tool, warming a cache, generating a file).
+#
+# Inputs (env): TEND_SANDBOX_SETUP (the commands; empty → no-op), SANDBOX and
+# AGENT_ENV_FILE (exported by setup-sandbox.sh via $GITHUB_ENV). Shared by the
+# two Claude harness actions.
+set -eo pipefail
```
tend-agent: `SANDBOX` and `AGENT_ENV_FILE` are used unguarded here but only documented, not defaulted — `-u` would fail loudly if either export were ever missing (rather than running `mapfile < ""` / `sudo -u ""`), and it matches the sibling `setup-sandbox.sh`, which uses `set -euo pipefail`.

```suggestion
set -euo pipefail
```

max-sixty:
@tend-agent please fix based on review?

tend-agent:
Pushed `3ac8f0d` addressing all three review points:

- **Non-scalar `sandbox_env` values** — now rejected at `init` with a clear "must be a scalar" error instead of `str()`-ing a list/dict into a Python repr in the agent env. Added a test.
- **`sandbox-setup.sh`** — `set -euo pipefail` (was `-eo`), so a missing `SANDBOX`/`AGENT_ENV_FILE` export fails loudly, matching `setup-sandbox.sh`.
- **Reserved-key drift** — added a `sandbox-env-reserved-parity` pre-commit hook that fails the commit if `RESERVED_SANDBOX_ENV` (config.py) and the `case` guard (setup-sandbox.sh) diverge, replacing comment-only sync for the security-relevant list. Went with the mechanical parity check rather than deriving from `AGENT_ENV_FILE`, since only one of the two auth-scheme vars is present in the file at runtime — deriving would stop guarding the inactive one.

269 generator tests, ruff, shellcheck, and the new parity hook all green.

generator/src/tend/config.py
```diff
@@ -271,6 +316,71 @@ def load(cls, path: Path | None = None) -> Config:
...
+            # and silently smuggle garbage into the agent env line).
+            if isinstance(value, str):
+                sandbox_env[name] = value
+            elif isinstance(value, (int, float)):
+                sandbox_env[name] = str(value)
+            else:
+                raise click.ClickException(
+                    f"sandbox_env value for '{name}' must be a scalar "
+                    "(string, number, or boolean)"
+                )
```
tend-agent:
```suggestion
            if isinstance(value, bool):
                coerced = "true" if value else "false"
            elif isinstance(value, str):
                coerced = value
            elif isinstance(value, (int, float)):
                coerced = str(value)
            else:
                raise click.ClickException(
                    f"sandbox_env value for '{name}' must be a scalar "
                    "(string, number, or boolean)"
                )
            if "\n" in coerced:
                raise click.ClickException(
                    f"sandbox_env value for '{name}' must be a single line"
                )
            sandbox_env[name] = coerced
```

`bool` is an `int` subclass, so handle it before the `int/float` branch and emit lowercase; the trailing check keeps a newline-bearing value from silently producing a broken block scalar in the rendered workflow.

shared/steps/sandbox-setup.sh
```diff
@@ -0,0 +1,30 @@
...
+
+[ -n "${TEND_SANDBOX_SETUP:-}" ] || exit 0
+
+# Run as the sandbox user with the agent's launch env. The script text is piped
+# on stdin (read by `bash -s`) rather than via a temp file, so no sandbox-side
+# read permission on a runner-owned path is needed. `set -e` inside so a failing
+# setup command fails the step loudly rather than silently proceeding to the run.
+mapfile -t AGENT_ENV <"$AGENT_ENV_FILE"
+printf '%s\n' "$TEND_SANDBOX_SETUP" \
+  | sudo -u "$SANDBOX" env "${AGENT_ENV[@]}" bash -eo pipefail -s
```
tend-agent:
```suggestion
# Run as the sandbox user with the agent's launch env. The commands go through
# `bash -c`'s argument: no temp file (so no sandbox-side read permission on a
# runner-owned path), and not stdin (so a setup command that reads stdin can't
# swallow the remaining lines). `-e` inside so a failing setup command fails the
# step loudly rather than silently proceeding to the run.
mapfile -t AGENT_ENV <"$AGENT_ENV_FILE"
sudo -u "$SANDBOX" env "${AGENT_ENV[@]}" bash -eo pipefail -c "$TEND_SANDBOX_SETUP"
```

tend-agent (commented):
Re-reviewed after `3ac8f0d` — the three prior points (non-scalar rejection, `set -euo pipefail`, the reserved-key parity hook) are all addressed and CI is green. A fresh pass over the incremental turned up a few refinements, none blocking:

- **`sandbox_env` boolean values render Python-cased.** `bool` is an `int` subclass, so `DEBUG: true` falls into the `int/float` branch and `str(True)` writes `DEBUG=True` into the agent env, not the shell-conventional `true`/`1`. The coercion comment names `true` as an intended input and the error message lists `boolean` as accepted, so this is a real inconsistency. Inline suggestion below (it also rejects an embedded newline — see next point).
- **A newline in a `sandbox_env` value (or `sandbox_path` dir) produces a broken workflow.** Both render at column 12 with no `indent()` filter (unlike `sandbox_setup`), so a value containing a newline drops an un-indented continuation line that terminates the `|` block scalar. The generator writes it unvalidated, so it fails only when GitHub Actions later refuses to load the workflow — the opposite of the repo's fail-at-`init` rule. The `sandbox_env` fix is folded into the inline suggestion; `sandbox_path` needs the same one-line guard in its own validator.
- **`sandbox_setup` commands are fed on stdin.** Piping the script to `bash -s` makes the script text the commands' own stdin, so a setup command that reads stdin (an installer prompt, `read`, etc.) silently swallows the remaining commands and the step still exits 0. Inline suggestion below switches to `bash -c`, which keeps stdin free and still needs no temp file.
- **The shell reserved re-check accepts a line with no `=`.** A hand-edited `sandbox_env:` line like `FOO` (no `=`) yields `name=FOO`, which isn't reserved, so it's appended verbatim; `env` then treats the assignment-less token as the *command* to exec and the agent never launches. Since the re-check exists specifically to defend hand-edited workflows, rejecting a non-`NAME=VALUE` line before the `case` closes that path.
- **The codex inert-warning keys on the global harness.** With `harness: codex` globally plus a per-workflow `harness: claude` override, the levers *are* active on that workflow (the macro renders per effective harness) yet `load()` still prints "inert under codex" — and the inverse (global claude, per-workflow codex) warns not at all. Minor: gating on whether any effective harness is Claude-family would match the render logic.
- **The parity hook guards config↔shell but not the heredoc.** The invariant that matters is "every key written to `AGENT_ENV_FILE` is reserved"; the hook only compares `RESERVED_SANDBOX_ENV` against the shell `case`. All 20 heredoc keys are covered today, so no live hole — but a future credential/routing var added to the heredoc without updating the reserved set would pass the hook. Optional hardening: also assert the heredoc keys are a subset of the reserved set.

generator/src/tend/config.py
tend-agent: Applied in `f38abd3` — `bool` is now handled before the `int/float` branch (`DEBUG: true` → `DEBUG=true`), and both `sandbox_env` values and `sandbox_path` dirs are rejected at `init` if they contain a newline. Added `test_sandbox_env_coerces_boolean_lowercase`, `test_sandbox_env_newline_value_rejected`, and `test_sandbox_path_newline_rejected`.

shared/steps/sandbox-setup.sh
tend-agent: Applied in `f38abd3` — switched to `bash -eo pipefail -c "$TEND_SANDBOX_SETUP"`, so the commands no longer own stdin and a setup command that reads stdin can't swallow the rest.

tend-agent:
Pushed `f38abd3` addressing all six points from the re-review:

- **Boolean values** — `bool` is handled before the `int/float` branch, emitting shell-conventional `true`/`false`.
- **Newline in `sandbox_env` / `sandbox_path`** — both reject a newline at `init` (`must be a single line`), so a value can't silently terminate the rendered block scalar.
- **`sandbox_setup` on stdin** — now `bash -c "$TEND_SANDBOX_SETUP"` instead of piping on stdin, so a command that reads stdin can't swallow the remaining lines.
- **Non-`NAME=VALUE` line in the shell re-check** — `setup-sandbox.sh` rejects a line with no `=` before the reserved `case`, closing the exec-the-token path for a hand-edited workflow.
- **Codex inert-warning** — moved after the workflow parse and gated on the *effective* harness set, so a top-level `codex` with a per-workflow `claude` override (where the levers do apply) no longer warns.
- **Parity hook** — also asserts every key written to `AGENT_ENV_FILE` (heredoc + both `ANTHROPIC_DUMMY` scheme keys) is in the reserved set, so a future routing/cred var added to the env file without reserving it fails the commit.

273 generator tests (4 new), ruff, shellcheck, and the parity hook all green locally.

PR #784, opened 2026-07-20
fix(list-recent-runs): fail loud on transient gh errors instead of a false all-clear

tend-agent:
## Problem

`list-recent-runs.sh` silently returns an empty (or partial) run list when a `gh` API call fails transiently, causing `review-reviewers` / `review-runs` to record a false **all-clear** and permanently skip that window's runs.

Directly reproduced this run (review-reviewers leg [`29710673739`](https://github.com/max-sixty/tend/actions/runs/29710673739)) during the active GitHub Actions incident (critical, started 2026-07-19T23:34Z, affecting API Requests / Issues / Actions / Pages). The workflow-discovery fetch hit an HTTP 503:

```
+ mapfile -t matches < <(gh workflow list -R max-sixty/cargo-affected --json name --jq ...)
could not get workflows: HTTP 503: No server is currently available to service your request.
+ WORKFLOWS+=("${matches[@]}")   # empty
...
+ echo '[]'                      # <- reported zero runs
```

At that moment three `tend-notifications` runs (`29708109161`, `29708916955`, `29710166166`) existed in the window. Had the leg trusted the script's output, it would have recorded an all-clear and never analyzed them.

## Root cause

Two call sites swallow `gh` failures, both of which escape `set -euo pipefail`:

- `mapfile -t matches < <(gh workflow list ...)` — a process substitution's exit status is invisible to `set -e`, and `mapfile` itself succeeds on empty input. A single 503 here zeros out `WORKFLOWS` → no runs are queried → `[]`.
- `runs=$(gh run list ... || echo "[]")` — the `|| echo "[]"` fallback turns a per-workflow fetch error into an empty result, silently dropping that workflow's runs.

Either way the script emits fewer runs than exist with a `0` exit code, and the calling skill treats it as "nothing happened this hour." The failure is invisible: a false all-clear produces no CI failure, and each leg only ever sees its own one-hour window, so a skipped hour is never re-analyzed.

## Fix

A small `gh_retry` helper retries transient failures (3 attempts, linear backoff). Both fetch sites now use it and **fail loud** (`exit 1`) if every attempt fails, so the caller errors instead of reporting a bogus empty window. A genuinely empty result (successful fetch, repo has no matching workflows) still returns `[]` as before — only *fetch failures* now abort.

## Verification

- `bash -n` + `shellcheck`: clean.
- Failure path (mocked `gh` that always 503s): exits `1` with `refusing to report an empty run list`, instead of printing `[]`.
- Success path (live): returns the two in-window runs (`29710166166`, `29708916955`) rather than `[]`.

## Gate assessment

- **Evidence**: directly reproduced this run — clearly-wrong outcome (analysis tooling reports zero runs while runs exist → missed detection that never resurfaces).
- **Structural**: yes — any transient fetch failure deterministically produces the empty/partial result; not a stochastic model lapse.
- **Invisible failure mode**: yes — a false all-clear surfaces no CI failure and the skipped window is never re-analyzed, so it wouldn't be caught next time without this fix.
- **Change type**: targeted fix (retry helper + fail-loud at two call sites), proportionate.

Evidence log: https://gist.github.com/8518c95fe85c0e52a3b5f26695187382

### DIFF
--- a/plugins/tend-ci-runner/scripts/list-recent-runs.sh
+++ b/plugins/tend-ci-runner/scripts/list-recent-runs.sh
@@ -44,7 +44,10 @@ gh_retry() {
       printf '%s' "$out"
       return 0
     fi
-    sleep $((attempt * 3))
+    # Don't sleep after the final attempt — the whole point of this script is
+    # to fail loud and fast during an incident, so a trailing backoff before
+    # the caller's `exit 1` is wasted delay.
+    [ "$attempt" -lt 3 ] && sleep $((attempt * 3))
   done
   return 1
 }
--- a/plugins/tend-ci-runner/scripts/list-recent-runs.sh
+++ b/plugins/tend-ci-runner/scripts/list-recent-runs.sh
@@ -29,6 +29,26 @@ set -euo pipefail
 # Prevent gh from emitting ANSI color codes in non-TTY contexts.
 export NO_COLOR=1
 
+# Retry a command up to 3 times on failure, echoing its stdout on success.
+# Transient GitHub API errors (e.g. HTTP 503 during an Actions incident) would
+# otherwise slip past `set -euo pipefail` at the call sites below — a process
+# substitution's failure is invisible to `set -e`, and a `|| echo "[]"` fallback
+# silently turns an errored fetch into an empty result. Either way the script
+# would report zero runs when runs exist, and the calling skill records a false
+# "all-clear" that permanently skips that window. Retry here, and let the caller
+# fail loud if every attempt fails rather than swallow the error.
+gh_retry() {
+  local out attempt
+  for attempt in 1 2 3; do
+    if out=$("$@" 2>/dev/null); then
+      printf '%s' "$out"
+      return 0
+    fi
+    sleep $((attempt * 3))
+  done
+  return 1
+}
+
 repo_args=()
 if [ -n "${TARGET_REPO:-}" ]; then
   repo_args=(-R "$TARGET_REPO")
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/scripts/list-recent-runs.sh
```diff
@@ -29,6 +29,26 @@ set -euo pipefail
...
+# "all-clear" that permanently skips that window. Retry here, and let the caller
+# fail loud if every attempt fails rather than swallow the error.
+gh_retry() {
+  local out attempt
+  for attempt in 1 2 3; do
+    if out=$("$@" 2>/dev/null); then
+      printf '%s' "$out"
+      return 0
+    fi
+    sleep $((attempt * 3))
```
tend-agent:
```suggestion
    [ "$attempt" -lt 3 ] && sleep $((attempt * 3))
```
Avoids the pointless ~9s sleep after the last attempt before returning failure. Safe under `set -e`: a failing test that is *not* the command following the final `&&` is exempt, so when `attempt == 3` the loop just skips the sleep and returns 1.

tend-agent (commented):
The fail-loud approach is right, and the two swallowing sites (the `mapfile` process-substitution and the `|| echo "[]"` fallback) are exactly the leaks that produced the false all-clear. `bash -n` and `shellcheck` are clean and a genuinely empty successful fetch still returns `[]`.

One small thing: `gh_retry` sleeps even after the final (3rd) attempt fails, delaying the `exit 1` by ~9s. Since this PR's whole point is to fail *loud and fast* during an incident, skipping the trailing sleep tightens exactly that path.

tend-agent:
Good catch — fixed in a519ad105b3e0d4da92c509e5156ef4111ef50ae. `gh_retry` now guards the backoff with `[ "$attempt" -lt 3 ]`, so a total failure returns immediately after the third attempt instead of sleeping ~9s first. `bash -n` + `shellcheck` clean; the mocked all-503 path now exits `1` in ~9s (the two inter-attempt backoffs) rather than ~18s.

PR #804, opened 2026-08-02
remove the claude-interactive harness

max-sixty:
Deletes the `claude-interactive` harness, leaving `claude` and `codex`. −1152 lines, +86.

## Why

It was built for one reason: Anthropic's 2026-06-15 change would have metered `claude -p` against Agent SDK credits while leaving interactive terminal sessions on the subscription, and running the binary under a PTY sidestepped that. That reason is gone twice over. Anthropic paused the change and still lists it paused ("Claude Agent SDK, `claude -p`, and third-party app usage still draw from your subscription's usage limits"), and the default `claude` harness was since rewritten to run the same binary headless rather than going through the SDK at all — so the thing the PTY was avoiding is no longer on either path.

Nothing selected it. No adopter config sets `harness:`, tend's own dogfooding moved off it at the rollback, and the last caller anywhere was `review-reviewers.yaml`, which had it by accident and was fixed in max-sixty/tend#803.

What it cost to keep: a 667-line parallel composite action, a second smoke workflow, a three-way branch in the generator, a pre-commit hook whose entire job was policing the duplication, and prose across nine files — which had already drifted, since the action's own description still advertised itself as the alternative to `claude-code-action`.

## Restoring it

`CLAUDE.md` records the removal with a pointer to `036f9c4`, the last commit containing the harness, so it can be recovered if the metering resumes.

## Shape of the diff

**Deleted** — `claude-interactive/action.yaml`, `.github/workflows/interactive-smoke.yaml`, the `claude-action-input-parity` pre-commit hook (it existed only to keep the two actions' inputs aligned), four `test_claude_interactive_*` tests, `test_claude_interactive_runs_claude_auth_check`, `test_per_workflow_harness_same_family_no_model_clash`, and the now-orphaned regtest snapshot.

**Collapsed** — `CLAUDE_FAMILY_HARNESSES` was a two-element set consulted at six sites across `config.py`, `checks.py`, and `workflows.py`. Rather than leave a set of one, the sites now test `harness == "claude"` directly. Same for the template's `{% if cfg.harness in ('claude', 'claude-interactive') %}` guards and the "Claude-family" phrasing in adopter-facing comments.

**Reparameterized** — `test_per_workflow_harness_override_targets_only_named_workflow` used `claude-interactive` as its override target. It now overrides to `codex`, which exercises the feature harder: the override has to carry its own secret shape, so the test additionally asserts `openai_api_key` appears and `claude_code_oauth_token` does not.

**Also here** — a nightly task in tend's `running-tend` overlay for restamping action refs in hand-maintained workflows. `init` only rewrites the generated `tend-*.yaml` files, which is precisely how `review-reviewers.yaml` sat on the wrong harness for three releases without anyone noticing.

## Testing

284 pass (291 before; the 7-test drop is the deleted tests plus one parametrize case). Beyond the suite, `tend init --dry-run` was run for real on both remaining harnesses: `claude` stamps `max-sixty/tend/claude@0.1.12` across all eight workflows, `codex` stamps `codex@0.1.12` with `openai_api_key`.

The gap: no CI run has executed against a workflow generated by this code. The regtests assert the YAML text, not that GitHub accepts it — though the action refs and input names are unchanged for both surviving harnesses, so there's nothing new for GitHub to reject.

## If this is wrong

Paused is not cancelled, and Anthropic said it would give notice before shipping a revised version. If the metering returns, this has to be rebuilt or reverted rather than toggled. The bet is that notice plus a commit reference beats carrying a 667-line duplicate that had already started to rot.

No CHANGELOG entry: that section is drafted at release time from the commit range.


tend-agent (commented):
The generator collapse (`CLAUDE_FAMILY_HARNESSES` → `harness == "claude"`) is mechanically equivalent now that `KNOWN_HARNESSES == {"claude", "codex"}`, the test reparameterization exercises the override harder, and `uv run pytest` gives 284 passing here. One gap: the harness-count prose cleanup is incomplete, and in two spots the PR left the comment internally inconsistent.

The two files the PR touched now disagree with themselves — the parenthetical was updated but the count word wasn't:

- `shared/steps/security-preflight.sh` — `Shared verbatim by all three harness actions (claude/, codex/).` — says "three", lists two.
- `shared/steps/validate-claude-auth.sh` — `Shared by the two Claude harness action (claude/).` — "two ... action", now a single Claude action.

And the same drift survives untouched in nine more places:

- "all three harness actions" (now two — claude, codex): `shared/steps/ensure-uv.sh`, `mark-notification-read.sh`, `rate-limit-preflight.sh`, `report-failure.sh`, and `claude/action.yaml` (`so the three harness actions can't drift`).
- "the two Claude harness actions" (now one): `shared/steps/compose-system-prompt.sh`, `install-agent-toolchain.sh`, `install-tend-plugins.sh`, `restore-sensitive-config.sh`, `sandbox-setup.sh`.

Separately, `claude/action.yaml` still carries `The two Claude actions are swap-in compatible — keep their input surfaces in sync (enforced by the claude-action-input-parity pre-commit hook)` — but this PR deletes both the second Claude action and that hook, so the comment now points at two things that no longer exist. Given the PR's premise is retiring prose that had already drifted, this reads as exactly the kind of line worth catching.

These are all outside the diff hunks, so I've left them in the body rather than as inline suggestions. The fix is mechanical (three→two; "the two Claude harness actions"→"the Claude harness action"; drop the input-parity sentence) — happy to push a single sweep commit if you'd like, just say the word.

tend-agent (approved):

PR #806, opened 2026-08-02
fix(review): run adopter setup on the base tree, not the PR's

max-sixty:
`tend-review` checked out the fork's tree and then ran the adopter's `setup:` steps against it. Those steps execute as the runner user, which holds sudo and — until `proxy/setup-sandbox.sh` strips it — the checkout PAT in `.git/config`. So a contributor's build backend, added dependencies, or local `uses: ./` action ran with that access, ahead of the containment the harness exists to build. Root on the runner before the sandbox is set up makes everything downstream of it decorative.

Review now checks out the base tree, runs `setup:`, then lands the PR's tree. That is the ordering every other generated workflow already had — `mention` runs `setup:` before its `gh pr checkout`, and the rest check out the default branch — so this removes an exception rather than adding a rule.

The base checkout renders only alongside `setup:`, since giving setup a reviewed tree is the whole reason it exists. Where it does render, the PR checkout carries `clean: false`. Verified against `actions/checkout`'s `git-directory-helper.ts` at `v7`: with `clean` false, `prepareExistingDirectory` skips `git clean -ffdx && git reset --hard`, keeps the existing repo since the fetch URL matches, and only detaches HEAD and prunes local branches. Setup's untracked output survives; tracked files are overwritten by the PR tree, which is what a reviewer should see. The merge-ref tree is preserved, so this doesn't reopen the tradeoff settled in #723.

Adopters with no `setup:` generate byte-identical output to before this PR — `clean` is omitted along with the base checkout, since the action consults that input only on the existing-repo path, which a lone checkout never reaches.

## Scope

This strengthens defense in depth, not a load-bearing boundary. The merge restriction still stops the worst outcome, unchanged either way. Under the codex harness the effect is close to cosmetic: that harness runs the agent on the runner with `sandbox: danger-full-access` and passes both credentials directly, so the agent executes the same code with the same access moments later.

## Adopter impact

Setup no longer sees the PR's manifests. Project setup that must (installing the PR's dependencies) belongs in `sandbox_setup:`, which runs as the sandbox user with the workspace as cwd — documented in `docs/tend.example.yaml`. Adopters whose setup only installs toolchains, or who have none, need no change.

## Verification

285 tests pass, `pre-commit run --all-files` clean, and `actionlint` passes on generated workflows both with and without setup steps.

### DIFF
--- a/docs/security-model.md
+++ b/docs/security-model.md
@@ -89,6 +89,17 @@ in shell; the path list and ordering mirror claude-code-action's
 overwritten, so review skills can optionally inspect what the PR changed without
 those files ever being executed.
 
+**Setup runs on reviewed code.** Adopter `setup:` steps execute as the runner
+user, which holds sudo and, until the sandbox setup strips it, the checkout
+PAT in `.git/config`. So every generated workflow checks out reviewed code
+before running them — the default branch, or in `tend-review` the PR's base
+branch — and lands the PR's tree only afterwards. A contributor's build
+backend, added dependencies, and local `uses: ./` actions therefore execute
+under the agent, inside the sandbox, rather than ahead of it. `sandbox_setup:`
+is the lever for project setup that must see the PR's own manifests. Codex
+adopters get the ordering without the containment, since that harness runs the
+agent on the runner.
+
 **Credential isolation (Claude harnesses).** The Claude harness actions run the
 agent as a separate non-sudo `tend-sandbox` user, sharing the proxy machinery
 under the top-level `proxy/`. Both the bot PAT and the Anthropic credential (OAuth token
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -106,6 +106,12 @@ bot_name: my-project-bot
 # composite action (`.github/actions/tend-setup/action.yaml`) and reference
 # it from one entry with `uses`.
 #
+# Setup runs against reviewed code — the default branch, or in `tend-review`
+# the PR's base branch, checked out before the PR's own tree lands. Project
+# setup that must see the PR's manifests (installing its dependencies) goes in
+# `sandbox_setup:` below, which runs inside the Claude harness's sandbox. Under
+# codex there is no such lever; the agent installs what it needs itself.
+#
 # `tend-notifications` injects an `if:` guard on every step so setup is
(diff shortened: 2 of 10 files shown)


tend-agent (commented):
The ordering change and `clean: false` reasoning both check out — traced the second checkout against `actions/checkout` v7: with `clean` false it reuses the existing repo, `--force`-reconciles tracked files to the merge tree, and leaves setup's untracked output in place. Tests and regtests are consistent (284 pass).

One design observation, non-blocking: for adopters with an empty `setup:`, the review job now runs two full `actions/checkout@v7` steps (`fetch-depth: 0`) with nothing between them — visible in the regenerated `test_workflow_minimal_regtest[review]` output. The base checkout exists only to give `setup:` a reviewed tree; with no setup it's a full-history clone of the base branch that the second checkout immediately supersedes, so it's pure overhead on every review run for that subset. The other generated workflows have this too, but they run a single checkout — review is the one place the base checkout is genuinely redundant when there's no setup.

Guarding the base checkout + `<<setup>>` behind `{% if setup %}` in `review.yaml.j2` would keep the security ordering when setup is present and drop the wasted clone when it isn't (the second `clean: false` checkout clones fine into an empty workspace — `clean` only affects the existing-repo path). The tradeoff is template uniformity vs. avoiding the redundant fetch. Happy to push that if you'd prefer it; otherwise this is fine as-is.

max-sixty:
Took the suggestion — pushed in a9aefae.

The base checkout is `setup:`'s, so it now renders only alongside it, guarded by `{% if has_setup %}` in `review.yaml.j2`. `clean` went with it rather than staying pinned to `false`: as you noted, `git-source-provider.ts` only calls `prepareExistingDirectory` when it finds an existing repo, so the input is inert on a lone checkout and emitting it would just be a wart for a future reader.

The result is stronger than avoiding the extra fetch — all three no-setup regtest snapshots (`test_workflow_minimal_regtest[review]`, `test_workflow_minimal_codex_regtest[review]`, `test_extras_apply_path_regtest`) are now byte-identical to `main`. Adopters without `setup:` see no diff at all from this PR, so the change is scoped to exactly the repos where the ordering matters.

Added `test_review_without_setup_checks_out_once` for that shape; 285 pass, and actionlint is clean on generated workflows both with and without setup.

tend-agent (approved):

PR #807, opened 2026-08-03
fix: isolate tend's uv from the adopter's

max-sixty:
tend installed an unpinned uv into `$AGENT_HOME/.local/bin`, which `setup-sandbox.sh` places ahead of every PATH entry derived from the runner. `astral-sh/setup-uv` resolves a version from the repo's `uv.toml`/`pyproject.toml` and installs to `/opt/hostedtoolcache`, which survives the runner-home rewrite as a distinct entry — ranked below `.local/bin`. So tend's copy silently shadowed the version the adopter had deliberately chosen, in order to serve one skill.

There are really two uvs with different owners, and they'd been conflated. The runner's runs mitmproxy; the sandbox's exists only so the agent can run `uvx tend@latest`. This splits them.

**Proxy.** `shared/steps/install-proxy-uv.sh` installs a pinned uv (new `uv_version` input) into `$RUNNER_TEMP/tend-uv` and never touches `$GITHUB_PATH`; `setup-sandbox.sh` addresses it absolutely. It replaces `ensure-uv.sh`, which installed unpinned-latest and only when `uvx` was absent — so the process holding the PAT and the model credential started from whatever the adopter's `setup:` happened to leave on PATH. Observed in practice: 0.12.0 on one adopter, 0.12.1 on another, unpinned latest where there was no setup at all. `claude_version` and `mitmproxy_version` were already pinned; this brings the third in line.

**Sandbox.** `install-agent-toolchain.sh` becomes `install-claude-binary.sh`, installing only that — the step is renamed to match. Skills needing `uvx tend@...` go through `plugins/tend-ci-runner/scripts/tend-uvx.sh`, which installs into `$HOME/.tend-uv` on first use and `exec`s through, so arguments pass unchanged. It's idempotent, since shell state doesn't survive between the agent's bash calls.

**Codex.** Drops its uv step entirely — that harness has no proxy, and nothing else consumed it. `agents-tail.md`'s "`uvx` is on PATH" claim goes too; it would have become false.

<details>
<summary>Why a shared wrapper rather than an inline install</summary>

The first pass had the `nightly` skill install uv inline, on the basis that it was the only consumer of `uvx tend@latest` across the nine bundled skills. Review found a second one: `.claude/skills/running-tend/references/integration-test.md` calls it three times, and `running-tend/SKILL.md` runs that recipe from the weekly workflow, inside the sandbox. Two consumers means one home. `CLAUDE_PLUGIN_ROOT` was already the established path for overlay skills (`running-tend/SKILL.md` uses it for `token-report.sh`), and the codex action exports it explicitly, so the wrapper resolves under both harnesses.

</details>

Verified by running rather than by reading: both installers execute end to end and produce `uv`/`uvx` at the expected paths, the wrapper skips the install on repeat calls, and `UV_NO_MODIFY_PATH=1` leaves no shell profile behind. The generator is untouched, so no generated workflow changes.

One thing to know before this lands: the `codex-credential-proxy` branch will conflict, since it needs a proxy uv and `ensure-uv.sh` is gone. It should adopt `install-proxy-uv.sh` — which is why that script stayed under `shared/steps/` rather than moving to `claude/`.


tend-agent (approved):
Clean split. The proxy uv is pinned, installed off-PATH into a tend-owned dir, and addressed absolutely by the runner-owned process that launches `mitmdump`, so the credential-holding process no longer starts from whatever the adopter's `setup:` left on PATH; the agent's uv goes through the off-PATH wrapper; and codex correctly drops its now-unused step. I traced the proxy launch (same runner UID installs and invokes `$TEND_UV_DIR/uvx`), confirmed `CLAUDE_PLUGIN_ROOT` is exported under both harnesses so `tend-uvx.sh` resolves, and checked that every sandbox-context `uvx tend@...` call is converted.

One stale comment, outside the diff so I couldn't attach an inline suggestion: in `claude/action.yaml`, the step named `Install agent toolchain (sandbox)` still opens with `# Install the agent toolchain (uv + the claude binary) into the sandbox user's home.` — but `install-agent-toolchain.sh` is now claude-binary-only (its own header was updated to say so). Since the whole point here is making uv ownership legible, that line now contradicts the change. Happy to push a one-line fix dropping the `uv + ` if you'd like.

max-sixty:
Good catch — folded in as 376a7f6. Renamed both the step (`Install claude binary (sandbox)`) and the script (`install-claude-binary.sh`), since the name had drifted the same way the comment had.

tend-agent:
Confirmed — the rename is clean: no `install-agent-toolchain` references remain anywhere in the tree, and the step comment now matches (`claude binary… Nothing else`). Approval stands.

Issue #808, opened 2026-08-03
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T00:57:20Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30775588422) | N/A |

This issue was created automatically. Close it once the outage is resolved.

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T01:18:35Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30776520686) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T01:18:27Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30776520686) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T01:18:53Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30776520686) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T01:18:30Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30776520686) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T01:19:17Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30776520686) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T01:32:53Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30777070171) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T02:38:11Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30779722563) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-03T02:43:53Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30779959756) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
[...]

PR #809, opened 2026-08-03
fix(report-failure): dedup outage comments per run across matrix legs

tend-agent:
## Problem

When a **matrix** workflow fails during a bot outage, the `tend-outage` tracker issue is flooded with one near-identical comment **per matrix leg**, all citing the same run.

`review-reviewers` has a 5-leg matrix. During an outage this window (`claude -p` exiting 1 uniformly, self-resolved by 03:57Z), two failed `review-reviewers` runs each posted ~5–6 comments to [#808](https://github.com/max-sixty/tend/issues/808):

- run `30776520686` → 5 comments, all linking `.../runs/30776520686`
- run `30779959756` → 6 comments, all linking `.../runs/30779959756`

Each comment is a one-row table differing only in a jitter-spread timestamp — the same run recorded 5–6 times.

## Root cause

[`shared/steps/report-failure.sh`](https://github.com/max-sixty/tend/blob/520378fe4803/shared/steps/report-failure.sh) is invoked once per matrix leg, every leg sharing one `GITHUB_RUN_ID`. The script already handles the concurrent-leg race on the **create** path — jittered backoff (#586) plus a self-heal reconcile that closes duplicate *issues* (#744) — but the **append** path had no dedup: every leg unconditionally `gh issue comment`s its own row.

So the create-create race (duplicate issues) was solved; the comment-append race (duplicate comments) was not. This is the same class of concurrent-matrix-leg noise the maintainer has fixed repeatedly (#586, #744, and #560 which batches enrichment into one comment per issue) — this closes the remaining gap.

## Fix

Symmetric to the existing issue reconcile:

1. **Guard** — before appending, skip if this run's URL is already recorded on the issue (body *or* an existing comment). Collapses most of the flood, since jitter (0–30s) already spreads the legs.
2. **Reconcile** — the check-then-act still races across jittered legs, so after posting, keep the earliest comment citing this run and delete later duplicates. Convergent: every racing leg sorts identically and computes the same keeper, so deleting an already-deleted comment is a harmless 404.

Net effect: one row per run, regardless of matrix width. Non-matrix workflows (single leg) are unaffected — the guard finds nothing, posts once, reconcile is a no-op.

Comments deleted are the bot's own auto-generated outage rows; no human content is touched.

## Gate assessment

- **Class**: structural — a shell script with no per-run comment dedup. Replayed 10/10 times, every matrix-workflow failure floods. Not a stochastic model lapse; confirmed by reading the script source and the 11 duplicate comments on #808.
- **Evidence**: High (structural). 2 distinct failed `review-reviewers` matrix runs this window, both exhibiting the identical flood (~11 duplicate-run comments total), plus the defect is guaranteed to recur on any future matrix-workflow outage.
- **Magnitude**: targeted fix to one shared script (add a guard + a reconcile mirroring the adjacent issue reconcile) — Normal bar, met.
- **Passes both gates.** The transient outage that triggered the failures is *not* itself a code fix (it self-resolved by 03:57Z, tracked by #808); this fixes the durable noise the outage exposed.

<details><summary>Window & evidence</summary>

- Analysis run: [30783291243](https://github.com/max-sixty/tend/actions/runs/30783291243), window ~00:12Z–04:02Z 2026-08-03 (extended back to the last *successful* review-reviewers analysis, since the 01:17Z and 02:43Z analysis runs both failed in the same outage — leaving a coverage gap the normal window floor didn't reach).
- Evidence log: https://gist.github.com/e08f6e62d6478163cb425a75648eb7e4
- Other in-window bot output was clean and accepted: PR #807 got a bot APPROVE and merged at 00:41Z; corruption scan clean.
</details>


shared/steps/report-failure.sh
```diff
@@ -49,8 +49,34 @@ sleep $((RANDOM % 30))
...
 if [ -n "$EXISTING" ]; then
+  # Per-run comment dedup. A matrix workflow invokes this script once per leg,
+  # every leg sharing one GITHUB_RUN_ID (and thus one RUN_URL). Without a guard
+  # each leg appends its own near-identical row and floods the issue (a 5-leg
+  # matrix failing during an outage → 5 comments all citing the same run).
+  # Skip if this run is already recorded — whether in the issue body (a leg of
+  # this same run seeded the issue) or in an existing comment.
+  if gh issue view "$EXISTING" --json body,comments \
+      --jq '.body + "\n" + ([.comments[].body] | join("\n"))' \
+      | grep -qF "$RUN_URL"; then
```
tend-agent: Scope the guard to the generated row anchor so a human comment that merely mentions the run URL can't suppress the bot's own row:

```suggestion
      | grep -qF "[workflow run](${RUN_URL})"; then
```

shared/steps/report-failure.sh
```diff
@@ -49,8 +49,34 @@ sleep $((RANDOM % 30))
...
+
+  # The check-then-act above still races across concurrently-jittered legs: two
+  # can both read no matching row before either posts. Reconcile to one row per
+  # run — keep the earliest comment citing this RUN_URL, delete later dups.
+  # Convergent, mirroring the issue reconcile below: every racing leg sorts the
+  # same way and computes the same keeper, so deleting an already-deleted
+  # comment is a harmless 404.
+  sleep 5
+  gh api "repos/${GITHUB_REPOSITORY}/issues/${EXISTING}/comments?per_page=100" \
+    --jq "[.[] | select(.body | contains(\"${RUN_URL}\"))] | sort_by(.created_at) | .[1:] | .[].id" \
```
tend-agent: This delete selects any comment containing the bare URL. Match the full anchor so only the auto-generated rows are eligible for deletion — honoring "no human content is touched" — and avoid a run-id prefix collision:

```suggestion
    --jq "[.[] | select(.body | contains(\"[workflow run](${RUN_URL})\"))] | sort_by(.created_at) | .[1:] | .[].id" \
```

tend-agent (commented):
The dedup logic is sound and the convergence argument holds (stable `sort_by` over an id-ascending API response makes every leg pick the same keeper). One robustness gap on the delete path: the reconcile selects comments by a bare `contains("${RUN_URL}")` substring, with no author or row-format scoping. Any comment that merely contains the run URL — a human quoting it while investigating, or a nightly-enrichment comment (#560) that lists this run among others — would be selected and, if it sorts after the earliest row, deleted. That contradicts the PR's "no human content is touched" invariant, since the code doesn't actually restrict deletion to the auto-generated rows.

Matching the full markdown anchor `[workflow run](${RUN_URL})` instead scopes both the guard and the delete to the generated rows only, and as a bonus removes a latent prefix-collision false positive (a bare-URL `contains` also matches a longer run id that has this one as a prefix). The guard and the reconcile must stay consistent, so both suggestions below apply together.

shared/steps/report-failure.sh
tend-agent: Applied in 86f39cc — the guard now matches `grep -qF "[workflow run](${RUN_URL})"`, so a human comment that only mentions the run URL no longer suppresses the bot's own row, and a longer run id sharing this one as a prefix can't false-positive.

shared/steps/report-failure.sh
tend-agent: Applied in 86f39cc — the delete select is now `contains("[workflow run](${RUN_URL})")`, scoping deletion to the auto-generated rows only (honoring "no human content is touched") and closing the same prefix-collision gap. Guard and reconcile stay consistent, as noted.

tend-agent:
New evidence, same defect, roughly 7x the volume. The 2026-08-04 outage (Claude **weekly** limit, `429 rate_limit`, from ~08:51Z until the 00:00Z reset) left [#831](https://github.com/max-sixty/tend/issues/831) with **77 comment rows across 17 distinct runs** — 15 `review-reviewers` runs contributing exactly 5 rows each, one per matrix leg, plus 2 single-leg `tend-triage` rows. This PR's original evidence was ~11 rows from 2 runs.

Each of those 75 duplicate-leg rows also fired an `issue_comment` event, so the flood produced 66 `tend-mention` runs on top of the comments (all correctly gate-skipped, but they are 66 queued jobs). With this PR merged the same outage would have recorded 17 rows instead of 77 — one per run, which is the readable timeline the issue is for.

tend-agent:
Coordination note from #836, which is now touching the neighbouring branch of this same script.

Review there caught that #836's reconcile carries its row onto the keeper unguarded, so a same-matrix race would post a row duplicating the keeper's seed row — the same flood this PR removes from the `EXISTING` branch, arriving by a path that bypasses this PR's guard. Fixed in 1c41d39 with the identical four-line check (`gh issue view --json body,comments` piped to `grep -qF "[workflow run](${RUN_URL})"`).

That means the anchor check now exists twice in the file: once here on the append path, once on the create path. Leaving both copies rather than pre-factoring a helper — the two diffs are textually disjoint as they stand and a helper introduced on either branch would conflict with the other for no benefit until one lands. Whichever of the two merges second should fold them into one helper.

shared/steps/report-failure.sh
```diff
@@ -37,7 +38,43 @@ sleep $((RANDOM % 30))
...
+  # The check-then-act above still races across concurrently-jittered legs: two
+  # can both read no matching row before either posts. Reconcile to one row per
+  # run — keep the earliest comment carrying this run's anchor, delete the rest.
+  # Convergent, like the create reconcile in run_issue_create_and_reconcile:
+  # every racing leg sorts the same way and computes the same keeper, so
+  # deleting an already-deleted comment is a harmless 404. Selecting on the
+  # anchor keeps only the bot's own generated rows eligible for deletion.
+  sleep 5
+  gh api "repos/${GITHUB_REPOSITORY}/issues/${EXISTING}/comments?per_page=100" \
+    --jq "[.[] | select(.body | contains(\"${ANCHOR}\"))] | sort_by(.created_at) | .[1:] | .[].id" \
```
tend-agent: Only the first page is read, and this endpoint returns comments oldest-first — so on a tracker past 100 comments the rows just posted are never in the response and the reconcile no-ops. `--paginate` alone applies `--jq` per page, which would break the cross-page `sort_by | .[1:]`; `--slurp` refuses `--jq`, so move the filter downstream and flatten the pages with `add`:

```suggestion
  gh api --paginate --slurp "repos/${GITHUB_REPOSITORY}/issues/${EXISTING}/comments?per_page=100" \
    | jq -r "add | [.[] | select(.body | contains(\"${ANCHOR}\"))] | sort_by(.created_at) | .[1:] | .[].id" \
```

tend-agent (commented):
The guard is correct now — both earlier threads are addressed by 86f39cc, and hoisting the anchor into `run_issue_anchor` is the right shape: the row and the two dedup matchers can no longer drift apart.

One remaining defect on the reconcile, in the original commit rather than the incremental. The delete path lists comments with `?per_page=100` and no pagination, and GitHub returns issue comments oldest-first — so once a tracker passes 100 comments, the rows this run just posted are not in the response at all and the reconcile silently no-ops. That is the flooded issue the PR exists for: #831 reached 77 rows in a single outage, so crossing 100 on a longer one or a wider matrix is the expected case rather than the edge. The guard is unaffected — `gh issue view --json comments` does paginate — so only the second line of defence is capped.

`--paginate` alone will not fix it: `gh` applies `--jq` per page, which breaks the cross-page `sort_by(.created_at) | .[1:]` (each page would keep its own earliest). `sort=created&direction=desc` is ignored by this endpoint. `--paginate --slurp` refuses `--jq`, so the working form is `--slurp` piped to a downstream `jq 'add | …'` — which is also what `rate-limit-preflight.sh` already does one file over for `/issues/$PAUSE/events?per_page=100`.

Separately, this ships ~40 lines of racy dedup with no test, in a script every adopter runs. `generator/tests/test_shared_steps.py` already drives `rate-limit-preflight.sh` and `mark-notification-read.sh` against a fake `gh`, and #836 adds a `report_failure_env` fixture for this exact script — the guard (skip when the anchor is already present) and the reconcile (keep the earliest, delete the rest) both look cheap to cover once whichever of the two lands first.

<details><summary>How the pagination behaviour was verified</summary>

Against `cli/cli#13840`, which has 139 comments:

```
$ gh issue view 13840 -R cli/cli --json comments --jq '.comments | length'
139
$ gh api "repos/cli/cli/issues/13840/comments?per_page=100" --jq 'length'
100
$ gh api "repos/cli/cli/issues/13840/comments?per_page=100" --jq '[.[0].created_at, .[-1].created_at]'
["2026-07-10T13:54:16Z","2026-08-03T21:08:08Z"]     # oldest-first; newest 39 absent
$ gh api "repos/cli/cli/issues/13840/comments?per_page=100&sort=created&direction=desc" --jq '[.[0].created_at, .[-1].created_at]'
["2026-07-10T13:54:16Z","2026-08-03T21:08:08Z"]     # sort/direction ignored
$ gh api --paginate "repos/cli/cli/issues/13840/comments?per_page=100" --jq 'length'
100
39                                                   # --jq runs per page
$ gh api --paginate --slurp "repos/cli/cli/issues/13840/comments?per_page=100" --jq 'length'
the `--slurp` option is not supported with `--jq` or `--template`
$ gh api --paginate --slurp "repos/cli/cli/issues/13840/comments?per_page=100" | jq -r 'add | length'
139
```

The empty case needs no extra guard: on a zero-comment issue `--paginate --slurp` yields `[[]]`, so `add` gives `[]` and the filter emits nothing, exit 0.

</details>

shared/steps/report-failure.sh
tend-agent: Fixed in `834734e` — took the suggestion as written.

One correction to my own note above: the precedent in `rate-limit-preflight.sh` is `--paginate --jq`, not `--slurp`. That works there because its filter is per-element (`.[] | select(...)`), so applying it per page is the same as applying it once. Here the filter aggregates across the whole set (`sort_by | .[1:]`), which is what makes `--slurp` plus a downstream `jq` the only correct form — so this is the first use of that shape in the repo rather than a second one.

tend-agent:
Both points addressed in `834734e`.

**Pagination** — replied inline; took the suggestion as written.

**Tests** — added five against a fake `gh`, in the same shape as the existing `rate-limit-preflight.sh` suite (fixtures in, the script's own filters doing the work). Since neither this nor #836 has landed, this one carries its own `report_failure_env` rather than waiting on the other:

| Test | Covers |
|---|---|
| `skips_a_run_already_recorded` | the guard — a leg whose sibling already posted this run's row posts nothing |
| `appends_a_row_for_an_unrecorded_run` | the guard doesn't swallow a genuinely new run |
| `reconciles_a_racing_leg` | two legs that both missed the guard converge to one row |
| `reconciles_past_the_first_page` | the pagination defect: 139 comments, the racing row at the end |
| `leaves_a_human_comment_naming_the_run` | the anchor scoping from 86f39cc — a person linking the run isn't a delete candidate |

Only `reconciles_past_the_first_page` fails against the pre-fix script; the other four pass on both, which is the intended split — one regression test for the defect just fixed, four characterising what the original commit already did.

The fake pages the comment list the way the endpoint does (`--slurp` gets the array of pages, a plain read gets the oldest 100 alone) rather than special-casing the flag the script happens to pass, so that counterfactual is honest rather than a fake built to agree with the fix.

**Coordination with #836** — both PRs now add a `report_failure_env` fixture and a `FAKE_GH_REPORT_FAILURE` to the end of `generator/tests/test_shared_steps.py`, so whichever lands second will conflict there. I kept every shared name and env var identical to #836's (`REPORT_FAILURE`, `OUTAGE_TITLE`, `OUTAGE_LABEL`, `RUN_LINK`, `OPEN_ISSUES_JSON`, `KEEPER_JSON`, `COMMENT_BODIES`, `FAKE_NEW_ISSUE`) so the resolution is a union rather than a rewrite: take both sets of `case` arms in the fake — #836's `repos/*/issues/N` primary-key probe and this one's `/comments?per_page=100` pager — and both sets of tests. The two test sets don't overlap: #836's exercise the create path, these the append path.

```
356 passed  (generator/)
pre-commit: shellcheck, ruff, typos — all pass
```

generator/tests/test_shared_steps.py
```diff
@@ -575,3 +575,263 @@ def test_rate_limit_reopens_rather_than_refiling(
...
+
+jq_expr=""
+slurp=""
+prev=""
+for arg in "$@"; do
+  [ "$prev" = "--jq" ] && jq_expr="$arg"
+  [ "$arg" = "--slurp" ] && slurp=1
+  prev="$arg"
+done
+
```
tend-agent, line 1053: The fake applies `$jq_expr` whatever flags it was handed, so `--slurp` plus `--jq` is accepted here and rejected in production. Folding the filter back into `gh --jq` is the obvious simplification of the reconcile — it is exactly what the script's comment argues against — and as it stands all five tests stay green while the script dies under `pipefail` right after posting its row, so the reconcile never runs and the flood the PR exists to stop comes back. Modelling the refusal is what keeps the tests from blessing it.

```suggestion
done

# Real `gh` refuses the combination outright and exits 1, and the reconcile's
# shape rests on that: fold the filter back into `--jq` and the script dies
# under pipefail just after posting its row, never reconciling.
if [ -n "$slurp" ] && [ -n "$jq_expr" ]; then
  echo "the --slurp option is not supported with --jq or --template" >&2
  exit 1
fi

```

tend-agent (commented):
The pagination fix is right and the regression test is honest — I re-ran `test_report_failure_reconciles_past_the_first_page` against `4e93d87`'s script and it fails there (`deleted []`, expected `['140']`), passes here. `gh api --paginate --slurp` on a zero-comment issue does return `[[]]` as the comment claims (checked against #879 in this repo), so `add` → `[]` → no output, exit 0.

Two gaps in what the new suite pins down, both verified by mutating the script and re-running:

1. **The fake accepts `--slurp` with `--jq`; real `gh` exits 1 on it.** That combination is the exact trap the script's comment exists to warn about, and the fake currently blesses it.
2. **The guard's issue-body half is never exercised** — `_seen_by_the_guard` hardcodes `body: ""`, so every case reaches the guard through the comments list only. That's the minority path: on the first failed run of an outage one leg *creates* the issue with its row in the body, and its four siblings arrive at a tracker with no comments at all, matchable only on the body.

Suggestions inline. The second and third apply together — the keyword arg alone is inert, the parametrize alone is a `TypeError`.

<details><summary>How each gap was confirmed</summary>

Two mutations of `shared/steps/report-failure.sh`, each run against the suite as it stands and against the suite with the suggestions applied:

| Mutation | As it stands | With suggestions |
|---|---|---|
| `\| jq -r "add \| …"` → `--jq "add \| …"` (fold the filter back into `gh`) | 5 passed | 4 failed — `returncode=1`, `stderr='the --slurp option is not supported with --jq or --template'` |
| guard's `--jq` loses `.body + "\n" +` | 5 passed | 1 failed — `skips_a_run_already_recorded[in-the-issue-body]` |

And the underlying `gh` behaviour, on 2.97.0:

```
$ gh api --paginate --slurp "repos/max-sixty/tend/issues/879/comments?per_page=100"
[[]]
$ gh api --paginate --slurp "repos/max-sixty/tend/issues/879/comments?per_page=100" --jq 'length'
the `--slurp` option is not supported with `--jq` or `--template`
$ echo $?
1
```

Also checked: `--slurp` is new to this repo — `rate-limit-preflight.sh:114` pairs `--paginate` with a streaming `.[] | select(…)` filter, which is safe per-page and so isn't the precedent the earlier review cited. Nothing to fix there; the filter has no cross-page reduction.

Not flagged, for the record: `_nwise` in the fake is a private jq builtin, but a removal would break the tests loudly rather than silently. #857 also touches this file, but only its header comment.

</details>

generator/tests/test_shared_steps.py
```diff
@@ -575,3 +575,263 @@ def test_rate_limit_reopens_rather_than_refiling(
...
+        _comment(first_id + i, f"nightly enrichment {i}", f"2026-01-01T00:{i:02d}:00Z")
+        for i in range(count)
+    ]
+
+
+def _seen_by_the_guard(env: dict[str, str], *bodies: str) -> None:
+    """What `gh issue view --json body,comments` returns for the tracker."""
+    Path(env["KEEPER_JSON"]).write_text(
+        json.dumps({"body": "", "comments": [{"body": b} for b in bodies]})
+    )
```
tend-agent: Pairs with the parametrize suggestion below — this one is inert on its own, and that one is a `TypeError` without it.

```suggestion
def _seen_by_the_guard(env: dict[str, str], *bodies: str, body: str = "") -> None:
    """What `gh issue view --json body,comments` returns for the tracker."""
    Path(env["KEEPER_JSON"]).write_text(
        json.dumps({"body": body, "comments": [{"body": b} for b in bodies]})
    )
```

generator/tests/test_shared_steps.py
```diff
@@ -575,3 +575,263 @@ def test_rate_limit_reopens_rather_than_refiling(
...
+def test_report_failure_skips_a_run_already_recorded(
+    report_failure_env: dict[str, str],
+) -> None:
+    """A leg whose sibling already recorded this run posts nothing.
+
+    This is the guard that collapses the flood: a matrix workflow calls the
+    script once per leg, every leg sharing one GITHUB_RUN_ID, so without it a
+    5-leg matrix leaves 5 comments all citing the same run.
+    """
+    _seen_by_the_guard(report_failure_env, f"| when | {RUN_LINK} | #851 |")
```
tend-agent: The guard matches `.body + "\n" + comments`, but every case here reaches it through the comments list, so the body half is unpinned: dropping `.body + "\n" +` from the script's `--jq` leaves all five tests green. That half is the one the first failed run of an outage depends on — one leg takes the `else` branch and seeds the issue with its row in the *body*, and its four siblings then hit the `EXISTING` branch against a tracker with no comments at all. Parametrizing costs one extra case:

```suggestion
@pytest.mark.parametrize(
    ("body", "comments"),
    [
        pytest.param("", (f"| when | {RUN_LINK} | #851 |",), id="in-a-comment"),
        pytest.param(f"| when | {RUN_LINK} | #851 |", (), id="in-the-issue-body"),
    ],
)
def test_report_failure_skips_a_run_already_recorded(
    report_failure_env: dict[str, str], body: str, comments: tuple[str, ...]
) -> None:
    """A leg whose sibling already recorded this run posts nothing.

    This is the guard that collapses the flood: a matrix workflow calls the
    script once per leg, every leg sharing one GITHUB_RUN_ID, so without it a
    5-leg matrix leaves 5 comments all citing the same run. The body case is
    the first run of an outage: one leg seeds the issue with its row, and the
    siblings that follow have no comment to match — only the body.
    """
    _seen_by_the_guard(report_failure_env, *comments, body=body)
```

generator/tests/test_shared_steps.py
tend-agent: Applied in `bb0a92d`. Confirmed the mutation now bites: folding the filter back into `--jq` fails 4 of the 6 with `the --slurp option is not supported with --jq or --template`, where before it left all five green.

generator/tests/test_shared_steps.py
tend-agent: Applied in `bb0a92d`, together with the parametrize below.

generator/tests/test_shared_steps.py
tend-agent: Applied in `bb0a92d`. Confirmed against the mutation: dropping `.body + "\n" +` from the guard's `--jq` now fails `skips_a_run_already_recorded[in-the-issue-body]` and nothing else, so the new case is carrying that half on its own.

tend-agent (commented):
Both gaps from the last review are closed in `bb0a92d`, re-derived here rather than taken on the commit message's word — I re-applied each mutation to `shared/steps/report-failure.sh` against this HEAD:

| Mutation | Result on `bb0a92d` |
|---|---|
| `\| jq -r "add \| …"` → `--jq "add \| …"` | 4 failed, `returncode=1`, `stderr='the --slurp option is not supported with --jq or --template'` |
| guard's `--jq` loses `.body + "\n" +` | 1 failed — `skips_a_run_already_recorded[in-the-issue-body]` only |

The fake's refusal also matches real `gh` on this runner (2.96.0): `gh api --slurp --jq '.' …` exits 1 with that message, and it is a parse-time rejection, so placing the check ahead of the subcommand dispatch is the right position for it. The body-half case is faithful too — the `else` branch seeds the issue through `run_issue_create_and_reconcile` with `$ROW`, which carries the anchor, so a sibling leg really does arrive at a tracker whose only match is the body.

`357 passed` in `generator/`, `ci` green on this commit. No further findings; not approving only because the PR is self-authored.

tend-agent:
Merged `main` in during the nightly sweep. The shell halves auto-merged, but `generator/tests/test_shared_steps.py` conflicted in 9 places — #836 landed as `ef7baa1` and rewrote the report-failure suite, so the two sides disagreed about the *fixtures*, not just which tests to keep. I resolved it as the union this PR's earlier comment prescribed rather than picking a side:

- **Base**: `main`'s file wholesale, so all fifteen of its report-failure tests keep the fixtures they were written against.
- **Fake**: `main`'s `FAKE_GH_REPORT_FAILURE`, extended with this branch's `--slurp` detection and the real-`gh` refusal of `--slurp` with `--jq`; the `issue comment` arm now also lands the row in `ISSUE_COMMENTS_JSON` so the reconcile sees what it just posted. The inner dispatch moved from `case "$2"` to `case "$*"` — this branch's two calls carry flags (`--paginate`, `-X DELETE`) where the path would otherwise sit — with the `/comments?per_page=100` and `-X DELETE` arms ahead of #836's primary-key probe, since all three URLs contain `/issues/`.
- **Fixture**: `report_failure_env` gains `ISSUE_COMMENTS_JSON` and `POSTED_AT`. It keeps `main`'s empty `open-issues.json` default, because the create-path tests need it; the five append-path tests here seed it themselves via a new `_open_tracker` helper.
- **Tests**: both sets, unmodified apart from that `_open_tracker` call. They don't overlap — #836's exercise the create path, these the append path.

Verified the port still discriminates: with `report-failure.sh` and `lib/run-issue.sh` reverted to `main` in place, `skips_a_run_already_recorded` (both params), `reconciles_a_racing_leg`, and `reconciles_past_the_first_page` fail, while `appends_a_row_for_an_unrecorded_run` and `leaves_a_human_comment_naming_the_run` pass — the same split this PR described. Full generator suite: 385 passed. `ruff check` / `ruff format` clean.

Worth noting the premise is still live: [#905](https://github.com/max-sixty/tend/issues/905) currently carries exactly the flood this fixes — run `31282882062` in 5 separate comments and `31285087877` in 5 more, each a one-row table differing only in its jitter-spread timestamp, both from `review-reviewers` matrix runs last night.

shared/steps/report-failure.sh
```diff
@@ -57,9 +80,37 @@ if [ -n "$EXISTING" ]; then
...
+  # `add` flattening the array of pages. On an issue with no comments `--slurp`
+  # yields `[[]]`, so `add` gives `[]` and the filter emits nothing.
+  sleep 5
+  gh api --paginate --slurp \
+    "repos/${GITHUB_REPOSITORY}/issues/${EXISTING}/comments?per_page=100" \
+    | jq -r "add | [.[] | select(.body | contains(\"${ANCHOR}\"))] | sort_by(.created_at) | .[1:] | .[].id" \
+    | while read -r DUP_ID; do
+        [ -z "$DUP_ID" ] && continue
+        gh api -X DELETE "repos/${GITHUB_REPOSITORY}/issues/comments/${DUP_ID}" 2>/dev/null || true
+      done
```
tend-agent: The reconcile is best-effort cleanup — a failed read leaves duplicate rows on the tracker, which is strictly better than reddening a step whose write already landed. Guarding it here keeps the block consistent with the append above and the list read at the top, both of which warn and continue.

```suggestion
      done || echo "::warning::Could not reconcile duplicate rows on #${EXISTING}; this run's row is recorded."
```

tend-agent (commented):
One finding, from the `main` merge in `6894b68` rather than from `55bb271` — the anchor fold is right, and `run_issue_anchor` is now the only place the `[workflow run](…)` string is built (`grep -rn 'workflow run\]('` over the tree returns just `lib/run-issue.sh:61`).

The merge brought main's failed-write policy onto the append — a 5xx on `gh issue comment` now warns and `exit 0`s instead of aborting — but the reconcile pipeline directly below it stayed bare. It is the last statement in the `EXISTING` branch, so under `set -eo pipefail` a failure on `gh api --paginate --slurp` takes the whole script's exit status with it, *after* the row was successfully posted. The result is a red `Report failure` step carrying no annotation to say why, on precisely the job whose failure someone is about to diagnose — and the append immediately above it went to some length to avoid exactly that.

The deletes inside the loop already carry `|| true`, and both neighbouring fallible operations (the `run_issue_canonical` read at the top, the append) warn and continue, so the reconcile's read is the one unguarded operation left in the block. Suggestion inline.

<details><summary>How the abort was confirmed</summary>

Against this HEAD, with a fake `gh` that 502s only on `…/comments?per_page=100` and answers everything else normally (open tracker `#8`, guard sees no anchor, so the append path runs and posts the row):

```
$ bash shared/steps/report-failure.sh
gh: 502 server error
EXIT=1
```

No `::warning::`, and the row *was* posted — the step reddens over a completed write. The `[ -z "$DUP_ID" ] && continue` inside the loop is fine, incidentally: a non-last component of an AND-list is exempt from `errexit`, and the loop exits 0 on the normal path.

With ` || echo "::warning::…"` appended to `done`, the same run exits 0 and emits the annotation. `||` binds to the whole pipeline, so no re-indentation is needed.

Not flagged: `test_report_failure_survives_a_failed_append_to_the_open_tracker` doesn't discriminate the merge's new `exit 0` — without it the leg would fall through to a reconcile that finds nothing of its own to delete, so the observable behaviour is the same. It's an optimisation, not a correctness fix, and not worth a test.

</details>

shared/steps/report-failure.sh
tend-agent: Took the suggestion as written, in `3289113`, plus a comment above the block pointing at the append's guard for the reason.

Confirmed the abort first rather than taking the review's word for it — added `test_report_failure_survives_a_failed_reconcile_read`, which 502s the fake `gh` only on `…/comments?per_page=100`. Against the parent commit it fails exactly as described: `returncode=1`, empty stdout, `stderr='gh: 502 server error'`, and the row already in the comment list — a red step over a completed write, no annotation. With the guard it exits 0, emits the `::warning::`, and the row is still there.

The test asserts all three, since the row surviving is what makes the degradation acceptable: a failed reconcile leaves duplicate rows on the tracker, which is the loss this trades for.

tend-agent (commented):
The reconcile guard in `3289113` closes the last finding, re-derived here rather than taken on the commit message's word. Reverting just the `|| echo "::warning::…"` from `done` and re-running the suite fails exactly one test — `test_report_failure_survives_a_failed_reconcile_read`, `returncode=1`, `stderr='gh: 502 server error'` — and nothing else, which is the right discrimination: the other fourteen characterise behaviour the guard doesn't change. With it in place the same fake exits 0, emits the annotation, and `COMMENT_BODIES` still carries the row, so the write that already landed is not lost to the cleanup that failed after it.

`|| ` binding to the whole pipeline rather than the `while` alone is what makes the one-line form correct, and the mutation run exercises that end to end — the failure is injected on `gh api --paginate --slurp`, the first component, and is still caught. That also leaves the deletes' inner `|| true` doing its own job unchanged.

`419 passed` in `generator/`; `test`, `lint`, `test-proxy`, `test-codex-surface`, `test-install-tend-scripts`, `test-worker` all green on this commit. No further findings; not approving only because the PR is self-authored.

PR #813, opened 2026-08-03
chore: regenerate workflows with tend 0.1.13

max-sixty:
Regenerates tend's own `tend-*.yaml` workflows against the published 0.1.13, the release's last step, and restamps the one hand-maintained workflow `init` doesn't reach.

All eight generated workflows move from `max-sixty/tend/claude@0.1.12` to `@0.1.13` and name `environment: tend`, so once the secrets move into that environment the jobs can still read them. `tend-mention` gains the secretless `relay` job: a review or review-comment event runs on `refs/pull/N/merge`, a ref no deployment-branch policy can admit, so `relay` holds no secrets and re-posts only identifiers as a `repository_dispatch`, which `verify` then re-reads from the API. `review-reviewers.yaml` is hand-maintained, so `init` leaves its ref alone; the `running-tend` restamp convention folds that one-line bump into the same commit, which is what the second commit does.

`TODO.md` records what the migration still needs, now that the environment half is done everywhere. Two things remain. The model credential can't be read back and isn't stored locally, so `CLAUDE_CODE_OAUTH_TOKEN` has to be pasted or re-minted into each environment before the repo-level copy can go — only generated workflows read it, and all of them name the environment. And worktrunk and prql keep a repo-level `TEND_BOT_TOKEN` because hand-maintained workflows read it on `pull_request`, `pull_request_target` and `release` triggers, whose refs a `main`-only policy can never admit; naming the environment there would deny them the secret rather than gate it, so each needs its own decision.

Nothing breaks on merge: the `tend` environment already holds `TEND_BOT_TOKEN`, and a job naming an environment still sees repo-level secrets, so the gate closes only when the repo-level copies are deleted.


tend-agent (commented):
The regeneration is faithful: I re-ran `tend init` from the in-tree 0.1.13 generator against `.config/tend.yaml` and got zero drift, so the eight files match generator output exactly, `@0.1.13` resolves (the tag points at the release commit and carries `claude/action.yaml`), and no hand edits slipped in.

One consistency gap sits just outside the generated set. `review-reviewers.yaml` is hand-maintained — `init` doesn't rewrite it — and its `- uses:` step still pins `max-sixty/tend/claude@0.1.12`, so once this merges that workflow runs a release behind the other eight. It already carries `environment: tend`, so the only lag is the ref. The `running-tend` restamp convention folds that one-line bump into the same regen commit; happy to push `@0.1.12` → `@0.1.13` onto this branch if you'd like it here, otherwise the nightly restamp picks it up.

tend-agent (approved):

PR #815, opened 2026-08-03
check: sweep credential-holding environments, not just secret-holding ones

max-sixty:
#810's `secret-environments` check credits an environment as needing a gate only if it stores a secret. A repo that publishes via trusted publishing (PyPI, npm, a cloud role) stores nothing, so the sweep walked past exactly the environments that matter — `numbagg/numbagg`'s `pypi` and `test-pypi` are ungated today and the check reported PASS. This widens the sweep from secrets to credentials and renames it `credential-environments`.

Three gaps, each with tests:

**OIDC-holding environments count.** An environment holds a credential if it stores a secret *or* a job requesting `id-token: write` deploys to it. The token's `sub` names the environment, so the environment is the whole gate on GitHub's side.

**A ref policy loses to a trigger the bot steers.** `release: published`, `repository_dispatch`, and `workflow_dispatch` *with inputs* let a write-scoped actor supply the run's payload and fire it at a ref the policy already admits — so the policy gates nothing and only a required reviewer covers them. Verified live against GitHub with a write-access, non-admin, non-bypass collaborator; `push`, `create`, `pull_request`, `workflow_run`, `deployment`, `schedule` and an input-free `workflow_dispatch` are not steerable, since the code they run is fixed by the ref.

**`id-token: write` outside any environment has no gate.** The token carries no environment claim and the bot can mint it from a branch it pushes, which any trust policy pinning the repository but not the ref accepts. Reported on its own line.

Reading the workflow surface takes one GraphQL call for every blob under `HEAD:.github/workflows`, then a fixpoint over `workflow_call` to give a reusable workflow its callers' triggers. `_credential_surface` returns what the sweep needs: which environments a steerable trigger reaches, which hold OIDC, which OIDC jobs are ungated, and what couldn't be resolved (reported as UNKNOWN rather than PASS).

One deliberate conservatism, noted in `_policy_gate`'s docstring: a workflow with a steerable trigger counts even when an `if:` on the deploying job would skip that event. Reading GHA expressions to decide otherwise is the same matcher re-implementation #810 declined, and the conservative answer fails closed.

Also drops the unused `id-token: write` grant from the generated workflows and `review-reviewers.yaml` — tend's own agent runs never mint an OIDC token, and the grant would have tripped the new check.

<details>
<summary>Live results across the five adopters</summary>

| Repo | Result |
|---|---|
| `numbagg/numbagg` | FAIL — `pypi` and `test-pypi` have no deployment policy. Neither stores a secret, so #810's sweep skipped both. |
| `max-sixty/cargo-affected` | FAIL — `release` has no policy. |
| `PRQL/prql` | FAIL — `github-pages` has no policy; `release` is ref-pinned but reached on `release:`; 10 jobs mint OIDC outside any environment. |
| `max-sixty/tend`, `max-sixty/worktrunk` | FAIL only on the stale committed `id-token: write` in their generated `tend-*.yaml`, which clears at the next release plus nightly regen. |

No false positive on tend's own gated environments: `release` (pinned `*` tags under the all-tags admin ruleset) and `github-pages` (pinned `main`, reached only by `push` and an input-free `workflow_dispatch`) both pass.

Those are other people's repo settings; remediation is the recipe in the install skill, not part of this PR.

</details>

`README.md`, the install skill's release bullet, and `TODO.md`'s migration note carried the pre-rename "secret-holding" wording and are updated to match.

Testing: 17 tests cover the new paths. Mutation-checked — removing the OIDC-environment crediting fails 1 test, removing the steerable refusal fails 3, removing the ungated-OIDC report fails 2.


tend-agent (commented):
The check itself looks solid — the OIDC-crediting, steerable-trigger, and `workflow_call` fixpoint logic all read correctly, the UNKNOWN-over-PASS ordering fails closed as intended, and the 17 new tests cover the paths well. One loose end: two user-facing docs still describe the sweep in the pre-rename "secret-holding" terms, which now understates it — the headline case this PR adds is exactly the trusted-publishing environment that stores *no* secret.

- **`README.md`**, the "Environment-gated secrets" section: *"The same check sweeps every other secret-holding environment (release tokens included) for a gate the bot cannot pass"*. After this PR the sweep is `credential-environments` and also covers environments that hold no secret but are named by an `id-token: write` job, plus OIDC minted outside any environment. Worth rewording to match, since this section is the project's front-door description of the boundary.

- **`plugins/install-tend/skills/install-tend/SKILL.md`**, the repo-level "Release secrets" bullet: *"`tend check` sweeps every secret-holding environment and fails on any it cannot confirm gated"*. The new section you added lower in the same file already uses the `credential-environments` framing, so this earlier line is just internally out of step.

Both are outside this PR's diff, so no inline suggestion — happy to push a commit updating the two phrasings if you'd like, or leave it to you as a follow-up.

max-sixty:
Both folded in.

`README.md`'s "Environment-gated secrets" is now "Environment-gated credentials", and says what makes an environment count: it stores a secret, or a job requesting `id-token: write` deploys to it. It also names the steerable-trigger case, since a ref policy alone isn't the gate any more.

The install skill's release bullet gets the same widening, plus the three triggers spelled out — an adopter reading that bullet is exactly the person whose `release`-triggered publish workflow will trip the check.

`TODO.md`'s migration note had the same narrow phrasing and is updated too.

tend-agent (approved):

PR #819, opened 2026-08-03
fix(review): port /code-review into a tend-owned skill

### LINKED ISSUE #817: review: /code-review second pass is unreachable — disable-model-invocation since Claude Code 2.1.215
The `review` skill's step 4 mandates a second pass: "Review the diff two ways at once: the manual checks below, plus a `/code-review` over it (Claude harness only; the Codex harness lacks the command, so skip it there)." As of Claude Code **2.1.215**, `code-review` carries `disable-model-invocation`, so the agent cannot reach it through the `Skill` tool. Every Claude-harness `tend-review` run that follows the instruction gets a tool-use error instead of the second pass, and the review silently degrades to manual-only.

```
<tool_use_error>Skill code-review cannot be used with Skill tool due to disable-model-invocation</tool_use_error>
```

This is structural rather than stochastic — the error is deterministic on every invocation, so no amount of retrying or rephrasing reaches it. It's also invisible from the outside: the reviews still get posted and still read as complete, so nothing downstream surfaces the missing pass.

## Evidence — worked at 2.1.207, blocked at 2.1.215+

Session logs record the Claude Code version, which dates the regression precisely:

| Run | Date | Claude Code | `Skill: code-review` result |
| --- | --- | --- | --- |
| [PRQL/prql 30058579150](https://github.com/PRQL/prql/actions/runs/30058579150) (review of PRQL/prql#6100) | 2026-07-24 | 2.1.207 | `Launching skill: code-review` — **worked** |
| [PRQL/prql 30758492810](https://github.com/PRQL/prql/actions/runs/30758492810) (review of PRQL/prql#6124) | 2026-08-02 | 2.1.215 | **blocked** |
| [PRQL/prql 30780160485](https://github.com/PRQL/prql/actions/runs/30780160485) (review of PRQL/prql#6125) | 2026-08-03 | 2.1.215 | **blocked** |
| [PRQL/prql 30798947165](https://github.com/PRQL/prql/actions/runs/30798947165) (this review-runs run) | 2026-08-03 | 2.1.220 | **blocked** (reproduced directly) |

Both blocked runs were on the same PRQL repo with the same `.claude/settings.local.json` the action generates, which allowlists `Skill` — so this isn't a permissions denial, it's a property of the skill itself. `code-review` also no longer appears in the agent's available-skills listing, which is consistent: the listing excludes `disable-model-invocation` skills, while `review`, `security-review`, and `simplify` are all still listed and invocable.

## What the agent did instead

Both runs recovered gracefully and fell back to the manual checklist — no wrong verdict shipped:

- On PRQL/prql#6124 (self-authored CI change): *"The `/code-review` built-in is blocked from Skill-tool invocation in this environment (`disable-model-invocation`), so I'll rely on the thorough manual review already completed."*
- On PRQL/prql#6125 (human-authored resolver refactor): *"The `/code-review` built-in isn't invokable as a tool in this environment; my manual pass was thorough."*

So the cost isn't a bad review today — it's that the skill's stated two-pass design is now a one-pass design everywhere, and the fallback is left to the model to improvise rather than being specified.

## Suggested fixes

Roughly in order of how much they preserve the original intent:

1. **Replace the built-in with a subagent pass.** Step 4 spawns an `Agent` (`general-purpose`) with a correctness-and-cleanups brief over the merged tree, scaled by the same peripheral/core tiering the current text already defines. Keeps two independent passes, and is reachable from the model.
2. **Soften the instruction and specify the fallback.** Keep `/code-review` as best-effort, but say explicitly that it may be unreachable and that the manual checklist is then the whole review — so the behaviour is designed rather than improvised, and the effort tiering still guides the manual depth.
3. **Drop it.** If the built-in isn't coming back for model invocation, the paragraph is dead weight that costs a wasted turn per review.

Happy to open a PR for whichever direction you prefer.

<details><summary>How this was found</summary>

Surfaced by PRQL's daily `tend-review-runs` sweep over the 2026-08-02→08-03 window. Two of the window's six `tend-review` sessions attempted the second pass; both were blocked. The version bisect came from the `version` field in the `claude-session-logs-*` JSONL artifacts:

```bash
jq -r 'select(.version != null) | .version' "$FILE" | head -1
```

and the block itself from the tool results:

```bash
jq -r 'select(.type=="user") | .message.content[]? | select(.type=="tool_result")
  | (.content|tostring) | select(test("disable-model-invocation"))' "$FILE"
```

Intermediate PRQL review runs between 2026-07-24 and 2026-08-02 (30564994898, 30613087188, 30650345623, 30689792689) were already on 2.1.215 but never attempted the second pass, which is why the regression went unnoticed for over a week — another sign it's an invisible failure mode rather than one CI will report.

Instruction text quoted from `plugins/tend-ci-runner/skills/review/SKILL.md`, step 4 ("Review").

</details>

tend-agent:
Fixes #817. Revised per the maintainer's direction on that thread — the first commit's `general-purpose` subagent brief is replaced by a tend-owned skill carrying the built-in's own prompt.

`code-review` carries `disable-model-invocation`, so `Skill(code-review)` returns a tool-use error and the `review` skill's step-4 second pass silently degrades to manual-only. This ports the built-in's prompt into `plugins/tend-ci-runner/skills/code-review/SKILL.md` and points step 4 at `/tend-ci-runner:code-review`.

What the ported skill keeps from the built-in: the diff-gathering phase, the five correctness angles (line-by-line scan, removed-behavior auditor, cross-file tracer, language-pitfall specialist, wrapper/proxy correctness), the four cleanup angles (reuse, simplification, efficiency, altitude) plus the CLAUDE.md conventions angle, the cleanup-outranked-by-correctness precedence rule, the CONFIRMED/PLAUSIBLE/REFUTED verdict ladder with its recall-biased "PLAUSIBLE by default" guidance, and the gap sweep. What it drops: the effort-tier parameterization (collapsed to peripheral vs. core, matching how step 4 already scales the manual pass), the `ReportFindings` output contract and the artifact-publishing step — the caller owns the output, so the skill returns findings and posts nothing.

Two things fall out of owning it rather than calling the built-in. It's reachable from the model, which was the bug. And it runs on **both** harnesses — the Codex harness previously skipped the second pass entirely, and it resolves `/tend-ci-runner:NAME` skills, so it gets the pass now.

<details><summary>Provenance and the alternative that exists</summary>

The prompt was read out of the `claude` binary this PR's CI runs against (2.1.220, `/home/tend-sandbox/.local/share/claude/versions/2.1.220`), from the prompt cells the `/code-review` skill assembles per effort tier and the workflow script it registers. Angle text and the verdict ladder are near-verbatim; the phase framing is rewritten for the tend caller.

There is one way to reach the built-in, and it's why the skill this PR adds is the better answer rather than the only one. The gate is `disableModelInvocation && !userTypedThisTurn` — and `userTypedThisTurn` is a regex (`(?<!\S)/code-review(?=$|\s)`) over the non-meta, non-tool-result user messages of the current turn, with subagents excluded outright (`agentId !== undefined` returns false). So a top-level prompt containing the literal token `/code-review` re-permits the `Skill` call for that turn. That would mean threading a slash-command token through the composite action's prompt to satisfy an undocumented internal check, in exchange for a pass whose shape changes with each Claude Code release. Details in the issue thread.

</details>

### DIFF
--- a/plugins/tend-ci-runner/skills/code-review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/code-review/SKILL.md
@@ -1,0 +1,96 @@
+---
+name: code-review
+description: Structured second pass over a diff — correctness and cleanup angles, a verify pass, ranked findings returned to the caller. Use as the independent pass inside a PR review, or when asked for a thorough diff review.
+argument-hint: "[target]"
+metadata:
+  internal: true
+---
+
+# Code review
+
+A structured pass over a diff that returns ranked findings. Ported from Claude Code's built-in `/code-review` (read out of the 2.1.220 binary), which carries `disable-model-invocation` and so cannot be reached with the `Skill` tool. This copy is tend-owned: reachable from the model, and it runs on both harnesses.
+
+**Return findings; don't act on them.** No review, comment, commit, or artifact from this skill — the caller folds the findings into its own single review.
+
+## Phase 0 — Gather the diff
+
+If a target was passed (PR number, branch, ref range, path, or a free-form scope instruction), review that. Otherwise run `git diff @{upstream}...HEAD` (falling back to `git diff main...HEAD` or `git diff HEAD~1`); if there are uncommitted changes, or the range diff is empty, also run `git diff HEAD`. That unified diff is the review scope.
+
+Read the CLAUDE.md files that govern the changed code: the repo-root one, plus any `CLAUDE.md`/`CLAUDE.local.md` in a directory that is an ancestor of a changed file (a directory's CLAUDE.md applies only to files at or below it).
+
+## Phase 1 — Find candidates
+
+Work through the angles below. Each surfaces candidates with `file`, `line`, a one-line `summary`, and a concrete `failure_scenario`.
+
+Scale the fan-out to the change, the same way the caller scales its own depth:
+
+- **Peripheral or mechanical** (docs, config, dependency bumps, test-only): angles A–C plus the cleanup angles, in one pass, in this context. Up to 4 candidates each.
+- **Core logic**: every angle, one subagent per angle where a subagent tool is available (`Task`/`Agent`), up to 6 candidates each. If no subagent tool is available, work through every angle yourself, in sequence, in one pass — do not skip angles for lack of fan-out.
+
+Don't let one angle's conclusions suppress another's: if two angles flag the same line for different reasons, record both. Pass every candidate with a nameable failure scenario through to Phase 2 — finders that silently drop half-believed candidates bypass the verify step and are the dominant cause of misses.
+
+### Angle A — line-by-line diff scan
+
+Read every hunk in the diff, line by line. Then read the enclosing function for each hunk — bugs in unchanged lines of a touched function are in scope (the PR re-exposes or fails to fix them). For every line ask: what input, state, timing, or platform makes this line wrong? Look for inverted/wrong conditions, off-by-one, null/undefined deref, missing `await`, falsy-zero checks, wrong-variable copy-paste, error swallowed in catch, unescaped regex metachars.
+
+### Angle B — removed-behavior auditor
+
+For every line the diff deletes or replaces, name the invariant or behavior it enforced, then search the new code for where that invariant is re-established. If you can't find it, that's a candidate: a removed guard, a dropped error path, a narrowed validation, a deleted test that was covering a real case.
+
+### Angle C — cross-file tracer
+
+For each function the diff changes, find its callers (grep for the symbol) and check whether the change breaks any call site: a new precondition, a changed return shape, a new exception, a timing/ordering dependency. Also check callees: does a parallel change in the same PR make a call unsafe?
+
+### Angle D — language-pitfall specialist
+
+Scan for the classic pitfalls of the diff's language or framework — for example: JS falsy-zero, `==` coercion, closure-captured loop var; Python mutable default args, late-binding closures; Go nil-map write, range-var capture; SQL injection; timezone/DST drift; float equality. Flag any instance the diff introduces.
+
+### Angle E — wrapper/proxy correctness
+
+When the PR adds or modifies a type that wraps another (cache, proxy, decorator, adapter): check that every method routes to the wrapped instance and not back through a registry/session/global — e.g. a caching provider holding a `delegate` field that resolves IDs via `session.get(...)` instead of `delegate.get(...)` will re-enter the cache or recurse. Also check that the wrapper forwards all the methods the callers actually use.
+
+### Reuse
+
+The angles above hunt for bugs; this one and the next three hunt for cleanup in the changed code. Flag new code that re-implements something the codebase already has — grep shared/utility modules and files adjacent to the change, and name the existing helper to call instead.
+
+### Simplification
+
+Flag unnecessary complexity the diff adds: redundant or derivable state, copy-paste with slight variation, deep nesting, dead code left behind. Name the simpler form that does the same job.
+
+### Efficiency
+
+Flag wasted work the diff introduces: redundant computation or repeated I/O, independent operations run sequentially, blocking work added to startup or hot paths. Also flag long-lived objects built from closures or captured environments — they keep the entire enclosing scope alive for the object's lifetime (a memory leak when that scope holds large values); prefer a class/struct that copies only the fields it needs. Name the cheaper alternative.
+
+### Altitude
+
+Check that each change is implemented at the right depth, not as a fragile bandaid. Special cases layered on shared infrastructure are a sign the fix isn't deep enough — prefer generalizing the underlying mechanism over adding special cases.
+
+### Conventions (CLAUDE.md)
+
+Check the diff against the CLAUDE.md files read in Phase 0. Only flag a violation when you can quote the exact rule and the exact line that breaks it — no style preferences, no "spirit of the doc" inferences. Name the CLAUDE.md path and quote the rule so the report can cite it. If no CLAUDE.md applies, return nothing for this angle.
+
+Cleanup, altitude, and conventions candidates use the same `file`/`line`/`summary` shape; in `failure_scenario`, state the concrete cost (what is duplicated, wasted, harder to maintain, or which CLAUDE.md rule is broken) instead of a crash. Correctness bugs always outrank cleanup, altitude, and conventions findings when the output cap forces a cut.
+
+## Phase 2 — Dedup and verify
+
+Dedup candidates that point at the same line and mechanism, keeping the one with the most concrete failure scenario. Then verify each remaining candidate against the diff and the relevant files — one subagent verifier per candidate where a subagent tool is available, otherwise yourself, in this context. Each candidate gets exactly one of:
+
+- **CONFIRMED** — you can name the inputs/state that trigger it and the wrong output or crash. Quote the line.
+- **PLAUSIBLE** — mechanism is real, trigger is uncertain (timing, env, config). State what would confirm it.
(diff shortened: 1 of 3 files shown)


plugins/tend-ci-runner/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+
+Read the CLAUDE.md files that govern the changed code: the repo-root one, plus any `CLAUDE.md`/`CLAUDE.local.md` in a directory that is an ancestor of a changed file (a directory's CLAUDE.md applies only to files at or below it).
+
+## Phase 1 — Find candidates
+
+Work through the angles below. Each surfaces candidates with `file`, `line`, a one-line `summary`, and a concrete `failure_scenario`.
+
+Scale the fan-out to the change, the same way the caller scales its own depth:
+
+- **Peripheral or mechanical** (docs, config, dependency bumps, test-only): angles A–C plus the cleanup angles, in one pass, in this context. Up to 4 candidates each.
```
tend-agent: The peripheral set is `angles A–C plus the cleanup angles`, and the Reuse section defines "the cleanup angles" as itself plus the next three (Simplification, Efficiency, Altitude) — the closing paragraph reinforces the split by naming "cleanup, altitude, and conventions" as three categories. So the Conventions (CLAUDE.md) angle is selected only on core-logic changes, and the peripheral list is exactly `docs, config, dependency bumps, test-only` — the changes where a CLAUDE.md rule violation is the *most* likely defect and often the only one available. This PR is itself the case: it's a docs-shaped change governed by CLAUDE.md's "Authoring skills" rules, and a peripheral run of this skill over it would skip the angle that checks them.

```suggestion
- **Peripheral or mechanical** (docs, config, dependency bumps, test-only): angles A–C plus the cleanup and conventions angles, in one pass, in this context. Up to 4 candidates each.
```

plugins/tend-ci-runner/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+- **Core logic**: every angle, one subagent per angle where a subagent tool is available (`Task`/`Agent`), up to 6 candidates each. If no subagent tool is available, work through every angle yourself, in sequence, in one pass — do not skip angles for lack of fan-out.
```
tend-agent: The branch keys on the subagent tool being *available*, but availability isn't the binding constraint under the Claude harness. `Task` is in `allowed_tools` (`claude/action.yaml`, default `Bash,Edit,Read,Write,Glob,Grep,WebSearch,WebFetch,Task,Skill`), so the tool is there — while Claude Code's own headless system prompt in this run carries a line instructing the agent not to call the Agent tool unless the user requested it, and the workflow prompt is `/tend-ci-runner:review {number}`, which requests no such thing. So on a core change the condition reads as satisfied, the instruction reads as blocked, and which one wins is left to the agent. Naming the inline pass as the default resolves it in the direction that always works, and keeps the escape hatch for harnesses where fan-out is permitted. Phase 2's verifier sentence has the same `where a subagent tool is available` shape and wants the same treatment.

```suggestion
- **Core logic**: every angle, up to 6 candidates each. Work through them yourself, in sequence, in one pass — that's the default. Fan out one subagent per angle (`Task`/`Agent`) only where this session permits subagent use; the Claude harness allowlists the tool but its system prompt limits it to user-requested calls. Either way, don't skip angles for lack of fan-out.
```

plugins/tend-ci-runner/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Phase 3 — Sweep for gaps
+
+On a core-logic change, take one more pass as a fresh reviewer holding the verified list. Re-read the diff and the enclosing functions looking **only** for defects not already listed — the job is gaps, not re-confirmation. Focus on what a first pass tends to miss: moved or extracted code that dropped a guard or anchor; second-tier footguns (dataclass default evaluated once, `hash()` non-determinism, lock-scope shrink, predicate methods with side effects); setup/teardown asymmetry in tests; config defaults flipped. If nothing new, return nothing — don't pad.
+
+## Output
+
+Return the findings to the caller as a list of at most 10 (at most 4 for a peripheral change), ranked most-severe first, each with `file`, `line`, `summary`, `failure_scenario`, and its verdict. If nothing survives verification, say so in one line. Don't call the `ReportFindings` tool and don't publish an artifact — the caller owns the output.
+
+Say which mode ran (fanned-out angles with a subagent verify, or a single inline pass) so whoever reads the review isn't misled about what actually ran.
```
tend-agent: "whoever reads the review" points at the wrong audience. This skill returns findings to the `review` skill, and that skill's step 5 rules out exactly this kind of process narration from the posted body — its don't-post column is headed "internal analysis", and silence is the signal for anything that isn't actionable for the author. As written this asks for mode disclosure in the artifact the caller is told to keep free of it. The mode is worth reporting, just to the caller, whose weighting of the findings depends on it.

```suggestion
Tell the caller which mode ran (fanned-out angles with a subagent verify, or a single inline pass) so it can weigh the findings — context for the caller, not content for the review it posts.
```

tend-agent (commented):
Self-review — three notes on the ported skill, all in its selection/reporting edges rather than the angle text.

One is outside the diff, so I'm pushing it rather than suggesting it: `codex/agents-tail.md` enumerates the plugin's skills by name (`review`, `triage`, `ci-fix`, `nightly`, `weekly`, `notifications`, `review-runs`, `running-in-ci`, `review-reviewers`) and that list is exhaustive over the other nine — #820 brought it current two commits ago. `code-review` isn't in it, which undercuts the "runs on both harnesses" claim in the description: `codex plugin add` discovers the directory, but the AGENTS.md a Codex session actually reads still says these are the skills that exist. Adding it in a follow-up commit on this branch.

plugins/tend-ci-runner/skills/code-review/SKILL.md
tend-agent: Applied. Confirmed the reading: the Reuse section scopes "the cleanup angles" to itself plus the next three, so Conventions sat outside the peripheral set — and a docs-shaped change is where a CLAUDE.md rule is the likeliest defect. Peripheral now selects [angles A–C plus the cleanup and conventions angles](https://github.com/max-sixty/tend/blob/d4fa3d8ae5b109503ff71c3ab86272bea67d90be/plugins/tend-ci-runner/skills/code-review/SKILL.md#L27).

plugins/tend-ci-runner/skills/code-review/SKILL.md
tend-agent: Applied to both sentences. Verified the premise: `Task` is in the [`allowed_tools` default](https://github.com/max-sixty/tend/blob/d4fa3d8ae5b109503ff71c3ab86272bea67d90be/claude/action.yaml#L39), and the harness system prompt in this run does carry the don't-call-the-Agent-tool-unless-requested line — so availability and permission disagree, and the old wording keyed on the wrong one. Both the Phase 1 core bullet and the Phase 2 verifier now name the inline pass as the default and make fan-out the conditional.

plugins/tend-ci-runner/skills/code-review/SKILL.md
tend-agent: Applied. The `review` skill's step 5 does head its don't-post column ["internal analysis"](https://github.com/max-sixty/tend/blob/d4fa3d8ae5b109503ff71c3ab86272bea67d90be/plugins/tend-ci-runner/skills/review/SKILL.md#L187), so the old line asked for exactly what the caller is told to strip. Now reports the mode to the caller as weighting context rather than as review content.

PR #820, opened 2026-08-03
Remove claude-smoke and the tend-manual environment it required

max-sixty:
Removes the `claude-smoke` workflow and, once this merges, the `tend-manual` environment that existed only to feed it secrets.

The smoke was scaffolding for building the headless proxy harness: it ran 4 times ever, all June 17–19 while #704–#712 were under construction, and none of the ~15 harness changes since (sandbox PATH, uv isolation, sandbox levers, removing `claude-interactive`, the proxy readiness fix) dispatched it. Its deterministic half is covered per-PR by CI's `test-proxy` job at the production-pinned mitmproxy, and a harness broken in a release surfaces within a day through tend's own scheduled workflows dogfooding it, with a fix-forward release costing about an hour.

`tend-manual` was the reviewer-gated companion environment `claude-smoke` needed because its whole point — dispatching on feature branches — is what the `tend` ref policy exists to refuse. With its only consumer gone, the environment goes too, which also drops one paste from the credential migration in `TODO.md`. The `credential-environments` sweep is untouched: it verifies any reviewer-gated environment a repo declares, keyed on the credential rather than the name (its tests assert exactly that), and `docs/security-model.md` keeps documenting the pattern as the answer for bot-steerable triggers.

Also carried on this branch: two TODO.md corrections from the migration work — dropping the deleted cargo-affected `BOT_TOKEN` leftover, and replacing the wrong claim that a `main`-only policy can never admit the worktrunk/prql hand-maintained workflows' refs (probed live: `pull_request_target` runs at `refs/heads/main` and is admitted; a tag push is admitted once the policy carries a tag entry).


TODO.md
```diff
@@ -16,25 +16,40 @@ generated workflow names it. What remains:
...
+   so `environment: tend` is the whole change: worktrunk's benchmark gist
+   append and its two create-issue-on-failure jobs, prql's
+   `update-rust-toolchain` (all four `schedule`-only, and already `if`-gated
+   to it) and prql's backport. A `pull_request_target` run reports
+   `GITHUB_REF=refs/heads/main` and is admitted — which also means the
+   environment is not what gates that job.
+
+   worktrunk's release jobs run on a tag *push*, which is not bot-steerable,
+   and its all-tags ruleset is admin-gated, so a tag entry on the `tend`
+   policy gates them. Winget and homebrew move in too, and the repo-level copy
```
tend-agent: A tag entry on the `tend` policy is the one option here that `tend check` refuses. `check_environment` (in `generator/src/tend/checks.py`) reads *every* deployment-branch-policy entry without filtering on `type`, then requires exact set equality with the admitted branches — `extra = names - set(admitted)` fails with "admits v*, which tend does not verify the bot is kept off". `admitted` comes from `admitted_refs()` over the branch-protection results, so a tag name can never be in it. `fix_environment` is worse than an assertion failure: its second loop deletes any policy whose name isn't in `admitted`, so `tend check --fix` would silently remove the tag entry on the next run and leave the release jobs unable to reach the secret.

The admin-gated-tag reasoning is real, but it lives in `_policy_gate` — the path `check_credential_environments` takes for environments *other* than `tend` (it's the branch that checks `p.get("type") == "tag"` against `_tags_admin_gated`). So worktrunk's release jobs want either a separate environment carrying the tag entry, which `_policy_gate` already credits, or `check_environment` taught to accept `type: tag` entries under the admin-only all-tags ruleset with `fix_environment` stopped from deleting them. Worth naming which, since prql's third option ("moving the workflow to a tag push, which makes it worktrunk's case") inherits whichever answer this gets.

tend-agent (commented):
The removal itself checks out: `TEND_MANUAL_ENVIRONMENT` had no reader anywhere in the tree (only its definition), `.github/claude-smoke-prompt.md` had no other consumer, and the `credential-environments` sweep really is keyed on holding a credential rather than on a name, so `docs/security-model.md`'s rewrite matches `check_credential_environments`. Generator tests and pre-commit both pass on the merged tree.

I checked the other two repos' workflows against the rewritten TODO item, and the ref accounting holds: worktrunk's benchmark gist append and both `create-issue-on-*-failure` jobs are `if: … github.event_name == 'schedule'`, prql's `update-rust-toolchain` is gated on `nightly-upstream`, which is `github.event_name == 'schedule' && github.repository_owner == 'prql'`, and prql's `backport` is the `pull_request_target` case. The remaining `TEND_BOT_TOKEN` readers in prql are `release.yaml`'s brew-dispatcher, `publish-winget`, and `push-web-branch` — all covered by the third paragraph. One conflict with tend's own check, inline.

TODO.md
max-sixty: Confirmed against both functions and committed the TODO to the first resolution in 3a5621c: the tag entry lives on a separate release environment, the shape `check_credential_environments` already credits under the admin-gated all-tags ruleset, and the `tend` policy keeps its exact protected-branches shape. prql's third option inherits it.

tend-agent (approved):

TODO.md
tend-agent: That resolution holds against both functions: `check_environment` and `fix_environment` each address `TEND_ENVIRONMENT` by name, so a second environment sits outside the reconcile loop that would have deleted the entry, and `_policy_gate` never matches the *name* of a `type: tag` entry — it credits the whole entry on `tags_ok` alone, so the release tag pattern is fine as a pattern (the "a pattern entry is refused rather than matched" rule in that docstring is about branch entries).

Worth recording for the new environment while it's fresh: `_policy_gate` also drops a ref-gated policy when a workflow reaching it carries a bot-steerable trigger, and counts that trigger even when an `if:` would skip the event. worktrunk's `release.yaml` is `pull_request` plus a tag `push`, and `BOT_STEERABLE_TRIGGERS` is `{release, repository_dispatch}` (plus `workflow_dispatch` only when it declares inputs), so `publish-winget` and `publish-homebrew` moving into `release` clear that clause too — no `if:`-reading needed. It's exactly prql's `release`-triggered jobs that don't, which is the split the next paragraph already draws.

PR #821, opened 2026-08-03
fix(review): dedupe pre-APPROVE rollup by name and decide cap-expiry on provenance

tend-agent:
Two `tend-review` runs on `PRQL/prql` hit an identical pre-APPROVE rollup state 18 minutes apart this window and resolved it opposite ways: [#6130](https://github.com/PRQL/prql/pull/6130) APPROVED over the red, [#6133](https://github.com/PRQL/prql/pull/6133) withheld approval because of it. Both review bodies correctly diagnosed the red as a superseded cancellation-cascade artifact, and in both cases it did clear. The divergence traces to two defects in the pre-APPROVE guard added by #798.

## Defect 1 — `$FAILED` never clears

`statusCheckRollup` returns **every** check run on the commit, not the latest per name. When a concurrency-cancelled run is replaced, both entries persist, so the superseded `FAILURE` sits alongside the replacement's `SUCCESS` indefinitely:

```
$ gh -R PRQL/prql pr view 6133 --json statusCheckRollup \
    --jq '.statusCheckRollup[] | select(.name == "check-ok-to-merge") | "\(.conclusion) started=\(.startedAt)"'
FAILURE started=2026-08-03T17:17:01Z    # run 30835923137, own conclusion: cancelled
SUCCESS started=2026-08-03T18:26:46Z    # run 30835928436, the replacement
```

The current un-deduped `$FAILED` query still reports `check-ok-to-merge` on both PRs right now, long after each replacement omnibus went green. So the guidance's premise — "a stale cancellation-cascade `FAILURE` clears once the replacement run's omnibus goes green" — does not hold, and the `$FAILED` empty branch is unreachable on any PR that ever suffered a cancellation cascade.

Fix: reduce the rollup to the latest entry per check name before computing `$FAILED`/`$PENDING`. Verified against both PRs — `FAILED=[check-ok-to-merge]` before, `FAILED=[]` after.

A third occurrence ([PRQL/prql#6163](https://github.com/PRQL/prql/pull/6163), detailed in the thread below) lands this defect on the **settled** branch rather than cap-expiry, which is the worse of the two: that session polled to `PENDING == 0`, re-peeked, still saw `FAILED=[check-ok-to-merge]`, and so hit *"genuine terminal red. Skip the close-out"* — a confident wrong instruction to withhold approval on a clean mechanical bump, rather than merely leaving the model out of rules. The dedupe fixes that branch too; re-verified against #6163's live rollup.

## Defect 2 — no terminal branch when the poll cap expires

The `$FAILED` set + `$PENDING > 0` branch prescribes polling to settlement. On a matrix-heavy repo that settlement never arrives inside the 9-minute cap: #6130 touches `nightly.yaml`, so the full nightly release matrix runs on it, and the bot polled ~29 minutes without settling. With no rule for cap expiry, the model improvises — and improvised differently on the two PRs.

Replaying each rollup as of its review timestamp shows the inputs were identical:

```
#6130 at 17:46:10Z: FAILED=[check-ok-to-merge]
    -> traces to run 30835854409, own conclusion: cancelled
#6133 at 18:04:28Z: FAILED=[check-ok-to-merge]
    -> traces to run 30835923137, own conclusion: cancelled
```

Fix: add a cap-expiry branch that decides on **provenance** rather than settlement — resolve each remaining `FAILURE` to its run and read that run's own conclusion. All `cancelled` → superseded, approve and name the unverified checks. Any real `failure` → terminal-red branch. Both runs already performed this exact lookup by hand and reported it in their review bodies, so this codifies what the bot does anyway, and makes it produce one answer instead of two.

## Outcome cost

#6133 is a 4-line mechanical `docker/login-action` bump whose diff the bot called "correct and complete", now fully green, left without an approval on the strength of a red the same review called "stale, not a real failure — no action needed on it". That is precisely the outcome #798 was written to prevent.

## Gate assessment

- **Evidence level**: High — 3 occurrences in independent sessions (two with opposite resolutions on cap-expiry, one on the settled branch), on top of the parent shape's cumulative 2 occurrences that justified #798 (merged 2026-07-31). Threshold for High is 2–3.
- **Structural vs stochastic**: structural for both defects. Defect 1 is a deterministic query bug — the superseded entry never leaves the commit, so the same input always produces the same wrong `$FAILED`. Defect 2 is a fixed 9-minute cap against a matrix that exceeds it. The *choice* the model then makes is stochastic, which is exactly why the missing rule matters.
- **Change type**: targeted fix — one query correction plus one branch in an existing decision list.
- **Passes both gates**: yes.


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -223,29 +223,43 @@ ALREADY_POSTED=$(gh api "repos/$REPO/pulls/<number>/reviews" \
...
+- **`$FAILED` set and `$PENDING > 0`** — the rollup hasn't settled. Foreground-poll until non-own checks are terminal (the Step 6 / `running-in-ci` CI-monitoring loop), then re-peek `$FAILED`. Judge the settled state, not the mid-flight snapshot — a stale cancellation-cascade `FAILURE` drops out of `$FAILED` once the replacement omnibus registers, but *only* via the latest-per-name reduction above; the superseded check run itself never leaves the commit.
+- **`$FAILED` set and the poll cap expired with `$PENDING > 0`** — settlement is out of reach this session; a release or nightly matrix routinely outlasts the cap. Decide on **provenance**, not on settlement: resolve each remaining `FAILURE` to its run and read that run's own conclusion.
+
+  ```bash
+  for url in $(jq -r '.[] | select((.conclusion // .state) == "FAILURE")
+                     | .detailsUrl // .targetUrl // empty' <<<"$ROLLUP"); do
+    RUN=$(sed -nE 's#.*/actions/runs/([0-9]+).*#\1#p' <<<"$url")
+    [ -n "$RUN" ] && gh run view "$RUN" --json conclusion --jq '.conclusion'
+  done
```
tend-agent, line 284: The loop silently drops any `FAILURE` it can't resolve to an Actions run. `sed` yields nothing for a third-party status context — `pre-commit.ci - pr` points at `results.pre-commit.ci`, `codecov/patch` at `codecov.io` — so `[ -n "$RUN" ]` short-circuits and the check emits no line at all. It also prints an empty conclusion for a run still in progress (a failed job inside a matrix whose other legs are still going), which is the common shape at cap expiry given the branch requires `$PENDING > 0`.

Both feed the rule below, which reads "every one `cancelled` → APPROVE". Over a set where the only genuine red produced no output, that reads vacuously true and approves over it — the same `codecov/patch` class the closing paragraph names as the reason the pre-APPROVE peek exists. Emitting a marker keeps the non-`cancelled` default red:

```suggestion
  for url in $(jq -r '.[] | select((.conclusion // .state) == "FAILURE")
                     | .detailsUrl // .targetUrl // empty' <<<"$ROLLUP"); do
    RUN=$(sed -nE 's#.*/actions/runs/([0-9]+).*#\1#p' <<<"$url")
    if [ -n "$RUN" ]; then
      gh run view "$RUN" --json conclusion --jq '.conclusion'
    else
      echo "unresolved: $url"   # third-party status context, not an Actions run
    fi
  done
```

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -223,29 +223,43 @@ ALREADY_POSTED=$(gh api "repos/$REPO/pulls/<number>/reviews" \
...
+  ```
+
+  Every one `cancelled` — the red is superseded, so APPROVE and name the still-unverified checks in the body. Any one a real `failure` — take the terminal-red branch below. Don't leave this to improvisation: the same stale red must not draw an APPROVE on one PR and a withheld approval on the next.
```
tend-agent: Pairing with the loop change above: making `cancelled` the *only* value that earns an approval closes the vacuous-truth path, so an unresolvable URL or an empty conclusion lands on the red branch rather than falling through the gap between the two clauses.

```suggestion
  Every one `cancelled` — the red is superseded, so APPROVE and name the still-unverified checks in the body. `cancelled` is the only conclusion that earns an approval here: a real `failure`, an empty conclusion (the run is still going, so the job failed on its own merits), or an unresolvable URL (a third-party status context like `codecov/patch`, never an Actions run) all take the terminal-red branch below. Don't leave this to improvisation: the same stale red must not draw an APPROVE on one PR and a withheld approval on the next.
```

tend-agent (commented):
I verified the dedupe claim against PRQL#6133's live rollup — un-deduped `FAILED=check-ok-to-merge`, `FAILED=` after `group_by | max_by(.startedAt)` — and confirmed `startedAt` is populated on `StatusContext` nodes too (`pre-commit.ci - pr`), so the reduction is safe across both rollup node types. One gap in the new cap-expiry branch, inline.

Worth considering before this lands: `gh pr checks <n> --json name,state,link,workflow` already returns the **latest entry per check name** — gh does the reduction itself — with `link` standing in for `detailsUrl // targetUrl` and `workflow` for `workflowName`. On PRQL#6133 it returns exactly one `check-ok-to-merge` row, the green replacement. That would let `$ROLLUP` drop the `group_by`/`max_by` dance and the implicit assumption that `startedAt` is always present and always orders replacements after supersedees. I checked exit status on a PR with a pending check (0) but not on one with a terminal `FAILURE`, so that's the one thing to confirm before swapping.

```
$ ROLLUP=$(gh -R PRQL/prql pr view 6133 --json statusCheckRollup \
    --jq '[.statusCheckRollup[]] | group_by(.name // .context) | map(max_by(.startedAt))')
$ jq -r '[.[] | select((.conclusion // .state) == "FAILURE") | .name // .context] | join(", ")' <<<"$ROLLUP"
            # empty
$ gh -R PRQL/prql pr view 6133 --json statusCheckRollup \
    --jq '[.statusCheckRollup[] | select((.conclusion // .state) == "FAILURE") | .name // .context] | join(", ")'
check-ok-to-merge

$ gh -R PRQL/prql pr checks 6133 --json name,state,bucket,link,workflow \
    --jq '.[] | select(.name == "check-ok-to-merge" or .name == "pre-commit.ci - pr")'
{"bucket":"pass","link":".../runs/30835928436/job/91776226906","name":"check-ok-to-merge","state":"SUCCESS","workflow":"tests"}
{"bucket":"pass","link":"https://results.pre-commit.ci/run/github/449091292/...","name":"pre-commit.ci - pr","state":"SUCCESS","workflow":""}
```

`gh pr checks` is already latest-per-name (both PRQL PRs have three superseded `check-ok-to-merge` runs on the commit; only one row comes back).

</details>

Skipped the `/code-review` pass — `code-review` isn't in this session's skill list, which is the same observation #819 is built on.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Applied in 91e8463.

PRQL#5613 turns out to be a live instance of exactly this shape: its rollup carries a `codecov/patch` `FAILURE` pointing at `app.codecov.io`, so `sed` yields nothing and the old loop emitted no line for it. Running the two loop versions over it:

```
old loop:                      new loop:
failure                        failure
failure                        unresolved: https://app.codecov.io/gh/PRQL/prql/pull/5613
                               failure
```

Here the two genuine `failure` lines carry the decision anyway — but had those been `cancelled`, "every one `cancelled`" would have read true over an unemitted codecov red and drawn an APPROVE. That is the vacuous-truth path you described, and the marker closes it.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Applied in 91e8463. Making `cancelled` the sole approving conclusion is the right pairing — it turns the three non-`cancelled` outcomes (real `failure`, empty conclusion, unresolvable URL) from a gap between the two clauses into an explicit fall-through to the red branch.

tend-agent:
Both inline suggestions applied in 91e8463. On the `gh pr checks --json` swap: I chased the exit-status question and it comes back clean, but the swap itself doesn't — two things it would cost, and one thing worth taking from it, which I did take.

**Exit status is a non-issue.** `--json` exits 0 regardless of check state. `checksRun` returns at the exporter ([`checks.go:190`](https://github.com/cli/cli/blob/e83adbc0642994fae7c39a9a012eb34b8c81f4f1/pkg/cmd/pr/checks/checks.go#L190)), well before the `counts.Failed > 0 → SilentError` / `counts.Pending > 0 → PendingError` branches at [`checks.go:248`](https://github.com/cli/cli/blob/e83adbc0642994fae7c39a9a012eb34b8c81f4f1/pkg/cmd/pr/checks/checks.go#L248). Confirmed on PRQL#5613, which carries three terminal `FAILURE`s: bare `gh pr checks` exits 1, `gh pr checks --json name,state` exits 0.

**But it isn't a drop-in, for a reason that only shows up on the empty case.** When the commit has no checks at all, `populateStatusChecks` returns an error — `no checks reported on the '<branch>' branch` ([`checks.go:303`](https://github.com/cli/cli/blob/e83adbc0642994fae7c39a9a012eb34b8c81f4f1/pkg/cmd/pr/checks/checks.go#L303)) — and that fires *before* the exporter, so `--json` gives a non-zero exit and empty stdout where `statusCheckRollup` gives `[]`. `$ROLLUP` would be empty, the `jq` calls over it would break, and the branch that breaks is `$FAILED` empty → APPROVE: the normal path for a docs-only PR in a repo with path-filtered CI. Silent, and on the approving side.

**And the `startedAt` assumption isn't actually avoided by switching** — it moves inside gh rather than going away. `eliminateDuplicates` sorts by `StartedAt` descending and keeps the first per key ([`aggregate.go:96-97`](https://github.com/cli/cli/blob/e83adbc0642994fae7c39a9a012eb34b8c81f4f1/pkg/cmd/pr/checks/aggregate.go#L96-L97)) — the same "always present, always orders replacements after supersedees" premise `max_by(.startedAt)` rests on. So that half of the case for swapping doesn't survive reading the implementation.

**What was worth taking: gh's key is better than ours was.** It keys on `name/workflow/event` ([`aggregate.go:110`](https://github.com/cli/cli/blob/e83adbc0642994fae7c39a9a012eb34b8c81f4f1/pkg/cmd/pr/checks/aggregate.go#L110)), not name alone — so our `group_by(.name // .context)` was the *more* aggressive reduction, and two workflows registering the same check name would have collapsed into one entry, hiding a genuine red behind an unrelated green. Tightened to `group_by([.name // .context, .workflowName])`; `event` isn't exposed by `gh pr view --json statusCheckRollup`, so the key stops one field short of gh's. No divergence on the PRs I checked — row counts identical to `gh pr checks` on all three — so this is hardening against a shape I haven't observed, not a fix for one I have.

Dedupe keys vs. `gh pr checks`, surviving rows and resulting `$FAILED`:

```
                raw   name-only   name+workflow   gh pr checks
PRQL#6133        76      55            55             55        FAILED=[] for all three reductions
PRQL#6130       109      94            94             94        FAILED=[] for all three reductions
PRQL#5613        41      38            38             38        FAILED=[check-ok-to-merge, codecov/patch,
                                                                        test-rust (wasm32…) / test-rust]
```

`gh pr checks` exit status, gh 2.96.0:

```
$ gh pr checks -R PRQL/prql 5613 >/dev/null 2>&1;                     echo $?
1
$ gh pr checks -R PRQL/prql 5613 --json name,state >/dev/null 2>&1;   echo $?
0
```

Full recipe as it now stands, run end-to-end against both PRs with `GITHUB_RUN_ID`/`GITHUB_WORKFLOW` set:

```
PRQL#6133 → FAILED=[]  PENDING=0                      (provenance loop: no output, nothing red)
PRQL#5613 → FAILED=[check-ok-to-merge, codecov/patch, test-rust (wasm32…) / test-rust]  PENDING=0
            provenance loop → failure
                              unresolved: https://app.codecov.io/gh/PRQL/prql/pull/5613
                              failure
```

I could not find a PR with an empty rollup to test the `no checks reported` path against live data — that claim rests on the source at `checks.go:303`, not on a run.

gh source read at `cli/cli@e83adbc`.

</details>

tend-agent:
Third occurrence, and it lands on a branch the PR body doesn't name: the **settled** one, not cap-expiry.

[PRQL/prql#6163](https://github.com/PRQL/prql/pull/6163) (`codeql-action` 4.37.5 → 4.37.6, +3/−3), `tend-review` run [31201552732](https://github.com/PRQL/prql/actions/runs/31201552732). Dependabot opened it at 17:16:44Z; four `tests` runs fired within 6 s and the concurrency group cancelled three. The session's pre-APPROVE peek read `FAILED=[check-ok-to-merge] PENDING=55`, polled two full loops to `pending=0` / `ALL TERMINAL` (~19 min), re-peeked — and got `FAILED=[check-ok-to-merge]` again.

That is `$FAILED` set with `$PENDING == 0`, which the current text routes to *"genuine terminal red. Skip the close-out (`exit 0`)"*. So on a settled rollup the guidance doesn't merely make the `$FAILED`-empty branch unreachable — it affirmatively prescribes withholding approval on a clean mechanical bump. The session got the right answer only by ignoring the branch it landed on and running this PR's provenance lookup by hand (`actions/runs?head_sha=…` → cancelled run `31201555451` owns the `FAILURE` check run, surviving run `31201562501`'s `check-ok-to-merge` job is `success`), then approving.

This also cuts against the two counter-observations I'd recorded in the evidence log on 2026-08-05 — "the stale-cancellation-cascade branch clears correctly under load". It cleared there; it does not clear here, and the settled state is the worse failure of the two, because cap-expiry at least leaves the model visibly out of rules while this one hands it a confident wrong instruction.

The dedupe in this PR fixes it, re-verified against #6163 as it stands now:

```
$ ROLLUP=$(gh -R PRQL/prql pr view 6163 --json statusCheckRollup \
    --jq '[.statusCheckRollup[]] | group_by([.name // .context, .workflowName]) | map(max_by(.startedAt))')
$ jq -r '[.[] | select((.conclusion // .state) == "FAILURE") | .name // .context] | join(", ")' <<<"$ROLLUP"
                                    # deduped:   (empty)
$ gh -R PRQL/prql pr view 6163 --json statusCheckRollup \
    --jq '[.statusCheckRollup[] | select((.conclusion // .state) == "FAILURE") | .name // .context] | join(", ")'
check-ok-to-merge                   # undeduped: still red, ~30 min after the replacement went green
```

Both entries are still on the commit — `FAILURE` started 17:16:59Z (run `31201555451`, own conclusion `cancelled`), `SUCCESS` started 17:36:35Z (run `31201562501`).

<details><summary>Session-log excerpt — the settled re-peek</summary>

```
FAILED=[check-ok-to-merge] PENDING=55
iter 1 pending=51 … iter 9 pending=6
CI still running after 9 minutes
iter 1 pending=6 … iter 9 pending=0
ALL TERMINAL
FAILED=[check-ok-to-merge]
HEAD=a88efee16b1d23a3e962ca3047a54f182d6845aa STATE=OPEN
```

Artifact `claude-session-logs-n6163` on run 31201552732.

</details>

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -246,29 +246,47 @@ ALREADY_POSTED=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
...
 
-- **`$FAILED` set and `$PENDING > 0`** — the rollup hasn't settled. Foreground-poll until non-own checks are terminal (the Step 6 / `running-in-ci` CI-monitoring loop), then re-peek `$FAILED`. Judge the settled state, not the mid-flight snapshot — a stale cancellation-cascade `FAILURE` clears once the replacement run's omnibus goes green.
+- **`$FAILED` set and `$PENDING > 0`** — the rollup hasn't settled. Foreground-poll until non-own checks are terminal (the Step 6 / `running-in-ci` CI-monitoring loop), then re-peek `$FAILED`. Judge the settled state, not the mid-flight snapshot — a stale cancellation-cascade `FAILURE` drops out of `$FAILED` once the replacement omnibus registers, but *only* via the reduction above; the superseded check run itself never leaves the commit.
```
tend-agent: `$ROLLUP` is a shell variable, and the poll it points at is a separate Bash invocation — so at the re-peek `$ROLLUP` is empty, `jq … <<<""` returns nothing with exit 0, and `$FAILED` reads clean. Say to re-run the block, not to re-read the variable.

```suggestion
- **`$FAILED` set and `$PENDING > 0`** — the rollup hasn't settled. Foreground-poll until non-own checks are terminal (the Step 6 / `running-in-ci` CI-monitoring loop), then **re-run the whole `ROLLUP=`/`FAILED=`/`PENDING=` block above** — the poll is its own Bash call and `$ROLLUP` does not survive it; re-reading `$FAILED` off a stale (empty) `$ROLLUP` yields no findings and falls through to APPROVE. Judge the settled state, not the mid-flight snapshot — a stale cancellation-cascade `FAILURE` drops out of `$FAILED` once the replacement omnibus registers, but *only* via the reduction above; the superseded check run itself never leaves the commit.
```

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -246,29 +246,47 @@ ALREADY_POSTED=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
...
+- **`$FAILED` set and the poll cap expired with `$PENDING > 0`** — settlement is out of reach this session; a release or nightly matrix routinely outlasts the cap. Decide on **provenance**, not on settlement: resolve each remaining `FAILURE` to its run and read that run's own conclusion.
```
tend-agent: Same shell-state problem, worse consequence: the loop below reads `$ROLLUP`, and by the time this branch is reached the poll has run in a separate Bash call, so `$ROLLUP` is empty and the loop executes zero iterations. "Every one `cancelled`" over an empty set reads vacuously true and approves over the red — the shape the unresolvable-URL marker closed in 91e8463, reached through a different door.

```suggestion
- **`$FAILED` set and the poll cap expired with `$PENDING > 0`** — settlement is out of reach this session; a release or nightly matrix routinely outlasts the cap. Re-derive `$ROLLUP` first (the poll was its own Bash call; an empty `$ROLLUP` makes the loop below iterate zero times, and "every one `cancelled`" then reads vacuously true). Decide on **provenance**, not on settlement: resolve each remaining `FAILURE` to its run and read that run's own conclusion.
```

tend-agent (commented):
One new finding on the incremental, inline: hoisting the rollup into `$ROLLUP` makes both later readers depend on shell state that cannot survive the poll.

Before this PR, `$FAILED` and `$PENDING` were each a self-contained `gh pr view` command, so "re-peek `$FAILED`" meant re-running one command in whatever Bash call came after the poll. Now they are `jq … <<<"$ROLLUP"`, and the poll is necessarily a separate Bash invocation (`running-in-ci` prescribes it as its own call with `timeout: 600000`). Step 1 of this same skill already states the constraint — "Shell state doesn't persist between tool calls" — but this section is the one place a variable now has to cross one.

Both failures land on the approving side, silently:

```
$ ROLLUP=""     # what the post-poll Bash call actually sees
$ FAILED=$(jq -r '[.[] | select((.conclusion // .state) == "FAILURE")
                  | .name // .context // "unknown"] | join(", ")' <<<"$ROLLUP")
$ echo "exit=$? FAILED=[$FAILED]"
exit=0 FAILED=[]                 # -> "$FAILED empty -> proceed with APPROVE"

$ for url in $(jq -r '.[] | select((.conclusion // .state) == "FAILURE")
                    | .detailsUrl // .targetUrl // empty' <<<"$ROLLUP"); do echo "iter: $url"; done
                                 # zero iterations -> "Every one cancelled" reads vacuously true -> APPROVE
```

jq exits 0 on empty input, so neither reads as an error. The second one is the same vacuous-truth shape the unresolvable-URL marker closed one commit ago, reachable through a different door — there the set was missing a member, here the set is empty.

(`$PENDING` also comes back as `""` rather than `0`, so `[ "$PENDING" -gt 0 ]` errors with `integer expression expected` and evaluates false — but `$FAILED` is read first, so the approve branch is reached before that matters.)

Both suggestions just say to re-derive `$ROLLUP`; the alternative is to fold the three assignments into one re-runnable block and tell the reader to run the block, not the line.

The `/tend-ci-runner:code-review` pass ran this time (single inline pass, short angle set) — it surfaced this and nothing else that survived verification.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Applied in e9bfd66, taking the alternative you named rather than the suggestion verbatim: the constraint now sits once under the `ROLLUP=`/`FAILED=`/`PENDING=` block ("re-run that whole block — not just the `FAILED=` line — after any poll", with both silent-exit-0 failure modes named), and this branch just says to re-run the block and read `$FAILED` off the fresh `$ROLLUP`. Stating it at the block covers every future reader of `$ROLLUP`, not only the two branches that read it today.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Applied in e9bfd66. The vacuous-truth half is called out at the block ("the provenance loop iterates zero times, making its 'every one `cancelled`' test vacuously true"), and this branch now says to re-run the rollup block first, since the loop below reads `$ROLLUP` and the expired poll was its own Bash call.

Issue #822, opened 2026-08-04
tend check: configuration drift on max-sixty/tend


tend-agent:
`tend check` currently reports one failure on this repo, down from three.

- **`credential-environments`** — the `release` environment has no required reviewers and admits tags, and "no active all-tags ruleset restricting creation and update to admins could be verified". This is a **false positive**, tracked as #824 — see below.

Previously resolved: `claude-auth` and `repo-secret-allowlist` both PASS. `CLAUDE_CODE_OAUTH_TOKEN` now resolves from the `tend` environment and no longer sits at repo level, which is item 1 of ["Finish moving the operational secrets into the `tend` environment"](https://github.com/max-sixty/tend/blob/4672f80/TODO.md#L13-L19) done for this repo. No bot action was possible on those two — the token can't be read back — so a maintainer moved them.

## `credential-environments` — the #824 false positive, firing on tend itself

Not a config regression. The repo does have the ruleset the check is asking for — `Tag operations` (`16490366`), active, `~ALL` tags, both `creation` and `update` restricted, and no `bypass_actors` key at all, which is stricter than the admin-only arrangement the remediation text requests:

```console
$ gh api repos/max-sixty/tend/rulesets/16490366 \
    --jq '{name, target, enforcement, current_user_can_bypass, conditions, rules: [.rules[].type], bypass_actors}'
{"bypass_actors":null,"conditions":{"ref_name":{"exclude":[],"include":["~ALL"]}},"current_user_can_bypass":"never","enforcement":"active","name":"Tag operations","rules":["creation","update"],"target":"tag"}
```

`_bypass_actors_above_bot` returns `None` for an absent list (["Returns … None when the list is withheld"](https://github.com/max-sixty/tend/blob/3b46066/generator/src/tend/checks.py#L244-L273)), which propagates to "could not be verified" and then to FAIL. Queried with the bot's own token the same ruleset reports `"current_user_can_bypass": "never"`, so the escalation path the check describes is in fact closed.

## Status: fixed on `main`, not yet released

#825 merged on 2026-08-09 (`82479bc`) and does exactly what this issue asks for: `_policy_gate` now returns a `_Gap(..., verified=False)` when `tags_ok()` is `None`, and `check_credential_environments` routes those to a SKIP ("Could not read the whole credential surface") instead of a FAIL. #826 was closed in favour of it.

[...]

PR #825, opened 2026-08-04
check: report an unreadable ruleset bypass list as unknown, not ungated

tend-agent:
## Problem

`check_credential_environments` (added in #815, unreleased) fails on this repo, and its verdict depends on the token's permission level rather than on the repo's configuration:

```
FAIL  credential-environments — A run the bot can cause reaches a credential: 'release' has no required reviewers, and admits tags, and no active all-tags ruleset restricting creation and update to admins could be verified.
```

The `release` environment (PyPI trusted publishing, `pypi-release.yaml`) admits tags, and this repo *does* carry the all-tags ruleset the check looks for — `Tag operations`: target `tag`, active, `include: ["~ALL"]`, rules `creation` + `update`. The gate is there. What the bot cannot see is its bypass list: GitHub omits `bypass_actors` from the ruleset response below repo admin, so `_bypass_actors_above_bot(None, …)` returns None, `_tags_admin_gated` returns None, and `_policy_gate` — which credited tag entries only on `is True` — reported the environment ungated.

Verified against the API rather than inferred: reading a ruleset with no bypass actors as **admin** returns `"bypass_actors": []`; reading the same field on this repo with the bot's **write**-scoped token omits the key entirely (all three of this repo's rulesets, including the branch one).

That breaks the invariant this module's own docstring states — *"Everything read here is readable with the bot's own write-scoped token, so the nightly run sees the same answers a maintainer does"* — and it breaks it on the release-environment shape [`TODO.md`](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/TODO.md#L20-L38) prescribes. Once this ships, every adopter whose nightly runs `tend check` under the bot's token gets a permanent failure on a correctly-gated environment, and `--fix` cannot clear it; a maintainer running the same command as admin sees it pass.

## Solution

Split "confirmed absent" from "could not see". `_policy_gate` now returns a `_Gap(reason, verified)`, and `check_credential_environments` routes an unverified gap into the existing unknown result (`Could not read the whole credential surface: …`) instead of the failure list. A tag entry with no gating ruleset still fails, unchanged. The one other permission-shaped reason — a deployment branch policy the token cannot list — moves with it, same cause.

This is the stance `check_branch_protection` already takes on the same unreadable field: *"Ruleset check was inconclusive — don't false-positive"* → unknown, not failure.

After the change, on this repo:

```
SKIP  credential-environments — Could not read the whole credential surface: 'release' has no required reviewers, and admits tags, and whether an all-tags ruleset gates them is unverifiable with this token — either the rulesets aren't readable, a bypass list is withheld (only a repo admin reads one), or a bypass actor names a principal tend cannot resolve: a team, app, or deploy key, or any user if `bot_name` itself does not resolve to an account
```

The two genuine failures this repo has — `claude-auth` and `repo-secret-allowlist`, both #822 — still fail, so the drift report is unaffected.

## Testing

`test_credential_environments_withheld_bypass_list_is_unknown` reproduces it: the admin-shaped tag ruleset with `bypass_actors` removed, asserting unknown. It fails before the change (`assert False is None`) and passes after. Three more tests pin the rest of the split — a confirmed bad ref outranking a withheld list (both entry orders), the unlistable-policy path, and a steerable trigger outranking a held unverified tag entry — each verified to fail on the mutation that removes what it pins. Full generator suite green (331 passed), pinned `ruff` clean, and `tend check` re-run live against this repo for the output above.

---

Found while triaging #822 — it does not fix that issue, which needs a maintainer to move the token.

### DIFF
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -917,14 +917,17 @@ def _policy_gate(
         if p.get("type") == "tag":
             gated = tags_ok()
             if gated is None:
-                # GitHub serves `bypass_actors` only to a repo admin, so the
-                # bot's own token reads a gating all-tags ruleset as
-                # unverifiable — the same configuration a maintainer's admin
-                # token confirms.
+                # Every unread input lands here, not just a withheld bypass
+                # list: an unlistable `/rulesets`, a ruleset that won't fetch,
+                # and a bypass actor naming a team, app, or deploy key are all
+                # None. So the message names the set rather than prescribing
+                # the admin re-run that settles only one of them.
                 unverified = _Gap(
-                    "admits tags under a ruleset this token cannot verify: only "
-                    "a repo admin reads a ruleset's bypass list, so re-run "
-                    "`tend check` as an admin to settle it",
+                    "admits tags, and whether an all-tags ruleset gates them is "
+                    "unverifiable with this token — either the rulesets aren't "
+                    "readable, a bypass list is withheld (only a repo admin "
+                    "reads one), or a bypass actor names a team, app, or deploy "
+                    "key whose membership isn't resolvable here",
                     verified=False,
                 )
             elif gated is False:
@@ -939,10 +942,14 @@ def _policy_gate(
             )
     if steerable:
         triggers = ", ".join(f"`{t}`" for t in sorted(steerable))
+        # Not "admits only verified refs": a held `unverified` means one entry
+        # didn't settle. The ref list is beside the point here anyway — the bot
+        # picks the ref — so the message states the trigger, which holds either
+        # way, and this stays a verified finding.
         return _Gap(
-            f"admits only verified refs, but a workflow reaching it runs on "
-            f"{triggers}, which the bot fires and steers against a ref the "
-            "policy already admits"
+            f"is reached by a workflow running on {triggers}, which the bot "
+            "fires and steers against a ref the policy already admits, so its "
+            "ref list does not gate it"
         )
     return unverified
 
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -908,6 +908,11 @@ def _policy_gate(
         return _Gap(
             "has a deployment branch policy this token cannot list", verified=False
         )
+    # An unverifiable entry is held, not returned: a later entry can name a ref
+    # confirmed out of the verified set, and that finding outranks this one —
+    # the precedence `check_credential_environments` already applies when both
+    # kinds arrive from different environments.
+    unverified: _Gap | None = None
     for p in policies:
         if p.get("type") == "tag":
             gated = tags_ok()
@@ -916,13 +921,13 @@ def _policy_gate(
                 # bot's own token reads a gating all-tags ruleset as
                 # unverifiable — the same configuration a maintainer's admin
                 # token confirms.
-                return _Gap(
+                unverified = _Gap(
                     "admits tags under a ruleset this token cannot verify: only "
                     "a repo admin reads a ruleset's bypass list, so re-run "
                     "`tend check` as an admin to settle it",
                     verified=False,
                 )
-            if gated is False:
+            elif gated is False:
                 return _Gap(
                     "admits tags, and no active all-tags ruleset restricting "
                     "creation and update to admins could be verified"
@@ -939,7 +944,7 @@ def _policy_gate(
             f"{triggers}, which the bot fires and steers against a ref the "
             "policy already admits"
         )
-    return None
+    return unverified
 
 
 def check_credential_environments(
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -242,7 +242,9 @@ def _bypass_actors_above_bot(actors: list[dict] | None, bot_name: str) -> bool |
     Returns False if one of them is the bot itself or a role at write or
     below, None when the list is withheld (only ruleset admins see
     `bypass_actors`) or names a principal this can't resolve (a team, app,
-    or deploy key). An empty list is True — nobody bypasses at all.
+    or deploy key, or any user once `bot_name` itself fails to resolve — the
+    ids have nothing to compare against). An empty list is True — nobody
+    bypasses at all.
     """
     if actors is None:
         return None
@@ -920,9 +922,10 @@ def _policy_gate(
             if gated is None:
                 # Every unread input lands here, not just a withheld bypass
                 # list: an unlistable `/rulesets`, a ruleset that won't fetch,
-                # and a bypass actor naming a team, app, or deploy key are all
-                # None. So the message names the set rather than prescribing
-                # the admin re-run that settles only one of them.
+                # a bypass actor naming a team, app, or deploy key, and a
+                # `User` actor left undecidable by an unresolvable `bot_name`
+                # are all None. So the message names the set rather than
+                # prescribing the admin re-run that settles only one of them.
                 unverified = _Gap(
                     "admits tags, and whether an all-tags ruleset gates them is "
                     "unverifiable with this token — either the rulesets aren't "
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -208,8 +208,9 @@ def check_branch_protection(repo: str, branch: str, bot_name: str) -> CheckResul
             None,
             f"Branch '{branch}' is protected but could not verify that the bot "
             "cannot bypass its rulesets — either they aren't readable with this "
-            "token, or a bypass actor names a team, app, or deploy key whose "
-            "membership isn't resolvable here. Check the bypass list manually.",
+            "token, or a bypass actor names a principal tend cannot resolve: a "
+            "team, app, or deploy key, or any user if `bot_name` itself does not "
+            "resolve to an account. Check the bypass list manually.",
         )
 
     return CheckResult(
@@ -926,8 +927,9 @@ def _policy_gate(
                     "admits tags, and whether an all-tags ruleset gates them is "
                     "unverifiable with this token — either the rulesets aren't "
                     "readable, a bypass list is withheld (only a repo admin "
-                    "reads one), or a bypass actor names a team, app, or deploy "
-                    "key whose membership isn't resolvable here",
+                    "reads one), or a bypass actor names a principal tend "
+                    "cannot resolve: a team, app, or deploy key, or any user "
+                    "if `bot_name` itself does not resolve to an account",
                     verified=False,
                 )
             elif gated is False:
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -856,6 +856,21 @@ def _reviewer_gate(env: dict, bot_name: str) -> str | None:
     return None
 
 
+@dataclass(frozen=True)
+class _Gap:
+    """Why a gate does not hold, and whether that verdict was verified.
+
+    `verified` is False when the token could not see enough to decide, which
+    is not the same finding as a gate confirmed absent: the module docstring's
+    invariant is that the nightly sees the answers a maintainer does, so where
+    it doesn't, the honest report is unknown. `check_branch_protection` takes
+    the same stance on an unreadable bypass list.
+    """
+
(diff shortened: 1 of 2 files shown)


generator/src/tend/checks.py
```diff
@@ -882,35 +897,54 @@ def _policy_gate(
...
+            gated = tags_ok()
+            if gated is None:
+                # GitHub serves `bypass_actors` only to a repo admin, so the
+                # bot's own token reads a gating all-tags ruleset as
+                # unverifiable — the same configuration a maintainer's admin
+                # token confirms.
+                unverified = _Gap(
+                    "admits tags under a ruleset this token cannot verify: only "
+                    "a repo admin reads a ruleset's bypass list, so re-run "
+                    "`tend check` as an admin to settle it",
```
tend-agent: `_tags_admin_gated` returns None for more than a withheld list: a candidate all-tags ruleset whose bypass list names a Team, app, or deploy key falls through `_bypass_actors_above_bot`'s `actor_type not in BYPASS_ACTOR_TYPES_ABOVE_BOT` branch to None, and so does an unreadable `/rulesets` listing. On those the message points at the wrong fix — the list was read, and re-running as admin settles nothing.

```suggestion
                    "admits tags under a ruleset this token cannot verify: its "
                    "bypass list is either withheld (only a repo admin reads "
                    "one) or names a principal tend cannot resolve, so re-run "
                    "`tend check` as an admin, or name individual bypass actors",
```

generator/tests/test_checks.py
```diff
@@ -1611,6 +1611,39 @@ def test_credential_environments_rejects_tags_without_a_ruleset() -> None:
...
+    withheld = {k: v for k, v in _ADMIN_TAG_RULESET.items() if k != "bypass_actors"}
+    for order in ("tag v*\nbranch staging", "branch staging\ntag v*"):
+        fake = _credential_env_gh(
+            {"release": (["PYPI_TOKEN"], _CUSTOM_POLICY, order)},
+            tag_rulesets={"7": withheld},
+        )
+        with patch("tend.checks._gh", side_effect=fake):
+            result = check_credential_environments("owner/repo", _config(), ["main"])
+        assert result.passed is False, f"{order}: {result.message}"
+        assert "staging" in result.message
```
tend-agent, line 1644: Pins the `_branch_policies` → None path, which currently passes either way.

```suggestion
        assert "staging" in result.message


def test_credential_environments_unlistable_branch_policy_is_unknown() -> None:
    """Listing the policy entries can fail on permission alone, and an unread
    list is not a gate confirmed absent — the same reason a withheld bypass
    list reports unknown."""
    inner = _credential_env_gh(
        {"release": (["PYPI_TOKEN"], _CUSTOM_POLICY, "branch main")}
    )

    def fake(*args, **kwargs):
        url = args[-3] if "--jq" in args else args[-1]
        if url.endswith("/deployment-branch-policies"):
            return _make_completed(returncode=1)
        return inner(*args, **kwargs)

    with patch("tend.checks._gh", side_effect=fake):
        result = check_credential_environments("owner/repo", _config(), ["main"])
    assert result.passed is None, result.message
    assert "cannot list" in result.message
```

tend-agent (commented):
The unknown-vs-failure split reads right, and the two new tests pin the halves that matter (withheld list → unknown; a settled bad ref outranking it, both orders). Two things below.

The other reason that moved to `verified=False` — the deployment branch policy the token cannot list — has no test. Dropping `verified=False` from that return leaves all 329 tests green, so nothing pins the half of the change the PR body calls out as moving for the same cause. Inline suggestion adds one (verified: it fails on that mutation, passes as-is, `ruff format` clean).

Separately, holding the unverified gap instead of returning it makes the `steerable` return reachable while a tag entry was unverifiable, and its message opens `admits only verified refs` — a claim about refs this token could not verify. The verdict is still a genuine failure there (the bot fires the trigger against an already-admitted ref either way), so this is wording, not logic — but that clause now asserts something the path didn't establish.

generator/src/tend/checks.py
tend-agent: Right — `_tags_admin_gated` reaches None four ways, and only one of them is settled by re-running as admin: an unlistable `/rulesets` ([checks.py#L308-L309](https://github.com/max-sixty/tend/blob/932922c96a53211e57807c1256fec47d37b316ec/generator/src/tend/checks.py#L308-L309)), a ruleset that won't fetch or parse ([#L313-L321](https://github.com/max-sixty/tend/blob/932922c96a53211e57807c1256fec47d37b316ec/generator/src/tend/checks.py#L313-L321)), a withheld bypass list, and an actor type outside `BYPASS_ACTOR_TYPES_ABOVE_BOT` ([#L266-L267](https://github.com/max-sixty/tend/blob/932922c96a53211e57807c1256fec47d37b316ec/generator/src/tend/checks.py#L266-L267)).

Took the substance rather than the literal wording, so it names the readability case too and matches the sentence `check_branch_protection` already uses on the same field ("either they aren't readable with this token, or a bypass actor names a team, app, or deploy key whose membership isn't resolvable here"):

```
'release' has no required reviewers, and admits tags, and whether an all-tags ruleset gates
them is unverifiable with this token — either the rulesets aren't readable, a bypass list is
withheld (only a repo admin reads one), or a bypass actor names a team, app, or deploy key
whose membership isn't resolvable here
```

The prescription is gone with it: "re-run as an admin" is wrong for two of the four, so the message names the causes and leaves the fix to whichever one the reader recognizes. Updated the PR body's sample output to match.

generator/tests/test_checks.py
tend-agent: Added, near-verbatim — the mutation check holds: dropping `verified=False` from that return fails it with `assert False is None`, and it passes as-is. `ruff check`/`ruff format` clean.

While in there I added the one other untested path this PR introduces: a steerable trigger arriving alongside a *held* unverified tag entry. That precedence is only asserted by the code today — returning `unverified` ahead of the `steerable` gap leaves the whole suite green — so [`test_credential_environments_steerable_trigger_outranks_an_unverified_tag`](https://github.com/max-sixty/tend/blob/932922c96a53211e57807c1256fec47d37b316ec/generator/tests/test_checks.py#L1835-L1851) pins it (verified failing on that mutation). 331 passing.

tend-agent:
On the third point from the review body — the `steerable` return opening `admits only verified refs` while a tag entry went unverified — agreed, and fixed in [932922c](https://github.com/max-sixty/tend/commit/932922c96a53211e57807c1256fec47d37b316ec). Rather than branch the wording on whether `unverified` is held, the reason now drops the ref claim entirely:

```
'pypi' has no required reviewers, and is reached by a workflow running on `release`, which the
bot fires and steers against a ref the policy already admits, so its ref list does not gate it
```

The ref list is beside the point on that path — the bot picks the ref, so what the policy admits doesn't decide the verdict — which is why the finding stays verified even with an unsettled entry held. The two inline threads carry the other two fixes.

generator/src/tend/checks.py
```diff
@@ -882,35 +897,61 @@ def _policy_gate(
...
+                # list: an unlistable `/rulesets`, a ruleset that won't fetch,
+                # and a bypass actor naming a team, app, or deploy key are all
+                # None. So the message names the set rather than prescribing
+                # the admin re-run that settles only one of them.
+                unverified = _Gap(
+                    "admits tags, and whether an all-tags ruleset gates them is "
+                    "unverifiable with this token — either the rulesets aren't "
+                    "readable, a bypass list is withheld (only a repo admin "
+                    "reads one), or a bypass actor names a team, app, or deploy "
+                    "key whose membership isn't resolvable here",
```
tend-agent: `_bypass_actors_above_bot` also returns `None` when a `User` bypass actor can't be compared — `_user_id(bot_name)` failing leaves `bot_id is None`, which sets `unresolved`. Adding it keeps the set closed; verified with `pytest` (331 pass) and under the 88-char line cap.

```suggestion
                    "reads one), or a bypass actor doesn't resolve here: a team, "
                    "app, or deploy key, or a `User` entry left undecidable "
                    "because the configured bot_name itself doesn't resolve",
```

tend-agent (commented):
One gap left in the enumeration the commit set out to complete. `_bypass_actors_above_bot` reaches `None` a fifth way: a `User` bypass actor whose id can't be compared because `_user_id(bot_name)` returned `None` (the `if bot_id is None: unresolved = True` branch). The likeliest cause isn't a permission wall — it's a `bot_name` in `.config/tend.yaml` that doesn't resolve, so `gh api users/<name>` 404s. On that input the reader gets a message naming three causes, none of which is theirs, and nothing else in `tend check` says the login is wrong either: `check_bot_permission`'s 404 path reports "Could not check (may require admin access to read)".

The same sentence in `check_branch_protection` — the `Branch '<b>' is protected but could not verify that the bot cannot bypass its rulesets` return — has the identical gap, and it's outside this diff. Happy to push a commit covering both if you want them to stay matched.

tend-agent:
Confirmed and fixed in [ce7d23e](https://github.com/max-sixty/tend/commit/ce7d23e). `_user_id` returns `None` on any non-zero exit, so a `bot_name` that 404s takes every `User` actor down the `unresolved` branch — the helper's behaviour is already pinned by `test_update_rule_user_bypass_unresolvable_login`; only the message was missing the case. Both sentences now read:

```
... or a bypass actor names a principal tend cannot resolve: a team, app, or deploy key, or
any user if `bot_name` itself does not resolve to an account
```

Took `check_branch_protection`'s copy in the same commit rather than a follow-up: it's the identical sentence over the identical helper return, and the reason the wording was aligned in the first place was so the two stay matched — splitting them would land this PR having just re-diverged them.

Left alone: `check_bot_permission` reporting an unresolvable login as "may require admin access to read". That's a different check reaching a different conclusion from the same 404, and fixing it means distinguishing 404 from 403 at the `_gh` layer rather than editing a string — out of scope here.

generator/src/tend/checks.py
```diff
@@ -882,35 +898,62 @@ def _policy_gate(
...
     for p in policies:
         if p.get("type") == "tag":
-            if tags_ok() is not True:
-                return (
+            gated = tags_ok()
+            if gated is None:
+                # Every unread input lands here, not just a withheld bypass
+                # list: an unlistable `/rulesets`, a ruleset that won't fetch,
+                # and a bypass actor naming a team, app, or deploy key are all
+                # None. So the message names the set rather than prescribing
```
tend-agent: The message below now names five inputs; this comment still names four, so it reads as the authoritative list and undercounts.

```suggestion
                # a bypass actor naming a team, app, or deploy key, and a `User`
                # actor left undecidable by an unresolvable `bot_name` are all
                # None. So the message names the set rather than prescribing
```

tend-agent (commented):
`ce7d23e` names the fifth way in both user-facing messages, but two internal enumerations of the same set still list only four — and one of them is the comment that exists to explain the message directly below it, so the two now contradict each other. Suggestion inline.

The other is `_bypass_actors_above_bot`'s own docstring (outside the diff), which is the canonical statement of when that function returns `None`:

> ...None when the list is withheld (only ruleset admins see `bypass_actors`) or names a principal this can't resolve (a team, app, or deploy key).

Same gap — the unresolvable-`bot_name` case is a `User` actor, which that parenthetical excludes. Happy to push both as one commit if hand-applying isn't convenient.

generator/src/tend/checks.py
tend-agent: Applied in [bfe70b6](https://github.com/max-sixty/tend/commit/bfe70b6), along with `_bypass_actors_above_bot`'s docstring from the review body — its parenthetical now reads "a team, app, or deploy key, or any user once `bot_name` itself fails to resolve — the ids have nothing to compare against", so the helper's own statement of its `None` covers the case its callers' messages do.

That's three enumerations of the same set (helper docstring, this comment, the two user-facing strings) kept in sync by hand, which is the shape that just drifted twice in one PR. Not restructuring it here — the honest fix is for the `None` to carry its cause rather than be re-described at each site, and that's a change to the helper's return type across every caller, not to this diff.

PR #826, opened 2026-08-04
check: settle a withheld ruleset bypass list with current_user_can_bypass

### LINKED ISSUE #824: secret-environments: false positive when an all-tags ruleset has no bypass actors
`tend check`'s `secret-environments` check reports a FAIL on max-sixty/worktrunk that appears to be a false positive: it says no qualifying all-tags ruleset "could be verified", but an active one has existed on the repo since 2026-05-16 and it blocks exactly the escalation path the check describes.

The reported line (from max-sixty/worktrunk#3729, filed by the nightly sweep):

> **`secret-environments`** — the `release` and `signing` environments each hold secrets, have no required reviewers, and admit tags, and no active all-tags ruleset restricting tag creation/update to admins could be verified. A run the bot can cause could therefore reach those secrets.

## The ruleset the check misses

`GET /repos/max-sixty/worktrunk/rulesets/16490367`:

```json
{
  "name": "Tag operations",
  "target": "tag",
  "enforcement": "active",
  "conditions": { "ref_name": { "include": ["~ALL"], "exclude": [] } },
  "rules": [{ "type": "creation" }, { "type": "update" }],
  "created_at": "2026-05-16T22:31:36.015Z",
  "current_user_can_bypass": "never"
}
```

Active, `~ALL` tags, both `creation` and `update` restricted. The detail response carries **no `bypass_actors` key at all** — so there are no bypass actors configured, which is stricter than the admin-only arrangement the check's remediation text asks for.

## Why this matters for the check's own threat model

The check's stated concern is that "a run the bot can cause could reach those secrets" — i.e. the bot creates a `v*` tag, which triggers `release.yaml`, whose `publish-cargo` and `publish-aur` jobs carry `environment: release`. Queried with the **bot's own token**, that same ruleset reports `"current_user_can_bypass": "never"`. The bot cannot create or update any tag, so it cannot reach the release/signing secrets by that path. The mitigation is in place and effective against precisely the actor the check is worried about.

## Possible causes

I can't tell from the consumer side which of these it is, but a permissions gap looks unlikely — I read the ruleset above with the bot token, so the check would have been able to as well:

- The check may enumerate rulesets without following up with a per-ruleset `GET` (the list response at `/rulesets` returns only `{id, name, target, enforcement, ...}` — `conditions` and `rules` are only present on the detail endpoint, so a list-only check can't see `~ALL` / `creation` / `update`).
- It may require `bypass_actors` to be present and to name admins, and treat an absent `bypass_actors` (no bypass at all) as unverifiable rather than as the strictest case.

If the second is the cause, the fix is to treat "no bypass actors" as satisfying the requirement — it's a superset of admin-only.

## Impact

The FAIL is durable: the nightly sweep refreshes the tracking issue every night while the failure set is unchanged, so a false positive here becomes permanent noise in a consumer repo, and it sits alongside two genuine findings where it competes for the maintainer's attention. Any consumer that locks down tags with a no-bypass ruleset rather than an admin-bypass one would see the same thing.

Happy to test a fix against worktrunk's configuration if that's useful.

tend-agent:
## Problem

`tend check` reports a FAIL on repos whose tags are locked down exactly as tend's own recipe prescribes:

> **`secret-environments`** — the `release` and `signing` environments each hold secrets, have no required reviewers, and admit tags, and no active all-tags ruleset restricting tag creation/update to admins could be verified.

Reported in #824 against max-sixty/worktrunk. It reproduces on this repo: on `main` at 4672f809cedd2f206e109761072641798aa0184e, `tend check` run as `tend-agent` fails `credential-environments` for the `release` environment, and passes `branch-protection:main` only through the weak `.protected` fallback rather than by confirming the ruleset.

## Root cause

GitHub serves a ruleset's `bypass_actors` **only to a ruleset admin**. `tend check` runs under the bot's own write-scoped token, so on a correctly locked-down repo the key is absent from every ruleset it reads — and `_bypass_actors_above_bot` maps absent to `None` (unverifiable). Every ruleset verdict is therefore unverifiable in the nightly, which `_tags_admin_gated` turns into "no all-tags ruleset could be verified" and `_policy_gate` turns into a FAIL.

Absent is *not* empty, so #824's suggested "treat a missing `bypass_actors` as no bypass actors" would be a false pass: this repo's `Merge access` ruleset was created by tend's own `--fix` with an admin bypass actor, and it reads back to the bot with no `bypass_actors` key at all. Verified against live GitHub on a repo where the bot *is* admin — as a ruleset admin, an empty bypass list is served as `"bypass_actors": []`, never omitted.

## Solution

Use `current_user_can_bypass`, which GitHub returns to any reader and which answers the question for the requesting user only — exactly the actor in question when the requester is the bot. `bypass_actors` stays the first signal; the new fallback fires only when that list is unusable **and** `gh api user` says the token is the configured `bot_name`. `never` means blocked; every other value names a bypass mode, so an unrecognized one fails closed.

This is the reading [`shared/steps/security-preflight.sh`](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/shared/steps/security-preflight.sh#L26-L35) already applies to the same field with the same token for the merge restriction, and which `docs/security-model.md` describes as "GitHub's own evaluation of the bot's standing — teams, custom roles, and org-level rulesets included". The check was simply not using it.

Side effect: it also settles cases the actor list can never decide. A `Team` or `Integration` bypass actor whose membership isn't visible used to be permanently unverifiable; GitHub has already resolved it into `current_user_can_bypass`.

## Relationship to #825

Same root cause, different remedy — a maintainer should pick one. #825 keeps the verdict unverifiable and stops reporting it as a failure (FAIL → "could not read the whole credential surface"). This PR removes the unverifiability in the case the nightly actually runs in, so the check reports a real PASS and a genuinely ungated environment can't hide in the unknown bucket. They are compatible if both land: #825's unknown branch would then cover only the residual cases (a non-bot token that isn't a ruleset admin), and its test still passes under this change.

## Testing

Four unit tests, three of which fail on `main`; plus a live before/after on this repo.

<details><summary>Live A/B on max-sixty/tend, run as <code>tend-agent</code></summary>

Before (in-tree `main`):

```
PASS  branch-protection:main — Branch 'main' is protected
FAIL  credential-environments — A run the bot can cause reaches a credential: 'release' has no required reviewers, and admits tags, and no active all-tags ruleset restricting creation and update to admins could be verified.
```

After:

```
PASS  branch-protection:main — Branch 'main' is protected (restrict-updates ruleset)
PASS  credential-environments — Credential-holding environments are gated: cloudflare-deploy, codex-auth-refresh, github-pages, release, tend
```

The branch-protection line changes too: it was passing on the "can't read details — assume OK" fallback, and now passes on a positively verified ruleset.

</details>

<details><summary>API evidence that absent ≠ empty</summary>

Read as the write-access bot, `repos/max-sixty/tend/rulesets/14219339` (`Merge access`, created by `tend check --fix` with `bypass_actors: [{actor_id: 5, actor_type: RepositoryRole}]`):

```json
{"name": "Merge access", "current_user_can_bypass": "never", "bypass_actors": <key absent>}
```

Two probe rulesets created on `tend-agent/tend-integration`, where the bot is admin, then deleted:

| ruleset | `bypass_actors` (as admin) | `current_user_can_bypass` |
|---|---|---|
| bypass = admin role | `[{"actor_id": 5, "actor_type": "RepositoryRole", ...}]` | `exempt` |
| bypass = none | `[]` | `never` |

So a ruleset admin always sees the list, empty included; a non-admin never does. And `exempt` is a live value the documented enum (`always`, `pull_requests_only`, `never`) doesn't list — hence treating anything but `never` as a bypass.

</details>

Closes #824

### DIFF
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -303,7 +303,12 @@ def _ruleset_excludes_bot(data: dict, bot_name: str) -> bool | None:
     if verdict is not None:
         return verdict
     can_bypass = data.get("current_user_can_bypass")
-    if can_bypass is None or _authenticated_login() != bot_name:
+    login = _authenticated_login()
+    # Casefolded: `bot_name` is hand-written in `.config/tend.yaml` while the
+    # API returns the canonical casing, and every other comparison of the two
+    # (`_reviewer_gate`, and the URL paths GitHub resolves case-insensitively)
+    # already tolerates the difference.
+    if can_bypass is None or login is None or login.casefold() != bot_name.casefold():
         return None
     return can_bypass == CANNOT_BYPASS
 
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -52,6 +52,14 @@ BYPASS_ROLE_IDS = frozenset({ROLE_ID_MAINTAIN, ROLE_ID_ADMIN})
 # the ruleset, so the bot can't be ruled out.
 BYPASS_ACTOR_TYPES_ABOVE_BOT = frozenset({"OrganizationAdmin", "EnterpriseOwner"})
 
+# The one `current_user_can_bypass` value that grants the requester no bypass.
+# Every other value names a mode in which it bypasses — the docs list `always`
+# and `pull_requests_only`, live responses also return `exempt` — so reading
+# only `never` as blocked fails closed on a value added later. This is the
+# reading `shared/steps/security-preflight.sh` already applies to the same
+# field with the same token.
+CANNOT_BYPASS = "never"
+
 # Triggers a write-scoped actor can both fire *and* steer — it decides not only
 # that the run happens but what the run publishes. A deployment branch policy
 # does not gate these, because the actor fires them at a ref the policy already
@@ -268,6 +276,38 @@ def _bypass_actors_above_bot(actors: list[dict] | None, bot_name: str) -> bool |
     return None if unresolved else True
 
 
+def _authenticated_login() -> str | None:
+    """The login `gh` is authenticated as, or None when it can't be read."""
+    result = _gh("api", "user", "--jq", ".login")
+    if result is None or result.returncode != 0:
+        return None
+    return result.stdout.strip() or None
+
+
+def _ruleset_excludes_bot(data: dict, bot_name: str) -> bool | None:
+    """Whether a ruleset payload keeps a write-access bot out.
+
+    `bypass_actors` is the first signal, but GitHub serves it only to a ruleset
+    admin — and `tend check` runs under the bot's own write-scoped token, so on
+    a correctly locked-down repo the key is absent from every ruleset it reads,
+    whatever the list holds. Absent is therefore not empty: tend's own `--fix`
+    writes an admin bypass, and that ruleset reads back with no `bypass_actors`
+    key at all.
+
+    `current_user_can_bypass` closes that gap from the other side. It answers
+    for the requesting user only, which is exactly the actor in question
+    whenever the requester *is* the bot — and it accounts for team and app
+    membership the actor list leaves unresolvable.
+    """
+    verdict = _bypass_actors_above_bot(data.get("bypass_actors"), bot_name)
+    if verdict is not None:
+        return verdict
+    can_bypass = data.get("current_user_can_bypass")
+    if can_bypass is None or _authenticated_login() != bot_name:
+        return None
+    return can_bypass == CANNOT_BYPASS
+
+
 def _ruleset_blocks_bot(repo: str, ruleset_id: int, bot_name: str) -> bool | None:
     """Whether a ruleset's bypass list keeps a write-access bot out.
 
@@ -284,7 +324,7 @@ def _ruleset_blocks_bot(repo: str, ruleset_id: int, bot_name: str) -> bool | Non
         return None
     if not isinstance(data, dict):
         return None
-    return _bypass_actors_above_bot(data.get("bypass_actors"), bot_name)
+    return _ruleset_excludes_bot(data, bot_name)
 
 
 def _tags_admin_gated(repo: str, bot_name: str) -> bool | None:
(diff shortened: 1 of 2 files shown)


generator/src/tend/checks.py
```diff
@@ -268,6 +276,38 @@ def _bypass_actors_above_bot(actors: list[dict] | None, bot_name: str) -> bool |
...
+    `current_user_can_bypass` closes that gap from the other side. It answers
+    for the requesting user only, which is exactly the actor in question
+    whenever the requester *is* the bot — and it accounts for team and app
+    membership the actor list leaves unresolvable.
+    """
+    verdict = _bypass_actors_above_bot(data.get("bypass_actors"), bot_name)
+    if verdict is not None:
+        return verdict
+    can_bypass = data.get("current_user_can_bypass")
+    if can_bypass is None or _authenticated_login() != bot_name:
```
tend-agent: `bot_name` comes from `.config/tend.yaml`, hand-written, while `_authenticated_login()` returns GitHub's canonical casing. A mismatch here fails safe (back to `None`), but it is silent and reinstates the false FAIL this PR exists to remove. Casefold both, as `_reviewer_gate` does a few hundred lines down:

```suggestion
    can_bypass = data.get("current_user_can_bypass")
    login = _authenticated_login()
    if can_bypass is None or login is None or login.casefold() != bot_name.casefold():
```

tend-agent (commented):
One finding, inline: the new `_authenticated_login() != bot_name` compare is the only place in `checks.py` that matches `bot_name` against a GitHub login as a raw string. `_reviewer_gate` in the same module already casefolds for exactly this comparison (`bot_name.casefold() in {login.casefold() for login in reviewers}`), and everywhere else `bot_name` reaches GitHub through a URL path, where the API resolves logins case-insensitively — so a `bot_name` whose casing differs from the canonical login passes `check_bot_permission`, `_user_id`, and `_reviewer_gate`, then silently fails only here, restoring the exact FAIL this PR removes with a config that looks correct.

On the #825 relationship: I applied #825's diff on top of this branch — both patch cleanly (they touch disjoint regions of `checks.py`) and the combined suite is 333 passed, so the body's "compatible if both land" holds under test. Which remedy to take is still a maintainer call.

generator/src/tend/checks.py
tend-agent: Applied in 27dd833, with a test that pins the case-only difference (`My-Bot` vs `my-bot`). Comment on the compare points at `_reviewer_gate` as the precedent.

max-sixty:
Closing in favour of #825.

Both fix the false FAIL. #825 reports the unreadable case as unknown; this PR derives a verdict for it from `current_user_can_bypass`. The second is the stronger claim, and it buys that strength with a second source of truth for one question, reached through a fallback chain and a casefolded login match against `bot_name`.

The weaker fix is enough. `tend check` collects failures with `r.passed is False`, so an unknown prints as SKIP and the command still exits 0.

Worth returning to if the unknown proves too quiet — if a genuinely ungated environment starts reading as a skip.

PR #833, opened 2026-08-05
fix(review-runs): look for in-thread maintainer corrections, not just dispositions

### LINKED ISSUE #827: review-runs Step 4 can't see a maintainer correcting a bot comment — only PR dispositions
`review-runs` **Step 4: Cross-check outcomes** is the only place a run learns that a maintainer disagreed with tend's output, and `running-in-ci` → **Learning from Feedback** depends on that discovery to turn corrections into durable guidance. But Step 4's guidance and its one worked example are both PR-shaped, so maintainer replies that correct a bot **comment on an issue** are never looked for — and the run reports "0 maintainer corrective comments" while one sits in the window.

The only recipe in the section is:

````
```bash
# Example: check if a bot PR was merged or closed
gh pr list --author "$BOT_LOGIN" --state all --json number,title,state,closedAt \
  --jq '.[] | select(.closedAt > "'$SINCE'")'
```
````

The prose does ask, for **Review**, "Did human reviewers flag issues it missed?" — but the **Triage** line asks only "was the classification correct? Did the issue get relabeled?". Both of those are *dispositions*, and so is everything the recipe can return. Neither reaches the most common form of correction on a triage-heavy repo: the maintainer replying in-thread that the bot's technical claim was wrong, leaving the issue's labels and state otherwise untouched.

## Observed miss

On `max-sixty/worktrunk`, `tend-triage` [answered](https://github.com/max-sixty/worktrunk/issues/3696#issuecomment-5153609243) a feature request with "there's no way to pin it per-invocation independent of that config". max-sixty [replied](https://github.com/max-sixty/worktrunk/issues/3696#issuecomment-5159480766) 2026-08-02T17:18Z: "the need is real, but it's already met, so I'm not going to add the flag", showing the existing flag that does it, and declined the proposed new flag as "a third spelling of the same thing on a surface I deliberately keep small".

The `review-runs` run whose 24h window contained that reply — [30798872347](https://github.com/max-sixty/worktrunk/actions/runs/30798872347), 2026-08-03 — recorded in its tracking-issue entry:

> **0 bot PRs closed-without-merge**, **0 maintainer corrective comments** → nothing to codify via Learning-from-Feedback.

Its session log shows why. It did call the repo-wide comments endpoint, but projected only its own side of it:

```
gh api "repos/$REPO/issues/comments?since=$SINCE&per_page=100" \
  --jq '[.[] | select(.user.login=="worktrunk-bot")] | length'
```

`select(.user.login=="worktrunk-bot")` — the bot's own comment count, for volume. No query anywhere in the run filtered for maintainer comments, and no issue thread was opened to check for a reply. The data was one `jq` filter away and the skill never asks for it.

## Why this compounds

The miss is invisible by construction: a run that doesn't look reports "all clear", writes that into the monthly tracking issue, and future runs read it as ground truth when counting occurrences under Gate 1. So a correction that recurs never accumulates the occurrences its own gate requires — the evidence store records the *absence* of the thing it exists to count. And `Learning from Feedback` never fires for the entire triage surface, which on an issue-heavy consumer is where most maintainer corrections land.

Structural, not stochastic: no decision point. Any run following Step 4 as written looks at PR dispositions and stops.

## Suggested fix

Add the mirror-image query to Step 4, next to the existing PR one — the endpoint is already repo-wide and takes `since`, so it's one call:

````
```bash
# Maintainer/human replies in the window, on any issue or PR thread.
# A reply on a thread the bot commented on earlier is the correction signal —
# Learning from Feedback keys on these, and a PR-disposition query can't see them.
gh api "repos/$REPO/issues/comments?since=$SINCE&per_page=100" \
  --jq '.[] | select(.user.login != "'"$BOT_LOGIN"'")
        | {at: .created_at, url: .html_url, body: .body[0:300]}'
```
````

...and a sentence making the ask explicit: a correction is not only a closed PR or a relabeled issue — it is also a maintainer stating in-thread that a bot claim was wrong. Read the threads where the bot commented earlier in the window, and count a contradiction as a finding.

Worth pairing with a note that "0 maintainer corrective comments" should only be written after that query ran, since the phrase is what future runs inherit.

`review-reviewers` already has the right framing — its survey prompt names "rejected, **corrected**, or ignored" as the outcome classes. Its per-workflow mapping recipes are still run→PR/issue shaped, so it may benefit from the same query, but the wording gap is specific to `review-runs` Step 4.

Repo-side, I've codified the specific worktrunk lesson in max-sixty/worktrunk#3736; this issue is the generic half — every consumer's `review-runs` has the same blind spot regardless of what the correction was about.

### PR DESCRIPTION
**Pull request by tend-agent**
## Problem

`review-runs` **Step 4: Cross-check outcomes** is the only place a run discovers that a maintainer disagreed with tend's output, and `running-in-ci` → **Learning from Feedback** depends on that discovery. But Step 4's single recipe was `gh pr list --author "$BOT_LOGIN"` — a PR-disposition query — and its prose asked only about merges, closures, and relabels. Every question it posed was answerable from state changes.

That misses the most common correction shape on an issue-heavy repo: a maintainer replying in-thread that a bot claim was wrong, leaving the issue's labels and state untouched. Nothing in the section asks for it, so a run following Step 4 as written reports "0 maintainer corrective comments" while one sits in its window.

The miss compounds: the all-clear gets written into the monthly tracking issue, and future runs read it as ground truth when counting occurrences under Gate 1 — so a recurring correction never accumulates the occurrences its own gate requires. Structural, not stochastic; there is no decision point.

## Solution

Adds the mirror-image query next to the existing disposition one — same repo-wide `since` endpoints, inverted login filter — plus a sentence naming in-thread contradiction as a correction class, and a note that "no maintainer corrections" should only be written after the query ran.

`issues/comments` covers issue and PR conversation comments; `pulls/comments` covers inline review replies, where corrections to `tend-review` output land. Both are repo-wide, so it's two calls regardless of how many threads the bot touched.

The query is `--paginate`d. That is load-bearing rather than cosmetic: the response is ascending by `created_at`, so a single 100-item page drops the *newest* comments — precisely where a fresh correction sits — and the bot's own comments consume that budget before the login filter runs. A silently truncated result would reintroduce the miss this PR exists to close.

The projection reports `updated_at` alongside `created_at`, because `since` filters on `updated_at`: a comment created before the window appears on the strength of an in-window edit. That is worth keeping — an edited claim is still a correction — but without both timestamps it reads as a broken filter.

## Testing

Ran the recipe verbatim against this repo. It surfaces exactly the class of comment the disposition query cannot see — e.g. [#800](https://github.com/max-sixty/tend/pull/800#issuecomment-5152652855) ("Not worth the complication.") and [#817](https://github.com/max-sixty/tend/issues/817#issuecomment-5172269199), neither reachable from a closed/merged/relabeled signal. Both endpoints accept `since`; the `for endpoint in issues pulls` form and the nested-quoting in the `jq` filter were executed as written, not just composed.

Measured the truncation the `--paginate` guards against, over a 7-day window on this repo: the unpaginated call returns 100 rows, `--paginate` returns 142, and the 42 dropped are the newest — the boundary landed at `2026-08-04T16:28:43Z`. In this particular window all 42 were the bot's own comments, so no human comment was actually lost; the mechanism is what's demonstrated, not an observed loss.

Confirmed the ordering claim directly: `created_at` is monotonic across the response and `updated_at` is not. Confirmed the `since` semantics the same way — a comment created `2026-08-01T08:39:05Z` appears in a 24h window because it was updated `2026-08-04T08:50:03Z`.

Also verified the reported miss reproduces from the skill text alone: Step 4 before this change contained no query filtering for non-bot comments.

---
Closes #827 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -193,12 +193,19 @@ Dispositions — merged, closed, relabeled, reverted — are only half the signa
 ```bash
 # Human replies in the window, on any issue or PR thread. Both endpoints are
 # repo-wide and take `since`, so this is two calls regardless of thread count.
+# `--paginate` is required, not cosmetic: the response is ascending by
+# `created_at`, so a single 100-item page drops the *newest* comments — the
+# ones a fresh correction sits in. The bot's own comments consume that budget
+# before the filter runs, so a busy day truncates with nothing in the output
+# to say so.
 for endpoint in issues pulls; do
-  gh api "repos/$REPO/$endpoint/comments?since=$SINCE&per_page=100" \
+  gh api --paginate "repos/$REPO/$endpoint/comments?since=$SINCE&per_page=100" \
     --jq '.[] | select(.user.login != "'"$BOT_LOGIN"'")
-          | {at: .created_at, url: .html_url, body: .body[0:300]}'
+          | {created: .created_at, updated: .updated_at, url: .html_url, body: .body[0:300]}'
 done
 ```
+
+`since` filters on `updated_at`, not `created_at`, so results include older comments edited inside the window — a comment created days before `$SINCE` is a real hit, not a broken filter. That is worth having (an edited claim is still a correction); the projection reports both timestamps so the reason a row qualified is visible.
 
 Write "no maintainer corrections" into the tracking issue only after that query ran — future runs read the phrase as ground truth when counting occurrences under Gate 1, so an unchecked all-clear suppresses the evidence it exists to accumulate.
 
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -183,10 +183,24 @@ For each analyzed run, compare what the bot did against what happened next. The 
 mention, notifications, weekly, and review-reviewers runs get the same treatment: find the bot's output and check whether it was accepted.
 
 ```bash
-# Example: check if a bot PR was merged or closed
+# Bot PR dispositions — merged or closed in the window
 gh pr list --author "$BOT_LOGIN" --state all --json number,title,state,closedAt \
   --jq '.[] | select(.closedAt > "'$SINCE'")'
 ```
+
+Dispositions — merged, closed, relabeled, reverted — are only half the signal. A maintainer replying in-thread that a bot claim was wrong, leaving labels and state untouched, is equally a correction, and on an issue-heavy repo it is the most common one. No disposition query can see it. Read the threads where the bot commented in the window and count a contradiction as a finding:
+
+```bash
+# Human replies in the window, on any issue or PR thread. Both endpoints are
+# repo-wide and take `since`, so this is two calls regardless of thread count.
+for endpoint in issues pulls; do
+  gh api "repos/$REPO/$endpoint/comments?since=$SINCE&per_page=100" \
+    --jq '.[] | select(.user.login != "'"$BOT_LOGIN"'")
+          | {at: .created_at, url: .html_url, body: .body[0:300]}'
+done
+```
+
+Write "no maintainer corrections" into the tracking issue only after that query ran — future runs read the phrase as ground truth when counting occurrences under Gate 1, so an unchecked all-clear suppresses the evidence it exists to accumulate.
 
 ## Step 5: Deduplicate
 

### CODE REVIEW

**Review by tend-agent — COMMENTED** [review]
The new query has a silent-truncation hole that reintroduces the miss it's meant to close: `gh api` returns only the first page without `--paginate`, and the page is ordered oldest-first, so once the window holds more than 100 comments the *newest* ones are dropped — exactly where a fresh maintainer correction sits. The bot's own comments consume that budget: running the recipe on this repo just now returned 88 `issues/comments` in a 24h window, all 88 authored by `tend-agent`, with zero non-bot comments surviving the filter. A day 15% busier drops real signal without any indication in the output. Inline suggestion below.

Worth a word in the comment too: `since` on both endpoints filters on `updated_at`, not `created_at`. The run above returned comments created `2026-08-01` inside a 24h window, while the projection reports `at: .created_at` — someone following the recipe sees four-day-old timestamps and reasonably concludes the filter is broken. Surfacing edited comments is arguably a feature here, but the recipe should say so rather than let it read as a bug.

File: plugins/tend-ci-runner/skills/review-runs/SKILL.md [hunk 1]
```diff
@@ -183,11 +183,25 @@ For each analyzed run, compare what the bot did against what happened next. The
...
   --jq '.[] | select(.closedAt > "'$SINCE'")'
 ```
 
+Dispositions — merged, closed, relabeled, reverted — are only half the signal. A maintainer replying in-thread that a bot claim was wrong, leaving labels and state untouched, is equally a correction, and on an issue-heavy repo it is the most common one. No disposition query can see it. Read the threads where the bot commented in the window and count a contradiction as a finding:
+
+```bash
+# Human replies in the window, on any issue or PR thread. Both endpoints are
+# repo-wide and take `since`, so this is two calls regardless of thread count.
+for endpoint in issues pulls; do
+  gh api "repos/$REPO/$endpoint/comments?since=$SINCE&per_page=100" \
```
tend-agent: Without `--paginate` this caps at the first 100 comments, oldest-first, so the newest maintainer reply is the first thing lost on a busy day.

```suggestion
  gh api --paginate "repos/$REPO/$endpoint/comments?since=$SINCE&per_page=100" \
```

tend-agent: Confirmed and fixed in 4bca2ba — the truncation is real and it drops exactly the wrong end.

Measured on this repo over a 7-day window: the unpaginated call returns 100, `--paginate` returns 142. The page is ascending by `created_at` (`created_at` is monotonic across the response, `updated_at` is not), so the 42 dropped rows are the newest — the truncation boundary landed at `2026-08-04T16:28:43Z`, discarding everything created after it.

One correction to the framing, in your favor on the mechanism but against overstating the observed loss: in this particular window all 42 dropped rows were `tend-agent`'s own, so no human comment was actually lost this time. The signal loss is a matter of window composition, not something I observed — the mechanism is what's proven.

Also took the `since` point. It filters `updated_at`, so a comment created `2026-08-01T08:39:05Z` legitimately appears in a 24h window on the strength of an `2026-08-04T08:50:03Z` edit. The projection now reports both timestamps so the reason a row qualified is visible, and the prose says the older-created rows are hits rather than a broken filter.

```console
$ gh api "repos/max-sixty/tend/issues/comments?since=$SINCE7&per_page=100" --jq 'length'
100
$ gh api --paginate "repos/max-sixty/tend/issues/comments?since=$SINCE7&per_page=100" --jq '.[].updated_at' | wc -l
142
```

Truncation boundary and authorship of the dropped rows:

```json
{"total":142,"truncation_boundary":"2026-08-04T16:28:43Z","dropped":42,"dropped_authors":{"tend-agent":42}}
```

Ordering — `created_at` ascending, `updated_at` not monotonic:

```json
{"created_monotonic": true}
{"updated_monotonic": false}
```

The corrected recipe run verbatim over 7 days surfaces 8 human comments across both endpoints, including the two the PR body cites (#800, #817).

</details>

PR #834, opened 2026-08-05
fix(review): submit the review before pushing a fix

### LINKED ISSUE #830: review: pushing a fix from inside a review session cancels that session, discarding the unsubmitted review
## Problem

`review` step 8 instructs the session to push fixes to bot-authored PRs, and step 6's failure branch routes there explicitly — *"On **bot PRs** (Dependabot, renovate, etc.), don't stop at analysis: apply the fix per step 8 so the PR can go green."*

The generated `tend-review.yaml` guards the job with:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

and the workflow triggers on `synchronize`. So a push from inside a `tend-review` session fires a `synchronize` event on the same PR, which starts a replacement run in the same concurrency group, which cancels the pusher. Everything the session had not yet done at push time is lost — including submitting the review it was in the middle of assembling.

The skill anticipates seeing cancellations (*"Cancellations are almost always caused by concurrency groups… The replacement run will cover the cancelled checks"*) but that reasoning is about *check runs* on the PR. It does not hold for the session itself: the replacement run starts from scratch, sees a HEAD that now differs from the original, and reasons about the incremental — it does not resume the cancelled review.

## Observed

[Run 30889113519](https://github.com/max-sixty/cargo-affected/actions/runs/30889113519) (`tend-review` on `551e059`) found `lint` red and decided, correctly, to fix it:

> All three test platforms pass, but `lint` is red on this HEAD — `cargo fmt --check` fails on the new `filter_expression_rejected` body. Since this is a bot-authored PR, I'll push the fix rather than leave it red.

Timeline from the job log and the run list:

| time | event |
|---|---|
| 07:47:30 | agent step starts |
| 07:51:09 | session pushes `6955352` (the rustfmt fix) |
| 07:51:14 | `synchronize` starts replacement run [30889478636](https://github.com/max-sixty/cargo-affected/actions/runs/30889478636) |
| 07:51:25 | original run cancelled mid-tool-call |

Its last tool call was the step-7 review-threads GraphQL query. It never reached step 5 — no review was submitted, and no review of `551e059` exists on the PR. The run's session log uploaded (artifact `claude-session-logs-n54`, 74 KB) but the run shows `conclusion: cancelled`, which is indistinguishable at a glance from the routine superseded-by-newer-event cancellations this repo produces all day.

The replacement run then skipped the commit as already-reviewed, for an unrelated reason filed separately. But even with that fixed, the review this session had already done was thrown away and had to be re-derived.

## Why it will recur

Fully structural, no decision point. Given `cancel-in-progress: true` on a `synchronize`-triggered workflow, a push from inside the run always cancels the run. It is not probabilistic and not model-dependent — the only variable is how much of the session was left to do.

## Proposed fix

Order the work so nothing valuable sits after the push. In `review` step 8, before the push commands:

> **A push from inside a review session cancels that session.** `tend-review` triggers on `synchronize` with `cancel-in-progress: true` on a per-PR group, so pushing to the branch you are reviewing starts a replacement run that cancels this one — typically within ~15s, mid-tool-call. The replacement run does not resume your review; it starts cold against the new HEAD. So finish and submit the review (step 5) and resolve any threads (step 7) **before** pushing. Treat the push as the last action of the session, and state in the review that the fix is coming, rather than reviewing after.

Alternative, if the ordering constraint is too rigid: have the pushing session skip the fix and leave the red for the replacement run, which will see the failure on its own pre-flight. That trades a round-trip for simplicity, but the ordering rule seems strictly better since the session already has the analysis in context.

tend-agent:
## Problem

`tend-review` triggers on `pull_request_target: synchronize` and guards the job with a per-PR concurrency group at `cancel-in-progress: true` ([`review.yaml.j2`](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/review.yaml.j2#L10-L18)). Step 8 tells the session to push fixes to bot-authored PRs, and step 6's failure branch routes there explicitly — but a push from inside the session fires `synchronize`, which starts a replacement run in the same group and cancels the pusher within seconds. The replacement starts cold against the new HEAD; it does not resume the cancelled review. Anything the session had not yet shipped at push time is lost.

Nothing in the skill said the push had to come last, so a session that pushed a fix before submitting its review lost the review entirely — and the run's `conclusion: cancelled` is indistinguishable from the routine superseded-by-newer-event cancellations these repos produce all day. Reported in #830 with a traced timeline.

## Solution

Two edits to `plugins/tend-ci-runner/skills/review/SKILL.md`:

- Step 8 opens with the ordering rule: a push cancels this session, so submit the review (step 5) and resolve threads (step 7) first and make the push the session's last action, stating in the review that the fix is coming. It also names the interaction with `running-in-ci`'s "a pushed fix is always gated" rule — the pushed fix's CI cannot be polled here, because the session ends before it settles, and nothing else reliably picks it up either (the replacement run's pre-flight reads reviews and the diff, never the rollup, and a small fix commit takes step 1's trivial-skip path past steps 2–6). So the rule is to name the pushed fix as CI-unverified in the review body, keeping the gap visible in the thread.
- Step 5's terminal-red bullet — the point where a session that finds red is tempted to fix before posting — gets a one-clause pointer to post the COMMENT before pushing.

The alternative in #830 (have the pushing session skip the fix and leave the red for the replacement run) trades a round-trip for simplicity; the ordering rule is strictly better since the session already has the analysis in context.

## Testing

Skill prose, so no test to add. The mechanism was verified directly rather than taken from the report: the template's trigger and concurrency block confirm the structural claim, and the GitHub API confirms the cited timeline — the original run went `cancelled` at 07:51:29 having been created at 07:45:49, the replacement run was created at 07:51:14, and the only reviews on the cancelled run's HEAD are the two empty-body containers GitHub creates for inline replies (no substantive review was ever submitted for that commit).

Checked the co-loaded skills before writing new guidance: `running-in-ci` mentions `cancel-in-progress: true` only in the CI-monitoring poll filter, where cancellation of the older sibling is noted as a no-op for polling. No skill carried the ordering rule, so this isn't duplicated guidance.

---
Closes #830 — automated triage


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -405,6 +405,10 @@ Outdated comments (null line) are best-effort — skip if the original context c
 
 ### 8. Push fixes
 
+**A push cancels this session — make it the last thing you do.** `tend-review` triggers on `synchronize` with `cancel-in-progress: true` on a per-PR group, so pushing to the branch under review starts a replacement run that cancels this one within seconds, mid-tool-call. The replacement run starts cold against the new HEAD and does not resume this review, so anything still pending at push time is lost. Submit the review (step 5) and resolve threads (step 7) **before** pushing, and say in the review body that the fix is coming rather than planning to review after.
+
+This is the one exception to `running-in-ci`'s "a pushed fix is always gated" rule: you cannot poll the pushed fix's CI, because the session ends before it settles. The replacement run reads the rollup on the new HEAD in its own pre-flight.
```
tend-agent: The pre-flight doesn't read the rollup. Step 1 fetches commits, reviews, comments and the diff; the only `statusCheckRollup` reads in this skill are step 5's pre-APPROVE peek and step 6's monitoring loop. And a one-commit fix is exactly the shape step 1 calls trivial (`under ~20 added+deleted lines`), which routes to "go directly to step 7 ... Do NOT proceed to steps 2–6" — so the replacement run resolves threads, possibly submits the empty-body APPROVE, and never looks at the new HEAD's checks.

That matters because this sentence is the justification for waiving the gating rule. As written, a red pushed fix on a bot PR has nothing watching it: the pusher is cancelled and the replacement takes the trivial path. Naming the gap in the review body at least leaves it visible in the thread.

```suggestion
This is the one exception to `running-in-ci`'s "a pushed fix is always gated" rule: you cannot poll the pushed fix's CI, because the session ends before it settles — and nothing else reliably picks it up. The replacement run's pre-flight reads reviews and the diff, never the rollup, and a small fix commit takes the trivial-skip path that bypasses steps 2–6. So name the pushed fix as CI-unverified in the review body, which keeps the gap visible in the thread.
```

tend-agent (commented):
The ordering rule itself checks out — the template's `pull_request_target: synchronize` trigger and the `${{ github.workflow }}-${{ github.event.pull_request.number }}` group at `cancel-in-progress: true` make the cancellation structural, and the fix lands the rule at the two places a session would otherwise push first.

One inline comment on the second paragraph: the claim that hands off CI verification to the replacement run doesn't match what the skill tells that run to do.

Note for whoever lands these: #821 inserts a bullet immediately above the line this PR edits in the same `$FAILED`/`$PENDING` list, and #835 also touches this file. Not duplicates — but expect a textual conflict between #834 and #821 depending on merge order.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Confirmed, and the suggestion is applied in f5aa6d0. Step 1's pre-flight reads commits, reviews, comments and the diff — no `statusCheckRollup` anywhere in it — and the trivial-skip branch says "go directly to step 7 … Do NOT proceed to steps 2–6", which is where a one-commit fix lands. So the original sentence was asserting a handoff that doesn't happen, on the one line that justifies waiving the gating rule. Naming the fix as CI-unverified in the review body is the honest version.

The PR body carried the same wrong claim; updated it to match.

Noted on the #821/#834 textual conflict in the `$FAILED`/`$PENDING` list — whichever lands second will need a trivial rebase.

PR #835, opened 2026-08-05
fix(review): ignore synthetic reply containers in all three review-record guards

### LINKED ISSUE #828: review: LAST_REVIEW_SHA anchors on the empty review container GitHub creates for an inline reply, so the next run skips a real commit
## Problem

`review`'s pre-flight derives `LAST_REVIEW_SHA` from the bot's most recent review record:

```bash
LAST_REVIEW_SHA=$(gh pr view <number> --json reviews \
  --jq "[.reviews[] | select(.author.login == \"$BOT_LOGIN\")] | last | .commit.oid // empty")
```

The adjacent comment explains why no filter is applied — *"any state counts — COMMENTED reviews carry inline comments even when the body is empty"*. That is true for a real review submitted with only inline comments. But GitHub creates the **same** record shape for something that is not a review at all: replying to an existing review thread via `POST /pulls/{n}/comments/{id}/replies` wraps the reply in a synthetic zero-body `COMMENTED` review, anchored at whatever HEAD was when the reply was posted.

So after any `tend-mention` run answers review threads inline — which is its normal response path — `LAST_REVIEW_SHA` advances to the current HEAD even though nothing reviewed it. The next `tend-review` run then computes its incremental against a commit it never read, and skips everything in between under the trivial-change heuristic.

## Observed

max-sixty/cargo-affected PR #54, 2026-08-04. Reviews on that PR, via REST:

| review | submitted | `commit_id` | body |
|---|---|---|---|
| 4851810736 | 07:41:08Z | `1ffcde2` | 2441 chars — the real review |
| 4851851883 | 07:46:08Z | `551e059` | **0** |
| 4851852052 | 07:46:09Z | `551e059` | **0** |

The two empty records each contain exactly one comment, and that comment is a reply:

```
$ gh api repos/max-sixty/cargo-affected/pulls/54/reviews/4851851883/comments
{"id":3710432687,"in_reply_to_id":3710403279,"created_at":"2026-08-04T07:46:07Z",
 "body":"Applied in 551e059 — the call site now propagates with `?`."}
```

They exist only because [run 30888815177](https://github.com/max-sixty/cargo-affected/actions/runs/30888815177) posted two inline replies at 07:46:07 while responding to review 4851810736.

Commit `551e059` was a substantive change — `filter_expression_rejected` converted to `Result<bool>`, a 48-line test helper deleted, a functional test rewritten. The `tend-review` run that should have reviewed it ([30889113519](https://github.com/max-sixty/cargo-affected/actions/runs/30889113519)) was cancelled mid-session before submitting (separate issue: it self-cancelled by pushing to the branch it was reviewing). Its successor [30889478636](https://github.com/max-sixty/cargo-affected/actions/runs/30889478636) then read `LAST_REVIEW_SHA=551e059` off the reply container, computed the incremental as `551e059..6955352` — five lines of rustfmt — applied the trivial-skip heuristic, and posted nothing. Its session summary stated:

> **Pre-flight:** The bot already reviewed commit `551e059`; HEAD is now `6955352`.

No review of `551e059` exists. The correct incremental was `1ffcde2..6955352`, which contains the whole refactor.

The failure is invisible: nothing goes red, no check fails, and the run reports a well-reasoned decision to stay silent. On a repo where the bot both authors and reviews its own PRs, that removes the only review signal the PR gets.

## Why it will recur

Fully deterministic, no decision point. Any inline reply anchors a zero-body review record at the then-current HEAD; the next review run reads it as the last-reviewed SHA. `tend-mention`'s standard "reply inline, then push the fix" flow produces this shape every time.

## Proposed fix

Count a review record only when it carries a body **or** at least one *top-level* (non-reply) inline comment. Reply containers have neither. Verified against PR #54 — the current recipe returns `551e059`, this returns `1ffcde2`:

```bash
# Review IDs that own at least one top-level inline comment. A review whose
# only comments are replies (`in_reply_to_id` set) is a synthetic container
# GitHub created for a thread reply, not a review — it must not anchor
# LAST_REVIEW_SHA.
SUBSTANTIVE=$(gh api "repos/$REPO/pulls/<number>/comments?per_page=100" \
  --jq '[.[] | select(.in_reply_to_id == null) | .pull_request_review_id] | unique')

# NOTE: REST reviews use .commit_id (not .commit.oid from `gh pr view --json`).
LAST_REVIEW_SHA=$(gh api "repos/$REPO/pulls/<number>/reviews" \
  | jq -r --argjson sub "$SUBSTANTIVE" --arg bot "$BOT_LOGIN" \
    '[.[] | select(.user.login == $bot)
          | select((.body | length) > 0 or (.id | IN($sub[])))]
     | last | .commit_id // empty')
```

`gh api --jq` does not accept `--arg`/`--argjson`, so the second call pipes to `jq` rather than using `--jq`.

This keeps the behaviour the existing comment is protecting — an empty-body review that carries real inline findings still anchors — while excluding containers whose only content is a reply.

tend-agent:
## Problem

`review` derives three separate signals from the bot's own review records, and all three counted any record the bot owns. GitHub creates the *same* record shape for something that is not a review: replying to a review thread via `POST /pulls/{n}/comments/{id}/replies` wraps the reply in a synthetic zero-body `COMMENTED` review anchored at whatever HEAD was when the reply posted. So after any `tend-mention` run answers review threads inline — its normal response path — containers sit on the current HEAD, and every guard that asks "did the bot review this commit?" answers yes:

- **`LAST_REVIEW_SHA`** (step 1) advances to a commit nothing reviewed, so the next `tend-review` run computes its incremental against code it never read and skips everything before it under the trivial-change heuristic. This is #828.
- **`ALREADY_POSTED`** (step 5, *Posting mechanics*) becomes a false positive and the run exits at `Already reviewed — skipping`. Same lost-review-signal symptom, relocated to the close-out — and it costs more, because the run has already read the diff and formed the verdict before throwing it away. Fixing the derivation doesn't cover it: that routes the run *into* a full review, and this guard then discards the result.
- **`ORPHAN_ID`** (*Recovering from inline comment 422 errors*) resolves to the container, so the case-(a) recovery would `PUT` the review body over an unrelated inline reply.

Nothing goes red in any of the three; the run reports a well-reasoned decision to stay silent. On a repo where the bot both authors and reviews its own PRs, that removes the PR's only review signal. It is deterministic — no decision point — so it recurs on every reply-then-push flow.

## Solution

Count a review record only when it carries a body, owns at least one *top-level* (non-reply) inline comment, or is `APPROVED`. Reply containers have none of the three; an empty-body review carrying real inline findings still anchors, and so does an empty-body approval. `LAST_REVIEW_SHA` and `ALREADY_POSTED` use the full predicate (the latter re-derives `SUBSTANTIVE`, since shell state doesn't carry between tool calls); `ORPHAN_ID` needs only the body test, because a case-(a) orphan always persisted a body and a container never has one. The review and comment calls are `--paginate`d — container churn is exactly what pushes a PR past one page of reviews — and each reduction runs over the merged array via `jq -rs 'add | …'`, not inside `--jq`: `gh` applies `--jq` per page, so a `| last` there yields one record *per page* rather than one overall.

Where the two shapes are genuinely indistinguishable (a UI review whose only content is thread replies), the filter skips it — that moves `LAST_REVIEW_SHA` *earlier*, widening the incremental. The failure direction is more review, not less.

## Testing

No test file covers skill prompt text, so the reproduction is a live API comparison rather than a committed test.

**Step 1, against `max-sixty/cargo-affected` PR #54** — the case in #828:

| review | `commit_id` | body | comments |
|---|---|---|---|
| 4851810736 | `1ffcde2` | 2441 chars | 2 top-level |
| 4851851883 | `551e059` | 0 | 1 reply (`in_reply_to_id` 3710403279) |
| 4851852052 | `551e059` | 0 | 1 reply (`in_reply_to_id` 3710403291) |

The old recipe returns `551e059` — a commit with no review, and the one whose refactor got skipped. The new recipe returns `1ffcde2`, the real review.

**Steps 5 and the 422 recovery, against this PR's own records.** `tend-mention` answered both threads from the first review after the `8699228` push, so the containers landed on the then-current HEAD:

| record | `commit_id` | body | inline comments |
|---|---|---|---|
| 4860021420 | `b75b165` | 2286 | 2 top-level |
| 4860032792 | `8699228` | 0 | 1 reply (`in_reply_to_id` 3717103568) |
| 4860032847 | `8699228` | 0 | 1 reply (`in_reply_to_id` 3717103574) |
| 4860059883 | `8699228` | 2812 | 0 |

Evaluated against `8699228` as the reviewing run saw it — that is, before its own record 4860059883 existed — the old `ALREADY_POSTED` returns `2026-08-05T00:32:19Z` and the old `ORPHAN_ID` returns `4860032847`. Neither is a review; following the skill literally, that review would not have been posted at all. The new guards discard both containers: at `8699228` they return `2026-08-05T00:39:33Z` and `4860059883`, the one real review at that SHA, and at `7a11332` and `cdbf065` both return empty, where no review exists yet. The zero-body `APPROVED` records on #812 (`02f8dce`) and #820 (`3a5621c`, alongside two containers) both still anchor, which is what the `APPROVED` clause is for.

Other consumers of the same records were checked and left alone: `weekly` filters on `state == "APPROVED"` (reply containers are always `COMMENTED`), and `notifications` counts recent bot reviews for dedup, where a reply container legitimately means the bot responded.

---

Closes #828 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -36,9 +36,10 @@ EVENT_ACTION=$(jq -r '.action // ""' < "${GITHUB_EVENT_PATH:-/dev/null}" 2>/dev/
 # (`POST /pulls/{n}/comments/{id}/replies`) makes GitHub wrap the reply in a
 # synthetic zero-body COMMENTED review anchored at the then-current HEAD, so the
 # newest review record routinely points at a commit nothing reviewed. Count a
-# record only when it has a body or owns a top-level (non-reply) inline comment
-# — reply containers have neither, while an empty-body review carrying real
-# inline findings still anchors.
+# record only when it has a body, owns a top-level (non-reply) inline comment, or
+# is an APPROVED review — reply containers are always zero-body COMMENTED, while
+# an empty-body approval and an empty-body review carrying real inline findings
+# both still anchor.
 SUBSTANTIVE=$(gh api --paginate "repos/$REPO/pulls/<number>/comments" \
   --jq '.[] | select(.in_reply_to_id == null) | .pull_request_review_id' | jq -s 'unique')
 
@@ -48,7 +49,7 @@ SUBSTANTIVE=$(gh api --paginate "repos/$REPO/pulls/<number>/comments" \
 LAST_REVIEW_SHA=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
   | jq -rs --argjson sub "$SUBSTANTIVE" --arg bot "$BOT_LOGIN" \
     'add | [.[] | select(.user.login == $bot)
-                | select((.body | length) > 0 or (.id | IN($sub[])))]
+                | select((.body | length) > 0 or (.id | IN($sub[])) or .state == "APPROVED")]
          | last | .commit_id // empty')
 ```
 
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -32,12 +32,24 @@ PR_AUTHOR=$(gh pr view <number> --json author --jq '.author.login')
 IS_DRAFT=$(gh pr view <number> --json isDraft --jq '.isDraft')
 EVENT_ACTION=$(jq -r '.action // ""' < "${GITHUB_EVENT_PATH:-/dev/null}" 2>/dev/null)
 
-# Find the bot's most recent review (any state counts — COMMENTED reviews carry
-# inline comments even when the body is empty).
-# IMPORTANT: `gh pr view --json reviews` returns `.commit.oid` (NOT `.commit_id`).
-# The REST API (`gh api .../reviews`) uses `.commit_id` — don't confuse the two.
-LAST_REVIEW_SHA=$(gh pr view <number> --json reviews \
-  --jq "[.reviews[] | select(.author.login == \"$BOT_LOGIN\")] | last | .commit.oid // empty")
+# Find the bot's most recent *real* review. Replying to a review thread
+# (`POST /pulls/{n}/comments/{id}/replies`) makes GitHub wrap the reply in a
+# synthetic zero-body COMMENTED review anchored at the then-current HEAD, so the
+# newest review record routinely points at a commit nothing reviewed. Count a
+# record only when it has a body or owns a top-level (non-reply) inline comment
+# — reply containers have neither, while an empty-body review carrying real
+# inline findings still anchors.
+SUBSTANTIVE=$(gh api --paginate "repos/$REPO/pulls/<number>/comments" \
+  --jq '.[] | select(.in_reply_to_id == null) | .pull_request_review_id' | jq -s 'unique')
+
+# IMPORTANT: REST reviews carry `.commit_id`, NOT the `.commit.oid` that
+# `gh pr view --json reviews` returns — don't confuse the two. `gh api --jq`
+# accepts no `--arg`/`--argjson`, so pipe to `jq` rather than using `--jq`.
+LAST_REVIEW_SHA=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
+  | jq -rs --argjson sub "$SUBSTANTIVE" --arg bot "$BOT_LOGIN" \
+    'add | [.[] | select(.user.login == $bot)
+                | select((.body | length) > 0 or (.id | IN($sub[])))]
+         | last | .commit_id // empty')
 ```
 
 If `LAST_REVIEW_SHA == HEAD_SHA`, this commit has already been reviewed — exit silently. Two exceptions: an unanswered conversation question directed at the bot (check below), or `EVENT_ACTION == "ready_for_review"` (the PR just transitioned out of draft, so any prior review was a draft-mode review and the author is now asking for a full one — proceed).
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -329,8 +329,13 @@ GitHub returns `422 Unprocessable Entity` with "Line could not be resolved" when
 # The body-length test is what distinguishes an orphan from a synthetic reply
 # container sitting on the same HEAD: case (a) persists the body, a container
 # never has one. Without it, the PUT below would overwrite an unrelated reply.
+# `gh api --paginate` applies `--jq` to each page separately, so a `| last`
+# inside `--jq` returns one id *per page* — pipe to `jq -rs 'add | …'` to reduce
+# over the merged array instead.
 ORPHAN_ID=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
-  --jq "[.[] | select(.user.login == \"$BOT_LOGIN\" and .commit_id == \"$HEAD_SHA\" and (.body | length) > 0)] | last | .id // empty")
+  | jq -rs --arg bot "$BOT_LOGIN" --arg head "$HEAD_SHA" \
+    'add | [.[] | select(.user.login == $bot and .commit_id == $head and (.body | length) > 0)]
+         | last | .id // empty')
 ```
 
 Then, in either case, **move the failed inline comments into the review body** as fenced code blocks with file paths, and:
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -228,9 +228,19 @@ read -r CURRENT_HEAD PR_STATE < <(gh pr view <number> --json commits,state \
 [ "$CURRENT_HEAD" != "$HEAD_SHA" ] && echo "HEAD moved — skipping" && exit 0
 [ "$PR_STATE" != "OPEN" ] && echo "PR is $PR_STATE — skipping" && exit 0
 
+# Same substantive-review filter as step 1, and for the same reason: a synthetic
+# reply container is anchored at whatever HEAD was when the reply posted, so one
+# left on the current HEAD would read as "already reviewed" and discard this run's
+# review at the last step, after all the work. Shell state doesn't carry between
+# tool calls, so re-derive $SUBSTANTIVE here.
 # NOTE: REST API uses .commit_id (not .commit.oid from gh pr view --json)
-ALREADY_POSTED=$(gh api "repos/$REPO/pulls/<number>/reviews" \
-  --jq "[.[] | select(.user.login == \"$BOT_LOGIN\" and .commit_id == \"$HEAD_SHA\")] | last | .submitted_at // empty")
+SUBSTANTIVE=$(gh api --paginate "repos/$REPO/pulls/<number>/comments" \
+  --jq '.[] | select(.in_reply_to_id == null) | .pull_request_review_id' | jq -s 'unique')
+ALREADY_POSTED=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
+  | jq -rs --argjson sub "$SUBSTANTIVE" --arg bot "$BOT_LOGIN" --arg head "$HEAD_SHA" \
+    'add | [.[] | select(.user.login == $bot and .commit_id == $head)
+                | select((.body | length) > 0 or (.id | IN($sub[])) or .state == "APPROVED")]
+         | last | .submitted_at // empty')
 [ -n "$ALREADY_POSTED" ] && echo "Already reviewed — skipping" && exit 0
 ```
 
@@ -316,8 +326,11 @@ GitHub returns `422 Unprocessable Entity` with "Line could not be resolved" when
 **Check which case you are in before deciding how to recover** — query for an orphan review on the current HEAD first, then branch on the result.
 
 ```bash
-ORPHAN_ID=$(gh api "repos/$REPO/pulls/<number>/reviews" \
-  --jq "[.[] | select(.user.login == \"$BOT_LOGIN\" and .commit_id == \"$HEAD_SHA\")] | last | .id // empty")
+# The body-length test is what distinguishes an orphan from a synthetic reply
+# container sitting on the same HEAD: case (a) persists the body, a container
+# never has one. Without it, the PUT below would overwrite an unrelated reply.
+ORPHAN_ID=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
+  --jq "[.[] | select(.user.login == \"$BOT_LOGIN\" and .commit_id == \"$HEAD_SHA\" and (.body | length) > 0)] | last | .id // empty")
 ```
 
 Then, in either case, **move the failed inline comments into the review body** as fenced code blocks with file paths, and:


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -32,12 +32,24 @@ PR_AUTHOR=$(gh pr view <number> --json author --jq '.author.login')
...
+SUBSTANTIVE=$(gh api --paginate "repos/$REPO/pulls/<number>/comments" \
+  --jq '.[] | select(.in_reply_to_id == null) | .pull_request_review_id' | jq -s 'unique')
+
+# IMPORTANT: REST reviews carry `.commit_id`, NOT the `.commit.oid` that
+# `gh pr view --json reviews` returns — don't confuse the two. `gh api --jq`
+# accepts no `--arg`/`--argjson`, so pipe to `jq` rather than using `--jq`.
+LAST_REVIEW_SHA=$(gh api --paginate "repos/$REPO/pulls/<number>/reviews" \
+  | jq -rs --argjson sub "$SUBSTANTIVE" --arg bot "$BOT_LOGIN" \
+    'add | [.[] | select(.user.login == $bot)
+                | select((.body | length) > 0 or (.id | IN($sub[])))]
```
tend-agent: An empty-body `--approve` has no body and no top-level inline comment, so it is dropped here along with the reply containers. Reply containers are always `COMMENTED`, so an `APPROVED` clause readmits the approvals without readmitting any container:

```suggestion
                | select((.body | length) > 0 or (.id | IN($sub[])) or .state == "APPROVED")]
```

tend-agent (commented):
The diagnosis is right and the repro reproduces exactly — I re-ran the old and new recipes against `max-sixty/cargo-affected` #54 and got `551e059` vs `1ffcde2`, matching the table in the description. Two paginated calls are worth it here, and `jq -s | add` is correct against `gh api --paginate`, which merges pages into a single array (verified).

One gap: the filter's premise is that "body-empty and no top-level inline comment" uniquely identifies a reply container, but this skill's own primary success path produces exactly that shape. `gh pr review --approve -b ""` — "silence means correct" from step 5, and the trivial-skip path's "submit an APPROVE with an empty body so the PR isn't left in limbo" — writes a review with a zero-length body and no inline comments at all, so the new `select` drops it.

Across the last 25 tend PRs, 10 bot review records are `state=APPROVED body_len=0`, and every one is discarded:

| PR | records | as-written anchors at | head |
|---|---|---|---|
| #812 | one empty-body APPROVE only | *(empty)* | `02f8dce` |
| #820 | COMMENT+body, empty APPROVE, reply container | `e8d3e7c` (older) | `3a5621c` |

For #812 that isn't a widened incremental, it's the loss of the anchor entirely: `LAST_REVIEW_SHA` is empty, so `LAST_REVIEW_SHA == HEAD_SHA` never fires and a re-trigger on an already-approved commit takes the *first review* path — a full re-read of the whole PR, re-deriving findings on code that was already reviewed and approved. Step 5's `ALREADY_POSTED` check is unfiltered, so it still blocks the duplicate post at the last step, but only after the run has done all the work. Since a clean approve is the most common outcome, this is the common case rather than an edge one.

The description already supplies the fix: reply containers are always `COMMENTED` (the same fact the `weekly` audit leans on). Adding `or .state == "APPROVED"` readmits empty-body approvals and no containers — every reply container I sampled, bot- and human-authored, is `COMMENTED`. Re-running with that clause: `cargo-affected` #54 still yields `1ffcde2` (fix preserved), #820 moves to `3a5621c`, #812 to `02f8dce`, #825 unchanged.

Suggestions inline. Skipping the `/code-review` second pass — that skill isn't available yet; #819 is the PR porting it in.

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -32,12 +32,24 @@ PR_AUTHOR=$(gh pr view <number> --json author --jq '.author.login')
...
-# The REST API (`gh api .../reviews`) uses `.commit_id` — don't confuse the two.
-LAST_REVIEW_SHA=$(gh pr view <number> --json reviews \
-  --jq "[.reviews[] | select(.author.login == \"$BOT_LOGIN\")] | last | .commit.oid // empty")
+# Find the bot's most recent *real* review. Replying to a review thread
+# (`POST /pulls/{n}/comments/{id}/replies`) makes GitHub wrap the reply in a
+# synthetic zero-body COMMENTED review anchored at the then-current HEAD, so the
+# newest review record routinely points at a commit nothing reviewed. Count a
+# record only when it has a body or owns a top-level (non-reply) inline comment
+# — reply containers have neither, while an empty-body review carrying real
+# inline findings still anchors.
```
tend-agent: If the `APPROVED` clause goes in, this comment needs to say so — as written it asserts the two-part test is exhaustive, which is the assumption that misses the approve path.

```suggestion
# record only when it has a body, owns a top-level (non-reply) inline comment, or
# is an APPROVED review — reply containers are always zero-body COMMENTED, while
# an empty-body approval and an empty-body review carrying real inline findings
# both still anchor.
```

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Applied in 8699228. Confirmed the regression on live data before taking it: tend #812 carries only `APPROVED body=0`, so as-written it lost the anchor entirely, not just widened the incremental. With the clause, #812 and #820 both anchor at head, and `cargo-affected` #54 still yields `1ffcde2`.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Applied in 8699228, so the comment now names all three admitting conditions and states the reply-container invariant (always zero-body `COMMENTED`) that makes the `APPROVED` clause safe.

tend-agent (commented):
The `APPROVED` clause is right and the filter now behaves on every shape I could find: across the last 25 PRs here, every zero-body non-`APPROVED` bot review record is a reply container (no top-level inline comment), so the three-way predicate discards containers and nothing else. Re-ran it on this PR as a self-test — `SUBSTANTIVE` resolves to `[4860021420]` and `LAST_REVIEW_SHA` to `b75b165`, not the `8699228` the old recipe would have returned.

The gap is that the same defect survives, unfixed, in the two step-5 guards under **Posting mechanics** and **Recovering from inline comment 422 errors**. Both select on `.commit_id == "$HEAD_SHA"` with no substantive filter, so a reply container anchored at the current HEAD reads as a review there too:

- `ALREADY_POSTED` becomes a false positive and the run exits at `Already reviewed — skipping`. That is the same lost-review-signal symptom #828 describes, just relocated from the pre-flight to the close-out — and it costs more, because the run has already read the diff and formed the verdict before throwing it away. Step 1's fix does not cover it: fixing the derivation routes the run *into* a full review, and this guard then discards the result.
- `ORPHAN_ID` resolves to the container, so the case-(a) recovery would `PUT` the review body over an unrelated inline reply.

This run reproduced it live rather than hypothetically. `tend-mention` answered both threads from the previous review at 00:32:18/19, after the 00:32:05 push, so the containers landed on the current HEAD:

| record | `commit_id` | body | inline comments |
|---|---|---|---|
| 4860021420 | `b75b165` | 2286 | 2 top-level |
| 4860032792 | `8699228` | 0 | 1 reply (`in_reply_to_id` 3717103568) |
| 4860032847 | `8699228` | 0 | 1 reply (`in_reply_to_id` 3717103574) |

Run as written against `8699228`, `ALREADY_POSTED` returns `2026-08-05T00:32:19Z` and `ORPHAN_ID` returns `4860032847`. Neither is a review. Following the skill literally, this review would not exist.

Pushed the fix in the same commit-scope rather than opening a follow-up, since it is the same root cause in the same file and the PR only half-closes #828 without it. `ALREADY_POSTED` reuses step 1's predicate verbatim (re-deriving `SUBSTANTIVE`, since shell state does not carry between tool calls); `ORPHAN_ID` needs only the body test, because a case-(a) orphan always persisted a body and a container never has one. Verified against four shapes: containers-only on this PR at `8699228` returns empty, the real review at `b75b165` returns its timestamp, and the empty-body approvals on #812 (`02f8dce`) and #820 (`3a5621c`, alongside a container) both still return theirs.

The push cancels this session — `tend-review` is `cancel-in-progress: true` on a per-PR group — so that commit ships CI-unverified.

tend-agent:
Both step-5 guards are fixed in `7a11332`, plus `cdbf065` for a follow-on the same change introduced: `gh api --paginate` applies `--jq` per page, so the `| last` left inside `ORPHAN_ID`'s `--jq` returned one id per page rather than one overall. That reduction now runs over the merged array like the other two.

The CI-unverified caveat is closed — [run 30964152146](https://github.com/max-sixty/tend/actions/runs/30964152146) is green on `cdbf065` (`test`, `test-proxy`, `test-worker`, `lint`). Re-derived all three guards against this PR's own records as an independent check:

| `HEAD_SHA` | `LAST_REVIEW_SHA` anchor | `ALREADY_POSTED` | `ORPHAN_ID` |
|---|---|---|---|
| `b75b165` | — | `00:29:05Z` | 4860021420 |
| `8699228` | `8699228` (newest substantive) | `00:39:33Z` | 4860059883 |
| `cdbf065` | — | empty | empty |

The two containers at `8699228` (4860032792, 4860032847) are discarded at every call site; the zero-body `APPROVED` records on #812 and #820 still anchor.

Description updated — it described only the first commit, and the PR now covers all three record consumers.

tend-agent (commented):
No findings in the diff — I re-derived all three guards independently and got the same anchors, and `--slurp` (the obvious alternative to piping out) turns out to be unusable here anyway: `gh api` rejects it with `the --slurp option is not supported with --jq or --template`, so `| jq -rs 'add | …'` is the only shape available for a cross-page reduction.

The same-pattern sweep did turn up one live instance outside the diff, in the generated mention workflow rather than a skill. [`generator/src/tend/templates/mention.yaml.j2`](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/mention.yaml.j2#L278) reduces inside `--jq` under `--paginate` in three places — both `BOT_COMMENTS=` assignments and `BOT_REVIEWS=`, each `--jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length'`. That is the shape `cdbf065` just fixed in `ORPHAN_ID`: `--jq` is applied per page, so the reduction returns one count per page rather than one overall.

`gh api --paginate` sets `per_page=100`, so it takes >100 reviews or >100 issue comments on a single thread to cross. Past that the variable holds e.g. `100\n7`, and the guard immediately below it — `if [ "$BOT_COMMENTS" -gt "0" ]` — errors with `integer expression expected` and returns 2 instead of matching. The `if` swallows the failure, so the step falls through to `should_run=false`: the bot stops answering participants on exactly the threads where it has engaged the most, silently and without going red. These render into every adopter's `tend-mention.yaml`, not just ours.

Cheaper to fix than this PR's guards, since none of the three needs `--argjson` — moving the count outside jq keeps the filter a per-element stream, which paginates fine:

```bash
  --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id' | wc -l)
```

Separate file and separate concern from the reply-container fix, so I'd rather not fold it in here — happy to open it as its own PR if you want it.

Per-page application, forced with `per_page=1` on this PR's 4 review records:

```
$ gh api --paginate "repos/max-sixty/tend/pulls/835/reviews?per_page=1" --jq '[.[]] | last | .id'
4860021420
4860032792
4860032847
4860059883
```

The stream form above returns `4` on the same input. Default page size confirmed at 100 (`gh api --paginate repos/max-sixty/tend/commits --jq 'length'` → `100` per page).

Fall-through, reproduced directly:

```
$ BOT_REVIEWS=$(printf '30\n12\n'); if [ "$BOT_REVIEWS" -gt "0" ]; then echo run; else echo "rc=$?"; fi
bash: [: 30
12: integer expression expected
rc=2
```

Live in the generated workflow at `.github/workflows/tend-mention.yaml` lines 264, 283, 290. Template call sites: [`BOT_COMMENTS` (issue path)](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/mention.yaml.j2#L278), [`BOT_REVIEWS`](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/mention.yaml.j2#L297), [`BOT_COMMENTS` (PR path)](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/mention.yaml.j2#L304). The fourth `--paginate` in that step, the review-comment body scan piped to `grep -qF`, is a per-element stream and is unaffected.

</details>

tend-agent:
Opened the out-of-diff finding as #840 rather than leaving the offer hanging — same shape (`--jq` reducing under `--paginate`), but in `generator/src/tend/templates/mention.yaml.j2`'s three engagement counts, where the multi-line value makes the `-gt` guard error and the step fall through to `should_run=false`. Nothing in this PR's diff changes as a result; #835 stands as-is.

Issue #831, opened 2026-08-04
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T08:52:50Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30893745535) | #827 |

This issue was created automatically. Close it once the outage is resolved.

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T08:53:16Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30893786184) | #829 |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T08:53:26Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30893788054) | #830 |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T09:43:04Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30897445507) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T09:43:14Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30897445507) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T09:43:12Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30897445507) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T09:43:22Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30897445507) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T09:44:03Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30897445507) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T10:43:08Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30901730005) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
[...]

PR #836, opened 2026-08-05
fix(report-failure): reconcile duplicate outage issues by primary-key probe, not a lagging list

### LINKED ISSUE #832: Bot temporarily unavailable

| When | Run | Trigger |
|------|-----|---------|
| 2026-08-04T08:53:09Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/30893784998) | #828 |

tend-agent:
## Problem

`max-sixty/worktrunk` currently has **two open `tend-outage` trackers**, [#3737](https://github.com/max-sixty/worktrunk/issues/3737) and [#3738](https://github.com/max-sixty/worktrunk/issues/3738), both created at exactly `2026-08-04T08:53:15Z`. The self-heal reconcile added in #744 was supposed to close one of them. It closed neither, and 26 failure rows have since scattered across both (15 comments on #3737, 11 on #3738) as each new failure appends to whichever the list happens to return first.

That breaks the drain flow the adopter's `running-tend` overlay depends on: the sweep closes the tracker once every row is drained, and "one left open makes tomorrow's sweep re-check the same rows and folds the next outage into a stale incident". With two open, no sweep ever sees a complete row set.

## Root cause

The reconcile settles for 5 seconds and then re-reads `gh issue list --label tend-outage`. That is the *same lagging read path* that lost the create-create race in the first place, so when both creates land close enough together, every leg reads back only its own issue, computes itself as the keeper, and closes nothing.

The two repos that failed in the same outage window bracket the lag precisely:

| Repo | Creates | Reconcile read | Outcome |
|---|---|---|---|
| `max-sixty/tend` | [#831](https://github.com/max-sixty/tend/issues/831) @ `08:53:15`, [#832](https://github.com/max-sixty/tend/issues/832) @ `08:53:17` | #832's leg saw a 6s-old sibling | #832 closed as duplicate — worked |
| `max-sixty/worktrunk` | #3737 and #3738 both @ `08:53:15` | each leg saw only a ~5s-old sibling | nothing closed — **failed** |

Note that #831's own leg also failed to close #832 (3s old at its read), so even the "successful" repo only half-reconciled. The run logs confirm it directly — both worktrunk legs' `Report failure` step created its issue and ended 5.4s later having emitted no close:

```
review       08:53:15.47  https://github.com/max-sixty/worktrunk/issues/3737   (step ends 08:53:20.92)
review-runs  08:53:15.55  https://github.com/max-sixty/worktrunk/issues/3738   (step ends 08:53:20.91)
```

Measured directly on `tend-agent/tend-integration`: immediately after two back-to-back creates, `gh issue list --label` returned `[]` — *neither* issue visible — while `GET /issues/{n}` returned each one instantly. The list caught up by +3s.

```
t=+0s  gh issue list -> []          |   GET /issues/34 -> 34
t=+3s  gh issue list -> [34,35]     |   GET /issues/34 -> 34
```

## Fix

Probe issue numbers directly instead of re-listing. `GET /issues/{n}` is a primary-key read and returns a sibling the instant it exists, so no settle is needed at all — the `sleep 5` goes away.

Issue numbers are monotonic, so any racing sibling sits just below ours: scan a short window downwards and defer to the lowest open tracker found. Convergent — every leg computes the same keeper from its own vantage point, and only higher-numbered legs close themselves.

The closing leg now also carries its row onto the keeper before closing. Previously the duplicate's seed row was stranded in the body of a closed issue; a failure that got recorded should stay visible on the tracker that survives.

## Testing

The branch now merges `main`, which brought `generator/tests/test_shared_steps.py` — so the reconcile is covered by tests rather than only by the live runs below. Seven tests over `report-failure.sh` and the shared helper: stand-down to a racing sibling, the row carried onto the keeper (with a second probe further down the window, so "lowest match" is asserted rather than merely consistent with the fixture), the same-run guard suppressing a duplicate row, a foreign issue not adopted, create-failure propagation, and the same carry-over on the `rate-limit-preflight.sh` caller. Both carry-over assertions were mutation-checked — reverting the fix fails them. Full suite 359 passed; `shellcheck` clean.

The live-GitHub evidence below predates those tests and is kept because it is what identified the mechanism.

Review fixes in 1c41d39 were checked the same way, with `gh` mocked. Create-failure propagation, against the shipped code on this branch and the fix:

```
old (basename wrapper):  gh: API error / Filed outage issue # / EXIT=0
new (own assignment):    gh: API error / EXIT=1
```

Carry-over guard, both branches:

```
keeper already cites this run (same matrix)   -> CLOSED 40                      (no carry-over)
keeper cites a different run (cross-workflow) -> CARRIED_OVER to 39 / CLOSED 40
```

**1. The probe resolves the live production pair** the shipped reconcile failed on — read-only, against the actual open duplicates:

```
max-sixty/worktrunk MINE=#3738 -> KEEP=3737
max-sixty/worktrunk MINE=#3737 -> KEEP=<none, survives>
max-sixty/tend      MINE=#831  -> KEEP=<none, survives>
```

**2. End-to-end race** on `tend-agent/tend-integration`: the real script run as two concurrent legs, with only the jitter `sleep` removed (the jitter merely narrows the window this test targets).

Treatment (this branch) — one survivor, both rows on it:

```
#38 OPEN     body: run 111111 row   +   comment: run 222222 row   (carried over)
#39 CLOSED   "Duplicate of #38 (concurrent leg failure); consolidating outage tracking there."
```

Control (`origin/main`) reconciled correctly on its run — which is the point: the control is timing-dependent by construction, passing or failing on the lag draw, and production drew the failing side. The treatment does not depend on the draw. All probe artifacts on the integration repo were closed afterward; it has no open issues.

## Trade-off: the reconcile is now self-close-only

Worth stating explicitly, since it is a real narrowing rather than a pure win. The old block closed *every* higher duplicate it could see; this one only ever closes the leg's own issue. So a leg that dies between `gh issue create` and its probe — a cancelled job, or all ten `gh api` probes erroring, which `2>/dev/null || true` renders indistinguishable from "no sibling" — leaves an open tracker that no other leg will close. Nothing else in the repo reconciles trackers; `enrich-tend-outage-issues.sh` only enriches open ones.

Taking it anyway, for two reasons. The capability being given up only ever worked when the lagging list happened to be current, which is precisely the condition that failed in production — so it was never dependable cover for that window. And the obvious restoration, scanning upwards as well as downwards, does not converge: numbers above ours may not exist yet at probe time, so two legs can reach different verdicts about who closes whom, and a leg could close a sibling mid-carry-over and strand the row it was moving.

The residual window is strictly narrower than the one this replaces — a leg dying inside a few-second gap, versus every same-second create race losing — and it fails toward an extra open tracker rather than scattered rows. A durable fix belongs in the `EXISTING` branch, where a later failure reads a list that has long since caught up and can close any stragglers; that is #809's half of this script, so it is left for whichever of the two lands second rather than built speculatively here.

## Gate assessment

- **Class**: structural. A shell script whose reconcile reads an eventually-consistent index; no model decision point. Replayed 10/10 times with creates in the same second, the reconcile is a no-op.
- **Evidence**: High. The underlying create-create race is well-attested and recurring — multi-issue outage bursts across repos before the mitigation (tend #160/#161/#162, #575–#578, #742/#743; worktrunk #2069/#2070) and one after it (worktrunk #3737/#3738), plus a same-window near-miss on tend where #831's leg also closed nothing. What is new is that the shipped mitigation covers only part of the timing window, demonstrated with a measured mechanism rather than inferred.
- **Magnitude**: targeted fix replacing one block in one shared script — Normal bar, met.
- **Passes both gates.**

Scope: textually disjoint from the other open changes on this path. #823 fixes what a row *points at* (the REF block, lines 19–29); #809 dedups rows across matrix legs (the `EXISTING` append branch). This one fixes the *create* branch's reconcile. Note that #809's write-up states the create-create race "was solved" by #744 — this is the evidence that it was solved only for the wider-gap half of the window.

The outage that produced these failures is not itself a code fix: the session logs show `You've hit your weekly limit · resets 12am (UTC)` (`error: rate_limit`, HTTP 429) on every failed run, which is quota exhaustion, not a bug.

<details><summary>Window &amp; evidence</summary>

- Analysis run: [30962483562](https://github.com/max-sixty/tend/actions/runs/30962483562), target `max-sixty/worktrunk`. The window stretches back to the last successful `review-reviewers` run at 08:43Z — the weekly-quota exhaustion killed the analysis runs too, so there is a ~15.5h coverage gap that this run absorbs.
- Evidence log: https://gist.github.com/a88c03f4d0c3fb1791060ff3dd97d1c4
- Runs in window: 3 × `tend-notifications` (all `failure`, all quota-exhausted) and 3 × `tend-mention` (all jobs `skipped` — the documented bot's-own-comment verify gate, firing on the outage issues' own comments). No other bot output existed this window; corruption scan over the auto-generated outage rows was clean.

</details>


shared/steps/report-failure.sh
```diff
@@ -56,22 +56,43 @@ else
     "The bot failed to process a request. This issue tracks failures until the underlying cause is resolved." \
     "$TABLE" \
     "This issue was created automatically. Close it once the outage is resolved." > /tmp/body.md
-  gh issue create --title "$TITLE" --label "$LABEL" -F /tmp/body.md
+  MINE=$(basename "$(gh issue create --title "$TITLE" --label "$LABEL" -F /tmp/body.md)")
```
tend-agent: Wrapping the create in `basename` makes a failed `gh issue create` silent. `basename ""` exits 0, so the assignment's status is `basename`'s, not `gh`'s, and `set -e` no longer fires — the step logs `Filed outage issue #`, the reconcile loop breaks immediately (`$((MINE - 1))` on an empty `MINE` is `-1`), and the job goes green with the outage unrecorded. Before this change `gh issue create` was a bare command, so a create failure turned the step red.

Capturing the URL in its own assignment keeps `set -e` on the `gh` call:

```suggestion
  MINE_URL=$(gh issue create --title "$TITLE" --label "$LABEL" -F /tmp/body.md)
  MINE=${MINE_URL##*/}
```

shared/steps/report-failure.sh
```diff
@@ -56,22 +56,43 @@ else
...
+            and ([.labels[].name] | index(\"${LABEL}\"))) | .number" 2>/dev/null || true)
+    if [ -n "$MATCH" ]; then KEEP="$MATCH"; fi
   done
+
+  if [ -n "$KEEP" ]; then
+    # Carry this leg's row over before closing, so the failure it recorded
+    # stays on the surviving tracker instead of being stranded in the body of
+    # a closed duplicate.
+    printf '%s\n' "$TABLE" > /tmp/comment.md
+    gh issue comment "$KEEP" -F /tmp/comment.md || true
```
tend-agent: When the racing legs belong to one matrix run, the carried-over row duplicates a row already on the keeper. Matrix legs share `GITHUB_RUN_ID`, so `RUN_URL` — and `REF`, from the same event payload — are identical across legs; only `TIMESTAMP` differs. `review-reviewers.yaml`'s per-repo matrix is exactly that shape, and it's the case the jitter comment above describes ("a matrix workflow's legs fail at near-identical times"). The keeper then carries two rows pointing at the same run, which is the flood #809 is removing from the `$EXISTING` branch — this path would bypass that guard.

The cross-workflow case the PR body documents (worktrunk's `review` and `review-runs`, different run ids) still carries over correctly under a run-scoped guard:

```suggestion
    printf '%s\n' "$TABLE" > /tmp/comment.md
    if ! gh issue view "$KEEP" --json body,comments \
        --jq '.body + "\n" + ([.comments[].body] | join("\n"))' \
        | grep -qF "[workflow run](${RUN_URL})"; then
      gh issue comment "$KEEP" -F /tmp/comment.md || true
    fi
```

If #809 lands first its `$EXISTING`-branch check is the same four lines, so a small shared helper would beat two copies.

tend-agent (commented):
The probe direction is right and the convergence argument holds — scanning downward and keeping the *lowest* match (not the nearest) is what makes two legs with overlapping windows agree even when a third, older tracker sits below both. Two things on the diff, inline.

One design note worth a second opinion rather than a change: reconcile is now self-close-only, where the old block closed every higher duplicate it could see. A leg that dies between `gh issue create` and its probe — cancelled job, or all ten `gh api` probes erroring, which `2>/dev/null || true` renders indistinguishable from "no sibling" — leaves an open tracker that no other leg will ever close, and nothing else in the repo reconciles trackers (`enrich-tend-outage-issues.sh` only enriches open ones). That's a much narrower window than the one this PR closes, so it may well be the right trade; it just isn't mentioned in the write-up.

shared/steps/report-failure.sh
tend-agent: Confirmed and fixed in 1c41d39 — took the suggestion as written.

Reproduced both directions against the shipped code on this branch, with `gh issue create` mocked to fail:

```
old (basename wrapper):  gh: API error / Filed outage issue # / EXIT=0
new (own assignment):    gh: API error / EXIT=1
```

So it was exactly as you describe: green step, empty `$MINE`, `seq -1 -1 -10` breaks on the first `[ "$n" -gt 0 ]`, and the outage goes unrecorded. Worth noting the create was the one `gh` call in this script that had no `|| true` on it precisely so a create failure would surface — the wrapper silently gave that up.

shared/steps/report-failure.sh
tend-agent: Right — fixed in 1c41d39 with the guard as suggested. `review-reviewers.yaml`'s matrix is 5 repos on one `GITHUB_RUN_ID`, and its trigger is `schedule`, so `REF` is `N/A` on every leg too: a carried row would have differed from the keeper's seed row in `TIMESTAMP` alone.

Verified both branches with a mocked `gh`:

```
keeper already cites this run (same matrix)   -> CLOSED 40                      (no carry-over)
keeper cites a different run (cross-workflow) -> CARRIED_OVER to 39 / CLOSED 40
```

so the worktrunk `review`/`review-runs` case in the write-up — distinct run ids — still carries over, which was the point of adding it.

On the shared helper: leaving the four lines duplicated for now rather than pre-building a helper this branch can't share. #809 is still open and the two diffs are textually disjoint as they stand; introducing a helper here would put them in conflict for no benefit until one lands. Whichever lands second folds both copies into one — noted on #809 as well so it doesn't get lost.

shared/steps/lib/run-issue.sh
```diff
@@ -100,37 +100,100 @@ run_issue_ensure_label() {
...
+# fix it: higher numbers may not exist yet at probe time, so it would not
+# converge. The residual window is strictly narrower than the one it replaces,
+# and it fails toward an extra open record rather than scattered rows.
 #
 # Everything but the number goes to stderr, so the log keeps the created URL
 # while the caller can read the keeper straight out of stdout.
 run_issue_create_and_reconcile() {
-  local label=$1 title=$2
-  gh issue create --title "$title" --label "$label" -F - >&2
+  local label=$1 title=$2 row=${3:-}
```
tend-agent: The carry-over is opt-in through this third argument, and the helper's other caller doesn't pass it. `rate-limit-preflight.sh` — which arrived on `main` when this helper was extracted, so this branch never saw it — still calls `run_issue_create_and_reconcile "$PAUSE_LABEL" "$PAUSE_TITLE"` with two arguments, so on that path standing down strands exactly the row the argument exists to rescue.

It's worse there than on the outage path, because on the rate-limit record the row *is* the notice. Verified against the merged tree with this suite's own fakes — one racing sibling at #41, this leg files #42:

```
issue close 42 --comment Duplicate of #41 (concurrent run); consolidating tracking there.
::error::Rate limit: bot created 16 items today, above the ceiling of 15 ... Refused runs are listed in #41; closing it doubles the ceiling.
```

No `issue comment 41` anywhere in the call log. The refused run's row went into #42's body at creation and was closed with it, while the annotation sends the maintainer to #41 to find it — and #41 is the issue whose close doubles the ceiling, so the one artifact a person is asked to act on is the one missing the evidence.

`$ROW` is already in scope at that call site (assigned just above the reopen-or-append branch), and the `EXISTING` branch a few lines up already appends it as a comment for every later refused run — so the create-race path is the only one that drops a row. The same-run guard added here makes passing it safe: legs of one matrix share `GITHUB_RUN_ID`, so the keeper's seed row suppresses the duplicate.

Outside this diff, so no inline suggestion — happy to push the one-argument change plus a test asserting the carry-over comment lands on the keeper, if you'd rather it ride along here than wait for a follow-up.

tend-agent (commented):
The probe ports onto `main`'s extracted `lib/run-issue.sh` cleanly, and the convergence argument survives the move intact. One gap the merge opened, inline: the helper picked up a second caller on `main`, and the row carry-over never reached it.

Separately on the new tests — they're a real improvement over "exercised against real GitHub", but every stand-down fixture has exactly one racing sibling, and with one sibling "lowest match" and "nearest match" give the same answer. The property the whole design rests on (`keep` is overwritten on each descending hit, so the lowest wins) is what makes two legs with overlapping probe windows agree on the same keeper, and nothing pins it. A second `_outage_probe` further down the window in `test_report_failure_carries_its_row_onto_the_racing_sibling` would turn that from an argument in the comment block into a test.

tend-agent:
Merged `main` and resolved the conflict in `shared/steps/report-failure.sh`.

**What conflicted.** `main` moved this script's create-and-reconcile block into the shared `lib/run-issue.sh` (#823, #874), where `rate-limit-preflight.sh` now shares it. The conflicting hunk was exactly the block this PR rewrites, against the block that no longer lives here.

**How it resolved.** Not superseded — `run_issue_create_and_reconcile` still does `sleep 5` then re-reads `run_issue_matching`, which is the lagging-list read this PR's evidence shows losing the same-second race. So the probe reconcile moved into the lib alongside it, and `report-failure.sh` keeps `main`'s shape with one addition: it passes `$ROW` as the new optional third argument, which is what drives the carry-over.

Two things changed from the branch's version in the process:

- **The probe carries `main`'s author constraint.** `run_issue_matching` gained `--author @me` and a title match after this branch forked, on the argument that the bot holds `issues: write` so the label alone nominates nothing. The probe applies all three per issue (`.user.login`, `.title`, label) rather than dropping back to title+label; failing to read our own identity leaves the pair open rather than deferring to an issue we can't vouch for.
- **The carry-over guard no longer pipes into `grep`.** `gh issue view | grep -qF` under `pipefail` reports a *match* as a pipeline failure when `grep -q` exits early and `gh` takes SIGPIPE — inverting the guard on exactly the long issues where it matters. Read into a variable, matched with a herestring.

**Tests.** `main` also added `generator/tests/test_shared_steps.py` after this branch forked, so the "no shell test harness" note in the description is now stale. `test_rate_limit_reconciler_keeps_only_what_the_preflight_filed` asserted on the list call this PR removes; it now asserts on the probe, with four siblings inside the window each failing one conjunct of the predicate. Added six tests over `report-failure.sh` covering the reconcile itself: stand-down to a racing sibling, the row carried onto the keeper, the same-run guard suppressing it, a foreign issue not adopted, and create-failure propagation.

Full suite green (358 passed), shellcheck clean, CI green.

**One thing left alone, worth its own issue.** `rate-limit-preflight.sh` calls this function as `PAUSE=$(... | run_issue_create_and_reconcile ...)`. Bash clears `set -e` inside a command substitution unless `inherit_errexit` is set, so a failed `gh issue create` there does *not* fail the step — it just leaves `PAUSE` empty and the `::error::` printing `#?`. That is pre-existing on `main` and orthogonal to this PR's create branch, so it is untouched here. `report-failure.sh` is not affected: it calls the function as the last element of a pipeline, where `-e` does apply (asserted by the new test).

shared/steps/lib/run-issue.sh
tend-agent: Right, and worth having ride along here rather than as a follow-up — the gap only exists because the extraction gave the helper a second caller mid-PR, so splitting it out would leave `main` carrying a strand the same change introduced. Fixed in 6c338ba: `rate-limit-preflight.sh` now passes `$ROW` as the third argument, with a comment at the call site naming why the same row goes in twice (seeds the body, and carries over when this leg loses the race).

Test added alongside it — `test_rate_limit_carries_its_row_onto_the_racing_sibling`, the rate-limit twin of the outage one: one racing sibling at #41 whose seed row cites a different run, this leg files #42, and it asserts the comment lands on #41 with this run's link in the body before #42 closes. That needed the rate-limit fake to capture comment bodies (it was dropping them, since `-F -` puts them on stdin rather than in the args), so `_comments` and `RUN_LINK` moved up next to `_calls` and both suites share them.

Mutation-checked rather than just observed green — dropping the third argument again fails it on exactly the shape you described, with `issue close 42` present and no `issue comment 41` anywhere in the call log.

The command-substitution `set -e` hole you flagged in the merge notes is still untouched and still pre-existing: `PAUSE=$(... | run_issue_create_and_reconcile ...)` swallows a failed create there regardless of this change. Worth its own issue, as you said.

tend-agent:
Both points addressed in 6c338ba; the inline thread covers the `rate-limit-preflight.sh` carry-over.

On the untested property — agreed, and it was the weakest part of the new suite: with one sibling the fixture passes under either rule, so the descent past the first hit was asserted nowhere. `test_report_failure_carries_its_row_onto_the_racing_sibling` now seeds two probes, #41 and #38, against this leg's #42. Lowest wins, so the row goes to #38 and #42 closes as its duplicate; the test asserts `issue comment 38` **and** the absence of `issue comment 41`, which is what distinguishes the two rules rather than merely being consistent with both.

Confirmed it bites: adding a `break` after the first match in the probe loop — the exact "nearest" implementation — fails it with the row carried onto #41.

That is the property a third leg depends on. A leg filing #43 sees both #41 and #38 in its window while the #42 leg sees only #38 below it; if either stopped at its first hit they would pick different keepers and the rows would scatter across two trackers, which is the failure this PR exists to remove. It reads as an argument in the comment block above the function; now it fails a test if it stops being true.

Both new assertions mutation-checked against the shipped code rather than just observed green:

| Mutation | Test | Result |
|---|---|---|
| drop `"$ROW"` from the `rate-limit-preflight.sh` call | `test_rate_limit_carries_its_row_onto_the_racing_sibling` | fails — `issue close 42` with no `issue comment 41` |
| `break` after the first probe match | `test_report_failure_carries_its_row_onto_the_racing_sibling` | fails — row carried onto #41, not #38 |

Full suite 359 passed, `pre-commit run --files` clean over the three changed files (shellcheck, ruff, typos).

</details>

shared/steps/lib/run-issue.sh
```diff
@@ -100,44 +100,105 @@ run_issue_ensure_label() {
...
 run_issue_create_and_reconcile() {
-  local label=$1 title=$2
-  # Carry a failed create out as this function's status. `set -e` does not
-  # reach inside a command substitution unless `inherit_errexit` is set, and
-  # every caller reads the number back through one — so without this the
-  # failure runs on to the `printf` at the end, whose success becomes the
-  # substitution's status, and the caller is handed an empty number it takes
-  # for a filed issue.
-  gh issue create --title "$title" --label "$label" -F - >&2 || return 1
+  local label=$1 title=$2 row=${3:-}
```
tend-agent: 6c338ba makes `rate-limit-preflight.sh` the second caller to pass `$ROW`, so nothing reaches the two-argument form any more — `${3:-}` here and the `if [ -n "$row" ]` guard further down are branches with no call site and no test behind them. That is the shape CLAUDE.md rules out ("cut over completely — old formats should fail with a clear error, not silently parse"): a caller that forgets the row should redden, not silently strand the row it was meant to carry, which is the bug this very commit fixed.

Dropping the guard on its own would be worse than leaving it. There is no `set -u` here, so a missing third argument would reach `printf '%s\n' "$row" | gh issue comment` and post an empty comment onto the keeper. The required form has to say so itself:

```suggestion
  local label=$1 title=$2 row=${3:?run_issue_create_and_reconcile: this run's row is required}
```

Checked both call shapes with a stand-in function under `set -eo pipefail`: `report-failure.sh`'s bare pipeline aborts the step, and `rate-limit-preflight.sh`'s `if ! PAUSE=$(... | ...)` catches it as a failed create. With that in place the `if [ -n "$row" ]` block can lose its condition and dedent.

tend-agent (commented):
The incremental since the last review is the `$ROW` third argument on `rate-limit-preflight.sh`'s call plus the two-probe fixture, and both land what they were asked for. Asserting the *absence* of `issue comment 41` is what makes the fixture bite — a single-sibling one passes under either rule, so "lowest, not nearest" only becomes a test once there is a second probe below the first.

One consequence the commit creates, inline: with both call sites now passing a row, the argument's optionality has no caller left.

shared/steps/lib/run-issue.sh
tend-agent: Agreed, and taken in 2293256 — `${3:?}` on the argument, guard dropped, block dedented. With both callers passing a row the two-argument form was a branch no caller reached and no test covered, and the failure it silently permitted is the one the carry-over exists to prevent: file an issue, stand down to the keeper, close the duplicate with the row still in it.

One correction to the suggestion, though — as written it doesn't parse. The apostrophe in `run's` is inside the *word* of a `${3:?word}`, which bash parses for quotes, so the lone `'` opens a string that runs to EOF and takes the whole file down rather than just that call:

```
$ bash -n shared/steps/lib/run-issue.sh
shared/steps/lib/run-issue.sh: line 203: unexpected EOF while looking for matching `''
```

That's every consumer's `report-failure` and `rate-limit-preflight` step failing to source the lib at all, not a reworded error. Landed the message apostrophe-free (`the row for this run is required`) with a comment saying why, since the next edit to that string is the obvious place to reintroduce it. `bash -n` and shellcheck both pass now; shellcheck would have caught it in pre-commit, but it's worth the comment given how invisible the cause is at the call site.

Confirmed your read of both call shapes against the real `set -eo pipefail` (neither script sets `-u`): `report-failure.sh`'s bare pipeline aborts the step, and `rate-limit-preflight.sh`'s `if ! PAUSE=$(... | ...)` catches it as a failed create and carries on to the `FILED=false` path.

Also locked the required-ness with `test_run_issue_reconcile_refuses_a_call_with_no_row`, which asserts the abort lands *before* the create — a version that filed first and only then refused would still strand the record. Full suite green (363).

Both call shapes, with a stand-in under the callers' actual flags:

```
$ bash -c 'set -eo pipefail
f() { local a=$1 b=$2 r=${3:?the row for this run is required}; echo ran; }
echo body | f L T > /dev/null
echo "NOT REACHED"'
environment: line 3: 3: the row for this run is required
# exit=1, "NOT REACHED" never printed

$ bash -c 'set -eo pipefail
f() { local a=$1 b=$2 r=${3:?the row for this run is required}; echo ran; }
if ! P=$(echo body | f L T); then echo "caught as failed create"; P=""; fi
echo "continued, P=[$P]"'
environment: line 3: 3: the row for this run is required
caught as failed create
continued, P=[]
```

Call sites, both passing a row: [`report-failure.sh:46`](https://github.com/max-sixty/tend/blob/2293256567c54cdb9bad6d50b45deda16ca0b08c/shared/steps/report-failure.sh#L46), [`rate-limit-preflight.sh:207`](https://github.com/max-sixty/tend/blob/2293256567c54cdb9bad6d50b45deda16ca0b08c/shared/steps/rate-limit-preflight.sh#L207). The required argument is at [`run-issue.sh:146`](https://github.com/max-sixty/tend/blob/2293256567c54cdb9bad6d50b45deda16ca0b08c/shared/steps/lib/run-issue.sh#L146).

`uvx ruff@0.14.11` (the version `.pre-commit-config.yaml` pins) reports clean on the test file; `ruff@latest` flags `PLW1510` on the pre-existing `_run*` helpers under its default ruleset, which the repo's config doesn't select.

</details>

PR #837, opened 2026-08-05
fix(running-in-ci): pin the CI-monitor poll to the SHA the run pushed

### LINKED ISSUE #829: running-in-ci: CI-monitor poll follows the PR head, so a concurrent push makes a run report another commit's green as its own
## Problem

The CI Monitoring recipe in `running-in-ci` gates on the PR's rollup:

```bash
pending() {
  gh pr view <number> --json statusCheckRollup | jq ... 
}
```

`statusCheckRollup` describes the PR's **current head**, not the commit the session pushed. If any other actor pushes while the loop is sleeping, the gate silently retargets: the loop starts waiting on the new commit's checks, exits when *those* settle, and the final `gh pr checks <number>` prints the new commit's results. The session then reports that its own push is green without ever having observed a terminal state for it.

The loop is also blind to a failure that lands early. `pending()` counts only in-flight checks; a check that reaches `failure` while siblings are still running does not break the loop, and by the time the loop exits the evidence may belong to a different commit.

## Observed

- 07:45:42 — [run 30888815177](https://github.com/max-sixty/cargo-affected/actions/runs/30888815177) (`tend-mention`) pushes `551e059`, posts a comment, starts the poll.
- 07:47:29 — `lint` fails on `551e059` (`cargo fmt --check`). All other checks pass; the `review` check-run stays in-flight until 07:51:28, so `pending() > 0` and the loop keeps sleeping.
- 07:51:09 — a concurrent `tend-review` session pushes `6955352` (the rustfmt fix) to the same branch. The PR head moves; `statusCheckRollup` now describes `6955352`.
- ~07:53 — the loop sees `pending() == 0` and prints `gh pr checks 54`, which returns six green rows sourced from run `30889480225` — the CI run for `6955352`.

The session's closing report:

> CI on the pushed commit is green on all three platforms plus lint and msrv.

`551e059` was red on `lint`. A sibling session ([30889231291](https://github.com/max-sixty/cargo-affected/actions/runs/30889231291)) repeated the same claim — *"CI on `551e059` is green on every check (lint, msrv, test linux/macos/windows, tend-review)"* — from the same rollup.

Outcome was benign here because the concurrent push happened to be the fix. In the general case the poll can report green for a commit that another actor's push masked, and the gated follow-up the recipe exists to drive — dismiss the approval, fix the failure — never fires.

## Why it will recur

Structural. The recipe has no SHA pin anywhere, so any concurrent push retargets the gate. This repo routinely has three or four tend sessions live on one PR (`tend-review`, `tend-mention` on each review event, `tend-notifications`), so concurrent pushes are the normal case, not an edge one.

## Proposed fix

Pin the poll to the SHA that was pushed, and treat a moved head as a distinct outcome rather than silently adopting it. Verified against `551e059` — this surfaces the `lint` failure the PR-scoped poll missed:

```bash
PUSHED_SHA=$(git rev-parse HEAD)

# Poll the commit, not the PR: `gh pr view --json statusCheckRollup` follows
# the PR head, so a concurrent push retargets the gate mid-loop and the run
# ends up reporting another commit's result as its own.
check_state() {
  gh api "repos/$REPO/commits/$PUSHED_SHA/check-runs?per_page=100" \
      [.check_runs[]
       | select((.html_url // "") | test($own) | not)
       | select((.name // "") != $wf)]
      | {pending: ([.[] | select(.status != "completed")] | length),
         failed:  [.[] | select(.conclusion == "failure") | .name]}'
}
```

Exit the loop when `pending == 0`, then branch on `failed`. Before reporting, confirm the branch is still where you left it:

```bash
HEAD_NOW=$(gh pr view <number> --json headRefOid --jq '.headRefOid')
[ "$HEAD_NOW" = "$PUSHED_SHA" ] || echo "branch advanced to $HEAD_NOW — $PUSHED_SHA's result is what this run is accountable for"
```

Two notes on the endpoint: `commits/{sha}/check-runs` returns check-runs only, so a repo using legacy commit statuses also needs `commits/{sha}/status`; and it has no `workflowName` field, so the same-workflow exclusion keys on `.name` instead.

A smaller variant, if pinning is too large a change: keep the rollup poll but capture `PUSHED_SHA` up front and re-read `headRefOid` before reporting, so a moved head is at least stated rather than absorbed.

tend-agent:
## Problem

The CI Monitoring loop in `running-in-ci` gated on `gh pr view <number> --json statusCheckRollup`, which describes the PR's **current head** rather than the commit the session pushed. Any other actor pushing while the loop slept retargeted the gate: the loop began waiting on the new commit, exited when *those* checks settled, and the closing `gh pr checks <number>` printed the new commit's results. The run then reported its own push as green without ever observing a terminal state for it — and the gated follow-up the recipe exists to drive (fix the failure, dismiss the approval) never fired.

The loop was also blind to an early failure: `pending()` counted only in-flight items, so a check reaching `failure` while siblings ran didn't break the loop, and nothing in the exit path named it.

Reported in #829, with a concrete instance on `max-sixty/cargo-affected` PR #54.

## Solution

Query the same rollup by commit OID (GraphQL `repository.object(oid:).statusCheckRollup`) instead of through the PR. One call keeps everything the PR-scoped version had — check-runs *and* legacy status contexts, the own-run URL filter, and the same-workflow filter (the workflow name lives on `checkSuite.workflowRun.workflow`, since a CheckRun's `name` is the job name) — while staying pinned to the SHA the run is accountable for.

The helper now returns `{pending, failed}` rather than a bare count, so the loop breaks as soon as its own SHA is red and the exit path names the failing checks with their run URLs. After the loop, `headRefOid` is re-read: a moved head is reported as a distinct outcome rather than silently absorbed.

`PINNED_SHA` covers both gated cases, and the block spells out a derivation for each rather than naming one. After your own push it's `git rev-parse HEAD`. In a review session it is not: `tend-review` checks out `refs/pull/N/merge`, and that ephemeral merge commit carries no `statusCheckRollup` at all — the query returns `null`, the loop breaks on iteration 1 with `{"pending":0,"failed":[]}`, and the approval is reported green without the commit it sits on ever being observed. That is this PR's own bug relocated onto the approve-then-dismiss gate, so the review-session derivation (`gh pr view <number> --json headRefOid`) ships alongside, commented.

## Testing

Replayed the reported instance against the live API. `551e059` on `max-sixty/cargo-affected` PR #54, whose head has since advanced to `6955352`:

```
$ # old recipe, PR-scoped
{"total":6,"failed":[]}

$ # new recipe, pinned to 551e059
{"pending":0,"failed":["lint https://github.com/max-sixty/cargo-affected/actions/runs/30889116309/job/91926839301"]}
branch advanced to 6955352192882134d0fb0df302e849fff27ab234 — the result above is still 551e059798afbdfe7d458e29484c0faa4049250c's, which is what this run is accountable for
```

The old form reports zero failures; the pinned form surfaces the `lint` failure that the session missed, and flags the moved head.

The merge-commit case was then confirmed against this PR's own review run: `4897ea37` (its `git rev-parse HEAD`) returns `statusCheckRollup: null`, while the PR head `8582f53` carries eight contexts — so the merge commit's `{"pending":0,"failed":[]}` is indistinguishable from settled green. Bucketing was exercised against synthetic nodes covering every conclusion: `STARTUP_FAILURE` and `ACTION_REQUIRED` now count as failed (they previously landed in neither bucket, so a job that never started read as green), `CANCELLED` stays excluded, and both the own-run and same-workflow filters still hold. The three-way outcome line was exercised across all of red, cap-hit, and green. The own-run and same-workflow filters were exercised in that replay too (`GITHUB_RUN_ID=30889113519`, `GITHUB_WORKFLOW=tend-review` — the `review` check-run was correctly excluded). The block as committed passes `bash -n`.

## Review rounds

Five rounds of review on this branch found the same false-green class surviving inside the fix itself; all are addressed:

- **`PINNED_SHA` derivation.** `git rev-parse HEAD` is wrong in a review session — `tend-review` checks out `refs/pull/N/merge`, and that merge commit's `statusCheckRollup` is `null`, so the loop broke on iteration 1 with a false green on the approve-then-dismiss gate. Both derivations are now spelled out, one commented.
- **Failure buckets.** `STARTUP_FAILURE` and `ACTION_REQUIRED` are terminal `CheckConclusionState` values that landed in neither bucket, so a job that never started read as green. `CANCELLED` stays excluded.
- **Cap signal.** The single trailing `echo "$R"` exited 0 whether the loop settled or timed out. Replaced with an explicit outcome line naming red, cap-hit, no-rollup, and green.
- **Break-on-red removed.** `review`'s pre-APPROVE path routes a `FAILURE` co-existing with in-flight checks *back through this loop* to be waited out, because that shape is usually a stale cancellation-cascade artifact. An early break returned ~60 s in with the same mid-flight snapshot, so the caller re-peeked an identical rollup.
- **Every no-rollup response now reaches the guard.** `// []` collapsed a GraphQL `errors` envelope, an unresolvable OID, and a nullable `statusCheckRollup` all to `{"pending":0,"failed":[]}` — byte-identical to settled green. `// empty` emits nothing so they route to `no rollup returned … unverified, not green`. Both in-loop queries also guard on an empty `$R`: the `[` tests errored on empty strings and fired neither `continue` nor `break`, costing 90 s per iteration — 810 s against the 600 s Bash cap, which killed the call before the verdict printed. Now 540 s.

- **Superseded check runs.** The rollup was read un-reduced. A concurrency-cancelled run's check runs stay on the commit forever, and an `if: always()` merge-gate omnibus whose dependency was cancelled concludes `FAILURE` rather than `CANCELLED`, so the bucket doesn't exclude it — once the replacement settled the loop broke with `PENDING == 0`, read the stale red, and printed `red on $PINNED_SHA`, dismissing an approval over a check whose replacement went green. Now reduced to the latest entry per (check name, workflow), matching the block `review`'s pre-APPROVE path got in #821 — which is what delegates here for exactly that shape.

- **Outcome ordering.** Removing the break left `FAILED > 0` with `PENDING > 0` reachable only at cap expiry — the same unsettled shape the loop is told to wait out — yet it printed `red`. `PENDING` is now tested first, so an unsettled `FAILURE` stays unverified; a settled red still reports red because the loop only breaks once `PENDING` is 0.

Round four's reduction was verified by extracting the committed block and running its jq filter verbatim against a synthetic rollup: a superseded `check-ok-to-merge` `FAILURE` (run 111) drops out while a same-named `FAILURE` from a *different* workflow (run 333) survives, a settled `FAILURE` superseded by a still-`QUEUED` replacement moves from `failed` into `pending`, and `STARTUP_FAILURE`, `CANCELLED`-exclusion, the own-run filter, the same-workflow filter, and the `codecov/patch` status context are all unchanged. That queued case rests on `startedAt` being set at queue time, checked against live `QUEUED` check runs (`pytorch/pytorch`) via both REST and GraphQL rather than assumed. Live commits still answer as before, and an unresolvable OID still emits nothing and routes to `no rollup returned`.

- **Where latest-wins doesn't hold.** The reduction presumes the later entry supersedes the earlier one, which is true for the cancellation cascade it targets and false for two independent runs of a workflow with no concurrency group — `tend-mention`'s group-less `relay` / `verify` / `handle` are that shape, and this branch's head carries two `relay` entries tied to the same second, where `max_by` falls back to input order. Kept the reduction (the recurring stale red is the case worth fixing) and documented the limit, pointing at the per-run conclusion check for when the distinction has to be exact.

Loop behaviour was simulated against a stubbed `rollup()` with `sleep` counting seconds rather than sleeping: settled green 90 s / 2 calls, transient blip then green 150 s / 3, stale cascade red then green 210 s / 4, genuine red 210 s / 4, never settles 540 s / 9 (cap hit), persistent API failure 540 s / 9 (no rollup returned).

## Not included

`review`'s pre-APPROVE bullet still says to re-peek the PR rollup after polling, which has the same head-move exposure. #821 is currently rewriting that exact line, so the one-line change is left out to avoid a conflict; it should adopt `PINNED_SHA=$HEAD_SHA` once #821 lands.

---
Closes #829 — automated triage


tend-agent (commented):
Querying the rollup by commit OID is the right shape, and the own-run / same-workflow filters carry over faithfully (the `checkSuite.workflowRun.workflow.name` hop is correct — a CheckRun's `name` really is the job name). One correctness issue in how `PINNED_SHA` is derived, plus two smaller gaps; all three have inline suggestions.

**`git rev-parse HEAD` is the wrong SHA in exactly the session the comment above it names.** The review workflow checks out `refs/pull/N/merge` — see the `Resolve PR checkout ref` step in `.github/workflows/tend-review.yaml`, which only falls back to `refs/pull/N/head` when the merge ref 404s on a conflicting PR. So in a review session HEAD is an ephemeral merge commit, and a merge commit carries no check rollup at all. The loop then breaks on iteration 1 with `{"pending":0,"failed":[]}` and reports green without ever observing the commit the approval sits on. That's the same false-green this PR is fixing, relocated from "the PR head moved" to "the pinned SHA was never the right one" — and it lands on the approve-then-dismiss gate, which is the one case the PR body says `PINNED_SHA` covers.

Since shell state doesn't persist between Bash calls, `${HEAD_SHA:-…}` won't rescue it either — `HEAD_SHA` from the review skill's step 1 is gone by the time this block runs. The derivation has to be spelled out per session type.

<details><summary>Verified from inside this review run</summary>

`git rev-parse HEAD` here is `4897ea376798028919212298b4138991f0df896f`, a merge of the PR head `8582f53499b8bbce01391f91c237e50a53cfdacd` into `4672f809cedd2f206e109761072641798aa0184e`. Running the PR's own `rollup()` query against each:

```
--- 4897ea37 (git rev-parse HEAD) ---
statusCheckRollup: null, 0 contexts   ->  {"pending":0,"failed":[]}
--- 8582f534 (PR head) ---
5 contexts: review IN_PROGRESS, test/test-proxy/test-worker/lint SUCCESS
```

The merge commit's `{"pending":0,"failed":[]}` is indistinguishable from a settled-green result, so nothing downstream can tell the gate never ran.

</details>

Smaller, both inline: `STARTUP_FAILURE` and `ACTION_REQUIRED` are terminal `CheckConclusionState` values that land in neither bucket (not pending, not failed), so a job that fails to start reads as green; and the cap-expiry signal is gone — the old block ended `echo "CI still running after 9 minutes"; exit 1`, while the new one falls through to the same `echo "$R"` whether the loop settled or timed out, always exiting 0. Numbered item 4 still asks the run to report still-pending checks as unverified, so it's worth keeping something that says which case it's in.

Not blocking, and leaving `review`'s pre-APPROVE peek to #821 reads right — it's still open and touching that line.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,56 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
-      [.statusCheckRollup[]
-       | select((.detailsUrl // .targetUrl // "") | test($own) | not)
-       | select((.workflowName // "") == $wf | not)
-       | (.status // .state)
-       | select(. == "IN_PROGRESS" or . == "QUEUED" or . == "PENDING" or . == "WAITING" or . == "REQUESTED" or . == "EXPECTED")
-      ] | length'
+
+# The SHA this run is accountable for: the commit you just pushed, or the
+# commit you approved (`HEAD_SHA` in a review session).
+PINNED_SHA=$(git rev-parse HEAD)
```
tend-agent: `tend-review` checks out `refs/pull/N/merge`, so here `git rev-parse HEAD` is the merge commit — no rollup, instant false green. Spelling both derivations out (one commented) keeps the block runnable verbatim:

```suggestion
# The SHA this run is accountable for. Pick the derivation that matches your
# session: after your own push it's the branch tip you pushed; in a review
# session the checkout is `refs/pull/N/merge`, an ephemeral merge commit that
# carries no rollup of its own, so `git rev-parse HEAD` would break the loop on
# iteration 1 with a false green — pin the PR head instead.
PINNED_SHA=$(git rev-parse HEAD)                                         # you pushed the commit
# PINNED_SHA=$(gh pr view <number> --json headRefOid --jq '.headRefOid')  # you approved it
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,56 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+        } } } } }
+      }
+    }' -F owner="${GITHUB_REPOSITORY%/*}" -F name="${GITHUB_REPOSITORY#*/}" -F oid="$PINNED_SHA" \
+  | jq -c --arg own "/runs/$GITHUB_RUN_ID/" --arg wf "$GITHUB_WORKFLOW" '
+      [(.data.repository.object.statusCheckRollup.contexts.nodes // [])[]
+       | select(((.detailsUrl // .targetUrl // "") | test($own)) | not)
+       | select((.checkSuite.workflowRun.workflow.name // "") != $wf)]
+      | {pending: [.[] | (.status // .state)
+                   | select(IN("IN_PROGRESS","QUEUED","PENDING","WAITING","REQUESTED","EXPECTED"))] | length,
+         failed:  [.[] | select((.conclusion // .state) | IN("FAILURE","TIMED_OUT","ERROR"))
```
tend-agent: `STARTUP_FAILURE` and `ACTION_REQUIRED` are terminal `CheckConclusionState` values, so today they count as neither pending nor failed and a job that never started reads as green. (`CANCELLED` staying out is right — the skill treats cancellations as no-ops.)

```suggestion
         failed:  [.[] | select((.conclusion // .state) | IN("FAILURE","TIMED_OUT","ERROR","STARTUP_FAILURE","ACTION_REQUIRED"))
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,56 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
-  gh pr checks <number>
-  exit 0
+  R=$(rollup)
+  read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
+  [ "$PENDING" -eq 0 ] && break
 done
-echo "CI still running after 9 minutes"
-exit 1
+
+echo "$R"
```
tend-agent, line 323: The cap-expiry path now prints the same thing as the settled path and exits 0, so item 4 ("comment the still-pending checks as unverified") has nothing to trigger on beyond reading `pending` out of the JSON:

```suggestion
echo "$R"
[ "$PENDING" -gt 0 ] && echo "cap hit — still pending on $PINNED_SHA after 9 minutes; report these as unverified, not as passing"
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Confirmed and applied in e6362ee — the suggestion is taken verbatim.

Verified the mechanism from this run rather than by reading alone: `refs/pull/N/merge` is what [`tend-review.yaml:51`](https://github.com/max-sixty/tend/blob/e6362ee/.github/workflows/tend-review.yaml#L51) checks out, and that merge commit's rollup is `null` — not "zero pending contexts", but no rollup object at all, which is why it collapsed to a green-looking result.

<details><summary>Both SHAs through the PR's own query</summary>

```
4897ea37 (the merge commit `git rev-parse HEAD` returned in your run): statusCheckRollup: null (no checks)
8582f53  (PR head): test=SUCCESS relay=SUCCESS review=SUCCESS verify=SKIPPED test-proxy=SUCCESS handle=SKIPPED test-worker=SUCCESS lint=SUCCESS
```

Eight real contexts on the head, none on the merge commit. Your point about shell state is right too — `HEAD_SHA` from the review skill's step 1 is gone by the time this block runs, so `${HEAD_SHA:-…}` would have silently fallen through to the merge commit and reproduced the same false green.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied verbatim in e6362ee. Both are in the live `CheckConclusionState` enum (`gh api graphql` on `__type(name: "CheckConclusionState")` returns `ACTION_REQUIRED TIMED_OUT CANCELLED FAILURE SUCCESS NEUTRAL SKIPPED STARTUP_FAILURE STALE`), and your read on `CANCELLED` is right, so it stays out.

<details><summary>Bucketing exercised against synthetic nodes</summary>

Fed the edited jq program a rollup covering every branch:

```
{"pending":1,"failed":["startup https://x/1","needs-approval https://x/2","legacy-status https://x/6"]}
```

`STARTUP_FAILURE` and `ACTION_REQUIRED` now land in `failed` (both read as green before); `CANCELLED` and `SUCCESS` in neither; `IN_PROGRESS` in `pending`; the legacy `StatusContext` with `state: FAILURE` still caught; and the own-run URL and same-workflow nodes still excluded — so the added conclusions don't disturb either filter.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Took the point, but not the literal suggestion — `[ "$PENDING" -gt 0 ]` alone also fires on the break-on-red path, where the loop can exit on iteration 1 with checks still queued. That would print "cap hit … after 9 minutes" about sixty seconds in, and the run's disposition there is "go diagnose", not "report as unverified". Since red-with-pending and cap-hit are genuinely different outcomes, e6362ee names all three instead:

```bash
echo "$R"
if   [ "$FAILED" -gt 0 ];  then echo "red on $PINNED_SHA — diagnose the failures above"
elif [ "$PENDING" -gt 0 ]; then echo "cap hit — still pending on $PINNED_SHA; report these as unverified, not as passing"
else                            echo "green on $PINNED_SHA"
fi
```

Dropped "after 9 minutes" from the string for the same reason the old `exit 1` line was worth losing: it hardcodes a duration the branch can't actually vouch for.

<details><summary>Each branch exercised</summary>

```
FAILED=2 PENDING=3 -> red on abc123 — diagnose the failures above
FAILED=0 PENDING=4 -> cap hit — still pending on abc123; report these as unverified, not as passing
FAILED=0 PENDING=0 -> green on abc123
FAILED=1 PENDING=0 -> red on abc123 — diagnose the failures above
```

Item 3 keys off the first and third, item 4 off the second. The extracted block still passes `bash -n`.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,64 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+         failed:  [.[] | select((.conclusion // .state) | IN("FAILURE","TIMED_OUT","ERROR","STARTUP_FAILURE","ACTION_REQUIRED"))
+                       | "\(.name // .context) \(.detailsUrl // .targetUrl // "")"]}'
 }
 for i in $(seq 1 9); do
   sleep 60
-  [ "$(pending)" -gt 0 ] && continue
+  R=$(rollup)
+  read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
+  [ "$FAILED" -gt 0 ] && break     # already red — stop waiting, go diagnose
+  [ "$PENDING" -gt 0 ] && continue
```
tend-agent: `review`'s pre-APPROVE section routes an unsettled red *back through this loop*: for `$FAILED` set and `$PENDING > 0` it says "the rollup hasn't settled. Foreground-poll until non-own checks are terminal (the Step 6 / `running-in-ci` CI-monitoring loop), then re-peek `$FAILED`." With the early break the loop can't carry that out — it returns ~60s in with the same mid-flight `FAILURE`, so the caller re-peeks an identical snapshot and either spins or judges a cancellation-cascade red as real, which is the exact outcome that bullet exists to prevent.

#821 leans on it harder: it adds a `$FAILED` set **and the poll cap expired** with `$PENDING > 0` branch, resolving each remaining `FAILURE` to its run conclusion. The early break makes that branch unreachable through this loop — with a `FAILURE` on the commit, the cap is never reached.

Dropping the break keeps everything this PR fixes (the `failed` bucket, the named failures with run URLs, the outcome line) and costs only the wait — which is the wait `review` is asking for. The push path still exits as soon as everything is terminal.

```suggestion
  [ "$PENDING" -gt 0 ] && continue
```

If the early exit is worth keeping instead, then those two bullets in `review/SKILL.md` need to move with it in this PR — they're written against the old behaviour either way.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,64 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+  [ "$PENDING" -eq 0 ] && break
 done
-echo "CI still running after 9 minutes"
-exit 1
+
+echo "$R"
+if   [ "$FAILED" -gt 0 ];  then echo "red on $PINNED_SHA — diagnose the failures above"
+elif [ "$PENDING" -gt 0 ]; then echo "cap hit — still pending on $PINNED_SHA; report these as unverified, not as passing"
+else                            echo "green on $PINNED_SHA"
+fi
```
tend-agent: When the `gh api graphql` call fails, `$R` is empty, `read` leaves both variables unset, both `[` tests error to stderr, and the `else` fires — so the run announces `green on $PINNED_SHA` on a rollup it never received. The old form printed `CI still running after 9 minutes` and `exit 1` here, so this is new. Reproduced with `R=""`:

```
$ read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
[: : integer expression expected
[: : integer expression expected
green on SHA
```

(A response that parses but carries no contexts — the null-`statusCheckRollup` merge-commit case — is a separate matter that the derivation comment above already covers.)

```suggestion
if   [ "${FAILED:-0}" -gt 0 ];  then echo "red on $PINNED_SHA — diagnose the failures above"
elif [ "${PENDING:-0}" -gt 0 ]; then echo "cap hit — still pending on $PINNED_SHA; report these as unverified, not as passing"
elif [ -z "$R" ];               then echo "no rollup returned for $PINNED_SHA — unverified, not green"
else                                 echo "green on $PINNED_SHA"
fi
```

tend-agent (commented):
The three fixes from the last round hold up: the two-derivation `PINNED_SHA` block is applied verbatim, `STARTUP_FAILURE`/`ACTION_REQUIRED` are in the live `CheckConclusionState` enum and belong in the failed bucket, and the three-way outcome line is a better answer than the suggestion it replaced — `[ "$PENDING" -gt 0 ]` alone really would have fired on the break-on-red path. Resolving those threads.

Two things left, both about the loop reporting a state it didn't observe.

**The break-on-red exit makes `review`'s settled-rollup path unreachable.** That path names *this* loop as the mechanism for waiting out a mid-flight `FAILURE`, and #821 adds a second branch that presumes the loop can reach its cap with a `FAILURE` present. The early break returns after one iteration instead. Inline on the break line, with the tradeoff.

**The `else` branch reports green when the rollup query returned nothing at all.** Inline, with the reproduction.

Both are inline suggestions. No approval — self-authored.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Taken — the break is gone in 0ca2c53. Checked the claim against the text rather than taking it on trust, and [`review/SKILL.md:244`](https://github.com/max-sixty/tend/blob/0ca2c53/plugins/tend-ci-runner/skills/review/SKILL.md#L244) says exactly what you quote: for `$FAILED` set with `$PENDING > 0` it sends the caller to "foreground-poll until non-own checks are terminal … then re-peek `$FAILED`". A loop that returns on the first red can't terminate that wait, so the re-peek saw the same snapshot — and the shape it was meant to filter (a cancelled contributor making an `if: always()` omnibus resolve to `FAILURE` rather than `cancelled`) is precisely the one that arrives with siblings still in flight.

Left a comment in place of the break recording why, so it doesn't get re-added as an optimisation:

```bash
# Don't break early on a red: a FAILURE alongside in-flight checks is often a
# stale cancellation-cascade artifact, and `review`'s pre-APPROVE path routes
# exactly that case back here to be waited out. Judge the settled rollup.
[ "$PENDING" -gt 0 ] && continue
```

No change needed in `review/SKILL.md` on this branch as a result — dropping the break is what makes both of those bullets true again, and #821's cap-expired branch is now reachable.

The cost you named is the only one I found: a genuinely-red commit whose siblings are slow now waits for them instead of returning early. That's the wait the caller asked for, and `echo "$R"` still prints the failed list with run URLs the moment the loop ends, so nothing is lost for diagnosis.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Real, and applied verbatim in 0ca2c53. Reproduced your exact output before taking it — with `R=""` both `[` tests error to stderr and the `else` announces green on a rollup that was never received. Worth flagging that this is the same false-green family the PR is about: the run reports a state it never observed, and here a transient API failure is enough to trigger it.

<details><summary>All four branches after the fix</summary>

```
graphql failed (R empty):    no rollup returned for abc123 — unverified, not green
red, settled:                red on abc123 — diagnose the failures above
red, cap hit with pending:   red on abc123 — diagnose the failures above
cap hit, no failures:        cap hit — still pending on abc123; report these as unverified, not as passing
settled green:               green on abc123
```

No stderr noise in the empty case now that the tests default. Row three is the `$FAILED` set / `$PENDING > 0` shape #821 will want: the summary line calls it red, but `echo "$R"` above it still carries both the pending count and the failed list, so the caller re-peeks the full rollup rather than the one-line summary and loses nothing.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,67 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+      repository(owner: $owner, name: $name) {
+        object(oid: $oid) { ... on Commit { statusCheckRollup { contexts(first: 100) { nodes {
+          ... on CheckRun { name status conclusion detailsUrl
+                            checkSuite { workflowRun { workflow { name } } } }
+          ... on StatusContext { context state targetUrl }
+        } } } } }
+      }
+    }' -F owner="${GITHUB_REPOSITORY%/*}" -F name="${GITHUB_REPOSITORY#*/}" -F oid="$PINNED_SHA" \
+  | jq -c --arg own "/runs/$GITHUB_RUN_ID/" --arg wf "$GITHUB_WORKFLOW" '
+      [(.data.repository.object.statusCheckRollup.contexts.nodes // [])[]
```
tend-agent: `// []` makes every no-rollup response — GraphQL `errors` envelope, unresolvable OID, `statusCheckRollup: null` — reduce to `{"pending":0,"failed":[]}`, which is non-empty and identical to a settled green, so the `[ -z "$R" ]` guard below can never see it. Emitting nothing instead routes all three into that guard.

```suggestion
      .data.repository.object.statusCheckRollup.contexts.nodes // empty
      | [ .[]
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,67 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+      | {pending: [.[] | (.status // .state)
+                   | select(IN("IN_PROGRESS","QUEUED","PENDING","WAITING","REQUESTED","EXPECTED"))] | length,
+         failed:  [.[] | select((.conclusion // .state) | IN("FAILURE","TIMED_OUT","ERROR","STARTUP_FAILURE","ACTION_REQUIRED"))
+                       | "\(.name // .context) \(.detailsUrl // .targetUrl // "")"]}'
 }
 for i in $(seq 1 9); do
   sleep 60
-  [ "$(pending)" -gt 0 ] && continue
+  R=$(rollup)
+  read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
```
tend-agent, line 311: When `$R` is genuinely empty these two `[` tests error (`integer expression expected`) and return 2, so neither `continue` nor `break` fires — the iteration falls through to the 30 s grace and costs 90 s. Nine of those is 810 s against the 600 s Bash cap, so the call is killed and the verdict block never prints. Retrying explicitly keeps each failed iteration at 60 s (540 s total) and leaves `$R` empty for the `no rollup returned` branch.

```suggestion
  R=$(rollup)
  # No rollup came back at all — retry rather than fall through the tests below
  # on empty strings, which neither continue nor break and blow the 600s cap.
  [ -z "$R" ] && { PENDING=; FAILED=; continue; }
  read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,67 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+  # Don't break early on a red: a FAILURE alongside in-flight checks is often a
+  # stale cancellation-cascade artifact, and `review`'s pre-APPROVE path routes
+  # exactly that case back here to be waited out. Judge the settled rollup.
+  [ "$PENDING" -gt 0 ] && continue
   sleep 30
-  [ "$(pending)" -eq 0 ] || continue
-  gh pr checks <number>
-  exit 0
+  R=$(rollup)
+  read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
```
tend-agent, line 319: Same guard on the grace re-query, so a rollup that vanishes between the two calls doesn't leave the previous iteration's counts standing in for a response that never arrived.

```suggestion
  R=$(rollup)
  [ -z "$R" ] && { PENDING=; FAILED=; continue; }
  read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
```

tend-agent (commented):
The break-on-red removal is right, and the comment above it now says why the loop has to outlast a mid-flight `FAILURE` — `review`'s settled-rollup path is reachable again. Resolving that thread.

The empty-`$R` fix is the half of the false-green it can reach, but the guard it added can't see the larger half. `rollup()`'s `// []` turns every *"this response carries no rollup"* shape into `{"pending":0,"failed":[]}` — byte-identical to a settled green, and non-empty, so `[ -z "$R" ]` never fires. Three ways in, all verified against the live API from this run: a GraphQL `errors` envelope (rate limit, transient server error — `gh` writes the body to stdout and jq happily reduces it), an OID GraphQL can't resolve, and a commit whose checks haven't registered yet. Each one breaks the loop on iteration 1 and prints `green on $PINNED_SHA` for a rollup the run never received. That's the PR's own thesis — reporting a state you didn't observe — surviving in the branch meant to catch it.

The second thread stays open rather than resolved, because the truly-empty case isn't fully closed either: the in-loop `[ "$PENDING" -gt 0 ]` / `-eq 0` tests still error on an empty string, so neither `continue` nor `break` fires and each iteration costs 90 s instead of 60 s. Nine of those is 810 s against the 600 s Bash cap this block is explicitly sized for — the call is killed mid-loop and the verdict block never runs, so a persistent API failure still ends with no verdict at all rather than "unverified, not green".

Three inline suggestions; together they route all of it into the guard you already wrote. No approval — self-authored.

<details><summary>Verified: the three no-rollup shapes are indistinguishable from green</summary>

Live responses, this run:

```
$ gh api graphql -f query='... object(oid:"deadbeef…") { ... statusCheckRollup … }'
{"data":{"repository":{"object":null}}}                       # exit 0

$ gh api graphql -f query='query { nonexistentField }'
{"errors":[{"message":"Field 'nonexistentField' doesn't exist on type 'Query'"…}]}   # exit 1, body on stdout
```

`statusCheckRollup` is schema-nullable (`{"name":"statusCheckRollup","type":{"kind":"OBJECT","name":"StatusCheckRollup"}}` — no `NON_NULL` wrapper), so a commit with no checks yet is the same shape.

Each fed through the PR's own jq filter:

```
errors envelope     -> {"pending":0,"failed":[]}
unresolvable OID    -> {"pending":0,"failed":[]}
statusCheckRollup:null -> {"pending":0,"failed":[]}
settled green       -> {"pending":0,"failed":[]}     <- same bytes
```

With the suggestions applied, simulating the loop against a stubbed `rollup()`:

```
persistent API failure          calls=9  R=[]  -> no rollup returned — unverified, not green
settled green                   calls=2        -> green
settled red                     calls=2        -> red
pending -> green                calls=3        -> green
red while pending -> green      calls=3        -> green      (stale cascade waited out)
one blip -> green               calls=3        -> green
never settles                   calls=9        -> cap hit — unverified
```

The failure case is 9 x 60 s = 540 s, inside the 600 s cap, so the verdict block still prints.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Right, and this was the important one — applied in 6eaa835. The guard I added last round only ever covered the case where `gh` wrote nothing at all; `// []` was manufacturing a settled-green value for everything else. Confirmed the collapse before and after, including against the live API:

```
                          old (// [])                new (// empty)
errors envelope           {"pending":0,"failed":[]}  <empty>
unresolvable OID          {"pending":0,"failed":[]}  <empty>
statusCheckRollup: null   {"pending":0,"failed":[]}  <empty>
settled green             {"pending":0,"failed":[]}  {"pending":0,"failed":[]}
```

Live, against real SHAs on this repo: the merge commit `4897ea37` and a bogus all-zero OID both return empty now, while the branch head `0ca2c53` still returns `{"pending":0,"failed":[]}`. So the merge-commit case from the first round has a mechanical backstop and not just the derivation comment — pinning the wrong SHA now reports `no rollup returned … unverified, not green` rather than green.

Also re-checked that `// empty` doesn't swallow a real rollup whose contexts are all filtered out: a node set of `{FAILURE, IN_PROGRESS, own-run, same-workflow}` still reduces to `{"pending":1,"failed":["lint …"]}`, so both filters and both buckets survive the rewrite. `[]` is truthy in jq, so only a genuinely absent `nodes` falls through to `empty`.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied verbatim in 6eaa835. Verified the cap arithmetic by simulating the loop against a stubbed `rollup()` with `sleep` counting seconds instead of sleeping, so the control flow is measured rather than estimated:

```
persistent API failure   before: elapsed=810s  calls=18   (killed at the 600s cap — no verdict)
                         after:  elapsed=540s  calls=9    -> no rollup returned — unverified, not green
```

810 s matches your read exactly: the failed test returns 2, so neither `continue` nor `break` fires and each iteration pays 60 s + 30 s.

<details><summary>The other paths, to check the guard doesn't swallow a real result</summary>

```
settled green                       elapsed=90s   calls=2  -> green
one transient blip -> green         elapsed=150s  calls=3  -> green
red while pending -> settles green  elapsed=210s  calls=4  -> green
pending -> settles genuinely red    elapsed=210s  calls=4  -> red
never settles                       elapsed=540s  calls=9  -> cap hit — unverified
```

Row three is the case `review/SKILL.md:244` cares about — a stale cancellation-cascade `FAILURE` that clears once the replacement run settles now reports green rather than red, which is only reachable because the break-on-red went in the previous round.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied verbatim in 6eaa835 alongside the first-query guard. Agreed on the reasoning — without it a rollup that vanished between the 60 s query and the 30 s grace re-query would leave the first query's `PENDING=0` standing, and the loop would `break` and report green on a response that never arrived. Since the grace re-query exists precisely to catch a late-registering omnibus, treating its failure as "nothing changed" would defeat the check it was added for.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -261,34 +272,72 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+  [ -z "$R" ] && { PENDING=; FAILED=; continue; }
+  read -r PENDING FAILED < <(jq -r '"\(.pending) \(.failed | length)"' <<<"$R")
+  [ "$PENDING" -eq 0 ] && break
 done
-echo "CI still running after 9 minutes"
-exit 1
+
+echo "$R"
+if   [ "${FAILED:-0}" -gt 0 ];  then echo "red on $PINNED_SHA — diagnose the failures above"
+elif [ "${PENDING:-0}" -gt 0 ]; then echo "cap hit — still pending on $PINNED_SHA; report these as unverified, not as passing"
```
tend-agent: `0ca2c53` removed the break-on-red, so `FAILED > 0` with `PENDING > 0` is now only reachable on cap expiry — the stale-cancellation-cascade shape the comment above tells the loop to wait out, and the case `review`'s pre-APPROVE path routes here on purpose. Testing `PENDING` first keeps an unsettled rollup labelled unverified; a settled red still reports red, because the loop only breaks when `PENDING` is 0.

```suggestion
if   [ "${PENDING:-0}" -gt 0 ]; then echo "cap hit — still pending on $PINNED_SHA; report these as unverified, not as passing"
elif [ "${FAILED:-0}" -gt 0 ];  then echo "red on $PINNED_SHA — diagnose the failures above"
```

tend-agent (commented):
The `// empty` rewrite and both `[ -z "$R" ]` guards close the no-rollup half — I re-ran the three shapes through the committed filter (`statusCheckRollup: null`, `object: null`, `errors` envelope) and all three now emit nothing, so they route to `no rollup returned … unverified, not green` instead of the byte-identical green. Resolving those threads.

One finding in the new code, and it's the mirror of this PR's own thesis rather than a repeat of it. The outcome line tests `FAILED` before `PENDING`. That ordering was right when it was written in `e6362ee`, because the loop could still break on red with checks in flight and the disposition there was "go diagnose". `0ca2c53` removed that break — so `FAILED > 0` alongside `PENDING > 0` is now reachable **only** on cap expiry, which is precisely the shape the comment three lines above says not to judge ("a `FAILURE` alongside in-flight checks is often a stale cancellation-cascade artifact … Judge the settled rollup"). A rollup that never settled now prints `red on $PINNED_SHA — diagnose the failures above`, indistinguishable from a settled red.

That lands on the gate `review` delegates here. Its pre-APPROVE path routes `$FAILED` set + `$PENDING > 0` back through this loop specifically to wait the cascade out; if the replacement matrix outlasts the 9-iteration cap, the caller reads "red", skips the approve or dismisses one — for a commit whose checks are still running and will likely go green. Item 4 of the numbered list ("if the cap hits with checks still running, comment the still-pending checks as unverified") also has nothing left to fire on, since the red branch has already claimed the case.

Swapping the two tests is safe: the loop only `break`s when `PENDING` is 0, so a settled red always has `PENDING == 0` and still reports red. Inline suggestion below. No approval — self-authored.

<details><summary>Simulated against a stubbed <code>rollup()</code>, sleeps counted rather than slept</summary>

Committed ordering:

```
stale cascade: FAILURE + in-flight, never settles    -> red                    (540s)
plain never-settles, no failure                      -> cap hit / unverified   (540s)
settled red                                          -> red                    (90s)
settled green                                        -> green                  (90s)
persistent API failure                               -> no rollup / unverified (540s)
```

With `PENDING` tested first:

```
stale cascade: FAILURE + in-flight, never settles    -> cap hit / unverified   (540s)
plain never-settles, no failure                      -> cap hit / unverified   (540s)
settled red                                          -> red                    (90s)
settled green                                        -> green                  (90s)
persistent API failure                               -> no rollup / unverified (540s)
```

Only the first row moves; the failing check names and run URLs are still printed by the `echo "$R"` immediately above, so nothing is lost from the cap-hit report.

The `// empty` check, same run, each shape fed through the committed filter:

```
statusCheckRollup: null   -> (no output)
object: null              -> (no output)
errors envelope           -> (no output)
empty nodes array         -> {"pending":0,"failed":[]}
mixed real nodes          -> {"pending":1,"failed":["lint …/runs/222/job/1","never-started …/runs/444/job/4"]}
```

The last row also confirms `STARTUP_FAILURE` still buckets as failed, `CANCELLED` stays excluded, and both the own-run and same-workflow filters hold.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Correct, and applied in 9c87a82. Checked the invariant the swap rests on before taking it: the loop's only `break` is `[ "$PENDING" -eq 0 ] && break`, so exiting via `break` implies `PENDING == 0` and a settled red still lands on the `red` branch. Cap expiry is the only other exit, which is exactly the case being relabelled.

Your reading of what the previous ordering cost is right — it contradicted the comment three lines above it. Added a note at the branch so the ordering doesn't get "tidied" back:

```bash
# PENDING is tested first on purpose: the loop only breaks once PENDING is 0, so
# a settled red still reports red, while a FAILURE that is still in flight at cap
# expiry stays unverified rather than being judged as a real failure.
```

<details><summary>Both orderings simulated, sleeps counted rather than slept</summary>

```
                                              committed (FAILED 1st)   swapped (PENDING 1st)
stale cascade: FAILURE + in-flight, no settle  red                      cap hit / unverified   540s
plain never-settles, no failure                cap hit / unverified     cap hit / unverified   540s
settled red                                    red                      red                     90s
settled green                                  green                    green                   90s
persistent API failure                         no rollup / unverified   no rollup / unverified 540s
stale cascade -> settles green                 green                    green                  210s
```

One row moves, as you predicted. The last row is the payoff case — a cascade `FAILURE` that clears once the replacement matrix settles reports green in both, so the swap only affects rollups that never settled at all. `echo "$R"` above still prints the failing names and run URLs, so the cap-hit report keeps everything item 4 needs.

</details>

That restores item 4's trigger too — it had nothing left to fire on while the red branch claimed the case.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -296,34 +307,75 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+# PINNED_SHA=$(gh pr view <number> --json headRefOid --jq '.headRefOid')  # you approved it
+
+rollup() {
+  gh api graphql -f query='
+    query($owner: String!, $name: String!, $oid: GitObjectID!) {
+      repository(owner: $owner, name: $name) {
+        object(oid: $oid) { ... on Commit { statusCheckRollup { contexts(first: 100) { nodes {
+          ... on CheckRun { name status conclusion detailsUrl
+                            checkSuite { workflowRun { workflow { name } } } }
+          ... on StatusContext { context state targetUrl }
```
tend-agent: `max_by` needs a sort key that is actually in the selection set, and neither timestamp is requested today. Confirmed by introspection that `CheckRun` exposes `startedAt` and `StatusContext` exposes `createdAt` — and that `startedAt` is populated while a run is still in flight, not only once it completes, so an in-flight replacement still beats a settled predecessor.

```suggestion
          ... on CheckRun { name status conclusion startedAt detailsUrl
                            checkSuite { workflowRun { workflow { name } } } }
          ... on StatusContext { context state createdAt targetUrl }
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -296,34 +307,75 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+        } } } } }
+      }
+    }' -F owner="${GITHUB_REPOSITORY%/*}" -F name="${GITHUB_REPOSITORY#*/}" -F oid="$PINNED_SHA" \
+  | jq -c --arg own "/runs/$GITHUB_RUN_ID/" --arg wf "$GITHUB_WORKFLOW" '
+      .data.repository.object.statusCheckRollup.contexts.nodes // empty
+      | [ .[]
+       | select(((.detailsUrl // .targetUrl // "") | test($own)) | not)
+       | select((.checkSuite.workflowRun.workflow.name // "") != $wf)]
```
tend-agent, line 349: The rollup needs reducing to the latest entry per check name and workflow before `failed` is read from it. A concurrency-cancelled run's check runs stay on the commit forever, and an `if: always()` merge-gate omnibus whose dependency was cancelled concludes `FAILURE` rather than `CANCELLED` — so it is not excluded by the bucket below. Once the replacement run settles, this loop breaks with `PENDING == 0`, reads that stale red, and prints `red on $PINNED_SHA`; step 6 then dismisses an approval over a check whose replacement went green.

`review`'s pre-APPROVE block already does this reduction after #821, which the base merge brought onto this branch — so the two blocks currently answer differently for the same commit, and the pre-APPROVE path routes exactly this shape here.

Keying on the workflow name as well as the check name is load-bearing: two workflows can register the same check name, and collapsing those would hide a genuine red behind an unrelated green.

```suggestion
       | select((.checkSuite.workflowRun.workflow.name // "") != $wf)]
      | group_by([(.name // .context), (.checkSuite.workflowRun.workflow.name // "")])
      | map(max_by(.startedAt // .createdAt))
```

tend-agent (commented):
The ordering swap in `9c87a82` is right, and it's the last of the four rounds' findings — every thread on this branch is now resolved.

One new finding, and it arrived with the base merge rather than with your commit. `c5b7977` brought #821 in, which established that a concurrency-cancelled run leaves *both* check runs on the commit, so a rollup has to be reduced to the latest entry per check name and workflow before it is read. `review`'s pre-APPROVE block now does that reduction; the loop in this PR — the one `review` delegates to for exactly that case — does not.

It lands harder here than it did there, because this PR is what introduces the `failed` bucket and the `red on $PINNED_SHA` verdict. A superseded `FAILURE` never leaves the commit, so once the replacement run settles the loop breaks with `PENDING == 0`, reads the stale red out of `failed`, and prints `red on $PINNED_SHA — diagnose the failures above`. Step 6's follow-up then dismisses an approval over a check whose replacement went green — the PR's own thesis, reporting a state you didn't observe, reaching the one gate `review` hands to this loop.

Two inline suggestions. `startedAt` / `createdAt` have to be in the selection set before anything can sort on them, then the same latest-per-(name, workflow) reduction #821 uses. Keying on the workflow name as well as the check name matters for the same reason it does there: two workflows can register the same check name, and collapsing those would hide a genuine red behind an unrelated green.

No approval — self-authored.

<details><summary>Verified: the OID-pinned query returns duplicates un-collapsed; the reduction drops only the superseded entry</summary>

GitHub does not collapse same-named check runs in `statusCheckRollup.contexts`. This PR's own commits carry live examples — `e6362ee` returns 17 contexts of which four are `handle` / `tend-mention`, three are `relay`, three are `verify`:

```
--- e6362eea2313065817a85cb215c29b4db59d3acb ---
{"total":17,"dupes":[{"n":"handle","w":"tend-mention",...},{"n":"handle","w":"tend-mention",...},
                     {"n":"handle","w":"tend-mention",...},{"n":"handle","w":"tend-mention",...},
                     {"n":"relay","w":"tend-mention",...} x3, {"n":"verify","w":"tend-mention",...} x3]}
```

Those all conclude `SKIPPED` / `SUCCESS`, so they cost nothing today; the point is that the duplicates arrive. Fed a synthetic rollup where the superseded entry is the red one — `check-ok-to-merge` `FAILURE` at 10:00 from run 111, replaced by `SUCCESS` at 10:05 from run 222, plus a genuine `check-ok-to-merge` `FAILURE` from a *different* workflow, a `CANCELLED`/`SUCCESS` pair, a `STARTUP_FAILURE`, and a `codecov/patch` status context:

```
committed: {"pending":0,"failed":["check-ok-to-merge","lint","codecov/patch","check-ok-to-merge"]}
deduped:   {"pending":0,"failed":["check-ok-to-merge","codecov/patch","lint"]}
```

The committed filter reports the superseded red. The deduped one drops it, keeps the other workflow's genuine `check-ok-to-merge` red (which is what the `workflowName` half of the key buys), and leaves `STARTUP_FAILURE`, `CANCELLED`-excluded, and the status-context handling exactly as they were.

`startedAt` and `createdAt` were confirmed by introspection rather than assumed — `CheckRun` exposes `startedAt` (and `completedAt`), `StatusContext` exposes `createdAt`. `startedAt` is populated while a run is still going, not only at completion: the live rollup on this PR's head shows `review` `IN_PROGRESS` with `startedAt: 2026-08-12T06:49:06Z`, so an in-flight replacement still wins `max_by` against a settled predecessor.

With both suggestions applied, the block passes `bash -n` and runs live:

```
e6362eea (17 contexts, 10 of them duplicates) -> {"pending":0,"failed":[]}
c5b79778 (live head, this run in flight)      -> {"pending":1,"failed":[]}
06b57475 (base tip)                           -> {"pending":0,"failed":[]}
```

The `// empty` guard still fires — `group_by` sits after the array construction, so an unresolvable OID and an all-zeros OID both still emit nothing and route to `no rollup returned … unverified, not green`. `pending` is unaffected either way: a superseded check run is `COMPLETED`, so it was never in that bucket.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied verbatim in `f519a7c`. Checked the sort key rather than taking it on trust, because `max_by` over a null key silently picks the wrong entry: introspection confirms `CheckRun.startedAt` / `StatusContext.createdAt`, and live `QUEUED` check runs carry a non-null `startedAt` in both REST and GraphQL — so a replacement that is still queued beats its settled predecessor and the `pending` signal survives the reduction. That was the case worth confirming: had `startedAt` been null until a run starts, the reduction would have dropped a queued replacement in favour of a settled `FAILURE` and broken the loop early on a stale red — the same class, one layer down.

<details><summary>Queue-time <code>startedAt</code>, live</summary>

`QUEUED` check runs on `pytorch/pytorch` heads, REST then GraphQL:

```
{"n":"macos-py3-arm64 / test (openreg, 1, 1, macos-m1-stable)","s":"queued","started":"2026-08-12T06:42:11Z"}
{"n":"torchtitan-x-pytorch-test / test (…)","s":"QUEUED","startedAt":"2026-08-12T06:09:49Z"}
```

Exercised through the committed filter too — a settled `FAILURE` superseded by a `QUEUED` replacement leaves `failed` and lands in `pending`, rather than the un-reduced form's report of both at once.

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Real, and applied verbatim in `f519a7c`. Confirmed the delegation the finding turns on before taking it: [`review/SKILL.md:276`](https://github.com/max-sixty/tend/blob/f519a7c7293d74633fb30b08c7201e213756303d/plugins/tend-ci-runner/skills/review/SKILL.md#L276) routes `$FAILED` set with `$PENDING > 0` into this loop to be waited out, and [`:255-259`](https://github.com/max-sixty/tend/blob/f519a7c7293d74633fb30b08c7201e213756303d/plugins/tend-ci-runner/skills/review/SKILL.md#L255-L259) does the reduction on the way in — so the caller deduped and the callee it hands the decision to did not, and the two answered differently for the same commit. Reduction added, plus a comment recording why the key includes the workflow name.

Verified by extracting the committed block and running its jq filter verbatim, so the test exercises the shipped text rather than a retyped copy:

```
committed: {"pending":1,"failed":["check-ok-to-merge …/runs/333","codecov/patch …","lint …/runs/555"]}
old:       {"pending":1,"failed":["check-ok-to-merge …/runs/111","check-ok-to-merge …/runs/333","lint …/runs/555","flaky …/runs/666","codecov/patch …"]}
```

The superseded `check-ok-to-merge` red (run 111) drops; the same-named red from a *different* workflow (run 333) survives, which is the `workflowName` half of the key paying for itself. `flaky` — a settled `FAILURE` whose replacement is still `QUEUED` — leaves `failed` for `pending` instead of being reported as both. `STARTUP_FAILURE` still counts, `CANCELLED` still doesn't, the own-run and same-workflow filters still hold, and the `codecov/patch` status context still resolves through `createdAt`.

The `// empty` guard is unaffected, as you noted — `group_by` sits after the array construction, so an unresolvable OID still emits nothing and routes to `no rollup returned … unverified, not green`. Live: `e6362ee` (17 contexts, 10 duplicates) → `{"pending":0,"failed":[]}`, `06b5747` → same, all-zeros OID → empty. `bash -n` and `pre-commit` pass.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -292,38 +303,92 @@ After pushing, what to do depends on whether a red result creates a follow-up.
...
+# Reduce to the latest entry per (check name, workflow) before reading the
+# buckets. A concurrency-cancelled run's check runs stay on the commit forever,
+# and an `if: always()` merge-gate omnibus whose dependency was cancelled
+# concludes FAILURE rather than CANCELLED — so the bucket below doesn't exclude
+# it. Un-reduced, the loop breaks once the replacement settles, reads that
+# superseded red, and reports red on a commit whose replacement went green.
+# Key on the workflow name too: two workflows can register the same check name,
+# and collapsing those would hide a genuine red behind an unrelated green.
+# `startedAt` is populated at queue time, so a still-queued replacement wins the
+# reduction over its settled predecessor and the pending signal survives.
```
tend-agent, line 315: The reduction reads "latest wins", which presumes the later entry supersedes the earlier one. That holds for the cancellation cascade it targets. It does not hold for two *independent* runs of a workflow with no concurrency group — both stay live on the commit, neither supersedes the other, and the reduction discards one run's real result. `tend-mention`'s `relay` / `verify` / `handle` are exactly that shape (the CLAUDE.md concurrency table lists their group as `none`), and this PR's own head carries them: runs `31572537904` and `31572538790` each registered all three, with `relay` tied to the same second. A tie makes `max_by` pick by input order, so if one of those pairs were ever red/green the loop's verdict is arbitrary.

Judgment call rather than a defect — the trade is a definite recurring stale red against a rare masked one, and #821 already made it for `review`. This documents the limit and points at the precise test (`review`'s cap-expiry path resolves each `FAILURE` to its run and reads that run's own conclusion, which distinguishes the two cases exactly). Landing as-is is defensible too.

```suggestion
# `startedAt` is populated at queue time, so a still-queued replacement wins the
# reduction over its settled predecessor and the pending signal survives.
# The reduction presumes later supersedes earlier — true for a cancellation
# cascade, false for two independent runs of a workflow with no concurrency
# group, where it discards one run's real result. Resolve a red to its own run's
# conclusion (`review`'s cap-expiry path) when that distinction has to be exact.
```

tend-agent (commented):
Both suggestions applied verbatim, and the reduced filter holds up when I run it rather than read it: fed a rollup carrying a superseded `check-ok-to-merge` `FAILURE` alongside its replacement's `SUCCESS`, plus a genuine same-named `FAILURE` from a second workflow, a `codecov/patch` status context, and a `CANCELLED` leg whose replacement is still `QUEUED`, it returns `{"pending":1,"failed":["check-ok-to-merge …/runs/333/…","codecov/patch …"]}` — superseded red dropped, the other workflow's red kept, and the queued replacement still holding the loop open. The `startedAt`-at-queue-time claim checks out against a live `QUEUED` check run too (`started_at` non-null while `status` is still `queued`), so the pending signal survives the reduction. Both threads resolved.

One inline note on the block's framing, not on its behaviour — a judgment call for you rather than a finding, and fine to land as-is.

No approval — self-authored.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Taken — applied verbatim in `2046e99`. Checked the counterexample rather than accepting the framing, and it reproduces on this PR's head: three independent `tend-mention` runs each registered all three checks, and the two `relay` entries from `31572537904` / `31572538790` are tied to the same second.

```
{"n":"relay","run":31572537904,"c":"SUCCESS","t":"2026-08-12T07:05:27Z"}
{"n":"relay","run":31572538790,"c":"SUCCESS","t":"2026-08-12T07:05:27Z"}
```

So the tie is not hypothetical, and on a tie `max_by` falls back to input order — which GitHub does not promise means anything. Agreed on the trade: a stale red from a cancellation cascade is the recurring case and this loop's whole job is to not report one, while the independent-runs case needs two live runs of a group-less workflow to disagree. Worth stating as a limit rather than leaving the reader to infer "latest wins" is always sound, and the pointer to the per-run conclusion check gives them the exact out.

Both of that shape's escape hatches now exist in the same place: `pending` still counts every un-reduced-away in-flight entry, and a red that has to be attributed precisely resolves to its own run.

max-sixty:
Superseded by #965, which lands the same fix as a tested script rather than a markdown recipe: `plugins/tend-ci-runner/scripts/poll-pr-checks.sh` queries the rollup by commit OID, so a concurrent push can no longer make the poll report another commit's green as this run's. #829 closed with it.

The edge cases this PR's review rounds surfaced are carried over into the script and pinned by pytest: null rollups on ephemeral merge-ref commits, superseded check runs from a cancellation cascade, and the STARTUP_FAILURE / ACTION_REQUIRED conclusions that fall into neither the pending nor the failed bucket if forgotten. Cases the recipe didn't reach are covered too — a commit with more contexts than the query's one 100-node page now routes to UNVERIFIED instead of reading green.

Closing rather than rebasing: the recipe this branch edits no longer exists.

PR #838, opened 2026-08-05
fix(list-recent-runs): recover the window on the last successful run, and stop the fetch limit re-truncating it

tend-agent:
`list-recent-runs.sh`'s dropped-tick recovery anchors the window floor on the previous **completed** run of the analyzing workflow. `completed` is a status, not a conclusion, so a run that failed before its agent produced any analysis advances the anchor exactly as if it had covered its hour. After an outage the next green run therefore resumes one tick back and reports an all-clear for a window nothing ever looked at — and the gap is unrecoverable, because the following run's floor advances past it too.

This is not hypothetical: it just consumed ~15 hours of `review-reviewers` coverage.

## What happened

Every `review-reviewers` run from [30897445507](https://github.com/max-sixty/tend/actions/runs/30897445507) (09:41Z) through [30959115609](https://github.com/max-sixty/tend/actions/runs/30959115609) (23:10Z) failed — 16 consecutive runs, all five matrix legs, `claude -p` exiting 1 after ~4s. The `result` event in every leg's session log is identical: `is_error: true`, `num_turns: 1`, `result: "You've hit your weekly limit · resets 12am (UTC)"`, with `token-usage.json` all zeros. The agent never started, so none of those runs analyzed anything. (The outage itself is already tracked by #831 and the annotation-legibility work in #816 / #818 — this PR is only about the window the outage swallowed.)

The last run that did analyze anything is [30893072876](https://github.com/max-sixty/tend/actions/runs/30893072876) at 08:42Z, which is also the last entry in the `numbagg/numbagg` evidence gist. At this run's tick (23:47Z intended), the recovery query picked `{"conclusion":"failure","createdAt":"2026-08-04T23:10:57Z","databaseId":30959115609}` and floored the window at 22:47Z, so `list-recent-runs.sh` returned exactly one row: a `tend-notifications` pre-check no-op.

Everything `numbagg-bot` actually did in the swallowed window sat outside that floor — PRs [#719](https://github.com/numbagg/numbagg/pull/719) and [#720](https://github.com/numbagg/numbagg/pull/720), issue [#721](https://github.com/numbagg/numbagg/issues/721), three reviews and four inline comments on #719, and comments on #716, #721 and #1. That is the densest window numbagg has had in days, and per the skill's Step 1 ("If empty, record the run as all-clear … then skip to Step 6") it would have been recorded as a quiet hour. I only found it by diffing the run list against the gist's last recorded boundary by hand.

## The change

Anchor on the previous **successful** run rather than any completed one, filtering server-side with `--status success` (the flag takes conclusions as well as statuses), and when the existing 6h cap clamps the recovered floor, say so on stderr so the caller records a coverage gap instead of a false all-clear. A partially-failed matrix run counts as a failure here, which only ever widens the window — overlap is re-offered work the caller dedups against its own evidence log, whereas a gap is silently unanalyzed.

Widening the window then exposed a second truncation one layer down, so the fetch loop moves with it. It passed `--limit 50` while a recovered window spans up to 8h, and `gh run list` returns newest-first — so a workflow over the limit silently drops its *oldest* runs, which are exactly the ones in the gap the anchor just reached back for. On `max-sixty/tend`, `tend-mention` alone produces 57 runs in an 8h window. The limit is now 200, and a workflow returning exactly the limit warns rather than truncating in silence.

Two smaller pieces: the anchor query routes through `gh_retry` and exits non-zero instead of `2>/dev/null || true`, so a transient API error can no longer masquerade as "no successful run" and emit a confident warning naming a cause that didn't happen; and `review-reviewers/SKILL.md`, which still said "if empty, record all-clear" with nothing about the stderr warnings, now treats any `WARNING:` as a coverage gap — without that, an agent following the skill literally would print the warning and record an all-clear anyway.

When every tick fires and succeeds the anchor is the previous tick, `prev_intended == intended - 3600 == COMPLETED_AFTER`, and the comparison is a no-op — output is byte-identical to today's on the healthy path.

<details><summary>Verification against live data</summary>

Same tick, same repo, before vs. after:

```
# anchor query — before (--limit 10, no conclusion filter)
{"conclusion":"failure","createdAt":"2026-08-04T23:10:57Z","databaseId":30959115609}

# anchor query — after (--limit 50, conclusion == "success")
{"conclusion":"success","createdAt":"2026-08-04T08:42:12Z","databaseId":30893072876}
```

`TARGET_REPO=numbagg/numbagg ./list-recent-runs.sh` after the change returns 5 runs instead of 1, preceded by:

```
WARNING: the last successful 'review-reviewers' run started 2026-08-04T08:42:12Z, more than 6h back. Window floored at 2026-08-04T17:47:00Z; runs that completed before it are NOT in this list. Record a coverage gap, not an all-clear.
```

The recovered anchor (08:42:12Z) matches the gist's last recorded coverage boundary exactly. Also exercised: the no-successful-run branch (warns, floors at the cap, exits 0) and the unset-`GITHUB_WORKFLOW` branch (recovery skipped, unchanged). `shellcheck` clean; `bash -n` clean. Also exercised after the review follow-ups: the fetch-limit truncation warning (via a copy with `RUN_LIMIT=5`) and the anchor-query failure path (exits 1 rather than degrading). See [the follow-up comment](https://github.com/max-sixty/tend/pull/838#issuecomment-5186164124) for the full branch table.

</details>

## Gates

- **Evidence level**: Critical — a clearly wrong outcome (the analysis records an all-clear for a window it never examined, and the window is then unreachable), which acts on 1 occurrence. Occurrences: 1 this run, 0 historical (no prior gist entry records an outage of a length that would trigger it).
- **Structural, not stochastic**: no decision point. The anchor query is deterministic; replayed ten times it picks the failed run ten times.
- **Change type**: targeted fix — one predicate, one query limit, two stderr warnings. Normal evidence bar under Gate 2, met.
- **Dedup**: checked open issues #831, #830, #829, #828, #827, #824, #822, #817, #816, #801, #799, #750, #624 and open PRs #826, #825, #823, #821, #819, #818, #809. #816 / #818 / #823 / #809 all concern outage *reporting* — naming the cause in the annotation, the stranded trigger, per-run comment dedup — none touch window recovery. Searching issues and PRs for `list-recent-runs` surfaces only the three merged predecessors #526, #753 and #784.

Evidence log: https://gist.github.com/19b5ab297bb7ac7e1e9a44d595ccde0f

### DIFF
--- a/plugins/tend-ci-runner/scripts/list-recent-runs.sh
+++ b/plugins/tend-ci-runner/scripts/list-recent-runs.sh
@@ -97,22 +97,37 @@ if [ -n "$cron_minute" ]; then
   # Default floor: one cron period back. Consecutive ticks tile exactly.
   COMPLETED_AFTER=$((intended - 3600))
 
-  # Dropped-tick recovery. GHA doesn't only *delay* scheduled ticks, it also
-  # *drops* them: a tick that fires zero times leaves that hour's completions
-  # in the gap between the previous and next cycle's windows (the skipped-tick
-  # case #526 deferred as acceptable). Rather than assume the previous tick
-  # fired, resume from where the previous *actual* completed run of this
-  # workflow left off: recover that run's intended tick and floor the window
-  # there. When every tick fires, the previous run's intended tick == the
+  # Dropped-tick and failed-run recovery. GHA doesn't only *delay* scheduled
+  # ticks, it also *drops* them: a tick that fires zero times leaves that hour's
+  # completions in the gap between the previous and next cycle's windows (the
+  # skipped-tick case #526 deferred as acceptable). Rather than assume the
+  # previous tick fired, resume from where the previous run that actually
+  # analyzed something left off: recover that run's intended tick and floor the
+  # window there.
+  #
+  # Anchor on the previous *successful* run, not merely a completed one. A run
+  # that failed before its agent produced any analysis — harness crash, model
+  # quota exhaustion, timeout — covered no window at all, so resuming from it
+  # discards every hour the outage spanned and the next green run reports a
+  # one-hour all-clear for a window nothing ever looked at. For the same reason
+  # the scan reaches well past one cron period: during an outage the most recent
+  # completed runs are all failures, and a short limit finds no success to
+  # anchor on. A partially-failed matrix run counts as a failure here, which
+  # only ever widens the window — overlap is re-offered work the caller dedups
+  # against its own evidence log, where a gap is silently unanalyzed.
+  #
+  # When every tick fires and succeeds, the previous run's intended tick == the
   # default (intended - 3600), so this is a byte-identical no-op — still no
-  # overlap between consecutive cycles. When a tick was dropped, it reaches
-  # back to cover the orphaned hour. Capped at 6h so a sustained outage can't
-  # create an unbounded window. The analyzing workflow runs on the current
-  # repo, so this query omits TARGET_REPO's -R.
+  # overlap between consecutive cycles. Capped at 6h so a sustained outage can't
+  # create an unbounded window; when the cap bites, say so on stderr so the
+  # caller records a coverage gap instead of a false all-clear. The analyzing
+  # workflow runs on the current repo, so this query omits TARGET_REPO's -R.
   if [ -n "${GITHUB_WORKFLOW:-}" ]; then
+    floor_cap=$((intended - 21600))   # never reach back more than 6h
+    floor_cap_iso=$(date -u -d "@$floor_cap" +%Y-%m-%dT%H:%M:%SZ)
     prev_start=$(gh run list --workflow "$GITHUB_WORKFLOW" --status completed \
-      --limit 10 --json databaseId,createdAt \
-      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0}))] | .[0].createdAt // empty" \
+      --limit 50 --json databaseId,createdAt,conclusion \
+      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0})) | select(.conclusion == \"success\")] | .[0].createdAt // empty" \
       2>/dev/null || true)
     if [ -n "$prev_start" ]; then
       prev_ts=$(date -u -d "$prev_start" +%s 2>/dev/null || echo "")
@@ -123,10 +138,15 @@ if [ -n "$cron_minute" ]; then
         else
           prev_intended=$((prev_hour_tick - 3600))
         fi
-        floor_cap=$((intended - 21600))   # never reach back more than 6h
-        [ "$prev_intended" -lt "$floor_cap" ] && prev_intended=$floor_cap
+        if [ "$prev_intended" -lt "$floor_cap" ]; then
+          echo "WARNING: the last successful '$GITHUB_WORKFLOW' run started $prev_start, more than 6h back. Window floored at $floor_cap_iso; runs that completed before it are NOT in this list. Record a coverage gap, not an all-clear." >&2
+          prev_intended=$floor_cap
+        fi
         [ "$prev_intended" -lt "$COMPLETED_AFTER" ] && COMPLETED_AFTER=$prev_intended
       fi
+    else
+      echo "WARNING: no successful '$GITHUB_WORKFLOW' run among the last 50 completed ones. Window floored at $floor_cap_iso; anything earlier is NOT in this list. Record a coverage gap, not an all-clear." >&2
+      [ "$floor_cap" -lt "$COMPLETED_AFTER" ] && COMPLETED_AFTER=$floor_cap
     fi
   fi
 
--- a/plugins/tend-ci-runner/scripts/list-recent-runs.sh
+++ b/plugins/tend-ci-runner/scripts/list-recent-runs.sh
@@ -109,12 +109,12 @@ if [ -n "$cron_minute" ]; then
   # that failed before its agent produced any analysis — harness crash, model
   # quota exhaustion, timeout — covered no window at all, so resuming from it
   # discards every hour the outage spanned and the next green run reports a
-  # one-hour all-clear for a window nothing ever looked at. For the same reason
-  # the scan reaches well past one cron period: during an outage the most recent
-  # completed runs are all failures, and a short limit finds no success to
-  # anchor on. A partially-failed matrix run counts as a failure here, which
-  # only ever widens the window — overlap is re-offered work the caller dedups
-  # against its own evidence log, where a gap is silently unanalyzed.
+  # one-hour all-clear for a window nothing ever looked at. `--status` takes
+  # conclusions as well as statuses, so the filter runs server-side and no scan
+  # depth is involved: an outage of any length can't bury the anchor. A
+  # partially-failed matrix run counts as a failure here, which only ever widens
+  # the window — overlap is re-offered work the caller dedups against its own
+  # evidence log, where a gap is silently unanalyzed.
   #
   # When every tick fires and succeeds, the previous run's intended tick == the
   # default (intended - 3600), so this is a byte-identical no-op — still no
@@ -125,10 +125,15 @@ if [ -n "$cron_minute" ]; then
   if [ -n "${GITHUB_WORKFLOW:-}" ]; then
     floor_cap=$((intended - 21600))   # never reach back more than 6h
     floor_cap_iso=$(date -u -d "@$floor_cap" +%Y-%m-%dT%H:%M:%SZ)
-    prev_start=$(gh run list --workflow "$GITHUB_WORKFLOW" --status completed \
-      --limit 50 --json databaseId,createdAt,conclusion \
-      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0})) | select(.conclusion == \"success\")] | .[0].createdAt // empty" \
-      2>/dev/null || true)
+    # Route through gh_retry and fail loud, like the other queries here: a
+    # transient API error must not masquerade as "no successful run ever", which
+    # would emit a confidently-worded warning naming a cause that didn't happen.
+    if ! prev_start=$(gh_retry gh run list --workflow "$GITHUB_WORKFLOW" \
+      --status success --limit 5 --json databaseId,createdAt \
+      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0}))] | .[0].createdAt // empty"); then
+      echo "ERROR: 'gh run list' for the window anchor failed after retries — refusing to guess a floor that would read as a false all-clear" >&2
+      exit 1
+    fi
     if [ -n "$prev_start" ]; then
       prev_ts=$(date -u -d "$prev_start" +%s 2>/dev/null || echo "")
       if [ -n "$prev_ts" ]; then
@@ -145,7 +150,7 @@ if [ -n "$cron_minute" ]; then
         [ "$prev_intended" -lt "$COMPLETED_AFTER" ] && COMPLETED_AFTER=$prev_intended
       fi
     else
-      echo "WARNING: no successful '$GITHUB_WORKFLOW' run among the last 50 completed ones. Window floored at $floor_cap_iso; anything earlier is NOT in this list. Record a coverage gap, not an all-clear." >&2
+      echo "WARNING: no successful '$GITHUB_WORKFLOW' run found at all. Window floored at $floor_cap_iso; anything earlier is NOT in this list. Record a coverage gap, not an all-clear." >&2
       [ "$floor_cap" -lt "$COMPLETED_AFTER" ] && COMPLETED_AFTER=$floor_cap
     fi
   fi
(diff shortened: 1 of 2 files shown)


plugins/tend-ci-runner/scripts/list-recent-runs.sh
```diff
@@ -97,22 +97,37 @@ if [ -n "$cron_minute" ]; then
...
+  # workflow runs on the current repo, so this query omits TARGET_REPO's -R.
   if [ -n "${GITHUB_WORKFLOW:-}" ]; then
+    floor_cap=$((intended - 21600))   # never reach back more than 6h
+    floor_cap_iso=$(date -u -d "@$floor_cap" +%Y-%m-%dT%H:%M:%SZ)
     prev_start=$(gh run list --workflow "$GITHUB_WORKFLOW" --status completed \
-      --limit 10 --json databaseId,createdAt \
-      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0}))] | .[0].createdAt // empty" \
+      --limit 50 --json databaseId,createdAt,conclusion \
+      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0})) | select(.conclusion == \"success\")] | .[0].createdAt // empty" \
       2>/dev/null || true)
```
tend-agent:
```suggestion
    prev_start=$(gh run list --workflow "$GITHUB_WORKFLOW" --status success \
      --limit 5 --json databaseId,createdAt \
      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0}))] | .[0].createdAt // empty" \
      2>/dev/null || true)
```

`--status` takes conclusions as well as statuses, so `success` filters server-side and the `--limit 50` scan depth stops carrying any meaning — verified against `review-reviewers`, it returns `30893072876` / `2026-08-04T08:42:12Z`, the anchor in the description's "after" block. The comment paragraph above and the `else` branch's "among the last 50 completed ones" both need to follow.

tend-agent (commented):
Two things on the recovery path, plus a smaller note. Reviewing my own PR, so this is a COMMENT.

**The widened window isn't matched by the per-workflow fetch limit.** The recovery now routinely pushes `COMPLETED_AFTER` to `intended - 6h` on exactly the outage path this PR targets, which makes `CREATED_SINCE` reach `intended - 8h`. But the fetch loop that actually collects the runs — the `gh_retry gh run list ... --created ">=${CREATED_SINCE}"` call guarded by `refusing to report a partial run list`, outside this diff — still passes `--limit 50`. `gh run list` returns newest-first, so more than 50 runs in that window silently drops the oldest ones: precisely the runs sitting in the recovered gap the anchor just reached back for. Measured just now on `max-sixty/tend`, which is itself a matrix leg in `review-reviewers.yaml`: `tend-mention` alone produced 48 runs in the last 8h, i.e. one workflow at the cap before the other seven are counted. Unlike the 6h clamp, this truncation prints nothing, so the caller records the all-clear this PR exists to prevent. `gh` paginates past 100 (`--limit 150` returns 150 here), so raising the limit is cheap; warning when a workflow comes back with exactly `--limit` rows would turn a silent truncation into a visible one. Happy to push that as a commit here if you'd rather it land together.

**`gh run list` can do the conclusion filter server-side** — inline suggestion below. `--status success` is accepted and returns the anchor directly: I ran it against `review-reviewers` and got run [30893072876](https://github.com/max-sixty/tend/actions/runs/30893072876) at `2026-08-04T08:42:12Z`, the same anchor the "after" block in the description shows. That drops the client-side `conclusion` filter and, more usefully, the scan-depth heuristic with it — the "50 completed runs might all be failures" reasoning stops applying, and an outage longer than 50 runs can no longer defeat the anchor. It does need the comment paragraph about the scan reaching past one cron period trimmed, and the `else` branch's wording changed from "among the last 50 completed ones" to something like "at all".

**Smaller:** the anchor query keeps `2>/dev/null || true`, so a transient `gh` failure is indistinguishable from a genuine "no successful run" — and now it produces a confidently-worded warning naming a cause that didn't happen, plus a 6h window. The script's other `gh` calls deliberately route through `gh_retry` and exit non-zero rather than degrade, on the same reasoning this PR is applying.

plugins/tend-ci-runner/scripts/list-recent-runs.sh
tend-agent: Applied in fb26376. Confirmed `--status` takes conclusions — it's in the flag's own help text (`{queued|completed|…|success|timed_out}`), and against `review-reviewers` it returns `30893072876` / `2026-08-04T08:42:12Z` directly, matching the description's "after" anchor.

Both follow-ons went with it: the comment paragraph drops the scan-depth reasoning for a note that the filter is server-side so no scan depth is involved, and the `else` branch now reads "no successful … run found at all".

I kept the `databaseId != $GITHUB_RUN_ID` exclusion. It's dead weight during a real run — the current run is `in_progress`, so `--status success` can't return it — but it's what makes the script replayable from a checkout after the run has gone green, which is how every branch below was exercised.

tend-agent:
All three taken, pushed as fb26376. The fetch-limit one was the important catch — it would have restored the exact failure this PR exists to prevent, one layer down.

**Fetch limit.** Reproduced the measurement: over an 8h window on `max-sixty/tend`, `tend-mention` alone returns 57 runs, so the old `--limit 50` was already truncating one workflow before the other seven were counted. Since `gh run list` is newest-first, the dropped rows are the oldest — i.e. the recovered gap itself, so recovery would reach back for a window and then hand back a list with that window cut off. Raised to 200 and added a warning when a workflow comes back with exactly the limit. That warns rather than exits: unlike a failed fetch, the rows in hand are still worth analyzing — the caller just can't read the list as complete.

**Anchor query.** Both applied as suggested — see the inline reply.

**`gh_retry` on the anchor.** Agreed, and it was worse with this PR than before it: the old code degraded to a silent one-hour window, whereas the new one degrades to a confidently-worded warning naming a cause that didn't happen. Now routed through `gh_retry`, exiting 1 with `refusing to guess a floor that would read as a false all-clear`. A genuine empty result still takes the no-anchor branch — `--jq … // empty` exits 0 with empty stdout, so the two cases are now distinguishable.

**One thing the review didn't raise**, folded in because it's the same false all-clear: the consuming skill still said *"If empty, record the run as all-clear … then skip to Step 6"*, with nothing about the stderr warnings. An agent following it literally would print `WARNING: … Record a coverage gap` and then record an all-clear anyway. `review-reviewers/SKILL.md` now says a `WARNING:` means a coverage gap regardless of whether the list came back empty.

<details><summary>Branch verification (live, this checkout)</summary>

`shellcheck` and `bash -n` clean; `pre-commit` clean on both files.

The anchor is recent again as of this run, so the recovery path needs the just-succeeded run excluded to reproduce — that's `GITHUB_RUN_ID=30962483562` below.

| Branch | Invocation | Result |
|---|---|---|
| Recovery + 6h clamp | `GITHUB_RUN_ID=30962483562`, `TARGET_REPO=numbagg/numbagg` | anchors `2026-08-04T08:42:12Z`, warns, returns 5 rows — matches the description's "after" |
| Healthy path | same, `GITHUB_RUN_ID=0` | anchor is the 00:11Z run, no warning, no widening |
| Truncation | copy with `RUN_LIMIT=5`, `TARGET_REPO=max-sixty/tend` | warns per capped workflow (`tend-mention`, `tend-review`, `tend-notifications`) |
| No anchor at all | stubbed `gh` returning empty for the anchor query | warns, floors at cap, exits 0 |
| Anchor query fails | `GITHUB_WORKFLOW=does-not-exist` | `ERROR: … refusing to guess a floor`, exits 1 |

At the real limit of 200, nothing on `max-sixty/tend` comes close over an 8h window — the busiest workflow is the 57 above.

</details>

plugins/tend-ci-runner/scripts/list-recent-runs.sh
```diff
@@ -97,23 +97,43 @@ if [ -n "$cron_minute" ]; then
...
-      --limit 10 --json databaseId,createdAt \
-      --jq "[.[] | select(.databaseId != (${GITHUB_RUN_ID:-0}))] | .[0].createdAt // empty" \
-      2>/dev/null || true)
+    floor_cap=$((intended - 21600))   # never reach back more than 6h
+    floor_cap_iso=$(date -u -d "@$floor_cap" +%Y-%m-%dT%H:%M:%SZ)
+    # Route through gh_retry and fail loud, like the other queries here: a
+    # transient API error must not masquerade as "no successful run ever", which
+    # would emit a confidently-worded warning naming a cause that didn't happen.
+    if ! prev_start=$(gh_retry gh run list --workflow "$GITHUB_WORKFLOW" \
+      --status success --limit 5 --json databaseId,createdAt \
```
tend-agent:
```suggestion
      --status success --event schedule --limit 5 --json databaseId,createdAt \
```

The anchor doesn't filter by event, so a successful `workflow_dispatch` run can become the floor — and a dispatch run never has `.schedule` in its event payload, so it takes the non-cron branch (`COMPLETED_AFTER=$(date -d '1 hour ago' +%s)`) and only ever covers a now-anchored 1h window. Manually dispatching the workflow mid-outage to check a fix is the natural thing to do, and it's the one action that silently discards the gap: the dispatch covers 1h, the next scheduled run anchors on it, and the outage hours fall between them with no WARNING — the exact false all-clear this PR exists to prevent. Failed runs are excluded for the same reason; a dispatch success is the same class of run, one that succeeded without covering the tiled window.

Verified against `review-reviewers` (which does declare `workflow_dispatch:`): with the filter the query returns `2026-08-05T00:11:09Z`, byte-identical to today's result, since there are no dispatch runs on record — so it's a no-op now and only bites the case above.

The `else` branch's message would want to follow ("no successful scheduled run"), and the comment paragraph above could note that the anchor is scheduled runs only.

tend-agent (commented):
One finding on the anchor query, otherwise the follow-up commit holds up. Reviewing my own PR, so this is a COMMENT.

I exercised each branch against this repo rather than reading them: the normal path returns 55 runs with no warnings; `--status success` is a valid conclusion filter (it's in the flag's own value list) and an existing workflow with zero matching runs returns `[]` at exit 0, so the "no successful run" branch is reachable rather than shadowed by the new `exit 1`; the clamp branch fires correctly on `tend-weekly` (last success 2026-08-02, floored at `2026-08-04T17:47:00Z`); and forcing `RUN_LIMIT=5` produces the truncation warning per workflow. The limit raise is well sized — `tend-mention` returned 80 runs over an 8h window just now, so 50 was truncating by a third and 200 leaves real headroom.

plugins/tend-ci-runner/scripts/list-recent-runs.sh
tend-agent: Good catch — applied in 3521f8b. The dispatch case is worse than the failure case it parallels, because it's the one that produces no `WARNING`: the anchor looks recent, so nothing in the script has a reason to complain, and the gap just disappears.

Verified `--event schedule` is a real filter (`-e, --event event` in `gh run list --help`) and that it's a no-op today — against `review-reviewers` the anchor is `30962483562` / `2026-08-05T00:11:09Z` with and without it, since there are no dispatch runs on record.

Both follow-ons went with it: the `else` message now reads "no successful **scheduled** … run found at all", and the comment paragraph gained a note on why dispatch runs are excluded, framed as the same class of hole as a failure rather than a separate rule.

The new branch is reachable rather than shadowed, and I exercised it against a workflow that genuinely has no scheduled runs rather than a stub — `GITHUB_WORKFLOW=tend-review` (PR-triggered only) hits `WARNING: no successful scheduled 'tend-review' run found at all`, floors at the cap, exits 0. The recovery and healthy paths are unchanged from the previous commit's table.

tend-agent:
Second independent occurrence, from a different cause — reproduced live on this tick, so recording it here rather than opening a divergent PR.

The motivating case was the weekly-limit outage: `claude -p` exiting 1 after the agent never started. Today's is infrastructure. GitHub opened a `critical` Actions incident at 2026-08-06T15:22:49Z (components: Actions, Pages) that is still unresolved as I write this, and five consecutive `review-reviewers` runs died in it — [31119653960](https://github.com/max-sixty/tend/actions/runs/31119653960) (16:23Z) failed at `Set up job` with `Failed to resolve action download info.` / `Service Unavailable`, and [31122877870](https://github.com/max-sixty/tend/actions/runs/31122877870), [31125649607](https://github.com/max-sixty/tend/actions/runs/31125649607), [31126978811](https://github.com/max-sixty/tend/actions/runs/31126978811) and [31128259981](https://github.com/max-sixty/tend/actions/runs/31128259981) each had `init-tracking` cancelled with **zero steps** and the annotation `The job was not acquired by Runner of type hosted even after multiple attempts`, which skipped the whole matrix. Different mechanism, identical consequence: five runs that analyzed nothing, each advancing the anchor exactly as if they had.

The anchor query on this tick, run both ways against live data:

```
# current: --status completed, --limit 10
{"conclusion":"failure","createdAt":"2026-08-06T21:22:19Z","databaseId":31128259981}

# this PR: --status success
{"conclusion":"success","createdAt":"2026-08-06T15:32:18Z","databaseId":31116304881}
```

The failed 21:22Z run floors the window at 20:47Z, so `list-recent-runs.sh` returned 2 rows. Six `tend-notifications` runs on `max-sixty/cargo-affected` completed between 16:12Z and 19:22Z — four of them cancelled with the same runner-acquisition annotation, two still stuck in `status: waiting` five hours after creation with `pending_deployments` showing `wait_timer: 0` and no reviewers — and none of them were in any leg's window. `31116304881` is also the last entry in this target's evidence gist, so the recovered anchor again matches the gist's last coverage boundary exactly.

Worth noting what the cap does here, since this run exercises the branch the PR added rather than the happy path: `15:32:18Z` is more than 6h before the `22:47Z` tick, so `prev_intended` clamps to `16:47Z` and the `WARNING:` fires. That is the design working — the leg records a coverage gap instead of a false all-clear — but the real gap is ~8h (last successful analysis 15:45Z, this tick 23:45Z), and today's incident has now been open 8.5h, so a 6h cap under-covers an incident-length outage by a couple of hours in practice. Not asking for a change; just the datapoint, since the cap was picked before there was a case that exceeded it.

Substantive impact this time was nil, which is only luck: `max-sixty/cargo-affected` saw zero mutations of any kind across the whole 15:45Z–23:45Z window (`issues?state=all&sort=updated`, repo-level `issues/comments` and `pulls/comments`, and `commits?since=` all returned empty), so the swallowed window held nothing to analyze. The numbagg case showed what it costs when the window isn't idle.

tend-agent:
The `max-sixty/worktrunk` leg of the same tick hit this too, and it supplies the one thing the comment above says it was missing: *"Substantive impact this time was nil, which is only luck… the swallowed window held nothing to analyze."* On worktrunk it wasn't idle. The window the failed-run anchor swallowed (15:35Z, the last recorded coverage boundary, → the 20:47Z floor) contained the day's **only** public bot artefact — a 6.9 KB comment on [max-sixty/worktrunk#3757](https://github.com/max-sixty/worktrunk/issues/3757) at 16:40Z, whose six external claims and six links I verified live only because this leg backfilled by hand — plus the `tend-triage` run for that issue dying in `Set up job`, and six more runs cancelled or stuck `waiting`. Reported as a one-hour all-clear, all of that would have gone unexamined.

That also puts a number on the cap question raised above without needing a change: with `--status success` the anchor is 31116304881 (15:32Z), clamped to 16:47Z, so four of the five lost hours come back and the `WARNING` flags the fifth. Under-covering by an hour while saying so is a different failure than covering five hours silently. Third datapoint for the `--event schedule` half, too: 31119653960 has been sitting in `waiting` since 16:23Z, which is exactly the state that invites a hand-dispatch to check on things mid-incident.

PR #840, opened 2026-08-05
fix(mention): count bot engagement outside jq so --paginate can't split it

tend-agent:
## Problem

`tend-mention`'s verify step derives three engagement counts by reducing inside a `--paginate`d `--jq`:

```bash
  --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
```

`gh` applies `--jq` once per page, so the reduction emits one count *per page* rather than one overall. `--paginate` requests `per_page=100`, so it takes more than 100 reviews or more than 100 issue comments on a single thread to cross — past that the variable holds e.g. `100\n7`, and the guard immediately below it errors:

```
$ BOT_REVIEWS=$(printf '100\n7\n'); if [ "$BOT_REVIEWS" -gt "0" ]; then echo run; else echo "rc=$?"; fi
bash: [: 100
7: integer expression expected
rc=2
```

`test` returning 2 makes the `if` false, so the step falls through to `should_run=false`. The bot stops answering participants on exactly the threads where it has engaged the most — silently, with nothing going red. All three sites are on the engagement path that decides whether a non-mention comment gets a session at all: the issue-comment count, the PR-review count, and the PR-comment count.

These render into every adopter's `tend-mention.yaml`, not just ours.

## Solution

Keep the filter a per-element stream — which paginates fine — capture it bare, and test it for emptiness. The values are only ever compared against zero, so no count is needed:

```bash
  --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id')
if [ -n "$BOT_REVIEWS" ]; then
```

Reducing through a pipe (`| wc -l`) would fix the count but move the substitution's exit status off `gh` and onto `wc`. The step declares no `shell:`, so it runs under GHA's default `bash -e {0}` — errexit on, pipefail off — and a failing `gh api` would then read as "no engagement" on a green job rather than aborting the step:

```
$ bash -e -c 'X=$(false | wc -l); echo "survived, X=[$X]"'
survived, X=[0]
$ bash -e -c 'X=$(false); echo survived'; echo "rc=$?"
rc=1
```

`set -o pipefail` is not a safe alternative in this step: `grep -qF` on the review-comment scan exits on first match, `gh` takes SIGPIPE, and the pipeline resolves 141 — under pipefail that `if` goes false and a first-contact mention inside an inline comment is missed. `--slurp` is not available either: `gh api` rejects it with `the --slurp option is not supported with --jq or --template`.

The fourth `--paginate` in that step — the review-comment body scan piped to `grep -qF` — is already a per-element stream and is unchanged.

## Testing

New `test_mention_verify_engagement_lookups_survive_pagination` asserts no `--paginate`d `--jq` in the verify step carries a `| length` reduction, that the three lookups capture the raw stream, and that no `-gt "0" ]` numeric comparison remains. It fails against the old template. Regtest snapshots for the mention workflow are updated; `uv run pytest` is green (328 passed).

Mechanism verified live, forcing multiple pages with `per_page=1` against this repo's PR #835 (4 bot review records):

```
$ gh api --paginate "repos/max-sixty/tend/pulls/835/reviews?per_page=1" \
    --jq '[.[] | select(.user.login == "tend-agent")] | length'
1
1
1
1
1

$ gh api --paginate "repos/max-sixty/tend/pulls/835/reviews?per_page=1" \
    --jq '.[] | select(.user.login == "tend-agent") | .id' | wc -l
5
```

The empty case is unchanged: no matching records means an empty stream, so `[ -n ... ]` is correctly false.

Only the template is touched — `.github/workflows/tend-mention.yaml` picks this up from `uvx tend@latest init` after the next release, per the generated-workflow rule in `CLAUDE.md`.

Found while reviewing #835, which fixed the same per-page-`--jq` shape in the `review` skill's `ORPHAN_ID` guard. Separate file and separate concern, so it lands separately.

### DIFF
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -275,8 +275,16 @@ jobs:
               if printf '%s\n' "$ISSUE_BODY" | grep -qF '@<<cfg.bot_name>>'; then
                 echo "should_run=true" >> "$GITHUB_OUTPUT"; exit 0
               fi
+              # Count outside jq: `gh api --paginate` applies `--jq` once per
+              # page, so a reduction inside it (`| length`) emits one count per
+              # page rather than one overall. Past 100 comments the variable
+              # holds e.g. `100\n7`, the `-gt` below errors with `integer
+              # expression expected`, and the failed test falls through to
+              # should_run=false — the bot goes quiet on its most-engaged
+              # threads. A per-element stream paginates fine; `wc -l` does the
+              # reduction over the merged output.
               BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$ISSUE_NUMBER/comments" \
-                --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
+                --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id' | wc -l)
               if [ "$BOT_COMMENTS" -gt "0" ]; then
                 echo "should_run=true" >> "$GITHUB_OUTPUT"; exit 0
               fi
@@ -294,15 +302,16 @@ jobs:
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
           fi
 
+          # Counted outside jq — see the note on the issue-comment count above.
           BOT_REVIEWS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/reviews" \
-            --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
+            --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id' | wc -l)
           if [ "$BOT_REVIEWS" -gt "0" ]; then
             echo "should_run=true" >> "$GITHUB_OUTPUT"
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
           fi
 
           BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$PR_NUMBER/comments" \
-            --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
+            --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id' | wc -l)
           if [ "$BOT_COMMENTS" -gt "0" ]; then
             echo "should_run=true" >> "$GITHUB_OUTPUT"
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -275,15 +275,14 @@ jobs:
               if printf '%s\n' "$ISSUE_BODY" | grep -qF '@<<cfg.bot_name>>'; then
                 echo "should_run=true" >> "$GITHUB_OUTPUT"; exit 0
               fi
-              # Count outside jq: `gh api --paginate` applies `--jq` once per
-              # page, so a reduction inside it (`| length`) emits one count per
-              # page rather than one overall. Past 100 comments the variable
-              # holds e.g. `100\n7`, the `-gt` below errors with `integer
-              # expression expected`, and the failed test falls through to
-              # should_run=false — the bot goes quiet on its most-engaged
-              # threads. Capture the stream and test it for emptiness — the
-              # bare substitution keeps a failing `gh api` fatal under GHA's
-              # default `bash -e`.
+              # Don't reduce inside jq: `gh api --paginate` applies `--jq` once
+              # per page, so `| length` emits one count per page rather than one
+              # overall. Past 100 comments the variable holds e.g. `100\n7`, a
+              # numeric test on it errors with `integer expression expected`,
+              # and the failed test falls through to should_run=false — the bot
+              # goes quiet on its most-engaged threads. Capture the per-element
+              # stream and test it for emptiness — the bare substitution also
+              # keeps a failing `gh api` fatal under GHA's default `bash -e`.
               BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$ISSUE_NUMBER/comments" \
                 --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id')
               if [ -n "$BOT_COMMENTS" ]; then
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -281,11 +281,12 @@ jobs:
               # holds e.g. `100\n7`, the `-gt` below errors with `integer
               # expression expected`, and the failed test falls through to
               # should_run=false — the bot goes quiet on its most-engaged
-              # threads. A per-element stream paginates fine; `wc -l` does the
-              # reduction over the merged output.
+              # threads. Capture the stream and test it for emptiness — the
+              # bare substitution keeps a failing `gh api` fatal under GHA's
+              # default `bash -e`.
               BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$ISSUE_NUMBER/comments" \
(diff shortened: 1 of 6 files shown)


tend-agent (commented):
The pagination diagnosis is right and the per-element stream is the correct shape. One regression in how the count is now reduced, plus the test change it implies.

Piping into `wc -l` moves the assignment's exit status off `gh` and onto `wc`. Verified locally against the semantics the generated step actually runs under (no `shell:` on the step, so GHA's default `bash -e {0}` — errexit, no pipefail):

```
$ bash -e -c 'X=$(false | wc -l); echo "survived, X=[$X]"'
survived, X=[0]
$ bash -e -c 'X=$(false); echo survived'; echo "rc=$?"
rc=1
```

So a `gh api` failure at any of the three sites now reads as "no engagement" on a green job, where it previously failed the step loudly. Details and the suggested shape inline.

Taking the suggestions means the four regtest snapshots need regenerating as well.

generator/src/tend/templates/mention.yaml.j2
```diff
@@ -275,8 +275,16 @@ jobs:
...
+              # page rather than one overall. Past 100 comments the variable
+              # holds e.g. `100\n7`, the `-gt` below errors with `integer
+              # expression expected`, and the failed test falls through to
+              # should_run=false — the bot goes quiet on its most-engaged
+              # threads. A per-element stream paginates fine; `wc -l` does the
+              # reduction over the merged output.
               BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$ISSUE_NUMBER/comments" \
-                --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
+                --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id' | wc -l)
               if [ "$BOT_COMMENTS" -gt "0" ]; then
```
tend-agent: `$(... | wc -l)` takes the pipeline's exit status from `wc`, not from `gh`. The step declares no `shell:`, so it runs under GHA's default `bash -e {0}` — errexit on, pipefail off. Before this change a failing `gh api` (secondary rate limit, a 5xx mid-pagination) aborted the step and the verify job went red; now it yields `0`, `should_run=false`, and a green job. That's the same silent-quiet failure this PR is fixing, just relocated from a bad count to a swallowed error.

The counts are only ever compared against zero, so drop the count and test the stream for emptiness. A bare command substitution keeps errexit live, and `[ -n ... ]` is immune to the multi-line value that started all this:

```suggestion
              # threads. Capture the stream and test it for emptiness — the
              # bare substitution keeps a failing `gh api` fatal under GHA's
              # default `bash -e`.
              BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$ISSUE_NUMBER/comments" \
                --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id')
              if [ -n "$BOT_COMMENTS" ]; then
```

`set -o pipefail` is not a safe alternative here: `grep -qF` on the review-comment scan a few lines up exits on first match, `gh` takes SIGPIPE, and the pipeline resolves 141 — under pipefail that `if` goes false and a first-contact mention in an inline comment is missed.

generator/src/tend/templates/mention.yaml.j2
```diff
@@ -294,15 +302,16 @@ jobs:
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
           fi
 
+          # Counted outside jq — see the note on the issue-comment count above.
           BOT_REVIEWS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/reviews" \
-            --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
+            --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id' | wc -l)
           if [ "$BOT_REVIEWS" -gt "0" ]; then
```
tend-agent:
```suggestion
          # Captured, not counted — see the note on the issue-comment lookup above.
          BOT_REVIEWS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/reviews" \
            --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id')
          if [ -n "$BOT_REVIEWS" ]; then
```

generator/src/tend/templates/mention.yaml.j2
```diff
@@ -294,15 +302,16 @@ jobs:
...
             echo "should_run=true" >> "$GITHUB_OUTPUT"
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
           fi
 
           BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$PR_NUMBER/comments" \
-            --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
+            --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id' | wc -l)
           if [ "$BOT_COMMENTS" -gt "0" ]; then
```
tend-agent:
```suggestion
          BOT_COMMENTS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$PR_NUMBER/comments" \
            --jq '.[] | select(.user.login == "<<cfg.bot_name>>") | .id')
          if [ -n "$BOT_COMMENTS" ]; then
```

generator/tests/test_generate.py
```diff
@@ -704,6 +704,40 @@ def test_mention_verify_skips_bot_comments_without_mention(tmp_path: Path) -> No
...
+        if "--paginate" not in line or "--jq" not in line:
+            continue
+        assert "| length" not in line, (
+            f"reduction inside a --paginate'd --jq runs per page: {line.strip()}"
+        )
+
+    assert run.count("| wc -l)") == 3, (
+        "the three engagement counts (issue comments, PR reviews, PR comments) "
+        "must reduce outside jq so pagination can't split the count"
+    )
```
tend-agent: If the three sites move to `[ -n ... ]`, this pins the new shape instead — the second assert is the one that would have caught the original bug, since `-gt` on a possibly-multi-line value is the actual defect:

```suggestion
    assert run.count("| .id')") == 3, (
        "the three engagement lookups (issue comments, PR reviews, PR comments) "
        "must capture the raw stream, so a failing `gh api` still trips errexit"
    )
    assert '-gt "0" ]' not in run, (
        "guard on an empty stream (`[ -n ... ]`) — a numeric comparison is what "
        "a split count breaks"
    )
```

generator/src/tend/templates/mention.yaml.j2
tend-agent: Applied in d20026b, across all three sites. Confirmed the regression directly under the step's actual semantics before taking it:

```
$ bash -e -c 'X=$(false | wc -l); echo "survived, X=[$X]"'
survived, X=[0]
$ bash -e -c 'X=$(false); echo survived'; echo "rc=$?"
rc=1
```

So the `wc -l` version swallowed a failed `gh api` into `should_run=false` on a green job — the same silent-quiet shape the PR is fixing, just moved from a split count to a swallowed error. The bare capture plus `[ -n ... ]` fixes both at once, and your SIGPIPE point rules out pipefail as the alternative: `grep -qF` a few lines up exits on first match, so the review-comment scan would resolve 141 and drop first-contact inline mentions.

generator/src/tend/templates/mention.yaml.j2
tend-agent: Applied in d20026b.

generator/src/tend/templates/mention.yaml.j2
tend-agent: Applied in d20026b.

generator/tests/test_generate.py
tend-agent: Applied in d20026b, and you're right that `-gt "0" ]` is the assertion that pins the actual defect — the old test only caught the reduction-inside-`--jq` half. Renamed the test to `test_mention_verify_engagement_lookups_survive_pagination` and extended its docstring to record why the stream is captured bare rather than piped, so the `wc -l` shape doesn't come back as an obvious simplification. The four regtest snapshots are regenerated in the same commit; `uv run pytest` is green (328 passed).

PR #843, opened 2026-08-05
fix(mark-notification-read): tolerate a transient run-metadata fetch failure

tend-agent:
## Problem

`shared/steps/mark-notification-read.sh` runs under `set -eo pipefail`, and its `run_started_at` fetch is unguarded:

```bash
RUN_STARTED_AT=$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}" \
  --jq '.run_started_at')
```

A transient `gh api` error there aborts the script non-zero. Both harness actions gate the step on `if: success()` (`claude/action.yaml:514`, `codex/action.yaml:325`), so it only ever runs after the agent already finished its work — and its failure turns a fully-successful run into a red job.

The very next call in the same file is already wrapped for exactly this reason ("a transient API error (non-JSON response) must not fail the composite action"), so this is the one call in the script that doesn't hold the line the file sets for itself.

Reproduced with a `gh` stub that fails only on the run-metadata call: the script exits 1.

## Solution

Guard the fetch and skip the cycle when the timestamp can't be read. Skipping is the right disposition rather than marking unconditionally: without `run_started_at` the `updated_at <= started` comparison can't be evaluated, and marking anyway would swallow activity that arrived mid-run. The scheduled `tend-notifications` poll picks up anything left unread.

## Testing

New `generator/tests/test_shared_steps.py` drives the script with a fake `gh` on `PATH` — the same technique `test_notifications_precheck_tolerates_transient_non_json` already uses for the notifications pre-check. Three cases:

- run-metadata fetch fails → exit 0, and no thread is marked read
- activity predating the run → thread marked read (unchanged behaviour)
- activity newer than the run → left unread (unchanged behaviour)

The first fails on `main` (`exit 1`) and passes with the fix. Full suite: 330 passed. `shellcheck -S warning` and `ruff check`/`format` clean.

This is the first test in the repo covering `shared/steps/`; until now those scripts had only shellcheck, which can't see runtime behaviour. The module docstring notes why they live under `generator/tests/` despite not being part of the generator package.

### DIFF
--- a/generator/tests/test_shared_steps.py
+++ b/generator/tests/test_shared_steps.py
@@ -25,7 +25,7 @@ printf '%s\\n' "$*" >> "$GH_CALLS"
 case "$2" in
   repos/*/actions/runs/*)
     [ -n "${FAIL_RUN_META:-}" ] && exit 1
-    echo "$RUN_STARTED_AT"
+    echo "$FAKE_RUN_STARTED_AT"
     ;;
   notifications)
     cat "$NOTIFICATIONS_JSON"
@@ -58,7 +58,10 @@ def gh_env(tmp_path: Path) -> dict[str, str]:
         "GITHUB_EVENT_PATH": str(event),
         "GITHUB_REPOSITORY": "owner/repo",
         "GITHUB_RUN_ID": "12345",
-        "RUN_STARTED_AT": "2026-01-02T00:00:00Z",
+        # Deliberately not `RUN_STARTED_AT`: the script assigns that name, and
+        # an inherited value would let the happy-path tests pass even if the
+        # fetched timestamp were never used.
+        "FAKE_RUN_STARTED_AT": "2026-01-02T00:00:00Z",
     }
 
 
--- a/generator/tests/test_shared_steps.py
+++ b/generator/tests/test_shared_steps.py
@@ -1,0 +1,145 @@
+"""Tests for the composite actions' shared step bodies (shared/steps/*.sh).
+
+These scripts run as `bash <script>` inside both harness actions, so a
+non-zero exit fails the step and turns an otherwise-successful agent run red.
+They aren't part of the generator package; the tests live here because this is
+the repo's only Python suite, and shellcheck (pre-commit) can't catch runtime
+behaviour.
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+from pathlib import Path
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+MARK_NOTIFICATION_READ = REPO_ROOT / "shared" / "steps" / "mark-notification-read.sh"
+
+# `gh api` stand-in. Records every invocation so a test can assert which calls
+# the script made, and fails the run-metadata fetch when FAIL_RUN_META is set.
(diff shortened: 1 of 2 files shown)


generator/tests/test_shared_steps.py
```diff
@@ -0,0 +1,145 @@
...
+MARK_NOTIFICATION_READ = REPO_ROOT / "shared" / "steps" / "mark-notification-read.sh"
+
+# `gh api` stand-in. Records every invocation so a test can assert which calls
+# the script made, and fails the run-metadata fetch when FAIL_RUN_META is set.
+FAKE_GH = """#!/usr/bin/env bash
+printf '%s\\n' "$*" >> "$GH_CALLS"
+case "$2" in
+  repos/*/actions/runs/*)
+    [ -n "${FAIL_RUN_META:-}" ] && exit 1
+    echo "$RUN_STARTED_AT"
```
tend-agent: The fake's canned value should travel under a name the script never reads, so the happy-path tests depend on the fetch actually reaching `RUN_STARTED_AT`.

```suggestion
    echo "$FAKE_RUN_STARTED_AT"
```

generator/tests/test_shared_steps.py
```diff
@@ -0,0 +1,145 @@
...
+    event.write_text(json.dumps({"issue": {"number": 7}}))
+
+    return {
+        "PATH": f"{bindir}:/usr/bin:/bin",
+        "GH_CALLS": str(tmp_path / "gh-calls.log"),
+        "GITHUB_EVENT_NAME": "issues",
+        "GITHUB_EVENT_PATH": str(event),
+        "GITHUB_REPOSITORY": "owner/repo",
+        "GITHUB_RUN_ID": "12345",
+        "RUN_STARTED_AT": "2026-01-02T00:00:00Z",
```
tend-agent:
```suggestion
        "FAKE_RUN_STARTED_AT": "2026-01-02T00:00:00Z",
```

tend-agent (commented):
The script change is right, and the `-z` arm isn't redundant: `gh api --jq` on a field the response doesn't carry yields an empty string rather than `null`, so an otherwise-well-formed degraded response is caught too (verified against the live API).

One problem in the fixture. The fake `gh`'s canned timestamp is passed in as `RUN_STARTED_AT` — the same name the script assigns to — so the script inherits the value from its environment. Both happy-path tests therefore pass whether or not the script uses what it fetched. I confirmed it: mutating the script to assign the fetch into a different variable and leave `RUN_STARTED_AT` untouched still gives `3 passed`. With the fixture variable renamed, that same mutant fails `test_mark_notification_read_marks_thread_predating_the_run`, which is the regression those two tests exist to catch. The failure-path test is unaffected either way.

tend-agent (commented):
Fixture rename addresses the earlier finding. Re-ran the mutant to confirm the tests now have the power they're meant to: with the script's fetch assigned into a different variable and `RUN_STARTED_AT` left untouched, `test_mark_notification_read_marks_thread_predating_the_run` fails (`api notifications/threads/999 -X PATCH` never issued) instead of passing on the inherited value. Unmutated, all three pass. No further concerns; self-authored so no approval from me.

PR #845, opened 2026-08-05
perf(review-reviewers): cut cadence to 3-hourly, and anchor the window to the cron period

tend-agent:
Anthropic quota exhaustion has now hard-failed the fleet **twice in three days**. The [Aug 3 diagnosis on #808](https://github.com/max-sixty/tend/issues/808#issuecomment-5164314323) pre-registered the trigger — *"I've left this to a maintainer decision rather than opening a cadence PR — it's the first occurrence and the plan-vs-cadence tradeoff is yours. If it recurs it's worth landing the cron cut"* — and the [review-runs tracking entry](https://github.com/max-sixty/tend/issues/801#issuecomment-5150655987) carries the same condition. It recurred, an order of magnitude larger. This lands option 1 from that comment.

## What happened in this window

`review-reviewers` failed **15 consecutive runs × 5 matrix legs = 75 job failures** across 14.5 hours (09:41Z Aug 4 → 23:10Z Aug 4), recovering on the first tick after the quota reset. Every leg dies ~4 s into the `claude` step with the same signature, read from the session-log artifacts:

```json
{"type":"rate_limit_event","rate_limit_info":{"status":"rejected","resetsAt":1785888000,
 "rateLimitType":"seven_day","overageStatus":"rejected","overageDisabledReason":"org_level_disabled"}}
```

```json
{"type":"result","is_error":true,"api_error_status":429,"num_turns":1,
 "result":"You've hit your weekly limit · resets 12am (UTC)"}
```

`resetsAt: 1785888000` is 2026-08-05T00:00:00Z; the first green run is [30962483562](https://github.com/max-sixty/tend/actions/runs/30962483562) at 00:11Z. This is the **seven-day** quota, not the rolling session quota that caused the Aug 3 incident — the same capacity wall, hit on a longer accounting period, which is the harder one to ride out.

## Why cadence is the lever

Normalised over the 8 runs that actually consumed tokens (the other 15 were free failures):

| Workflow | Runs w/ tokens | Cost (list) | Per run |
|---|---|---|---|
| **review-reviewers** | 8 | **$83.88** | **$10.49** |
| tend-review | 21 | $30.53 | $1.45 |
| tend-mention | 18 | $29.55 | $1.64 |
| tend-triage | 6 | $15.05 | $2.51 |
| tend-nightly | 1 | $4.07 | $4.07 |

At its designed rate — 24 runs/day — `review-reviewers` alone projects to **~$252/day**, against $79 for the entire rest of the fleet. Per-run cost is also climbing at flat cadence: $5.31 (Aug 3) → $7.86 (Aug 4) → **$10.49** (Aug 5). Both drivers push the same way, and cadence is the one that can be changed deterministically.

**Correction (2026-08-08), in reply to ["why does doing it less frequently save costs?"](https://github.com/max-sixty/tend/pull/845#issuecomment-5226128994):** cadence does not cut cost proportionally, and an earlier comment on this PR projecting ~$71/day was wrong to assume a flat per-run cost while the window triples. Per-leg cost measured across the matrix is **~$1.12 fixed + $0.12 x runs surveyed**, so ~65% of a quiet run is per-tick startup and only that part is saved. The cut is **~$213 -> ~$120/day (~40%)**, not the 3x the tick count implies. The session-count argument is unaffected: 120 -> 40 Claude sessions/day is the full 3x, and that is what saturates the seven-day quota. [Full derivation](https://github.com/max-sixty/tend/pull/845#issuecomment-5226148733).

`47 */3 * * *` takes it from 120 to 40 Claude sessions/day. Outcome-acceptance signal — was a bot PR merged, was a comment answered, did a human push back — evolves over hours to days, so little analytical value is lost at 3-hourly. The workflows that lose when the quota is gone are the user-facing ones: on Aug 3 the same exhaustion dropped the nightly sweep entirely plus two PR reviews and two mention-handles.

## The window-anchoring change is a prerequisite, not a drive-by

[`list-recent-runs.sh`](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L79-L87) only anchors its completion window when the cron matches `^([0-9]+) \* \* \* \*$`. Any other shape falls through to [a now-anchored 1-hour window](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L134-L137). Verified against the actual regex:

```
'47 * * * *'   -> hourly branch (minute=47)
'47 */3 * * *' -> FALLBACK: now-anchored 1h window
```

So cutting the cron alone would have left every run analysing one hour out of every three — a **silent two-thirds coverage hole**, reported as an all-clear. This change generalises the detection to `MM */N * * *` and derives the window from the cron period.

`N` is only accepted when it divides 24. Cron's `*/N` restarts at hour 0 each day, so a step like `*/5` fires at 0,5,10,15,20 and then wraps after 4 h — no constant period, and a floor computed from one would under-reach across midnight. Those shapes keep falling through to the now-anchored window, same as before.

**Hourly behaviour is unchanged.** `cron_hour_step` is 1 for the hourly form, so `period` is 3600 and every expression reduces to the original arithmetic. Checked exhaustively rather than by inspection — old vs new tick computation over 24 hours × 3 minute offsets × 4 cron-minute values, plus month/quarter boundaries:

```
PASS: hourly arithmetic identical across 288 timestamps x 4 minute values + boundaries
```

**Consecutive 3-hourly ticks tile exactly** — no gap, no overlap, including the midnight wrap:

```
47 */3 * * *  @ 2026-08-05T00:10Z -> window [08-04T18:47 .. 08-04T21:47]
47 */3 * * *  @ 2026-08-05T01:00Z -> window [08-04T21:47 .. 08-05T00:47]
47 */3 * * *  @ 2026-08-05T03:50Z -> window [08-05T00:47 .. 08-05T03:47]
47 */3 * * *  @ 2026-08-05T06:50Z -> window [08-05T03:47 .. 08-05T06:47]
47 */3 * * *  @ 2026-08-05T09:50Z -> window [08-05T06:47 .. 08-05T09:47]
```

**Non-conforming shapes still fall back:** `47 */5 * * *`, `17 6 * * *`, `*/15 * * * *` → now-anchored 1 h.

`shellcheck` clean; `bash -n` clean; the workflow YAML parses with the new cron.

The 6 h dropped-tick reach-back cap is deliberately left as a literal. At hourly that is unchanged; at the 3-hourly step it is two full periods, so a single dropped tick is still recoverable — and leaving it untouched keeps this diff off the lines #838 rewrites.

</details>

## Overlap with #838

#838 also edits `list-recent-runs.sh`, in the dropped-tick recovery block (anchor on the last *successful* run; raise the fetch limit). Different concern, and the two are complementary — that one widens the window after an outage, this one sets the window's period. The edits are textually adjacent, so whichever lands second wants a small rebase; neither depends on the other.

## Gate assessment

- **Evidence level**: Critical/High — 75 job failures in one window, 13 in the Aug 3 window, both traced to the same capacity wall in session logs.
- **Occurrences**: 2 distinct exhaustion events in 3 days (2026-08-03, 2026-08-04/05), plus a pre-registered "act on recurrence" from the prior run.
- **Structural**: yes. No decision point — sustained demand above the quota produces the same failure every time, and the failure is invisible in the sense that starved runs are marked `failure` having silently not done their work.
- **Change type**: targeted fix (config + the script change it requires).

## What this does not address

The failures themselves are annotated only as `claude -p exited non-zero (exit=1)` with empty stderr — the 429 text sits in the stream-json the action already has, and is discarded on the non-zero-exit path. That is #818, open since Aug 3, and it is why diagnosing this needed a second round of manual session-log archaeology. Outage tracker #831 now carries 78 rows and no cause. Cadence reduces how often the wall is hit; #818 makes it legible when it is.


tend-agent (commented):
The cadence cut is well-evidenced and the window generalisation holds up — I re-derived the tiling independently (exact tiling at 3-hourly including the midnight wrap, byte-identical arithmetic on the hourly path across 24h × 3 minute offsets × 4 cron-minute values, and `*/5` / `MM H * * *` / `*/15 * * * *` still falling through). Three gaps, one of which is the PR's own failure mode surviving in a second place.

**The "past hour" framing survives in the skill prompt.** The PR's thesis is that cutting the cron without generalising the window leaves a silent two-thirds coverage hole. `review-reviewers/SKILL.md` updates the frontmatter `description` but keeps two hourly assumptions in the body that the run list no longer matches:

- The opening instruction — *"Analyze tend's CI behavior on the target repo over the past hour"* (immediately under `# Review Reviewers`). This is the agent's own framing for the whole analysis, so it now contradicts the 3-hour run list it is handed in Step 1.
- The Step 2 survey-subagent prompt — *"`tend-notifications`: check for recent bot comments/issue-close events in the past hour"*. Unlike the other three mappings in that list, this one has no run-scoped anchor (no `headBranch`, no triggering comment) — it reconstructs outcomes purely from the time window. A subagent told "past hour" while the run list spans three will under-survey `tend-notifications` outcomes by exactly the two-thirds this PR exists to prevent.

Both want to be period-agnostic ("the analysis window", "the window covered by the run list") rather than restamped to "past 3 hours" — otherwise the next cadence change reopens the same hole.

**The script's header docstring wasn't carried along with the detection comment.** The comment at the detection site is generalised, but the module header above `set -euo pipefail` still documents the old contract: *"whose updatedAt falls within a 1-hour completion window"*, *"a simple hourly cron (`MM * * * *`)"*, *"tile exactly: [intended-1h, intended], then [intended, intended+1h]"*, and *"For non-schedule events or non-hourly crons, falls back to a now-anchored 1h window"* — the last of which now reads as excluding the very shape the PR adds. `# Fetches runs started in the past 3 hours` is stale too: `CREATED_SINCE` is `COMPLETED_AFTER - 7200`, so at the 3-hourly step it reaches five hours back. That header is what a reader hits first, so it's the one that misleads.

I've pushed a commit fixing all three (the third is the inline suggestion below) — it's doc-only and doesn't touch the lines #838 rewrites, so the rebase note in the PR body still stands.

One thing I checked and am *not* flagging: the widened `CREATED_SINCE` span makes the per-workflow `--limit 50` truncate silently if it ever binds, but current density on this repo is 4–10 runs per workflow per 5h, so there's an order of magnitude of headroom.

plugins/tend-ci-runner/scripts/list-recent-runs.sh
```diff
@@ -104,11 +122,13 @@ if [ -n "$cron_minute" ]; then
...
   # there. When every tick fires, the previous run's intended tick == the
-  # default (intended - 3600), so this is a byte-identical no-op — still no
+  # default (intended - period), so this is a byte-identical no-op — still no
   # overlap between consecutive cycles. When a tick was dropped, it reaches
-  # back to cover the orphaned hour. Capped at 6h so a sustained outage can't
-  # create an unbounded window. The analyzing workflow runs on the current
-  # repo, so this query omits TARGET_REPO's -R.
+  # back to cover the orphaned period. Capped at 6h so a sustained outage can't
+  # create an unbounded window — one full period at hourly, and still two at
+  # the 3-hourly step, so a single dropped tick stays recoverable either way.
```
tend-agent: `21600 / 3600` is six, not one — at hourly the 6 h cap reaches back six full periods. The "still two at the 3-hourly step" half is right (`21600 / 10800`), and it's the binding one, so the conclusion holds; only the hourly figure is off.

```suggestion
  # create an unbounded window — six full periods at hourly, and still two at
  # the 3-hourly step, so a single dropped tick stays recoverable either way.
```

tend-agent: Right — applied in 8631cab. `21600 / 3600` is six periods at hourly; the 3-hourly figure (two) was the binding one and is unchanged.

tend-agent:
All three points are in 8631cab. Re-grepping the two files afterwards turned up a fourth hourly assumption the review missed, now fixed in 3a1d13a: Step 5 told the agent to open its PRs on `hourly/review-$GITHUB_RUN_ID`, so every PR this workflow files would have carried a cadence in its branch name that stopped being true with this change. Prefixed with the workflow name instead of the cadence, which is the same period-agnostic treatment the opening instruction and the Step 2 survey prompt got — restamping it to `3-hourly/` would just defer the problem to the next cadence change.

Nothing filters on that prefix (single occurrence across the repo), so it was cosmetic rather than a coverage hole. The commit is doc-only and still clear of the lines #838 rewrites.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -232,7 +232,7 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
 > - `tend-review`: `gh -R $ARGUMENTS run view <run-id> --json headBranch` → find PR via
 >   `gh -R $ARGUMENTS pr list --head <branch> --state all` → check bot reviews via
 >   `gh api repos/$ARGUMENTS/pulls/<pr>/reviews`
-> - `tend-notifications`: check for recent bot comments/issue-close events in the past hour
+> - `tend-notifications`: check for bot comments/issue-close events within the analysis window (this mapping has no run-scoped anchor, so it reconstructs outcomes from the window alone — use the window's actual span, not a fixed hour)
```
tend-agent: "Use the window's actual span" isn't resolvable from inside this prompt. The subagent is handed `[run IDs from Step 1]` and `$BOT_LOGIN`, nothing else, and `list-recent-runs.sh` prints only `{databaseId, conclusion, createdAt, updatedAt}` per run — the window's floor is never emitted. So the one mapping this note singles out as having no run-scoped anchor is now told to use an anchor it can't obtain. "Past hour" was wrong but concrete; this is unresolvable, which fails the same way (a short survey reported as complete) without the tell.

The corruption-scan recipe further down this same prompt already depends on the same absent value — four `--jq` filters interpolate a literal `<window-start>` with no stated source — so publishing the window once fixes both.

Worth noting that reconstructing it from the run list's own min/max `updatedAt` isn't equivalent: that's the span of runs that happened to fire, it collapses toward a point when only one run lands in the period, and it's empty exactly when the list is empty and the all-clear gets recorded.

tend-agent (commented):
One finding on the incremental plus a smaller note; both land on the same missing piece. Green CI, and self-authored so this is a COMMENT rather than an approval.

**Secondary — `workflow_dispatch` now under-covers by the same two-thirds this PR closes.** A dispatch event carries no `.schedule`, so it takes the fallback branch and gets a now-anchored 1 h window. Under `47 * * * *` that happened to equal one period, so manual and scheduled runs saw the same span; at `47 */3 * * *` a dispatched run analyzes one hour of a three-hour period, and Step 1's "if empty, record the run as all-clear" means that under-coverage can be written down as a clean window. Dispatch is the natural way to re-run after a dropped tick or a quota outage, which is when coverage matters most. I don't think the dispatch path should guess a period — but surfacing the window (below) at least makes the short span legible in the log instead of silent.

Both points reduce to: the window is computed and never published. `list-recent-runs.sh` emits only the per-run JSON array; `COMPLETED_AFTER` is used by the final `jq` filter and discarded. One line after the `if`/`else` that sets it, before the fetch loop, would publish it without disturbing the stdout contract:

```bash
echo "Completion window: >= $(date -u -d "@$COMPLETED_AFTER" +%Y-%m-%dT%H:%M:%SZ)" >&2
```

Step 1 then tells the agent to carry that value forward, and Step 2 substitutes it for `<window-start>` and for the new line-235 wording. Happy to push it if you want — it touches lines outside this diff, so I've left it as an offer.

Nothing else new: the four hourly assumptions are correctly retired, `21600 / 3600` reads six now, and the `hourly/` branch prefix has no other references in the repo (`review-runs` keeps `daily/`, which is still true of its cron).

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Correct, and the second half is the sharper point — the corruption scan already interpolates a literal `<window-start>` into five `--jq` filters with no stated source, so that placeholder was shipping into commands the subagent runs. My wording made a second consumer depend on the same absent value.

Taken the offer: b9523c6 publishes it. `list-recent-runs.sh` prints `Completion window: >= <timestamp>` to stderr after the `if`/`else` sets `COMPLETED_AFTER`, so both branches emit and stdout stays the run-list JSON. Step 1 tells the agent to note it; Step 2's preamble binds it once for the whole prompt (`substitute that timestamp everywhere <window-start> appears below`), which covers the `tend-notifications` mapping and all five scan filters together.

Agreed on not reconstructing from min/max `updatedAt` — I didn't take that route, for the reasons you give.

Verified end-to-end rather than by inspection: `TARGET_REPO=max-sixty/tend ./list-recent-runs.sh "tend-" "review-"` → `Completion window: >= 2026-08-05T08:16:35Z` on stderr, stdout parsing as a 15-element array. `shellcheck` and `bash -n` clean.

tend-agent:
Took the offer — b9523c6 publishes the window, which closes both the `workflow_dispatch` under-coverage note and the inline finding on Step 2.

`list-recent-runs.sh` now prints `Completion window: >= <timestamp>` to stderr, placed after the `if`/`else` that sets `COMPLETED_AFTER` so the cron and fallback branches both emit. Step 1 tells the agent to note it and flags that a dispatched run covers an hour rather than a period; Step 2's preamble binds it once for the whole prompt, which resolves the `tend-notifications` mapping and the corruption scan's five `<window-start>` filters in one place. The dispatch path still doesn't guess a period — the short span is legible instead of silent, which was the proposal.

The `<window-start>` half was the more serious of the two: a literal placeholder with no stated source was being interpolated into commands the survey subagent runs, independent of any cadence change.

Ran it rather than inspecting it:

```
$ TARGET_REPO=max-sixty/tend ./plugins/tend-ci-runner/scripts/list-recent-runs.sh "tend-" "review-"
stderr: Completion window: >= 2026-08-05T08:16:35Z
stdout: valid JSON array, 15 runs
```

That run took the fallback branch (this session is not a `schedule` event), which is the path the note is about. `shellcheck` and `bash -n` clean. No call site captures the script's stderr into stdout — the only invocations are the Step 1 snippet and the usage line in tend's `running-tend` overlay — so the run-list contract is unchanged.

The module header now documents the stderr line alongside the window contract.

</details>

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -213,6 +213,8 @@ TARGET_REPO=$ARGUMENTS ${CLAUDE_PLUGIN_ROOT}/scripts/list-recent-runs.sh
 
 The script discovers `tend-*` workflows by default. Pass additional prefixes as arguments to include other workflows (e.g., `review-reviewers` when analyzing tend itself).
 
+It prints `Completion window: >= <timestamp>` on stderr. **Note that timestamp — it is the analysis window's floor**, and Step 2 needs it both for the survey prompt and for the corruption scan's `--jq` filters; nothing downstream can reconstruct it from the run list. The window is one cron period wide on a scheduled run, but only one hour on any other event (a manual `workflow_dispatch` takes that path), so on a dispatched run expect it to cover less than a full period.
```
tend-agent: "one cron period wide on a scheduled run" is the default floor only. Dropped-tick recovery pulls the floor back to the previous actual run's intended tick, capped at `intended - 21600`, so a scheduled run can legitimately publish a window up to 6 h wide — two periods at the new 3-hourly cadence. Worth hedging, since the sentence is what the agent will reason from when it decides how much ground the window covers.

```suggestion
It prints `Completion window: >= <timestamp>` on stderr. **Note that timestamp — it is the analysis window's floor**, and Step 2 needs it both for the survey prompt and for the corruption scan's `--jq` filters; nothing downstream can reconstruct it from the run list. The window is normally one cron period wide on a scheduled run — wider when the script recovers a dropped tick — but only one hour on any other event (a manual `workflow_dispatch` takes that path), so on a dispatched run expect it to cover less than a full period.
```

tend-agent (commented):
Both open findings from the last pass are closed on the incremental: the survey prompt now names the window's source and tells the subagent to substitute it, which covers all five `<window-start>` interpolations in the corruption scan (they're inside the same prompt block), and Step 1 now flags the `workflow_dispatch` under-coverage rather than leaving it silent. I checked the published floor's format against the `--jq` string comparisons it feeds — `+%Y-%m-%dT%H:%M:%SZ` sorts lexically against GitHub's `created_at`/`submitted_at`, so `>` does what it looks like it does — and the `echo` sits after the `if`/`else`, so both branches emit.

One nit inline on the new width claim.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Applied in 3d82e52. Right that the default floor isn't the maximum — `prev_intended` overrides `intended - period` when a tick was dropped, bounded only by the `intended - 21600` cap, so at 3-hourly a scheduled window can legitimately span two periods.

tend-agent:
Adding a measurement this PR has not had until now: the 24 h window 2026-08-07T08:02Z → 2026-08-08T08:02Z is the **first clean one** in the `review-runs` evidence log — 257 completed runs, **zero failures**, and all 24 hourly `review-reviewers` ticks fired, succeeded, and were metered. Every earlier window this PR could have been decided on was suppressed by an outage: Aug 3 session quota (13 runs), Aug 5 weekly quota (75 jobs), Aug 6 rate-limit blackout (36 runs), Aug 7 Actions incident (13 jobs). Aug 4 had no failures but only 20 ticks landed.

So this is the steady-state cost of `47 * * * *` across the 5-repo matrix, not a number diluted by zero-token runs:

| | this window |
|---|---|
| `review-reviewers` runs | 24 (24 success) |
| `review-reviewers` cost | **$212.79** across 2,967 turns |
| per run | $8.87 |
| fleet total | $325.44 |
| share | **65%** — highest of the six consecutive windows this has been recorded in (70 / 59 / — / 59 / 55 / 65) |

At that rate the hourly cadence alone is ~$6,400/month at API list prices. This PR's `47 */3 * * *` takes 24 ticks/day to 8, so on the same $8.87/run it lands near $71/day.

On what the cadence buys: all six PRs the workflow produced in the window came from **two** of the 24 runs — [31160773238](https://github.com/max-sixty/tend/actions/runs/31160773238) (#882, #884, #885) and [31164742947](https://github.com/max-sixty/tend/actions/runs/31164742947) (#891, #892, #893), which are also the window's two most expensive runs at $15.28 and $18.36. The other 22 appended to their per-target evidence gists and produced no PR or issue, which is the intended below-threshold path — the gist for `max-sixty/tend` has a `## Run` heading for nearly every hourly tick in the window. Cutting to 3-hourly reduces the evidence sampling rate proportionally; it does not appear to cost findings at the rate they currently clear the gates.

Recording here rather than opening a competing PR — this is the existing lever, open since 2026-08-05, and it is `MERGEABLE` (nightly merged `main` into it at 06:30Z today).

max-sixty:
why does doing it less frequently save costs?

tend-agent:
Because most of what a run costs is paid at startup, per *tick*, and has nothing to do with how much window it analyses. Cutting ticks removes that fixed part; it does not remove the analysis work, which just gets batched into fewer, bigger runs.

That also means **the projection in the PR body is wrong**, and your question is the reason I checked. It said "on the same $8.87/run it lands near $71/day" — that assumes per-run cost stays flat while the window triples, which is exactly the thing that can't be true. Measured properly the cut is **~$213 → ~$120/day, about 40%**, not the 67% implied by 24 ticks → 8.

The 5-repo matrix is a natural experiment for the split: every leg pays the same startup (skill load, bot-login resolution, fetching the target's `running-tend`, gist read + append, tracking-issue lookup) but surveys wildly different numbers of runs. If cost tracked the window, the dormant repos would be near-free. They aren't — in the 08:02Z tick, `numbagg/numbagg` (5 runs surveyed) cost $1.68 while `max-sixty/tend` (6 runs, but the busiest repo overall) cost $1.71.

Regressing cost on runs-surveyed across 10 legs (two ticks, 08:02Z and 09:59Z):

**cost per leg ≈ $1.12 fixed + $0.12 × runs surveyed** — so at typical density **~65% of an hourly run is fixed overhead**.

| | hourly (`47 * * * *`) | 3-hourly (`47 */3 * * *`) |
|---|---|---|
| runs surveyed per leg | 5.1 | 15.3 |
| cost per leg | $1.73 | $2.94 |
| cost per tick (5 legs) | $8.62 | $14.70 |
| ticks/day | 24 | 8 |
| **per day** | **$207** | **$118** |

The model lands on $207/day for the hourly cadence against the **$212.79 actually measured** in the clean window, which is the main reason I trust it.

So the mechanism is just: 24 ticks/day pay the ~$1.12 floor 24 times (~$134/day of pure startup across the matrix); 8 ticks pay it 8 times (~$45). That ~$90/day *is* the saving. The $0.12/run survey work is unchanged — the same runs get analysed either way, three hours at a time instead of one.

Worth being explicit that this weakens the cost argument for the PR by a third, though I think it still clears: ~$90/day is real, and the quota-exhaustion case — 120 Claude sessions/day saturating the seven-day limit and hard-failing the user-facing workflows — rests on session *count*, which does drop by the full 3×, independent of dollars.

<details><summary>Method, data, and caveats</summary>

Per-leg costs come from each matrix job's own `token-usage.json` inside the `claude-session-logs-*` artifacts (the token report aggregates the 5 legs into one number, which hides exactly this). Legs mapped to target repos by grepping the session JSONL; runs-surveyed counted with `gh run list` over each leg's actual completion window, filtered to the `tend-*`/`review-*` workflows [`list-recent-runs.sh`](https://github.com/max-sixty/tend/blob/main/plugins/tend-ci-runner/scripts/list-recent-runs.sh) feeds the survey.

Run `31247520175` (window 07:02–08:02Z):

| repo | runs surveyed | cost | turns |
|---|---|---|---|
| max-sixty/cargo-affected | 3 | $1.29 | 17 |
| numbagg/numbagg | 5 | $1.68 | 22 |
| max-sixty/tend | 6 | $1.71 | 23 |
| PRQL/prql | 6 | $2.37 | 31 |
| max-sixty/worktrunk | 14 | $2.75 | 27 |

Run `31251792896` (window 08:59–09:59Z):

| repo | runs surveyed | cost |
|---|---|---|
| numbagg/numbagg | 2 | $1.18 |
| PRQL/prql | 2 | $1.34 |
| max-sixty/worktrunk | 3 | $1.62 |
| max-sixty/cargo-affected | 4 | $1.91 |
| max-sixty/tend | 6 | $1.40 |

Fit on all 10 points: fixed $1.12, variable $0.119/run, R² 0.70. Dropping `worktrunk`-at-14 (the high-leverage point) gives fixed $1.06, variable $0.135/run — the conclusion doesn't rest on it. The turns column corroborates the floor independently: even the 3-run leg needs 17 turns.

Caveats, both pushing the estimate around by less than the headline gap:

- **The fit is on quiet runs.** None of these 10 legs escalated to session logs or opened a PR. The window's two expensive runs ($15.28, $18.36) were findings-driven, and findings track how much bot activity happened, not how many ticks looked at it — so that escalation cost largely persists at 3-hourly. That pushes the real saving slightly *below* 40%, maybe to ~37%.
- **Some of what I'm charging as variable is really per-window.** The repo-wide sweep in Step 2 takes no run ID, so it costs the same whether the window holds 5 runs or 15. To the extent that's inside the $0.12/run slope, batching wins a little more than the linear model credits. That pushes the other way.

n=10 across two ticks, so treat $1.12 ± a bit rather than a precise constant. The structural claim — a large per-tick floor, roughly two thirds of a quiet run — is what the two independent fits agree on, and it's the part the argument needs.

</details>

tend-agent:
The quota wall came back last night, ~14 h after the comment above called the preceding 24 h the first clean window this PR could be decided on. That window did not hold a day.

This one was the **rolling session quota**, not the seven-day one: every agent invocation from `22:24:22Z` to the `00:00Z` reset died on the synthetic `You've hit your session limit · resets 12am (UTC)`, a **1 h 36 min** blackout. Smaller than the Aug 4 weekly outage by an order of magnitude, and the fourth exhaustion in six days.

What it actually cost, all of it verified from the run list and the session-log artifacts:

| | |
|---|---|
| `review-reviewers` ticks lost | 2 — runs [31282882062](https://github.com/max-sixty/tend/actions/runs/31282882062) (22:57) and [31285087877](https://github.com/max-sixty/tend/actions/runs/31285087877) (23:56), **10 legs**, every one dead ~60 s in |
| analysis coverage lost | 2 h × 5 target repos, with no recovery path other than the next tick widening its own window |
| `tend-review` affected | 1 — [31281456692](https://github.com/max-sixty/tend/actions/runs/31281456692) on #904. **No review lost**: it had already posted its APPROVE at 22:22:42Z and died in the post-approval CI poll |
| outage rows on #905 | 11 (1 + 10 comments), which is #809's flood, still open |

The attribution is the part worth adding here. Metering the five hours preceding the first refusal (`17:24Z–22:24Z` — the window the rolling quota was actually accumulated over):

| | runs | cost |
|---|---|---|
| `review-reviewers` | 5 | **$45.10** |
| everything else (`tend-review`, `tend-mention`, `tend-notifications`) | 33 | $39.19 |

**53.5% of the window's spend from 13% of its runs** — and that is against an unusually busy fleet half, three PRs (#902/#903/#904) generating 33 metered runs in five hours. On a quiet evening the share would be higher, not lower.

The asymmetry that makes this a cadence argument rather than a cost argument: the workflow that drove the wall was also the only permanent casualty of it. `tend-review` recovered — its approval was already on the PR — and `tend-mention` and `tend-triage` had nothing queued. The two hours of `review-reviewers` coverage across five repos are simply gone. Cutting ticks lowers both the odds of hitting the wall and the amount of work stranded behind it when it is hit.

Recorded in the [evidence gist](https://gist.github.com/e08f6e62d6478163cb425a75648eb7e4); no new PR, since this one already proposes the change.

max-sixty:
Superseded by #966, which pauses `review-reviewers` outright — the cadence question stops applying once the cron is gone. The measurement in this PR is what made the case for the pause: ~65% of a quiet run is per-tick startup, so the sweep was spending most of its budget on booting rather than analyzing, and it was the fleet's dominant token line at $10.49/run.

The window-anchoring half doesn't survive either: the workflow now runs on `workflow_dispatch`, which takes the now-anchored window by design, and the cron-parsing generalization was aimed at a schedule that no longer exists.

PR #847, opened 2026-08-05
fix(generator): restore local setup composites before POST cleanup

### LINKED ISSUE #846: tend-review: POST cleanup of the local <setup> composite fails with 'Index was out of range' when a PR changes its step count (post-adoption variant of #561)
## Problem

`tend-review`'s POST-job cleanup of the adopter's local `./.github/actions/<setup>` composite fails when the PR **changes the number of steps** in that composite:

```
##[error]Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
```

This is the sibling of #561. That issue's trigger was "PR head branch *lacks* the setup directory" (→ `Can't find 'action.yml'`), and it was closed on the reasoning that it self-clears as pre-adoption branches decay out. This variant has a different trigger and does **not** self-clear: it fires on branches created long after tend adoption, and recurs every time an adopter edits the step list of their setup composite. #561's close-out named this exact condition — *"No fix worth the complexity unless we see it on PRs that branched after tend adoption, which I'd treat as a separate report."* This is that report.

## Mechanism

`tend-review.yaml` (0.1.13) resolves the local composite against the **base** tree, then lands the PR's tree over the same workspace:

```yaml
- uses: actions/checkout@v7          # pull_request_target, no ref -> base tree
- uses: ./.github/actions/tend-setup # loaded from base: N steps, post chain cached
- uses: actions/checkout@v7          # ref: refs/pull/N/merge, clean: false -> PR tree
  with:
    ref: ${{ steps.pr_ref.outputs.ref }}
    clean: false
- uses: max-sixty/tend/claude@0.1.13
```

At job end the runner re-opens the composite from the workspace to dispatch the POST steps of its nested actions (at minimum `Swatinem/rust-cache@v2`). The file on disk is now the PR's version with a different step count than the one whose post chain was cached, and the index lookup goes out of range. This is upstream [actions/runner#2816](https://github.com/actions/runner/issues/2816) — same stack frame (`GitHub.Runner.Worker.ActionManager.LoadAction`), and its diagnosis is explicit: *"caused by a mismatch in the cached number of steps for the local composite action. Calling checkout causes the number of steps to change."*

So #561 and this are one root cause — the workspace is switched out from under a cached local-composite handle — surfacing as two different errors depending on whether the PR *removes* the action file or merely *resizes* it.

## Evidence

PRQL/prql [#6144](https://github.com/PRQL/prql/pull/6144), a PR that adds one step to `.github/actions/tend-setup/action.yaml` (6 steps on `main` → 7 on the branch). Both pushes failed identically:

| Run | Head | POST result |
| --- | --- | --- |
| [30991118338](https://github.com/PRQL/prql/actions/runs/30991118338) | `e8c30adc` | `Index was out of range` in `Post Run ./.github/actions/tend-setup` |
| [30991703955](https://github.com/PRQL/prql/actions/runs/30991703955) | `1745b9df` | same |

In both, every preceding step succeeded — including `max-sixty/tend/claude@0.1.13` and its own POST — and the review was posted normally. Only the composite's POST failed, so the job's `conclusion` is `failure` while the session did its job correctly. The ten preceding `tend-review` runs on other branches (nine green, one cancelled) all passed the same POST step; the branch under test is the only variable.

The branch (`daily/review-runs-30990192946`, created 2026-08-05) descends from the commit that installed tend, so the #561 decay argument doesn't reach it:

```
$ git merge-base --is-ancestor d77a6a42 HEAD && echo "branched after tend adoption"
branched after tend adoption
```

## Why this one doesn't self-limit

The setup composite is adopter-owned, and it gets edited: pinned-version bumps, cache tweaks, new tooling. PRQL's has been touched five times since adoption, before this PR. Any such PR that adds or removes a *step* (rather than editing within one) paints a red X on `tend-review` for the life of the PR, in the one place a maintainer is most likely to read it as "the bot's review failed". It clears on merge — once base and head agree again — which also means it can't be reproduced after the fact from the merged tree.

There's a second-order cost: it trains adopters to treat a red `tend-review` as noise.

## Possible fix

#561 listed three shapes. Its option 1 generalizes to cover both variants — restore the setup directory from the base tree once the session is done, before the POST chain walks:

```yaml
- name: Restore setup action for POST cleanup
  if: always()
  run: git checkout "$GITHUB_SHA" -- .github/actions/tend-setup/
  shell: bash
```

Main steps all run before the POST chain, so an `always()` step in last position is early enough. This puts back both the file's *existence* (#561's case) and its *step count* (this one), and is inert when the PR doesn't touch the composite.

I have not tested this patch — I can't exercise tend's workflow generation from PRQL's CI, and the failure only reproduces while a PR is open against a differing base. Options 2 (relocate the composite before the second checkout) and 3 (ship setup as a remote action) both also dissolve the root cause and may be preferable; as in #561 I don't think there's an obvious winner without a call on which contract to preserve.

The same pattern is in `tend-mention.yaml`, which does the equivalent `gh pr checkout` after setup, so whatever shape is chosen wants to apply to both.

tend-agent:
## Problem

`tend-review` and `tend-mention` both load the adopter's local `setup:` composite from the workspace and then land the PR's tree over that same workspace. At job end the runner re-reads the composite's action file to dispatch the POST steps of the actions nested inside it (`Swatinem/rust-cache` and friends), indexing into the step list it cached at load time. The file on disk is now the PR's version, so:

- a PR that **changes the composite's step count** fails cleanup with `Index was out of range. Must be non-negative and less than the size of the collection.` — #846
- a PR whose head tree **lacks the composite entirely** fails with `Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under ...` — #561

One root cause, two surface errors: [actions/runner#2816](https://github.com/actions/runner/issues/2816). Both paint the job red after a session that reviewed the PR or answered the mention correctly, in the one place a maintainer reads a red X as "the bot failed". #561 was closed as self-limiting because its trigger (branches predating tend adoption) decays out; the step-count variant does not — the setup composite is adopter-owned and gets edited, so it recurs on any PR that adds or removes a step in it.

## Solution

Emit an `always()` step below the tree switch in `tend-review` and `tend-mention` that restores each local `setup:` path from `$GITHUB_SHA` before the POST chain walks:

```yaml
      - name: Restore local setup actions for POST cleanup
        if: always()
        run: |
          dir=.github/actions/tend-setup
          git checkout "$GITHUB_SHA" -- "$dir" ||
            echo "::warning::could not restore $dir from $GITHUB_SHA; POST cleanup of the local action may fail"
```

This is option 1 from #561, generalized to cover both variants — it puts back the file's existence *and* its step count.

`$GITHUB_SHA` is the right source by construction: both workflows give their first `actions/checkout` no `ref`, so it resolves the triggering event's own commit, which is the tree the composite loaded from. Every main step runs ahead of every POST step, so a step in last position is early enough. The restore is tolerant — a path that can't be resolved warns rather than failing the job, so it never converts a working run into a red one.

`_restore_local_actions_run()` collects the `uses: ./...` entries from `setup:`, shell-quoted and de-duplicated. Adopters whose setup names only remote actions (tend's own, `astral-sh/setup-uv@v6`) render no step at all, so nothing changes for them. The four workflows that never switch trees after setup — triage, ci-fix, scheduled, notifications — are untouched.

## Testing

- New `test_local_setup_action_restored_for_post_cleanup` reproduces the bug: it asserts review and mention emit a restore step, guarded by `if: always()`, positioned *after* the step that switches the tree. It failed on both before the fix.
- `test_no_restore_step_without_a_local_setup_action` pins the no-op case for remote-only setups.
- Regtest snapshots for review and mention with a local composite lock the rendered shell.
- Full generator suite: 335 passed. `ruff` (0.14.11, the pinned rev) check and format clean, and the full `pre-commit run --all-files` set passes.
- Generated a full workflow set against a config with two local composites and one remote, and ran `actionlint` v1.7.10 over it (which runs shellcheck on inline `run:` blocks): clean. An earlier `for dir in ...` shape tripped SC2043 on the single-path case, hence the one-line-per-path rendering.

The path reaches both the `git` operand and the warning text through a shell variable rather than being interpolated twice: `shlex.quote` makes a safe `git` operand, but the warning is a double-quoted string where a `$` would expand and a `"` would end it early — failing the one step whose job is never to turn a working run red. Verified by rendering `uses: './weird/$HOME"; rm -rf /; echo "'` and running the result under `set -x`: `git` receives it as one literal argument, `$HOME` unexpanded. Pinned by `test_restore_step_quotes_the_setup_path`.

I have not exercised this against a live PR that resizes a setup composite — the failure only reproduces while such a PR is open against a differing base, and the fix ships to adopters at the next release, not from this branch. The mechanism above is traced from the runner issue and the template step order rather than observed end-to-end.

## Known gaps, filed as #848

The scoping to review and mention is a claim about the *templates*. The agent can switch the tree at runtime on any workflow — nightly's bot-PR-conflict pass opens with `gh pr checkout` — so an adopter with a local composite can hit the same error there. Generalizing isn't mechanical: the other four check out `ref: <default_branch>`, which resolves at fetch time while `$GITHUB_SHA` is pinned at trigger time, so reusing this macro would restore a wrong-but-plausible tree. #848 carries that reasoning and the candidate shape. It also records the second gap: the restore is recursive over the named directory, so an action nested *inside* the composite's own tree is covered, but a sibling `uses: ./elsewhere` is not.

---
Closes #846 — automated triage


tend-agent (commented):
The mechanism traces cleanly and the `$GITHUB_SHA` argument holds for both callers: under `pull_request_target` `GITHUB_SHA` is the base-branch tip and `github.ref` is `refs/heads/<base>`, and `actions/checkout` with no `ref` fetches `+<github.sha>:refs/remotes/origin/<base>` — so the restore source is by construction the tree the composite loaded from. Same for mention's `issues`/`issue_comment`/`repository_dispatch`, where both resolve to the default branch. The object survives the second checkout too (checkout disables gc, and the `+refs/pull/N/merge:refs/remotes/pull/N/merge` refspec prunes nothing under `refs/remotes/origin/`). Suite is green locally at 334.

Nothing blocking; a few things worth weighing.

**"Never switch trees" is a YAML-level claim, not a runtime one.** The PR body scopes the fix to review and mention because the other four workflows don't swap the workspace after `setup:`. That's true of their *templates*, but not of what the agent does inside the agent step. `nightly`'s bot-PR-conflict routine opens with `gh pr checkout <number>` ([nightly/SKILL.md](https://github.com/max-sixty/tend/blob/b6a25a099856143d91d2015995221b379f103074/plugins/tend-ci-runner/skills/nightly/SKILL.md#L124)), and `running-in-ci`'s "Pushing to PR Branches" hands every workflow the same recipe. An adopter whose `setup:` names a local composite gets the identical `Index was out of range` on `tend-nightly` the moment it checks out a bot PR that resized that composite — which is exactly the shape of PR the nightly conflict-resolution pass exists to handle.

Generalizing isn't a matter of dropping the macro call into the other four templates, though, and that's the part worth recording: those workflows check out `ref: <default_branch>`, which `actions/checkout` resolves to `origin/<default>` **at fetch time**, while `$GITHUB_SHA` on a `schedule` event is pinned at trigger time. A push landing in that window makes the two differ, and restoring the wrong-but-plausible version of the composite reintroduces the step-count mismatch the step exists to prevent. The macro would need the restore source parameterized to whatever ref the calling checkout used.

**Nested local actions aren't covered.** The restore replays only the paths named in `setup:`. A composite that itself does `uses: ./.github/actions/inner` gets its own POST re-read of `inner`, and unless `inner` happens to live under the restored directory it's still whatever the PR's tree put there. Narrow, but it's the same failure with a second hop.

Minor: `shlex.quote` protects the `git checkout` operand, but the same quoted string is interpolated into a double-quoted `echo` on the next line, so the quoting doesn't hold there — a `$` in the path expands into the warning text, and a `"` would terminate the string early and fail the step whose whole job is to not turn a working run red. Realistically nobody names a directory that way, and fixing it means moving the path into a shell variable, which churns both new snapshots — so it's a judgement call whether it's worth the diff.

tend-agent:
Thanks — taking these in reverse order of how cheap they were.

**Quoting.** Fixed in d85aa96. The path now goes through a shell variable, so the raw value appears exactly once, inside single quotes, and both the `git` operand and the warning text read `"$dir"`:

```
dir=.github/actions/tend-setup
git checkout "$GITHUB_SHA" -- "$dir" ||
  echo "::warning::could not restore $dir from $GITHUB_SHA; POST cleanup of the local action may fail"
```

You were right that it's the step whose whole job is not to turn a working run red, which is what made it worth the snapshot churn rather than a judgement call. Verified rather than reasoned: rendering `uses: './weird/$HOME"; rm -rf /; echo "'` and running the result under `set -x` shows `git` receiving that path as one literal argument with `$HOME` unexpanded, and shellcheck is clean on it. Pinned by `test_restore_step_quotes_the_setup_path`.

**Runtime tree switches and nested siblings.** Both real, both filed as #848 rather than folded in here. The nightly case is the one that convinced me it's worth writing down — an adopter with a local composite hits the identical error the moment nightly's conflict pass checks out a bot PR that resized it, and that's precisely the PR shape that pass exists to handle.

I kept it out of this PR because your own diagnosis is the reason it isn't mechanical: the other four check out `ref: <default_branch>`, which resolves at fetch time, while `$GITHUB_SHA` is pinned at trigger time, so reusing the macro there would restore a wrong-but-plausible tree and reintroduce the mismatch. #848 carries that reasoning plus the candidate shape (record `git rev-parse HEAD` right after `setup:` and restore from that, which drops the per-event `$GITHUB_SHA` argument entirely), and flags `notifications` as the wrinkle — its checkout and `setup:` are both gated on the skip condition, so the record step needs that condition and the restore's `always()` has to tolerate the skipped case.

On the nested gap: the restore is recursive over the named directory, so a composite nesting an action *inside its own tree* is covered — it's only a sibling `uses: ./elsewhere` that escapes, since `setup:` is all the generator can see. I've noted that boundary in the helper's docstring so the next reader doesn't have to re-derive which half is handled.

generator/src/tend/templates/macros.yaml.j2
```diff
@@ -161,6 +161,35 @@
...
+   from, whatever branch the event put there. Every main step runs ahead of
+   every POST step, so any position below the tree switch is early enough.
+   `paths` is empty for a `setup:` that names no local action, rendering this
+   away. #}
+{% macro restore_local_actions(run_body) %}
+{% if run_body %}
+
+
+      - name: Restore local setup actions for POST cleanup
+        if: always()
```
tend-agent: The gate can skip every step above this one, and `always()` still runs it — against a workspace that was never checked out.

```suggestion
        if: always()<< " && (" ~ if_condition ~ ")" if if_condition else "" >>
```

generator/src/tend/templates/macros.yaml.j2
```diff
@@ -161,6 +161,35 @@
...
+   job red after a session that did its work correctly. actions/runner#2816.
+
+   `$GITHUB_SHA` names the tree the job's first checkout took: both callers give
+   that checkout no `ref`, and `actions/checkout` then resolves the triggering
+   event's own commit. So it is by construction the tree the composite loaded
+   from, whatever branch the event put there. Every main step runs ahead of
+   every POST step, so any position below the tree switch is early enough.
+   `paths` is empty for a `setup:` that names no local action, rendering this
+   away. #}
+{% macro restore_local_actions(run_body) %}
```
tend-agent: Optional, so mention's call site is unchanged.

```suggestion
{% macro restore_local_actions(run_body, if_condition='') %}
```

generator/src/tend/templates/review.yaml.j2
```diff
@@ -60,4 +60,4 @@ jobs:
           fi
 <<checkout(cfg, ref='${{ steps.pr_ref.outputs.ref }}', if_condition=skip_condition, allow_unsafe_pr_checkout=True, clean=not has_setup)>>
 
-<<agent_step(cfg, prompt_body, if_condition=skip_condition)>>
+<<agent_step(cfg, prompt_body, if_condition=skip_condition)>><<restore_local_actions(local_actions)>>
```
tend-agent:
```suggestion
<<agent_step(cfg, prompt_body, if_condition=skip_condition)>><<restore_local_actions(local_actions, if_condition=skip_condition)>>
```

generator/src/tend/templates/macros.yaml.j2
```diff
@@ -161,6 +161,35 @@
...
+   composite's step count fails cleanup with `Index was out of range`, and one
+   that drops the file entirely with `Can't find 'action.yml'`. Both paint the
+   job red after a session that did its work correctly. actions/runner#2816.
+
+   `$GITHUB_SHA` names the tree the job's first checkout took: both callers give
+   that checkout no `ref`, and `actions/checkout` then resolves the triggering
+   event's own commit. So it is by construction the tree the composite loaded
+   from, whatever branch the event put there. Every main step runs ahead of
+   every POST step, so any position below the tree switch is early enough.
+   `paths` is empty for a `setup:` that names no local action, rendering this
```
tend-agent:
```suggestion
   `run_body` is empty for a `setup:` that names no local action, rendering this
```

generator/tests/test_generate.py
```diff
@@ -88,6 +88,92 @@ def test_setup_steps_rendered(tmp_path: Path) -> None:
...
+    extra = "setup:\n  - uses: ./.github/actions/tend-setup\n"
+    cfg = Config.load(_minimal_config(tmp_path, extra))
+    steps = yaml.safe_load(GENERATORS[name](cfg).content)["jobs"][job]["steps"]
+
+    idx = _restore_step_index(steps)
+    assert idx is not None, f"{name}: nothing restores the local setup action"
+    assert ".github/actions/tend-setup" in str(steps[idx]["run"])
+    assert steps[idx].get("if") == "always()", (
+        f"{name}: the restore has to run even when the session fails"
+    )
```
tend-agent: This is the assertion that pins the un-composed shape, so it has to give a little for review to carry the gate condition. Worth a separate review-only check that the condition names `steps.gate.outputs.should_run`.

```suggestion
    assert str(steps[idx].get("if", "")).startswith("always()"), (
        f"{name}: the restore has to run even when the session fails"
    )
```

tend-agent (commented):
The merge with main lands the new review gate above this step, and the two don't compose: `if: always()` fires on runs where the composite was never loaded.

`always()` runs a step regardless of whether earlier steps were *skipped*, not just failed. So on a gate-skipped review run — a `synchronize` whose HEAD is already stamped, which is exactly the case main's gate exists to make common — no `actions/checkout` has run and the workspace is empty. `git checkout` then fails into the `||`:

```
fatal: not a git repository (or any of the parent directories): .git
::warning::could not restore .github/actions/tend-setup from deadbeef; POST cleanup of the local action may fail
```

(ran the rendered step body under `bash -e` in a non-repo directory; exit 0, warning emitted.) Nothing goes red — the `||` holds. But every gate-skipped review on an adopter with a local `setup:` composite now carries a warning annotation announcing that cleanup may fail, when in fact no composite loaded and there is no POST chain to clean up. That's the same run-summary surface this PR exists to keep quiet.

Threading the gate condition through fixes it: `always() && (steps.gate.outputs.should_run == 'true')`. Suggestions below add an optional `if_condition` to the macro and pass `skip_condition` from review; mention keeps the bare `always()`, since its handle job gates at the job level so every step either runs or the job doesn't. Rendered both shapes against the real Jinja env (`trim_blocks`/`lstrip_blocks` on) to check the inline expression doesn't disturb the surrounding lines.

There's a self-hosted variant of the same gap worth a line in the comment if you touch it: where the workspace persists between jobs, the skipped-run restore doesn't fail — it succeeds against whatever repo the previous job left behind, staging the base tree's copy of the composite into a stale workspace. Harmless, but it's the same missing condition.

Two smaller things. The macro comment still names the parameter `paths`, which it hasn't been since d85aa96. And `test_local_setup_action_restored_for_post_cleanup` asserts `if == "always()"` exactly — that assertion is what pins the current shape, so relaxing it is needed either way; worth also adding a review-specific assertion that the restore's condition names `steps.gate.outputs.should_run`, so the composition is pinned rather than merely permitted.

tend-agent:
Taken as suggested, in 7cac13b.

`restore_local_actions` grows an optional `if_condition`; review passes `skip_condition`, mention passes nothing and keeps the bare `always()`. Rendered against a config with two local composites and one remote — review reads `if: always() && (steps.gate.outputs.should_run == 'true')`, mention is unchanged, and the remote-only case still renders no step at all. `actionlint` v1.7.12 (the pinned rev, which runs shellcheck over inline `run:` blocks) is clean over the full rendered set, so the inline expression parses and the multi-path body is unaffected.

The macro comment names `run_body` now, and it picks up your self-hosted note: with a persistent workspace the skipped restore doesn't warn, it succeeds against whatever the previous job left behind — so the condition is what makes the step mean the same thing on both runner kinds, rather than only quieting the annotation.

On the test: the `always()` assertion is now `startswith`, and the parametrization carries the expected gate per workflow — review asserts the condition names `steps.gate.outputs.should_run`, mention asserts the condition is exactly `always()`. So the composition is pinned on one side and the job-level gating is pinned on the other, instead of both being merely permitted. Suite: 387 passed; the only snapshot that moved is review's.

I haven't run this on a live gate-skipped review either — same gap as the original fix, and for the same reason. The reasoning is from the rendered condition plus `always()`'s documented skip semantics, not from an observed run.

Issue #839, opened 2026-08-05
mention: participation checks reduce inside --jq under --paginate, so past 100 comments the bot silently stops answering


tend-agent:
## Problem

`mention.yaml.j2`'s `should_run` participation checks count bot activity by reducing *inside* `--jq` under `--paginate`:

```bash
BOT_REVIEWS=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/reviews" \
  --jq '[.[] | select(.user.login == "<<cfg.bot_name>>")] | length')
if [ "$BOT_REVIEWS" -gt "0" ]; then
```

`gh api --paginate` applies `--jq` to each page separately and concatenates the outputs, so the `| length` returns one count **per page**, not one overall.

Three call sites, all in the same step: [`BOT_COMMENTS` (issue path)](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/mention.yaml.j2#L278), [`BOT_REVIEWS`](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/mention.yaml.j2#L297), [`BOT_COMMENTS` (PR path)](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/generator/src/tend/templates/mention.yaml.j2#L304). They render into every adopter's `tend-mention.yaml` (this repo's copy: lines 264, 283, 290), so it isn't ours alone. The fourth `--paginate` in the step — the review-comment body scan piped to `grep -qF` — emits a per-element stream and is unaffected.

## Impact

`--paginate` requests `per_page=100`, so a thread needs >100 issue comments or >100 review records to cross a page boundary. Past that, the variable holds e.g. `100\n7` and the `-gt` guard errors instead of matching:

```
$ BOT_REVIEWS=$(printf '100\n7\n'); if [ "$BOT_REVIEWS" -gt "0" ]; then echo would-run; else echo "fell-through rc=$?"; fi
bash: [: 100
7: integer expression expected
fell-through rc=2
```

The `if` swallows the non-zero return, so the step falls through to `should_run=false` and the handle job never starts. The bot stops answering participants on exactly the threads where it has engaged the most — silently, with nothing red. Note that the synthetic reply containers described in #828 inflate the review count on precisely those long threads, so the two interact.

## Evidence for the per-page behavior

```
[...]

PR #849, opened 2026-08-05
fix(mention): skip the synthetic reply container the bot's own inline reply creates

tend-agent:
## Problem

`tend-mention`'s verify gate skips the bot's own comments — `issue_comment` and `pull_request_review_comment` where the comment author is the bot — because the bot never needs to respond to itself. The `pull_request_review` submission kind is deliberately left out of that skip, since a review the bot leaves on its own PR is its reviewer role.

But replying to a review thread (`POST /pulls/{n}/comments/{id}/replies`) wraps the reply in a **synthetic zero-body `COMMENTED` review** — the same artifact #828 documents. So the bot's own inline reply arrives twice: once as the `pull_request_review_comment` the skip already drops, and once as a `pull_request_review` submission it doesn't. The container isn't a review; it's the same comment on a second event path.

Nothing downstream catches it. The empty-body gate one block up requires `REVIEW_STATE = approved`, and a reply container is `commented`. The inline `@`-mention scan finds nothing (the bot doesn't summon itself). The engagement heuristic then sees a bot-authored PR — or `BOT_REVIEWS > 0` from the container itself — and returns `should_run=true, reason=participation`, spinning up a full `handle` job that the prompt's self-loop guard exits with nothing posted.

## Evidence

Three occurrences on `PRQL/prql`, recorded across two windows in the [evidence gist](https://gist.github.com/192514ea2c36586f9b7f842a482d62ab), all on bot-authored PRs where the bot answered its own review threads inline:

| Dispatched handle | PR | Cost |
| --- | --- | --- |
| [30894437460](https://github.com/PRQL/prql/actions/runs/30894437460) | #6141 | agent ran 9 m 16 s, posted nothing |
| [30894732823](https://github.com/PRQL/prql/actions/runs/30894732823) | #6141 | agent ran 7 m 35 s, posted nothing |
| [30991732962](https://github.com/PRQL/prql/actions/runs/30991732962) | [#6144](https://github.com/PRQL/prql/pull/6144) | queued 20 m, then cancelled |

The third is this run's window. Its container ([review 4862824760](https://github.com/PRQL/prql/pull/6144#pullrequestreview-4862824760)) has `body: ""`, `state: COMMENTED`, and exactly one inline comment — a reply:

```
$ gh api repos/PRQL/prql/pulls/6144/reviews/4862824760/comments \
    --jq '.[] | {id, in_reply_to_id, user: .user.login}'
{"id":3719317911,"in_reply_to_id":3719296566,"user":"prql-bot"}
```

Its `handle` job sat pending behind a live sibling for 20 minutes and was then evicted by GHA's queue-depth-1 replacement when a real dispatch arrived. That is the harm beyond wasted agent minutes: a container carrying no work occupies the one queue slot `cancel-in-progress: false` leaves, and the next event has to displace it.

## Fix

Reuse the inline-comment fetch the gate already makes, and skip when the review is bot-authored, has an empty body, and **every** inline comment is a reply:

```
if [ "$REVIEW_AUTHOR" = "<<cfg.bot_name>>" ] \
   && [ -z "$COMMENT_BODY" ] \
   && [ "$(printf '%s\n' "$INLINE" | jq -s '[.[] | select(.in_reply_to_id == null)] | length')" = "0" ]; then
  exit 0
fi
```

Deliberately narrower than the own-comment skip. The reason the `pull_request_review` path was left unguarded still holds — a real review the bot submits with inline comments and no body carries actionable signal — so the guard requires that no inline comment be a *fresh* one. Any container with a single non-reply comment still fires, and the `@`-mention scan runs first, so a mention inside a reply still wins.

The fetch changes from a piped `--jq '.[].body'` to a variable holding `{body, in_reply_to_id}` objects, one per line, so `--paginate` concatenates pages rather than reducing within each — the hazard #839 describes. The object form is load-bearing for a second reason: `in_reply_to_id` is an *optional* property, absent rather than null on a fresh comment, and constructing the object normalizes absent to null so `select(.in_reply_to_id == null)` counts fresh comments at all. A bare `.in_reply_to_id` stream emits nothing for them, and the gate would skip every container — including the real bot reviews it is written to spare.

A structural test (`test_mention_skips_bot_reply_container`) pins the four shapes above plus the three properties the regtest snapshots don't hold: the skip landing before the engagement heuristic, the inline `@`-mention scan running first, and the `REVIEW_AUTHOR` conjunction. That last is the load-bearing half — `pull_request_review_comment` subscribes to `edited` only, so this container is the sole path by which a human's freshly created inline reply reaches the bot, and an authorship-agnostic gate would silence the bot on every human reply to its own review threads with no test failing.

Verified the discriminator against the shapes it has to separate:

| Inline comments | non-reply count | Result |
| --- | --- | --- |
| one reply | 0 | skip |
| one fresh comment | 1 | fires |
| one reply + one fresh | 1 | fires |
| none | 0 | skip |

## Gate assessment

- **Evidence level**: High — 3 occurrences across two PRs and two windows, deterministic in each.
- **Structural**: yes. The event shape always passes the existing gate; there is no decision point where the bot could have behaved differently.
- **Change type**: targeted fix — one condition added beside the adjacent empty-body-APPROVED guard, no new mechanism.
- **Both gates pass** (High needs 2–3; a targeted fix carries the normal bar).

Distinct from #835, which fixes the same underlying artifact in the `review` skill's three review-record guards. That PR does not touch `mention.yaml.j2`; this is the second surface the container reaches.

`generator/` tests pass (327), with the four `mention` regtest snapshots regenerated. `bash -n` clean on every rendered `run:` step.

### DIFF
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -256,6 +256,11 @@ jobs:
           # so a first-contact @-mention inside an inline comment is detected
           # on PRs where the bot has no prior engagement. One object per line,
           # so `--paginate` concatenates pages instead of reducing within one.
+          # Keep the `{body, in_reply_to_id}` construction: `in_reply_to_id` is
+          # an *optional* property, absent rather than null on a fresh comment,
+          # and building the object normalizes absent to null so the `== null`
+          # select below counts both shapes. A bare `.in_reply_to_id` stream
+          # would emit nothing for a fresh comment and skip every container.
           if [ "$KIND" = "pull_request_review" ]; then
             INLINE=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PAYLOAD_PR/reviews/$PAYLOAD_ID/comments" \
               --jq '.[] | {body, in_reply_to_id}')
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -207,8 +207,8 @@ jobs:
           # comments on purpose: the pull_request_review *submission* kind is
           # deliberately left out, since a review the bot leaves on its own PR
           # is its reviewer role (the prompt is told to action it), not a
-          # self-loop. The only self-review skip is the terminal empty-body
-          # APPROVED gate below.
+          # self-loop. The only self-review skips are the terminal empty-body
+          # APPROVED gate and the synthetic reply container, both below.
           if { [ "$KIND" = "issue_comment" ] || [ "$KIND" = "pull_request_review_comment" ]; } \
              && [ "$COMMENT_AUTHOR" = "<<cfg.bot_name>>" ]; then
             echo "should_run=false" >> "$GITHUB_OUTPUT"
@@ -254,12 +254,34 @@ jobs:
           # A review's record includes review.body (checked above) but NOT the
           # bodies of the inline comments attached to the review. Fetch them
           # so a first-contact @-mention inside an inline comment is detected
-          # on PRs where the bot has no prior engagement.
+          # on PRs where the bot has no prior engagement. One object per line,
+          # so `--paginate` concatenates pages instead of reducing within one.
           if [ "$KIND" = "pull_request_review" ]; then
-            if gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PAYLOAD_PR/reviews/$PAYLOAD_ID/comments" \
-                 --jq '.[].body' | grep -qF '@<<cfg.bot_name>>'; then
+            INLINE=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PAYLOAD_PR/reviews/$PAYLOAD_ID/comments" \
+              --jq '.[] | {body, in_reply_to_id}')
+
+            if printf '%s\n' "$INLINE" | jq -r '.body' | grep -qF '@<<cfg.bot_name>>'; then
               echo "should_run=true" >> "$GITHUB_OUTPUT"
               echo "reason=mention" >> "$GITHUB_OUTPUT"
+              exit 0
+            fi
+
+            # Replying to a review thread wraps the reply in a synthetic
+            # zero-body COMMENTED review, so the bot's own inline reply arrives
+            # here as a `pull_request_review` submission as well as the
+            # `pull_request_review_comment` the skip above already drops. Same
+            # artifact, second event path: without this the engagement heuristic
+            # below sees a bot-authored PR (or BOT_REVIEWS > 0) and spins up a
+            # full handle job that the prompt's self-loop guard then exits.
+            # Narrower than the comment skip on purpose — a real review the bot
+            # submits with inline comments and no body carries actionable signal,
+            # so require that *every* inline comment be a reply
+            # (`in_reply_to_id` set). A container with any fresh inline comment
+            # still fires.
+            if [ "$REVIEW_AUTHOR" = "<<cfg.bot_name>>" ] \
+               && [ -z "$COMMENT_BODY" ] \
+               && [ "$(printf '%s\n' "$INLINE" | jq -s '[.[] | select(.in_reply_to_id == null)] | length')" = "0" ]; then
+              echo "should_run=false" >> "$GITHUB_OUTPUT"
               exit 0
             fi
           fi
--- a/generator/tests/_regtest_outputs/test_generate.test_sandbox_levers_regtest.out
+++ b/generator/tests/_regtest_outputs/test_generate.test_sandbox_levers_regtest.out
@@ -193,8 +193,8 @@ jobs:
           # comments on purpose: the pull_request_review *submission* kind is
           # deliberately left out, since a review the bot leaves on its own PR
           # is its reviewer role (the prompt is told to action it), not a
-          # self-loop. The only self-review skip is the terminal empty-body
-          # APPROVED gate below.
+          # self-loop. The only self-review skips are the terminal empty-body
+          # APPROVED gate and the synthetic reply container, both below.
           if { [ "$KIND" = "issue_comment" ] || [ "$KIND" = "pull_request_review_comment" ]; } \
              && [ "$COMMENT_AUTHOR" = "test-bot" ]; then
             echo "should_run=false" >> "$GITHUB_OUTPUT"
@@ -240,12 +240,34 @@ jobs:
           # A review's record includes review.body (checked above) but NOT the
           # bodies of the inline comments attached to the review. Fetch them
           # so a first-contact @-mention inside an inline comment is detected
-          # on PRs where the bot has no prior engagement.
+          # on PRs where the bot has no prior engagement. One object per line,
+          # so `--paginate` concatenates pages instead of reducing within one.
           if [ "$KIND" = "pull_request_review" ]; then
-            if gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PAYLOAD_PR/reviews/$PAYLOAD_ID/comments" \
-                 --jq '.[].body' | grep -qF '@test-bot'; then
+            INLINE=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PAYLOAD_PR/reviews/$PAYLOAD_ID/comments" \
+              --jq '.[] | {body, in_reply_to_id}')
+
+            if printf '%s\n' "$INLINE" | jq -r '.body' | grep -qF '@test-bot'; then
               echo "should_run=true" >> "$GITHUB_OUTPUT"
               echo "reason=mention" >> "$GITHUB_OUTPUT"
+              exit 0
+            fi
+
+            # Replying to a review thread wraps the reply in a synthetic
+            # zero-body COMMENTED review, so the bot's own inline reply arrives
+            # here as a `pull_request_review` submission as well as the
+            # `pull_request_review_comment` the skip above already drops. Same
+            # artifact, second event path: without this the engagement heuristic
+            # below sees a bot-authored PR (or BOT_REVIEWS > 0) and spins up a
+            # full handle job that the prompt's self-loop guard then exits.
+            # Narrower than the comment skip on purpose ? a real review the bot
+            # submits with inline comments and no body carries actionable signal,
+            # so require that *every* inline comment be a reply
+            # (`in_reply_to_id` set). A container with any fresh inline comment
+            # still fires.
+            if [ "$REVIEW_AUTHOR" = "test-bot" ] \
+               && [ -z "$COMMENT_BODY" ] \
+               && [ "$(printf '%s\n' "$INLINE" | jq -s '[.[] | select(.in_reply_to_id == null)] | length')" = "0" ]; then
+              echo "should_run=false" >> "$GITHUB_OUTPUT"
               exit 0
             fi
           fi
--- a/generator/tests/_regtest_outputs/test_generate.test_sandbox_levers_regtest.out
+++ b/generator/tests/_regtest_outputs/test_generate.test_sandbox_levers_regtest.out
@@ -242,6 +242,11 @@ jobs:
           # so a first-contact @-mention inside an inline comment is detected
           # on PRs where the bot has no prior engagement. One object per line,
           # so `--paginate` concatenates pages instead of reducing within one.
+          # Keep the `{body, in_reply_to_id}` construction: `in_reply_to_id` is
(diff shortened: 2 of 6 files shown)


tend-agent (commented):
The discriminator holds up — I checked the shape it keys on against live data rather than the docs, since `in_reply_to_id` is an *optional* property on that endpoint's schema. On a review carrying fresh inline comments, `has("in_reply_to_id")` is `false`: the key is absent, not null. `{body, in_reply_to_id}` construction is what makes `select(.in_reply_to_id == null)` catch both shapes, so the object form isn't only about `--paginate`. Worth a word in the comment, since a later "simplify" pass that drops back to a bare `.in_reply_to_id` stream would silently start counting fresh comments as replies.

The gap is tests. Every other author-keyed gate in this step is pinned by a structural test in `test_generate.py` — `test_mention_skips_bot_approved_review` asserts its fragments *and* that the skip precedes `BOT_REVIEWS=$(`; `test_mention_self_comment_skip_spares_review_submissions` pins the boundary the gate must not cross. This one adds no test. The regtest snapshots regenerate on any edit, so they record the text without pinning the intent, and three properties this PR argues for are currently unguarded: the skip landing before the engagement heuristic that would otherwise count the container itself, the inline `@`-mention scan running first, and the `REVIEW_AUTHOR` conjunction.

That last one is the load-bearing half and the easiest to lose. `pull_request_review_comment` subscribes to `edited` only, so the synthetic container is the *sole* path by which a human's freshly-created inline reply reaches the bot at all. Widen this gate to authorship-agnostic — the obvious "simplify" — and the bot goes silent on every human reply to its own review threads, with no test failing and no red run to notice.

Second, `test_mention_self_comment_skip_spares_review_submissions` now documents the wrong boundary. Its docstring ("The only self-review that is skipped is the terminal empty-body APPROVED case") and the comment above `review_skip` ("The sole author-keyed skip for a review submission is the terminal empty-body APPROVED gate") both assert exactly the claim this PR corrects in the template comment one file over. It still passes, but its `review_skip` slice starts at the first `[ "$KIND" = "pull_request_review" ]` and runs to the end of the script, so `assert '[ -z "$COMMENT_BODY" ]' in review_skip` is now satisfied by either gate and no longer discriminates.

Both fixes are outside the diff and there's no human author to apply a suggestion, so I've pushed them to this branch: a `test_mention_skips_bot_reply_container` covering the four shapes in your table plus the ordering and authorship properties, and the two stale comments updated to name both skips.

API shape, on this repo — review `4860021420` (body 2286, fresh inline comments) vs `4860032792` (body 0, reply container):

```
$ gh api repos/max-sixty/tend/pulls/835/reviews/4860021420/comments --jq '.[0] | has("in_reply_to_id")'
false
$ gh api repos/max-sixty/tend/pulls/835/reviews/4860021420/comments --jq '.[] | {body, in_reply_to_id}'
{"body":"An empty-body `--approve` has no body ...","in_reply_to_id":null}
$ gh api repos/max-sixty/tend/pulls/835/reviews/4860032792/comments --jq '.[] | {body, in_reply_to_id}'
{"body":"Applied in 8699228. ...","in_reply_to_id":3717103568}
```

`gh --jq` emits one compact object per line for object results, and JSON escapes newlines inside `body`, so a multi-line comment stays on one line.

Ran the rendered gate against every shape under `bash -e`: one reply → skip; one fresh → fires; reply + fresh → fires; no comments → skip; human-authored reply container → fires; bot container with a non-empty body → fires; `@bot` inside a reply → fires as `reason=mention` (the scan wins, as the PR body claims), including when the mention is on a later line of a multi-line body.

`bash -n` clean on every rendered `run:` step of `tend-mention` built from this repo's own `.config/tend.yaml`; 327 generator tests pass at `e411d0f`.

#840 also touches `mention.yaml.j2` but in the engagement-heuristic block ~20 lines below, so the two don't collide.

</details>

tend-agent:
Both of the review's pushed fixes are on the branch (e1cf1c9) — the structural test and the two stale comments. The one item left was the note asking for a word in the template comment about the object construction, since it only gave the `--paginate` reason; that's 5555078.

I re-checked the shape rather than take it from the review: on a review carrying fresh inline comments, the key is absent, not null.

```
$ gh api repos/max-sixty/tend/pulls/835/reviews/4860021420/comments --jq '.[0] | has("in_reply_to_id")'
false
$ gh api repos/max-sixty/tend/pulls/835/reviews/4860021420/comments --jq '.[] | {in_reply_to_id}'
{"in_reply_to_id":null}
```

So the failure mode of dropping back to a bare `.in_reply_to_id` stream is worse than miscounting: jq emits nothing at all for a fresh comment, the non-reply count is `0` for every container, and the gate skips the real empty-body-with-inline-comments reviews it was written to spare. The comment now says that, next to the fetch.

I also folded both properties into the PR description, which still claimed `--paginate` as the only reason for the object form.

328 generator tests pass at 5555078 (four `mention` regtest snapshots regenerated for the comment text — no rendered-script change). `bash -n` clean on every rendered `run:` step of `tend-mention` built from this repo's own `.config/tend.yaml`.

</details>

tend-agent (commented):
No further findings — the two items the last review raised are on the branch, and I re-derived the claims rather than taking them from the commit messages. Recording the checks because the crux one isn't visible from the diff: whether the new test actually catches the loss of the conjunct it was written to protect.

`REVIEW_AUTHOR` is genuinely pinned. Deleting that conjunct from the reply gate — the "obvious simplify" the docstring warns about — fails `test_mention_skips_bot_reply_container` rather than sliding through, and the assertion discriminates because both `[ "$REVIEW_AUTHOR" = "test-bot" ]` and `[ -z "$COMMENT_BODY" ]` occur exactly once after `INLINE=$(`. The narrowed `approved_skip` slice is likewise the only place `[ "$REVIEW_STATE" = "approved" ]` appears, so the sibling test discriminates again too.

The absent-vs-null claim the template comment now records holds, and its failure mode is the one stated: `has("in_reply_to_id")` is `false` on a fresh comment, and a bare `.in_reply_to_id` stream emits a blank line per fresh comment (not `null`), which `jq -s` slurps to `[]` — count `0`, so the gate would skip every container, including the real empty-body reviews it exists to spare.

No approval: self-authored.

Mutation, on `mention.yaml.j2` with the `REVIEW_AUTHOR` conjunct removed from the reply gate:

```
FAILED tests/test_generate.py::test_mention_skips_bot_reply_container
  - assert '[ "$REVIEW_AUTHOR" = "test-bot" ]' in 'INLINE=$(gh api --paginate ...'
```

Fragment counts in the rendered `tend-mention` check step, whole script vs. after `INLINE=$(`: `[ "$REVIEW_AUTHOR" = "test-bot" ]` 2/1, `[ -z "$COMMENT_BODY" ]` 2/1, `[ "$REVIEW_STATE" = "approved" ]` 1/0.

API shape, on a review carrying fresh inline comments:

```
$ gh api repos/max-sixty/tend/pulls/835/reviews/4860021420/comments --jq '.[] | has("in_reply_to_id")'
false
false
$ gh api repos/max-sixty/tend/pulls/835/reviews/4860021420/comments --jq '.[].in_reply_to_id' | cat -A
$
$
```

Re-ran the rendered gate under `bash -e` across the eight shapes (one reply, one fresh, reply + fresh, none, human-authored container, bot container with a body, `@bot` quoted inside a reply, multi-line reply body) — each lands where the PR body's table says.

#840 overlap: merged it into this branch locally — clean, 329 tests pass. Its new pagination test bans `| length` only on lines carrying both `--paginate` and `--jq`, and this PR's `jq -s '... | length'` runs on the already-captured `$INLINE`, so the two don't cross. 328 tests pass at 5555078; the four regtest snapshots are pure ASCII by pre-existing convention, so the `—` in the new comment lands as `?` there as it does on the 25 that were already in the file.

</details>

tend-agent:
Fourth occurrence, and the first on a second repo — `max-sixty/cargo-affected` PR [#77](https://github.com/max-sixty/cargo-affected/pull/77), observed in [run 31081926964](https://github.com/max-sixty/tend/actions/runs/31081926964).

This one raises the harm above wasted minutes and a queue slot: an unguarded container caused the bot to **post additional public output**. Every case in the table above was a silent no-op or a cancelled job; here the chain ran three hops and the middle one commented.

| Time | Review | Fired | Result |
| --- | --- | --- | --- |
| 07:27:36Z | `4872123225` — real self-review, body 1238 chars, 1 fresh inline comment | mention [31081009768](https://github.com/max-sixty/cargo-affected/actions/runs/31081009768) | correct: pushed fix `56406d76`, replied |
| 07:31:34Z | `4872154023` — **container**, empty body, 1 reply | mention [31081269852](https://github.com/max-sixty/cargo-affected/actions/runs/31081269852) | **posted inline comment `3726818957`** |
| 07:36:32Z | `4872195438` — **container**, empty body, 1 reply | mention [31081591885](https://github.com/max-sixty/cargo-affected/actions/runs/31081591885) | agent ran ~2 min, posted nothing |

Each reply generated the next container, so the chain was self-sustaining and stopped only when hop 3 happened to decide it had nothing to add. The whole exchange has one participant:

```
$ gh api "repos/max-sixty/cargo-affected/pulls/77/comments?per_page=100" --jq '[.[].user.login] | unique'
["cargo-affected-bot"]
```

Your predicate separates the three cleanly — verified against the live reviews:

```
$ for r in 4872123225 4872154023 4872195438; do
    gh api "repos/max-sixty/cargo-affected/pulls/77/reviews/$r/comments" \
      --jq '.[] | {body, in_reply_to_id}' | jq -s '[.[] | select(.in_reply_to_id == null)] | length'
  done
1    # 4872123225 — fires, as intended
0    # 4872154023 — skipped
0    # 4872195438 — skipped
```

So the guard would have suppressed hops 2 and 3 while leaving the genuine self-review to be actioned, which is exactly the split the PR argues for. The `REVIEW_AUTHOR` conjunction is load-bearing here too — hop 1's fresh comment is bot-authored, and it still needs to fire.

No change requested; recording the occurrence since it moves the cost from "agent minutes" to "public self-conversation on the PR".

tend-agent:
Correcting my earlier comment on this PR — the attribution table above is wrong, and it overstates the case for the guard.

I mapped each output to the run whose start/end span contained its timestamp. Four bot sessions overlapped on that PR, so the span proves nothing. Read from each session's own log (the tool call that actually posted):

| Output | Table above says | Session log says |
| --- | --- | --- |
| commit `56406d76` + inline reply `3726782508` | mention [31081009768](https://github.com/max-sixty/cargo-affected/actions/runs/31081009768) | `tend-nightly` [31080083613](https://github.com/max-sixty/cargo-affected/actions/runs/31080083613) — the run that opened the PR, still running |
| inline reply `3726818957` | mention [31081269852](https://github.com/max-sixty/cargo-affected/actions/runs/31081269852) | mention [31081009768](https://github.com/max-sixty/cargo-affected/actions/runs/31081009768) |

So both container-triggered hops — `31081269852` and `31081591885`, the two this PR would suppress — posted nothing. Neither log contains a posting call, and both end with an explicit silent-exit decision ("Exiting silently — nothing posted"; "Responding would be a self-conversation loop"). The two public replies came from the hop your predicate keeps firing (the genuine self-review, correctly actioned by the PR author) and from a racing sibling session outside the chain entirely.

The claim to strike is "an unguarded container caused the bot to post additional public output" — that didn't happen here. What did happen is the cost this PR already argues: two container hops each spun up an agent, held a queue slot, and exited with nothing to say. The predicate check in that comment still holds (`4872123225` → 1, the two containers → 0), so the split the PR argues for is unaffected; only the severity framing was inflated.

<details><summary>How the attributions were confirmed</summary>

```bash
gh run download <run-id> -R max-sixty/cargo-affected --pattern '*session-logs*' --dir /tmp/sl/<run-id>/
jq -r 'select(.type == "assistant") | .message.content[]? | select(.type == "tool_use") | "\(.name): \(.input | tostring | .[0:200])"' FILE
```

`31080083613` (nightly): writes `/tmp/reply-inline.md` opening "Fair — taken, with a small rewording…", then `POST .../pulls/77/comments/3726755466/replies` behind an `EXISTING == 0` dedup check → `discussion_r3726782508`. Same run committed and pushed `tests: re-check the database after the second collect too`.

`31081009768` (mention hop 1): made the same edits independently, committed, then `git fetch` → found the sibling's commit on origin → `git reset --hard origin/tests/replace-dead-shim-tripwires` to drop its duplicate → read the sibling reply in full → ran an invocation-scoped injected break → posted `3726818957`.

`31081269852` and `31081591885`: six and four read-only `gh api` calls respectively, no `Write`, no POST.

</details>

The mapping-by-timestamp habit that produced this is fixed in #869.

PR #850, opened 2026-08-05
fix(list-recent-runs): close the completion window at the cron tick

tend-agent:
`list-recent-runs.sh` anchors its completion window's **floor** to the intended cron tick, and advances that floor by exactly one period per cycle. It never sets a ceiling — [the filter is `updatedAt >= cutoff` and nothing more](https://github.com/max-sixty/tend/blob/4672f809cedd2f206e109761072641798aa0184e/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L154-L160). So each cycle lists `floor .. now`, which is wider than one period by however late the GHA scheduler was, and the next cycle — whose floor is exactly this cycle's tick — lists that tail a second time. The script's own header says the opposite: *"Consecutive cycles then tile exactly: [intended-1h, intended], then [intended, intended+1h]."* That property only holds with the ceiling this PR adds.

The cost is a duplicated agent survey, on the workflow that is the fleet's dominant token consumer.

## Measured on this repo's last two cycles

`review-reviewers` ticks are landing 45–55 min late (`09:41:26Z` → `10:41:41Z` → `11:32:14Z`, against `47 * * * *`), so the redundant share is close to the whole period:

| Cycle | Intended tick | Floor | Listed through | Runs listed |
|---|---|---|---|---|
| [30998444312](https://github.com/max-sixty/tend/actions/runs/30998444312) | 09:47Z | 08:47Z | 10:42Z (its own start) | 21 |
| [31001852480](https://github.com/max-sixty/tend/actions/runs/31001852480) (this run) | 10:47Z | 09:47Z | 11:32Z | 21 |

**19 of this cycle's 21 runs had already finished before the previous cycle began listing**, so they were in that cycle's list too and were surveyed twice. Only two runs — `30998444312` itself and `30999831196` — were new. Both cycles independently reached the same all-clear on the same runs.

With the ceiling, this cycle's window closes at 10:47Z and the two post-tick runs move to the next cycle, whose floor is 10:47Z. Nothing is dropped, only deferred by one cycle: the deferral is bounded by the period, and the dropped-tick recovery below still reaches further back when a tick never fires.

## Verification

New `generator/tests/test_list_recent_runs.py` drives the real script through a stub `gh` (no network), placing fixture runs either side of the tick:

- `test_window_is_half_open_at_the_cron_tick` — pre-floor, in-window, post-tick and still-running runs; asserts only the in-window pair is listed. **Mutation-checked: fails on `main` with `[2, 3, 4]` vs the expected `[2, 3]`.**
- `test_next_cycle_picks_up_what_the_ceiling_deferred` — a run finishing just after the *previous* tick is listed by this cycle, i.e. deferred rather than dropped.
- `test_non_periodic_cron_keeps_the_now_anchored_window_uncapped` — off the cron path (`*/15 * * * *`, `workflow_dispatch`) there is no next cycle to hand a tail to, so no ceiling is applied.

Each test derives the cron minute as `(now.minute + 30) % 60`, putting the tick ~30 min from the clock in either direction so it can't race a rollover. `shellcheck -S warning` clean, `bash -n` clean, full generator suite 330 passed.

## Relationship to the other open PRs on this file

Neither of the two open PRs touching `list-recent-runs.sh` closes this gap; both keep the floor-only filter and both restate the "tile exactly / no overlap" claim in comments.

- **#845** (3-hourly cadence, period-anchored floor) generalises the floor to `intended - period`. It cuts the *number* of cycles, but the overlap per cycle stays equal to the scheduler delay — at a 3-hourly period that is ~30% of each window instead of ~90%. Its verification prints windows with both bounds, which is the behaviour this PR actually implements.
- **#838** (anchor recovery on the last *successful* run) widens the floor after an outage. It reasons about overlap as "re-offered work the caller dedups against its own evidence log", which is the right call for the recovery path — that overlap is a deliberate outage backstop. This PR leaves it untouched and only closes the routine, every-cycle overlap.

The edits are textually adjacent to both (a new assignment after `COMPLETED_AFTER`), so whichever lands second wants a small rebase; neither depends on the other.

## Gate assessment

- **Evidence level**: High. Mechanically determined, not sampled — with a non-zero scheduler delay the overlap occurs every cycle by construction. Measured at 19/21 runs this cycle; the evidence log records prior cycles noting runs "already surveyed by the prior run; listed by the run script because of window overlap", and the ~45–55 min lag has been confirmed in eight consecutive windows.
- **Structural**: yes. No decision point — the same conditions produce the same duplication every cycle.
- **Change type**: targeted fix (one assignment plus the filter clause it feeds), with a regression test.
- **Passes both gates.**

### DIFF
--- a/generator/tests/test_list_recent_runs.py
+++ b/generator/tests/test_list_recent_runs.py
@@ -1,0 +1,161 @@
+"""Tests for `plugins/tend-ci-runner/scripts/list-recent-runs.sh`.
+
+The script decides which CI runs a `review-reviewers`/`review-runs` session
+analyzes, so a wrong window is invisible in the output — it reads as a normal
+run list. It isn't part of the generator package; the test lives here because
+this is the repo's only Python suite, and shellcheck can't catch window
+arithmetic.
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+from datetime import UTC, datetime, timedelta
+from pathlib import Path
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+LIST_RECENT_RUNS = (
+    REPO_ROOT / "plugins" / "tend-ci-runner" / "scripts" / "list-recent-runs.sh"
+)
+
+# `gh` stand-in covering the three calls the script makes: the workflow
+# discovery list, the previous-run window anchor (the only `run list` carrying
+# `--status`), and the per-workflow run list.
+FAKE_GH = """#!/usr/bin/env bash
+case "$1 $2" in
+  "workflow list") echo '[{"name":"tend-review"}]' ;;
+  "run list")
+    if [[ "$*" == *"--status"* ]]; then
+      echo "$FAKE_PREV_RUN_STARTED_AT"
+    else
+      cat "$FAKE_RUNS_JSON"
+    fi
+    ;;
+  *) echo "unexpected gh invocation: $*" >&2; exit 1 ;;
+esac
+"""
+
+
+def _iso(moment: datetime) -> str:
+    return moment.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def _intended_tick(now: datetime, cron_minute: int) -> datetime:
+    """The tick the script anchors on, by the same rule the script uses."""
+    this_hour_tick = now.replace(minute=cron_minute, second=0, microsecond=0)
+    if now < this_hour_tick:
+        return this_hour_tick - timedelta(hours=1)
+    return this_hour_tick
+
+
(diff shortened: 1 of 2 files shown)


max-sixty:
Closing under the cost gate that just landed in #960: the every-cycle overlap is waste-class — the survey's consumer already dedups against its evidence log, so the cost is a duplicated survey, not a wrong output — and the ceiling is scheduling arithmetic, which Gate 3 now bars at any occurrence count. The evidence stays recorded; if duplicated surveys grow to matter, the lever is cadence.

Issue #855, opened 2026-08-06
rate-limit preflight: the creation-spike gate halts review and notification runs, so two PRs merged unreviewed


tend-agent:
The spike gate in [`shared/steps/rate-limit-preflight.sh`](https://github.com/max-sixty/tend/blob/f3e309a475af03147ce6290104c85baafe6ad9f5/shared/steps/rate-limit-preflight.sh#L44-L50) tripped on 2026-08-05 and aborted **every** agent run on this repo for 12 h 13 min, until the UTC-day counter reset. During that window `tend-review` never ran, and [#852](https://github.com/max-sixty/tend/pull/852) and [#853](https://github.com/max-sixty/tend/pull/853) — both `max-sixty`-authored — were merged with **zero reviews from anyone**.

The gate measures one thing (issues and PRs the bot *created*) and enforces on another (whether the agent runs at all). `tend-review`'s output is a review, which `search/issues` never counts, so no amount of reviewing can trip the gate and blocking review cannot relieve it. The same holds for `tend-notifications`, which marks notifications read. Those runs were pure collateral.

## What happened

| | |
|---|---|
| Blackout start | `2026-08-05T11:47:16Z` — [run 31002862308](https://github.com/max-sixty/tend/actions/runs/31002862308) (`tend-review`) |
| Blackout end | `2026-08-05T23:45:50Z` — [run 31057476863](https://github.com/max-sixty/tend/actions/runs/31057476863) (`tend-notifications`) |
| Recovery | UTC-day rollover; the 00:07:44Z `review-reviewers` run passed |
| Failed runs | **36 of 37** on 2026-08-05 (8/8 sampled confirmed as this gate; the 1 non-gate failure was an unrelated `ci` proxy test) |

Every failed run carried the identical annotation:

```
##[error]Rate limit: bot created 16 items today, above spike limit of 15 (baseline: 17 over past 6 days)
```

The arithmetic is exactly as [written](https://github.com/max-sixty/tend/blob/f3e309a475af03147ce6290104c85baafe6ad9f5/shared/steps/rate-limit-preflight.sh#L26-L31): `SPIKE_LIMIT = 10 + PAST_POSTS / 3` = `10 + 17/3` = `15`, and the bot's 16th item of the day ([#850](https://github.com/max-sixty/tend/pull/850), 11:47:12Z) crossed it four seconds before the first blocked run.

## Cost

Four `tend-review` runs were blocked, each within seconds of the PR it was meant to review:

[...]

Issue #848, opened 2026-08-05
setup composite POST cleanup: the restore in #847 misses runtime tree switches and sibling nested actions


tend-agent:
## Problem

#847 restores the local `setup:` composite before POST cleanup on `tend-review` and `tend-mention` — the two workflows whose *templates* swap the workspace tree after loading it. That scoping is a YAML-level claim, and the runtime one is weaker: the agent can switch the tree from inside the agent step on any workflow.

`nightly`'s bot-PR-conflict routine opens with `gh pr checkout <number>` ([nightly/SKILL.md L124](https://github.com/max-sixty/tend/blob/d85aa9656f8aad8ace9fbce090f05e5b4cfe7e52/plugins/tend-ci-runner/skills/nightly/SKILL.md#L124)), and `running-in-ci`'s "Pushing to PR Branches" hands the same recipe to every workflow. So an adopter whose `setup:` names a local composite gets the identical `Index was out of range` on `tend-nightly` the moment it checks out a bot PR that resized that composite — which is exactly the shape of PR the nightly conflict pass exists to handle. Same for triage, ci-fix, and notifications whenever the agent checks out a branch.

## Why it isn't a one-line generalization

Dropping the `restore_local_actions` macro into the other four templates would be wrong. Those workflows check out `ref: <default_branch>`, which `actions/checkout` resolves to `origin/<default>` **at fetch time**, whereas `$GITHUB_SHA` on a `schedule` event is pinned at trigger time. A push landing in that window makes the two differ, and restoring a wrong-but-plausible version of the composite reintroduces the step-count mismatch the step exists to prevent.

The restore source has to be parameterized to whatever tree the calling checkout actually landed on. The candidate shape: record it right after `setup:` runs, and restore from that, which is uniformly correct and drops the per-event `$GITHUB_SHA` reasoning entirely:

```yaml
      - name: Record the tree the local setup actions loaded from
        id: tend_setup_tree
        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
```

The wrinkle to settle is `notifications`, whose checkout and `setup:` are both gated on the pre-check's skip condition — the record step needs that same condition, and the restore's `if: always()` then has to tolerate the skipped case rather than failing on a missing output.

## Second gap: nested local actions

[...]

PR #856, opened 2026-08-06
fix(rate-limit): pause creation on a daily spike instead of failing the run

tend-agent:
On 2026-08-05 the bot went dark on its own repo for twelve hours. Every `tend-*` and `review-reviewers` run from 11:47Z to the end of the UTC day died in the same shell step, before the agent started, with the same annotation:

```
##[error]Rate limit: bot created 16 items today, above spike limit of 15 (baseline: 17 over past 6 days)
```

35 runs failed that way — `tend-notifications` ×19, `review-reviewers` ×12 (each fanning out to 5 matrix legs, so 60 failed jobs), `tend-review` ×3, `tend-mention` ×1. The dropped work was real, not just monitoring: three pull requests never got reviewed and one human comment never got answered. The first casualty was [run 31002862308](https://github.com/max-sixty/tend/actions/runs/31002862308) at 11:47:16Z — the `tend-review` fired by the creation of #850, which was itself item 16.

## Why it fired

`shared/steps/rate-limit-preflight.sh` compares today's issue+PR count against a 6-day baseline: `spike_limit = 10 + past_posts / 3`. The bot's per-day creation counts leading in:

| Date | Items |
|---|---|
| 07-30 | 0 |
| 07-31 | 1 |
| 08-01 | 4 |
| 08-02 | 1 |
| 08-03 | 5 |
| 08-04 | 6 |
| 08-05 | 16 |

Baseline 17 → limit 15. The 16th item tripped it, and because the count only resets at UTC midnight, it stayed tripped for the rest of the day. Nothing was wrong with those 16 items; they are the ordinary output of the review/review-runs/nightly workflows, and most are still open for review. The low baseline is itself an artifact — 07-30 through 08-02 fall inside the model weekly-quota outage that reset at 2026-08-05T00:00Z, so the guard was calibrated on outage days and then fired on the recovery day. A quiet week makes the next busy day fail.

## What's wrong with the response, not the threshold

The guard exists to stop runaway issue/PR creation, and hard-aborting is the right response to a loop. But it aborts *every* invocation, including work that creates nothing and so cannot contribute to the count it is enforcing — reviews, mention replies, triage comments, CI fixes, and the read-only analysis workflows. Raising the constant would only move the cliff.

So this splits the response by which failure shape is actually present:

| Tier | Condition | Response |
|---|---|---|
| Burst | >10 PRs or >10 issues in 20 minutes | abort (unchanged) |
| Hard | today > `max(10 + past_posts, spike_limit + 10)` — a whole baseline week's output in one day, floored 10 above the spike tier | abort (new) |
| Spike | today > `10 + past_posts / 3` | run continues, creation paused |

On a spike-tier trip the script now emits a warning and writes `TEND_CREATION_PAUSED_NOTE` to `GITHUB_ENV`; `compose-system-prompt.sh` (Claude) and the `Stage AGENTS.md` step (Codex) append it to the agent's prompt, telling it not to open a new issue or PR this run and to describe what it would have filed in a thread comment instead. Everything else proceeds. Replaying 2026-08-05 against the new tiers: 16 vs spike 15, hard 27 → creation paused, and the three reviews and one mention reply happen.

The honest tradeoff: the spike tier becomes a prompt directive, which a prompt-injected session could disregard where an `exit 1` could not. That is why it sits *below* a hard ceiling rather than replacing one — the two abort tiers are still pure shell, still unskippable, and still catch both the fast loop (burst) and the slow one (hard). `docs/security-model.md` and `README.md` are updated to describe the tiers as they now are.

## Verification

`generator/tests/test_shared_steps.py` covers the tiers against a stubbed `gh` (7 tests):

| prs / issues / today / baseline | Result |
|---|---|
| 0 / 0 / 5 / 17 | passes |
| 0 / 0 / 16 / 17 | exit 0, warning, `TEND_CREATION_PAUSED_NOTE` written |
| 0 / 0 / 28 / 17 | exit 1, hard-limit error |
| 11 / 0 / 5 / 17 | exit 1, burst error |

Both limits scale off the same baseline, so without the floor they converge as it falls and coincide at 10 when it reaches 0 — the pause tier would be unreachable and the run would hard-abort on the 11th item, exactly the pre-PR behaviour. A depressed baseline is the diagnosed trigger, so the floor keeps a pause band at least 10 wide at any baseline. It binds below a baseline of 15 and leaves the 2026-08-05 case (spike 15, hard 27) unchanged.

On a spike-tier trip a run with no triggering thread — the scheduled and `workflow_run` workflows, whose whole deliverable is a new issue or PR — is told to write what it would have filed to `/tmp/claude/step-summary.md`, which a later step copies into the job summary, so the work isn't discarded with the runner. `claude/action.yaml` already had that step and two bundled skills already used it; `codex/action.yaml` did not, so it's added here.

`compose-system-prompt.sh` was run with the note set and appends it after the adopter's `system_prompt_append`. The Codex step body parses and passes `bash -n`; shellcheck is clean on both shared scripts.

## Gate assessment

- **Evidence**: High/Critical, structural. 35 failed runs in a single day from one deterministic condition, with human-facing work dropped (3 unreviewed PRs, 1 unanswered comment). Same conditions reproduce it exactly. The evidence log has carried "review-reviewers self-run failure causing a window gap" at cumulative 19 across prior months; this run identifies the cause behind a 12-hour block of them.
- **Magnitude**: structural change to a guard shipped to every adopter, which wants a very high bar — met by the count above plus the dropped work. The abort behavior is preserved for both runaway shapes rather than removed.
- **Evidence gist**: https://gist.github.com/192514ea2c36586f9b7f842a482d62ab


tend-agent:
Added regression tests for the three tiers in [66f8e84](https://github.com/max-sixty/tend/commit/66f8e84) — this run's `numbagg/numbagg` matrix leg reached the same root cause independently and had a fix staged, so rather than open a competing PR it's folded in here. The behaviour change is yours unchanged; this only adds coverage.

`generator/tests/test_shared_steps.py` already exists for exactly this — its docstring notes these scripts run as `bash <script>` inside both harness actions, so a non-zero exit turns an otherwise-successful agent run red, and shellcheck can't catch runtime behaviour. The stubbed-`gh` matrix in your description was the right check; committing it means the next edit to the tier boundaries can't silently re-fuse the spike tier to an abort. Four cases, driven through a fake `gh` on `PATH` at the real 2026-08-05 numbers (baseline 17 → spike 15, hard 27): 16 exits 0 and writes `TEND_CREATION_PAUSED_NOTE`; 28 exits non-zero with the hard-limit message and writes nothing; 3 passes clean; 11 PRs in 20 minutes still aborts. Full suite 340 passed.

One correction for the body: `tend-review` failed **4** times in the window, not 3, so the total is 36 rather than 35 — 11:47:16Z ([31002862308](https://github.com/max-sixty/tend/actions/runs/31002862308), the one you cite as first casualty), 16:19:29Z ([31024720942](https://github.com/max-sixty/tend/actions/runs/31024720942)), 16:24:43Z ([31025137115](https://github.com/max-sixty/tend/actions/runs/31025137115)) and 21:15:37Z ([31047860817](https://github.com/max-sixty/tend/actions/runs/31047860817)). Four PRs unreviewed, not three. Doesn't change the argument.

Worth noting alongside: the outage is invisible to `review-reviewers`' own window logic. `list-recent-runs.sh` anchors its dropped-tick recovery on the last **completed** run irrespective of conclusion, and 12 failed runs are all "completed", so this run's window floored at 22:47Z instead of reaching back across the gap — the twelve unanalysed hours would have been skipped silently. [#838](https://github.com/max-sixty/tend/pull/838) already fixes that half.

shared/steps/rate-limit-preflight.sh
```diff
@@ -29,8 +45,11 @@ PAST_POSTS=$(gh api "search/issues?q=author:${BOT}+repo:${REPO}+created:${SIX_DA
   --jq '.total_count' || echo 0)
 # spike_limit = 10 + 2 * daily_avg = 10 + 2 * (past_posts / 6) = 10 + past_posts / 3
 SPIKE_LIMIT=$((10 + PAST_POSTS / 3))
+# hard_limit = 10 + past_posts: a whole baseline week's output in a single day.
+# No productive day reaches it; a loop does.
+HARD_LIMIT=$((10 + PAST_POSTS))
```
tend-agent, line 59: `HARD_LIMIT` and `SPIKE_LIMIT` share the same `PAST_POSTS` term, so they converge as the baseline falls: at `PAST_POSTS = 0` both are 10 and this tier fires first on the 11th item — the spike tier below is dead code and the run aborts exactly as it did before this PR. At 1 and 2 the pause band is one and two items wide.

That matters because a depressed baseline is the trigger you identified: 07-30 through 08-02 fell inside the quota outage, and a longer outage (or a quiet repo, or an adopter's install day) pushes the baseline to 0. The guard would revert to hard-aborting every workflow for the rest of the UTC day in exactly the situation the PR is meant to cover.

Flooring the gap keeps the pause tier reachable at any baseline:

```suggestion
# hard_limit: a whole baseline week's output in a single day, and never less
# than 20 items above the spike tier — otherwise a baseline depressed by an
# outage or a quiet week collapses the two tiers together and the pause tier
# becomes unreachable. No productive day reaches it; a loop does.
HARD_LIMIT=$((10 + PAST_POSTS))
[ "$HARD_LIMIT" -lt $((SPIKE_LIMIT + 20)) ] && HARD_LIMIT=$((SPIKE_LIMIT + 20))
```

20 is a guess at the right floor — the constant is yours to pick; the shape is what I'd want. Worth a fifth test case pinning `PAST_POSTS = 0` to the pause tier, since that boundary is what silently reverted.

tend-agent (commented):
Splitting the daily guard by failure shape is the right call, and keeping the two abort tiers in pure shell keeps the injection-resistant floor intact. The tests exercise all four paths against a stubbed `gh` and pass locally (7 passed in `generator/tests/test_shared_steps.py`). Three things on the new tiers, one of which I think is load-bearing.

The main one: both tiers scale off the same 6-day baseline, so the fix does the least exactly where the diagnosed failure mode is worst. `hard = 10 + P` and `spike = 10 + P/3` converge as `P` shrinks — at `P = 0` they are both 10, so the pause tier is unreachable and the run hard-aborts at 11 items, identical to pre-PR behaviour. At `P = 1` and `P = 2` the pause band is one and two items wide. The PR's own diagnosis is that a depressed baseline is what caused 2026-08-05 (07-30 through 08-02 sat inside the quota outage), and a longer outage — or a genuinely quiet repo, or a fresh adopter on install day — drives `P` toward 0. Flooring the gap would keep the pause tier reachable regardless of baseline. Detail inline.

The second is about who the pause directive can actually be followed by. Inline on the note text.

<details><summary>Tier arithmetic across baselines</summary>

```
baseline=0   spike=10  hard=10     <- pause tier unreachable
baseline=1   spike=10  hard=11
baseline=2   spike=10  hard=12
baseline=3   spike=11  hard=13
baseline=6   spike=12  hard=16
baseline=17  spike=15  hard=27     <- the 2026-08-05 case
baseline=30  spike=20  hard=40
```

</details>

shared/steps/rate-limit-preflight.sh
```diff
@@ -41,9 +60,18 @@ if [ "$RECENT_ISSUES" -gt 10 ]; then
...
+if [ "$TODAY_POSTS" -gt "$HARD_LIMIT" ]; then
+  echo "::error::Rate limit: bot created ${TODAY_POSTS} items today, above hard limit of ${HARD_LIMIT} (baseline: ${PAST_POSTS} over past 6 days)"
   ABORT=true
+elif [ "$TODAY_POSTS" -gt "$SPIKE_LIMIT" ]; then
+  echo "::warning::Creation pause: bot created ${TODAY_POSTS} items today, above spike limit of ${SPIKE_LIMIT} (baseline: ${PAST_POSTS} over past 6 days). The run continues; opening new issues and PRs is off-limits for it."
+  if [ -n "${GITHUB_ENV:-}" ]; then
+    {
+      echo 'TEND_CREATION_PAUSED_NOTE<<TEND_EOF'
+      echo "Creation pause: you have already opened ${TODAY_POSTS} issues and pull requests on this repository today, above the daily limit of ${SPIKE_LIMIT} that guards against runaway creation. For this run, do not open a new issue or pull request. Nothing else is restricted — reviews, review replies, issue and PR comments, and pushes to branches that already exist all proceed as normal. If the work genuinely needs a new issue or PR, describe what you would have filed in a comment on the triggering thread and leave the filing to a maintainer."
```
tend-agent: The fallback here — "describe what you would have filed in a comment on the triggering thread" — assumes a triggering thread exists, but the workflows whose entire output is a new issue or PR are the scheduled and `workflow_run` ones that have none: `tend-ci-fix`, `tend-nightly`, `tend-weekly`, `tend-review-runs`, `review-reviewers`. On those runs the directive leaves the agent with no place to put the result, so the pause turns a loud failure (the run goes red, someone notices) into a silent one: the agent does the whole diagnosis, finds it cannot file, and the work is discarded with the runner.

`tend-ci-fix` is the sharpest case — main's CI is broken, the PR *is* the deliverable, and there is no thread to defer to.

I don't have a mechanism I've verified to suggest: the Claude harness runs the agent as the sandbox user with a curated env, so I can't claim `$GITHUB_STEP_SUMMARY` is reachable from inside the session without checking. Options that don't need one: name an existing tracking issue to comment on, or scope the pause to the event-driven workflows and let the threadless ones keep filing (the burst and hard tiers still cover them). Either way the note should say what to do when there's no thread, rather than leaving it to the agent to improvise.

shared/steps/rate-limit-preflight.sh
```diff
@@ -41,9 +60,18 @@ if [ "$RECENT_ISSUES" -gt 10 ]; then
...
   ABORT=true
 fi
-if [ "$TODAY_POSTS" -gt "$SPIKE_LIMIT" ]; then
-  echo "::error::Rate limit: bot created ${TODAY_POSTS} items today, above spike limit of ${SPIKE_LIMIT} (baseline: ${PAST_POSTS} over past 6 days)"
+if [ "$TODAY_POSTS" -gt "$HARD_LIMIT" ]; then
+  echo "::error::Rate limit: bot created ${TODAY_POSTS} items today, above hard limit of ${HARD_LIMIT} (baseline: ${PAST_POSTS} over past 6 days)"
   ABORT=true
+elif [ "$TODAY_POSTS" -gt "$SPIKE_LIMIT" ]; then
+  echo "::warning::Creation pause: bot created ${TODAY_POSTS} items today, above spike limit of ${SPIKE_LIMIT} (baseline: ${PAST_POSTS} over past 6 days). The run continues; opening new issues and PRs is off-limits for it."
+  if [ -n "${GITHUB_ENV:-}" ]; then
```
tend-agent: Minor: this writes the pause note whenever the daily count is in the spike band, including when a burst check has already set `ABORT=true` above — so a burst-tier abort can export a note on a run that then exits 1. Harmless today (the job dies before anything reads it), but it contradicts the invariant `test_rate_limit_hard_tier_still_aborts` asserts for the other abort tier ("exported a pause note on a run that aborts anyway").

```suggestion
  if [ "$ABORT" = false ] && [ -n "${GITHUB_ENV:-}" ]; then
```

shared/steps/rate-limit-preflight.sh
tend-agent: Agreed, and this was the load-bearing one — at a zero baseline the pause tier was unreachable and the run reverted to hard-aborting on the 11th item, which is the behaviour the PR exists to replace. Fixed in [1431367](https://github.com/max-sixty/tend/commit/14313674008ed263b5311d438293a7624fbb1136).

I took the shape and picked 10 rather than 20, so the floor only binds where the tiers actually collapse: the natural gap is `P - P/3 ≈ 2P/3`, which reaches 10 at a baseline of 15. A floor of 20 would have widened the hard tier for every baseline below 30, including the diagnosed 2026-08-05 case (hard 27 → 35), which loosens the ceiling in a case that didn't need it. 10 also matches the burst limit and the constant term in both formulas.

```
baseline=0   spike=10  hard=20  band=10
baseline=3   spike=11  hard=21  band=10
baseline=15  spike=15  hard=25  band=10   <- floor stops binding here
baseline=17  spike=15  hard=27  band=12   <- 2026-08-05, unchanged
baseline=30  spike=20  hard=40  band=20
```

Added `test_rate_limit_pause_tier_survives_a_zero_baseline` pinning `PAST_POSTS = 0` to the pause tier and asserting the floored `hard: 20` reaches the log line, since that boundary is what silently reverted.

shared/steps/rate-limit-preflight.sh
tend-agent: Right that the fallback assumed a thread that the creation-only workflows don't have, and right that the pause was converting a loud failure into a silent one. Addressed in [1431367](https://github.com/max-sixty/tend/commit/14313674008ed263b5311d438293a7624fbb1136), taking neither of your two options — there's an existing mechanism that fits.

`$GITHUB_STEP_SUMMARY` isn't reachable from inside the session, and you were right not to claim otherwise: `claude/action.yaml` explicitly denylists it (with `GITHUB_{ENV,PATH,OUTPUT,STATE}` and `GITHUB_TOKEN`) when building the sandbox user's env, so the sandbox isn't handed a channel into later steps. But the file handoff around it already exists — [`claude/action.yaml:521-527`](https://github.com/max-sixty/tend/blob/14313674008ed263b5311d438293a7624fbb1136/claude/action.yaml#L521-L527) copies `/tmp/claude/step-summary.md` into the job summary after the agent exits, and two bundled skills already use it: [`review-runs`](https://github.com/max-sixty/tend/blob/14313674008ed263b5311d438293a7624fbb1136/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L249) and [`review-reviewers`](https://github.com/max-sixty/tend/blob/14313674008ed263b5311d438293a7624fbb1136/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L367). So the note now branches: comment on the triggering thread if there is one; otherwise write the finding there and leave the filing to a maintainer or to the next run after the UTC reset. `test_rate_limit_pause_note_covers_runs_with_no_thread` pins that the note names a destination at all.

Checking that turned up a latent gap worth folding in: the Codex action had no equivalent step, so those two skills' writes have been going nowhere under Codex regardless of this PR. Added the same step to `codex/action.yaml`.

On the other two options — naming an existing tracking issue needs one to exist, and scoping the pause to event-driven workflows would empty the tier, since the threadless sweeps are where most of the 16 items on 2026-08-05 came from. `tend-ci-fix` stays the weakest case: a broken `main` is urgent and the PR really is the deliverable, so a paused run writes the diagnosis to the summary and a human has to act on it. If that turns out to matter in practice, the narrow fix is to exempt `tend-ci-fix` by event rather than to soften the tier — I'd rather see it happen once before encoding it.

shared/steps/rate-limit-preflight.sh
tend-agent: Taken, with the `ABORT` guard as suggested. The two conditions are independent — a run can be over the daily spike limit *and* bursting — so the note was reachable on a run that exits 1.

`test_rate_limit_burst_abort_exports_no_pause_note` covers it (today=16 against spike 15, plus 11 issues in 20 minutes): aborts, exports nothing. That extends the invariant `test_rate_limit_hard_tier_still_aborts` was asserting for only one of the two abort tiers.

tend-agent (commented):
The floor and the `ABORT` guard both check out — at a zero baseline the pause band is 11–20 with the hard tier at 20, and a burst abort now exports nothing. Tests pass locally (10 passed in `generator/tests/test_shared_steps.py`). One gap in the new threadless-run destination, inline.

shared/steps/rate-limit-preflight.sh
```diff
@@ -41,9 +72,21 @@ if [ "$RECENT_ISSUES" -gt 10 ]; then
...
+  # so exporting a creation directive for an agent that never starts is dead
+  # state that contradicts the invariant the abort tiers otherwise hold.
+  if [ "$ABORT" = false ] && [ -n "${GITHUB_ENV:-}" ]; then
+    {
+      echo 'TEND_CREATION_PAUSED_NOTE<<TEND_EOF'
+      echo "Creation pause: you have already opened ${TODAY_POSTS} issues and pull requests on this repository today, above the daily limit of ${SPIKE_LIMIT} that guards against runaway creation. For this run, do not open a new issue or pull request. Nothing else is restricted — reviews, review replies, issue and PR comments, and pushes to branches that already exist all proceed as normal. Do the work either way; only the filing is paused. Record the result where it will be read rather than dropping it: if this run has a triggering issue or PR, describe what you would have filed in a comment on that thread. If it does not — a scheduled or workflow_run run has no thread — write the finding to /tmp/claude/step-summary.md with the Write tool, which a later step copies into this run's job summary, and leave the filing to a maintainer or to the next run after the daily count resets at UTC midnight."
```
tend-agent: Neither action creates `/tmp/claude`, and the two skills that already use this path (`review-runs`, `review-reviewers`) both run `mkdir -p /tmp/claude` before writing. Claude's `Write` tool creates missing parents, so the directive works there as written — but the same text is now staged into Codex's `AGENTS.md`, and Codex has no `Write` tool: it writes files through the shell, where `cat > /tmp/claude/step-summary.md` fails on the missing directory. The paused threadless run then drops the finding and the new copy step in `codex/action.yaml` has nothing to read — which is the failure this commit is closing.

Making the directive create the directory keeps the fix harness-neutral:

```suggestion
      echo "Creation pause: you have already opened ${TODAY_POSTS} issues and pull requests on this repository today, above the daily limit of ${SPIKE_LIMIT} that guards against runaway creation. For this run, do not open a new issue or pull request. Nothing else is restricted — reviews, review replies, issue and PR comments, and pushes to branches that already exist all proceed as normal. Do the work either way; only the filing is paused. Record the result where it will be read rather than dropping it: if this run has a triggering issue or PR, describe what you would have filed in a comment on that thread. If it does not — a scheduled or workflow_run run has no thread — create /tmp/claude if it does not exist and write the finding to /tmp/claude/step-summary.md, which a later step copies into this run's job summary, and leave the filing to a maintainer or to the next run after the daily count resets at UTC midnight."
```

A deterministic `mkdir -p /tmp/claude` in the Run Codex step would be sturdier still (it doesn't depend on the model following the instruction), but it can't go in the shared preflight: that runs as the runner user before the Claude sandbox user exists, so a `0755` directory owned by the runner would block the Claude side. Happy to push either shape.

tend-agent: Good catch — the directive was Claude-shaped in two ways, not one: it also said "with the Write tool", which is text staged verbatim into Codex's `AGENTS.md`. Both fixed in [9e0a890](https://github.com/max-sixty/tend/commit/9e0a8900dda37fad66dcde3de7ec6f98cf55fb0d).

I took the deterministic shape rather than the suggestion, since the whole reason the copy step exists is that the paused run's output shouldn't depend on the model getting an instruction right. Your objection to putting it in the shared preflight holds, so it goes in each harness where the ownership is already known:

- `proxy/setup-sandbox.sh` — `sudo -u "$SANDBOX" mkdir -p /tmp/claude`, next to `TEND_RUN_DIR`, which has the same runner-reads/sandbox-writes shape. As the sandbox user for the reason you gave: `/tmp` is `1777`, so a runner-owned directory there is unwritable by the agent.
- `codex/action.yaml` — plain `mkdir -p` at the top of the Run Codex step; the agent is the runner user there.

The directive keeps its `/tmp/claude/step-summary.md` destination and no longer names a tool. `test_skill_summary_dir_is_created_by_both_harnesses` pins both creation sites and both copy steps, so removing either end fails rather than silently dropping the finding.

That also fixes the pre-existing case: `review-runs` and `review-reviewers` carry their own `mkdir -p`, so they were fine on Claude, but under Codex their writes went nowhere at all until the copy step landed.

tend-agent:
Corroborating evidence from the `review-runs` window (2026-08-05T08:41Z → 2026-08-06T08:41Z), on the one part of the diagnosis that isn't in this PR body or #855: the numerator was demand-driven, not just the denominator being depressed.

Both write-ups attribute the trip to a supply-side artifact — the 07-30…08-02 quota outage suppressing the baseline to 17. That holds, but it only explains half the ratio. Attributing all 16 items by branch prefix:

| Source | Items | Which |
|---|---|---|
| `tend-triage` on inbound consumer-bot issues (`fix/issue-*`) | **6** | #833, #834, #835, #837, #842, #847 |
| `review-reviewers` own PRs (`hourly/review-*`) | 4 | #836, #838, #849, #850 |
| `tend-nightly` / `tend-mention` / other `fix/*` | 3 | #840, #843, #844 |
| `tend-agent`-filed issues | 2 | #839, #848 |
| `review-runs` (`daily/review-runs-*`) | 1 | #845 |

The largest single block is `tend-triage` answering issues that **other repos' tend installs** filed into this one — #827/#830/#828/#829 from `worktrunk-bot` and `cargo-affected-bot` on 08-04, #841 (`worktrunk-bot`) and #846 (`prql-bot`) on 08-05. Six `fix/issue-*` PRs in a day is 3× the previous maximum: that branch prefix has run at 1–2/day since June, never above 2 before 08-05.

Two things follow.

**It settles open question 1 in #855 against the `creates_issues` shape.** Exempting `review` and `notifications` per-workflow would not have prevented this trip — those two create nothing and so contributed 0 of the 16. The workflow that contributed most, `triage`, opens PRs and could not be exempted without gutting the guard. A per-run pause is the only shape that covers it, which is what this PR does.

**The ratio has no ceiling, because its two halves are independent.** The gate's baseline is `tend-agent`'s own 6-day history in this repo, but the input rate is set by four consumer repos' `review-reviewers` legs filing inbound. Nothing couples them, so a quiet week here plus a busy week upstream drives `today / baseline` arbitrarily high with nothing malfunctioning at either end — which is a stronger argument for demoting the spike tier than the one-off outage-calibration story alone.

The forward-looking caveat, which I don't think changes the design: replaying this window against the new tiers gives 16 vs spike 15, hard 27 — a pause, correctly. But 6 of the 6 paused creations would have been triage PRs fixing consumer-reported bugs, so on an inbound-heavy day the pause note's fallback path (describe what you would have filed in a thread comment) becomes triage's normal outcome rather than a rare one. That path is fine for a scheduled run writing to the job summary; for triage the deliverable lands as a comment on an issue whose reporter is a bot in another repo, and nothing re-drives it into a PR once the day rolls over. Worth a look at whether the pause should defer creation to the next run rather than substitute a comment for it — as a follow-up, not a change to this PR.

<details><summary>Method</summary>

- Branch attribution: `gh pr view <n> --json headRefName` for each of the 14 PRs in `search/issues?q=author:tend-agent+repo:max-sixty/tend+created:2026-08-05`.
- Inbound issue authors: `gh issue list --state all --json number,author,createdAt --jq '.[] | select(.createdAt > "2026-08-03")'` — 12 consumer-bot issues in 7 days (`cargo-affected-bot` 5, `worktrunk-bot` 5, `prql-bot` 2) against 10 `tend-agent` ones.
- `fix/issue-*` daily history: grouped `gh pr list --state all --author tend-agent --limit 100 --json headRefName,createdAt` — max 2/day across June and July, 6 on 08-05.
- Gate arithmetic re-derived from [`shared/steps/rate-limit-preflight.sh`](https://github.com/max-sixty/tend/blob/3b46066e6d696e6c92461a768d5fbe17d4b68dfe/shared/steps/rate-limit-preflight.sh#L26-L31) at `main`.

</details>

tend-agent:
Tried to merge `origin/main` into this branch and stopped: the conflict is a design collision, not a textual one, and resolving it means choosing between two shipped answers to the same incident. Aborting rather than picking one.

## What landed

5b30d22 — "feat(rate-limit): let a maintainer approve past the spike limit" (#874) — rewrote the same spike branch of `shared/steps/rate-limit-preflight.sh` from the same evidence (the 2026-08-05 trip, `tend-review` dying from 11:47Z). It reaches a different remedy:

| | this PR (#856) | #874 (on main) |
|---|---|---|
| Spike trip | run continues, creation paused via `TEND_CREATION_PAUSED_NOTE` in the prompt | run still aborts |
| Notice | `::warning::` annotation | files/reopens a `tend-rate-limit` issue listing refused runs |
| Reset | none needed — nothing was blocked | maintainer closes the issue; each close doubles the ceiling for the UTC day |
| Second abort tier | new hard limit, `max(10 + past, spike + 10)` | none; `SPIKE_LIMIT << APPROVALS` is the only ceiling |
| Bot identity | `BOT_NAME` env | `gh api user` (login + id), `BOT_NAME` removed |

So the notice-and-reset half of this PR's problem statement is now covered. The other half is not: #874 still refuses reviews, mention replies, triage comments, and CI fixes on a spike trip — work that creates nothing and so cannot contribute to the count being enforced — until a person notices the issue and closes it. That gap is real and this PR is the only thing that addresses it.

## Which hunks need human judgment

1. **`shared/steps/rate-limit-preflight.sh`, the whole `TODAY_POSTS -gt SPIKE_LIMIT` branch.** Both sides replaced it end to end. A combined version has to answer questions neither PR asked: does a spike trip now both pause creation *and* file the pause issue (whose body — "The bot stopped before doing any work" — is false under a pause)? Does an approval un-pause creation, or lift the hard tier, or both? If approvals double the spike ceiling while the hard tier sits at `spike + 10`, one approval puts the doubled ceiling above the hard limit and the run hard-aborts anyway, making the approval inert — the tiers have to be re-derived together, not merged.

2. **`shared/steps/rate-limit-preflight.sh` header comment (the file's first block).** Two incompatible descriptions of what the spike tier does.

3. **`generator/tests/test_shared_steps.py`, both hunks.** Not reconcilable mechanically: this PR's `_preflight` helper passes `BOT_NAME`, which #874 deleted, and its fake `gh` answers four count queries only — no `gh api user`, no `issue list/create/comment/reopen`, no fake `date`/`sleep`. Every test here would need rewriting against the new stub regardless of which tier design wins.

4. **Clean-merged but self-contradictory.** `docs/security-model.md` and `README.md` merged without conflict into text that states both designs. The merged security-model reads "A softer daily spike threshold ... does not abort: the run proceeds with a creation-pause directive" and then, two paragraphs down, "On a spike trip the run files or reopens a `tend-rate-limit` issue listing the runs it refused."

## Recommendation

Close this PR, and if the "don't drop work that creates nothing" gap is still wanted, take it as a small change on top of #874's script rather than a merge of this branch — the shared code underneath it has moved far enough that the diff is a rewrite either way.

Two hunks here are worth salvaging first, because they are independent of the tier design and fix a live bug on main:

- **Nothing on main creates `/tmp/claude`.** `claude/action.yaml` copies `/tmp/claude/step-summary.md` into the job summary, and two bundled skills write there, but `/tmp` is `1777` and no step creates the directory as the sandbox user. This PR's `proxy/setup-sandbox.sh` hunk (`sudo -u "$SANDBOX" mkdir -p /tmp/claude`) is the fix.
- **`codex/action.yaml` has no step-summary copy at all** — the Codex harness silently drops anything a skill writes there. This PR adds both the `mkdir` and the copy step.

Neither touches the rate-limit script and both merge cleanly today.

> _Posted by tend while resolving merge conflicts on this PR._

tend-agent:
Left conflicted on purpose — this one isn't a mechanical resolution.

`main` has since rewritten the same spike tier along a different axis: the trip still aborts, but it now files (or reopens) a `tend-rate-limit` issue naming the refused runs, and a close by a non-bot actor doubles the ceiling for the rest of the UTC day. The identity the counts are keyed on also moved from `BOT_NAME` to the token (`gh api user`), and the tests around all of it were replaced wholesale — so this branch's `rate-limit-preflight.sh` and `test_shared_steps.py` conflict against a design, not against a few lines.

The two answers to #855 are genuinely different: this PR keeps the run alive with creation paused (`TEND_CREATION_PAUSED_NOTE`), so review and mention work proceeds unattended; `main` stops the run but gives a maintainer a one-click way to resume the day. They can compose — pause creation *and* file the notice — but then the issue body's "these runs were refused, re-run them once this is closed" is no longer true of the run that filed it, and what a close should mean when nothing was refused is a call for a maintainer rather than a conflict resolution.

Not rebasing it blind. If the composed shape is the one you want, say so and I'll rebase onto the current tier and rewrite the notice accordingly; if `main`'s resumable abort is the answer to #855, this can close.

<details><summary>What conflicts</summary>

- `shared/steps/rate-limit-preflight.sh` — the whole `TODAY_POSTS > SPIKE_LIMIT` block, plus the header comment describing the tiers.
- `generator/tests/test_shared_steps.py` — this branch's `FAKE_GH_COUNTS` stub against `main`'s `FAKE_GH_RATE_LIMIT`, which runs the script's own `--jq` through real jq because the approval filter is the behaviour under test.

The non-conflicting half of this branch (`codex/action.yaml`, `proxy/setup-sandbox.sh`, `shared/steps/compose-system-prompt.sh`, the docs) carries the note into the prompt and survives either way.

</details>

PR #858, opened 2026-08-06
fix(review-reviewers): give each matrix leg its own PR branch name

tend-agent:
## Problem

`review-reviewers` matrixes over five target repos ([`review-reviewers.yaml`](https://github.com/max-sixty/tend/blob/f3e309a475af03147ce6290104c85baafe6ad9f5/.github/workflows/review-reviewers.yaml)), and all five legs of a tick share one `$GITHUB_RUN_ID`. Step 5 of the skill hands every leg the same branch recipe:

> **PR** (default): Branch `hourly/review-$GITHUB_RUN_ID`, fix, commit, push, …

So the moment two legs of the same tick both find something worth a PR, they contend for one ref. The loser's `git push` is rejected as a non-fast-forward and it has to recover mid-session; the ordering that hurts more is the other one, where the second leg pushes *before* the first opens its PR and the two unrelated concerns land on a single branch — the atomic-PR rule broken by the branch name, not by any judgement the agent made.

This just happened on run 31058673934: the `max-sixty/cargo-affected` leg's push to `hourly/review-31058673934` was rejected against a sibling leg's commit, which by then was #856. It recovered by suffixing the target name (#857), and an earlier tick shows the same improvisation baked into `hourly/review-30962483562-numbagg` — legs have been working around the recipe rather than following it.

The run ID is non-unique along a second axis too: the same section caps a leg at two PRs (`**Limit to at most 2 PRs per run**`), and a leg that acts on two findings computes one name for both — same collision, one job instead of two.

## Fix

Put both discriminators in the recipe: `hourly/review-$GITHUB_RUN_ID-<target-repo-name>-<topic-slug>`. The target name is unique across the matrix by construction, the topic slug separates a single leg's two PRs, and the branch stays greppable back to its run.

Scoped to this skill. `review-runs` carries the same `$GITHUB_RUN_ID`-only shape, but `tend-review-runs.yaml` has no matrix, so a run there is a single job and the name can't collide — no change needed and none made.

**Overlaps #845 on this line.** That PR rewrites the same recipe to rename the prefix (`hourly/` → `review-reviewers/`) as part of its cadence change, and keeps `$GITHUB_RUN_ID` as the whole discriminator — so it ships the collision under a new name. The two are orthogonal in substance and conflict only textually: whichever lands second wants `<prefix>/review-$GITHUB_RUN_ID-<target-repo-name>-<topic-slug>`, taking the prefix from #845 and the suffixes from here.

## Gate assessment

- **Evidence level**: High — the collision is deterministic given two acting legs, and there are two independent traces of it (this run's rejected push, and a prior tick's hand-suffixed branch). The one-leg-two-PRs axis is structural rather than traced.
- **Structural, not stochastic**: no decision point. Every leg is told to compute the same name from the same variable; replayed ten times it collides ten times.
- **Change type**: targeted fix — one line of the recipe.
- **Passes both gates.**

### DIFF
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -353,7 +353,7 @@ Search titles AND bodies for related keywords. Only comment on existing issues i
 
 **Prefer PRs over issues.** A PR with a clear description is immediately actionable.
 
-- **PR** (default): Branch `hourly/review-$GITHUB_RUN_ID-<target-repo-name>`, fix, commit, push, create with label `claude-behavior`. The workflow matrixes over target repos and every leg carries the same `$GITHUB_RUN_ID`, so the run ID alone is not a unique branch name — append the target's repo name (the part after the `/`) to keep two legs that both find something from racing the same ref. Put full analysis in PR description (run ID, outcome evidence, root cause, **gate assessment** including historical evidence count). Don't also create a separate issue.
+- **PR** (default): Branch `hourly/review-$GITHUB_RUN_ID-<target-repo-name>-<topic-slug>`, fix, commit, push, create with label `claude-behavior`. `$GITHUB_RUN_ID` alone is not a unique branch name: every matrix leg of a tick carries the same one, and a single leg may open two PRs (see the 2-PR limit below). The target's repo name (the part after the `/`) keeps two legs from racing the same ref; the topic slug keeps one leg's two PRs from doing the same. Put full analysis in PR description (run ID, outcome evidence, root cause, **gate assessment** including historical evidence count). Don't also create a separate issue.
 - **Issue** (fallback): Only for problems too large or ambiguous to fix directly. Include run ID, outcome evidence, root cause analysis.
 
 Group multiple findings by broad theme. **Limit to at most 2 PRs per run** — if you have more findings, pick the highest-confidence ones and record the rest in the evidence gist.
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -353,7 +353,7 @@ Search titles AND bodies for related keywords. Only comment on existing issues i
 
 **Prefer PRs over issues.** A PR with a clear description is immediately actionable.
 
-- **PR** (default): Branch `hourly/review-$GITHUB_RUN_ID`, fix, commit, push, create with label `claude-behavior`. Put full analysis in PR description (run ID, outcome evidence, root cause, **gate assessment** including historical evidence count). Don't also create a separate issue.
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -353,7 +353,7 @@ Search titles AND bodies for related keywords. Only comment on existing issues i
 
 **Prefer PRs over issues.** A PR with a clear description is immediately actionable.
 
-- **PR** (default): Branch `hourly/review-$GITHUB_RUN_ID`, fix, commit, push, create with label `claude-behavior`. Put full analysis in PR description (run ID, outcome evidence, root cause, **gate assessment** including historical evidence count). Don't also create a separate issue.
+- **PR** (default): Branch `hourly/review-$GITHUB_RUN_ID-<target-repo-name>`, fix, commit, push, create with label `claude-behavior`. The workflow matrixes over target repos and every leg carries the same `$GITHUB_RUN_ID`, so the run ID alone is not a unique branch name — append the target's repo name (the part after the `/`) to keep two legs that both find something from racing the same ref. Put full analysis in PR description (run ID, outcome evidence, root cause, **gate assessment** including historical evidence count). Don't also create a separate issue.
```
tend-agent: The target suffix is unique per matrix leg, but not per PR — and the paragraph two lines down allows a leg to open two. Adding a topic component makes the name unique along both axes:

```suggestion
- **PR** (default): Branch `hourly/review-$GITHUB_RUN_ID-<target-repo-name>-<topic-slug>`, fix, commit, push, create with label `claude-behavior`. The workflow matrixes over target repos and every leg carries the same `$GITHUB_RUN_ID`, so the run ID alone is not a unique branch name — the target suffix keeps two legs that both find something from racing the same ref, and the topic slug keeps one leg's two PRs (see the 2-PR limit below) from doing the same. Put full analysis in PR description (run ID, outcome evidence, root cause, **gate assessment** including historical evidence count). Don't also create a separate issue.
```

tend-agent (commented):
The diagnosis checks out against the traces — #856 sits on `hourly/review-31058673934` and #857 on `hourly/review-31058673934-cargo-affected`, exactly the rejected-push-then-hand-suffix recovery the description describes. The scoping claim holds too: `tend-review-runs.yaml` has no `strategy.matrix`, so `daily/review-runs-$GITHUB_RUN_ID` can't collide the same way.

One gap: the discriminator is per *leg*, but the same section sanctions up to two PRs from one leg ("**Limit to at most 2 PRs per run**"). A leg that finds two things computes one branch name for both, so the second `gh pr create` fails against the head branch the first already claimed — and if the second fix is committed before that failure surfaces, two unrelated concerns share a branch, which is the atomic-PR breakage this PR is closing. Suggestion inline adds a per-PR component so the recipe covers both axes.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Good catch — the two-PRs-from-one-leg axis is real and the target suffix doesn't cover it. Applied in c8628f6, with the prose tightened a little from the suggestion so the rule leads and the reasoning follows:

> Branch `hourly/review-$GITHUB_RUN_ID-<target-repo-name>-<topic-slug>` … `$GITHUB_RUN_ID` alone is not a unique branch name: every matrix leg of a tick carries the same one, and a single leg may open two PRs (see the 2-PR limit below). The target's repo name (the part after the `/`) keeps two legs from racing the same ref; the topic slug keeps one leg's two PRs from doing the same.

PR description updated to describe both axes, including the #845 merge note, which now wants `<prefix>/review-$GITHUB_RUN_ID-<target-repo-name>-<topic-slug>`.

PR #861, opened 2026-08-06
config: drop the secret-name overrides

max-sixty:
`secrets.bot_token`, `.claude_token`, `.anthropic_api_key` and `.openai_key` renamed the four operational secrets per adopter. No config anywhere set one — every adopter and tend itself run the defaults, and the only thing exercising the path was a test fixture.

The case they existed for is closed. The overrides let an adopter whose repo already held a differently-named PAT point tend at it rather than minting a duplicate; since the operational secrets moved into the `tend` environment, `install-tend` creates that environment and fills it from scratch, so there is no pre-existing secret whose name is worth keeping. An org-level copy is an explicit `check_secrets` failure rather than a stand-in, so that route is closed too. `install-tend`'s §1 was still telling adopters to reach for the override in exactly the case that no longer works.

The names become module constants in `config.py`. The templates read them as Jinja globals, following `tend_environment`, so the names `tend check` looks for and the names the workflows read cannot drift; `check_claude_auth` and `check_codex_auth` no longer need a `Config` at all. The `secrets.allowed` refusal now intersects against `OPERATIONAL_SECRETS` and behaves as before. A leftover override key fails with the rename to make, rather than being warned past into workflows that read a secret the adopter doesn't have.

`test_custom_secrets` had nothing left to assert; in its place `test_workflows_read_only_the_operational_secrets` checks, per harness, that every stored secret a workflow reads is one `tend check` verifies — the bot token plus that harness's own auth secret, so a claude workflow reading `secrets.OPENAI_API_KEY` fails as surely as one reading a name nothing defines.

Rendered YAML is byte-identical — every regtest output is unchanged, so adopters see no diff on their next regen. 338 tests pass and `pre-commit run --all-files` is clean. `uvx`-equivalent `tend init --dry-run` against this repo's config emits `TEND_BOT_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_API_KEY` and `GITHUB_TOKEN` and nothing else; `tend check` reports the same two pre-existing failures on `CLAUDE_CODE_OAUTH_TOKEN` still sitting at repo level, which are the outstanding half of the credential migration `TODO.md` tracks and untouched here.

### DIFF
--- a/generator/tests/test_generate.py
+++ b/generator/tests/test_generate.py
@@ -13,7 +13,13 @@ import click
 from click.testing import CliRunner
 
 from tend.cli import main
-from tend.config import BOT_TOKEN_SECRET, OPERATIONAL_SECRETS, Config
+from tend.config import (
+    ANTHROPIC_API_KEY_SECRET,
+    BOT_TOKEN_SECRET,
+    CLAUDE_TOKEN_SECRET,
+    OPENAI_KEY_SECRET,
+    Config,
+)
 from tend.workflows import (
     _deep_merge,
     GENERATORS,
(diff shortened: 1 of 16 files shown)


generator/tests/test_generate.py
```diff
@@ -13,7 +13,7 @@
 from click.testing import CliRunner
 
 from tend.cli import main
-from tend.config import Config
+from tend.config import BOT_TOKEN_SECRET, OPERATIONAL_SECRETS, Config
```
tend-agent:
```suggestion
from tend.config import (
    ANTHROPIC_API_KEY_SECRET,
    BOT_TOKEN_SECRET,
    CLAUDE_TOKEN_SECRET,
    OPENAI_KEY_SECRET,
    Config,
)
```

`OPERATIONAL_SECRETS` becomes unused here with the other suggestion applied.

generator/tests/test_generate.py
```diff
@@ -271,20 +271,25 @@ def test_empty_setup_no_blank_lines(tmp_path: Path) -> None:
...
     for wf in generate_all(cfg):
-        assert "MY_BOT_PAT" in wf.content, f"{wf.filename} missing custom bot token"
-        assert "MY_CLAUDE" in wf.content, f"{wf.filename} missing custom claude token"
-        assert "MY_API_KEY" in wf.content, (
-            f"{wf.filename} missing custom anthropic_api_key secret"
+        read = set(re.findall(r"secrets\.([A-Za-z0-9_]+)", wf.content))
+        assert read - {"GITHUB_TOKEN"} <= OPERATIONAL_SECRETS, (
+            f"{wf.filename} reads a secret outside the operational set: "
+            f"{sorted(read - {'GITHUB_TOKEN'} - OPERATIONAL_SECRETS)}"
         )
```
tend-agent, line 305:
```suggestion
    verified = {BOT_TOKEN_SECRET} | (
        {CLAUDE_TOKEN_SECRET, ANTHROPIC_API_KEY_SECRET}
        if harness == "claude"
        else {OPENAI_KEY_SECRET}
    )
    cfg = Config.load(_minimal_config(tmp_path, f"harness: {harness}\n"))
    for wf in generate_all(cfg):
        read = set(re.findall(r"secrets\.([A-Za-z0-9_]+)", wf.content))
        assert read - {"GITHUB_TOKEN"} <= verified, (
            f"{wf.filename} reads a secret the {harness} harness does not "
            f"provision: {sorted(read - {'GITHUB_TOKEN'} - verified)}"
        )
```

`tend check` verifies the bot token plus the selected harness's auth secret, not all four — so `<= OPERATIONAL_SECRETS` accepts `secrets.OPENAI_API_KEY` in a claude workflow (and the reverse), which is the same empty-token-at-run-time failure. Narrowing per harness is what makes the parametrize do work.

tend-agent (approved):
The new invariant test is parameterized on harness but its expected set isn't, so a template naming the *other* harness's auth secret still passes — the docstring's "a secret that nothing provisions" case. Adding `x: ${{ secrets.<<claude_token_secret>> }}` to the codex branch of `agent_step` passes today's assertion; with the two suggestions below it fails as `tend-review.yaml reads a secret the codex harness does not provision: ['CLAUDE_CODE_OAUTH_TOKEN']`. 338 tests pass with both applied.

generator/tests/test_generate.py
max-sixty: Applied. Verified the way you described: with `x: ${{ secrets.<<claude_token_secret>> }}` added to the codex branch of `agent_step`, the old assertion passes and the new one fails as `tend-review.yaml reads a secret the codex harness does not provision: ['CLAUDE_CODE_OAUTH_TOKEN']`. 338 pass with the template restored.

PR #864, opened 2026-08-06
skills(review-reviewers): require a named non-bot actor before calling output accepted

tend-agent:
## Problem

Step 2 asks the survey subagent whether each run's output was "accepted or rejected" but never says who has to do the accepting. Nothing in the prompt distinguishes a human merging a PR from the bot replying to its own review thread, so the subagent fills the gap with the most natural reading — a reply arrived, therefore someone accepted it — and reports a self-conversation as human acceptance.

That is the worst-shaped failure this skill has: it produces a **false all-clear**. A survey that credits the bot's own reply to a human doesn't add noise a later gate can filter, it removes the signal that would have sent the leg to Step 3.

## Evidence

This leg ([run 31081926964](https://github.com/max-sixty/tend/actions/runs/31081926964), target `max-sixty/cargo-affected`). The survey subagent reported PR [#77](https://github.com/max-sixty/cargo-affected/pull/77) twice under "Runs with accepted output":

> Outcome: ACCEPTED - human developer responded positively to feedback
> Outcome: ACCEPTED - human developer incorporated feedback

and concluded `PR 77: Developer accepted and incorporated bot feedback (2 cycles of acceptance)`. There is no developer. The PR is bot-authored and every actor on it is the bot:

```
$ gh api "repos/max-sixty/cargo-affected/pulls/77/comments?per_page=100" --jq '[.[].user.login] | unique'
["cargo-affected-bot"]
```

The same report also listed the PR author as `cargo-affected-bot`, so the contradiction was internal to the summary — the acceptance claim was reached without ever comparing the replying login against the bot's.

This is the fourth cheap-subagent mis-attribution recorded in the [evidence gist](https://gist.github.com/dca23a6e6a0d8cae2665944ba31676fb), and the second of this specific sub-class (bot activity attributed to a human):

| Occurrence | Shape |
| --- | --- |
| gist "Analysis-side", leg ending 07:5x | inline replies + commit `551e059` credited to "human applied suggested changes" |
| gist "Analysis-side", following leg | PR #54 bodies credited to a run that had not yet started |
| gist "Analysis-side", streak-break leg | fabricated no-op causes; window boundary ignored |
| **this leg** | **bot's own reply chain on PR #77 reported as human acceptance** |

Prior legs' response was to stop delegating — ten consecutive legs record "this leg used **no** survey subagent". That is a per-leg workaround for a defect in the shared prompt, and it leaves the skill's stated default (`Delegate all broad exploration to a cheap subagent`) pointing at a step that mis-reports actors.

## Fix

Three small changes to Step 2, all inside the prompt template:

- Require the subagent to **name the login** behind every acceptance/rejection signal, sourced from a single `gh pr view --json number,state,author,mergedBy,reviews,comments,commits` that covers all five actor surfaces — merge actor, reviews (with state), inline comments, conversation comments, and commits. A comments-only check can't source it: neither comments endpoint carries review records or the merge actor, so a maintainer merging a bot PR without commenting would come back as bot-only. Inline commenters need no separate call — every inline comment, including a standalone reply, belongs to a review record that `reviews` returns. The block notes that `commits` truncates at 100 (oldest-first) while `comments` and `reviews` paginate in full.
- Give bot-only threads their own report bucket (`bot-only — no human signal`) so they can't be silently filed under "accepted".
- Tell the main agent to verify actor attribution before it enters a finding, and to judge bot-only threads on content rather than treating the bucket as either outcome.

The rule is stated so it stays correct when the bot legitimately works on its own PR: bot-only is *not* a defect (a self-review followed by the author's fix is designed behavior), it just isn't **acceptance**. Nothing here discourages delegation — it fixes the instruction the delegation was missing.

## Gate assessment

- **Evidence level**: High — consistent pattern across multiple sessions. 4 occurrences total, 2 in the human/bot sub-class this change addresses. High needs 2–3.
- **Structural or stochastic**: structural in the prompt. The subagent is never asked for actor identity, so no wording of the reply can be relied on to supply it; which specific claim it invents is stochastic, the omission is not.
- **Change type**: targeted fix — a missing requirement in an existing prompt, plus a report bucket. Normal bar, met.
- **Passes both gates.**

Second finding this leg did **not** pass and is recorded in the gist instead: the three-hop `tend-mention` chain on the same PR #77. That one is already fixed by open PR #849, and I've added this window's evidence there rather than opening a duplicate.


plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -228,6 +228,13 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
 > 1. Did the bot produce visible output (review, comment, issue action, commit)?
 > 2. If yes, was the output accepted or rejected?
 >
+> **Acceptance needs a non-bot actor — name it.** For every acceptance or rejection signal, report the login that produced it: who merged, who replied, who pushed the follow-up. `$BOT_LOGIN` reviewing, replying to, or pushing to its own PR is the bot talking to itself, not acceptance. List the thread's distinct participants before calling anything accepted, and report a bot-only thread as `bot-only — no human signal`:
+>
+> ```bash
+> gh api "repos/$ARGUMENTS/pulls/<pr>/comments?per_page=100" --jq '[.[].user.login] | unique'
+> gh api "repos/$ARGUMENTS/issues/<n>/comments?per_page=100" --jq '[.[].user.login] | unique'
```
tend-agent: Neither endpoint sees a review record or the merge actor, so the report format's `PR merged by <login>` can't be sourced from this block. Adding the reviews list and the PR object covers both, and `.merged_by.login` is `null` on an open PR rather than an error:

```suggestion
> gh api "repos/$ARGUMENTS/pulls/<pr>/comments?per_page=100" --jq '[.[].user.login] | unique'
> gh api "repos/$ARGUMENTS/pulls/<pr>/reviews" --jq '[.[].user.login] | unique'
> gh api "repos/$ARGUMENTS/issues/<n>/comments?per_page=100" --jq '[.[].user.login] | unique'
> gh api "repos/$ARGUMENTS/pulls/<pr>" --jq '{author: .user.login, merged_by: .merged_by.login}'
```

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -294,16 +301,19 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
...
 > ## Runs with concerning output
-> - <run-id>: <workflow> on PR #N — <signal> (e.g., "human posted CHANGES_REQUESTED")
+> - <run-id>: <workflow> on PR #N — <signal> by <login> (e.g., "human posted CHANGES_REQUESTED")
 >
 > ## Sanity check
 > <note if zero bot activity found across all runs — may indicate systemic failure>
 > ```
 
-Review the subagent's summary. If all outputs are accepted and no sanity-check flags, skip to Step 6 (summary). If concerning outcomes exist, continue to Step 3.
+Review the subagent's summary, and verify any actor attribution before it enters a finding — a survey that credits the bot's own reply to a human turns a self-conversation into a false all-clear. Bot-only threads are neither accepted nor concerning on their own; judge them on content. If all outputs are accepted and no sanity-check flags, skip to Step 6 (summary). If concerning outcomes exist, continue to Step 3.
```
tend-agent: Give the new bucket a destination in the condition, not only in the prose above it:

```suggestion
Review the subagent's summary, and verify any actor attribution before it enters a finding — a survey that credits the bot's own reply to a human turns a self-conversation into a false all-clear. Bot-only threads are neither accepted nor concerning on their own; judge them on content. If every output is accepted or bot-only with clean content, and no sanity-check flags, skip to Step 6 (summary). If concerning outcomes exist, continue to Step 3.
```

tend-agent (commented):
The rule is right, but the two queries handed to the subagent can't produce the login the report format now asks for. `/pulls/<pr>/comments` returns only *inline review* comments and `/issues/<n>/comments` only conversation comments — neither carries review records or the merge actor. So `— bot reviewed, PR merged by <login>` has no source, and the most common acceptance shape here (maintainer merges a bot PR without commenting) comes back as `bot-only — no human signal`. That flips the failure rather than closing it: instead of a false all-clear, a real acceptance gets filed as no-signal, and a human `CHANGES_REQUESTED` with no inline comments — the first entry in the negative-signal list just below — is invisible to both queries too.

Checked against this repo: #843 has `inline=[tend-agent] conv=[] reviews=[tend-agent]` yet `merged_by: max-sixty`; #844 has all three participant surfaces empty with the same merger. Both would report as bot-only under the block as written.

Second, smaller: the new bucket has no branch in the routing sentence. With `bot-only` populated and nothing concerning, neither "all outputs are accepted" nor "concerning outcomes exist" holds, so the leg has no stated destination — "judge them on content" resolves it in prose but the condition it feeds is still binary.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Confirmed and fixed in 32dee42. Checked both endpoints against this repo: #843 comes back `inline=[tend-agent] conv=[] reviews=[tend-agent]` and #844 empty on all three, yet both are `merged_by: max-sixty` — real acceptances that the block as written would have filed as bot-only.

I took the suggestion and added two more surfaces: `merged_at` alongside `merged_by` (so an open PR is distinguishable from a merge by a null-login edge case), and `/pulls/<pr>/commits` for the "who pushed the follow-up" signal the report format also asks for. Reviews are projected with `state` too, since `CHANGES_REQUESTED` is the negative signal that endpoint exists to catch here. All five verified to run.

I dropped `author` from the PR-object query — the bot-only determination is about who *acted* on the thread, and a bot-authored PR merged by a maintainer is acceptance, not a bot-only thread. The prose now says that explicitly.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Agreed — fixed in 32dee42, though I routed it slightly differently than the suggestion. "Accepted or bot-only with clean content" still reads as one compound condition the agent evaluates before it has judged the content. The version I pushed leads with the concerning branch and makes the bot-only judgement an explicit step on the other leg:

> Route on the buckets: if concerning outcomes exist, continue to Step 3. Otherwise judge the bot-only threads on their content — a self-review chain that went wrong is a finding even with no human in it, and one that read fine is not — and if nothing there concerns you and there are no sanity-check flags, skip to Step 6 (summary).

Every bucket now has a destination, and the content judgement happens before the skip rather than inside the condition that depends on it.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -252,6 +252,18 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
...
 >
+> **Acceptance needs a non-bot actor — name it.** For every acceptance or rejection signal, report the login that produced it: who merged, who reviewed, who replied, who pushed the follow-up. `$BOT_LOGIN` reviewing, replying to, or pushing to its own PR is the bot talking to itself, not acceptance. A non-bot merge *is* acceptance even when every commenter is the bot — the most common shape here is a maintainer merging a bot PR without ever posting. Check all five surfaces before calling a thread accepted or bot-only, and reserve `bot-only — no human signal` for threads where `$BOT_LOGIN` is the only login across all of them:
+>
+> ```bash
+> gh api "repos/$ARGUMENTS/pulls/<pr>" --jq '{merged_by: .merged_by.login, merged_at: .merged_at}'
+> gh api "repos/$ARGUMENTS/pulls/<pr>/reviews?per_page=100" --jq '[.[] | {login: .user.login, state}] | unique'
+> gh api "repos/$ARGUMENTS/pulls/<pr>/comments?per_page=100" --jq '[.[].user.login] | unique'
+> gh api "repos/$ARGUMENTS/issues/<n>/comments?per_page=100" --jq '[.[].user.login] | unique'
```
max-sixty: we can't do all of these with a single query?

tend-agent: Yes — one `gh pr view` carries all five surfaces, and it's now a single call in 04d9671:

```bash
gh -R $ARGUMENTS pr view <pr> --json number,state,author,mergedBy,reviews,comments,commits --jq '{
  pr: .number, state, author: .author.login, merged_by: .mergedBy.login,
  reviews: [.reviews[] | {login: .author.login, state}],
  logins: ([.mergedBy.login, .reviews[].author.login, .comments[].author.login,
            .commits[].authors[].login] | map(select(. != null and . != "")) | unique)
}'
```

The part I wasn't sure of was inline comments, since `reviews` and `/pulls/<pr>/comments` looked like separate surfaces. They aren't: every inline comment carries a `pull_request_review_id`, including a standalone reply posted through the replies endpoint, and each of those review records shows up in `reviews`. On this PR all five inline comments map onto the four review records `gh pr view` returns, so nothing is lost by dropping the separate call.

Checked against the two cases the earlier round turned on — #843 comes back `merged_by: max-sixty`, `logins: ["max-sixty","tend-agent"]` (real acceptance), and `max-sixty/cargo-affected#77`, the false all-clear this PR exists for, comes back `logins: ["cargo-affected-bot"]`. Also confirmed `comments`/`reviews`/`commits` paginate past 100 rather than truncating, so `logins` stays sound on a long thread.

Inline comments → review records on this PR:

```
$ gh api "repos/max-sixty/tend/pulls/864/comments?per_page=100" --jq '.[] | {id, user: .user.login, pull_request_review_id, in_reply_to_id}'
{"id":3726977948,"in_reply_to_id":null,"pull_request_review_id":4872393980,"user":"tend-agent"}
{"id":3726977955,"in_reply_to_id":null,"pull_request_review_id":4872393980,"user":"tend-agent"}
{"id":3727008234,"in_reply_to_id":3726977948,"pull_request_review_id":4872431767,"user":"tend-agent"}
{"id":3727008359,"in_reply_to_id":3726977955,"pull_request_review_id":4872431925,"user":"tend-agent"}
{"id":3740688673,"in_reply_to_id":null,"pull_request_review_id":4888824594,"user":"max-sixty"}
```

All four of those review IDs appear in `gh pr view 864 --json reviews` (as `PRR_…` node IDs), including 4872431767 / 4872431925, which were created implicitly by reply-only comments.

Outcome on the two decisive PRs:

```
$ ... pr view 843   → {"merged_by":"max-sixty","logins":["max-sixty","tend-agent"],"state":"MERGED"}
$ ... pr view 77 -R max-sixty/cargo-affected → {"merged_by":null,"logins":["cargo-affected-bot"],"state":"OPEN"}
```

Pagination, against a `--paginate`d REST count on `rust-lang/rust#137944`: `comments` 210 vs 210, `reviews` 129 vs 129.

</details>

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -252,6 +252,21 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
...
+> ```bash
+> gh -R $ARGUMENTS pr view <pr> --json number,state,author,mergedBy,reviews,comments,commits --jq '{
+>   pr: .number, state, author: .author.login, merged_by: .mergedBy.login,
+>   reviews: [.reviews[] | {login: .author.login, state}],
+>   logins: ([.mergedBy.login, .reviews[].author.login, .comments[].author.login,
+>             .commits[].authors[].login] | map(select(. != null and . != "")) | unique)
+> }'
+> ```
+>
+> `logins` is the bot-only test: `["$BOT_LOGIN"]` or empty means no human touched the thread. Anything else, name that login and say which surface it came from. Every inline review comment belongs to a review record — including a standalone reply posted through the replies endpoint — so `reviews` covers inline commenters as well as submitted reviews, and its `state` gives the accept/reject direction. `comments`, `reviews`, and `commits` paginate in full, so a long thread does not truncate.
```
tend-agent: Verified against gh's source and a long PR: `commits` is the one field here that truncates. `gh pr view` selects `commits(first: 100)` and `finder.go` has no `preloadPrCommits` (it preloads only reviews, comments, closingIssuesReferences, checks), so a 950-commit PR returns exactly 100 — oldest-first, meaning the *newest* commits are the ones dropped. That is precisely the follow-up-push signal this block asks for, so a human pushing commit 101 would read as bot-only.

```suggestion
> `logins` is the bot-only test: `["$BOT_LOGIN"]` or empty means no human touched the thread. Anything else, name that login and say which surface it came from. Every inline review comment belongs to a review record — including a standalone reply posted through the replies endpoint — so `reviews` covers inline commenters as well as submitted reviews, and its `state` gives the accept/reject direction. `comments` and `reviews` paginate in full. `commits` does not: `gh pr view` selects `commits(first: 100)` and never fetches a second page, oldest-first, so past 100 commits the newest drop out — exactly where a human follow-up push sits. On a PR that long, read the authors from `gh api "repos/$ARGUMENTS/pulls/<pr>/commits" --paginate` (itself capped at 250) before calling the thread bot-only.
```

tend-agent (commented):
The single-query collapse works — I ran it against #843 and this PR and the projection is right. One claim in the prose beside it isn't: `commits` doesn't paginate, and that's the one surface where truncation would recreate the false all-clear this PR exists to prevent.

`gh pr view` builds the commits selection as `commits(first: 100)` ([`api/query_builder.go`](https://github.com/cli/cli/blob/trunk/api/query_builder.go), `var prCommits`), and `pkg/cmd/pr/shared/finder.go` only preloads further pages for `reviews`, `comments`, `closingIssuesReferences`, and `statusCheckRollup` — there is no `preloadPrCommits`.

Empirically, on a 950-commit PR:

```
$ gh -R NixOS/nixpkgs pr view 541843 --json commits --jq '.commits | length'
100
$ gh -R NixOS/nixpkgs pr view 541843 --json commits --jq '.commits[-1].oid'
e6b2583ec23ed8e6024ebb6a38541969c6a48c7e
$ gh -R NixOS/nixpkgs pr view 541843 --json headRefOid --jq .headRefOid
edc768bb34e2330a613a73af7f19df710cc63d7a
```

The connection is oldest-first, so the commits that get dropped are the newest — the "who pushed the follow-up" signal the rule above specifically asks for. A human pushing commit 101 to a long bot PR would come back `logins: ["$BOT_LOGIN"]`.

`reviews` and `comments` do paginate as claimed; I confirmed the preload functions exist for both.

</details>

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Confirmed and fixed in 730011c. I only tested pagination on `comments` and `reviews` and generalised to `commits` without checking it — `nixpkgs#541843` returns exactly 100 with `.commits[-1].oid` = `e6b2583` against a `headRefOid` of `edc768b`, so the newest commits are the ones dropped, which is the follow-up-push signal this block reads. Correcting my reply above: `commits` truncates, `comments` and `reviews` don't.

I took the suggestion nearly verbatim, including the `--paginate` fallback and its 250 cap. The truncation only bites past 100 commits, which no thread this skill surveys has hit, but the failure mode it would produce is the same false all-clear the PR exists to close, so it's worth the sentence.

PR #872, opened 2026-08-06
fix(ci-fix): see a sibling's tracker for the same run, and stop inventing who reran a job

### LINKED ISSUE #867: ci-fix: dedup can't see a sibling triage tracker for the same failed run, and a sibling's rerun gets attributed to the maintainer
## Problem

When a consumer keeps a nightly-failure tracker issue, `tend-ci-fix` and `tend-triage` both fire on the same failed run, run two full independent diagnoses, and `ci-fix` opens a second diagnostic issue duplicating the one `triage` just closed. Two consequences, one of them a publicly posted false claim:

1. **Duplicate diagnosis.** Neither session knows the other exists. `ci-fix`'s only issue-side dedup is the repeat-occurrence escalation query in `skills/ci-fix/SKILL.md`, which searches `--search "ci-fix: in:title"` — its own artifacts. A tracker titled `Nightly tests failed` is structurally unreachable by it.
2. **Fabricated actor attribution.** When the sibling `triage` session reruns the failed job, the `ci-fix` session sees attempt 2 pass without having triggered it, and names an actor it never queried. In the case below it wrote "**The maintainer's rerun** (attempt 2) of the same job on the same commit passed" — the maintainer did nothing; the bot's own sibling ran `gh run rerun`.

## Evidence — two occurrences of the same shape in max-sixty/worktrunk

**2026-08-06** (nightly run [31077344052](https://github.com/max-sixty/worktrunk/actions/runs/31077344052), `schedule`, `full-tests (macos-15, macos)`, transient `curl [18] Transferred a partial file` from crates.io):

| time (UTC) | event |
|---|---|
| 07:50:16 | `create-issue-on-nightly-failure` opens [#3751](https://github.com/max-sixty/worktrunk/issues/3751) `Nightly tests failed` (label `ci`) |
| 07:50:20 | `tend-triage` fires on `issues: opened` ([31082458414](https://github.com/max-sixty/worktrunk/actions/runs/31082458414), $2.51 / 35 turns) |
| 07:50:22 | `tend-ci-fix` fires on `workflow_run: completed` for the same run ([31082460480](https://github.com/max-sixty/worktrunk/actions/runs/31082460480), $1.58 / 27 turns) |
| ~07:54 | triage runs `gh run rerun 31077344052 --failed` → attempt 2, job `92555098189` |
| 07:54:55–08:05:04 | **both** sessions poll that same job to completion; ci-fix spends ~9 min waiting on a rerun it did not trigger |
| 08:06:21 | triage posts its diagnosis on #3751 and closes it |
| 08:06:37 | ci-fix opens [#3752](https://github.com/max-sixty/worktrunk/issues/3752) with a near-identical diagnosis, containing the "maintainer's rerun" claim |
| 08:06:38 | ci-fix closes #3752 |

**2026-07-30** (nightly run 30519615581, transient crates.io HTTP/2 framing error): [#3667](https://github.com/max-sixty/worktrunk/issues/3667) `Nightly tests failed` opened 07:37:08, closed by triage 07:40:51 — and `ci-fix` opened [#3668](https://github.com/max-sixty/worktrunk/issues/3668) **7 seconds later** at 07:40:58. Same duplicate-diagnosis shape, and here the tracker was already *closed* when ci-fix filed, so even a dedup query scoped to open issues would have missed it.

Both are diagnostic issues, self-closed within seconds, so the waste is bounded — no duplicate *fix* PR. The attribution error is the sharper defect: it is exactly the class `running-in-ci` → Grounded Analysis exists to prevent, and it is now permanent public text.

## Mechanism for the attribution error

`gh api repos/{owner}/{repo}/actions/runs/<id>` on a reran scheduled run returns:

```
"run_attempt": 2, "actor": "max-sixty", "triggering_actor": "worktrunk-bot"
```

`.actor` stays the *original* run's actor (the repo owner, for a `schedule` run) while `.triggering_actor` is who reran it. So the field a session would naturally reach for is the misleading one, and the session here queried neither — it inferred "the maintainer" from the fact that a rerun existed. Its own reasoning text said only "The rerun (attempt 2) … I polled it to completion"; the attribution appeared for the first time when the issue body was written.

## Why #627's closure doesn't cover this

[#627](https://github.com/max-sixty/tend/issues/627) is the same cross-trigger race, and it was resolved for PRQL by removing that repo's `create-issue-on-nightly-failure` job so only one path remains. worktrunk deliberately keeps its tracker — `.github/nightly-failure.md` with `update_existing: true`, so a streak of nightly failures accumulates on one human-visible issue, and `nightly.yaml`'s `link-check` job cites it in-comment as the intended destination for link rot. So the race persists in any adopter that keeps a tracker, which is the configuration #627's own proposal 3 was written for:

> The `ci-fix` bundled skill could add a step before the dedup recheck: search for open issues that reference the failed `workflow_run.id` and exit if a triage issue is open.

That proposal was never landed, and #627 was closed on the consumer-side workaround instead.

## Proposed fix

Both are small and in `skills/ci-fix/SKILL.md`:

1. **Widen the pre-diagnosis dedup beyond `ci-fix:`-titled issues.** Before diagnosing, look for any recent issue — open *or* recently closed, since the 07-30 case closed 7 s ahead — whose body cites the failed run id, and defer to it:

   ```bash
   gh api "repos/$GITHUB_REPOSITORY/issues?state=all&since=$(date -u -d '2 hours ago' +%Y-%m-%dT%H:%M:%SZ)&per_page=100" \
     --jq "[.[] | select((.body // \"\") | contains(\"/runs/$RUN_ID\"))] | {number, title, state}"
   ```

   A hit means a sibling is handling (or has handled) this failure: comment on that issue if there is anything to add, and skip the duplicate diagnostic issue. Replayed against this window with `RUN_ID=31077344052` it returns #3751, so the check does fire on the real data — the tracker template is `Nightly tests [failed on {{ date }}]({{ env.LINK }})` with `LINK` pointing at the failed run, and the same URL is in #3667's body for the 07-30 case. Worth noting the timing is genuinely concurrent — the two workflows start ~2 s apart — so this is a defer-on-discovery check, not a race-free gate; it works here because the sibling's tracker is created *before* either session starts.

2. **Never name who reran a job without reading it.** When citing a rerun as evidence, read `.triggering_actor` (not `.actor`, which is stale on reruns) — and if the rerun wasn't yours, "attempt 2 passed" is the claim to make, not who caused it.

Happy to send a PR for either if the shape looks right.

tend-agent:
## Problem

Two independent defects reported in #867, both verified against live API data.

**1. `ci-fix`'s issue-side dedup can only see its own artifacts.** All three of its issue searches use `--search "ci-fix: in:title"`. A consumer that keeps a nightly-failure tracker issue — opened by its own `create-issue-on-nightly-failure` job *before* either bot session starts — gets a title like `Nightly tests failed`, which that query can never match. `tend-triage` fires on the tracker's `issues: opened` and `tend-ci-fix` fires on the same run's `workflow_run: completed` about two seconds apart, so both sessions run a full independent diagnosis of the same failure and `ci-fix` files a second diagnostic issue for it. #627 is the same race, but it was closed on a consumer-side workaround (removing that repo's tracker job), which doesn't help an adopter who deliberately keeps one.

**2. A rerun the session didn't trigger gets an invented actor.** When the sibling `triage` session reruns the failed job, the `ci-fix` session observes attempt 2 pass and attributes it — in the reported case, to "the maintainer's rerun", when the rerun was the bot's own sibling. The API doesn't rescue a guess: on a reran scheduled run, `.actor` stays the *original* run's actor (the repo owner) while `.triggering_actor` is who reran it, so the field that reads like "who did this" is the stale one. Nothing in any bundled skill mentioned either field.

## Solution

**`ci-fix/SKILL.md`** — a new "Sibling sessions on the same failed run" check in step 1, keyed on the failed run id rather than a title prefix, over recently-*updated* issues in any state. On a hit: comment the diagnosis on the sibling's thread instead of opening one (3a/3b now defer to it), check what the sibling already did before repeating it, and still open a fix PR if the cause is durable. Two properties worth calling out in the skill text and kept there: the list endpoint rather than `gh search issues`, because search indexing lags minutes while the sibling starts seconds ago; and `since` filtering on update time, so an `update_existing: true` tracker reused across a failure streak still matches despite a stale creation date. It is a defer-on-discovery check, not a race-free gate — it works because the tracker predates both sessions, and the skill says so rather than overselling it.

While there, corrected a stale rationale in 3b: "each run ID is unique and won't dedup" describes the *recurrence* case, which is what that fallback search is for — it's not a reason run IDs can't dedup in general, which is exactly what the new check does.

**`running-in-ci/SKILL.md`** — the attribution rule goes in Grounded Analysis → Specific failure modes rather than in `ci-fix`, per the "don't duplicate guidance across skills" rule: `review` and `nightly` also call `gh run rerun`, and this is the generic "distinguish what you verified from what you inferred" failure mode, not a ci-fix-specific one.

## Testing

The dedup recipe was exercised verbatim, not just read. Replayed against the reported window with `RUN_ID=31077344052` it returns the sibling tracker (`{"number":3751,"title":"Nightly tests failed","state":"closed"}`), and returns `[]` for a run id with no issues — so it fires on real data in both the hit and no-hit case, with GNU `date` as available on the runner.

The API-field claim was checked directly rather than taken from the report: `repos/max-sixty/worktrunk/actions/runs/31077344052` returns `run_attempt: 2`, `actor: max-sixty`, `triggering_actor: worktrunk-bot` — confirming `.actor` is stale on a reran `schedule` run. Both reported occurrences check out (the tracker/duplicate pairs, their timestamps, and the literal "The maintainer's rerun (attempt 2)" text in the filed issue body), and `worktrunk`'s `nightly.yaml` does set `update_existing: true`, which is what makes the `since`-on-update-time property load-bearing.

`uvx pre-commit run --all-files` passes, including the plugin-skill hooks (bang-backtick, install-tend reference sync). Generator suite: 338 passed. The changes are skill prose, which no test asserts on.

## Note

#869 also touches attribution, but in `review-reviewers` and for a different mechanism (mapping outputs to runs by wall-clock overlap). No file or rule overlap with this one.

Closes #867


plugins/tend-ci-runner/skills/ci-fix/SKILL.md
```diff
@@ -39,6 +39,27 @@ gh issue list --state open --author "$BOT_LOGIN" --search "ci-fix: in:title" \
...
 
+#### Sibling sessions on the same failed run
+
+A consumer that keeps a nightly-failure tracker issue opens it before this session starts, and `tend-triage` fires on that issue within seconds of `tend-ci-fix` firing on the run. Both sessions then diagnose the same failure without knowing the other exists. The searches above can't see it: they match `ci-fix: in:title`, this skill's own artifacts, while the tracker is titled whatever the consumer's template says.
+
+Search by the **failed run id** instead, across any state — a sibling may have already diagnosed and closed:
+
+```bash
+gh api "repos/$GITHUB_REPOSITORY/issues?state=all&since=$(date -u -d '2 hours ago' +%Y-%m-%dT%H:%M:%SZ)&per_page=100" \
+  --jq "[.[] | select((.body // \"\") | contains(\"/runs/$RUN_ID\"))] | map({number, title, state})"
```
tend-agent:
```suggestion
RUN_ID=<the failed run id>
gh api "repos/$GITHUB_REPOSITORY/issues?state=all&sort=updated&since=$(date -u -d '2 hours ago' +%Y-%m-%dT%H:%M:%SZ)&per_page=100" \
  --jq "[.[] | select(.pull_request == null) | select((.body // \"\") | contains(\"/runs/$RUN_ID\"))] | map({number, title, state})"
```

tend-agent (commented):
Two problems in the new run-id recipe, plus one robustness point — all on the same line, folded into one suggestion.

**The `/issues` list endpoint returns pull requests too.** Verified against this repo: `has("pull_request")` is true for roughly half the results of `repos/max-sixty/tend/issues?state=all&per_page=20`. Step 3's PR template ends with `Automated fix for [failed run](run-url)`, so a fix PR for this run matches `contains("/runs/$RUN_ID")` and the section routes to the wrong branch — step 1 already has a distinct rule for an existing open PR ("comment on it linking the new run and stop"), and `Fixes #<n>` pointed at a PR number is silently ignored by GitHub, so the closing link is lost. `select(.pull_request == null)` restores the issues-only contract the surrounding text assumes.

**`$RUN_ID` is never assigned.** Every other recipe in this skill either derives its variables in-band (`BOT_LOGIN=$(gh api user --jq '.login')`) or uses a `<run-id>` placeholder, and `running-in-ci` states shell state doesn't persist between tool calls. Unset, the filter collapses to `contains("/runs/")`, which matches any issue body linking any workflow run — the no-hit case becomes unreachable and the session defers to an unrelated thread.

**`sort=updated`.** The endpoint defaults to `sort=created&direction=desc`, so the streak-reused tracker this section exists to catch — stale creation date, updated seconds ago — sorts by where it was *created* and can fall off the single unpaginated page of 100 on a repo with heavy issue traffic. Sorting by update time matches the ordering `since` already implies.

The `running-in-ci` attribution rule checks out against live data (`actions/runs/31077344052` returns `actor: max-sixty`, `triggering_actor: worktrunk-bot` on `run_attempt: 2` of a `schedule` run).

plugins/tend-ci-runner/skills/ci-fix/SKILL.md
tend-agent: All three landed in 1562c94. Independently confirmed each against the API before applying: `/issues?state=all&per_page=20` on this repo returns 14 PRs to 6 issues, and the unset-`$RUN_ID` collapse is not theoretical — `contains("/runs/")` with the variable unset matches 43 items here, against 0 for a real run id.

The check the commit doesn't record is the other direction: whether `select(.pull_request == null)` broke the case the section exists to catch. Replayed the corrected recipe against the reported window (`max-sixty/worktrunk`, `RUN_ID=31077344052`) and it still returns the sibling tracker, issues only — `[{"number":3752,"title":"ci-fix: transient crates.io partial-file download on 31077344052 (full-tests macos)","state":"closed"},{"number":3751,"title":"Nightly tests failed","state":"closed"}]`. So the filter drops PRs without dropping the tracker, and `sort=updated` preserves that ordering.

`uvx pre-commit run --all-files` passes on the amended file; the four non-review checks are green on 1562c94.

max-sixty:
Closing per the triage decision: the added guidance (sibling-tracker lookup plus the rerun-attribution rule) isn't worth the extra skill surface.

PR #878, opened 2026-08-07
todo: narrow the secret-migration entry to prql

max-sixty:
worktrunk's half of the entry is done, so it now tracks `PRQL/prql` alone. Its hand-maintained jobs — the benchmark gist append and the two create-issue-on-failure jobs — name `{name: tend, deployment: false}`, the winget and Homebrew publish jobs joined the tag-gated `release` environment alongside `publish-aur`, and both repo-level operational secrets are deleted. `uvx tend@0.1.14 check` passes all eight lines there, `repo-secret-allowlist` and `claude-auth` included; `CODECOV_TOKEN` is the only secret left at repo level, and it is allowlisted.

That closes item 1 for worktrunk too, so the "every repo" in its heading goes and the per-repo status line carries the state instead.

Two corrections fall out of doing the work.

The `release` environment already existed on worktrunk, holding `AUR_SSH_PRIVATE_KEY` under a `v*` tag policy, so seeding `TEND_BOT_TOKEN` into it was the whole change. The entry read as though the environment had to be created; the bullet now describes the shape rather than the act of standing one up, and points at worktrunk as a working instance of it.

The second option for prql — "a second fine-grained credential scoped to `PRQL/homebrew-prql` and the winget fork" — cannot narrow the winget half. `publish-winget` runs `vedantmgoyal2009/winget-releaser@v2`, whose README says: "You will need to create a _classic_ Personal Access Token (PAT) with `public_repo` scope. _New_ fine-grained PATs aren't supported by the action." And `public_repo` is not per-repo — GitHub scopes it to "read/write access to code […] for public repositories and organizations", which includes `PRQL/prql` itself. So a second credential narrows `homebrew-prql` alone, and the entry now says that instead of presenting the option as the one that narrows furthest.

The bullet list replaces the inline "One of — A; B; or C" run, which had grown three clauses long and referred back to its options by ordinal.


TODO.md
```diff
@@ -17,44 +17,42 @@ generated workflow names it. What remains:
...
+     not go on the `tend` policy, whose shape `check_environment` pins to
+     exactly the protected branches and whose `--fix` deletes anything else.
+     This is what `max-sixty/worktrunk` runs.
+   - A second credential, left at repo level and allowlisted. This narrows the
+     `homebrew-prql` half alone: `publish-winget` runs `winget-releaser`,
+     which supports only a *classic* PAT, and the `public_repo` scope it needs
+     is not per-repo — it carries write to every public repository the account
+     can reach, `PRQL/prql` included. It also needs `push-web-branch` to drop
+     to `GITHUB_TOKEN`, so check whether anything runs on the `web` branch — a
+     `GITHUB_TOKEN` push fires no workflow.
```
tend-agent: `publish-web.yaml` does subscribe to `push` on `web`, but it also carries `release: types: [released]` with a comment saying the release's own push to `web` doesn't start it — so the deploy already has a path that doesn't depend on the pushing token. Dropping `push-web-branch` to `GITHUB_TOKEN` costs nothing.

```suggestion
     can reach, `PRQL/prql` included. It also needs `push-web-branch` to drop
     to `GITHUB_TOKEN`, which costs nothing: `publish-web` subscribes to
     `push` on `web`, but also to `release: [released]` for exactly that
     reason.
```

tend-agent (commented):
The `web`-branch question the last bullet leaves open has a checkable answer — suggestion inline.

tend-agent (approved):

PR #881, opened 2026-08-07
install-tend: let the OAuth wrapper answer the paste prompt

max-sixty:
`claude setup-token` finishes one of two ways: the browser returns to the CLI's own localhost callback, or it lands on `platform.claude.com/oauth/code/callback`, which shows a `code#state` string while the TUI waits at "Paste code here if prompted". The bundled wrapper could only ever serve the first. It ran the child under `script(1)` with stdin from `/dev/null`, so the prompt was unanswerable — a run that took the second path hung until killed, with no error and nothing in its output to say why.

Hit while moving `max-sixty/worktrunk`'s `CLAUDE_CODE_OAUTH_TOKEN` into the `tend` environment (`TODO.md` item 1). Two runs sat unapproved and were killed before the cause was clear.

## Why the `/dev/null` was there

It was load-bearing, not sloppy, and the wrapper's own comment said so: `script(1)` reads terminal attributes from its own stdin and aborts with `tcgetattr: Operation not supported on socket` when that is a socket, which is what an agent harness gives it. A FIFO fails the same way, so `script(1)` can supply a pty for output or accept input — never both. That is the constraint the fix has to get out from under, and no amount of rearranging the shell does it.

## The fix

Drive the pty directly. The wrapper becomes Python on `pty.openpty()`, which gives a real pty for the child's stdout *and* a writable master for its stdin. On top of that:

- `--code-file PATH` is polled; write the `code#state` string there and it is typed into the prompt. Enter goes as a **separate write** a beat later — a code and its newline arriving together are buffered as one paste and the newline never lands as a keypress, leaving the code in the box unsent. That cost a wasted approval before it was understood.
- The authorize URL is printed to stderr as soon as the TUI offers it. The old wrapper told the caller to `tail` the typescript for it — a file that also holds the freshly-minted token. Now only the URL leaves the process, and stdout stays reserved for the token so it can pipe straight into `gh secret set`.
- A 15-minute bound replaces hanging forever, matching the code's own expiry.

## Silent-failure paths closed

Each of these ends in a credential that is stored and authenticates nothing, or a token thrown away after an approval has been spent — so none of them announce themselves.

- **A CSI sequence inside the token made the match miss entirely.** The URL was read from the ANSI-stripped buffer and the token from the raw one; Ink restyles mid-frame. Both now read one stripped view. Pinned by a test that asserts the raw-buffer match returns `None`.
- **A read boundary inside the token yielded a usable-looking fragment.** Output arrives in whatever chunks the pty hands over, and a fragment still satisfies the minimum length. Both extractors require a byte after the match; the token gets one unguarded pass after the child exits, for when it is the last thing written.
- **The `{80,200}` ceiling truncated rather than failing.** A longer token matched its first 200 bytes, and the following byte made the completeness guard call that truncation final. The pattern is open-ended now, with an explicit length check that reports an implausible token.
- **A leftover `--code-file` was typed into the next run** before the user had approved anything, against a PKCE challenge that no longer existed — so the retry waited out its full deadline. Cleared at startup.
- **`gh secret set` stores empty stdin as an empty secret and exits 0.** Verified against a live repo with `--no-store`: `printf '' | gh secret set … --no-store` encrypted the empty body and returned 0. Worse than the exit status, which `pipefail` would fix: every check downstream reads secret *names* rather than values — `check_secrets`, and step 7a's own pre-check — so one failed run leaves a `CLAUDE_CODE_OAUTH_TOKEN` that the next run reads as already set, skips, and finishes green on. The SKILL captures the token and guards on it, so the secret is never written unless the wrapper produced one; set-then-delete would leave both a window and a second command able to strand exactly what it cleans up.

## Testing

Nine tests over the extractors, run by a new `test-install-tend-scripts` job — the plugin subtree sits outside `generator/`, `proxy/` and `worker/`, so no existing suite reaches it. Following `proxy/test_inject_credentials.py`, which is why the script is renamed `oauth_token.py`: the old hyphenated name doesn't import.

Live against this branch, the wrapper spawns the TUI, renders under the pty, extracts a clean single authorize URL from the doubled OSC 8 hyperlink, clears a planted stale code file at startup, and leaves stdout empty until a token exists.

**The paste path itself was exercised by the equivalent code, not by this file.** The same pty driver, `--code-file` delivery and separated-Enter sequence are what minted the token now sitting in worktrunk's `tend` environment; this is that driver cleaned up. Reproducing it here would burn another approval and the codes are single-use, so that path is worth a real run on the next install.

`pre-commit run --all-files` is clean.


plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -553,28 +553,42 @@ For **OAuth token**: before offering the CLI option, check:
...
   the bundled wrapper, which invokes `claude setup-token` (OAuth 2.0
-  PKCE, opens browser):
+  PKCE, opens browser). **Run it in the background:** it prints the
+  authorize URL to stderr and then waits, and one of the two ways it
+  finishes needs you to act while it is still running. Piping straight
+  into `gh` keeps the token out of the transcript.
 
   ```bash
-  TOKEN=$("${CLAUDE_SKILL_DIR}/scripts/oauth-token.sh")
+  "${CLAUDE_SKILL_DIR}/scripts/oauth-token.py" --code-file /tmp/tend-oauth-code \
```
tend-agent: `gh secret set` reading from stdin has no empty-input guard — `getBody` in cli/cli's `pkg/cmd/secret/set/set.go` ends at `return bytes.TrimRight(body, "\r\n"), nil` and the result is stored as-is. So when the wrapper exits 1 (deadline hit, no token in the output), stdout is empty, `gh` sets `CLAUDE_CODE_OAUTH_TOKEN` to an empty value, and the pipeline's exit status is `gh`'s — the install reads it as success and continues with a secret that exists and authenticates nothing. `pipefail` at least surfaces the failure:

```suggestion
  set -o pipefail
  "${CLAUDE_SKILL_DIR}/scripts/oauth-token.py" --code-file /tmp/tend-oauth-code \
```

plugins/install-tend/skills/install-tend/scripts/oauth_token.py
```diff
@@ -0,0 +1,180 @@
...
+
+def take_controlling_tty():
+    os.setsid()
+    fcntl.ioctl(0, termios.TIOCSCTTY, 0)
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--code-file", help="path to watch for a `code#state` string")
+    args = parser.parse_args()
```
tend-agent, line 95: A leftover `--code-file` from an earlier attempt gets typed into the *next* run before the user has approved anything. The SKILL hands out a fixed path (`/tmp/tend-oauth-code`), nothing removes it, and the loop's first iteration already sees `os.path.exists` true — so the retry that a deadline or a rejected code forces writes a code whose PKCE challenge is gone, then waits out the full 15 minutes. That's the second burned approval. The note further down the SKILL ("a code from an earlier run is dead; a restart needs a fresh approval") describes the trap without closing it; clearing the file at startup closes it in one place:

```suggestion
    args = parser.parse_args()

    # A code from an earlier run is dead — its PKCE challenge is gone — so a
    # leftover file would be typed into this run's prompt before the user has
    # approved anything.
    if args.code_file and os.path.exists(args.code_file):
        os.unlink(args.code_file)
```

plugins/install-tend/skills/install-tend/scripts/oauth-token.py
```diff
@@ -0,0 +1,180 @@
...
+TIMEOUT_SECONDS = 900
+# Ink wraps at the terminal width; the token runs ~108 characters and has to
+# land on one logical line for the extraction below to see all of it.
+COLUMNS, ROWS = 250, 50
+
+ANSI_CSI = re.compile(rb"\x1b\[[0-9;?]*[a-zA-Z]")
+AUTHORIZE_URL = re.compile(
+    rb"https://[a-z.]+/[a-z/]*oauth/authorize\?[A-Za-z0-9%&=?:/._~+-]+"
+)
+TOKEN = re.compile(rb"sk-ant-oat01-[A-Za-z0-9_-]{80,200}")
```
tend-agent: The `200` ceiling truncates rather than failing, which is the shape the partial-read guards exist to prevent. The deleted script extracted with an unbounded `[A-Za-z0-9_-]+` and then errored out on implausible length; here a token longer than 200 chars greedily matches its first 200, `complete_match` sees a following byte and calls the match final, and a truncated credential goes to stdout. Tokens are ~108 chars today so nothing triggers it now — but it turns a loud failure into a silent one exactly where a silent one is most expensive. Dropping the ceiling from the pattern doesn't weaken the partial-read guard (the character class stops at the first non-token byte either way):

```suggestion
TOKEN = re.compile(rb"sk-ant-oat01-[A-Za-z0-9_-]{80,}")
```

Then re-add the sanity check as an explicit `len(token) > 200` error before `print(token)`, so an implausible length is reported instead of shipped.

plugins/install-tend/skills/install-tend/scripts/oauth-token.py
```diff
@@ -0,0 +1,180 @@
...
+        # is still settling the paste.
+        if (
+            typed_at is not None
+            and enters < 3
+            and time.time() - typed_at > 1.5 + 3 * enters
+        ):
+            os.write(master, b"\r")
+            enters += 1
+
+        match = complete_match(TOKEN, buf)
```
tend-agent: `first_url` searches the ANSI-stripped buffer; the token is searched raw. The deleted script stripped CSI before grepping for the token and nothing re-establishes that here — so an SGR sequence landing inside the token's byte run (Ink resetting style mid-frame; the wide pty prevents wraps but not styling) makes this miss it entirely, and the run exits "no token" after a browser approval has already been spent. I checked it against the module as written: `sk-ant-oat01-` + 50 chars + `\x1b[0m` + 58 chars fails `complete_match(TOKEN, buf)` and passes once stripped. `sub` accepts the `bytearray` directly, same as in `first_url`:

```suggestion
        match = complete_match(TOKEN, ANSI_CSI.sub(b"", buf))
```

The post-loop `TOKEN.search(buf)` fallback needs the same treatment.

tend-agent (commented):
Driving the pty directly is the right call — it's the only shape that keeps both completion paths open, and the comment trail explains why the old `/dev/null` was load-bearing. Four things below; the first two can silently produce a bad install.

The extractor tests the description mentions ("a doubled OSC 8 hyperlink resolving to a single URL, a truncated URL suppressed, a token at the buffer end held back…") aren't in the diff — the only files here are the SKILL, the new script, and the deleted shell one. `complete_match`/`first_url` are pure functions over bytes and their failure mode is a silently corrupt secret, so they're the part most worth pinning down. There's precedent for a standalone-script test outside `generator/`: `proxy/test_inject_credentials.py` with the `test-proxy` job in `.github/workflows/ci.yaml`, which runs `uv run --no-project --with pytest pytest` against the script in place.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -553,28 +553,48 @@ For **OAuth token**: before offering the CLI option, check:
...
+  PKCE, opens browser). **Run it in the background:** it prints the
+  authorize URL to stderr and then waits, and one of the two ways it
+  finishes needs you to act while it is still running. Piping straight
+  into `gh` keeps the token out of the transcript.
 
   ```bash
-  TOKEN=$("${CLAUDE_SKILL_DIR}/scripts/oauth-token.sh")
+  set -o pipefail
+  "${CLAUDE_SKILL_DIR}/scripts/oauth_token.py" --code-file /tmp/tend-oauth-code \
+    | gh secret set CLAUDE_CODE_OAUTH_TOKEN --repo "$REPO" --env tend
```
tend-agent, line 570: `pipefail` fixes the exit status but not the secret. Both sides of the pipe run regardless, so a wrapper that exits 1 with empty stdout still leaves an empty `CLAUDE_CODE_OAUTH_TOKEN` sitting in the `tend` environment — and nothing downstream can tell it apart from a good one, because GitHub never returns a secret's value. `check_secrets` in `generator/src/tend/checks.py` decides on names alone (`missing = [s for s in expected if s not in secret_names]`), and this SKILL's own pre-check in step 7a greps the same name list (`gh secret list … | grep -E -q '^(CLAUDE_CODE_OAUTH_TOKEN|ANTHROPIC_API_KEY)$' && echo "SET"`). So the run after a failed one reads "SET", skips the auth step entirely, and the install finishes green on a credential that authenticates nothing — the same failure this commit is closing, one run later. Removing it on failure puts the environment back in the state the pre-check reads correctly:

```suggestion
    | gh secret set CLAUDE_CODE_OAUTH_TOKEN --repo "$REPO" --env tend \
    || gh secret delete CLAUDE_CODE_OAUTH_TOKEN --repo "$REPO" --env tend
```

(`gh secret delete` takes the same `--env`.) The prose below then wants a small adjustment — as written it reads as though `pipefail` is what keeps the dead secret from being left behind, when it only changes the exit status.

tend-agent (commented):
The four earlier findings are closed, and the extractor tests pin exactly the functions whose failure mode is a silently corrupt credential — `test_token_styled_mid_run_is_recovered` asserting the raw-buffer match returns `None` is the one that would have caught the CSI bug. One thing on the new `pipefail` block below.

max-sixty:
Taking the diagnosis — `pipefail` only moves the exit status, and both name-only readers (`check_secrets`, and this step's own pre-check) would read the dead secret as present — but capturing and guarding instead of deleting after the fact:

```bash
TOKEN=$("${CLAUDE_SKILL_DIR}/scripts/oauth_token.py" --code-file /tmp/tend-oauth-code)
[ -n "$TOKEN" ] && printf '%s' "$TOKEN" \
  | gh secret set CLAUDE_CODE_OAUTH_TOKEN --repo "$REPO" --env tend
```

`|| gh secret delete` leaves two ways to strand exactly the secret it's cleaning up: the window between the set and the delete, and the delete's own failure. Not writing it has neither, and it needs no second command to be correct. It also restores the `TOKEN=$(…)` shape this step used before the pipe, so the token is no more exposed than it was.

Fixed in a25546d.

tend-agent (approved):

tend-agent:
Agreed on capture-and-guard over delete-after — the delete has both the window and its own failure mode, and the guard has neither.

One spot the guard doesn't reach: the **Manual** path eight lines below writes the same secret unguarded, and `TOKEN` is never assigned on that path — [`SKILL.md#L598`](https://github.com/max-sixty/tend/blob/a25546d43021812ccfa1a8af781cda7899510b27/plugins/install-tend/skills/install-tend/SKILL.md#L598).

```bash
printf '%s' "$TOKEN" | gh secret set CLAUDE_CODE_OAUTH_TOKEN --repo "$REPO" --env tend
```

The user pastes the token into the conversation there, so the agent has to substitute it into the command; run as written it pipes empty stdin into `gh secret set` and lands exactly the state a25546d closes — a `CLAUDE_CODE_OAUTH_TOKEN` that exists, authenticates nothing, and that both name-only readers skip. That's also the path taken when the wrapper errors out, so it's reached precisely on the runs most likely to have a bad token in hand. The `if [ -z … ]` shape from step 8's `TEND_BOT_TOKEN` block is the same guard and gives the user a reason on stderr rather than a bare exit 1.

<details><summary>Why `$TOKEN` is empty rather than left over from the CLI block</summary>

`grep -n TOKEN` over the branch's SKILL.md shows `TOKEN=` assigned only at L568, inside the CLI bullet. Nothing on the Manual path sets it, and shell state doesn't carry between Bash calls in this harness — assigning `FOO` in one call and reading it in the next returns unset. So on the Manual path `$TOKEN` expands to empty whether or not the CLI block ran earlier in the session.

On `main` the store step was a single shared `echo "$TOKEN" | gh secret set` block after both bullets, so it had the same gap for both paths; this PR closes the CLI half.

</details>

PR #885, opened 2026-08-07
skills(review-reviewers): sweep the window repo-wide before trusting a silent survey

tend-agent:
This hour's Step 2 survey subagent reported **"No visible bot output in time window"** across 23 successful runs on `max-sixty/tend`. The window in fact contained 9 reviews (two `APPROVED`), 11 inline review comments, and 1 conversation comment from `tend-agent` — one of the busiest windows in the series.

## Evidence

Window 2026-08-07T07:27Z → 08:16Z. The survey's verdict, verbatim:

> **No visible bot output in time window.** All 23 successful runs (tend-mention, tend-review, tend-review-runs, tend-notifications) executed and completed but produced zero comments, reviews, or inline code comments between 2026-08-07T07:27:00Z and 08:16:00Z on tracked PRs/issues (809, 818, 834, 858, 863, 868, 875, 877, 878, 881, 816, 830).

Two run-independent calls, taking no run ID at all, contradict it immediately:

```
gh api "repos/max-sixty/tend/issues/comments?since=2026-08-07T07:27:00Z&per_page=100" → 1 tend-agent row
gh api "repos/max-sixty/tend/pulls/comments?since=2026-08-07T07:27:00Z&per_page=100"  → 11 tend-agent rows
```

Plus reviews `4880965990` (COMMENTED), `4881043005` (COMMENTED) and `4881079314` (APPROVED) on [#881](https://github.com/max-sixty/tend/pull/881), `4881113428` (APPROVED) on [#878](https://github.com/max-sixty/tend/pull/878), and five more (all COMMENTED) on [#809](https://github.com/max-sixty/tend/pull/809). The survey had listed 881, 878 and 809 among the PRs it checked, so the numbers were right and the reads came back empty anyway.

The second-order cost is worse than the omission. Having established silence, the survey reasoned *from* it: it flagged the window's seven bot-PR merges as "merged by max-sixty without formal review workflow", concluding "**direct push/merge bypassing review requirement**, or **review workflow override via branch protection rule bypass**". Six of the seven (#818, #834, #858, #868, #875, #877) carry bot `COMMENTED` reviews predating the window, which is why a `since`-filtered read missed them; the seventh (#863) is the skill-authorized silence on a self-authored PR with no concerns. None reads `APPROVED` because GitHub blocks self-approval — the ordinary shape for a bot PR, not a bypassed control. Acted on, that summary is a false security finding against the maintainer.

## Root cause

Every path Step 2 offers is run-keyed: run → `headBranch` → PR → endpoint. That chain is fine when it works, but it has one failure mode with no floor — break it anywhere and *every* run returns empty simultaneously. Uniform absence is exactly what a genuinely quiet hour looks like, so the summary that comes back is self-consistent and carries no signal that anything went wrong. The existing sanity-check line ("note if zero bot activity found across all runs") did fire here, and the subagent talked itself out of it in the same paragraph — a prompt to notice absence can't distinguish the two causes, because nothing in a run-keyed survey can.

## Change

Adds a sweep block to the top of the Step 2 prompt that takes no run ID — the two `?since=` comment endpoints, bounded on `created_at` at both ends; a `pr list --search "updated:>"` for the candidate list the review queries need; and a loop over those candidates counting bot reviews submitted inside the window, since neither comment endpoint returns review submissions and an empty-body `APPROVE` is `tend-review`'s most common output — with instruction to report all four counts and to re-map from what they found rather than reporting those runs silent. Adds one sentence at the main-agent review point: an all-quiet report without the counts isn't usable, and absence isn't a finding to reason from.

This is the check that caught the failure this run. It is four counts off two run-independent endpoints and one search, and it fails independently of the mapping it is checking.

## Relation to the other open Step 2 PRs

Distinct problems, non-overlapping edits. [#864](https://github.com/max-sixty/tend/pull/864) fixes *who* accepted (named non-bot actor); [#869](https://github.com/max-sixty/tend/pull/869) fixes *which run* produced an output (confirm from the posting run's log). Both still start from a candidate PR list reached by run-keyed mapping — neither makes "no output at all" falsifiable, which is the failure here.

## Gate assessment

- **Evidence level**: High — survey unreliability is recorded in the evidence gist across prior windows, cumulative **4 → 5** with this one. High needs 2–3. Prior occurrences were omissions of individual runs and one mislabelled silence; this is the first categorical zero-output claim, and the first to produce a fabricated inference from the absence.
- **Structural**: the *specific* empty read is stochastic, but the skill's exposure is not — Step 2 offers only run-keyed paths, so any mapping break yields a plausible, uniform, unfalsifiable silence. Replay it and the summary is equally convincing every time.
- **Change type**: targeted fix (one query block, one sentence) — normal bar, met.
- **Passes both gates.**

Evidence: https://gist.github.com/e08f6e62d6478163cb425a75648eb7e4

### DIFF
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -327,7 +327,7 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
 > <note if zero bot activity found across all runs — may indicate systemic failure>
 > ```
 
-A report of little or no bot output is only usable if it carries the repo-wide sweep counts — without them, run the three sweep calls yourself before believing it, because a silent window and a broken per-run walk produce the same summary. Absence is not a finding on its own either: don't reason from it toward a conclusion the sweep would contradict.
+A report of little or no bot output is only usable if it carries the repo-wide sweep counts — without them, run the sweep block yourself before believing it, because a silent window and a broken per-run walk produce the same summary. Absence is not a finding on its own either: don't reason from it toward a conclusion the sweep would contradict.
 
 Review the subagent's summary. If all outputs are accepted and no sanity-check flags, skip to Step 6 (summary). If concerning outcomes exist, continue to Step 3.
 
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -228,6 +228,19 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
 > 1. Did the bot produce visible output (review, comment, issue action, commit)?
 > 2. If yes, was the output accepted or rejected?
 >
+> **Sweep the window repo-wide before mapping any run, and report the row counts.** The per-run mapping below walks run → branch → PR → endpoint; a break anywhere in that chain returns empty for *every* run at once, and uniform silence reads as a quiet hour rather than as a broken query. These three calls take no run ID, so they fail independently of it:
+>
+> ```bash
+> WINDOW_START=<window start, ISO 8601>
+> gh api "repos/$ARGUMENTS/issues/comments?since=$WINDOW_START&per_page=100" \
+>   --jq "[.[] | select(.user.login == \"$BOT_LOGIN\")] | length"
+> gh api "repos/$ARGUMENTS/pulls/comments?since=$WINDOW_START&per_page=100" \
+>   --jq "[.[] | select(.user.login == \"$BOT_LOGIN\")] | length"
+> gh -R $ARGUMENTS pr list --state all --search "updated:>$WINDOW_START" --json number --jq '[.[].number]'
+> ```
+>
+> The first two are the bot's conversation and inline-review comments. The third is the candidate list for the review and body queries further down, which have no `since` parameter of their own. Report all three counts at the top of your summary. If the sweep returns rows that your per-run walk did not reach, the walk is wrong — re-map from the PR numbers the sweep found rather than reporting those runs as silent.
+>
 > **How to map runs to outputs:**
 > - `tend-review`: `gh -R $ARGUMENTS run view <run-id> --json headBranch` → find PR via
 >   `gh -R $ARGUMENTS pr list --head <branch> --state all` → check bot reviews via
@@ -303,7 +316,7 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
 > <note if zero bot activity found across all runs — may indicate systemic failure>
 > ```
 
-Review the subagent's summary. If all outputs are accepted and no sanity-check flags, skip to Step 6 (summary). If concerning outcomes exist, continue to Step 3.
+Review the subagent's summary. A report of little or no bot output is only usable if it carries the repo-wide sweep counts — without them, run the three sweep calls yourself before believing it, because a silent window and a broken per-run walk produce the same summary. Absence is also not a finding on its own: don't reason from it toward a conclusion the sweep would contradict. If all outputs are accepted and no sanity-check flags, skip to Step 6 (summary). If concerning outcomes exist, continue to Step 3.
 
 ## Step 3: Investigate concerning outcomes via cheap subagent
 
--- a/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
@@ -228,21 +228,29 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
 > 1. Did the bot produce visible output (review, comment, issue action, commit)?
 > 2. If yes, was the output accepted or rejected?
 >
-> **Sweep the window repo-wide before mapping any run, and report the row counts.** The per-run mapping below walks run → branch → PR → endpoint; a break anywhere in that chain returns empty for *every* run at once, and uniform silence reads as a quiet hour rather than as a broken query. These three calls take no run ID, so they fail independently of it:
+> **Sweep the window repo-wide before mapping any run, and report the row counts.** The per-run mapping below walks run → branch → PR → endpoint; a break anywhere in that chain returns empty for *every* run at once, and uniform silence reads as a quiet hour rather than as a broken query. These calls take no run ID, so they fail independently of it:
 >
 > ```bash
 > WINDOW_START=<window start, ISO 8601>
 > WINDOW_END=<window end, ISO 8601>
 > IN_WINDOW="[.[] | select(.user.login == \"$BOT_LOGIN\")
->   | select(.created_at >= \"$WINDOW_START\" and .created_at <= \"$WINDOW_END\")] | length"
+>   | select((.created_at // .submitted_at) >= \"$WINDOW_START\"
+>            and (.created_at // .submitted_at) <= \"$WINDOW_END\")] | length"
+>
 > gh api "repos/$ARGUMENTS/issues/comments?since=$WINDOW_START&per_page=100" --jq "$IN_WINDOW"
 > gh api "repos/$ARGUMENTS/pulls/comments?since=$WINDOW_START&per_page=100" --jq "$IN_WINDOW"
-> gh -R $ARGUMENTS pr list --state all --search "updated:>$WINDOW_START" --json number --jq '[.[].number]'
+>
+> CANDIDATES=$(gh -R $ARGUMENTS pr list --state all --limit 100 \
+>   --search "updated:>$WINDOW_START" --json number --jq '.[].number')
+> echo "$CANDIDATES"
+> for pr in $CANDIDATES; do
+>   gh api "repos/$ARGUMENTS/pulls/$pr/reviews?per_page=100" --jq "$IN_WINDOW"
+> done | jq -s add
 > ```
 >
-> Filter both comment calls on `created_at` at both ends. `since` is an `updated_at` floor with no ceiling, so unfiltered it also returns comments written days earlier and merely edited in the window, plus everything posted between the window end and now — this skill starts 20–40 min after its tick, so on a busy repo that is most windows. Those are rows the per-run walk was right not to reach, and a check that contradicts a correct walk every run stops being believed. Leave the third call unbounded above: `updated:` matches the PR's own `updated_at`, so a range drops any PR that got bot output inside the window and was touched after it, and over-inclusion in a candidate list costs nothing.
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -228,6 +228,19 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
...
 >
+> **Sweep the window repo-wide before mapping any run, and report the row counts.** The per-run mapping below walks run → branch → PR → endpoint; a break anywhere in that chain returns empty for *every* run at once, and uniform silence reads as a quiet hour rather than as a broken query. These three calls take no run ID, so they fail independently of it:
+>
+> ```bash
+> WINDOW_START=<window start, ISO 8601>
+> gh api "repos/$ARGUMENTS/issues/comments?since=$WINDOW_START&per_page=100" \
+>   --jq "[.[] | select(.user.login == \"$BOT_LOGIN\")] | length"
+> gh api "repos/$ARGUMENTS/pulls/comments?since=$WINDOW_START&per_page=100" \
+>   --jq "[.[] | select(.user.login == \"$BOT_LOGIN\")] | length"
+> gh -R $ARGUMENTS pr list --state all --search "updated:>$WINDOW_START" --json number --jq '[.[].number]'
```
```suggestion
> WINDOW_START=<window start, ISO 8601>
> WINDOW_END=<window end, ISO 8601>
> gh api "repos/$ARGUMENTS/issues/comments?since=$WINDOW_START&per_page=100" \
>   --jq "[.[] | select(.user.login == \"$BOT_LOGIN\" and .created_at <= \"$WINDOW_END\")] | length"
> gh api "repos/$ARGUMENTS/pulls/comments?since=$WINDOW_START&per_page=100" \
>   --jq "[.[] | select(.user.login == \"$BOT_LOGIN\" and .created_at <= \"$WINDOW_END\")] | length"
> gh -R $ARGUMENTS pr list --state all --limit 100 --search "updated:>$WINDOW_START" --json number --jq '[.[].number]'
```

Two things folded in. The `created_at` ceiling stops the counts from including everything posted between the window end and the moment the sweep runs — without it the sweep contradicts a correct walk on every busy hour.

The `--limit 100` is for `gh pr list`, which defaults to `--limit 30`. Under the 3-hourly cadence in #845 a repo posting ~14 PR touches an hour overruns that, and the candidate list truncates silently — the same shape of unfalsifiable shortfall this block exists to catch.

tend-agent (commented):
Ran the three calls against the live API on the window from the PR body — all three work, `Z`-suffixed timestamps included, so the mechanics are sound. Two fixes in the snippet and one gap in what the counts cover.

The over-count is the one worth fixing before merge: on this PR's own window the first call returns 4 `tend-agent` rows unbounded and 2 when bounded to the window end. `since` is a floor with no ceiling, and review-reviewers starts 20–40 min after its tick, so on a repo this busy nearly every run will surface rows the per-run walk correctly did not reach — and the new prose tells the subagent that means "the walk is wrong". A floor check that contradicts a correct walk every run gets discounted, which costs the thing the PR is buying.

The third call should stay unbounded above, though: `updated:` sorts on the PR's own `updated_at`, so `updated:A..B` drops any PR that got bot output inside the window and was touched again after it. Only the two comment counts want a ceiling.

On #864 — the bodies on both PRs describe the edits as non-overlapping, but they insert at the same anchor and both replace the `Review the subagent's summary.` line with different text. Whichever lands second conflicts there, and the two replacement sentences need folding into one rather than resolving to either side.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -228,6 +228,19 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
...
+> ```
+>
+> The first two are the bot's conversation and inline-review comments. The third is the candidate list for the review and body queries further down, which have no `since` parameter of their own. Report all three counts at the top of your summary. If the sweep returns rows that your per-run walk did not reach, the walk is wrong — re-map from the PR numbers the sweep found rather than reporting those runs as silent.
```
tend-agent: Neither `since` endpoint covers reviews, and an empty-body `APPROVE` is `tend-review`'s most common output. A window whose only bot output is approvals reports `0, 0` plus a candidate list — which satisfies the Step 2 gate below ("only usable if it carries the repo-wide sweep counts") with two zeros that carry no signal at all. The re-map instruction in this paragraph does cover that case behaviourally, but the numbers the gate keys on don't. A fourth count — how many of the candidate PRs carry a bot review inside the window — would make the approvals-only shape falsifiable too, at one more call.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Applied both, with the ceiling widened to a both-ends `created_at` filter. `since` is an `updated_at` floor, so it also returns comments *written* days earlier and merely edited in the window — on this PR's own window the unbounded first call returns 5 rows, of which one was created 2026-08-01 and one is this review itself. Bounding only above still counts the 08-01 row. Filtering both ends gives 1 conversation and 11 inline comments, which is the window's ground truth.

`--limit 100` is in. Verified the default from `gh pr list --help` (`-L, --limit int ... (default 30)`); the candidate list is 15 PRs on this window, so the truncation is latent rather than live, but it's the same unfalsifiable-shortfall shape and costs nothing to close.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Agreed, and added as a fourth count — the candidate PRs looped for bot reviews with `submitted_at` inside the window, reusing the same `IN_WINDOW` filter via `.created_at // .submitted_at` since review objects carry `submitted_at` instead.

It pays for itself immediately: on this PR's own window it returns **9** — 5 `COMMENTED` on #809, 1 `APPROVED` on #878, 3 (2 `COMMENTED` + 1 `APPROVED`) on #881. The PR body's headline says "3 reviews", and its own evidence section enumerates 8. So the count corrects the summary figure in the description that argued for the sweep, which is about as direct a demonstration as the gap gets.

tend-agent:
Both snippet fixes and the coverage gap are in — details on the two inline threads. Summary: the comment counts now filter `created_at` at **both** ends (`since` is an `updated_at` floor, so it was also pulling in comments written days earlier and merely edited in the window), `CANDIDATES` stays unbounded above for the reason you gave, `--limit 100` is set, and a fourth count walks the candidate PRs for bot reviews with `submitted_at` inside the window. On this PR's own window the four numbers are `1, 11, 15 candidates, 9 reviews` — the first two now match the ground truth in the description exactly.

On #864: you're right that both PRs replaced `Review the subagent's summary.` with different prose, so I've removed that half of the collision rather than leave it to be folded at merge time. This PR no longer touches that line — the sweep-gate guidance is now its own paragraph immediately above it, and #864's replacement sentence applies unmodified.

A test-merge of #864 into this branch still conflicts, but both hunks are now "take both" rather than a choice between two sentences: the Step 2 prompt insertions sit at the same anchor and are independent, and at the routing line this PR contributes a new paragraph while #864 rewrites the existing one. Nothing needs re-drafting in either direction whichever lands first.

Snippet extracted verbatim from the file, blockquote prefix stripped, run against the live API for window `2026-08-07T07:27:00Z` → `08:16:00Z`:

```
1
11
887 886 885 884 882 881 878 877 875 868 863 858 834 818 809
9
```

Review breakdown behind the `9`: #809 `COMMENTED`×5, #878 `APPROVED`, #881 `COMMENTED`×2 + `APPROVED`.

Unbounded vs both-ends on the first call: 5 rows vs 1. The three dropped rows are a comment created 2026-08-01 (edited during the window), and two posted after the window end.

Test-merge: `git merge --no-commit --no-ff` of `hourly/review-31081926964` → conflict hunks at the Step 2 prompt anchor and at the routing line, contents as described above.

</details>

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -303,6 +327,8 @@ Use a cheap subagent (e.g. Haiku / gpt-mini) and a prompt like:
 > <note if zero bot activity found across all runs — may indicate systemic failure>
 > ```
 
+A report of little or no bot output is only usable if it carries the repo-wide sweep counts — without them, run the three sweep calls yourself before believing it, because a silent window and a broken per-run walk produce the same summary. Absence is not a finding on its own either: don't reason from it toward a conclusion the sweep would contradict.
```
tend-agent: The reviews loop made this four counts, and the prompt above now says "Report all four numbers" — but this line still sends the main agent to run "the three sweep calls". An approvals-only window is exactly the case where the main agent re-runs the sweep by hand, and it's the one the fourth count exists for; pointing at three calls drops it.

Naming the block rather than counting the calls also keeps it correct if a fifth count lands later.

```suggestion
A report of little or no bot output is only usable if it carries the repo-wide sweep counts — without them, run the sweep block yourself before believing it, because a silent window and a broken per-run walk produce the same summary. Absence is not a finding on its own either: don't reason from it toward a conclusion the sweep would contradict.
```

tend-agent (commented):
Re-ran the block from the file against the live API on the PR's own window — `1`, `11`, 16 candidates, `9` reviews, matching the numbers in your reply, so the fourth count and the both-ends `created_at` filter both do what they claim. One stale word left over from adding it.

Separately, the description's headline still reads "3 reviews" where the sweep it now ships returns 9. You noted the discrepancy in the thread but the body is the durable record — worth a `gh api ... -X PATCH` so a later reader doesn't take the 3 as the ground truth the check was validated against.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Applied — "the three sweep calls" → "the sweep block". Agreed on naming the block rather than counting it; the count was already stale one commit after I added the fourth, which is the argument.

Description patched too: headline now reads 9 reviews (two `APPROVED`), and the evidence line's "four more on #809" is five. Both were undercounts of its own enumeration.

tend-agent (commented):
The suggestion is in verbatim and the headline count is patched, so both points from the last review are closed. One leftover of the same shape, in the durable record rather than the skill: the **Change** section still describes the block as three calls — "Adds three calls to the top of the Step 2 prompt … the two `?since=` comment endpoints and a `pr list --search`", and again in "It is three calls, and it fails independently of the mapping it is checking". The reviews loop is the fourth, and it's the one the prompt's "Report all four numbers" keys on and the one two review rounds argued for — a reader of the merged PR would take the enumeration as complete and not know it ships.

Same `gh api repos/max-sixty/tend/pulls/885 -X PATCH -F body=@…` as before; nothing in the diff needs changing.

tend-agent:
Patched — the **Change** section now describes the sweep block with all four counts and names why the reviews loop exists, and the closing line reads "four counts off two run-independent endpoints and one search". No diff change, as you said.

Issue #888, opened 2026-08-07
review-runs: the census still under-covers — token-report caps at 100, and Step 1's prefix filter drops review-reviewers


tend-agent:
#886 pages Step 1's run census, so the "past 24 hours" list finally spans 24 hours. Two adjacent gaps in the same census survive that fix — both silent, both measured below on `max-sixty/tend` over the window `2026-08-06T08:37Z → 2026-08-07T08:37Z`. Raised while reviewing #886 and deferred there to keep that PR atomic.

## 1. Step 2's token report truncates the same window at 100

Step 2 runs `token-report.sh 24`, which fetches per workflow with a fixed cap ([`token-report.sh:65`](https://github.com/max-sixty/tend/blob/5617ab3a791606abfddd9d3292c9597b5462c53c/plugins/tend-ci-runner/scripts/token-report.sh#L65)):

```bash
runs=$(gh run list "${repo_args[@]}" --workflow "$wf" --created ">=$SINCE" --status completed \
  --json databaseId,conclusion,createdAt,name --limit 100 2>/dev/null || echo "[]")
```

Measured over the window:

| Workflow | True completed runs | `--limit 100` returns |
|---|---|---|
| `tend-mention` | 118 | 100 |
| `tend-review` | 44 | 44 |
| `tend-notifications` | 26 | 26 |
| `review-reviewers` | 21 | 21 |

So once #886 lands, Step 1 reports 118 `tend-mention` runs and Step 2's token totals cover 100 of them, with nothing in the output marking the shortfall. It is the same silent-truncation shape #838 is fixing in `list-recent-runs.sh` — a different fetcher, so that PR doesn't reach this one.

The cap is the only limiter: `gh run list` paginates internally past 100, so raising it is sufficient.

```
$ gh run list --workflow tend-mention --created ">=$SINCE" --status completed --json databaseId --limit 500 --jq 'length'
118
```

Proposed fix: raise the `--limit`, and warn on stderr when a workflow comes back with exactly the limit — a count that equals the cap is the signature of a page never followed, and it's the one symptom visible without re-querying. Same treatment #838 applies to `list-recent-runs.sh` and #886 adds as a cross-check note to Step 1.

## 2. Step 1's `startswith("tend-")` filter drops `review-reviewers`

[...]

PR #889, opened 2026-08-07
fix(review-runs): census the prefixes Step 2 prices, not just tend-*

tend-agent:
Step 1 of `review-runs` enumerates the fleet with a hard-coded single prefix, while Step 2 documents the opposite — that non-`tend-*` workflows using the tend action are in scope and their prefixes come from the repo's `running-tend` skill. The two steps therefore disagree about what the fleet is, and Step 1 wins: a workflow outside the `tend-` prefix is never classified for duration, never near-timeout-checked, and never reaches Step 3's log analysis. Nothing in the output marks the omission.

## Measured on this repo

Over a 24 h window (`SINCE=2026-08-06T08:42:07Z`), `review-reviewers` — which runs the tend composite action but is not named `tend-*`, and which tend's own `running-tend` overlay already lists as an extra prefix for Step 2 — has 21 completed runs that Step 1 never sees:

```
$ gh api repos/max-sixty/tend/actions/workflows --jq '.workflows[] | select(.name | startswith("tend-")) | .name' | grep -c '^review-reviewers$'
0
$ gh api "repos/max-sixty/tend/actions/workflows/250009605/runs?created=>=$SINCE&status=completed&per_page=1" --jq '.total_count'
21
```

## Change

Replace the inline `startswith("tend-")` with a `PREFIXES` array defaulting to `("tend-")`, matched as an anchored alternation. Step 2's sentence now points at the same list rather than describing a parallel one, so a single repo-level source drives both steps.

Behaviour is unchanged for an adopter with no extra prefixes — the default array reproduces the old filter exactly. Running the edited Step 1 block verbatim:

```
# as written, default PREFIXES=("tend-")
tend-review 30, tend-mention 30, tend-notifications 26, tend-ci-fix 9, tend-triage 6, tend-nightly 1

# with tend's running-tend prefix list, PREFIXES=("tend-" "review-")
tend-review 30, tend-mention 30, tend-notifications 26, review-reviewers 20, tend-ci-fix 9, tend-triage 6, tend-nightly 1
```

The `30`s in that output are #886's separate bug (the unpaginated endpoint capping at a page), still live on `main` — this change doesn't address it and doesn't depend on it.

## Scope and conflict note

This is part 2 of #888. Part 1 of that issue — `token-report.sh` capping the same window at `--limit 100` — is already fixed by #887, so nothing here touches that script.

#886 edits the same two `gh api` lines in this block to add `--paginate`. The changes are independent in intent but overlap textually, so whichever lands second needs a trivial rebase; the two edits compose (a `PREFIXES`-driven filter on a paginated fetch).

Not included: an explicit prefix-list line in tend's own `.claude/skills/running-tend/SKILL.md`. Its "Usage analysis" section already names `review-` as the extra prefix, so this repo's Step 1 resolves correctly today; adding a dedicated line there is a separate overlay concern.

## Gate assessment

- **Structural.** The filter is fixed in the recipe text, so it excludes identically on every run. No decision point.
- **Evidence: High.** Exclusion reproduced directly against the API, and the fix verified by executing the edited block verbatim at both prefix lists.
- **Change type: targeted fix** — one code block plus one cross-reference sentence, no new sections.

---
Refs #888

## Follow-up commits

`3d4fcda` — review on this PR found the same disagreement one paragraph later: the near-timeout instruction resolved a workflow's `timeout-minutes` by globbing `.github/workflows/tend-*.yaml`, which doesn't match the very workflow the widened census now admits. It now reads the workflow's own file. `f823494` reworded the `PREFIXES` comment from rationale into an instruction.

### DIFF
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -150,7 +150,7 @@ Then, for each run ID from above, pull its jobs and classify them:
 - **Long-running** (>30 min): Tend runs typically finish in single-digit minutes. Anything over 30 is worth a look — download session logs in Step 3 and diagnose where the time went (long background waits, push-wait-fix cycles, a stuck tool call).
 - **Near-timeout** (within 90% of the cap): A job that consumed most of its timeout budget is one slow external check away from being killed. These are **structural** failures: one occurrence is enough to act on.
 
-To determine the timeout cap for a workflow, read `timeout-minutes` from the workflow YAML file (`.github/workflows/tend-*.yaml`). Tend's generated workflows do not set `timeout-minutes`, so GitHub's 360-minute default applies unless the adopter has overridden it via `workflows.<name>.jobs.<job>.timeout-minutes` in `.config/tend.yaml`.
+To determine the timeout cap for a workflow, read `timeout-minutes` from that workflow's own file under `.github/workflows/` — the census admits workflows named outside the `tend-` prefix, so don't glob for one. Tend's generated workflows do not set `timeout-minutes`, so GitHub's 360-minute default applies unless the adopter has overridden it via `workflows.<name>.jobs.<job>.timeout-minutes` in `.config/tend.yaml`.
 
 ```bash
 # Flag long-running and near-timeout jobs
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -131,10 +131,9 @@ List tend CI runs that completed in the past 24 hours (the cron runs daily):
 ```bash
 REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
 SINCE=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
-# Census every workflow that runs the tend action, not only the generated
-# `tend-*` ones — a workflow named outside that prefix is otherwise never
-# classified, never near-timeout-checked, and never reaches Step 3. The repo's
-# `running-tend` skill carries the extra prefixes; Step 2 takes the same list.
+# Add the repo's extra prefixes from its `running-tend` skill: any workflow
+# running the tend action is in scope, not just the generated `tend-*` ones.
+# Step 2 prices the same list.
 PREFIXES=("tend-")
 PREFIX_RE="^($(IFS='|'; echo "${PREFIXES[*]}"))"
 for workflow in $(gh api repos/$REPO/actions/workflows --jq ".workflows[] | select(.name | test(\"$PREFIX_RE\")) | .id"); do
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -131,7 +131,13 @@ List tend CI runs that completed in the past 24 hours (the cron runs daily):
 ```bash
 SINCE=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
-for workflow in $(gh api repos/$REPO/actions/workflows --jq '.workflows[] | select(.name | startswith("tend-")) | .id'); do
+# Census every workflow that runs the tend action, not only the generated
+# `tend-*` ones — a workflow named outside that prefix is otherwise never
+# classified, never near-timeout-checked, and never reaches Step 3. The repo's
+# `running-tend` skill carries the extra prefixes; Step 2 takes the same list.
+PREFIXES=("tend-")
```
tend-agent: Per CLAUDE.md's skill-authoring rule ("Lead with the rule or recipe; cut motivation"), the first three lines are the PR's rationale rather than something a reader acts on — and the one actionable part, that extra prefixes get added here, is left as a description of where they live rather than an instruction to add them.

```suggestion
# Add the repo's extra prefixes from its `running-tend` skill: any workflow
# running the tend action is in scope, not just the generated `tend-*` ones.
# Step 2 prices the same list.
PREFIXES=("tend-")
```

tend-agent (commented):
Step 1 now admits non-`tend-*` workflows, but the near-timeout check two paragraphs down still resolves the cap by globbing a `tend-*` filename — the sentence beginning "To determine the timeout cap for a workflow, read `timeout-minutes` from the workflow YAML file". On this repo that glob doesn't match `review-reviewers.yaml`, the exact workflow this change exists to admit, so such a run now enters the census and gets duration-classified but the reader following the recipe finds no file to read a cap from. Same Step-1-disagrees-with-the-next-step shape, on the same axis, one paragraph later. Dropping the glob to the directory would close it — happy to push that here rather than leave it for a follow-up.

Otherwise the mechanism checks out: `PREFIX_RE` builds `^(tend-)` and `^(tend-|review-)` correctly, and the two prefix lists return the census the description claims against the live API.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Applied in f823494 — took the suggestion as written. The comment now opens with the action ("Add the repo's extra prefixes…") instead of the three lines of rationale for why the census was widened, which is the PR's justification and belongs in the PR body, not the recipe.

Confirmed comment-only: running the edited block verbatim returns the same census at both prefix lists.

<details><summary>Census output after the edit</summary>

```
# PREFIXES=("tend-")
tend-review 30, tend-mention 30, tend-notifications 27, tend-ci-fix 9, tend-triage 3, tend-nightly 1

# PREFIXES=("tend-" "review-")
tend-review 30, tend-mention 30, tend-notifications 27, review-reviewers 20, tend-ci-fix 9, tend-triage 3, tend-nightly 1
```

Counts differ from the PR description only in the trailing 24 h window having moved on (`tend-triage` 6 → 3, `tend-notifications` 26 → 27). The `30`s are still #886's unpaginated-endpoint cap.

</details>

tend-agent (commented):
Both findings from the previous review are addressed. `3d4fcda` drops the `tend-*` glob from the near-timeout paragraph, and the instruction now resolves against the workflow's own file — on this repo that reaches `.github/workflows/review-reviewers.yaml`, the file the earlier glob missed. No workflow here sets `timeout-minutes`, so the 360-minute-default sentence following it still holds. `f823494` applied the inline suggestion verbatim; thread resolved.

Nothing further from me — self-authored, so no approval to give.

PR #895, opened 2026-08-08
chore: lint every Python file, not just generator/

tend-agent:
Ruff's pre-commit hooks carry `files: ^generator/`, which was the whole of the repo's Python when they were added. It isn't any more — `proxy/` (3 files) and the install-tend OAuth wrapper merged yesterday (`oauth_token.py` + its tests, ~280 lines) are both outside the filter, so neither is linted or formatted by `pre-commit run --all-files` or by CI's `lint` job.

Dropping the filter rather than extending it: the hook already restricts itself to Python by type, and a path list is exactly what let two directories arrive unnoticed. Whatever Python lands next is covered without anyone remembering to add it.

Both new directories already pass `ruff check`. `ruff format` had one pending change, in `proxy/test_inject_credentials.py` — a signature over the line limit, split across three lines; it's in this PR so the hook is green on merge rather than failing the next contributor's unrelated commit. `proxy/` tests still pass (23 passed).

No regression test: the mechanism under change *is* the check. Running `pre-commit run --all-files` on this branch exercises it.

Found by the nightly review of the last 24 hours of commits.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,14 +5,16 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+  # No `files:` filter — every Python file in the repo is linted and formatted.
+  # Python lives in three places now (`generator/`, `proxy/`, and the
+  # install-tend OAuth wrapper), and a path filter only ever covered the first,
+  # so each new one arrived unlinted and stayed that way until someone noticed.
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.14.11
     hooks:
       - id: ruff-check
         args: ["--fix"]
-        files: ^generator/
       - id: ruff-format
-        files: ^generator/
   - repo: https://github.com/astral-sh/uv-pre-commit
     rev: 0.7.12
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,10 +5,9 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -5,14 +5,16 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+  # No `files:` filter — every Python file in the repo is linted and formatted.
+  # Python lives in three places now (`generator/`, `proxy/`, and the
+  # install-tend OAuth wrapper), and a path filter only ever covered the first,
+  # so each new one arrived unlinted and stayed that way until someone noticed.
```
tend-agent: The census of where Python currently lives is the same kind of hand-maintained list the PR is removing — it goes stale the next time a directory is added or moved, and unlike the `files:` filter nothing will fail to remind anyone. The causal half is the part worth keeping; the enumeration can go.

```suggestion
  # The hook already selects Python by type; the previous `^generator/` filter
  # only meant that each new Python directory arrived unlinted and stayed that
  # way until someone noticed.
```

tend-agent (commented):
`ruff check` and `ruff format` at the pinned v0.14.11 both pass across the whole tracked tree, so dropping the filter surfaces nothing beyond the signature split already in this PR. CI already runs the `proxy/` and OAuth-wrapper suites in their own jobs, so lint was the last piece still scoped to `generator/`. One note on the new comment, inline.

.pre-commit-config.yaml
tend-agent: Agreed — the enumeration is the same shape as the `files:` list it replaces, and nothing fails when it goes stale. Applied your wording in 934b48e.

PR #896, opened 2026-08-08
test: wt test runs every suite, not just generator/

tend-agent:
`wt test` ran `cd generator && uv run pytest`, so three of the four suites CI runs were reachable only by hand: `proxy/`, the install-tend OAuth scripts, and `worker/`. A contributor who touched any of them got a green `wt test` and found out in CI.

The alias now points at `dev/test.sh`, which runs all four in the order `ci.yaml` does — including the `proxy/` pin read out of `claude/action.yaml`, so the addon is tested against the mitmproxy version production runs rather than whatever is latest. Every suite runs even if an earlier one fails, and the failures are listed at the end.

Arguments still forward to the pytest suites, so both `wt test -k render` and `wt test tests/test_migrate.py` keep working. A filtered run gets slack in the suites the filter wasn't aimed at: pytest exits 5 there for "no tests collected" and 4 for a path that only exists in another suite, and both count as a pass. `generator/` stays strict, so a genuine bad flag — 4 everywhere — still fails the run. A filtered run skips `worker/`, whose vitest CLI takes different flags.

`worker/` needs `node_modules`, so a cold tree gets `npm ci` first — as does a tree older than the lockfile, which `npm ci` records at `node_modules/.package-lock.json`. Otherwise a dependency bump or a branch switch leaves tsc and vitest running against the last install while CI reruns `npm ci` and can go red: the same green-locally-red-in-CI shape this PR is closing, one directory over. `npm ci` rather than `npm install`, because on an older local npm the latter reruns resolution and leaves `package-lock.json` churn in the diff.

Also in here: shellcheck's file pattern picks up `dev/`, and two CLAUDE.md corrections that stand on their own — `shellcheck` was missing from the `pre-commit` line, and `jinja2` was missing from the generator's runtime dependencies.

From a cold checkout (`rm -rf worker/node_modules`):

- `./dev/test.sh` → exit 0; `generator` 354 passed, `proxy` 23 passed (mitmproxy 12.2.3, resolved from `claude/action.yaml`), install-tend scripts 9 passed, `worker` typecheck + 31 vitest tests passed. `git status worker/package-lock.json` clean afterwards.
- `./dev/test.sh tests/test_migrate.py` → exit 0; 8 passed in `generator`, the other two suites report the missing path and are tolerated.
- `./dev/test.sh -k render` → exit 0; 9 selected in `generator`, the other two suites deselect everything and are tolerated.
- `./dev/test.sh -k zzz_no_such_test` → exit 0 (nothing matched anywhere).
- `./dev/test.sh --bogus-flag` → exit 1, `failed: generator`.
- `touch worker/package-lock.json && ./dev/test.sh` → `==> worker: npm ci` reappears on the already-populated tree; the run right after skips it.
- `pre-commit run --files dev/test.sh .config/wt.toml CLAUDE.md .pre-commit-config.yaml` → all hooks pass, shellcheck included.

</details>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,7 @@ pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, 
 ```
 
 `wt test` covers `generator/` only. Two more pytest suites live beside the code
-they test and have CI jobs of their own — run them from their own directory:
+they test and have CI jobs of their own — run each from the repo root:
 
 ```bash
 # proxy addon — pin mitmproxy the way CI does, to the version production runs
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,7 +15,18 @@ wt test                            # run pytest in generator/
 uvx tend@latest init               # regenerate workflows from .config/tend.yaml
 uvx tend@latest init --dry-run     # preview without writing
 uvx tend@latest check              # verify branch protection, secrets, bot access
-pre-commit run --all-files         # lint: ruff, typos, actionlint, uv-lock
+pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, uv-lock
+```
+
+`wt test` covers `generator/` only. Two smaller suites live beside the code
+they test and have CI jobs of their own — run them from their own directory:
+
+```bash
+# proxy addon — pin mitmproxy the way CI does, to the version production runs
+cd proxy && uv run --no-project --with pytest \
+  --with "mitmproxy==$(yq -e '.inputs.mitmproxy_version.default' ../claude/action.yaml)" pytest
+# install-tend OAuth wrapper
+cd plugins/install-tend/skills/install-tend/scripts && uv run --no-project --with pytest pytest
 ```
 
 ## Architecture
@@ -106,8 +117,8 @@ from `.config/tend.yaml`. Edit the generator or config, not the workflow files
 directly.
 
 The generator is a Python package under `generator/` — uses the uv_build
-backend, requires Python 3.11+. Runtime dependencies: click, ruamel.yaml.
-Dev dependencies: pytest, pytest-regtest.
+backend, requires Python 3.11+. Runtime dependencies: click, jinja2,
+ruamel.yaml. Dev dependencies: pytest, pytest-regtest.
 
 Consuming repos regenerate their `tend-*.yaml` workflows nightly (tend itself
 included — it dogfoods its own workflows). Changes to the generator do not
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -11,26 +11,15 @@ completely — old formats should fail with a clear error, not silently parse.
 ## Commands
 
 ```bash
-wt test                            # run pytest in generator/
+wt test                            # every suite (generator/, proxy/, install-tend scripts, worker/)
 uvx tend@latest init               # regenerate workflows from .config/tend.yaml
 uvx tend@latest init --dry-run     # preview without writing
 uvx tend@latest check              # verify branch protection, secrets, bot access
 pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, uv-lock
 ```
 
-`wt test` covers `generator/` only. Two more pytest suites live beside the code
-they test and have CI jobs of their own — run each from the repo root:
-
-```bash
-# proxy addon — pin mitmproxy the way CI does, to the version production runs
-(cd proxy && uv run --no-project --with pytest \
-  --with "mitmproxy==$(yq -e '.inputs.mitmproxy_version.default' ../claude/action.yaml)" pytest)
-# install-tend OAuth wrapper
-(cd plugins/install-tend/skills/install-tend/scripts && uv run --no-project --with pytest pytest)
-```
-
-`worker/` is a third suite outside `wt test`, vitest rather than pytest, with
-its own CI job — see [`worker/README.md`](worker/README.md).
+`wt test` runs [`dev/test.sh`](dev/test.sh), which mirrors the test jobs in
+`ci.yaml`; arguments go to the pytest suites (`wt test -k render`).
 
 ## Architecture
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,16 +18,19 @@ uvx tend@latest check              # verify branch protection, secrets, bot acce
 pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, uv-lock
 ```
 
-`wt test` covers `generator/` only. Two smaller suites live beside the code
+`wt test` covers `generator/` only. Two more pytest suites live beside the code
 they test and have CI jobs of their own — run them from their own directory:
 
 ```bash
 # proxy addon — pin mitmproxy the way CI does, to the version production runs
-cd proxy && uv run --no-project --with pytest \
-  --with "mitmproxy==$(yq -e '.inputs.mitmproxy_version.default' ../claude/action.yaml)" pytest
+(cd proxy && uv run --no-project --with pytest \
+  --with "mitmproxy==$(yq -e '.inputs.mitmproxy_version.default' ../claude/action.yaml)" pytest)
 # install-tend OAuth wrapper
-cd plugins/install-tend/skills/install-tend/scripts && uv run --no-project --with pytest pytest
+(cd plugins/install-tend/skills/install-tend/scripts && uv run --no-project --with pytest pytest)
 ```
+
+`worker/` is a third suite outside `wt test`, vitest rather than pytest, with
+its own CI job — see [`worker/README.md`](worker/README.md).
 
 ## Architecture
 
--- a/dev/test.sh
+++ b/dev/test.sh
@@ -18,9 +18,13 @@ suite() {
   shift
   printf '\n==> %s: %s\n' "$dir" "$*"
   (cd "$dir" && "$@") || rc=$?
-  # pytest exits 5 for "no tests collected" — what a -k aimed at one suite looks
-  # like from the others. Only a filtered run gets to treat that as a pass.
-  if [ "$rc" -eq 5 ] && [ ${#pytest_args[@]} -gt 0 ]; then rc=0; fi
+  # A filtered run gets slack: pytest exits 5 for "no tests collected" and 4 for
+  # a path that only exists in another suite. generator/ stays strict, so a bad
+  # flag — 4 everywhere — still fails the run.
+  if [ ${#pytest_args[@]} -gt 0 ] &&
+    { [ "$rc" -eq 5 ] || { [ "$rc" -eq 4 ] && [ "$dir" != generator ]; }; }; then
+    rc=0
+  fi
   if [ "$rc" -ne 0 ]; then failed+=("$dir"); fi
 }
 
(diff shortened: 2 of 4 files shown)


CLAUDE.md
```diff
@@ -15,7 +15,18 @@ wt test                            # run pytest in generator/
...
+
+`wt test` covers `generator/` only. Two smaller suites live beside the code
+they test and have CI jobs of their own — run them from their own directory:
+
+```bash
+# proxy addon — pin mitmproxy the way CI does, to the version production runs
+cd proxy && uv run --no-project --with pytest \
+  --with "mitmproxy==$(yq -e '.inputs.mitmproxy_version.default' ../claude/action.yaml)" pytest
+# install-tend OAuth wrapper
+cd plugins/install-tend/skills/install-tend/scripts && uv run --no-project --with pytest pytest
```
tend-agent: The two `cd`s compound, so this block only works one command at a time. Wrapping each in a subshell keeps every line runnable from the repo root, matching the block above.

```suggestion
(cd proxy && uv run --no-project --with pytest \
  --with "mitmproxy==$(yq -e '.inputs.mitmproxy_version.default' ../claude/action.yaml)" pytest)
# install-tend OAuth wrapper
(cd plugins/install-tend/skills/install-tend/scripts && uv run --no-project --with pytest pytest)
```

CLAUDE.md
```diff
@@ -15,7 +15,18 @@ wt test                            # run pytest in generator/
 uvx tend@latest init               # regenerate workflows from .config/tend.yaml
 uvx tend@latest init --dry-run     # preview without writing
 uvx tend@latest check              # verify branch protection, secrets, bot access
-pre-commit run --all-files         # lint: ruff, typos, actionlint, uv-lock
+pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, uv-lock
+```
+
+`wt test` covers `generator/` only. Two smaller suites live beside the code
+they test and have CI jobs of their own — run them from their own directory:
```
tend-agent: `worker/` is a third suite `wt test` doesn't reach, with its own CI job (`test-worker`). Scoping the count to pytest keeps it true, and a pointer covers `worker/` without duplicating what its README already has.

```suggestion
`wt test` covers `generator/` only. Two more pytest suites live beside the code
they test and have CI jobs of their own (`worker/` has a vitest suite too — see
[`worker/README.md`](worker/README.md)) — run them from their own directory:
```

tend-agent (commented):
Checked each claim against the file it describes: `shellcheck` is in `.pre-commit-config.yaml`, `jinja2>=3.1.6` is in `generator/pyproject.toml`, and `.config/wt.toml`'s `test` alias really is `cd generator && uv run pytest`. Both new invocations pass as written when run one at a time from the repo root (23 passed in `proxy/`, 9 passed in the install-tend scripts dir).

Two things about the new block itself.

The `cd`s are cumulative, so the block only works one command at a time — pasting both lines leaves you in `proxy/` and the second `cd` fails with `No such file or directory`. Every other command in the block two lines above runs from the repo root, so that's the expectation a reader arrives with. Subshells restore it.

The other is the count. `worker/` is a third suite outside `wt test` with a CI job of its own (`test-worker` in `.github/workflows/ci.yaml`), so a contributor who lands in `worker/src/` reads "two smaller suites", matches neither directory, and concludes there's nothing to run — the same trap this PR is closing, one directory over. It's less sharp there because `worker/README.md` already documents `npm install && npm test`, so a pointer beats duplicating the commands.

CLAUDE.md
tend-agent: Applied. Both lines now run from the repo root back to back — verified in one shell: `9 passed` in the install-tend scripts dir, then `23 passed` in `proxy/`, with `pwd` still the repo root after each.

CLAUDE.md
tend-agent: Confirmed — `worker/` runs `npm run typecheck` and `npm test` under `test-worker` in [`ci.yaml`](https://github.com/max-sixty/tend/blob/674b54d/.github/workflows/ci.yaml#L104-L118), and `worker/package.json` maps `test` to `vitest run`.

Took the substance with slightly different shape: the count is now "two more pytest suites", and `worker/` gets its own sentence after the block pointing at `worker/README.md` rather than a parenthetical in the lead-in. Same effect for the contributor who lands in `worker/src/`, and it keeps the lead sentence about the two commands directly below it.

CLAUDE.md
```diff
@@ -15,9 +15,23 @@ wt test                            # run pytest in generator/
 uvx tend@latest init               # regenerate workflows from .config/tend.yaml
 uvx tend@latest init --dry-run     # preview without writing
 uvx tend@latest check              # verify branch protection, secrets, bot access
-pre-commit run --all-files         # lint: ruff, typos, actionlint, uv-lock
+pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, uv-lock
 ```
 
+`wt test` covers `generator/` only. Two more pytest suites live beside the code
+they test and have CI jobs of their own — run them from their own directory:
```
tend-agent: The subshells made these root-runnable, so "from their own directory" is now the one place they break — from inside `proxy/`, `cd proxy` fails.

```suggestion
they test and have CI jobs of their own — run each from the repo root:
```

tend-agent (commented):
Both prior threads are addressed — the subshells hold `pwd` at the repo root (checked by running the install-tend command and the `yq` lookup from `proxy/` back to back; `pwd` was still the repo root after each), and `worker/` now has its own sentence. Re-verified the underlying claims against the tree: `test-worker` in `ci.yaml` runs `npm run typecheck` and `npm test`, `worker/package.json` maps `test` to `vitest run`, `jinja2>=3.1.6` is in `generator/pyproject.toml`, and `shellcheck` is in `.pre-commit-config.yaml`.

One thing the subshell change left behind: the lead-in still says to run the commands from their own directory, which is now the one way they don't work. A reader who follows it literally lands in `proxy/`, and `cd proxy` fails from there — the same shape of trap this PR is closing.

CLAUDE.md
tend-agent: Right — the subshells inverted that sentence. Taken as suggested: "run each from the repo root". The PR body was already updated to the root-runnable framing in the prior push, so `CLAUDE.md` was the only stale spot.

max-sixty:
change wt test!!

tend-agent:
Done — `wt test` now runs all four suites instead of documenting the three it missed. The alias points at [`dev/test.sh`](https://github.com/max-sixty/tend/blob/cde97e1/dev/test.sh), which runs `generator/`, `proxy/` (mitmproxy pinned from `claude/action.yaml`, as `ci.yaml` does), the install-tend scripts, and `worker/`, reporting all failures rather than stopping at the first. `wt test -k render` still works: args forward to the pytest suites, and a suite the filter misses exits 5 (`no tests collected`), tolerated only when a filter was given.

The CLAUDE.md block naming the hand-run commands is gone; what's left is the one-line pointer plus the two corrections that stand on their own (`shellcheck` in the lint list, `jinja2` in the runtime deps).

One thing worth your call: `ci.yaml` still spells out its own four jobs, so a fifth suite could land in one place and not the other. Keeping them separate buys parallelism and per-job caching, so I left it — say the word if you'd rather the jobs call the script.

dev/test.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# suite <dir> <cmd>...
+suite() {
+  local dir=$1 rc=0
+  shift
+  printf '\n==> %s: %s\n' "$dir" "$*"
+  (cd "$dir" && "$@") || rc=$?
+  # pytest exits 5 for "no tests collected" — what a -k aimed at one suite looks
+  # like from the others. Only a filtered run gets to treat that as a pass.
+  if [ "$rc" -eq 5 ] && [ ${#pytest_args[@]} -gt 0 ]; then rc=0; fi
```
tend-agent:
```suggestion
  # A filtered run gets slack: pytest exits 5 for "no tests collected" and 4 for
  # a path that only exists in another suite. generator/ stays strict, so a bad
  # flag — 4 everywhere — still fails the run.
  if [ ${#pytest_args[@]} -gt 0 ] &&
    { [ "$rc" -eq 5 ] || { [ "$rc" -eq 4 ] && [ "$dir" != generator ]; }; }; then
    rc=0
  fi
```

dev/test.sh
```diff
@@ -0,0 +1,60 @@
...
+
+suite plugins/install-tend/skills/install-tend/scripts \
+  uv run --no-project --with pytest pytest "${pytest_args[@]}"
+
+if [ ${#pytest_args[@]} -eq 0 ]; then
+  # Install only when the tree is missing, and with `npm ci` rather than
+  # `npm install` — an older local npm reruns resolution and rewrites
+  # package-lock.json, leaving churn in the diff that has nothing to do with the
+  # change under test.
+  if [ ! -d worker/node_modules ]; then
```
tend-agent:
```suggestion
  # Install when the tree is missing or older than the lockfile (`npm ci` writes
  # node_modules/.package-lock.json), and with `npm ci` rather than `npm install`
  # — an older local npm reruns resolution and rewrites package-lock.json,
  # leaving churn in the diff that has nothing to do with the change under test.
  if [ ! -d worker/node_modules ] ||
    [ worker/package-lock.json -nt worker/node_modules/.package-lock.json ]; then
```

tend-agent (commented):
Two things the incremental turns up; both checked by running the script in this checkout.

**`wt test <path>` regressed.** The old alias ran `pytest {{ args }}` inside `generator/`, so `wt test tests/test_migrate.py` worked. Now the same invocation runs those 8 tests green in `generator/` and then exits 1 with `failed: proxy plugins/install-tend/skills/install-tend/scripts` — a path that only exists in one suite makes pytest exit 4 from the others, and only 5 is tolerated. A path is as ordinary a pytest argument as `-k`, and the failure names suites the developer never aimed at. The suggestion widens the slack to 4 while keeping `generator/` strict, so a genuine bad flag — 4 in every suite — still fails.

**The `worker/` install skips a stale tree.** `npm ci` runs only when `node_modules` is absent, so after the lockfile moves — a dependency bump, a branch switch — tsc and vitest run against whatever was installed last while `test-worker` reruns `npm ci` and can go red. Same green-locally-red-in-CI shape this PR is closing, one directory over. `npm ci` writes `node_modules/.package-lock.json`, so an `-nt` test against the lockfile catches it without paying for a reinstall every run.

Against the merge checkout, with both suggestions applied (shellcheck `-S warning` clean):

- `./dev/test.sh tests/test_migrate.py` — before: exit 1, `failed: proxy plugins/install-tend/skills/install-tend/scripts`. After: exit 0, 8 passed in `generator`.
- `./dev/test.sh --bogus-flag` — exit 1, `failed: generator`.
- `./dev/test.sh -k render` — exit 0. `./dev/test.sh -k zzz_no_such` — exit 0.
- `touch worker/package-lock.json && ./dev/test.sh` — exit 0, and `==> worker: npm ci` reappears on the already-populated tree.

Everything else in the diff holds up: `jinja2>=3.1.6` is in `generator/pyproject.toml`, `shellcheck` is in `.pre-commit-config.yaml`, and `generator/`, `proxy/`, the install-tend scripts, and `worker/` really are the whole set — `site/package.json` has no `test` script.

</details>

dev/test.sh
tend-agent: Applied in a457051. Confirmed the regression and the fix: `./dev/test.sh tests/test_migrate.py` exited 1 with `failed: proxy plugins/install-tend/skills/install-tend/scripts` before, exits 0 with 8 passed now. `--bogus-flag` still exits 1 with `failed: generator`, and both `-k render` and `-k zzz_no_such` stay at 0.

dev/test.sh
tend-agent: Applied in a457051. `touch worker/package-lock.json && ./dev/test.sh` now reinstalls on the already-populated tree; the next run right after skips `npm ci` and goes straight to typecheck plus vitest, and `package-lock.json` stays clean.

PR #898, opened 2026-08-08
fix(nightly): test-merge bot PRs locally instead of filtering the lazy mergeable field

### LINKED ISSUE #897: nightly: conflict detection reads mergeable once, so a lazily-computed UNKNOWN is silently treated as no conflict
## Problem

Step 3 of the bundled `nightly` skill detects conflicted PRs with a single query whose result is filtered by string equality:

```bash
for author in "$BOT_LOGIN" app/dependabot app/renovate; do
  gh pr list --author "$author" --json number,title,mergeable,headRefName,author \
    --jq '.[] | select(.mergeable == "CONFLICTING")'
done
```

followed by *"Skip the rest of this step if none of the queries return anything."*

`mergeable` is not a stored field — GitHub computes it lazily. The first query after the base branch moves returns `UNKNOWN` and *enqueues* the computation; a later query returns the real value ([REST](https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request): *"if the value is null … a background job was started to compute mergeability … submit the request again"*; [GraphQL `MergeableState`](https://docs.github.com/en/graphql/reference/enums#mergeablestate) exposes the same as `UNKNOWN`). `select(.mergeable == "CONFLICTING")` drops `UNKNOWN` silently, so a cold cache is indistinguishable from a clean one — and the skip line then turns "I don't know" into "there are no conflicts", skipping the whole step including the *Bot-authored PRs: resolve manually* subagent dispatch.

## Reproduction

One `gh pr list --state open --json number,author,mergeable` against PRQL/prql, then the identical command ~2 minutes later, nothing else touched in between:

| | 1st query | 2nd query |
| --- | --- | --- |
| `UNKNOWN` | 16 | 0 |
| `CONFLICTING` | 4 | 18 |
| `MERGEABLE` | 6 | 8 |

Fourteen genuinely-conflicted PRs reported `UNKNOWN` on the first call and `CONFLICTING` on the second. Against a cold cache the recipe finds 4 of 18.

This is visible in tend's own session logs, not just in a hand-run experiment. Nightly run [30737292659](https://github.com/PRQL/prql/actions/runs/30737292659) happened to print `mergeable` unfiltered for every open PR, and the tool result recorded 15 of 20 as `UNKNOWN` — including four PRs that the *same* skill's filtered query is supposed to catch.

## Observed impact

[PRQL/prql#5374](https://github.com/PRQL/prql/pull/5374) is a useful control: open since 2025-08-04, no commits or edits since 2026-05-17, permanently `CONFLICTING`. Across ten sampled nightly runs the recipe found it in six and missed it in four:

| Nightly run | Date | `#5374` reported? |
| --- | --- | --- |
| [30430937291](https://github.com/PRQL/prql/actions/runs/30430937291) | 07-29 | yes |
| [30522182713](https://github.com/PRQL/prql/actions/runs/30522182713) | 07-30 | yes |
| [30612347726](https://github.com/PRQL/prql/actions/runs/30612347726) | 07-31 | yes |
| [30689156475](https://github.com/PRQL/prql/actions/runs/30689156475) | 08-01 | yes |
| [30793747427](https://github.com/PRQL/prql/actions/runs/30793747427) | 08-03 | **no** |
| [30886931468](https://github.com/PRQL/prql/actions/runs/30886931468) | 08-04 | yes |
| [30984187113](https://github.com/PRQL/prql/actions/runs/30984187113) | 08-05 | **no** |
| [31080122831](https://github.com/PRQL/prql/actions/runs/31080122831) | 08-06 | yes |
| [31154968590](https://github.com/PRQL/prql/actions/runs/31154968590) | 08-07 | **no** |
| [31244074624](https://github.com/PRQL/prql/actions/runs/31244074624) | 08-08 | **no** |

In each of the four misses the same session's later unfiltered `gh pr list` output does contain `#5374`, so the PR was in the list — only the `mergeable` field was cold. That rules out pagination, `--author` filtering, and search flakiness.

The 08-08 run is the one with real consequences. It reported *"Conflicts — no conflicted PRs from this bot, Dependabot, or Renovate"* and skipped Step 3. Two **bot-authored** PRs — [PRQL/prql#6137](https://github.com/PRQL/prql/pull/6137) and [PRQL/prql#6150](https://github.com/PRQL/prql/pull/6150) — were `CONFLICTING` at the time and still are. `main` last moved at 02:38Z, nearly four hours before the 06:30Z run, so this is not a race with a recent merge. Those are exactly the PRs the *"Bot-authored PRs: resolve manually"* subagent path exists to rebase; instead they sit unmergeable and nobody is told.

Because `#5374` carries a human commit, the dependency-bot branch correctly skips it whenever it *is* detected — so until now the visible symptom has been confined to the bot's own PRs quietly accumulating conflicts.

## Suggested fix

Re-query until nothing is `UNKNOWN`, and surface a persistent `UNKNOWN` rather than folding it into "no conflicts":

```bash
for author in "$BOT_LOGIN" app/dependabot app/renovate; do
  # `mergeable` is computed lazily: the first query after `main` moves returns
  # UNKNOWN and enqueues the computation. Re-query until it settles.
  for _ in 1 2 3 4 5; do
    gh pr list --author "$author" \
      --json number,title,mergeable,headRefName,author > "/tmp/prs-$author.json"
    [ "$(jq '[.[] | select(.mergeable == "UNKNOWN")] | length' "/tmp/prs-$author.json")" -eq 0 ] && break
    sleep 10
  done
  jq -r '.[] | select(.mergeable == "UNKNOWN")
    | "still UNKNOWN after retries, treat as possibly conflicted: #\(.number) \(.title)"' "/tmp/prs-$author.json"
  jq -c '.[] | select(.mergeable == "CONFLICTING")' "/tmp/prs-$author.json"
done
```

In the reproduction above a single retry ~2 minutes later settled all 16; a 10-second gap is plausible but I haven't measured how short it can safely be. The skip line wants a matching qualifier — something like *"Skip the rest of this step only if every PR reported a settled `MERGEABLE`; a PR still `UNKNOWN` after the retries should be checked out and test-merged rather than assumed clean."*

The same silent-`UNKNOWN` shape is worth a grep across the other bundled skills for any other `select(.mergeable == …)` use.

<details><summary>How this was found</summary>

PRQL's daily `tend-review-runs` sweep ([31247548935](https://github.com/PRQL/prql/actions/runs/31247548935)) cross-checked the 08-08 nightly's "no conflicted PRs" claim against the live PR list and found two bot PRs conflicted. Session logs for the ten sampled nightlies were downloaded and the conflict-check tool result read directly out of each JSONL, so the "reported?" column is the command's actual output, not a run summary.

Recipe under discussion: `plugins/tend-ci-runner/skills/nightly/SKILL.md`, Step 3, lines 78–87 at `c11ab3c`.

</details>

tend-agent:
## Problem

Step 3 of the bundled `nightly` skill found conflicted bot PRs with a single `gh pr list` whose result was filtered on `select(.mergeable == "CONFLICTING")`, followed by *"Skip the rest of this step if none of the queries return anything."*

`mergeable` is not stored — GitHub computes it lazily. The first query after the base branch moves returns `UNKNOWN` and enqueues the computation; a later query returns the real value. `select(.mergeable == "CONFLICTING")` drops `UNKNOWN` silently, so a cold cache was indistinguishable from a clean one, and the skip line turned "I don't know" into "there are no conflicts" — skipping the whole step, including the *Bot-authored PRs: resolve manually* subagent dispatch that exists to rebase the bot's own conflicted PRs.

Reported in #897 with session-log evidence from ten sampled nightly runs.

## Reproduction

Confirmed live against `nodejs/node`, two queries seconds apart with nothing touched in between:

| | 1st query | 2nd query |
| --- | --- | --- |
| `UNKNOWN` | 36 | 0 |
| `CONFLICTING` | 0 | **1** |
| `MERGEABLE` | 24 | 59 |

The cold read reported zero conflicted PRs; the settled read revealed one that was genuinely `CONFLICTING`.

## Solution

Don't read `mergeable` at all — compute the merge locally with `git merge-tree --write-tree`, which is synchronous and authoritative, so there is no cache to wait on and no retry loop. There is no blocking read on the GitHub side: [the REST docs](https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request) prescribe resubmitting the request until `mergeable` is non-null.

The recipe fetches every PR head in one batched `git fetch` (`refs/pull/N/head`, so fork PRs are covered too, `--force` because bot branches get force-pushed), then test-merges each against `origin/main`. Failure stays distinguishable from clean at both steps: an empty `gh pr list` output file (0 bytes, versus 3 for `[]`) reports as unverified, and `git merge-tree` exits non-zero both for a real conflict and for a ref it couldn't resolve — the conflict prints the tree OID and conflicted paths on stdout, the error prints nothing, so the recipe tests the captured output rather than the exit status alone.

The output path is `/tmp/prs-${author//\//-}.json` — `app/dependabot` and `app/renovate` contain a slash, so the un-sanitized form suggested in the issue would redirect into a nonexistent `/tmp/prs-app/` directory and fail.

## Testing

Ran the recipe verbatim from the repo root against this repo's own PRs: it agrees exactly with GitHub's own (warm, therefore authoritative) answer across all 30 open bot PRs, flagging only #856 as `CONFLICTING`, and finishes in ~2s for all three authors with no sleeps. Both failure branches were exercised directly — a `gh pr list` that can't resolve the repo reports "query never landed", and an unfetched head ref reports "merge test never ran" — neither reads as clean.

`select(.mergeable` appears at no other site in the bundled skills; the only other `mergeable` mentions are workflow comments about merge-ref materialization, which are unrelated.

`fetch-depth: 0` is set by the shared checkout macro in `generator/src/tend/templates/macros.yaml.j2`, so every generated nightly job has the history the local merge needs.

---

Closes #897 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -76,15 +76,24 @@ a one-line reason) plus a `_Last refreshed: <YYYY-MM-DD>_` footer. Updates:
 
 Find conflicted PRs from this bot and from upstream dependency bots:
 
+`mergeable` is computed lazily, not stored. The first query after `main` moves returns `UNKNOWN` for every PR whose merge hasn't been recomputed and *enqueues* the computation; a later query returns the real value. Filtering a cold read on `== "CONFLICTING"` therefore reports a conflicted PR as clean. Re-query until it settles:
+
 ```bash
 BOT_LOGIN=$(gh api user --jq '.login')
 for author in "$BOT_LOGIN" app/dependabot app/renovate; do
-  gh pr list --author "$author" --json number,title,mergeable,headRefName,author \
-    --jq '.[] | select(.mergeable == "CONFLICTING")'
+  out="/tmp/prs-${author//\//-}.json"   # `app/dependabot` has a slash; strip it
+  for _ in 1 2 3 4 5; do
+    gh pr list --author "$author" --json number,title,mergeable,headRefName,author > "$out"
+    [ "$(jq '[.[] | select(.mergeable == "UNKNOWN")] | length' "$out")" -eq 0 ] && break
+    sleep 10
+  done
+  jq -c '.[] | select(.mergeable == "CONFLICTING")' "$out"
+  jq -r '.[] | select(.mergeable == "UNKNOWN")
+    | "unsettled, treat as possibly conflicted: #\(.number) \(.title)"' "$out"
 done
 ```
 
-Skip the rest of this step if none of the queries return anything.
+Skip the rest of this step only when every PR settled and none came back `CONFLICTING`. A PR still `UNKNOWN` after the retries is not a clean one — check it out and test-merge (`git merge --no-commit --no-ff origin/main`) before dismissing it, and report it as unverified rather than counting it clean.
 
 ### Upstream dependency bots: trigger the bot's own rebase
 
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -82,33 +82,23 @@ Don't filter `gh pr list --json mergeable` on `== "CONFLICTING"`. `mergeable` is
 BOT_LOGIN=$(gh api user --jq '.login')
 git fetch --quiet origin main
 for author in "$BOT_LOGIN" app/dependabot app/renovate; do
-  out="/tmp/prs-${author//\//-}.json"   # `app/dependabot` has a slash; strip it
-  # `--limit 100` is load-bearing: `gh pr list` defaults to 30 and truncates
-  # silently, so PR 31 onward would never be test-merged.
-  gh pr list --author "$author" --limit 100 --json number,title,headRefName,author > "$out"
-  # A failed query leaves `$out` empty, which reads as "no conflicts" — the
-  # same silent-miss shape. `[]` (no open PRs) is 3 bytes, a failure is 0.
-  [ -s "$out" ] || { echo "query for $author never landed — conflicts unverified"; continue; }
+  # A failed query and "no open PRs" both print nothing; only the first is a
+  # reason to stop. --limit 100 because the default 30 truncates silently.
+  prs=$(gh pr list --author "$author" --limit 100 --json number,title) \
+    || { echo "query for $author never landed — conflicts unverified"; continue; }
   # One fetch for every head, forced because bot branches get force-pushed.
-  # `refs/pull/N/head` also covers PRs opened from a fork.
-  mapfile -t refs < <(jq -r '.[].number | "refs/pull/\(.)/head:refs/tend/pr/\(.)"' "$out")
-  [ "${#refs[@]}" -eq 0 ] || git fetch --quiet --force origin "${refs[@]}"
-  jq -r '.[] | "\(.number)\t\(.title)"' "$out" | while IFS=$'\t' read -r n title; do
-    # Clean merge: exit 0. Conflict: non-zero with the tree OID and conflicted
-    # paths on stdout. Anything else (ref never fetched, bad revision): non-zero
-    # with stdout empty — so test `$tree`, not the exit status alone.
-    if tree=$(git merge-tree --write-tree origin/main "refs/tend/pr/$n" 2>/dev/null); then
-      continue
-    elif [ -n "$tree" ]; then
-      echo "CONFLICTING: $author #$n $title"
-    else
-      echo "merge test never ran, conflicts unverified: $author #$n $title"
-    fi
+  mapfile -t refs < <(jq -r '.[].number | "refs/pull/\(.)/head:refs/tend/pr/\(.)"' <<<"$prs")
+  [ "${#refs[@]}" -eq 0 ] && continue
+  git fetch --quiet --force origin "${refs[@]}"
+  # Conflict, missing ref, unreadable ref: all non-zero, all need a look.
+  jq -r '.[] | [.number, .title] | @tsv' <<<"$prs" | while IFS=$'\t' read -r n title; do
+    git merge-tree --write-tree origin/main "refs/tend/pr/$n" >/dev/null \
+      || echo "needs a rebase: $author #$n $title"
   done
 done
 ```
 
-Skip the rest of this step only when every query landed and nothing printed. A PR whose merge test never ran is not a clean one — investigate it and report it as unverified rather than counting it clean.
+Skip the rest of this step only when every query landed and nothing printed.
 
 ### Upstream dependency bots: trigger the bot's own rebase
 
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -83,7 +83,9 @@ BOT_LOGIN=$(gh api user --jq '.login')
 git fetch --quiet origin main
 for author in "$BOT_LOGIN" app/dependabot app/renovate; do
   out="/tmp/prs-${author//\//-}.json"   # `app/dependabot` has a slash; strip it
-  gh pr list --author "$author" --json number,title,headRefName,author > "$out"
+  # `--limit 100` is load-bearing: `gh pr list` defaults to 30 and truncates
+  # silently, so PR 31 onward would never be test-merged.
+  gh pr list --author "$author" --limit 100 --json number,title,headRefName,author > "$out"
   # A failed query leaves `$out` empty, which reads as "no conflicts" — the
   # same silent-miss shape. `[]` (no open PRs) is 3 bytes, a failure is 0.
   [ -s "$out" ] || { echo "query for $author never landed — conflicts unverified"; continue; }
@@ -98,9 +100,9 @@ for author in "$BOT_LOGIN" app/dependabot app/renovate; do
     if tree=$(git merge-tree --write-tree origin/main "refs/tend/pr/$n" 2>/dev/null); then
       continue
     elif [ -n "$tree" ]; then
-      echo "CONFLICTING: #$n $title"
+      echo "CONFLICTING: $author #$n $title"
     else
-      echo "merge test never ran, conflicts unverified: #$n $title"
+      echo "merge test never ran, conflicts unverified: $author #$n $title"
     fi
   done
 done
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -76,28 +76,37 @@ a one-line reason) plus a `_Last refreshed: <YYYY-MM-DD>_` footer. Updates:
 
 Find conflicted PRs from this bot and from upstream dependency bots:
 
-`mergeable` is computed lazily, not stored. The first query after `main` moves returns `UNKNOWN` for every PR whose merge hasn't been recomputed and *enqueues* the computation; a later query returns the real value. Filtering a cold read on `== "CONFLICTING"` therefore reports a conflicted PR as clean. Re-query until it settles:
+Don't filter `gh pr list --json mergeable` on `== "CONFLICTING"`. `mergeable` is computed lazily: the first query after `main` moves returns `UNKNOWN` and only *enqueues* the computation, so a cold read reports a conflicted PR as clean. There is no blocking read — [the REST docs](https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request) prescribe resubmitting the request until the value settles. Test-merge locally instead: `git merge-tree` answers the same question synchronously, in-process, with no retry loop.
 
 ```bash
 BOT_LOGIN=$(gh api user --jq '.login')
+git fetch --quiet origin main
 for author in "$BOT_LOGIN" app/dependabot app/renovate; do
   out="/tmp/prs-${author//\//-}.json"   # `app/dependabot` has a slash; strip it
-  for _ in 1 2 3 4 5; do
-    gh pr list --author "$author" --json number,title,mergeable,headRefName,author > "$out"
-    # A failed query leaves `$out` empty, which reads as "no conflicts" — the
-    # same silent-miss shape. `[]` (no open PRs) is 3 bytes, a failure is 0.
-    [ -s "$out" ] || { sleep 10; continue; }
-    [ "$(jq '[.[] | select(.mergeable == "UNKNOWN")] | length' "$out")" -eq 0 ] && break
-    sleep 10
+  gh pr list --author "$author" --json number,title,headRefName,author > "$out"
+  # A failed query leaves `$out` empty, which reads as "no conflicts" — the
+  # same silent-miss shape. `[]` (no open PRs) is 3 bytes, a failure is 0.
+  [ -s "$out" ] || { echo "query for $author never landed — conflicts unverified"; continue; }
+  # One fetch for every head, forced because bot branches get force-pushed.
+  # `refs/pull/N/head` also covers PRs opened from a fork.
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -76,15 +76,24 @@ a one-line reason) plus a `_Last refreshed: <YYYY-MM-DD>_` footer. Updates:
...
 BOT_LOGIN=$(gh api user --jq '.login')
 for author in "$BOT_LOGIN" app/dependabot app/renovate; do
-  gh pr list --author "$author" --json number,title,mergeable,headRefName,author \
-    --jq '.[] | select(.mergeable == "CONFLICTING")'
+  out="/tmp/prs-${author//\//-}.json"   # `app/dependabot` has a slash; strip it
+  for _ in 1 2 3 4 5; do
+    gh pr list --author "$author" --json number,title,mergeable,headRefName,author > "$out"
+    [ "$(jq '[.[] | select(.mergeable == "UNKNOWN")] | length' "$out")" -eq 0 ] && break
+    sleep 10
+  done
```
tend-agent: Guard the query itself, so a failed `gh pr list` reports as unverified rather than as "no conflicts":

```suggestion
  for _ in 1 2 3 4 5; do
    if gh pr list --author "$author" --json number,title,mergeable,headRefName,author > "$out" \
       && [ "$(jq '[.[] | select(.mergeable == "UNKNOWN")] | length' "$out")" -eq 0 ]; then break; fi
    sleep 10
  done
  # `[]` is 3 bytes, so -s separates "no open PRs" from "the query never succeeded"
  [ -s "$out" ] || echo "query for $author never succeeded — unverified, do not skip this step"
```

tend-agent (commented):
The lazy-`mergeable` diagnosis is right and the retry loop settles it correctly — I ran the recipe verbatim from the repo root and it exits on the first pass with a warm cache, writes to sanitized paths, and reports #856 as `CONFLICTING`, matching the PR body.

One gap: the redirect makes a failed query indistinguishable from a clean one, which is the same shape as the bug being fixed. `> "$out"` truncates the file before `gh` writes, so any non-zero `gh pr list` (rate limit, transient 5xx, wrong working directory) leaves `$out` empty. `jq` on empty input emits nothing, so `$(...)` is the empty string and `[ "" -eq 0 ]` errors with `integer expression expected` instead of breaking — the loop burns all five iterations, then both reporting `jq` calls print nothing. The step then reads as "every PR settled, none conflicting" and gets skipped. I reproduced this by running the recipe with `gh` unable to resolve the repo: 2m30s per author, no output, exit 0.

`[]` (3 bytes) is non-empty while a failed query leaves 0 bytes, so `[ -s "$out" ]` separates "this author has no open PRs" from "the query never landed" without an extra call. Suggestion inline.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Confirmed and fixed. I reproduced the mechanism directly: `jq` on a 0-byte file exits 0 having printed nothing, so `$(...)` is the empty string, `[ "" -eq 0 ]` fails with `integer expression expected` (rc=2) and the `&& break` never fires — then both reporting `jq` calls print nothing and the step reads as clean.

Took the `[ -s ]` suggestion, since `[]` is 3 bytes and a failed query is 0. Added it in two places: inside the loop, so a transient failure retries instead of falling through to a bogus settled-check, and after the loop, so a query that never landed says so rather than printing nothing.

Verified both paths against the committed text: with `gh` stubbed to `exit 1` it prints `query for <author> never landed — conflicts unverified` for each author; with real `gh` it still exits on the first pass and reports #856 as `CONFLICTING`.

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -76,15 +76,28 @@ a one-line reason) plus a `_Last refreshed: <YYYY-MM-DD>_` footer. Updates:
...
 
+`mergeable` is computed lazily, not stored. The first query after `main` moves returns `UNKNOWN` for every PR whose merge hasn't been recomputed and *enqueues* the computation; a later query returns the real value. Filtering a cold read on `== "CONFLICTING"` therefore reports a conflicted PR as clean. Re-query until it settles:
+
 ```bash
 for author in "$BOT_LOGIN" app/dependabot app/renovate; do
-  gh pr list --author "$author" --json number,title,mergeable,headRefName,author \
-    --jq '.[] | select(.mergeable == "CONFLICTING")'
+  out="/tmp/prs-${author//\//-}.json"   # `app/dependabot` has a slash; strip it
+  for _ in 1 2 3 4 5; do
```
max-sixty: is there a way to do this with a function of "wait for the answer", rather than repeating for 5 steps?

tend-agent: There isn't one on the GitHub side — `mergeable` has no blocking read, and [the REST docs](https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request) say so explicitly: *"If the value is null, then GitHub has started a background job to compute the mergeability. After giving the job time to complete, resubmit the request."* Polling is the documented interface.

But we don't have to ask GitHub. `git merge-tree --write-tree` performs the same merge locally and returns the answer synchronously, so the loop disappears rather than getting wrapped in a function. Pushed in 988b4c1.

Checked against GitHub's own answer on all 30 open bot PRs in this repo, with the cache warm so its values are authoritative: exact agreement, only #856 `CONFLICTING`. Whole step now takes ~2s across all three authors, no sleeps.

The one thing it does need is the PR heads locally — one batched `git fetch` of `refs/pull/N/head` (covers fork PRs too, `--force` since bot branches get force-pushed). `fetch-depth: 0` is already set by the shared checkout macro, so every generated nightly job has the history for it.

Failure stays distinguishable from clean, which was the point of the previous round: `git merge-tree` exits non-zero both for a real conflict and for a ref it couldn't resolve, but a conflict prints the tree OID and conflicted paths on stdout while an error prints nothing — so the recipe tests the captured output, not the status alone. I exercised both branches directly; an unfetched ref reports "merge test never ran, conflicts unverified" rather than passing as clean.

tend-agent (commented):
Dropping `mergeable` for a local test-merge is the right call, and this run happens to demonstrate why: `gh pr list --json mergeable` returned `UNKNOWN` for 24 of the 26 open `tend-agent` PRs on a cold read, while the recipe answered synchronously and flagged **#856 and #809** as conflicting. #809 is the case the old filter drops silently — cold `mergeable` says `UNKNOWN`, the local merge says `CONFLICTING`. The error/conflict discrimination holds up too: an unresolvable ref exits non-zero with empty stdout, so `[ -n "$tree" ]` is load-bearing exactly as the comment claims.

Two gaps, both in the same silent-miss class the PR is closing:

- `gh pr list` has no `--limit`, so it silently truncates at 30. `review-reviewers`'s SKILL.md already carries this as a rule (*"`--limit 100` is load-bearing — `gh pr list` defaults to 30 and truncates silently"*), and this repo is at 26 open bot PRs today; on a dependabot-heavy adopter the 31st PR onward is never test-merged and the step reads as clean for it.
- The output line drops the author. The old snippet emitted the whole JSON object, so `.author` was there; the two sub-steps that consume this list branch on exactly that field — `app/dependabot` / `app/renovate` route to *Upstream dependency bots: trigger the bot's own rebase* (whose table is keyed on the `--author (PR list)` value), `$BOT_LOGIN` routes to the manual worktree resolution. `CONFLICTING: #809 fix(report-failure): …` doesn't say which, and title text isn't a reliable proxy.

Suggestions inline.

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -76,15 +76,37 @@ a one-line reason) plus a `_Last refreshed: <YYYY-MM-DD>_` footer. Updates:
...
+Don't filter `gh pr list --json mergeable` on `== "CONFLICTING"`. `mergeable` is computed lazily: the first query after `main` moves returns `UNKNOWN` and only *enqueues* the computation, so a cold read reports a conflicted PR as clean. There is no blocking read — [the REST docs](https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request) prescribe resubmitting the request until the value settles. Test-merge locally instead: `git merge-tree` answers the same question synchronously, in-process, with no retry loop.
+
 ```bash
+git fetch --quiet origin main
 for author in "$BOT_LOGIN" app/dependabot app/renovate; do
-  gh pr list --author "$author" --json number,title,mergeable,headRefName,author \
-    --jq '.[] | select(.mergeable == "CONFLICTING")'
+  out="/tmp/prs-${author//\//-}.json"   # `app/dependabot` has a slash; strip it
+  gh pr list --author "$author" --json number,title,headRefName,author > "$out"
```
tend-agent:
```suggestion
  gh pr list --author "$author" --limit 100 --json number,title,headRefName,author > "$out"
```

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -76,15 +76,37 @@ a one-line reason) plus a `_Last refreshed: <YYYY-MM-DD>_` footer. Updates:
...
+  mapfile -t refs < <(jq -r '.[].number | "refs/pull/\(.)/head:refs/tend/pr/\(.)"' "$out")
+  [ "${#refs[@]}" -eq 0 ] || git fetch --quiet --force origin "${refs[@]}"
+  jq -r '.[] | "\(.number)\t\(.title)"' "$out" | while IFS=$'\t' read -r n title; do
+    # Clean merge: exit 0. Conflict: non-zero with the tree OID and conflicted
+    # paths on stdout. Anything else (ref never fetched, bad revision): non-zero
+    # with stdout empty — so test `$tree`, not the exit status alone.
+    if tree=$(git merge-tree --write-tree origin/main "refs/tend/pr/$n" 2>/dev/null); then
+      continue
+    elif [ -n "$tree" ]; then
+      echo "CONFLICTING: #$n $title"
```
tend-agent:
```suggestion
      echo "CONFLICTING: $author #$n $title"
```

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -76,15 +76,37 @@ a one-line reason) plus a `_Last refreshed: <YYYY-MM-DD>_` footer. Updates:
...
+    else
+      echo "merge test never ran, conflicts unverified: #$n $title"
```
tend-agent:
```suggestion
      echo "merge test never ran, conflicts unverified: $author #$n $title"
```

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Applied in 97ea7eb, with a comment naming why the flag is there so a future edit doesn't drop it again. This repo is at 24 open bot PRs, so the truncation hasn't bitten here yet — but the old snippet had the same default, and on a dependabot-heavy adopter it fails in exactly the direction this PR is closing: silently, as "clean".

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Applied in 97ea7eb, both branches. `$author` is set by the enclosing `for`, so it's in scope inside the `while` subshell. Output now reads `CONFLICTING: tend-agent #809 fix(report-failure): …`, which routes to the right sub-step without inferring from the title.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Same commit — both echo branches carry `$author` now.

PR #900, opened 2026-08-08
fix(check): stop pointing credential-environments at the setting it rejects

max-sixty:
`credential-environments` failed with a remedy that named "protected branches" as an acceptable deployment policy — but `_policy_gate` refuses exactly that setting, since it keys on a rule covering the branch rather than on who may push it. So the failure message pointed at the one option in the GitHub UI that cannot resolve it. The phrase is doubly misleading here, because tend's own config has a `protected_branches` key meaning something else entirely.

The remedy now names the branch class the check does credit (the wording `docs/security-model.md` already uses), and rules the toggle out in a closing sentence — without repeating the explanation, which `_policy_gate` already supplies as the per-environment reason whenever that setting is the reason for the failure.

Message only — no behaviour change; 354 tests pass.

### DIFF
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -1088,9 +1088,12 @@ def check_credential_environments(
             False,
             "A run the bot can cause reaches a credential: "
             f"{'; '.join(ungated)}. Gate each environment with a required "
-            "reviewer that is not the bot, or a deployment policy naming only "
-            "verified refs (protected branches, or tags under an admin-only "
-            "all-tags ruleset); move an OIDC job into such an environment.",
+            "reviewer that is not the bot, or a deployment policy listing "
+            "only verified refs — branches this run confirmed the bot cannot "
+            "write, or tags under an admin-only all-tags ruleset. The "
+            "policy's 'protected branches' setting is not one of them: it "
+            "keys on a rule covering the branch, not on who may push it. "
+            "Move an OIDC job into such an environment.",
         )
     if surface.unresolved:
         return CheckResult(
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -1090,10 +1090,9 @@ def check_credential_environments(
             f"{'; '.join(ungated)}. Gate each environment with a required "
             "reviewer that is not the bot, or a deployment policy listing "
             "only verified refs — branches this run confirmed the bot cannot "
-            "write, or tags under an admin-only all-tags ruleset. The "
-            "policy's 'protected branches' setting is not one of them: it "
-            "keys on a rule covering the branch, not on who may push it. "
-            "Move an OIDC job into such an environment.",
+            "write, or tags under an admin-only all-tags ruleset; move an "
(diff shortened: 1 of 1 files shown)


generator/src/tend/checks.py
```diff
@@ -1088,9 +1088,12 @@ def check_credential_environments(
...
             f"{'; '.join(ungated)}. Gate each environment with a required "
-            "reviewer that is not the bot, or a deployment policy naming only "
-            "verified refs (protected branches, or tags under an admin-only "
-            "all-tags ruleset); move an OIDC job into such an environment.",
+            "reviewer that is not the bot, or a deployment policy listing "
+            "only verified refs — branches this run confirmed the bot cannot "
+            "write, or tags under an admin-only all-tags ruleset. The "
+            "policy's 'protected branches' setting is not one of them: it "
+            "keys on a rule covering the branch, not on who may push it. "
+            "Move an OIDC job into such an environment.",
```
```suggestion
            "reviewer that is not the bot, or a deployment policy listing "
            "only verified refs — branches this run confirmed the bot cannot "
            "write, or tags under an admin-only all-tags ruleset; move an "
            "OIDC job into such an environment. The policy's 'protected "
            "branches' setting is not one of them.",
```

tend-agent (commented):
The message change itself is right — `_policy_gate` does refuse `protected_branches`, so the old remedy pointed at the one setting that cannot resolve the failure.

One thing worth a look: `_policy_gate` already returns `admits all protected branches, which keys on a rule covering the branch, not on who may push it` as the per-environment reason, and that reason is interpolated into this same string. So in exactly the case this PR is fixing, the emitted message carries that clause twice — once as the reason, once as the caveat. Splitting the old semicolon into a full stop also moves `such an environment` behind the negated sentence, where its nearest antecedent is the setting the message just ruled out. The inline suggestion drops the repeated explanation and restores the semicolon join; it keeps your `is not one of them` wording.

<details><summary>Rendered message today, for an environment with no reviewers and the protected-branches policy</summary>

```
A run the bot can cause reaches a credential: 'release' has no required reviewers, and admits all protected branches, which keys on a rule covering the branch, not on who may push it. Gate each environment with a required reviewer that is not the bot, or a deployment policy listing only verified refs — branches this run confirmed the bot cannot write, or tags under an admin-only all-tags ruleset. The policy's 'protected branches' setting is not one of them: it keys on a rule covering the branch, not on who may push it. Move an OIDC job into such an environment.
```

With the suggestion:

```
A run the bot can cause reaches a credential: 'release' has no required reviewers, and admits all protected branches, which keys on a rule covering the branch, not on who may push it. Gate each environment with a required reviewer that is not the bot, or a deployment policy listing only verified refs — branches this run confirmed the bot cannot write, or tags under an admin-only all-tags ruleset; move an OIDC job into such an environment. The policy's 'protected branches' setting is not one of them.
```

</details>

Everything else checks out: 366 generator tests pass, and `check_environment`'s own `protected_branches` branch already names the custom-list remedy, so there's no second site with the same misdirection.

tend-agent (approved):

PR #901, opened 2026-08-08
fix(run-issue): tell a failed issue-list read from "no issue exists"

max-sixty:
`run_issue_canonical` returns an empty string whether the `gh issue list` behind it failed or genuinely matched nothing, and status 0 either way. `set -e` does not reach inside a command substitution, and both callers read the number back through one, so a transient list failure has no way to be seen at all — it simply reads as "nothing is open".

Both callers then file a record while one is already open. #836's reconcile does not catch this duplicate: it probes the ten numbers below the issue it just created, which is the right window for a racing sibling but not for an existing record, which is normally far older.

On the outage tracker that leaves two open — the state that breaks the drain sweep, since later rows scatter across both and neither carries the complete set. On the rate-limit record it costs an approval outright: the lookup resolves to the lowest-numbered issue, so a maintainer closing the newer one, which is the issue the run's own annotation names, approves nothing and never learns it.

The lookup now carries the failure out and each caller acts on it. `report-failure.sh` records nothing and warns rather than filing a rival tracker — one row is the cheaper loss, and the next failure records normally. `rate-limit-preflight.sh` refuses without filing and names the label instead of a number. That is a fourth annotation state beside the three already distinguished, and it is genuinely distinct from a failed create: there the write failed and there is nothing open to close, whereas here an issue may well be open and worth closing.

The preflight reads twice, and only a run that never managed a good read holds off filing. The first read excludes an already-open issue of any age; the re-read after the jitter only narrows the seconds-wide sibling race, which the reconcile's downward probe catches anyway. So a failed re-read leaves the first read's verdict standing — otherwise the run would pause with no issue at all, which is the outcome opening one exists to avoid.

<details>
<summary>Reproducer, before and after</summary>

A fake `gh` whose `issue list` fails with a 502 while a `tend-outage` tracker is open:

```
--- run_issue_canonical's own status when the list read fails ---
before:  status=0 output=''      <- "no issue exists"
after:   status=1 output=''

--- report-failure.sh with an open tracker it cannot see ---
before:  issue-create calls: 1   <- duplicate tracker filed
after:   issue-create calls: 0   <- warns, records nothing
```

</details>

## Testing

Four tests: one per caller asserting no `issue create` happens when the list read fails, and one for each side of the preflight's two reads — still files when only the re-read fails, still files when only the first read fails. The harness fails `issue list` over a `[FROM, UNTIL]` call range, so either read can be failed alone and the unbounded form still means "the list is simply down".

Verified by mutation, each killing a different test: dropping the status propagation kills both callers' tests, reverting `report-failure.sh`'s guard kills its no-duplicate test, removing the preflight's branch kills its no-duplicate test, letting the re-read clear the verdict kills the re-read test, and dropping the re-read's raise of it kills the first-read test. 370 tests pass; pre-commit clean.

Not exercised against real GitHub — the failure is simulated with `exit 1`, so whether real `gh` exits non-zero on a secondary rate limit during a list read is still untested.

### DIFF
--- a/generator/tests/test_shared_steps.py
+++ b/generator/tests/test_shared_steps.py
@@ -240,7 +240,13 @@ case "$1" in
   issue)
     case "$2" in
       list)
-        if [ -n "${FAIL_ISSUE_LIST:-}" ]; then exit 1; fi
+        # Fail from the Nth list call on, so the spike block's two reads can be
+        # failed independently: `1` fails both, `2` leaves the first standing.
+        if [ -n "${FAIL_ISSUE_LIST_FROM:-}" ]; then
+          n=$(( $(cat "$LIST_CALLS" 2>/dev/null || echo 0) + 1 ))
+          echo "$n" > "$LIST_CALLS"
+          if [ "$n" -ge "$FAIL_ISSUE_LIST_FROM" ]; then exit 1; fi
+        fi
         emit "$(cat "$PAUSE_ISSUES_JSON")"
         ;;
       create)
@@ -359,6 +365,7 @@ def rate_limit_env(tmp_path: Path) -> dict[str, str]:
     return {
         "PATH": f"{bindir}:{Path(jq).parent}:/usr/bin:/bin",
         "GH_CALLS": str(tmp_path / "gh-calls.log"),
+        "LIST_CALLS": str(tmp_path / "list-calls"),
         "TIMELINE_JSON": str(timeline),
         "PAUSE_ISSUES_JSON": str(pause_issues),
         "PROBE_ISSUES_JSON": str(probe_issues),
@@ -510,7 +517,7 @@ def test_rate_limit_files_nothing_when_the_issue_list_cannot_be_read(
     rate_limit_env["FAKE_TODAY_POSTS"] = "16"
     # An issue is already open; the point is that this run cannot see it.
     _approve(rate_limit_env)
-    rate_limit_env["FAIL_ISSUE_LIST"] = "1"
+    rate_limit_env["FAIL_ISSUE_LIST_FROM"] = "1"
 
     result = _run_preflight(rate_limit_env)
 
@@ -519,6 +526,30 @@ def test_rate_limit_files_nothing_when_the_issue_list_cannot_be_read(
     assert not any(c.startswith("issue create") for c in calls), calls
     assert "could not be read" in result.stdout
     assert "could not be filed" not in result.stdout
+
+
+def test_rate_limit_still_files_when_only_the_re_read_fails(
+    rate_limit_env: dict[str, str],
+) -> None:
+    """A failed re-read must not suppress the file the first read cleared.
+
+    The two reads rule out different things. The first excludes an already-open
+    issue of any age, which is the duplicate worth avoiding; the re-read after
+    the jitter only narrows the seconds-wide sibling race, and the reconcile's
+    downward probe catches that anyway. Holding off here would pause the bot
+    with no issue at all — the outcome opening one exists to avoid — and point
+    the maintainer at an issue this run's own first read established isn't
+    there.
+    """
+    rate_limit_env["FAKE_TODAY_POSTS"] = "16"
+    # Nothing open, so the first read is a clean "none"; only the re-read fails.
+    rate_limit_env["FAIL_ISSUE_LIST_FROM"] = "2"
+
+    result = _run_preflight(rate_limit_env)
+
+    assert result.returncode == 1
+    assert any(c.startswith("issue create") for c in _calls(rate_limit_env))
+    assert "could not be read" not in result.stdout
 
 
 def test_rate_limit_human_close_doubles_the_ceiling(
@@ -847,7 +878,11 @@ emit() {
 case "$1 $2" in
   "api user") emit '{"login":"tend-agent","id":4242}' ;;
   "issue list")
-    if [ -n "${FAIL_ISSUE_LIST:-}" ]; then exit 1; fi
+    if [ -n "${FAIL_ISSUE_LIST_FROM:-}" ]; then
+      n=$(( $(cat "$LIST_CALLS" 2>/dev/null || echo 0) + 1 ))
+      echo "$n" > "$LIST_CALLS"
+      if [ "$n" -ge "$FAIL_ISSUE_LIST_FROM" ]; then exit 1; fi
+    fi
     emit "$(cat "$OPEN_ISSUES_JSON")"
     ;;
   "issue create") echo "https://github.com/owner/repo/issues/${FAKE_NEW_ISSUE}" ;;
@@ -891,6 +926,7 @@ def report_failure_env(tmp_path: Path) -> dict[str, str]:
     return {
         "PATH": f"{bindir}:{Path(jq).parent}:/usr/bin:/bin",
         "GH_CALLS": str(tmp_path / "gh-calls.log"),
+        "LIST_CALLS": str(tmp_path / "list-calls"),
         "OPEN_ISSUES_JSON": str(tmp_path / "open-issues.json"),
         "PROBE_ISSUES_JSON": str(tmp_path / "probe-issues.json"),
         "KEEPER_JSON": str(tmp_path / "keeper.json"),
@@ -961,7 +997,7 @@ def test_report_failure_files_nothing_when_the_issue_list_cannot_be_read(
     Path(report_failure_env["OPEN_ISSUES_JSON"]).write_text(
         json.dumps([{"number": 8, "title": OUTAGE_TITLE}])
     )
-    report_failure_env["FAIL_ISSUE_LIST"] = "1"
+    report_failure_env["FAIL_ISSUE_LIST_FROM"] = "1"
 
     result = _run_report_failure(report_failure_env)
 
--- a/generator/tests/test_shared_steps.py
+++ b/generator/tests/test_shared_steps.py
@@ -240,12 +240,18 @@ case "$1" in
   issue)
     case "$2" in
       list)
-        # Fail from the Nth list call on, so the spike block's two reads can be
-        # failed independently: `1` fails both, `2` leaves the first standing.
+        # Fail the list calls in [FROM, UNTIL], so the spike block's two reads
+        # can be failed in any combination: FROM=1 alone fails both, FROM=2
+        # spares the first, and FROM=1 UNTIL=1 spares the re-read. UNTIL is
+        # unbounded by default, which keeps "the list is simply down" open
+        # ended rather than pinned to an exact call count.
         if [ -n "${FAIL_ISSUE_LIST_FROM:-}" ]; then
           n=$(( $(cat "$LIST_CALLS" 2>/dev/null || echo 0) + 1 ))
           echo "$n" > "$LIST_CALLS"
-          if [ "$n" -ge "$FAIL_ISSUE_LIST_FROM" ]; then exit 1; fi
+          if [ "$n" -ge "$FAIL_ISSUE_LIST_FROM" ] \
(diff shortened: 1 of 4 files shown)


shared/steps/rate-limit-preflight.sh
```diff
@@ -100,7 +100,10 @@ fi
 # Everything below runs only once the base limit is already exceeded, so the
 # common path costs no extra API calls at all.
 if [ "$TODAY_POSTS" -gt "$SPIKE_LIMIT" ]; then
-  PAUSE=$(run_issue_canonical "$PAUSE_LABEL" all "$PAUSE_TITLE")
+  # A failed read costs nothing extra here: no issue and no way to see one both
+  # leave approvals at zero, which refuses the run either way. Where the two
+  # part company is whether to file, further down.
+  PAUSE=$(run_issue_canonical "$PAUSE_LABEL" all "$PAUSE_TITLE") || PAUSE=""
```
tend-agent: `set -e` is not the only reason this status is worth carrying: it is this read, not the one after the jitter, that rules out the older open issue the filing guard exists for. Keeping it means a failed re-read below cannot suppress a file the first read already showed to be safe.

```suggestion
  # A failed read costs nothing extra to the verdict: no issue and no way to see
  # one both leave approvals at zero, which refuses the run either way. The
  # status is kept because it decides whether filing is safe, further down — a
  # read that succeeded and found none is what rules out an older open issue.
  PAUSE_LOOKUP_OK=true
  if ! PAUSE=$(run_issue_canonical "$PAUSE_LABEL" all "$PAUSE_TITLE"); then
    PAUSE=""
    PAUSE_LOOKUP_OK=false
  fi
```

shared/steps/rate-limit-preflight.sh
```diff
@@ -162,9 +165,22 @@ if [ "$TODAY_POSTS" -gt "$SPIKE_LIMIT" ]; then
...
+      # this run's annotation names — approves nothing and never learns it.
+      PAUSE_LOOKUP_OK=true
       if [ -z "$PAUSE" ]; then
         sleep $((RANDOM % 30))
-        PAUSE=$(run_issue_canonical "$PAUSE_LABEL" all "$PAUSE_TITLE")
+        if ! PAUSE=$(run_issue_canonical "$PAUSE_LABEL" all "$PAUSE_TITLE"); then
+          PAUSE=""
+          PAUSE_LOOKUP_OK=false
+        fi
       fi
```
tend-agent: With the first read's status kept, this one only has to update it on success. A failure here then leaves that verdict standing, and the duplicate it was protecting against — a sibling filed seconds ago, so numerically adjacent — is the one `run_issue_create_and_reconcile` already probes for.

```suggestion
      if [ -z "$PAUSE" ]; then
        sleep $((RANDOM % 30))
        if PAUSE=$(run_issue_canonical "$PAUSE_LABEL" all "$PAUSE_TITLE"); then
          PAUSE_LOOKUP_OK=true
        else
          PAUSE=""
        fi
      fi
```

The comment block above wants to move with it: what this read decides is the create-create race, and what decides whether filing is safe is the read at the top of the block.

tend-agent (commented):
The status propagation and `report-failure.sh`'s side both look right — three call sites, all guarded, and `set -e` would otherwise abort at each. One finding on the preflight.

**The filing guard keys on the second lookup, but it's the first one that rules out the case the guard exists for.** When the read at the top of the spike block succeeds and reports no open pause issue, an already-open far-older issue is excluded — the only duplicate the re-read after the jitter protects against is a sibling filed seconds ago, and that is exactly what `run_issue_create_and_reconcile`'s downward probe catches. So a failing re-read should leave the first read's verdict standing rather than suppressing the file.

I ran the preflight against the suite's fake `gh` patched so only the *second* `issue list` fails, with no pause issue open:

```
base fef82e9:  issue create called; "Refused runs are listed in #42; closing it doubles the ceiling."
this PR:       no issue create;     "…could not be read…; if an open `tend-rate-limit` issue exists, closing it doubles the ceiling."
```

Two costs in that window. The run pauses with nothing but the `::error::` on a job nobody opens — the shape the header comment says opening the issue exists to avoid — and the recovery line sends the maintainer after an issue this run's own first read established isn't there. It self-heals on the next refused run, but not this one. The two inline suggestions carry the first read's status forward; with both applied the new test still passes (all 31 do) and the probe above files again.

Worth noting the harness can't express that window: `FAIL_ISSUE_LIST` fails every `issue list`, so the two reads can't be failed independently. Failing from the Nth call rather than all of them would let each be covered.

shared/steps/rate-limit-preflight.sh
max-sixty: Applied in 7393975. The first read now sets `PAUSE_LOOKUP_OK`, and the re-read only ever raises it, so a failure there leaves the first read's verdict standing.

You were right that the harness couldn't express the window: `FAIL_ISSUE_LIST_FROM` now fails from the Nth `issue list` on, so `1` fails both reads and `2` fails only the re-read. `test_rate_limit_still_files_when_only_the_re_read_fails` covers the second case and fails if the re-read is allowed to clear the flag.

shared/steps/rate-limit-preflight.sh
max-sixty: Applied in 7393975 — the re-read now only sets `PAUSE_LOOKUP_OK=true` on success and never clears it, so the failure case falls through to the first read's verdict and the file still happens.

tend-agent: Confirmed, and this covers the first-read half above too. `test_shared_steps.py` passes at 7393975 (32 tests), and mutating this block back to the `PAUSE_LOOKUP_OK=false` form — so the re-read clears the flag again — fails `test_rate_limit_still_files_when_only_the_re_read_fails` and nothing else. So the new test pins that window specifically rather than riding on coverage another test already provided.

`FAIL_ISSUE_LIST_FROM` closes the harness gap: `2` reaches the re-read with the first read's clean "none" standing, which the all-or-nothing flag couldn't express. Both findings resolved from my side.

generator/tests/test_shared_steps.py
```diff
@@ -239,7 +239,16 @@ def test_mark_notification_read_leaves_activity_newer_than_the_run(
...
   issue)
     case "$2" in
-      list) emit "$(cat "$PAUSE_ISSUES_JSON")" ;;
+      list)
+        # Fail from the Nth list call on, so the spike block's two reads can be
+        # failed independently: `1` fails both, `2` leaves the first standing.
+        if [ -n "${FAIL_ISSUE_LIST_FROM:-}" ]; then
+          n=$(( $(cat "$LIST_CALLS" 2>/dev/null || echo 0) + 1 ))
+          echo "$n" > "$LIST_CALLS"
+          if [ "$n" -ge "$FAIL_ISSUE_LIST_FROM" ]; then exit 1; fi
```
tend-agent: `FAIL_ISSUE_LIST_FROM` fails from the Nth call *on*, so "only the first read fails" is the one shape it cannot express — and that is the shape the re-read's own `PAUSE_LOOKUP_OK=true` exists for. An `UNTIL` bound closes it and leaves the three existing `FROM` call sites alone.

```suggestion
        # Fail the list calls in [FROM, UNTIL], so each of the spike block's two
        # reads can be failed on its own: `1` fails both, `2` only the re-read,
        # and `1` with UNTIL=1 only the first — the read the re-read rescues.
        if [ -n "${FAIL_ISSUE_LIST_FROM:-}" ]; then
          n=$(( $(cat "$LIST_CALLS" 2>/dev/null || echo 0) + 1 ))
          echo "$n" > "$LIST_CALLS"
          if [ "$n" -ge "$FAIL_ISSUE_LIST_FROM" ] \
            && [ "$n" -le "${FAIL_ISSUE_LIST_UNTIL:-$n}" ]; then exit 1; fi
```

generator/tests/test_shared_steps.py
```diff
@@ -492,6 +502,56 @@ def test_rate_limit_keeps_its_annotation_when_the_row_cannot_be_appended(
...
+    """
+    rate_limit_env["FAKE_TODAY_POSTS"] = "16"
+    # Nothing open, so the first read is a clean "none"; only the re-read fails.
+    rate_limit_env["FAIL_ISSUE_LIST_FROM"] = "2"
+
+    result = _run_preflight(rate_limit_env)
+
+    assert result.returncode == 1
+    assert any(c.startswith("issue create") for c in _calls(rate_limit_env))
+    assert "could not be read" not in result.stdout
```
tend-agent, line 558: The third case, alongside the two here. It passes as written and fails if the re-read is not allowed to raise the flag, which the two above do not catch.

```suggestion
    assert "could not be read" not in result.stdout


def test_rate_limit_files_when_only_the_first_read_fails(
    rate_limit_env: dict[str, str],
) -> None:
    """A first read the re-read rescues must still file.

    Raising the flag on a good re-read is the other half of keeping the two
    reads apart. Without it a transient failure on the first read alone would
    refuse the run, file nothing, and send the maintainer after issues it says
    it could not read — while the re-read had just established there are none.
    """
    rate_limit_env["FAKE_TODAY_POSTS"] = "16"
    rate_limit_env["FAIL_ISSUE_LIST_FROM"] = "1"
    rate_limit_env["FAIL_ISSUE_LIST_UNTIL"] = "1"

    result = _run_preflight(rate_limit_env)

    assert result.returncode == 1
    assert any(c.startswith("issue create") for c in _calls(rate_limit_env))
    assert "could not be read" not in result.stdout
```

tend-agent (commented):
The applied fix reads right: the first read's verdict now survives a failed re-read, and `test_rate_limit_still_files_when_only_the_re_read_fails` does fail when the re-read is allowed to clear the flag. One gap on the other half of it.

**The re-read's own raise of the flag is load-bearing and uncovered.** Dropping `PAUSE_LOOKUP_OK=true` from the re-read's success branch leaves all 32 tests green — so of the four mutations the description lists as each killing a different test, this fifth one kills none. It is not inert: with the first read failing and the re-read succeeding on an empty list, that mutation refuses the run, files nothing, and annotates

```
Rate limit: bot created 16 items today ... This repo's issues could not be read (see the error above), so none was filed; if an open `tend-rate-limit` issue exists, closing it doubles the ceiling.
```

which is the same shape the last round was about, from the opposite side: the bot pauses with no issue at all, and the recovery line sends the maintainer after issues the re-read had just established aren't there.

The reason it is uncovered is the harness rather than an oversight — `FAIL_ISSUE_LIST_FROM` fails from the Nth call on, so `1` and `2` reach both reads and the first-only case has no encoding. The two suggestions add an `UNTIL` bound and the test it enables; with both applied 33 pass, and the mutation above kills the new one while the existing 32 stay green.

Happy to push the pair as a commit if you'd rather not apply them by hand.

generator/tests/test_shared_steps.py
max-sixty: Applied in 56b6f1d. `FAIL_ISSUE_LIST_UNTIL` bounds the failing range, so `FROM=1 UNTIL=1` now reaches the first read alone; unbounded still means "the list is simply down". `test_rate_limit_files_when_only_the_first_read_fails` covers it and dies when the re-read's `PAUSE_LOOKUP_OK=true` is dropped, with the other 32 staying green.

generator/tests/test_shared_steps.py
max-sixty: Applied in 56b6f1d. `FAIL_ISSUE_LIST_UNTIL` bounds the failing range, so `FROM=1 UNTIL=1` now reaches the first read alone; unbounded still means "the list is simply down". `test_rate_limit_files_when_only_the_first_read_fails` covers it and dies when the re-read's `PAUSE_LOOKUP_OK=true` is dropped, with the other 32 staying green.

tend-agent (approved):

PR #902, opened 2026-08-08
feat(generator): make the Claude prompt prose, so /code-review is reachable

max-sixty:
The built-in `/code-review` carries `disable-model-invocation`, and the `Skill` tool waives that only for a turn whose own user message names the command. The scan behind that waiver skips messages wrapped in `<command-message>` — which is exactly how Claude Code stores a prompt beginning with `/`. Since `default_prompt` built every Claude prompt as `/tend-ci-runner:<skill>`, no eligible message ever existed, so the waiver could never fire in any tend workflow.

The Claude prompt now names its skill in prose and declares `/code-review` alongside the skills the `Skill` tool lists:

```
Run the /tend-ci-runner:review skill for 891. Beyond the skills the Skill tool lists, /code-review is available too.
```

Codex prompts are unchanged, and a test asserts no Codex workflow declares a skill that harness hasn't got.

Only the *leading* position routes a prompt down the slash-command path, so the skill mention keeps its own `/`: it's the invocable form, and it matches the `/code-review` beside it rather than reading as a bare identifier.

The declaring sentence lives in `Config.code_review_notice`, because `mention` builds its prompt in `mention.yaml.j2` rather than through `default_prompt` — one definition, so the two producers can't drift on a token whose whitespace is load-bearing.

Prose relies on the model electing to call `Skill`, where the slash command was routed by the harness. `CLAUDE_DIRECTIVE` in `compose-system-prompt.sh` now covers that without touching the prompt shape: if the prompt names a tend-ci-runner skill, invoke it. The failure it guards is silent, since a run that never loads its skill still exits 0. The clause binds to a tend-ci-runner skill rather than to any skill because `mention`'s prompt names none — it's a task description — so an unconditional version would have pointed a mention run at the `/code-review` the notice had just added.

## What this does and doesn't buy

Nothing consumes the new availability yet. Step 4 of the `review` skill still mandates `/tend-ci-runner:code-review`, so as it stands the declaration is inert — and the prompt now advertises one code-review skill while the skill body mandates another. Resolving that is a separate call, because pointing step 4 at the built-in would cost the Codex harness the second pass it gained when the port landed.

This also changes nothing in CI until the next release: the committed workflows are regenerated by the published generator, not the in-tree one.

<details><summary>Verification, and what is not verified</summary>

Checked on the folded scalar the action actually receives, rather than the serialized file: all seven Claude workflows carry a `/code-review` token with whitespace on both sides, and none leads with a slash.

Verified against a real session log: a prompt of this shape is stored as a plain string user message — not `isMeta`, no `<command-message>` wrapper, no tool result — so it clears all four of the scan's skip rules, and both tokens clear their own regexes with the slash mid-string.

Not verified: the end-to-end unlock at a version that hardcodes the flag. That A/B needed a local 2.1.222 binary, which Claude Code pruned mid-session; the remaining local builds gate the flag open, where the test can't distinguish. The blocked half is confirmed — this PR's own review run reports the built-in returning `disable-model-invocation` under the old slash-command prompt.

The token must clear `(?<!\S)/code-review(?=$|\s)`, where a trailing period or a backtick silently costs the run its second pass. `test_default_prompt_unlocks_code_review` pins that at the rendered layer, along with the no-leading-slash invariant and the unchanged Codex form.

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,8 @@ resolve bot identity. They differ in how the agent runs:
   (`claude -p`) as a non-sudo sandbox user behind a local
   credential-injecting proxy, so the bot token and Anthropic credential
   never enter the agent's environment. Each workflow's prompt names the
-  skill to run in prose, which also keeps the prompt eligible for the
-  waiver that makes the built-in `/code-review` invocable.
+  skill to run in prose rather than as a leading slash command, which
+  keeps the built-in `/code-review` reachable during the run.
 - **Codex harness** — installs the `@openai/codex` CLI on the runner and
   shells out to `codex exec`. An AGENTS.md staged into `$CODEX_HOME`
   teaches Codex to resolve `/tend-ci-runner:NAME` references to the
--- a/README.md
+++ b/README.md
@@ -105,8 +105,9 @@ resolve bot identity. They differ in how the agent runs:
 - **Claude harness** — runs the official `claude` binary headless
   (`claude -p`) as a non-sudo sandbox user behind a local
   credential-injecting proxy, so the bot token and Anthropic credential
-  never enter the agent's environment. Each workflow's prompt is a slash
-  command (`/tend-ci-runner:review`) that loads the matching skill.
+  never enter the agent's environment. Each workflow's prompt names the
+  skill to run in prose, which also keeps the prompt eligible for the
+  waiver that makes the built-in `/code-review` invocable.
 - **Codex harness** — installs the `@openai/codex` CLI on the runner and
   shells out to `codex exec`. An AGENTS.md staged into `$CODEX_HOME`
   teaches Codex to resolve `/tend-ci-runner:NAME` references to the
--- a/generator/src/tend/config.py
+++ b/generator/src/tend/config.py
@@ -203,15 +203,35 @@ class Config:
     def default_prompt(self, skill: str, args: str = "") -> str:
         """Default prompt invoking a tend-ci-runner skill in harness-native syntax.
 
-        Claude resolves `/tend-ci-runner:NAME` as a slash command. Codex resolves
-        `$NAME` as a skill mention (or matches by description); the
+        Codex resolves `$NAME` as a skill mention (or matches by description); the
         `tend-ci-runner` namespace prefix isn't needed at the prompt site because
         skill names within the plugin are unique. `args` is appended raw so
         callers can splice their own placeholders (`{pr_number}` etc.) and run
         the existing replace step.
+
+        Claude's prompt is prose naming the skill, rather than the
+        `/tend-ci-runner:NAME` slash command it also resolves, so that the
+        built-in `/code-review` is reachable. That skill carries
+        `disable-model-invocation`, which the `Skill` tool waives only for a turn
+        whose own user message names the command; a prompt starting with `/` is
+        stored wrapped in `<command-message>` and is skipped by that scan, so no
+        eligible message ever exists. Prose leaves the prompt eligible, and the
+        `/code-review` token in it unlocks the skill for the whole run.
+
+        The token has to survive that scan's regex, `(?<!\\S)/code-review(?=$|\\s)`:
+        whitespace on both sides, so a trailing period or a backtick silently
+        costs the run its second pass. `test_default_prompt_unlocks_code_review`
+        pins it.
         """
-        prefix = f"/tend-ci-runner:{skill}" if self.harness == "claude" else f"${skill}"
-        return f"{prefix} {args}".rstrip()
+        if self.harness != "claude":
+            return f"${skill} {args}".rstrip()
+        invocation = f"Run the tend-ci-runner:{skill} skill"
+        if args:
+            invocation += f" for {args}"
+        return (
+            f"{invocation}. Beyond the skills the Skill tool lists, "
+            "/code-review is available too."
+        )
 
     @classmethod
     def load(cls, path: Path | None = None) -> Config:
--- a/generator/src/tend/config.py
+++ b/generator/src/tend/config.py
@@ -204,8 +204,9 @@ class Config:
     def code_review_notice(self) -> str:
         """Sentence declaring the built-in `/code-review`; empty off the Claude harness.
 
-        Every Claude prompt carries it, including `mention`, whose prompt
-        `mention.yaml.j2` builds rather than `default_prompt`. One definition, so
+        Every generated Claude prompt carries it, including `mention`, whose prompt
+        `mention.yaml.j2` builds rather than `default_prompt` — but not an adopter's
+        `workflows.<name>.prompt`, which replaces the whole prompt. One definition, so
         the two can't drift — and the token's regex (see `default_prompt`) is
         fragile enough that a second copy is a second way to break it.
         """
--- a/generator/src/tend/config.py
+++ b/generator/src/tend/config.py
@@ -200,6 +200,19 @@ class Config:
     sandbox_env: dict[str, str] = field(default_factory=dict)
     sandbox_setup: list[str] = field(default_factory=list)
 
+    @property
+    def code_review_notice(self) -> str:
+        """Sentence declaring the built-in `/code-review`; empty off the Claude harness.
+
+        Every Claude prompt carries it, including `mention`, whose prompt
+        `mention.yaml.j2` builds rather than `default_prompt`. One definition, so
+        the two can't drift — and the token's regex (see `default_prompt`) is
+        fragile enough that a second copy is a second way to break it.
+        """
+        if self.harness != "claude":
+            return ""
+        return "Beyond the skills the Skill tool lists, /code-review is available too."
+
     def default_prompt(self, skill: str, args: str = "") -> str:
         """Default prompt invoking a tend-ci-runner skill in harness-native syntax.
 
@@ -232,10 +245,7 @@ class Config:
         invocation = f"Run the /tend-ci-runner:{skill} skill"
         if args:
             invocation += f" for {args}"
-        return (
-            f"{invocation}. Beyond the skills the Skill tool lists, "
-            "/code-review is available too."
-        )
+        return f"{invocation}. {self.code_review_notice}"
 
     @classmethod
     def load(cls, path: Path | None = None) -> Config:
--- a/generator/src/tend/config.py
+++ b/generator/src/tend/config.py
@@ -209,14 +209,18 @@ class Config:
         callers can splice their own placeholders (`{pr_number}` etc.) and run
         the existing replace step.
 
-        Claude's prompt is prose naming the skill, rather than the
+        Claude's prompt is prose naming the skill, rather than the bare
         `/tend-ci-runner:NAME` slash command it also resolves, so that the
         built-in `/code-review` is reachable. That skill carries
         `disable-model-invocation`, which the `Skill` tool waives only for a turn
-        whose own user message names the command; a prompt starting with `/` is
+        whose own user message names the command; a prompt *starting* with `/` is
         stored wrapped in `<command-message>` and is skipped by that scan, so no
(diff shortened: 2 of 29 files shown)


generator/tests/test_generate.py
```diff
@@ -395,6 +395,37 @@ def test_custom_prompt(tmp_path: Path) -> None:
...
+        assert user_typed_this_turn.search(prompt), (
+            f"{skill} prompt does not carry a bare /code-review token: {prompt!r}"
+        )
+        assert f"/tend-ci-runner:{skill}" in prompt
+
+    codex = Config.load(_minimal_config(tmp_path, "harness: codex\n"))
+    assert codex.default_prompt("review", "{pr_number}") == "$review {pr_number}"
+
+    workflows = {wf.filename: wf for wf in generate_all(cfg)}
+    assert "/code-review" in workflows["tend-review.yaml"].content
```
tend-agent: The invariant this test is defending is the regex, but the rendered-layer assert is a bare substring — so it passes even if rendering leaves `/code-review` adjacent to a non-whitespace character, which is exactly the silent failure the docstring warns about. `tend-review.yaml` is the only workflow whose prompt goes through `_escape_braces`, `''`-quoting and GHA `format()`, and it's also what re-serializes when `workflow_extra` is present, so it's the right place to run the real check.

```suggestion
    assert user_typed_this_turn.search(workflows["tend-review.yaml"].content), (
        "the rendered review workflow lost the bare /code-review token"
    )
```

README.md
```diff
@@ -105,8 +105,9 @@ resolve bot identity. They differ in how the agent runs:
 - **Claude harness** — runs the official `claude` binary headless
   (`claude -p`) as a non-sudo sandbox user behind a local
   credential-injecting proxy, so the bot token and Anthropic credential
-  never enter the agent's environment. Each workflow's prompt is a slash
-  command (`/tend-ci-runner:review`) that loads the matching skill.
+  never enter the agent's environment. Each workflow's prompt names the
+  skill to run in prose, which also keeps the prompt eligible for the
+  waiver that makes the built-in `/code-review` invocable.
```
tend-agent: "the waiver" has no antecedent here — an adopter reading the README hasn't been told there is one, or what it waives. The mechanism detail belongs in the docstring (where it now is); the README just needs the consequence.

```suggestion
  never enter the agent's environment. Each workflow's prompt names the
  skill to run in prose rather than as a leading slash command, which
  keeps the built-in `/code-review` reachable during the run.
```

tend-agent (commented):
Confirmed the blocked half of the A/B from inside this run. This workflow's prompt is still the old `/tend-ci-runner:review 902` form (committed workflows track the released generator), and invoking the built-in here returns `Skill code-review cannot be used with Skill tool due to disable-model-invocation`. So the premise holds under a slash-command prompt. The unlock half stays unverified, as the PR says.

**`tend-mention` doesn't get the token.** Its prompt is built in `mention.yaml.j2`, not by `default_prompt`, so it's the one Claude workflow the change misses. Running the generator over a minimal config, `/code-review` clears `(?<!\S)/code-review(?=$|\s)` in every generated workflow except `tend-mention.yaml`, which has no match at all. That prompt is already prose — it opens with the `${{ steps.delay.outputs.seconds && format(...) || '' }}Before acting,` expression — so it's eligible for the waiver and just carries no token. Mention is the path where a maintainer asks for changes and the bot commits and pushes, so it's plausibly the workflow that wants the second pass most.

**Nothing re-establishes skill loading.** The PR names the trade — the harness used to route the prompt, prose relies on the model electing to call `Skill` — but ships no compensating instruction, and `shared/system-prompt.md` says nothing about running the skill the prompt names. A line there would cover the loss without touching the prompt shape. Worth weighing because the failure is silent: a run that never loads its skill still exits 0 and reads as a successful run.

Two inline suggestions below.

generator/tests/test_generate.py
max-sixty: Right, and it was defending the wrong thing — a substring assert can't fail on the one mutation the docstring warns about. Applied, and extended past the suggestion: the regex now runs over `tend-mention.yaml` as well, since that prompt skips `default_prompt` entirely (your other finding), and a Codex pass asserts no generated workflow declares a skill that harness doesn't have.

I also checked the layer beneath the one you named. The assert reads `wf.content`, which is the serialized file, whereas the action receives the folded scalar — so a token could be whitespace-clean in the file and not in the value. Parsing each workflow and testing the folded `with.prompt`: all seven Claude workflows carry a whitespace-clean token, none leads with a slash. `mention` folds to `}} Beyond the skills…`, so the space comes from the fold rather than the source.

`05744c3`.

README.md
max-sixty: Applied verbatim. You're right that the antecedent was missing — I'd carried "the waiver" over from the docstring, where the mechanism is actually explained, into a paragraph an adopter reads without any of that context. The consequence is the only part that belongs here.

`05744c3`.

max-sixty:
Both body findings taken, in `05744c3`. Thanks for confirming the blocked half from inside the run — that's the half I couldn't reach locally, since the binary that hardcodes the flag was pruned mid-session and every build left gates it open.

**`tend-mention` doesn't get the token.** Correct, and I'd talked myself out of it on the grounds that mention doesn't run a review. Your reason is better: it's the path where the bot commits and pushes on a maintainer's say-so, so it's the one where an unreviewed change actually ships. Rather than copy the sentence into `mention.yaml.j2`, it now lives in `Config.code_review_notice` and both producers read it — a second copy of a string whose whitespace is load-bearing is a second way to break it silently. The property is empty off the Claude harness, so Codex renders unchanged, and a test now asserts no Codex workflow declares the skill.

**Nothing re-establishes skill loading.** Taken. I put it in `CLAUDE_DIRECTIVE` in `compose-system-prompt.sh` rather than `shared/system-prompt.md`: that file is consumed by both engines and its header asks for edits to stay in sync, whereas the loss here is Claude-only, and `CLAUDE_DIRECTIVE` is already the Claude-specific skill-loading directive. It now reads: `Invoke the skill the prompt names — it drives the whole run, and a run that skips it exits clean having done nothing. Use /tend-ci-runner:running-in-ci before starting work.` Your point about the silent failure is the reason the second clause is there rather than a bare instruction.

Your first finding also sent me a layer deeper than either of us checked: the test asserted against `wf.content`, the serialized file, while the action receives the folded scalar — so a token could be whitespace-clean in the file and not in the value. Parsing every generated workflow and testing the folded `with.prompt`: all seven Claude workflows carry a whitespace-clean token and none leads with a slash. `mention` gets its leading space from the fold rather than from the source, which is exactly the case a file-level assert would have missed.

Left alone deliberately, both stated in the description: step 4 still mandates `/tend-ci-runner:code-review`, so the declaration is inert for now — switching it would cost Codex its second pass, which is a maintainer's call — and `.github/workflows/` is not regenerated, per CLAUDE.md on the in-tree generator stamping an unreleased tag.

shared/steps/compose-system-prompt.sh
```diff
@@ -15,7 +15,7 @@
 set -eo pipefail
 
 SHARED="$SYSTEM_PROMPT_FILE"
-CLAUDE_DIRECTIVE="Use /tend-ci-runner:running-in-ci before starting work."
+CLAUDE_DIRECTIVE="Invoke the skill the prompt names — it drives the whole run, and a run that skips it exits clean having done nothing. Use /tend-ci-runner:running-in-ci before starting work."
```
tend-agent: `tend-mention`'s prompt names no tend skill, so this clause has no correct antecedent there — the only skill name it can bind to is the `/code-review` the notice just added. Conditioning it keeps the directive doing its job on the six workflows whose prompt does name a skill, and makes it inert on mention.

```suggestion
CLAUDE_DIRECTIVE="If the prompt names a tend-ci-runner skill, invoke it — it drives the whole run, and a run that skips it exits clean having done nothing. Use /tend-ci-runner:running-in-ci before starting work."
```

generator/src/tend/config.py
```diff
@@ -200,18 +200,52 @@ class Config:
     sandbox_env: dict[str, str] = field(default_factory=dict)
     sandbox_setup: list[str] = field(default_factory=list)
 
+    @property
+    def code_review_notice(self) -> str:
+        """Sentence declaring the built-in `/code-review`; empty off the Claude harness.
+
+        Every Claude prompt carries it, including `mention`, whose prompt
+        `mention.yaml.j2` builds rather than `default_prompt`. One definition, so
```
tend-agent: Not every one: `review`/`triage`/`ci-fix`/`scheduled`/`notifications` all render `wf.prompt or eff.default_prompt(...)`, so an adopter with `workflows.<name>.prompt` set gets no notice and silently loses the second pass, while `mention` gets it unconditionally from the template. Worth saying so here, since this docstring is where someone goes to find out whether the token is present.

```suggestion
        Every generated Claude prompt carries it, including `mention`, whose prompt
        `mention.yaml.j2` builds rather than `default_prompt` — but not an adopter's
        `workflows.<name>.prompt`, which replaces the whole prompt. One definition, so
```

tend-agent (commented):
Both prior findings look addressed. One new thing, from the interaction between the two fixes.

**`CLAUDE_DIRECTIVE` resolves to `/code-review` on a `tend-mention` run.** The directive is unconditional across every Claude workflow, but `tend-mention`'s prompt is a task description that names no tend skill — so after this PR the only slash-command-shaped skill name in it is the notice's `/code-review`. "Invoke the skill the prompt names — it drives the whole run" then points a mention run at a code review. The shape of the failure is a maintainer commenting "@bot rename this to X" and getting a code-review pass as the run's main act.

Rendered the workflows with the in-tree generator over a minimal config, parsed each file, and dumped the folded `with.prompt` value the action actually receives — not the serialized text. Six of the seven start `Run the /tend-ci-runner:<skill> skill …`. `tend-mention.yaml`'s value opens `${{ steps.delay… }}Before acting, check recent comments: …` and closes `…}} Beyond the skills the Skill tool lists, /code-review is available too.`, so `/code-review` is its only `/`-prefixed skill token.

</details>

The inline suggestion narrows the antecedent to a tend-ci-runner skill, which makes the clause a no-op on mention — it already gets its drive-the-run instruction from the prompt body, so nothing is lost there.

Second inline is narrower: the docstring's coverage claim, and the silent gap behind it for an adopter who sets `workflows.<name>.prompt`.

shared/steps/compose-system-prompt.sh
max-sixty: Good catch, and it's a bug I introduced rather than one you found in the original design — the two fixes were fine apart and broke on contact. I wrote the directive against the six workflows I'd just changed and never checked it against the one whose prompt I'd only appended to.

Applied verbatim. Composed the system prompt the way the action does (`SYSTEM_PROMPT_FILE`/`BOT_NAME`/`EXTRA` into the script, reading back the `value` output) to confirm the clause lands as written.

Worth naming why the conditional is the right shape rather than a per-workflow toggle: the directive is composed in a shell step shared by every Claude run, which has no idea which workflow invoked it. Binding the antecedent is the only place the distinction can live.

`52ff90d`.

generator/src/tend/config.py
max-sixty: Correct — "every Claude prompt" was written from the two producers I'd just wired together, and an adopter's `workflows.<name>.prompt` is a third path that replaces the lot. Applied verbatim.

Documenting it is the right call rather than appending the notice to a custom prompt: it's the adopter's prompt, and if theirs opens with a slash the message is `<command-message>`-wrapped and ineligible anyway, so appending would look like it worked while changing nothing.

Added an assert to `test_custom_prompt` so the claim is pinned rather than just written down — a custom prompt renders with no `/code-review` token, which is now what the docstring says.

`52ff90d`.

tend-agent (approved):

PR #903, opened 2026-08-08
feat(review): queue pushes behind an examined-HEAD gate instead of cancelling

max-sixty:
## Problem

A push to a PR under review cancels the in-flight `tend-review` run (`cancel-in-progress: true`), throwing away whatever the session had already read and assembled. On an active PR — e.g. one where a mention-handle session pushes a fix after each review round — this both wastes the partial review work and cold-starts a fresh session per push.

## Solution

Flip review to queue-instead-of-cancel, with a deterministic gate so the queued run is nearly free:

- **`cancel-in-progress: false`** on the review concurrency group: a mid-review push queues a replacement run while the live session keeps going.
- **Fold-in (review skill, new step 9)**: before ending, the session re-fetches the live HEAD and draft state; a push that landed mid-session is judged as an increment (the existing step-1 incremental logic), capped at three fold-ins before handing off to the queued run.
- **Examined-HEAD stamp**: the session stamps every fully-examined commit with a `tend-review/<pr>` commit status — including silent examinations, which previously left no machine-readable trace. The stamp is PR-scoped because one branch can be two open PRs with different bases.
- **Gate** (`review-gate.sh`, inlined into the generated workflow): a queued `synchronize` run exits before booting an agent when the live HEAD is stamped or the PR closed; fails open on API errors. `opened`/`reopened`/`ready_for_review` always run.

Consequences that fall out: the posting-mechanics "HEAD moved — discard" guard becomes a fold-in (nothing else will pick up the discarded review any more, since the queued run gates out), and step 8's pushed fixes are no longer CI-unverified — the session survives its own push and polls the fix to green, dissolving the documented exception to `running-in-ci`'s "a pushed fix is always gated".

Also folded in, from review: the notifications pre-check moves from inline template bash to `notifications-check.sh` consumed the same way as the gate (bot name via `BOT_NAME` env instead of a template substitution), so both workflow pre-checks ship as one mechanism — shellchecked and bash-tested; `review-reviewers` learns that a gate-skipped run is expected silence, not a missing review.

## Testing

Six bash tests drive `review-gate.sh` directly under `bash -e` with a fake `gh` (stamped/unstamped/foreign-context HEADs, non-`synchronize` events make no API calls, closed PRs, fail-open on API errors), plus a workflow-shape test asserting the gate is the first step and every later step is conditioned on it. Snapshot diffs for the notifications extraction verified to be exactly the env-parameterization. 378 tests, `shellcheck`, and `pre-commit` green.

Deferred, deliberately: an adopter `setup:` step carrying its own explicit `if:` still runs on gate-skipped runs (the pre-existing notifications warn-only behavior); AND-composing it with the gate changes semantics for `if: always()`-style steps, so that's left as a follow-up decision.

One property the flip makes strictly worse, accepted deliberately: with no cancel-in-progress, a wedged review session holds queued reviews on that PR until the action's own timeout (~5h50m) instead of being replaced by the next push — the same trade `tend-mention`'s handle job already makes.

The live queue → fold-in → stamp → skip cycle can only be observed after the next release (the action ref pins to a released tag), via tend's own dogfooding.

### DIFF
--- a/generator/src/tend/templates/review-gate.sh
+++ b/generator/src/tend/templates/review-gate.sh
@@ -1,0 +1,57 @@
+# shellcheck shell=bash
+# Pre-check for tend-review: decide whether the agent needs to boot at all.
+#
+# The review job runs without cancel-in-progress, so a push mid-review queues
+# a replacement run while the live session keeps going, folds the push in, and
+# stamps each commit it examined with a `tend-review/<pr>` commit status (review
+# skill, "Stamp examined HEADs"). The concurrency group holds this run until
+# that session ends; by then the live HEAD is usually stamped and there is
+# nothing left to do. Judged against the live PR, not the event payload — a
+# queued run's payload is stale by construction.
+#
+# Inlined into the generated workflow (adopter repos have no copy of this
+# file), so it stays self-contained: env in, GITHUB_OUTPUT out. Any write-scoped
+# actor could forge the stamp to suppress a review; that actor can already
+# cancel the run itself, so the merge gate — not this check — remains the
+# security boundary.
+#
+# env: PR, EVENT_ACTION, GITHUB_REPOSITORY, GITHUB_OUTPUT, GITHUB_TOKEN
+
+# Only `synchronize` can be a stale duplicate of an examination that already
+# happened: `opened` has no prior run, and `reopened` / `ready_for_review`
+# ask for a fresh pass even on a stamped commit.
+if [ "$EVENT_ACTION" != "synchronize" ]; then
+  echo "should_run=true" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
+
+# Fail open on API errors: a redundant agent run beats a silently skipped
+# review.
+if ! PR_INFO=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR" 2>/dev/null); then
+  echo "PR #$PR fetch failed — proceeding without the pre-check"
+  echo "should_run=true" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
+
+STATE=$(echo "$PR_INFO" | jq -r '.state')
+if [ "$STATE" != "open" ]; then
+  echo "PR #$PR is $STATE — skipping"
+  echo "should_run=false" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
+
+# The stamp context carries the PR number: one branch can be two open PRs
+# (same head, different base), and each base means a different diff, so an
+# examination of one must not gate the other.
+HEAD=$(echo "$PR_INFO" | jq -r '.head.sha')
+STAMPED=$(gh api "repos/$GITHUB_REPOSITORY/commits/$HEAD/status?per_page=100" 2>/dev/null \
+  | jq --arg ctx "tend-review/$PR" \
+    '[.statuses[]? | select(.context == $ctx and .state == "success")] | length' \
+  || echo 0)
+if [ "${STAMPED:-0}" -gt 0 ]; then
+  echo "HEAD $HEAD already examined (tend-review/$PR) — skipping"
+  echo "should_run=false" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
+
+echo "should_run=true" >> "$GITHUB_OUTPUT"
--- a/generator/src/tend/templates/review-gate.sh
+++ b/generator/src/tend/templates/review-gate.sh
@@ -26,14 +26,16 @@ if [ "$EVENT_ACTION" != "synchronize" ]; then
 fi
 
 # Fail open on API errors: a redundant agent run beats a silently skipped
-# review.
-if ! PR_INFO=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR" 2>/dev/null); then
+# review. The parse belongs inside the guard — GitHub sometimes returns an
+# HTML error page with a 200 during a blip, so a zero `gh` exit doesn't mean
+# the body is JSON, and an unguarded `jq` under the run block's `bash -e`
+# would fail the step (fail-closed) instead.
+if ! PR_INFO=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR" 2>/dev/null) \
+  || ! STATE=$(echo "$PR_INFO" | jq -re '.state'); then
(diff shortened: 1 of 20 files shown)


generator/src/tend/templates/review-gate.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Fail open on API errors: a redundant agent run beats a silently skipped
+# review.
+if ! PR_INFO=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR" 2>/dev/null); then
+  echo "PR #$PR fetch failed — proceeding without the pre-check"
+  echo "should_run=true" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
+
+STATE=$(echo "$PR_INFO" | jq -r '.state')
```
tend-agent: The “fail open on API errors” guarantee only covers a non-zero `gh` exit. A 200 carrying an HTML error page — the blip `notifications-check.sh` spends fifteen lines retrying, in this same PR — gets past it, and `jq -r '.state'` then exits 5 under `bash -e`, killing the step before `should_run` is ever written. Every later step is `if: steps.gate.outputs.should_run == 'true'`, so the review is skipped and the job is red.

Folding the parse into the same guard covers both (`jq -re` also fails open on a body with no `.state`):

```suggestion
if ! PR_INFO=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR" 2>/dev/null) \
  || ! STATE=$(echo "$PR_INFO" | jq -re '.state'); then
  echo "PR #$PR fetch failed — proceeding without the pre-check"
  echo "should_run=true" >> "$GITHUB_OUTPUT"
  exit 0
fi

```

generator/tests/test_shared_steps.py
```diff
@@ -876,6 +878,139 @@ def test_rate_limit_reopens_rather_than_refiling(
...
+REVIEW_GATE = REPO_ROOT / "generator" / "src" / "tend" / "templates" / "review-gate.sh"
+
+FAKE_GH_REVIEW_GATE = """#!/usr/bin/env bash
+printf '%s\\n' "$*" >> "$GH_CALLS"
+case "$2" in
+  repos/*/pulls/*)
+    [ -n "${FAIL_PR:-}" ] && exit 1
+    cat "$PR_JSON"
+    ;;
+  repos/*/commits/*/status\?per_page=100)
```
tend-agent: `\?` isn't a recognized Python escape, so this raises a new `SyntaxWarning: invalid escape sequence '\?'` on every `pytest` run (and is slated to become a `SyntaxError`). Doubling the backslash keeps the bash `case` pattern byte-identical — `r"""` would not, since the `printf '%s\\n'` above it relies on the escape being processed.

```suggestion
  repos/*/commits/*/status\\?per_page=100)
```

tend-agent (commented):
The queue-instead-of-cancel design holds up on the races I traced: the gate judging the *live* HEAD rather than the (stale by construction) event payload is what makes queue-depth-1 replacement loss-free, and the stamp being PR-scoped, written last, and skipped on close all line up. The notifications extraction is a clean env-parameterization — I diffed it against the old inline block and the only substantive change is `<<cfg.bot_name>>` → `$BOT_NAME`.

One correctness issue on the gate's fail-open, plus a nit. Both inline.

On the gate: `review-gate.sh` fails open on a non-zero `gh` exit, but not on the failure mode this PR's sibling script documents at length — GitHub returning an HTML error page with a 200 during an API blip. `gh api` exits 0, `PR_INFO` holds HTML, and `STATE=$(echo "$PR_INFO" | jq -r '.state')` exits 5 under the `bash -e` Actions gives a `run:` block. Because it's the first step and every later step is `if: steps.gate.outputs.should_run == 'true'`, `should_run` is never written, the whole review is skipped, and the job goes red — fail-*closed*, and a shape `review-reviewers` will read as a missing review rather than expected silence. Confirmed locally: `jq -r '.state'` on an HTML body exits 5 and the following line never runs.

<details><summary>Repro</summary>

```
$ cat t.sh
PR_INFO='<html>oops</html>'
STATE=$(echo "$PR_INFO" | jq -r '.state')
echo "reached: STATE=$STATE"
$ bash -e t.sh; echo "exit=$?"
jq: parse error: Invalid numeric literal at line 2, column 0
exit=5
```

The suggested form was checked against all three inputs — `gh` non-zero, HTML-200, and valid JSON — and against a `{"head":{}}` body with no `.state` (which `jq -re` also treats as fail-open).

</details>

`HEAD=$(… jq -r '.head.sha')` further down is fine as-is: by then the body has parsed once.

Not blocking, but worth a line in the PR body: with `cancel-in-progress: false` and no `timeout-minutes` on the job, a wedged review session now holds every later review on that PR for up to the action's 5h50m `timeout_seconds` (plus job overhead), where the old behavior replaced it on the next push. `tend-mention`'s handle job already accepts that trade, so this looks deliberate rather than missed — flagging only because it's the one property the queue flip makes strictly worse.

generator/src/tend/templates/review-gate.sh
max-sixty: Applied in 46b7e47 — the parse is folded into the guard with `jq -re` (which also fails open on a JSON body missing `.state`), and a new test drives the HTML-200 shape through the gate.

generator/tests/test_shared_steps.py
max-sixty: Fixed in 46b7e47 (doubled backslash; `-W error` run confirms the suite is warning-free).

generator/src/tend/templates/review-gate.sh
tend-agent: Confirmed at 46b7e47 — the folded guard covers the HTML-200 shape, and `jq -re` fails open on `{"state": null}` too (`-e` exits 1 on null). The `.head.sha` parse below stays safe for the same reason as before: by then the body has parsed once.

tend-agent (approved):

generator/tests/test_shared_steps.py
tend-agent: Confirmed — the doubled backslash still renders one backslash into the fake's `case` pattern (`repos/*/commits/*/status\?per_page=100`), so the match is unchanged, and the module compiles clean with warnings as errors. Full suite at 46b7e47: 379 passed under `-W error::SyntaxWarning`.

PR #904, opened 2026-08-08
revert(generator): restore the slash-command prompt, and record why

max-sixty:
Reverts #902.

Making the Claude prompt prose existed only so the `Skill` tool's waiver could fire and the built-in `/code-review` become reachable during a run. Reading the built-in out of the 2.1.226 binary says it isn't worth reaching.

**The two share their method almost verbatim.** Same ten angles (A line-by-line, B removed-behavior, C cross-file, D language-pitfall, E wrapper/proxy, plus Reuse, Simplification, Efficiency, Altitude, Conventions), the same three-verdict verify with the same "PLAUSIBLE by default" rule, and the same Phase 3 sweep down to the same focus list.

**But the built-in isn't one prompt — it's a model × effort matrix, and Opus 5 gets a deliberately minimal cell.** For `claude-opus-5`, both `medium` and `high` resolve to a cell whose entire prompt is `minimal prompt → single careful diff pass → ≤15 findings`: no angles, no verify, no sweep, no fan-out. It carries `measuredExternal: true`, so that is a measured choice rather than an oversight. The angle-based versions are what `xhigh`/`max` and the `default` model row get.

tend CI runs `model: opus` and sets no effort, so it lands on whichever row model-id normalization yields — and both candidates point the same way. The `default` row at medium is 3+5 angles → verify → ≤8; the `claude-opus-5` row is the minimal single pass. The port's core-logic band is 5+5 angles → verify → sweep → ≤10, broader than either.

Two further costs: the built-in reports through `ReportFindings`, which `claude/action.yaml` does not allowlist, and the Codex harness has no built-in at all.

So the port stays, and the prompt goes back to the slash command the harness routes — which also retires the compensating `CLAUDE_DIRECTIVE` clause the prose form needed.

## What stays behind

`TODO.md` gains an entry holding those measurements, so the question can be reopened without re-reading a binary. It lives there rather than in the skill because it argues about whether the port should exist, which is maintainer work — a skill body is loaded into every session that runs the skill, so a note there is carried by every review run and acted on by none.

Its preamble changes too. It used to say the built-in "cannot be reached with the `Skill` tool", which #902 made false — and which was over-strong even before, since the gate is `disableModelInvocation && !userTypedThisTurn` and a prompt naming `/code-review` is exactly what lifts it. It now names the two durable reasons the port exists: Codex has no equivalent, and no tend prompt lifts the Claude gate.

<details><summary>How the built-in was read, and what is not verified</summary>

2.1.220 — the version `claude/action.yaml` pins, and the one the port was taken from — has been pruned from local disk, so the comparison is against 2.1.226. The angle, verify and sweep texts still match the port near-verbatim, so the method itself has not drifted; what moved is the parameterization around it.

Not verified: which matrix row `model: opus` resolves to at runtime, since that depends on model-id normalization inside the binary. The conclusion holds under both candidate rows, which is why it did not need pinning down.

</details>

### DIFF
--- a/plugins/tend-ci-runner/skills/code-review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/code-review/SKILL.md
@@ -10,16 +10,6 @@ metadata:
 
 A structured pass over a diff that returns ranked findings. Ported from Claude Code's built-in `/code-review`, read out of the 2.1.220 binary `claude/action.yaml` pins. The Codex harness has no built-in equivalent, and on Claude the built-in carries `disable-model-invocation`, which no tend prompt lifts. This copy is tend-owned: reachable from the model on either harness.
 
-<!-- TODO(2026-08-08): decide whether the built-in `/code-review` should replace this
-     port. Making the prompt prose, so the `Skill` tool's waiver fires and the built-in
-     becomes reachable, was tried and reverted (#902). Measured against the 2.1.226
-     binary, the two share their angles, verify pass and sweep near-verbatim — but the
-     built-in resolves per model and effort, and `claude-opus-5` at medium/high renders
-     a minimal single-pass cell with no verify and no sweep, then reports through
-     `ReportFindings`, which `claude/action.yaml` does not allowlist. Codex has no
-     built-in at all. The port stays until one of those changes. -->
-
-
 **Return findings; don't act on them.** No review, comment, commit, or artifact from this skill — the caller folds the findings into its own single review.
 
 ## Phase 0 — Gather the diff
--- a/plugins/tend-ci-runner/skills/code-review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/code-review/SKILL.md
@@ -8,7 +8,17 @@ metadata:
 
 # Code review
 
-A structured pass over a diff that returns ranked findings. Ported from Claude Code's built-in `/code-review` (read out of the 2.1.220 binary), which carries `disable-model-invocation` and so cannot be reached with the `Skill` tool. This copy is tend-owned: reachable from the model, and it runs on both harnesses.
+A structured pass over a diff that returns ranked findings. Ported from Claude Code's built-in `/code-review`, read out of the 2.1.220 binary `claude/action.yaml` pins. The Codex harness has no built-in equivalent, and on Claude the built-in carries `disable-model-invocation`, which no tend prompt lifts. This copy is tend-owned: reachable from the model on either harness.
+
+<!-- TODO(2026-08-08): decide whether the built-in `/code-review` should replace this
+     port. Making the prompt prose, so the `Skill` tool's waiver fires and the built-in
+     becomes reachable, was tried and reverted (#902). Measured against the 2.1.226
+     binary, the two share their angles, verify pass and sweep near-verbatim — but the
+     built-in resolves per model and effort, and `claude-opus-5` at medium/high renders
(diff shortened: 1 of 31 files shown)


plugins/tend-ci-runner/skills/code-review/SKILL.md
```diff
@@ -8,7 +8,17 @@ metadata:
...
+
+<!-- TODO(2026-08-08): decide whether the built-in `/code-review` should replace this
+     port. Making the prompt prose, so the `Skill` tool's waiver fires and the built-in
+     becomes reachable, was tried and reverted (#902). Measured against the 2.1.226
+     binary, the two share their angles, verify pass and sweep near-verbatim — but the
+     built-in resolves per model and effort, and `claude-opus-5` at medium/high renders
+     a minimal single-pass cell with no verify and no sweep, then reports through
+     `ReportFindings`, which `claude/action.yaml` does not allowlist. Codex has no
+     built-in at all. The port stays until one of those changes. -->
+
```
tend-agent: This note is maintainer-facing — it argues about whether the port should exist — but it lives in a file that is loaded into every session running the skill, so every review run carries it and no run acts on it. `TODO.md` is the repo's stated home for exactly this: "Deferred work and unimplemented options. Each entry should justify the cost of building it if revisited."

It also runs into two of CLAUDE.md's own skill-authoring rules: "Be brief. Skills are loaded into every relevant session — extra prose is overhead. Lead with the rule or recipe; cut motivation, anecdotes, and historical context unless required to apply the rule", and "No specific past-case references. Don't cite individual past PRs, issues, or commits as supporting precedent for a rule (e.g. `[#123] tried X and was closed`)" — the `(#902)` citation is that shape. The measurements are worth keeping; the argument is that they're worth keeping somewhere a run doesn't read.

The preamble rewrite above it is the part a run does need, and it stands on its own. The suggestion drops the block here (it also removes the stray double blank line it leaves behind at 21–22) — say the word and I'll push the other half as a `TODO.md` entry so nothing is lost.

```suggestion
```

tend-agent (commented):
The revert itself is faithful: for every file #902 touched, the tree here is byte-identical to its pre-#902 state, and the merge with #903 is clean on substance too — #903 doesn't touch `config.py`, `mention.yaml.j2`, or `compose-system-prompt.sh`, and its `test_generate.py` addition sits well below the removed block. The generator suite passes on the merged tree (378 passed). Nothing dangles: `code_review_notice` has no remaining references anywhere, `re` is still used elsewhere in `test_generate.py`, and the Codex `$review {pr_number}` shape that the deleted test asserted is still pinned by the codex regtests.

One observation, on where the decision record landed rather than on the revert.

plugins/tend-ci-runner/skills/code-review/SKILL.md
max-sixty: Agreed, and the sharper half of it is that no run could act on the note — a skill body reaches only the sessions that run the skill, which are exactly the sessions with no standing to decide whether the skill should exist. Carried by every review, actionable by none.

Moved to `TODO.md` in `4862bc4`, where the entry can justify its own cost at proper length instead of compressing into a comment, and the `(#902)` citation drops out with it. The preamble stays, as you say — it's the part a run needs.

> say the word and I'll push the other half as a `TODO.md` entry

Thanks — already written, so nothing to pick up.

tend-agent (approved):

PR #910, opened 2026-08-09
fix(install): give the codex CLI install the same retry window

### LINKED ISSUE #909: codex: the CLI install is the one pre-agent installer not on the shared retry window
`install-claude-binary.sh` and `install-proxy-uv.sh` both retry their CDN fetch on the shared window in `shared/steps/lib/retry.sh` (#906, #907, #908). The Codex harness has a third pre-agent installer that isn't on it: [`codex/action.yaml:120-124`](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/codex/action.yaml#L120-L124) runs `npm install -g "@openai/codex@${{ inputs.codex_version }}"` bare — one attempt, no timeout, no backoff.

The failure mode is the one the shared window exists for. The step runs ahead of `Run Codex`, so an npm registry blip costs the whole run: the job goes red having done none of the work the trigger asked for, exactly as a CDN blip on claude.ai did before #906. Five steps sit ahead of the agent in that action (security preflight, rate-limit preflight, auth validation, the CLI install, plugin install, AGENTS.md staging), and this is the only one that reaches out to a third-party registry.

## Proposed fix

`retry_install` can't be called from where the install currently lives — it's an inline `run:` block, and the lib is only reachable from a script under `shared/steps/`. So the change is an extraction first, matching what `install-proxy-uv.sh` already does:

```bash
# shared/steps/install-codex-cli.sh
set -euo pipefail
SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
# shellcheck source=lib/retry.sh
. "${SCRIPT_DIR}/lib/retry.sh"

retry_install "codex ${CODEX_VERSION}" "npm install -g '@openai/codex@${CODEX_VERSION}'"
codex --version
```

with the action step becoming a `run: bash "${{ github.action_path }}/../shared/steps/install-codex-cli.sh"` plus `env: CODEX_VERSION: ${{ inputs.codex_version }}`.

## One thing to settle first

`retry_install` hardcodes `timeout 60`, which suits a `curl | sh` that fetches a single installer script. A global npm install resolves and unpacks a dependency tree, and I haven't measured how long `@openai/codex` takes on a cold runner — if it can legitimately exceed 60s, dropping it under the current lib would convert slow installs into failures rather than riding out blips. Whoever picks this up should time the install on a runner and either confirm 60s is generous or add an optional third parameter (`retry_install LABEL COMMAND [TIMEOUT]`, defaulting to 60) before wiring codex in.

Raised in the review of #908, where the shared lib lands.

tend-agent:
## Problem

`install-claude-binary.sh` and `install-proxy-uv.sh` both ride out a CDN blip on the shared window in `shared/steps/lib/retry.sh`. The codex harness's third pre-agent installer wasn't on it: `npm install -g "@openai/codex@..."` ran as a bare inline `run:` block — one attempt, no timeout, no backoff. It is the codex action's only reach to a third party, and it sits ahead of `Run Codex`, so a registry blip took the whole run red having done none of the work the trigger asked for.

## Solution

Extracted the install to `shared/steps/install-codex-cli.sh`, sourcing `lib/retry.sh` the way `install-proxy-uv.sh` does; the action step becomes a `bash .../install-codex-cli.sh` with `CODEX_VERSION` in `env:`. The version-pin rationale stays in `codex/action.yaml` next to the input it explains.

### The 60s timeout stands

#909 flagged that `retry_install` hardcodes `timeout 60`, sized for a `curl | sh` that fetches one installer script, and asked whether a global npm install could legitimately exceed it — in which case dropping codex under the current lib would convert slow installs into failures.

Measured on a GitHub-hosted runner with an empty npm cache, three cold installs of `@openai/codex@0.131.0-alpha.22`: **3.08s, 2.60s, 2.63s** (218 MB unpacked, 2 packages). The package ships a prebuilt per-platform binary, so there is no dependency tree to resolve and no toolchain to run — the shape that would justify a longer window isn't there. 60s leaves ~20x headroom, so no optional third parameter on `retry_install`; the measurement and its date are recorded in the script's header so a future bump can re-check it rather than re-derive it.

## Testing

Four tests in `generator/tests/test_shared_steps.py` mirroring the claude and uv suites, with a fake `npm` answering `503 Service Unavailable` on a schedule: rides out a four-failure burst, backs off 5/10/20/40 with jitter, reddens after five attempts, and installs first-try without sleeping. The happy-path test asserts the installed version echoes back, which keeps the pin honest — the step has to install the `codex_version` it was handed, not npm's `latest`.

Against the bare install the burst, backoff and redden tests fail (one attempt, no `after 5 attempts` message); against the new script all 52 tests in the file pass. `shellcheck -S warning` and the rest of pre-commit are clean on the touched files.

## Stacking

Based on #908, where `lib/retry.sh` lands — it will retarget to `main` when that merges. The diff against #908 is this commit alone.

---

Closes #909

### DIFF
--- a/codex/action.yaml
+++ b/codex/action.yaml
@@ -110,18 +110,18 @@ runs:
           exit 1
         fi
 
-    # Install the Codex CLI from npm. `@openai/codex` ships a prebuilt
-    # per-platform binary, so this is ~3 s with no Rust toolchain. We pin
-    # an explicit version (`codex_version`, default a `0.131.0-alpha.*`)
-    # because tend needs `codex plugin add` (PR #21396, first in
-    # `rust-v0.131.0-alpha.17`) and as of 2026-05-15 npm's `latest`
-    # (0.130.0) still predates it — only the `alpha` dist-tag carries it.
-    # Bump the default once a stable npm release ships PR #21396.
+    # Install the Codex CLI from npm, on the retry window the other pre-agent
+    # installers share (shared/steps/lib/retry.sh). We pin an explicit version
+    # (`codex_version`, default a `0.131.0-alpha.*`) because tend needs
+    # `codex plugin add` (PR #21396, first in `rust-v0.131.0-alpha.17`) and as
+    # of 2026-05-15 npm's `latest` (0.130.0) still predates it — only the
+    # `alpha` dist-tag carries it. Bump the default once a stable npm release
+    # ships PR #21396.
     - name: Install Codex CLI
       shell: bash
-      run: |
-        npm install -g "@openai/codex@${{ inputs.codex_version }}"
-        codex --version
+      run: bash "${{ github.action_path }}/../shared/steps/install-codex-cli.sh"
+      env:
+        CODEX_VERSION: ${{ inputs.codex_version }}
 
     # Install the tend-ci-runner plugin from the action's own checkout.
     # `github.action_path` is .../<runner-cache>/max-sixty/tend/<ref>/codex/;
--- a/generator/tests/test_shared_steps.py
+++ b/generator/tests/test_shared_steps.py
@@ -1390,8 +1390,8 @@ def _run_install(env: dict[str, str]) -> subprocess.CompletedProcess[str]:
     )
 
 
-def _attempts(env: dict[str, str]) -> int:
-    return int(Path(env["CURL_ATTEMPTS"]).read_text().strip())
+def _attempts(env: dict[str, str], key: str = "CURL_ATTEMPTS") -> int:
+    return int(Path(env[key]).read_text().strip())
 
 
 def _sleeps(env: dict[str, str]) -> list[int]:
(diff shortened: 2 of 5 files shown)


tend-agent (commented):
The extraction is faithful and the retry contract is now identical across the three installers. Passing `CODEX_VERSION` through `env:` instead of interpolating `${{ inputs.codex_version }}` into the `run:` body is a real improvement over the block it replaces. Tests pass locally (52 in `test_shared_steps.py`), `shellcheck -S warning` is clean on the touched scripts.

One gap the new `lib/retry.sh` header surfaces rather than causes: it now enumerates the pre-agent third-party reaches as three, but there is a fourth. `proxy/setup-sandbox.sh` warms the uvx cache with `"$UVX" --from "$MITMPROXY" mitmdump --version` (under the "Warm the uvx cache first" comment) before it backgrounds the proxy — that resolves `mitmproxy==$MITMPROXY_VERSION` from PyPI on a cold `UV_CACHE_DIR`, under `set -euo pipefail`, ahead of the agent step in the Claude action. A cache miss (first run on a new pin, or an evicted `actions/cache` entry) plus a PyPI blip reddens the run having done none of the work the trigger asked for — the same failure this stack exists to close. It's the wrong PR to fix it in (different harness, different issue, and CLAUDE.md's atomic-PR rule), but happy to open a follow-up that puts that resolve on the window and updates the header's enumeration.

Non-blocking on the test suite: `FAKE_NPM` reads the package spec off `$3` positionally, so it silently mis-parses if the install command ever grows a flag. The version assertion in the happy-path test is what would catch it, which is enough — just worth knowing the coupling is there.

tend-agent:
Both points from the review, as author.

**The fourth reach.** Confirmed and out of scope here: `proxy/setup-sandbox.sh:319` resolves `mitmproxy==$MITMPROXY_VERSION` under `set -euo pipefail` three steps ahead of `Run Claude`, with nothing riding out a PyPI blip. It's tracked in #911, which also carries the header's three-to-four enumeration update — that sentence should change in the commit that actually puts the resolve on the window, not ahead of it. #911 leaves the timeout open too: unlike npm's prebuilt binary (2.6-3.1 s measured here), a cold mitmproxy resolve pulls a real dependency tree, so whether 60 s is generous there needs timing on a cleared cache, and that is where the optional third `retry_install` parameter #909 raised may earn its place.

**`FAKE_NPM`'s positional read.** Removed rather than left as a known coupling — `6420622` reads the spec off the last argument (`spec=${*: -1}`) instead of `$3`, so a flag added ahead of it can't make the fake echo `-g` as the installed version. 52 tests in `test_shared_steps.py` still pass; ruff and typos clean.

Still based on #908 and will retarget to `main` when that lands.

max-sixty:

PR #912, opened 2026-08-09
fix(proxy): put the mitmproxy cache warm on the shared retry window

### LINKED ISSUE #911: proxy: the mitmproxy cache warm is the fourth pre-agent reach, and it's not on the retry window
`lib/retry.sh`'s header (as of #910) enumerates the pre-agent third-party reaches as three — `install-claude-binary.sh` (claude.ai), `install-proxy-uv.sh` (astral.sh), `install-codex-cli.sh` (registry.npmjs.org). There is a fourth, and it is not on the window.

[`proxy/setup-sandbox.sh:319`](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/proxy/setup-sandbox.sh#L319) warms the uvx cache before backgrounding the proxy:

```bash
MITMPROXY="mitmproxy==${MITMPROXY_VERSION}"
UVX="${TEND_UV_DIR}/uvx"
"$UVX" --from "$MITMPROXY" mitmdump --version >/dev/null
```

On a cold `UV_CACHE_DIR` that resolves and downloads mitmproxy from PyPI. The script runs under `set -euo pipefail` ([line 33](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/proxy/setup-sandbox.sh#L33)), so a failed resolve exits non-zero, and the calling step — `Set up credential-isolation sandbox`, [`claude/action.yaml:205`](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/claude/action.yaml#L205) — runs three steps ahead of `Run Claude` at [line 273](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/claude/action.yaml#L273). A PyPI blip there costs the whole run having done none of the work the trigger asked for: exactly the failure #906/#907/#908/#910 exist to close, on the one remaining reach.

## Why the cache doesn't cover it

The `actions/cache` key is version-scoped — `tend-mitmproxy-${{ runner.os }}-${{ inputs.mitmproxy_version }}` ([`claude/action.yaml:203`](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/claude/action.yaml#L203)) — so a miss isn't a rare event, it's scheduled. Every `mitmproxy_version` bump misses on the first run for every workflow, and the weekly bump PR is the moment a PyPI blip is most expensive: it would redden the PR that carries the security-relevant pin. Evicted entries (7-day idle, or the repo's 10 GB cap) miss the same way.

## Proposed fix

`retry_install` is sourceable from `proxy/setup-sandbox.sh` the way the installer scripts source it — the lib is at `shared/steps/lib/retry.sh` and the script already resolves `${ACTION_PATH}`. So:

```bash
. "${ACTION_PATH}/shared/steps/lib/retry.sh"
retry_install "mitmproxy ${MITMPROXY_VERSION}" \
  "'$UVX' --from '$MITMPROXY' mitmdump --version >/dev/null"
```

Two things to check when picking it up, neither settled here:

- **The 60s timeout.** Unlike the npm install measured in #910 (2.6-3.1s, prebuilt binary), a cold mitmproxy resolve pulls a real dependency tree; `claude/action.yaml:196-197` calls it a "~20s resolve" but that figure isn't dated and predates several version bumps. Time it on a runner with a cleared cache before assuming 60s is generous. If it isn't, that's the case for the optional third parameter (`retry_install LABEL COMMAND [TIMEOUT]`) that #909 raised and #910 declined — npm didn't need it; this might.
- **The backgrounded `mitmdump` on the next line is a separate question.** It's the same PyPI reach on a cache miss, but it's `nohup`'d with a readiness wait after it, so a retry there isn't the same shape and shouldn't be folded in without thinking about the wait.

Also update `lib/retry.sh`'s header enumeration from three reaches to four.

Raised in the review of #910.

tend-agent:
## Problem

`lib/retry.sh` covers three pre-agent third-party reaches. There is a fourth: `proxy/setup-sandbox.sh` warms the uvx cache before backgrounding the proxy, and on a cache miss that resolves and downloads mitmproxy from PyPI. The script runs under `set -euo pipefail`, and its step — `Set up credential-isolation sandbox` — sits three steps ahead of `Run Claude`, so a PyPI blip there loses the whole run having done none of the work the trigger asked for.

The miss is scheduled rather than rare: the `actions/cache` key is version-scoped (`tend-mitmproxy-${{ runner.os }}-${{ inputs.mitmproxy_version }}`), so every `mitmproxy_version` bump misses on each workflow's first run — which is exactly when a lost run is most expensive, since it reddens the weekly bump PR that carries the security-relevant pin.

## Solution

Source the shared lib from `setup-sandbox.sh` and wrap the warm in `retry_install`. The source sits up top, next to the env checks, so a wrong path fails before the user creation and the workspace chown rather than mid-setup.

Stacked on #910 — the lib it sources lands there.

## The two questions #911 left open

**Is 60s enough for a cold mitmproxy resolve?** Yes, with room to spare. The `~20s resolve` figure in `claude/action.yaml` was undated and is now wrong. Measured on a GitHub-hosted runner (this triage job), 2026-08-09, with `uv 0.12.3` and the pinned `mitmproxy==12.2.3`:

| | elapsed |
|---|---|
| cold `UV_CACHE_DIR` (43 packages, ~100MB downloaded) | 2.28 / 2.26 / 2.27 / 2.24 s |
| warmed cache | 0.516 / 0.514 / 0.515 s |

So no third `TIMEOUT` parameter is needed — the case #909 raised and #910 declined stays declined. This PR also corrects the stale figure in the action comment, keeping the cache's real justification (not re-pulling ~100MB from PyPI on every run) rather than the seconds.

**Does the backgrounded `mitmdump` need its own retry?** No — retrying the warm already covers it. The launch shares `UV_CACHE_DIR`, and a cache warmed by this line serves the same `--from` invocation with no network at all:

```
$ UV_CACHE_DIR=$D uvx --from 'mitmproxy==12.2.3' mitmdump --version   # warm
$ UV_CACHE_DIR=$D uvx --offline --from 'mitmproxy==12.2.3' mitmdump --version
Mitmproxy: 12.2.3
```

Once the warm succeeds, PyPI is off the launch path, so the `nohup`-plus-readiness-wait shape never has to be retried. That's recorded in the comment at the warm.

## Testing

Two tests in `generator/tests/test_shared_steps.py`, both failing before the change:

- `test_setup_sandbox_resolves_the_shared_retry_window` runs the script's own `. "${ACTION_PATH}/…"` line (matched by content, not position) with `ACTION_PATH` set the way the action sets it, and asserts `retry_install` becomes a defined function. A path that doesn't resolve wouldn't degrade to an unretried warm — under `set -euo pipefail` it loses *every* run, not just the ones that hit a blip.
- `test_setup_sandbox_warms_the_mitmproxy_cache_through_the_window` asserts the warm is an argument to `retry_install` rather than a bare command.

The second is a call-site assertion rather than the behavioural suite the three installers get. `setup-sandbox.sh` creates the sandbox user, chowns the workspace and installs a CA before reaching the warm, so faking a path to that line would assert more about the fakes than about the script; what the window itself does is already covered by the installer suites. Flagging the asymmetry rather than hiding it.

`pre-commit run` (shellcheck, ruff, actionlint) and the full 393-test generator suite pass.

---

Closes #911 — automated triage

### DIFF
--- a/claude/action.yaml
+++ b/claude/action.yaml
@@ -194,8 +194,10 @@ runs:
     # plugins, and the run all execute as the sandbox user. Cache the proxy's
     # mitmproxy download (a sizeable dependency) in an isolated uv cache dir,
     # keyed on the pinned version. First run populates; every run after restores
-    # in seconds instead of a cold ~20s resolve. Scoped to the proxy's uv; a
-    # skill that installs its own uv gets that uv's default cache, untouched.
+    # instead of pulling ~100MB from PyPI again (measured 2026-08-09 on a
+    # GitHub-hosted runner: 2.2s cold, 0.5s restored — the saving is the reach,
+    # more than the seconds). Scoped to the proxy's uv; a skill that installs
+    # its own uv gets that uv's default cache, untouched.
     - name: Cache mitmproxy
       uses: actions/cache@v4
       with:
--- a/generator/tests/test_shared_steps.py
+++ b/generator/tests/test_shared_steps.py
@@ -1691,3 +1691,79 @@ def test_install_codex_cli_installs_first_try_without_sleeping(
     )
     assert _sleeps(codex_cli_env) == [], _sleeps(codex_cli_env)
     assert "codex-cli 0.131.0-alpha.22" in result.stdout, result.stdout
+
+
+# ---------------------------------------------------------------------------
+# proxy/setup-sandbox.sh — the mitmproxy cache warm, the reach that isn't an
+# installer
+# ---------------------------------------------------------------------------
+
+SETUP_SANDBOX = REPO_ROOT / "proxy" / "setup-sandbox.sh"
+
+
+def _logical_lines(path: Path) -> list[str]:
+    """Script lines with backslash continuations folded, so a call reads as one."""
+    out: list[str] = []
+    buf = ""
+    for raw in path.read_text().splitlines():
+        stripped = raw.strip()
+        if stripped.endswith("\\"):
+            buf += stripped[:-1].rstrip() + " "
+            continue
+        out.append(buf + stripped)
+        buf = ""
+    if buf:
+        out.append(buf)
+    return out
+
+
+def test_setup_sandbox_resolves_the_shared_retry_window() -> None:
+    """The lib path has to resolve from ACTION_PATH the way the action passes it.
+
+    The action points ACTION_PATH at the repo root, and the script runs under
+    `set -euo pipefail`, so a path that doesn't resolve doesn't degrade to an
+    unretried warm — it exits the step, and then every run is lost, not only
+    the ones that hit a blip. Sourcing the script's own line (matched by
+    content) is what makes this a check on the shipped path rather than on a
+    copy of it.
+    """
+    sources = [
+        line
+        for line in _logical_lines(SETUP_SANDBOX)
+        if line.startswith(". ") and "lib/retry.sh" in line
+    ]
+    assert len(sources) == 1, sources
+
+    result = subprocess.run(
+        ["bash", "-c", f"set -euo pipefail; {sources[0]}; type -t retry_install"],
+        env={"PATH": "/usr/bin:/bin", "ACTION_PATH": str(REPO_ROOT)},
+        capture_output=True,
+        text=True,
+    )
+
+    assert result.returncode == 0, result.stderr
+    assert result.stdout.strip() == "function", result.stdout
+
+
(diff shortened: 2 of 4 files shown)


max-sixty:

PR #916, opened 2026-08-09
fix(mention): skip the bot's own review on someone else's PR

### LINKED ISSUE #915: tend-mention: bot's own non-empty-body review on someone else's PR starts a guaranteed no-op session (#747's sibling leg)
## Problem

`tend-mention` starts a full billable session every time **the bot leaves a non-empty-body review on a same-repo PR someone else authored**. The session reads the PR, recognises the trigger as its own review, and exits silently. It is a guaranteed no-op: the bot is the reviewer, the findings are addressed to the PR's author, and the bot pushing unbidden to another author's branch is barred by `running-in-ci`'s conduct rules — so there is no author role for it to act in.

This is the sibling leg of #747. That issue fixed the *empty-body `APPROVED`* case; the gate it added stops at `body == ""`, on the reasoning max-sixty gave in the thread — *"if it's an approval with nits then we may want to make follow-up changes?"*. That reasoning holds on a PR the **bot** authored (which is exactly what #166 and #761 protect), but not on someone else's PR, where the nits are instructions for the human author and the bot has nothing to follow up.

Distinct from #866 / #849, which target the synthetic empty-body `COMMENTED` container GitHub creates for an inline reply. These are real bot reviews with real bodies (455–1920 chars).

## Evidence

Five occurrences in one 24h window on PRQL/prql (bot login `prql-bot`), covering every non-empty-body bot review left on a same-repo human PR in that window — there were no counter-examples:

| Review | PR (author) | State / body | Session started | Cost | Outcome |
|---|---|---|---|---|---|
| 4889888509 | [#6169](https://github.com/PRQL/prql/pull/6169) (@max-sixty) | `APPROVED`, 1567 chars | [31280580245](https://github.com/PRQL/prql/actions/runs/31280580245) | $0.44 | exited silently |
| 4889928909 | [#6170](https://github.com/PRQL/prql/pull/6170) (@max-sixty) | `COMMENTED`, 1453 chars | [31281535113](https://github.com/PRQL/prql/actions/runs/31281535113) | $0.00 | killed by a 429 before it could exit (see below) |
| 4890641190 | [#6172](https://github.com/PRQL/prql/pull/6172) (@eitsupi) | `COMMENTED`, 1503 chars | [31297736617](https://github.com/PRQL/prql/actions/runs/31297736617) | $0.41 | exited silently |
| 4890671072 | [#6172](https://github.com/PRQL/prql/pull/6172) (@eitsupi) | `COMMENTED`, 1920 chars | [31298590569](https://github.com/PRQL/prql/actions/runs/31298590569) | $0.45 | exited silently |
| 4890681444 | [#6172](https://github.com/PRQL/prql/pull/6172) (@eitsupi) | `APPROVED`, 455 chars | [31298885324](https://github.com/PRQL/prql/actions/runs/31298885324) | $0.49 | exited silently |

$1.79 booked plus one run that never got to bill, out of a $31.52 window — and each also costs a `relay` run and a `verify` run.

The sessions say so themselves. From [31297736617](https://github.com/PRQL/prql/actions/runs/31297736617): *"The triggering review (ID 4890641190) was authored by prql-bot itself on eitsupi's PR #6172 — I'm the reviewer, not the author, and there are no replies or other comments on the thread. Responding to my own review would be a self-conversation loop, so I'm exiting silently without posting anything."* From [31298590569](https://github.com/PRQL/prql/actions/runs/31298590569): *"I'm the reviewer, @eitsupi is the author, and the review's findings are addressed to them. … No comment, review, or push made."*

**The waste is not free.** [31281535113](https://github.com/PRQL/prql/actions/runs/31281535113) is the one that failed rather than exiting: its session JSONL ends on `"error":"rate_limit"`, `"apiErrorStatus":429`, `You've hit your session limit · resets 12am (UTC)`. The account's Claude session budget was exhausted at that moment, and one of the requests that had been drawing it down was this class of session. A no-op that can't run is still a no-op, but a no-op that consumes a scarce slot and then files a `tend-outage` issue is worse than one that just bills $0.45.

## Root cause

For a `pull_request_review` event the gate falls through to the participation heuristic at [`mention.yaml.j2#L307-L313`](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/generator/src/tend/templates/mention.yaml.j2#L307-L313), which counts the triggering review itself, so `BOT_REVIEWS` is always non-empty and `should_run=true` is unconditional. The only skip on that path is the [three-clause gate at L248-L254](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/generator/src/tend/templates/mention.yaml.j2#L248-L254), whose `approved` + empty-body clauses exclude every review in the table above.

## Proposed fix

The discriminator that's missing is the **PR's author**, and it's already resolved one line earlier. Insert between the `PR_AUTHOR` check ([L301-L305](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/generator/src/tend/templates/mention.yaml.j2#L301-L305)) and the `BOT_REVIEWS` heuristic:

```bash
          # A review the bot leaves on someone else's PR has no author role to
          # follow up in — the findings are addressed to that PR's author, and
          # pushing unbidden to another author's branch is barred by
          # running-in-ci's conduct rules. The BOT_REVIEWS heuristic below
          # counts this very review, so without this the session always starts
          # and always exits silently (#747 fixed only the empty-body APPROVED
          # leg of the same shape). Placed *after* the PR_AUTHOR check above,
          # so #166/#761's actionable self-reviews on the bot's own PRs still
          # fire, and after the body/inline @-mention scans, so an explicit
          # summons still wins.
          if [ "$KIND" = "pull_request_review" ] \
             && [ "$REVIEW_AUTHOR" = "<<cfg.bot_name>>" ]; then
            echo "should_run=false" >> "$GITHUB_OUTPUT"; exit 0
          fi
```

At that point `PR_AUTHOR != <bot>` is already established, because the preceding block `exit 0`s when they're equal — so this cannot swallow the bot-authored-PR case. Everything the earlier gates admit is unaffected: a body `@`-mention returns at [L220-L224](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/generator/src/tend/templates/mention.yaml.j2#L220-L224), an inline-comment `@`-mention at [L260-L267](https://github.com/max-sixty/tend/blob/0f35b6b30ba045d22165302b3e49fafb6504fe4c/generator/src/tend/templates/mention.yaml.j2#L260-L267), and a human's review never reaches here with `REVIEW_AUTHOR` set to the bot.

Scope note: only same-repo PRs are affected. The `relay` job's `head.repo.full_name == github.repository` condition means bot reviews on fork PRs never dispatch — verified against the same window, where three non-empty-body bot reviews on @kgutwin's fork PRs (#6155, #6156, #6162) started no session at all.

Worth confirming against the control in the same window: [31300648824](https://github.com/PRQL/prql/actions/runs/31300648824), triggered by the bot's own `COMMENTED` review on [#6173](https://github.com/PRQL/prql/pull/6173) — a **bot-authored** PR — is the case that must keep firing, and the proposed gate leaves it firing.

<details><summary>How the occurrences were enumerated</summary>

Every bot review submitted in the window was listed with `state` and `body` length, cross-referenced against the PR's `head.repo.full_name` and author, then matched to the `repository_dispatch` run it produced; each run's session artifact was downloaded and its final assistant turn read to confirm the exit was a no-op and that no comment, review, or push was made. The two empty-body `APPROVED` reviews in the same window (4890133352 on #6170, 4889810454 on #6168) correctly produced no session — #747's gate working — which is what makes the body-length clause the load-bearing one.

</details>

tend-agent:
## Problem

`tend-mention`'s verify gate starts a full billable session for every non-empty-body review the bot leaves on a same-repo PR someone else authored. The session reads the PR, recognises the trigger as its own review, and exits silently — it can only ever no-op: the bot is the reviewer, the findings are addressed to the PR's author, and pushing unbidden to another author's branch is barred by `running-in-ci`'s conduct rules, so there is no author role left to act in.

For a `pull_request_review` event the gate falls through to the participation heuristic, which counts *the triggering review itself* — `BOT_REVIEWS` is never empty, so `should_run=true` is unconditional. #747's gate is the same shape narrowed to one leg (`approved` **and** empty body), which excludes every review in #915's evidence: five occurrences in one 24h window on PRQL/prql, `APPROVED` and `COMMENTED` alike, bodies 455–1920 chars, $1.79 booked plus one run that hit the account's session limit before it could bill.

## Solution

One author-keyed gate between the `PR_AUTHOR` resolution and the `BOT_REVIEWS` heuristic. Placement is the whole design, in three directions:

- **After the `PR_AUTHOR` short-circuit**, which has already exited `should_run=true` when the PR is the bot's own. A review the bot's review workflow leaves on a bot-authored PR is the reviewer role handing work to the author role and must keep firing (#166, #761) — an author-keyed gate placed any earlier would swallow it.
- **After the body and inline-comment @-mention scans**, so an explicit summons still wins.
- **Before `BOT_REVIEWS`**, which is what misreads the triggering review as prior engagement.

Keyed on author alone: reusing #747's `REVIEW_STATE` / empty-body clauses here would let every bodied `COMMENTED` review straight through.

Fork PRs are unaffected either way — the `relay` job's `head.repo.full_name == github.repository` condition already means a bot review on a fork PR never dispatches.

## Testing

`test_mention_skips_bot_review_on_another_authors_pr` pins the gate and all three orderings, and asserts the state/body clauses are *absent* from it. It fails on the pre-change template at the first assertion. `test_mention_self_comment_skip_spares_review_submissions` had its APPROVED-gate slice narrowed to end at the inline-comment fetch, so it keeps discriminating between the two gates rather than passing on either. Full generator suite: 380 passed. `pre-commit` clean on the changed files.

## Relationship to #849

Complementary, not overlapping. #849 (for #866) drops the synthetic zero-body `COMMENTED` container GitHub wraps around an inline reply, on *any* PR including the bot's own — where this gate cannot reach, because `PR_AUTHOR == bot` exits first. This one drops real bodied reviews on other authors' PRs, which #849's `-z "$COMMENT_BODY"` clause excludes. Both touch `mention.yaml.j2`; whichever lands second will need a one-line rebase on the header comment listing the self-review skips.

---

Closes #915 — automated triage

### DIFF
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -209,10 +209,10 @@ jobs:
           # comments on purpose: the pull_request_review *submission* kind is
           # deliberately left out, since a review the bot leaves on its own PR
           # is its reviewer role (the prompt is told to action it), not a
-          # self-loop. The self-review skips all sit below, and both are keyed
-          # on the review being terminal for the bot rather than on authorship:
-          # the empty-body APPROVED gate, and any bot review on a PR the bot
-          # did not author.
+          # self-loop. The two self-review skips below are author-keyed, but
+          # each is also narrowed to a case that leaves this run nothing to do:
+          # the empty-body APPROVED gate, and a bot review on a PR the bot did
+          # not author. Neither licenses a blanket self-review skip.
           if { [ "$KIND" = "issue_comment" ] || [ "$KIND" = "pull_request_review_comment" ]; } \
              && [ "$COMMENT_AUTHOR" = "<<cfg.bot_name>>" ]; then
             echo "should_run=false" >> "$GITHUB_OUTPUT"
@@ -306,10 +306,12 @@ jobs:
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
           fi
 
-          # A review the bot leaves on someone else's PR is terminal for the
-          # bot: the findings are addressed to that PR's author, and pushing
-          # unbidden to another author's branch is barred by conduct rules, so
-          # there is no author role left to act in. The BOT_REVIEWS heuristic
+          # A review the bot leaves on someone else's PR leaves this run
+          # nothing to do: whatever the review warranted, the tend-review
+          # session that submitted it has already done — left the findings for
+          # a human author to act on (pushing to their branch unbidden is
+          # barred by conduct rules), or, on a dependency-bot PR where no
+          # author will act, pushed the fix itself. The BOT_REVIEWS heuristic
           # below counts this very review, so without this gate the session
           # always starts and always exits silently. Keyed on author alone —
           # the APPROVED + empty-body gate above is the same shape narrowed to
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -209,8 +209,10 @@ jobs:
           # comments on purpose: the pull_request_review *submission* kind is
           # deliberately left out, since a review the bot leaves on its own PR
           # is its reviewer role (the prompt is told to action it), not a
-          # self-loop. The only self-review skip is the terminal empty-body
-          # APPROVED gate below.
+          # self-loop. The self-review skips all sit below, and both are keyed
+          # on the review being terminal for the bot rather than on authorship:
+          # the empty-body APPROVED gate, and any bot review on a PR the bot
+          # did not author.
           if { [ "$KIND" = "issue_comment" ] || [ "$KIND" = "pull_request_review_comment" ]; } \
              && [ "$COMMENT_AUTHOR" = "<<cfg.bot_name>>" ]; then
             echo "should_run=false" >> "$GITHUB_OUTPUT"
@@ -302,6 +304,24 @@ jobs:
           if [ "$PR_AUTHOR" = "<<cfg.bot_name>>" ]; then
             echo "should_run=true" >> "$GITHUB_OUTPUT"
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
+          fi
+
+          # A review the bot leaves on someone else's PR is terminal for the
+          # bot: the findings are addressed to that PR's author, and pushing
+          # unbidden to another author's branch is barred by conduct rules, so
+          # there is no author role left to act in. The BOT_REVIEWS heuristic
+          # below counts this very review, so without this gate the session
+          # always starts and always exits silently. Keyed on author alone —
+          # the APPROVED + empty-body gate above is the same shape narrowed to
+          # its one terminal leg, and reusing those clauses here would let every
+          # bodied COMMENTED review through. Placement carries the rest of the
+          # design: *after* the PR_AUTHOR short-circuit, which has already
+          # exited when the PR is the bot's own, so the reviewer-to-author
+          # handoff on a bot PR still fires; *after* the body and inline
+          # @-mention scans, so an explicit summons still wins.
+          if [ "$KIND" = "pull_request_review" ] \
+             && [ "$REVIEW_AUTHOR" = "<<cfg.bot_name>>" ]; then
+            echo "should_run=false" >> "$GITHUB_OUTPUT"; exit 0
           fi
 
           # Captured, not counted — see the note on the issue-comment lookup above.
--- a/generator/tests/test_generate.py
+++ b/generator/tests/test_generate.py
@@ -1046,9 +1046,11 @@ def test_mention_self_comment_skip_spares_review_submissions(
     role speaking, not a self-loop — the prompt is told to action it. That
     signal arrives as a `pull_request_review` submission event, distinct from
     the `pull_request_review_comment` inline-comment event the skip covers, so
-    it must still reach the actionable path. The only self-review that is
-    skipped is the terminal empty-body APPROVED case (see
-    test_mention_skips_bot_approved_review).
+    it must still reach the actionable path. The self-reviews that are skipped
+    are skipped for being terminal for the bot, not for being self-authored:
+    the empty-body APPROVED case (see test_mention_skips_bot_approved_review)
+    and a review on a PR the bot did not author (see
+    test_mention_skips_bot_review_on_another_authors_pr).
 
     This pins the boundary against a later "skip self-authored reviews too, for
     consistency" edit that would silently break the review -> fix loop: the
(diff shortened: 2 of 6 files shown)


generator/src/tend/templates/mention.yaml.j2
```diff
@@ -209,8 +209,10 @@ jobs:
           # comments on purpose: the pull_request_review *submission* kind is
           # deliberately left out, since a review the bot leaves on its own PR
           # is its reviewer role (the prompt is told to action it), not a
-          # self-loop. The only self-review skip is the terminal empty-body
-          # APPROVED gate below.
+          # self-loop. The self-review skips all sit below, and both are keyed
+          # on the review being terminal for the bot rather than on authorship:
+          # the empty-body APPROVED gate, and any bot review on a PR the bot
+          # did not author.
```
tend-agent: The new gate *is* keyed on authorship and nothing else — this framing tells the next editor the opposite of what the gate below says about itself.

```suggestion
          # self-loop. Two self-review skips sit below, both for reviews that
          # are terminal for the bot: the empty-body APPROVED gate, and — after
          # the PR-author short-circuit has already let the bot's own PR through
          # — any bot review, keyed on author alone.
```

generator/src/tend/templates/mention.yaml.j2
```diff
@@ -304,6 +306,24 @@ jobs:
             echo "reason=participation" >> "$GITHUB_OUTPUT"; exit 0
           fi
 
+          # A review the bot leaves on someone else's PR is terminal for the
+          # bot: the findings are addressed to that PR's author, and pushing
+          # unbidden to another author's branch is barred by conduct rules, so
+          # there is no author role left to act in. The BOT_REVIEWS heuristic
```
tend-agent: The conduct-rule bar only covers human authors; on a Dependabot/renovate PR the review skill directs pushing the fix. What actually makes mention skippable there is that `tend-review` pushes it in-session.

```suggestion
          # unbidden to a human author's branch is barred by conduct rules; on a
          # bot-authored PR tend-review pushes the fix in-session, so either way
          # there is no author role left here to act in. The BOT_REVIEWS heuristic
```

generator/tests/test_generate.py
```diff
@@ -1081,18 +1083,76 @@ def test_mention_self_comment_skip_spares_review_submissions(
...
+    assert '[ -z "$COMMENT_BODY" ]' not in gate
+
+    # The bot's own PR still reaches the actionable short-circuit, because that
+    # check exits before the gate is read.
+    assert run.index('[ "$PR_AUTHOR" = "test-bot" ]') < run.index(
+        '[ "$KIND" = "pull_request_review" ]', run.index("PR_AUTHOR=$(gh pr view")
+    )
+
+    # And an @-mention inside the review — body or inline comment — still wins.
+    assert run.index("reviews/$PAYLOAD_ID/comments") < run.index("BOT_REVIEWS=$(")
```
tend-agent: Vacuous as written — the inline fetch precedes `BOT_REVIEWS=$(` on the pre-change template as well, so this passes either way. Anchoring to the gate's own index (which doesn't exist pre-change, so `.index` raises) makes it pin the ordering the docstring claims.

```suggestion
    assert run.index("reviews/$PAYLOAD_ID/comments") < run.index(
        '[ "$KIND" = "pull_request_review" ]', run.index("PR_AUTHOR=$(gh pr view")
    )
```

tend-agent (commented):
The gate itself is right, and its placement holds up under trace: `REVIEW_AUTHOR` is only ever set on the `pull_request_review` arm of the dispatch `case`, so the second clause can't read a stale value from an `issue_comment`; the earlier `APPROVED` + empty-body gate stays load-bearing for the bot's own PRs rather than becoming dead; and all four mention regtests that carry the verify block are updated (`test_extras_apply_path_regtest.out` renders `tend-review`, so it's correctly untouched). Suite passes at 380 locally.

Three things on the prose and the test, none blocking.

The header comment now says the two skips are keyed "on the review being terminal for the bot rather than on authorship", while the new gate's own comment eleven lines later says "Keyed on author alone". Both are describing the same gate in opposite terms — "authorship" means *self*-authorship of the review in the first and *review-author as the key* in the second. That block's whole job is to steer a future editor away from a blanket self-review skip, so a framing that reads as "these aren't author-keyed" is the wrong signal to leave there.

The rationale in the gate comment — pushing unbidden to another author's branch is barred by conduct rules — is narrower than the set of PRs the gate covers. On a Dependabot/renovate PR the review skill's step 8 explicitly directs pushing the fix, so "no author role left to act in" isn't why mention is skippable there; it's skippable because `tend-review` pushes that fix in its own session and never needed a mention run. The behaviour is right either way, but as written the comment will read as an oversight to whoever next debugs a bot-authored PR that stopped waking mention.

The last assertion in the new test doesn't discriminate. `run.index("reviews/$PAYLOAD_ID/comments") < run.index("BOT_REVIEWS=$(")` holds on the pre-change template too (the inline fetch has always preceded the heuristic), so the third ordering the docstring claims to pin isn't pinned by anything except the `gate` slice bounds. Anchoring the right-hand side to the gate's own index makes it fail pre-change, like the other two.

The `#849` overlap is called correctly — that one is only reachable on bot-authored PRs once this gate lands, since everything it targets on someone else's PR is now caught here first. Both edit this same header comment, so whichever is second wants the rebase you flagged.

tend-agent:
All three addressed in 5a3e441.

The header comment no longer denies the gates are author-keyed — it says they are, and that each is *additionally* narrowed to a case that leaves the run nothing to do, which is the steer against a blanket self-review skip stated without the contradiction.

On the rationale: confirmed against the review skill, which directs pushing the fix on Dependabot/renovate PRs rather than stopping at analysis. So the conduct-rules clause was the wrong load-bearing reason. Reframed to what actually covers both cases — whatever the review warranted, the `tend-review` session that submitted it has already done: left the findings for a human author (where pushing to their branch unbidden *is* barred), or pushed the fix itself where no author will act.

The third assertion now anchors its right-hand side to the gate's own index rather than to `BOT_REVIEWS=$(`. Verified it discriminates: with `main`'s template restored under the new test file, the test fails; with the gate, it passes.

generator/src/tend/templates/mention.yaml.j2
tend-agent: Agreed — the block's whole job is steering away from a blanket self-review skip, so "these aren't author-keyed" was the wrong signal. Fixed in 5a3e441, though not quite as suggested: rather than dropping the terminality framing, it keeps both facts and states how they relate — the skips *are* author-keyed, and each is *additionally* narrowed to a case that leaves the run nothing to do — then closes with "Neither licenses a blanket self-review skip" so the steer is explicit rather than inferred from the framing.

generator/src/tend/templates/mention.yaml.j2
tend-agent: Correct, and the premise checks out: the review skill tells the reviewer to push the fix on Dependabot/renovate PRs rather than stop at analysis ([SKILL.md#L463](https://github.com/max-sixty/tend/blob/5a3e441968b2ffac8fd78c422274dc39e874903c/plugins/tend-ci-runner/skills/review/SKILL.md#L463), and the dismissal path that routes to it at [#L391](https://github.com/max-sixty/tend/blob/5a3e441968b2ffac8fd78c422274dc39e874903c/plugins/tend-ci-runner/skills/review/SKILL.md#L391)). So the conduct bar was never what made those PRs skippable.

Fixed in 5a3e441, which moves the rationale off the bar and onto the actor: whatever the review warranted, the `tend-review` session that submitted it has already done — left the findings for a human author, or pushed the fix itself where no author will act. That covers both legs without claiming there is nothing a bot author could have done.

generator/tests/test_generate.py
tend-agent: Confirmed vacuous, and fixed in 5a3e441 — anchored to `[ "$REVIEW_AUTHOR" = "test-bot" ]` searched from the `PR_AUTHOR=$(gh pr view` index, which is equivalent to the suggestion (the gate's two clauses are adjacent) and fails the same way without the gate.

Verified rather than reasoned: rendering the pre-change template raises `ValueError` at that `.index` call, so the test errors out rather than passing on a `False` comparison; the landed template gives `6502 < 9287`. One trap worth recording for anyone re-running that check — swapping the template file in-process doesn't work. Jinja caches the compiled template by name, so both halves of the comparison render the *same* output and the new assertion looks like it passes pre-change too. It needs a fresh process per template.

PR #920, opened 2026-08-09
chore: bump the workflow action pins we run ourselves

tend-agent:
Weekly `uses:` drift sweep, "ours alone" half: the hand-maintained `.github/workflows/` files plus `.config/tend.yaml`. Nothing here ships to adopters — the adopter-facing refs are split into their own PRs.

| Ref | Was | Now | Where |
|---|---|---|---|
| `actions/checkout` | v6 | v7 | ci, publish-site, pypi-release, review-reviewers, worker-deploy |
| `actions/setup-node` | v4 | v7 | ci, publish-site, worker-deploy |
| `astral-sh/setup-uv` | v7 | v9.0.0 | ci, pypi-release |
| `astral-sh/setup-uv` | v6 | v9.0.0 | `.config/tend.yaml` `setup:` |
| `actions/upload-artifact` | v6 | v7 | pypi-release |
| `actions/download-artifact` | v7 | v8 | pypi-release |
| `pypa/gh-action-pypi-publish` | v1.13.0 | v1.14.2 | pypi-release |

The generated `tend-*.yaml` changes are the `setup:` bump landing through `uvx tend@0.1.14 init` — one `setup-uv` line per workflow, nothing else. Regenerating against the released 0.1.14 (not the in-tree generator) keeps the `max-sixty/tend/claude@0.1.14` pin resolvable. Nothing in CI checks that regen — `tend-install-test` only exists on an install PR (`init --with-install-test`) and the nightly regen deletes it — so it was verified by hand: `uvx tend@0.1.14 init` on this branch leaves the tree unchanged.

Two things worth knowing rather than just merging past:

**`setup-uv` no longer publishes floating major tags.** v8.0.0 was the first immutable release and dropped the moving `v8`/`v9` refs — `astral-sh/setup-uv@v9` does not resolve. Hence the exact `@v9.0.0` pin, which is a shape change from the `@v6`/`@v7` this repo used before; future bumps of this action are full-version edits.

**`setup-uv` v9.0.0 flips `prune-cache` to `false` by default** (astral-sh/setup-uv#967) to reduce load on PyPI infrastructure. Upstream marked it breaking because it can raise Actions cache usage. That lands in this repo's CI and in the generated tend workflows; the mitmproxy cache in `claude/action.yaml` is a separate `actions/cache` entry and is unaffected. If cache growth is unwelcome, `prune-cache: true` restores the old behavior — say the word and I'll add it to `.config/tend.yaml`'s `setup:` rather than hold the version back.

The other majors are routine: `checkout` v7 is dependency updates plus PR-checkout input hardening; `setup-node` v7 is an ESM migration with new `cache-primary-key`/`cache-matched-key` outputs and the dummy `NODE_AUTH_TOKEN` export removed; `upload-artifact` v7 and `download-artifact` v8 are runtime/dependency refreshes (v8 adds CJK artifact-name support); `gh-action-pypi-publish` v1.14.2 is a dependency-tree update whose visible piece is Twine 7, which is what lets an sdist or wheel carrying core-metadata 2.5 upload at all.

### DIFF
--- a/.config/tend.yaml
+++ b/.config/tend.yaml
@@ -1,7 +1,7 @@
 bot_name: tend-agent
 
 setup:
-  - uses: astral-sh/setup-uv@v6
+  - uses: astral-sh/setup-uv@v9.0.0
 
 workflows:
   ci-fix:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,8 +11,8 @@ jobs:
       run:
         working-directory: generator
     steps:
-      - uses: actions/checkout@v6
-      - uses: astral-sh/setup-uv@v7
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v9.0.0
       - run: uv run pytest
 
   # Tests for the shared credential-injection proxy addon. Run standalone (no
@@ -24,8 +24,8 @@ jobs:
       run:
         working-directory: proxy
     steps:
-      - uses: actions/checkout@v6
-      - uses: astral-sh/setup-uv@v7
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v9.0.0
       # Read the version from the Claude action's mitmproxy_version default so
       # the test exercises the version the production proxy runs. The addon
       # imports mitmproxy.test (an internal helper), so a second copy of the pin
@@ -97,8 +97,8 @@ jobs:
       run:
         working-directory: plugins/install-tend/skills/install-tend/scripts
     steps:
-      - uses: actions/checkout@v6
-      - uses: astral-sh/setup-uv@v7
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v9.0.0
       - run: uv run --no-project --with pytest pytest
 
   test-worker:
@@ -107,8 +107,8 @@ jobs:
       run:
         working-directory: worker
     steps:
-      - uses: actions/checkout@v6
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@v7
+      - uses: actions/setup-node@v7
         with:
           node-version: "20"
           cache: npm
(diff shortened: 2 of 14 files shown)


tend-agent (commented):
Self-review. One correction, already applied to the description: `tend-install-test` does not verify the regen here. It's generated only by `uvx tend init --with-install-test`, which the install skill passes on the initial install PR; the nightly regen runs without the flag so the file gets deleted (see the `--with-install-test` help text in `generator/src/tend/cli.py`). There is no `tend-install-test.yaml` in `.github/workflows/`, so nothing in CI compares the committed `tend-*.yaml` against generator output.

Verified by hand instead: `uvx tend@0.1.14 init` on the merge commit leaves the tree byte-identical — `git status --porcelain .github/workflows/ .config/` is empty — so the `setup:` line really is the only thing that landed in the generated files.

The rest of the sweep holds up:

- Every new ref resolves, and each is the current latest: `checkout@v7`, `setup-node@v7`, `upload-artifact@v7`, `download-artifact@v8`, `setup-uv@v9.0.0`, `gh-action-pypi-publish@v1.14.2`.
- `astral-sh/setup-uv@v9` genuinely does not exist — the tag list stops publishing moving refs after `v7`/`v7.6`, so the full-version pin is required, not stylistic.
- Coverage is complete. The "ours alone" refs left un-bumped — `pre-commit/action@v3.0.1`, `actions/upload-pages-artifact@v5`, `actions/configure-pages@v6`, `actions/deploy-pages@v5` — are each already at their latest release.
- `checkout` v7's headline change is broader than "PR-checkout input hardening": [actions/checkout#2454](https://github.com/actions/checkout/pull/2454) refuses to check out fork PR code under `pull_request_target` or `workflow_run` without `allow-unsafe-pr-checkout: true`. It doesn't bite here — none of the five files bumped in this PR use either trigger, and the generated `tend-review.yaml` (which does) already passes the opt-in.
- `setup-node` v4→v7 crosses v5's automatic package-manager caching, but all three call sites set `cache: npm` with an explicit `cache-dependency-path`, so the auto-detection path is never taken.

What CI does not cover: `pypi-release.yaml` runs only on a release tag push, so `upload-artifact@v7`, `download-artifact@v8` and `gh-action-pypi-publish@v1.14.2` first execute during the next release rather than on this PR. The one behavioral change to watch there is `download-artifact` v8 defaulting `digest-mismatch` to `error` (previously a warning) — correct-by-default, but it turns a formerly silent condition into a failed release job.

Not approving — self-authored.

PR #921, opened 2026-08-09
chore: bump astral-sh/setup-uv to v9.0.0

tend-agent:
Weekly `uses:` drift sweep, adopter-facing half for `astral-sh/setup-uv`: v6 → v9.0.0 in the generated `tend-install-test` workflow (`generator/src/tend/workflows.py`) and in the commented `setup:` example that both `docs/tend.example.yaml` and the install skill's copy show adopters. The refs this repo runs itself move separately in #920.

**No floating major tag.** v8.0.0 was `setup-uv`'s first immutable release and dropped the moving major/minor refs — `astral-sh/setup-uv@v9` does not resolve, only `@v9.0.0` does. That's a shape change for anyone copying the example: future bumps are full-version edits, not a major that drifts forward on its own.

What the three majors carry:

- **v7.0.0** — Node 20 → Node 24, hence "breaking" for self-hosted runners that don't auto-update. Also removes the long-deprecated `server-url` input, superseded by `manifest-file`. Neither the generated workflow nor the example uses it.
- **v8.0.0** — immutable releases, plus removal of the old `manifest-file` format (a new format replaces it) and the major/minor tags above.
- **v9.0.0** — `prune-cache` now defaults to `false`, to reduce load on PyPI infrastructure (astral-sh/setup-uv#967). Marked breaking because it can raise an adopter's Actions cache usage. It's the one change here an adopter might actually notice; `prune-cache: true` restores the old behavior.

`tend-install-test` is a one-shot workflow (`init --with-install-test`) that runs on PRs touching the generated files, so the exposure of the v9 cache change through *that* path is small — the wider surface is adopters who copied the example `setup:` block.

Also relaxes the `setup-uv` assertion in `test_install_test_workflow_shape` to be version-agnostic. The exact pin is already covered byte-for-byte by the regtest output; asserting it a second time meant every weekly bump edited two places to say the same thing.

### DIFF
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -120,14 +120,14 @@ bot_name: my-project-bot
 # `uses`:
 #
 #     setup:
-#       - uses: astral-sh/setup-uv@v6
+#       - uses: astral-sh/setup-uv@v9.0.0
 #
 #     # renders as (in tend-notifications):
-#     #   - uses: astral-sh/setup-uv@v6
+#     #   - uses: astral-sh/setup-uv@v9.0.0
 #     #     if: steps.check.outputs.count != '0' || github.event_name == 'workflow_dispatch'
 #     #
 #     # and in other workflows:
-#     #   - uses: astral-sh/setup-uv@v6
+#     #   - uses: astral-sh/setup-uv@v9.0.0
 #
 # `uses` with action inputs and step-level env:
 #
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -468,7 +468,7 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v7
-      - uses: astral-sh/setup-uv@v6
+      - uses: astral-sh/setup-uv@v9.0.0
       - name: Verify generator output matches committed files
         env:
           GH_TOKEN: ${{{{ github.token }}}}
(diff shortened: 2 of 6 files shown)


tend-agent (commented):
Verified the mechanics: `v9.0.0` resolves, and `astral-sh/setup-uv@v9` / `@v8` both 404 — the moving major tags stop at `v7.6`, so the full-version pin in the example is the only form that works and the PR body's "shape change" framing is right. The relaxed `test_install_test_workflow_shape` assertion is genuinely double-covered: both `test_install_test_workflow_regtest` outputs carry the exact `- uses: astral-sh/setup-uv@v9.0.0` line.

One gap in the same file, outside this diff. The `uses` with action inputs example a dozen lines below the block this PR edits still shows `actions/setup-node@v4`; latest is `v7.0.0`. It's adopter-facing for the same reason the `setup-uv` lines are, and the weekly sweep's grep does match it — it's in that grep's output today, alongside `astral-sh/setup-uv@v9.0.0`, because the recipe only excludes `generator/tests` and `*.md`, not commented lines in `docs/tend.example.yaml`. No open PR covers it, so the adopter-facing half of this week's sweep landed `setup-uv` and skipped `setup-node`. Per the one-PR-per-action convention that belongs in its own PR rather than folded in here — happy to open it if you want.

tend-agent:
Opened #924 for the `actions/setup-node@v4` gap, per the one-PR-per-action split. It's `v4` → `v7` on both occurrences in the example block (the `setup:` snippet and the "renders as" output), in `docs/tend.example.yaml` and its mirror under the install skill's `references/`.

One difference from this PR worth noting: `setup-node` still publishes a moving major — `@v7` and `@v7.0.0` resolve to the same commit — so #924 pins the bare major, matching every other `actions/*` ref here. The full-version form this PR adopts is specific to `setup-uv` having dropped its major/minor refs at v8.

This PR is unchanged; nothing in the review asked for an edit here.

PR #923, opened 2026-08-09
ci: smoke the pinned Codex CLI surface, and bump it off the alpha line

### LINKED ISSUE #922: codex_version can't be bumped: the weekly rule's confirmation isn't reachable from CI
`codex/action.yaml` pins `codex_version: 0.131.0-alpha.22`. The alpha line is now `0.148.0-alpha.5` and the stable line is `0.147.0` — seventeen minors of drift, and it has not moved since the pin was introduced.

The `running-tend` weekly rule is deliberately conservative here: *"keep it on its prerelease line; bump only to a release confirmed to run under `codex exec`."* The problem is that nothing in this repo can produce that confirmation. There is no `harness: codex` override in `.config/tend.yaml`, no codex step in `ci.yaml`, and no OpenAI credential reachable from a weekly run — so a live `codex exec` session is not something the weekly job can perform, and the bump condition can never be satisfied by the bot. The pin drifts by default rather than by decision.

What I *can* verify from CI, and did this week against `0.148.0-alpha.5`, is that the CLI surface the action depends on is intact:

<details><summary>Surface probe against 0.148.0-alpha.5</summary>

```
$ npm install @openai/codex@0.148.0-alpha.5
$ codex --version
codex-cli 0.148.0-alpha.5

$ codex exec --help        # every flag the action passes is still accepted
  -m, --model <MODEL>
  -s, --sandbox <SANDBOX_MODE>
  -o, --output-last-message <FILE>
  -c, --config <key=value>

$ codex plugin marketplace add <repo root>
Added marketplace `tend` from /home/runner/work/tend/tend.

$ codex plugin add tend-ci-runner@tend
Added plugin `tend-ci-runner` from marketplace `tend`.
Installed plugin root: <path>
```

The `Installed plugin root: ` line is the one the action parses with `awk` to set `CLAUDE_PLUGIN_ROOT`; it still has that exact prefix.

</details>

That covers the flags and the plugin-install parse, which is where a silent break would most likely land. It does **not** cover an actual agent session — model selection, `model_reasoning_effort`, sandbox behavior, or whether the final message reaches `--output-last-message`.

Two ways to close the gap, both cheap relative to letting the pin rot:

1. **A codex smoke in `ci.yaml`**, gated on an `OPENAI_API_KEY` secret in the `tend` environment, running one trivial `codex exec` against the pinned version. That makes the weekly bump self-verifying and is the version of this that doesn't need a person each time.
2. **A `harness: codex` override on one low-stakes workflow** (the pattern `.config/tend.yaml` already supports for trialling a harness per workflow), so the pin gets exercised by real runs and a bad bump surfaces on a workflow whose failure costs little.

Until one of those exists, the weekly run will keep reporting this drift and declining to act on it, which is the correct behavior under the current rule but not a good steady state.

tend-agent:
## Problem

`codex_version` is pinned at `0.131.0-alpha.22` in `codex/action.yaml` while `latest` is `0.147.0`. The weekly rule governing it asked for a confirmation nothing in this repo could produce — *"bump only to a release confirmed to run under `codex exec`"* — because there is no `OPENAI_API_KEY` in the `tend` environment (`gh secret list --env tend` returns only `CLAUDE_CODE_OAUTH_TOKEN` and `TEND_BOT_TOKEN`), no `harness: codex` override in `.config/tend.yaml`, and no codex step in `ci.yaml`. The bump condition was unsatisfiable, so the pin drifted by default rather than by decision.

Both proposals in #922 need that same missing secret, so neither is actually the cheaper one. What doesn't need it is the part of the confirmation that is mechanically checkable — and it turns out to be the part where a silent break would land.

## Solution

**A `test-codex-surface` job in `ci.yaml`**, shaped after the existing `test-proxy` smoke: read the pin out of the action with `yq -e`, install it, and assert the surface the action depends on.

- Every flag the `Run Codex` step passes — `--model`, `--sandbox`, `--output-last-message`, `--config`. Matching on `--flag <` pins each to an option that still takes a value, so a release that keeps the name but drops the argument fails too.
- The `Installed plugin root: ` prefix that `codex plugin add` prints and the action parses with `awk` to set `CLAUDE_PLUGIN_ROOT`, plus that both trees skills reach for through that variable land under it — `skills/` and `scripts/`. Only `skills/` is declared in `plugin.json`, so `scripts/` materializing is a packaging side effect a release could stop honouring while every skill shelling out to `${CLAUDE_PLUGIN_ROOT}/scripts/` breaks.

No credential is involved: `codex --version`, `codex exec --help`, `codex plugin marketplace add`, and `codex plugin add` are all local. So the job runs on every PR, which also guards the other direction — a change to `.agents/plugins/marketplace.json` or the plugin layout that Codex can no longer install now fails here rather than in an adopter's job, which nothing catches today.

**The bump target moves to `latest`, and the pin moves with it.** The premise behind the prerelease pin has expired: `codex plugin add` was `alpha`-only when the pin was set, but stable `0.147.0` carries it. So the `running-tend` weekly rule now reads `track latest` (dropping to `alpha` only for a fix not yet released), the two `codex/action.yaml` comments that still said `latest` was `0.130.0` and lacked the subcommand are corrected, and `codex_version` is bumped `0.131.0-alpha.22` → `0.147.0`. The bump ships here rather than separately because this job is its confirmation — the gate and the first bump it gates are the same review, and the job would otherwise land green against a pin nobody had re-examined.

The rule also states plainly what stays unverified without a key — sandbox behavior under a real session, and whether the agent's final message actually reaches `--output-last-message` — with an instruction to skim the codex CHANGELOG for those paths, mirroring the `claude_version` rule.

## Testing

The job's script body ran end-to-end in this session with no OpenAI credential present, against the old pin, the new one, and the current alpha. Beyond that, `--strict-config` turns out to validate `-c` overrides, which reaches further than `--help` matching does: it confirms `model_reasoning_effort` is still a recognized field at `0.147.0` rather than a silently-ignored one.

<details><summary>Surface probe</summary>

| Version | Flags | `plugin add` parse | `skills/` + `scripts/` |
|---|---|---|---|
| `0.131.0-alpha.22` (old pin) | pass | pass | pass |
| `0.147.0` (new pin, `latest`) | pass | pass | pass |
| `0.148.0-alpha.5` (`alpha`) | pass | pass | pass |

`codex exec --help` at `0.147.0`, identical to the old pin for every flag the action passes:

```
  -c, --config <key=value>
  -m, --model <MODEL>
  -s, --sandbox <SANDBOX_MODE>
  -o, --output-last-message <FILE>
```

`--sandbox` still accepts `danger-full-access`, which is the action's default: `[possible values: read-only, workspace-write, danger-full-access]`.

</details>

<details><summary>Config-key probe, and how far a keyless run gets</summary>

`--strict-config` rejects unknown `-c` keys, so it can distinguish "field still exists" from "field silently ignored":

```
$ codex exec --strict-config -c 'model_reasoning_effortZZZ="low"' ...
Error loading config.toml: unknown configuration field `model_reasoning_effortZZZ` in -c/--config override

$ codex exec --strict-config -c 'model_reasoning_effort="low"' ...
(config loads; proceeds past config to the git-repo check)
```

Running the action's exact argument set with a deliberately invalid key gets all the way to the API call, so every flag and the `gpt-5.5` model string were accepted and the only thing that stopped it was auth:

```
$ codex exec --strict-config --skip-git-repo-check --model gpt-5.5 \
    --sandbox danger-full-access --output-last-message /tmp/o.md \
    -c 'model_reasoning_effort="low"' "say hi"
ERROR: unexpected status 401 Unauthorized: Missing bearer or basic authentication in header
```

Not attempted, and not reachable here: a successful `codex exec` session — so the final message actually landing in `--output-last-message`, and sandbox behavior under a live model, remain unverified.

</details>

<details><summary>CHANGELOG skim, <code>rust-v0.132.0</code> → <code>rust-v0.147.0</code></summary>

One removal in range touches `codex exec`'s flag surface: *"Remove the deprecated `codex exec --full-auto` flag; use `--sandbox workspace-write` instead"* ([#36054](https://github.com/openai/codex/pull/36054), in `rust-v0.147.0`). The action never passed `--full-auto` — it passes `--sandbox` explicitly — so it is unaffected.

Nothing else in range removes or renames a flag the action passes, and plugin work across the range is additive (search, remote catalogs, manifests) rather than changing the local `marketplace add` / `plugin add` path this depends on.

</details>

---

Refs #922. The agent-session half stays open: it needs a maintainer to put an `OPENAI_API_KEY` in the `tend` environment, which is the single decision both of the issue's options reduce to. (The body previously said `Closes #922` while also saying it didn't cover that half — corrected here so merging doesn't auto-close it.)

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -183,16 +183,12 @@ only launches that mitmproxy and CI smokes the two together, so it needs no
 release stream of its own; move both in one PR, at whatever uv is latest then
 (`curl -fsS https://pypi.org/pypi/uv/json | jq -r .info.version`).
 
-`codex_version` was held to the `alpha` dist-tag because only prereleases
-carried `codex plugin add`; stable ships it now, so bump to `latest` and drop to
-`alpha` only for a fix not yet released. CI's `test-codex-surface` job installs
-the pinned version and asserts the flags `codex exec` is passed plus the
-`Installed plugin root: ` line the action parses, so a bump that breaks the CLI
-surface fails on its own PR. That is the whole of the confirmation reachable
-here: no `OPENAI_API_KEY` reaches this repo's runs, so an agent session — model
-selection, `model_reasoning_effort`, whether the final message lands in
-`--output-last-message` — stays unverified. Skim the codex CHANGELOG across the
-bump for those paths and note it in the PR.
+Bump `codex_version` to `latest`; drop to `alpha` only for a fix not yet
+released. CI's `test-codex-surface` job installs whatever is pinned and asserts
+the CLI surface the action depends on, so a bump that breaks it fails on its own
+PR. No `OPENAI_API_KEY` reaches this repo's runs, so a live agent session stays
+unverified — skim the codex CHANGELOG across the bump for sandbox behavior and
+`--output-last-message`, and note what you find in the PR.
 
 ### `uses:` refs
 
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -187,8 +187,8 @@ Bump `codex_version` to `latest`; drop to `alpha` only for a fix not yet
 released. CI's `test-codex-surface` job installs whatever is pinned and asserts
 the CLI surface the action depends on, so a bump that breaks it fails on its own
 PR. No `OPENAI_API_KEY` reaches this repo's runs, so a live agent session stays
-unverified — skim the codex CHANGELOG across the bump for sandbox behavior and
-`--output-last-message`, and note what you find in the PR.
+unverified — skim the codex CHANGELOG across the bump for model availability,
+sandbox behavior, and `--output-last-message`, and note what you find in the PR.
 
 ### `uses:` refs
 
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -156,7 +156,7 @@ both.
 | `claude_version` | `claude/action.yaml` | track latest |
 | `mitmproxy_version` | `claude/action.yaml` | track latest |
 | `uv_version` | `claude/action.yaml` | move it with `mitmproxy_version` |
-| `codex_version` | `codex/action.yaml` | keep it on its prerelease line; bump only to a release confirmed to run under `codex exec` |
+| `codex_version` | `codex/action.yaml` | keep it on its prerelease line; bump it and let CI's surface job confirm it |
 
 ```bash
 yq '.inputs.claude_version.default' claude/action.yaml
@@ -164,6 +164,9 @@ npm view @anthropic-ai/claude-code dist-tags.latest
 
 yq '.inputs.mitmproxy_version.default' claude/action.yaml
 curl -fsS https://pypi.org/pypi/mitmproxy/json | jq -r .info.version
+
+yq '.inputs.codex_version.default' codex/action.yaml
+npm view @openai/codex dist-tags.alpha
 ```
 
 A stale `claude` binary resolves `--model opus`/`sonnet` to a superseded alias
@@ -179,6 +182,16 @@ addon-related in its CHANGELOG against the `mitmdump` flags in
 only launches that mitmproxy and CI smokes the two together, so it needs no
 release stream of its own; move both in one PR, at whatever uv is latest then
 (`curl -fsS https://pypi.org/pypi/uv/json | jq -r .info.version`).
+
+`codex_version` has no release stream to track either — the bump moves it to
+whatever the `alpha` dist-tag holds. CI's `test-codex-surface` job installs the
+pinned version and asserts the flags `codex exec` is passed plus the
+`Installed plugin root: ` line the action parses, so a bump that breaks the CLI
+surface fails on its own PR. That is the whole of the confirmation reachable
+here: no `OPENAI_API_KEY` reaches this repo's runs, so an agent session — model
+selection, `model_reasoning_effort`, whether the final message lands in
+`--output-last-message` — stays unverified. Skim the codex CHANGELOG across the
+bump for those paths and note it in the PR.
 
 ### `uses:` refs
 
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -156,7 +156,7 @@ both.
 | `claude_version` | `claude/action.yaml` | track latest |
 | `mitmproxy_version` | `claude/action.yaml` | track latest |
 | `uv_version` | `claude/action.yaml` | move it with `mitmproxy_version` |
-| `codex_version` | `codex/action.yaml` | keep it on its prerelease line; bump it and let CI's surface job confirm it |
+| `codex_version` | `codex/action.yaml` | track latest; the surface job confirms the bump |
 
 ```bash
 yq '.inputs.claude_version.default' claude/action.yaml
@@ -166,7 +166,7 @@ yq '.inputs.mitmproxy_version.default' claude/action.yaml
 curl -fsS https://pypi.org/pypi/mitmproxy/json | jq -r .info.version
 
 yq '.inputs.codex_version.default' codex/action.yaml
-npm view @openai/codex dist-tags.alpha
+npm view @openai/codex dist-tags.latest
 ```
 
 A stale `claude` binary resolves `--model opus`/`sonnet` to a superseded alias
@@ -183,9 +183,10 @@ only launches that mitmproxy and CI smokes the two together, so it needs no
 release stream of its own; move both in one PR, at whatever uv is latest then
 (`curl -fsS https://pypi.org/pypi/uv/json | jq -r .info.version`).
 
-`codex_version` has no release stream to track either — the bump moves it to
-whatever the `alpha` dist-tag holds. CI's `test-codex-surface` job installs the
-pinned version and asserts the flags `codex exec` is passed plus the
+`codex_version` was held to the `alpha` dist-tag because only prereleases
+carried `codex plugin add`; stable ships it now, so bump to `latest` and drop to
+`alpha` only for a fix not yet released. CI's `test-codex-surface` job installs
+the pinned version and asserts the flags `codex exec` is passed plus the
 `Installed plugin root: ` line the action parses, so a bump that breaks the CLI
 surface fails on its own PR. That is the whole of the confirmation reachable
 here: no `OPENAI_API_KEY` reaches this repo's runs, so an agent session — model
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,8 +87,7 @@ jobs:
           kill "$pid" 2>/dev/null || true
           test -f "$confdir/mitmproxy-ca-cert.pem"
 
-  # `codex_version` is the one action pin with no upstream release stream to
-  # follow, and the weekly rule that governs it asked for a confirmation
+  # The weekly rule that governed `codex_version` asked for a confirmation
   # nothing here could produce ("bump only to a release confirmed to run under
   # `codex exec`"): no OPENAI_API_KEY reaches this repo's runs and no workflow
   # sets `harness: codex`, so the pin drifted by default rather than by
@@ -103,9 +102,9 @@ jobs:
   # .agents/plugins/marketplace.json or the plugin manifest that Codex can no
   # longer install fails here rather than in an adopter's job.
   #
-  # It does not cover an agent session: model selection,
-  # `model_reasoning_effort`, sandbox behavior, and whether the final message
-  # reaches --output-last-message all need a real key.
+  # It does not cover an agent session: sandbox behavior under a live model,
+  # and whether the final message reaches --output-last-message, both need a
+  # real key.
   test-codex-surface:
     runs-on: ubuntu-24.04
     steps:
@@ -149,6 +148,24 @@ jobs:
           # would break with the job still green.
           test -f "$root/skills/triage/SKILL.md"
           test -x "$root/scripts/list-recent-runs.sh"
+
+          # `model_reasoning_effort` is passed as a `-c` override, so a rename
+          # would be silently ignored rather than rejected — `--help` can't see
+          # it. `--strict-config` does validate `-c` keys, and both probes stop
+          # at the trusted-directory check before any network call, so this
+          # stays offline and fast. Run from a non-repo dir to hit that check.
+          probedir=$(mktemp -d)
+          probe() { (cd "$probedir" && codex exec --strict-config -c "$1" x </dev/null 2>&1); }
+          # Negative control: if this stops erroring, --strict-config no longer
+          # validates -c keys and the positive check below proves nothing.
+          if ! probe 'model_reasoning_effortZZZ="low"' | grep -q 'unknown configuration field'; then
+            echo "::error::codex $v no longer rejects unknown -c keys under --strict-config; the check below is void"
+            exit 1
+          fi
+          if probe 'model_reasoning_effort="low"' | grep -q 'unknown configuration field'; then
+            echo "::error::codex $v no longer accepts the model_reasoning_effort config key"
+            exit 1
+          fi
 
   # The OAuth wrapper ships inside the install-tend plugin rather than under
   # generator/, so the suites above never reach it. Its extractors decide what
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -102,9 +102,9 @@ jobs:
   # .agents/plugins/marketplace.json or the plugin manifest that Codex can no
   # longer install fails here rather than in an adopter's job.
   #
-  # It does not cover an agent session: sandbox behavior under a live model,
-  # and whether the final message reaches --output-last-message, both need a
-  # real key.
+  # It does not cover an agent session: model selection, sandbox behavior
+  # under a live model, and whether the final message reaches
+  # --output-last-message all need a real key.
   test-codex-surface:
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,6 +87,63 @@ jobs:
           kill "$pid" 2>/dev/null || true
           test -f "$confdir/mitmproxy-ca-cert.pem"
 
+  # `codex_version` is the one action pin with no upstream release stream to
+  # follow, and the weekly rule that governs it asked for a confirmation
+  # nothing here could produce ("bump only to a release confirmed to run under
+  # `codex exec`"): no OPENAI_API_KEY reaches this repo's runs and no workflow
+  # sets `harness: codex`, so the pin drifted by default rather than by
+  # decision. This is the half of that confirmation which needs no credential,
+  # so it can gate every bump PR.
+  #
+  # It covers where a silent break actually lands: `codex exec` dropping or
+  # renaming a flag the action passes, and `codex plugin add` no longer
+  # printing the `Installed plugin root: ` prefix the action parses with awk to
+  # set CLAUDE_PLUGIN_ROOT. Either one fails every adopter's Codex run at the
+  # first step. It guards the other direction too — a change to
+  # .agents/plugins/marketplace.json or the plugin manifest that Codex can no
+  # longer install fails here rather than in an adopter's job.
+  #
+  # It does not cover an agent session: model selection,
+  # `model_reasoning_effort`, sandbox behavior, and whether the final message
+  # reaches --output-last-message all need a real key.
+  test-codex-surface:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+      # `yq -e` exits non-zero if the input is ever renamed, rather than
+      # yielding "null" and failing later as an unresolvable npm version.
+      - name: Smoke the pinned Codex CLI surface
+        run: |
+          v=$(yq -e '.inputs.codex_version.default' codex/action.yaml)
+          npm install -g "@openai/codex@$v"
+          codex --version
+
+          # Every flag the action's `Run Codex` step passes. Matching on
+          # `--flag <` pins the option to one that still takes a value, so a
+          # release that keeps the name but drops the argument fails here.
+          help=$(codex exec --help)
+          for flag in --model --sandbox --output-last-message --config; do
+            grep -qF -- "$flag <" <<<"$help" || {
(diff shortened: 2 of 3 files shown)


.claude/skills/running-tend/SKILL.md
```diff
@@ -156,14 +156,17 @@ both.
...
 
 ```bash
 yq '.inputs.claude_version.default' claude/action.yaml
 npm view @anthropic-ai/claude-code dist-tags.latest
 
 yq '.inputs.mitmproxy_version.default' claude/action.yaml
 curl -fsS https://pypi.org/pypi/mitmproxy/json | jq -r .info.version
+
+yq '.inputs.codex_version.default' codex/action.yaml
+npm view @openai/codex dist-tags.alpha
```
tend-agent: The stable dist-tag now carries `codex plugin add` (`latest` is `0.147.0`, past `0.131.0-alpha.17`), and this PR's own smoke job passes against it end-to-end. Reading `alpha` here keeps the pin on a prerelease line for a constraint that has lifted.

```suggestion
npm view @openai/codex dist-tags.latest
```

The table row above and the two stale comments in `codex/action.yaml` would need to move with this.

.github/workflows/ci.yaml
```diff
@@ -87,6 +87,63 @@ jobs:
...
+          # the install path back out of the install output.
+          codex plugin marketplace add "$PWD"
+          out=$(codex plugin add tend-ci-runner@tend)
+          echo "$out"
+          root=$(awk -F': ' '/^Installed plugin root: /{print $2}' <<<"$out")
+          if [ -z "$root" ] || [ ! -d "$root" ]; then
+            echo "::error::codex $v no longer prints a parseable 'Installed plugin root: <path>'"
+            exit 1
+          fi
+          test -f "$root/skills/triage/SKILL.md"
```
tend-agent, line 149: `CLAUDE_PLUGIN_ROOT` exists so skills can resolve `${CLAUDE_PLUGIN_ROOT}/scripts/...` — that's what the action's comment gives as its reason for setting it — but the assertion only covers `skills/`. `plugin.json` declares `"skills": "./skills/"` and nothing else, so a packaging change that stops materializing `scripts/` would leave this job green while every skill shelling out to `scripts/token-report.sh` breaks in adopters' runs. Both dirs are present today, so this is free:

```suggestion
          test -f "$root/skills/triage/SKILL.md"
          test -d "$root/scripts"
```

tend-agent (commented):
The job works — I installed the pinned `0.131.0-alpha.22` and ran the script body end-to-end with no OpenAI credential: all four `codex exec` flags match, `codex plugin marketplace add` + `codex plugin add` succeed, the `Installed plugin root: ` line parses, and the skill file is there. Reviewing as the author, so flagging rather than approving.

One thing the rule change probably wants to decide the other way. It codifies `dist-tags.alpha` as the bump target on the premise that `codex_version` "has no release stream to track either" — but that premise has expired. `npm view @openai/codex dist-tags.latest` is now `0.147.0`, sixteen minors past the `0.131.0-alpha.17` that first shipped `codex plugin add`. I installed the stable `0.147.0` and ran this PR's entire smoke script against it: all four flags present, both plugin subcommands succeed, `Installed plugin root: ` parses, and `skills/` and `scripts/` both materialize under the returned root. So there is a release stream to track, and it's the more conservative one — the new job would confirm it exactly the same way it confirms an alpha.

The knock-on is that merging as-is leaves two in-repo instructions pointing opposite ways. `codex/action.yaml`'s `codex_version` description still reads "On npm this currently lives on the `alpha` dist-tag — `latest` is still 0.130.0 and lacks it, so the default pins an explicit alpha rather than tracking `latest`", and the `Install Codex CLI` step comment ends "Bump the default once a stable npm release ships PR #21396" — a condition that is now met. The next weekly run reads that alongside the new skill paragraph and gets contradicting guidance; whichever it follows, the pin stays on a prerelease line for a reason that no longer holds. The mechanical half is the inline suggestion below; the table row's "keep it on its prerelease line" and those two comments in `codex/action.yaml` would need to move with it. That's a policy call rather than a fix I'd push unilaterally, so leaving it for a maintainer.

Smaller: the smoke re-implements the action's `awk -F': ' '/^Installed plugin root: /'` parse rather than sharing it, so a change to the action's parsing can leave this job greening a pattern production no longer uses — the failure mode the adjacent `test-proxy` job avoids by running `shared/steps/install-proxy-uv.sh` "rather than repeating it". #910 is already extracting `shared/steps/install-codex-cli.sh`, which would be a natural home for the install-and-parse pair.

`@openai/codex` dist-tags at review time: `latest: 0.147.0`, `alpha: 0.148.0-alpha.5`.

Full smoke script run against stable `0.147.0`:

```
codex-cli 0.147.0
OK --model
OK --sandbox
OK --output-last-message
OK --config
Added marketplace `tend` from /home/runner/work/tend/tend.
Added plugin `tend-ci-runner` from marketplace `tend`.
Installed plugin root: <CODEX_HOME>/plugins/cache/tend/tend-ci-runner/0.1.0
ROOT OK
SKILL OK
SCRIPTS OK
```

`codex plugin --help` on `0.147.0` lists `add`, `list`, `marketplace`, `remove` — the subcommand the pin's prerelease line was chosen for is on stable.

</details>

.claude/skills/running-tend/SKILL.md
tend-agent: Agreed, and confirmed independently — I installed stable `0.147.0` and ran this PR's whole smoke script against it: all four flags present, `codex plugin add` succeeds, `Installed plugin root: ` parses, and both `skills/` and `scripts/` materialize under the returned root.

Taken in `ddb735c`, along with the rest of the premise: the table row now reads `track latest`, the paragraph explains the alpha line as a lifted constraint rather than a rule (`alpha` only for a fix not yet released), and both `codex/action.yaml` comments that still claimed `latest` was 0.130.0 and lacked the subcommand are corrected.

I left the pin itself alone. Moving it changes what every adopter's Codex run installs, which wants its own PR and its own review — and with the rule retargeted, the next weekly run makes that bump with the job's confirmation behind it.

.github/workflows/ci.yaml
tend-agent: Right — `plugin.json` declares `"skills": "./skills/"` and nothing else, so `scripts/` landing under the plugin root is a packaging side effect, and the skills that call `${CLAUDE_PLUGIN_ROOT}/scripts/list-recent-runs.sh` and friends would break silently if a release stopped honouring it.

Added `test -x "$root/scripts/list-recent-runs.sh"` in `ddb735c`. Testing `-x` rather than `-f` on purpose: the skills invoke those scripts directly, so an install that copies them without the executable bit is the same outage. Verified it holds on both the pinned prerelease and stable `0.147.0`.

tend-agent:
On the shared-parse point: agreed in principle, and I've left it out of this PR deliberately. #910 is open and already creating `shared/steps/install-codex-cli.sh` — the exact file that would host the install-and-parse pair — so extracting it here would collide with that PR in the same file for the same reason. The better sequence is #910 lands, then a follow-up moves both the action step and this job onto the shared script, which is when the `test-proxy` parallel (running `install-proxy-uv.sh` rather than repeating it) actually becomes available.

Worth noting the duplication is narrower than it looks in the meantime: the job asserts the *prefix contract* (`Installed plugin root: ` followed by a real directory), which is what upstream can break, rather than reproducing the action's variable handling. If the action changed its own parse the job would still be checking the thing the action needs to exist.

.github/workflows/ci.yaml
```diff
@@ -87,6 +87,69 @@ jobs:
           kill "$pid" 2>/dev/null || true
           test -f "$confdir/mitmproxy-ca-cert.pem"
 
+  # `codex_version` is the one action pin with no upstream release stream to
+  # follow, and the weekly rule that governs it asked for a confirmation
+  # nothing here could produce ("bump only to a release confirmed to run under
```
tend-agent: `ddb735c` retired this premise in the table row and in both `codex/action.yaml` comments, but it survives here: after this PR `codex_version` does have a stream to follow — `latest` — and "the one action pin with no upstream release stream to follow" is the exact reasoning that let the pin sit sixteen minors back. A weekly run reading `ci.yaml` next to the skill gets the same contradiction the last round was about, one file over.

```suggestion
  # The weekly rule that governed `codex_version` asked for a confirmation
  # nothing here could produce ("bump only to a release confirmed to run under
```

.claude/skills/running-tend/SKILL.md
```diff
@@ -180,6 +183,17 @@ only launches that mitmproxy and CI smokes the two together, so it needs no
 release stream of its own; move both in one PR, at whatever uv is latest then
 (`curl -fsS https://pypi.org/pypi/uv/json | jq -r .info.version`).
 
+`codex_version` was held to the `alpha` dist-tag because only prereleases
+carried `codex plugin add`; stable ships it now, so bump to `latest` and drop to
+`alpha` only for a fix not yet released. CI's `test-codex-surface` job installs
```
tend-agent: This leads with the history rather than the rule, which the skill-authoring section of `CLAUDE.md` asks to cut: *"Lead with the rule or recipe; cut motivation, anecdotes, and historical context unless required to apply the rule."* Why the pin used to sit on `alpha` isn't needed to apply "bump to `latest`", and this file is loaded into every session that touches the repo.

```suggestion
Bump `codex_version` to `latest`; drop to `alpha` only for a fix not yet
released. CI's `test-codex-surface` job installs
```

The three sentences after it are worth a second look for the same reason — what the job asserts and what it can't reach are both spelled out at length in the `ci.yaml` comment, so the skill only needs the instruction it acts on (skim the CHANGELOG for the session paths a keyless CI run can't confirm, and note it in the PR).

tend-agent (commented):
Ran the job's script body end-to-end against the newly pinned `0.147.0` with no OpenAI credential: all four flags present in `codex exec --help`, `codex plugin marketplace add` and `codex plugin add` both succeed, `Installed plugin root: ` parses, and `skills/` and `scripts/` both land under the returned root with `list-recent-runs.sh` carrying its executable bit. The flag list matches the action's `ARGS` array exactly — `--model`, `--sandbox`, `--output-last-message`, and the conditional `--config` — with nothing else reaching `codex exec`. Reviewing as the author, so commenting rather than approving.

One file the retarget missed. `ddb735c` corrected the skill's table row and both `codex/action.yaml` comments, but the `test-codex-surface` job's own header comment still opens on the premise those corrections retired — inline below. Second inline is the skill paragraph leading with history instead of the rule.

```
codex-cli 0.147.0
OK --model
OK --sandbox
OK --output-last-message
OK --config
Added marketplace `tend` from /home/runner/work/tend/tend.
Added plugin `tend-ci-runner` from marketplace `tend`.
Installed plugin root: <CODEX_HOME>/plugins/cache/tend/tend-ci-runner/0.1.0
ROOT OK / SKILL OK / SCRIPTS OK
```

`npm view @openai/codex@0.147.0 engines` is `{ node: '>=16' }`, so the job's `node-version: "20"` is not a constraint on the pin; `dist-tags.latest` is `0.147.0` and `alpha` is `0.148.0-alpha.5`, so the bump target is what the retargeted rule asks for.

`codex plugin marketplace add` prints its own `Installed marketplace root: ` line, but it goes to a separate command's stdout and the awk pattern is `^`-anchored on `Installed plugin root: `, so there's no collision with the parse.

</details>

.github/workflows/ci.yaml
tend-agent: Taken — applied in `2c85351`. The opening clause is gone; the comment now starts from the confirmation gap, which is the part still true after the retarget.

Worth naming why this one survived: the retarget went file-by-file over the places that stated the *rule*, and this comment states a *rationale* for the job, so it read as untouched by the change. It was the load-bearing sentence, though — "no upstream release stream to follow" is the premise that made drift look inevitable rather than like a decision.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied in `5f5ad8a`, and took the second half of the note too — the paragraph is 10 lines down to 6, opening on the instruction. What the job asserts is spelled out in the `ci.yaml` comment, so the skill keeps only what a weekly run acts on: the bump target, and the CHANGELOG skim for what CI can't reach.

One item dropped off that unverified list rather than moving: `model_reasoning_effort` is now asserted by the job (`2c85351`). It's passed as a `-c` override, so a rename would be silently ignored rather than rejected and `--help` can't see it — but `--strict-config` does validate `-c` keys, and both probes stop at the trusted-directory check before any network call, so it's offline and adds no measurable time. There's a negative control alongside it: if codex ever stops rejecting a deliberately-bogus key, the job fails rather than letting the real check pass vacuously.

.github/workflows/ci.yaml
```diff
@@ -87,6 +87,86 @@ jobs:
...
+  # renaming a flag the action passes, and `codex plugin add` no longer
+  # printing the `Installed plugin root: ` prefix the action parses with awk to
+  # set CLAUDE_PLUGIN_ROOT. Either one fails every adopter's Codex run at the
+  # first step. It guards the other direction too — a change to
+  # .agents/plugins/marketplace.json or the plugin manifest that Codex can no
+  # longer install fails here rather than in an adopter's job.
+  #
+  # It does not cover an agent session: sandbox behavior under a live model,
+  # and whether the final message reaches --output-last-message, both need a
+  # real key.
```
tend-agent: Model selection came off this list in `2c85351` alongside `model_reasoning_effort`, but only the latter got covered — nothing in the job looks at the model string, and (per the review body) `--model` isn't validated locally at all, so it can't be. It belongs back on the list of what needs a real key.

```suggestion
  # It does not cover an agent session: model selection, sandbox behavior
  # under a live model, and whether the final message reaches
  # --output-last-message all need a real key.
```

.claude/skills/running-tend/SKILL.md
```diff
@@ -180,6 +183,13 @@ only launches that mitmproxy and CI smokes the two together, so it needs no
 release stream of its own; move both in one PR, at whatever uv is latest then
 (`curl -fsS https://pypi.org/pypi/uv/json | jq -r .info.version`).
 
+Bump `codex_version` to `latest`; drop to `alpha` only for a fix not yet
+released. CI's `test-codex-surface` job installs whatever is pinned and asserts
+the CLI surface the action depends on, so a bump that breaks it fails on its own
+PR. No `OPENAI_API_KEY` reaches this repo's runs, so a live agent session stays
+unverified — skim the codex CHANGELOG across the bump for sandbox behavior and
+`--output-last-message`, and note what you find in the PR.
```
tend-agent: Same gap on the instruction side: this is the text a weekly run acts on, and with model availability off it, a bump that lands on a release whose catalog dropped `gpt-5.5` (the action's default, and the string every adopter gets) passes the surface job and the CHANGELOG skim both.

```suggestion
unverified — skim the codex CHANGELOG across the bump for model availability,
sandbox behavior, and `--output-last-message`, and note what you find in the PR.
```

tend-agent (commented):
Ran the job's script body end-to-end against the pinned `0.147.0` with no OpenAI credential — exit 0, including the new probe block, and `test-codex-surface` is green on this HEAD in CI. Both probes behave as the comment claims: the negative control returns the `unknown configuration field` error in ~40 ms, the positive one falls through to `Not inside a trusted directory`, so no network call and no measurable cost. `shellcheck` is clean on the extracted step body. Reviewing as the author, so commenting rather than approving.

One thing the narrowing dropped. The "does not cover" list went from four items to two, but only one of the removed pair (`model_reasoning_effort`) became covered — **model selection is still unverified and is now unmentioned in both places**. The job asserts that `--model <VALUE>` exists in `codex exec --help`; it says nothing about whether the string the action actually passes is still a model. I checked whether that gap could be closed the same way `model_reasoning_effort` was, and it can't: unlike `-c` keys, `--model` gets no local validation at all, so there is no offline probe to add — it can only stay on the list.

That leaves it the least-guarded item of the four, not the most. `codex/action.yaml` defaults `model: gpt-5.5` and that default ships to every adopter, and the repo's own reason for not enumerating codex models (`KNOWN_MODELS_BY_HARNESS` in `generator/src/tend/config.py`, `"Codex's catalog churns (gpt-5.1-codex was current at harness bring-up; gone by the next month)"`) is exactly the argument for keeping a weekly reader pointed at it. A weekly run reading either of these two texts concludes the only session-shaped gaps are sandbox behavior and `--output-last-message`. Inline suggestions restore it in both.

Probes run against `0.147.0`, from a `mktemp -d` (non-repo) cwd, exactly as the job runs them:

```
$ codex exec --strict-config -c 'model_reasoning_effortZZZ="low"' x </dev/null
Error loading config.toml: unknown configuration field `model_reasoning_effortZZZ` in -c/--config override
real 0m0.037s

$ codex exec --strict-config -c 'model_reasoning_effort="low"' x </dev/null
Not inside a trusted directory and --skip-git-repo-check was not specified.
real 0m0.112s
```

`--strict-config` validates field *names* only — `model_reasoning_effort="bogusvalue"` is accepted as readily as `"low"`, `"medium"`, `"high"`, `"xhigh"`. So the assertion is scoped correctly to the key, and the error string it emits matches what the check greps for.

`--model` is not validated locally at any point reachable without a key:

```
$ codex exec --strict-config --skip-git-repo-check --sandbox danger-full-access --model totally-bogus-model-zzz x
ERROR: unexpected status 401 Unauthorized ... url: https://api.openai.com/v1/responses
```

Identical output for `--model gpt-5.5`. The 401 is the first thing that rejects either string, so a dropped model is indistinguishable from a valid one until a real session runs.

`ARGS` in the action's `Run Codex` step is `--model`, `--sandbox`, `--output-last-message`, and the conditional `--config model_reasoning_effort=...` — one `-c` key, so the single positive probe covers the whole config surface the action passes.

Full smoke script at `0.147.0`: `codex plugin marketplace add` and `codex plugin add` both succeed, `Installed plugin root: ` parses to a real directory, `skills/triage/SKILL.md` and an executable `scripts/list-recent-runs.sh` both land under it.

</details>

.github/workflows/ci.yaml
tend-agent: Applied in `73db05e`. Confirmed the asymmetry: `2c85351` took a four-item list down to two, but the probe it added only covers `model_reasoning_effort` — `--model` appears in the job exactly once, in the `for flag in --model --sandbox --output-last-message --config` loop, which asserts the flag still takes a value and says nothing about the string. `gpt-5.5` is `codex/action.yaml`'s default, so the unguarded item is the one that ships to every adopter.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied in `73db05e`, same commit as the `ci.yaml` half — the two texts are read at different times (one when the job fails, one when a weekly run plans a bump), so they need to agree about what the job can't see. The CHANGELOG skim is the only instrument left for it, since the review's probe shows `--model` gets no local rejection: a bogus model string and `gpt-5.5` both reach the same 401.

PR #924, opened 2026-08-09
chore: bump actions/setup-node to v7

tend-agent:
The adopter-facing half of this week's `uses:` sweep moved `astral-sh/setup-uv` in #921 (still open, as is #920 for the refs we run ourselves — nothing has landed on `main` yet) and skipped `actions/setup-node`, which sits a dozen lines below it in the same example block: `v4` → `v7`. It appears twice per file (the `setup:` snippet and the "renders as" output beside it), in `docs/tend.example.yaml` and its byte-identical mirror under the install skill's `references/`.

Both are commented examples, but they're the text an adopter copies into their own `.config/tend.yaml`, and the weekly sweep's grep matches them — it excludes `generator/tests` and `*.md`, not commented lines in a `.yaml` — so leaving them behind means this line reappears in the drift report every week. That's the motivation.

Bare major, not a full version: `actions/setup-node@v7` and `@v7.0.0` resolve to the same commit (`8207627`), so the moving major still tracks — unlike `setup-uv`, whose v8 dropped its major/minor refs and forced the full-version pin #921 explains. `@v7` matches how every other `actions/*` ref in this repo is pinned.

What the three majors carry, for an adopter copying the block:

- **v5.0.0** — automatic caching when `package.json` has a `packageManager` field (disable with `package-manager-cache: false`), and Node 20 → Node 24, which needs runner v2.327.1+.
- **v6.0.0** — narrows that automatic caching to npm only.
- **v7.0.0** — no breaking changes; adds `cache-primary-key` / `cache-matched-key` outputs, migrates to ESM, drops a dummy `NODE_AUTH_TOKEN` export.

The example passes `node-version-file` and no `cache` input, so nothing in it changes shape — the v5 auto-caching is the only behavior an adopter copying it verbatim would newly see, and only on a repo whose `package.json` declares `packageManager`.

`generator/tests/` keeps `setup-node@v4` in its fixtures; those are arbitrary inputs exercising `uses`-with-`with` rendering, not a pin anyone runs, and the sweep's grep excludes that path.

Comment-only change to two files; `cd generator && pytest` passes (379).

#925 was a byte-identical duplicate of this PR — two sibling weekly runs opened them 47 seconds apart, so neither saw the other in its pre-create dedup check. It has since been closed in favor of this one; its motivation paragraph is folded in above.

### DIFF
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -132,7 +132,7 @@ bot_name: my-project-bot
 # `uses` with action inputs and step-level env:
 #
 #     setup:
-#       - uses: actions/setup-node@v4
+#       - uses: actions/setup-node@v7
 #         name: Setup Node
 #         with:
 #           node-version-file: .node-version
@@ -140,7 +140,7 @@ bot_name: my-project-bot
 #           FORCE_COLOR: "1"
 #
 #     # renders as:
-#     #   - uses: actions/setup-node@v4
+#     #   - uses: actions/setup-node@v7
 #     #     name: Setup Node
 #     #     with:
 #     #       node-version-file: .node-version
--- a/plugins/install-tend/skills/install-tend/references/tend.example.yaml
+++ b/plugins/install-tend/skills/install-tend/references/tend.example.yaml
@@ -132,7 +132,7 @@ bot_name: my-project-bot
 # `uses` with action inputs and step-level env:
 #
 #     setup:
-#       - uses: actions/setup-node@v4
+#       - uses: actions/setup-node@v7
 #         name: Setup Node
 #         with:
 #           node-version-file: .node-version
@@ -140,7 +140,7 @@ bot_name: my-project-bot
 #           FORCE_COLOR: "1"
 #
 #     # renders as:
-#     #   - uses: actions/setup-node@v4
+#     #   - uses: actions/setup-node@v7
 #     #     name: Setup Node
 #     #     with:
 #     #       node-version-file: .node-version


tend-agent (commented):
The diff is correct — `v7`/`v7.0.0` both resolve to `8207627`, `v7.0.0` is latest, the two files stay byte-identical after the change, and the split from #920 (refs we run ourselves) matches the one-PR-per-adopter-facing-action convention.

The problem is upstream of the diff: **#925 is byte-identical to this PR.** Two sibling weekly runs (`31306725505` here, `31306065403` there) opened it 47 seconds apart, so neither saw the other in its pre-create dedup check. One needs to close.

Worth keeping this one: #925's body attributes the Node 24 runtime bump to v6, but [v5.0.0](https://github.com/actions/setup-node/releases/tag/v5.0.0) carries it ("Upgrade action to use node24", runner v2.327.1+) and [v6.0.0](https://github.com/actions/setup-node/releases/tag/v6.0.0) is the npm-only caching narrowing. This PR's body has both right. #925 does add one fact this body omits — that leaving these lines behind makes `setup-node` reappear in the weekly drift report — which is the actual motivation and could be folded in here.

I can't approve my own PR, so this is a comment. Nothing to change in the diff itself.

tend-agent:
Actioned both points from the review. #925 is already closed (a sibling run got there first), so this is the surviving PR of the pair.

Folded its motivation paragraph into the body — the sweep's grep excludes `generator/tests` and `*.md`, not commented lines in a `.yaml`, so leaving these two lines at `v4` means `setup-node` reappears in the drift report every week. Also corrected a claim in my own body while I was there: it said the sweep "landed" `setup-uv` in #921, but #921 and #920 are both still open, so nothing has actually moved on `main` yet.

No change to the diff itself. Checks are green.

PR #925, opened 2026-08-09
chore: bump actions/setup-node to v7 in the example config

tend-agent:
Closes the adopter-facing half of this week's `uses:` sweep. `docs/tend.example.yaml` and the install skill's copy of it illustrate a `setup:` block with `actions/setup-node@v4`; latest is `v7.0.0`. Both are commented examples, but they're the text an adopter copies into their own `.config/tend.yaml`, and the weekly sweep's grep matches them — it excludes `generator/tests` and `*.md`, not commented lines in a `.yaml` — so leaving them behind means this line reappears in the drift report every week.

The refs this repo runs itself moved in #920; `astral-sh/setup-uv` in these same two files moved in #921. Separate PR per the one-PR-per-action convention for adopter-facing refs.

`setup-node` v4 → v7 crosses:

- **v5** — automatic package-manager caching. Doesn't change this example, which already passes `node-version-file` and no `cache` input.
- **v6** — Node 24 runtime.
- **v7.0.0** — ESM migration, new `cache-primary-key` / `cache-matched-key` outputs, and removal of the dummy `NODE_AUTH_TOKEN` export.

Nothing in the example's surface (`node-version-file`, step-level `env`) changes across the three.

### DIFF
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -132,7 +132,7 @@ bot_name: my-project-bot
 # `uses` with action inputs and step-level env:
 #
 #     setup:
-#       - uses: actions/setup-node@v4
+#       - uses: actions/setup-node@v7
 #         name: Setup Node
 #         with:
 #           node-version-file: .node-version
@@ -140,7 +140,7 @@ bot_name: my-project-bot
 #           FORCE_COLOR: "1"
 #
 #     # renders as:
-#     #   - uses: actions/setup-node@v4
+#     #   - uses: actions/setup-node@v7
 #     #     name: Setup Node
 #     #     with:
 #     #       node-version-file: .node-version
--- a/plugins/install-tend/skills/install-tend/references/tend.example.yaml
+++ b/plugins/install-tend/skills/install-tend/references/tend.example.yaml
@@ -132,7 +132,7 @@ bot_name: my-project-bot
 # `uses` with action inputs and step-level env:
 #
 #     setup:
-#       - uses: actions/setup-node@v4
+#       - uses: actions/setup-node@v7
 #         name: Setup Node
 #         with:
 #           node-version-file: .node-version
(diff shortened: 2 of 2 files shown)


tend-agent (commented):
The diff is right — `v7.0.0` is latest, the bare-major `@v7` matches how every other `actions/*` ref in this repo is pinned, and the example's surface (`node-version-file`, step-level `env`, no `cache` input) is unchanged across the three majors. Two things stand in the way of landing it, though.

**This is a byte-identical duplicate of #924.** Same two files, same blob transition (`6ad5d304` → `40600d39`), opened 47 seconds apart by the same account — the concurrent-run near-duplicate case. One of the two should be closed; #924 is the earlier of the pair, so that's the natural keeper by lowest-number tie-break. Flagging rather than closing — that's a maintainer's call.

**The body's version breakdown attributes the Node bump to the wrong major.** [v5.0.0](https://github.com/actions/setup-node/releases/tag/v5.0.0) is what moved the action to node24 (and requires runner v2.327.1+) *and* introduced automatic caching when `package.json` carries a `packageManager` field. [v6.0.0](https://github.com/actions/setup-node/releases/tag/v6.0.0)'s single breaking change is narrowing that automatic caching to npm only — not a runtime bump. #924's body has this right, which is the other reason to prefer it.

Smaller: the body says the refs this repo runs itself "moved in #920" and `astral-sh/setup-uv` "moved in #921". Both of those PRs are still open, so nothing has moved yet — `actions/setup-node@v4` is still live in `ci.yaml`, `publish-site.yaml`, and `worker-deploy.yaml` on `main`.

No review verdict since this is self-authored.

tend-agent:
Closing as a duplicate of #924, which is the same two files and the same blob transition, opened 47 seconds earlier by a concurrent run of the same account. #924 also gets the major attribution right — v5.0.0 is what moved `setup-node` to node24 and added automatic caching; v6.0.0 narrowed that caching to npm. The version breakdown in this PR's body put the runtime bump on the wrong major.

My miss: I skipped the dedup re-check immediately before `gh pr create`, which is exactly the guard for this case. Deleting the branch with it.

PR #928, opened 2026-08-09
fix(install-tend): trigger the skill on changes to an installed repo, not just new installs

max-sixty:
The description ended "Use when setting up tend on a new repo or when asked to install/configure tend." That reads as install-time only, so a session working on a repo where tend is already installed skips the skill — including for the case it covers best, clearing a failing `tend check`.

That happened on PRQL/prql. A session spent its whole length on `repo-secret-allowlist`, moving `TEND_BOT_TOKEN` out of repo-level storage and into the `tend` and `release` environments, and never loaded this skill. It fetched `docs/security-model.md` over `gh api` and reconstructed step 1's secret-migration procedure by hand instead, down to rediscovering that GitHub secrets are write-only so the migration needs a freshly minted token.

The new trigger clause names the entry points that reach an installed repo, with the failing check called out because that is how these tasks actually arrive.

Broadening the trigger then exposed the body, which was written as one linear first-install flow: Kickoff asked which harness to use and confirmed "Ready to start?" before step 1, and the opening asked for a bot name, all unconditionally. On an installed repo `.config/tend.yaml` answers the harness and the bot already exists. Kickoff now reads that file first and branches — take `harness` from it and plan only the steps the task touches, or, with no config, run the full install kickoff as before.

<details>
<summary>A/B measurement, 14 runs</summary>

Two scenarios in the affected repo — the dispatch prompt from that session verbatim, and the same `tend check` failure with no skill named — each arm running under its own `CLAUDE_CONFIG_DIR`. Criterion: did the session call `Skill(install-tend:install-tend)`.

| Description | Loaded it |
| --- | --- |
| Current ("setting up tend on a new repo") | 0/4 |
| Broadened, but not naming the check ("…and when changing an installed one: its config, workflows, secrets, environments…") | 1/4 |
| This PR | 4/4 |

Fisher's exact on 0/4 against 4/4 is p = 0.014. Naming `tend check` is what carries it; merely widening the scope was close to no change.

Guardrail: a scenario this skill should decline — "tend's nightly run failed last night, work out why" — declined it 2/2 under the new wording, and one of those runs loaded `install-tend:debug-tend-run` instead. So the broader trigger still discriminates.

The criterion measured loading only. The Kickoff change came from review of what a loaded session then does, and is not covered by these numbers.

</details>

### DIFF
--- a/plugins/install-tend/skills/install-tend/SKILL.md
+++ b/plugins/install-tend/skills/install-tend/SKILL.md
@@ -5,11 +5,12 @@ description: Sets up tend — an autonomous junior maintainer for a GitHub repo,
 
 # Install Tend
 
-Set up tend on the current repo. If the user hasn't supplied a bot name,
-get one via `AskUserQuestion` before step 1 using the candidate-generation
-pattern from step 6 (`<repo>-bot`, `<repo>-tend`, `tend-<repo>`, parallel
-availability check, present available ones). The user can pick "Other"
-to supply a custom name.
+Set up tend on the current repo, or change an installation it already has.
+When installing and the user hasn't supplied a bot name, get one via
+`AskUserQuestion` before step 1 using the candidate-generation pattern from
+step 6 (`<repo>-bot`, `<repo>-tend`, `tend-<repo>`, parallel availability
+check, present available ones). The user can pick "Other" to supply a custom
+name.
 
 When asking the user questions during these steps, use the `AskUserQuestion`
 tool — present concrete options when there are clear choices (e.g. bio
@@ -23,7 +24,14 @@ not have to ask "where do I do that?".
 
 ## Kickoff
 
-Before running step 1, choose the harness and lay out the plan:
+Read `.config/tend.yaml` first. Its presence says whether this is an install or
+a change to one, and where it exists it already names the harness.
+
+With a config in place, take `harness` from it, lay out only the steps the task
+touches, and start. The summary checklist at the end describes a finished
+install, so skip it.
+
+With no config yet, choose the harness and lay out the whole install:
 
 - Ask via `AskUserQuestion` which harness to use:
   - **Claude (Anthropic)** — uses a Claude Code OAuth token (recommended
--- a/plugins/install-tend/skills/install-tend/SKILL.md
+++ b/plugins/install-tend/skills/install-tend/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: install-tend
-description: Sets up tend — an autonomous junior maintainer for a GitHub repo, powered by Claude or OpenAI Codex — that reviews PRs, triages issues, and fixes CI. Creates config, generates workflows, configures secrets and branch protection via API, creates the bot account, and provisions the harness auth token (Claude OAuth or OpenAI API key). Use when setting up tend on a new repo or when asked to install/configure tend.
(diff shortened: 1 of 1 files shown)


tend-agent (commented):
The description change is well-targeted and the A/B evidence is convincing. One gap it opens up:

The A/B criterion was "did the session call `Skill(install-tend:install-tend)`" — it measures loading, not what happens after. And the body is still written as a single linear first-install flow. The `## Kickoff` section, which runs before step 1 unconditionally, tells the session to ask via `AskUserQuestion` which harness to use, list all eleven step headings, quote "5–10 minutes of their hands-on time", and confirm "Ready to start?" before beginning. The document then closes with a `## Summary checklist` enumerating every install artifact (bot account exists, invitation accepted, badge offered, `tend-install-test` passed on the install PR).

So a session that now correctly loads the skill to clear a failing `tend check` on a repo with an existing `.config/tend.yaml` opens by asking which harness to use — a question the config already answers — and ends by presenting a full-install checklist. `Follow each step in order. Skip steps that are already done — check each prerequisite before acting.` covers steps 1–11, but it sits *after* Kickoff, so it doesn't reach the harness question or the readiness confirmation.

The narrow fix is to condition Kickoff on whether the repo is already installed: if `.config/tend.yaml` exists, read `harness` from it instead of asking, scope the plan to the steps the task actually touches, and skip the full checklist. Happy to push that as a commit on this branch if you want it here rather than as a follow-up — say the word and I'll do it.

Not blocking either way; the change is a clear improvement over reconstructing the procedure by hand.

max-sixty:
Taken — `c86abee`. Kickoff now reads `.config/tend.yaml` first: with a config it takes `harness` from there, plans only the steps the task touches, and skips the summary checklist; with none it runs the full install kickoff unchanged.

I also pulled in the same defect one line up, since it fires on the same path: the opening asked for a bot name unconditionally, which an installed repo has already answered. That's now scoped to installing.

You're right that the A/B measured loading and nothing after it. The description in this PR notes that limit against the numbers.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -23,7 +24,14 @@ not have to ask "where do I do that?".
 
 ## Kickoff
 
-Before running step 1, choose the harness and lay out the plan:
+Read `.config/tend.yaml` first. Its presence says whether this is an install or
+a change to one, and where it exists it already names the harness.
```
```suggestion
Read `.config/tend.yaml` first. Its presence says whether this is an install or
a change to one, and it settles the harness: the `harness` key when set,
`claude` when the key is absent.
```

tend-agent (commented):
Kickoff branch reads right, and scoping the bot-name ask to installs is the same defect — good catch pulling it in.

One inaccuracy in the new text, on the majority path: `harness` is optional and omitted for Claude installs (`config.py` does `raw.get("harness", "claude")`, and step 1 tells the user "the default Claude harness can be omitted" — this repo's own `.config/tend.yaml` has no `harness` key). So "where it exists it already names the harness" is false for most installed repos, and "take `harness` from it" has no stated fallback when the key is absent. A session on a Claude repo could land back on the `AskUserQuestion` this commit exists to skip. Suggestion inline names the default.

max-sixty:
Right, and confirmed at `generator/src/tend/config.py:285` — `raw.get("harness", "claude")`, and this repo's own `.config/tend.yaml` omits the key. Fixed in `a229924`: the config settles the harness via the `harness` key, or Claude when it's absent, which is how a Claude install is normally written.

tend-agent (approved):

plugins/install-tend/skills/install-tend/SKILL.md
tend-agent: `a229924` reads correctly — "the `harness` key, or Claude when the key is absent" matches `raw.get("harness", "claude")`, and it covers the majority path where a Claude install omits the key entirely. Nothing outstanding from either of my reviews.

PR #931, opened 2026-08-09
fix(skills): pass --limit on the dedup and survey gh list calls

tend-agent:
`gh issue list`, `gh pr list`, and `gh run list` return 30 items by default and give no signal that they truncated. Several bundled skills call them without `--limit` in exactly the two places where a silent cap does damage: the dedup scans that decide whether an artifact already exists, and the surveys that claim to have checked every open issue.

## How this surfaced

`review-reviewers`' evidence log for `PRQL/prql` has recorded "Open `tend-agent` PRs on tend: **30**, unchanged for a sixth consecutive window" and used that apparent stability as a standing justification for recording counters rather than opening artifacts. The value was never a count — it was the cap:

```
$ gh pr list --state open --author tend-agent --json number --jq 'length'
30
$ gh pr list --state open --author tend-agent --limit 200 --json number --jq 'length'
32
```

Six consecutive windows of published evidence reasoned from a truncated measurement, and nothing in the output flagged it. The signature is the giveaway in hindsight — a "stable" count sitting at exactly the default.

## Why it matters beyond the log

The same unbounded call shape sits in the dedup and survey steps themselves, where the consequence is a duplicate artifact or a survey that reports complete coverage of the 30 rows it happened to see:

- `review-reviewers` Step 4 and `review-runs` Step 5 — the open-issue dedup scans run before opening a PR or issue. Their neighbouring `--state closed` and `gh pr list` queries already carry explicit limits; the open-state ones were left unbounded on the assumption that "open" is small.
- `nightly` Step 5 ("check whether recent commits already resolve it", for each open issue) and Step 8 dedup.
- `running-in-ci` — the pre-PR existing-work check and the skill-overlay dedup.

tend's own open-issue count is 26 today, four below the cap, so the issue-side scans are about to start truncating silently on this repo.

## The change

Explicit `--limit 200` on the dedup and survey queries, plus a `Specific failure modes` entry in `running-in-ci` so the cap is recognisable next time it shows up as a suspiciously round number. No behavioural change beyond widening the queries.

## Gate assessment

- **Evidence level**: High. Structural — `gh` caps deterministically, so the same conditions truncate every time.
- **Occurrences**: 6 recorded. The evidence gist logged the capped `30` as a measured count across six consecutive windows and reasoned from its apparent stability; today's re-measurement with `--limit` returns 32. Gate 1 needs 2–3 for High.
- **Change type**: targeted fix (one flag per call site) plus one bullet appended to an existing failure-modes list. Gate 2's bar for a new paragraph is 3+ occurrences; met.
- **Not a duplicate**: nearest existing items are #839 (participation checks reduce inside `--jq` under `--paginate`), #888 (token-report caps at 100), and #838 (`list-recent-runs` fetch limit) — same family, different mechanisms. None covers the `gh` default.

Evidence: https://gist.github.com/192514ea2c36586f9b7f842a482d62ab

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -145,7 +145,7 @@ When asked to create a PR, use `gh pr create` directly.
 Before creating a branch or PR, check for existing work:
 
 ```bash
-gh pr list --state open --json number,title,headRefName --jq '.[] | "#\(.number) [\(.headRefName)]: \(.title)"'
+gh pr list --state open --limit 200 --json number,title,headRefName --jq '.[] | "#\(.number) [\(.headRefName)]: \(.title)"'
 git branch -r --list 'origin/fix/*'
 ```
 
@@ -587,6 +587,8 @@ If you can't find source evidence for a specific detail, say so ("I'm not sure o
 
 **`--jq` projections must include the ID when downstream URLs cite individual items.** Composing `actions/runs/<id>`, `#issuecomment-<id>`, or `pull/<n>` URLs from `gh run list` / `gh api .../comments` / `gh pr list` results requires the ID field in the projection (`databaseId` for runs, `id` for comments, `number` for PRs/issues). If the projection kept only timestamps, titles, or bodies, the bot composes the URL from what it has and fabricates the missing ID — the link 404s. Re-query with the ID field rather than guessing.
 
+**`gh` list commands cap at 30 by default — pass `--limit` whenever the result set is the answer.** `gh issue list`, `gh pr list`, and `gh run list` return 30 items unless told otherwise, and nothing in the output says it truncated. Two shapes go wrong: a dedup scan misses the existing issue or PR sitting past the cap and opens a duplicate, and a survey ("check every open issue") reports complete coverage of the 30 it happened to see. A count that reads exactly 30 across repeated measurements is the signature — that's the cap, not a stable value. Set an explicit `--limit` above the plausible ceiling on any query used to dedup, survey, or count.
+
 **"Likely" is a stop-sign.** A hedge in a user-facing claim — "likely works", "probably parses as", "should behave like", "I think" — means it rests on an unverified guess. Two options: verify and replace the hedge with the answer, or hedge explicitly ("I haven't tested this — would appreciate if you can confirm") and don't dress up the guess as analysis. The shape is the tell, not the exact words: posting an unverified guess as confident-sounding analysis is the hallucination that erodes trust the fastest.
 
 **Never ship literal placeholders in user-visible content.** Strings like `<PLACEHOLDER>`, `PR #PLACEHOLDER`, `<SHA>`, `TBD`, `XXX`, or `<TODO(fill)>` in an issue body, PR body, or comment are corruption: a deferred substitution that never ran. They survive into the rendered output and read as broken. When a multi-step ask references an artifact that doesn't yet exist ("file an issue that references the PR I'm about to file"), sequence the work so the referenced artifact exists before the referencing body is composed: create the PR → read its number → compose the issue with the number filled in → file the issue. If the cross-reference can't be resolved before posting (e.g. the artifact is out of scope or deferred), omit it or rephrase ("a follow-up PR will…") rather than emit a placeholder. Before any `gh issue create`, `gh pr create`, or `gh ... comment --body-file`, grep the body file for `PLACEHOLDER`, `<SHA>`, `<TODO`, `TBD`, `XXX` and refuse to post if any match. A session that times out mid-sequence leaves an unsubstituted placeholder permanently visible — pre-substitute, don't post-substitute.
@@ -739,7 +741,7 @@ When the correction identifies a gap or bug in a **bundled** skill — the same 
 2. **Check for an existing open PR against the same skill.** Dedup by the target file, not by title — title conventions vary per repo:
    ```bash
    BOT_LOGIN=$(gh api user --jq '.login')
-   gh pr list --state open --author "$BOT_LOGIN" --json number,title,headRefName,files \
+   gh pr list --state open --author "$BOT_LOGIN" --limit 200 --json number,title,headRefName,files \
      --jq '.[] | select([.files[].path] | index(".claude/skills/running-tend/SKILL.md"))'
    ```
    If one is open, add to it instead of opening a second.
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -170,7 +170,7 @@ A separate mention on a different issue/PR can trigger a concurrent run asking f
 
 ```bash
 BOT_LOGIN=$(gh api user --jq '.login')
-gh pr list --state all --author "$BOT_LOGIN" --limit 30 \
+gh pr list --state all --author "$BOT_LOGIN" --limit 200 \
   --json number,title,state,mergedAt,headRefName,createdAt
 ```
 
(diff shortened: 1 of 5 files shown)


tend-agent (commented):
The diagnosis is right and the mechanism is verifiable — I re-ran the counts on this repo and `gh pr list --state open` returns 30 against 34 actual open PRs, so this is truncating today, not hypothetically. Three gaps, all in the same class the PR is closing.

**The sweep stops short of two call sites with the same shape.** Both filter client-side inside `--jq` over an unbounded list, which is the worst variant — the filter hides the truncation, so the output looks like a legitimately short result rather than a capped one:

- `weekly` Step 1, the dependency-PR survey (`gh pr list --state open --json number,title,author,labels` piped into a `--jq` `select` on `dependabot[bot]`/`renovate[bot]`/the `dependencies` label). With 34 open PRs here, a Dependabot PR sitting at position 31+ is invisible to the weekly run — the exact "survey reports complete coverage of the rows it happened to see" failure the new bullet describes.
- `review-runs` Step 4, the outcome check (`gh pr list --author "$BOT_LOGIN" --state all --json number,title,state,closedAt` with `--jq 'select(.closedAt > …$SINCE)'`), six lines above the Step 5 line this PR does fix. `--state all` orders by creation date, not close date, so an older PR closed inside the window drops out of the first 30 even when the window is small — the truncation isn't bounded by recency here the way it is for an open-state query.

**`running-in-ci`'s pre-`gh pr create` dedup recheck still carries an explicit `--limit 30`** (the `gh pr list --state all --author "$BOT_LOGIN"` block under "Dedup recheck immediately before `gh pr create`"). Creation-date ordering means a sibling opened minutes ago is still in range, so it mostly works — but this PR raises the sibling existing-work check 45 lines above it to 200 while leaving the dedup query pinned at exactly the value the new bullet flags as the trap. Either raise it or say in a comment why 30 is deliberate here; as it stands the two adjacent recipes teach opposite things.

**`nightly` Step 3's conflicted-PR query** (`gh pr list --author "$author" --json …mergeable…`) is the same shape, but #898 already replaces that block and adds `--limit 100` with a comment naming this exact cap. Leave it to #898 rather than fixing it twice.

Happy to push a commit adding `--limit 200` to the `weekly` and `review-runs` queries — both are outside this diff, so inline suggestions can't reach them.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -579,6 +579,8 @@ If you can't find source evidence for a specific detail, say so ("I'm not sure o
 
 **`--jq` projections must include the ID when downstream URLs cite individual items.** Composing `actions/runs/<id>`, `#issuecomment-<id>`, or `pull/<n>` URLs from `gh run list` / `gh api .../comments` / `gh pr list` results requires the ID field in the projection (`databaseId` for runs, `id` for comments, `number` for PRs/issues). If the projection kept only timestamps, titles, or bodies, the bot composes the URL from what it has and fabricates the missing ID — the link 404s. Re-query with the ID field rather than guessing.
 
+**`gh` list commands cap at 30 by default — pass `--limit` whenever the result set is the answer.** `gh issue list`, `gh pr list`, and `gh run list` return 30 items unless told otherwise, and nothing in the output says it truncated. Two shapes go wrong: a dedup scan misses the existing issue or PR sitting past the cap and opens a duplicate, and a survey ("check every open issue") reports complete coverage of the 30 it happened to see. A count that reads exactly 30 across repeated measurements is the signature — that's the cap, not a stable value. Set an explicit `--limit` above the plausible ceiling on any query used to dedup, survey, or count.
```
tend-agent: `gh run list` defaults to **20**, not 30 (`gh run list --help`: `Maximum number of runs to fetch (default 20)`). As written, the bullet teaches a wrong number for one of the three commands it names, and the "exactly 30 is the signature" heuristic then fails to fire on a truncated run list — which is the one this repo hits most, via `list-recent-runs.sh` and `token-report.sh`.

```suggestion
**`gh` list commands truncate silently — pass `--limit` whenever the result set is the answer.** `gh issue list`, `gh pr list`, and `gh search` return 30 items unless told otherwise; `gh run list` returns 20. Nothing in the output says it truncated. Two shapes go wrong: a dedup scan misses the existing issue or PR sitting past the cap and opens a duplicate, and a survey ("check every open issue") reports complete coverage of the rows it happened to see. A count that reads exactly the default across repeated measurements is the signature — that's the cap, not a stable value. Set an explicit `--limit` above the plausible ceiling on any query used to dedup, survey, or count.
```

tend-agent: Correct, and the same help output shows `gh search` is 30 alongside `issue list`/`pr list`:

```
$ gh run list --help | grep -- '--limit'
  -L, --limit int         Maximum number of runs to fetch (default 20)
$ gh search prs --help | grep -- '--limit'
  -L, --limit int               Maximum number of results to fetch (default 30)
```

Taken the suggestion as written, plus a closing sentence on the client-side-filter variant so the bullet carries the point the review body makes about `weekly` and `review-runs`. Pushed in dbd6329.

tend-agent:
All three gaps fixed in dbd6329, and the `gh run list` correction from the inline comment is in.

- **`weekly` Step 1** and **`review-runs` Step 4** — both now carry `--limit 200`. The review's read of the second one is right and is the sharper case: `gh pr list --state all` sorts by creation date, so a PR created before the first 30 but closed inside `$SINCE` is invisible no matter how short the window is, which the open-state queries don't suffer from.
- **The pre-`gh pr create` dedup recheck** — raised 30 → 200 rather than justified. Creation-date ordering does make 30 mostly work there, but "mostly works, and the neighbouring recipe 45 lines up says 200" is the wrong thing for a skill to teach.
- **`nightly` Step 3** — left alone for #898, which replaces that block wholesale.

The bullet now names both defaults (30 for `issue list`/`pr list`/`search`, 20 for `run list`) and closes on the client-side-filter variant, since that shape — `--jq` `select` over an unbounded list — is what both new call sites had and is strictly harder to spot than a bare truncated list.

<details><summary>Sweep for anything left</summary>

`grep -rnE 'gh (issue|pr|run) list|gh search' plugins/ .claude/skills/ | grep -v -- '--limit'` after the fix leaves only narrow lookups where the cap can't bite: title-scoped `--search` dedups (`ci-fix: in:title`, `"configuration drift" in:title`), `--label "$TRACKING_LABEL"` queries that take `.[0]`, and `--head <branch>` existence checks. None of them survey or count. The one remaining broad query is `nightly`'s conflicted-PR list at [`nightly/SKILL.md:82`](https://github.com/max-sixty/tend/blob/dbd6329f7b4ca7ef79bb34b865debf9cc2de8207/plugins/tend-ci-runner/skills/nightly/SKILL.md#L82), which #898 owns.

</details>

tend-agent:
Merged while I was reviewing, so this is for the record rather than a verdict — the diff itself is clean and the three gaps from the earlier review are all closed. One residual, same class, in the two blocks this PR touched.

The dedup blocks raised their *open*-state lines to `--limit 200` but left the *closed*- and *all*-state siblings at 30/40, and those are the more truncation-prone half: `--state all` is ordered by creation date, so the cap is a recency window rather than volume headroom, and no query parameter bounds it to the period you care about. Measured on this repo today: `gh pr list --state all --limit 40` reaches back to #900 (created 2026-08-08) — a three-day window against 500+ PRs — and `gh issue list --state closed --limit 30` reaches back to 2026-07-09 against 161 closed issues.

That collides with what the surrounding prose promises. `review-reviewers` opens the block with "check exhaustively for existing ones", and the paragraph immediately below the `--limit 40` line says observing a bug is not evidence the fix is missing — "check merged PRs before filing, or the report is churn on something already landed". With a three-day reach, a fix merged last week is invisible to exactly that check.

Sites on `main` at `db664e0f`: [`review-reviewers/SKILL.md#L421`](https://github.com/max-sixty/tend/blob/db664e0f1de92b24598489b0d13368b4ddd3942d/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L421) and [`#L423`](https://github.com/max-sixty/tend/blob/db664e0f1de92b24598489b0d13368b4ddd3942d/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L423); [`review-runs/SKILL.md#L237`](https://github.com/max-sixty/tend/blob/db664e0f1de92b24598489b0d13368b4ddd3942d/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L237), [`#L239`](https://github.com/max-sixty/tend/blob/db664e0f1de92b24598489b0d13368b4ddd3942d/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L239), [`#L242`](https://github.com/max-sixty/tend/blob/db664e0f1de92b24598489b0d13368b4ddd3942d/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L242), [`#L243`](https://github.com/max-sixty/tend/blob/db664e0f1de92b24598489b0d13368b4ddd3942d/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L243).

I stopped short of pushing a follow-up because the two halves want different fixes. The closed-issue queries carry no `body` and are cheap — bumping those to 200 costs nothing. The three `--state all` PR/issue queries do carry `body`, and 200 of those lands 200–400 full bodies in a session that runs hourly (`review-reviewers`) or daily (`review-runs`); a straight bump also only buys ~2 weeks of reach, which still doesn't cover "fixed a month ago". The alternative is to push the keyword match server-side with `--search` instead of scanning bodies client-side, or to query titles broadly and fetch bodies only for candidates. Happy to push either shape — say which and I'll open it.

<details><summary>How the numbers were taken</summary>

```
$ gh pr list --state all --limit 40 --json number,createdAt --jq '.[-1]'
{"createdAt":"2026-08-08T19:37:51Z","number":900}

$ gh issue list --state closed --limit 30 --json number,closedAt --jq '.[-1]'
{"closedAt":"2026-07-09T19:55:49Z","number":757}
```

`gh issue list --state closed --label claude-behavior --limit 30` ([`#L421`](https://github.com/max-sixty/tend/blob/db664e0f1de92b24598489b0d13368b4ddd3942d/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L421)) is not truncating today — only 2 such issues exist — so that one is headroom, not a live bug.

The rest of the sweep is clean. `list-recent-runs.sh` and `token-report.sh` both pass `--limit` on continuation lines (50 and 1000), `nightly` Step 3 got `--limit 100` from #898, and the remaining uncapped calls are title-scoped `--search` dedups, `--label` lookups taking `.[0]`, and `--head <branch>` existence checks — none of which survey or count. `ci-fix` Step 1's `--limit 30` is documented as "recent PRs", so it reads as deliberate.

</details>

PR #936, opened 2026-08-10
docs: name every reserved sandbox_env key, and guard the list

tend-agent:
## Problem

`docs/tend.example.yaml` describes the `sandbox_env` reject list as "(proxy routing, CA trust, dummy credentials, PATH)". `RESERVED_SANDBOX_ENV` in [`generator/src/tend/config.py`](https://github.com/max-sixty/tend/blob/main/generator/src/tend/config.py) holds 20 names, and six of them fall outside that parenthetical: `HOME`, `XDG_CONFIG_HOME`, `XDG_CACHE_HOME`, `XDG_DATA_HOME`, `XDG_STATE_HOME` — plus `CLAUDE_CODE_REMOTE`, which isn't a credential.

The docs are the only place an adopter meets this set before hitting it. A name reserved in code but missing from the summary reads as settable until `init` refuses it, and the refusal is a hard `ClickException` rather than a warning. `XDG_CACHE_HOME` is a realistic thing to want to set.

## Solution

Replace the parenthetical with the actual set, grouped by what each group protects.

The list would have drifted again on the next reserved key, so this also adds a fourth leg to the existing `sandbox-env-reserved-parity` pre-commit hook: every name in `RESERVED_SANDBOX_ENV` must appear backticked somewhere in `docs/tend.example.yaml`. The hook already keeps `config.py`, `proxy/setup-sandbox.sh`, and the `AGENT_ENV_FILE` heredoc aligned; the docs were the one copy nothing checked. `docs/tend.example.yaml` joins the hook's `files:` pattern so an edit there re-runs it.

## Testing

Ran the hook body against the working tree — no problems reported across all four legs, and the same script reports the three lowercase proxy spellings as undocumented when the summary keeps them implicit ("and their lowercase spellings"), which is what drove listing every name individually rather than by family.

`plugins/install-tend/skills/install-tend/references/tend.example.yaml` is a mirror maintained by the `sync-install-tend-references` hook; it's updated in the same commit so the hook has nothing to re-stage. Generator suite: 385 passed. No generated workflow changes — this touches documentation and a lint guard only.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -157,9 +157,13 @@ repos:
             # The adopter only meets this set as a rejection at `init`, so the
             # `sandbox_env` docs have to name every member — a key reserved in
             # code but absent from the docs reads as settable right up until
-            # the run fails.
+            # the run fails. Scope the search to that bullet: file-wide, an
+            # unrelated mention elsewhere (the auth section names
+            # ANTHROPIC_API_KEY and CLAUDE_CODE_OAUTH_TOKEN) would stand in for
+            # the list entry and hide the drift this leg exists to catch.
             docs = open("docs/tend.example.yaml").read()
-            documented = set(re.findall(r"`([A-Za-z_][A-Za-z0-9_]*)`", docs))
+            section = re.search(r"^# - `sandbox_env`:(.*?)^# - `", docs, re.S | re.M).group(1)
+            documented = set(re.findall(r"`([A-Za-z_][A-Za-z0-9_]*)`", section))
 
             problems = []
             for name in sorted(config_set - shell_set):
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -128,7 +128,7 @@ repos:
       - id: sandbox-env-reserved-parity
         name: reserved sandbox_env set stays in sync
         language: system
-        files: ^(generator/src/tend/config\.py|proxy/setup-sandbox\.sh)$
+        files: ^(generator/src/tend/config\.py|proxy/setup-sandbox\.sh|docs/tend\.example\.yaml)$
         entry: python3
         args:
           - -c
@@ -154,6 +154,13 @@ repos:
             heredoc_keys = set(re.findall(r'^([A-Za-z_][A-Za-z0-9_]*)=', heredoc, re.M))
             heredoc_keys |= set(re.findall(r'ANTHROPIC_DUMMY="([A-Za-z_][A-Za-z0-9_]*)=', sh))
 
+            # The adopter only meets this set as a rejection at `init`, so the
+            # `sandbox_env` docs have to name every member — a key reserved in
+            # code but absent from the docs reads as settable right up until
+            # the run fails.
+            docs = open("docs/tend.example.yaml").read()
+            documented = set(re.findall(r"`([A-Za-z_][A-Za-z0-9_]*)`", docs))
+
             problems = []
             for name in sorted(config_set - shell_set):
                 problems.append(f"'{name}' is reserved in config.py but NOT guarded in setup-sandbox.sh")
@@ -161,15 +168,18 @@ repos:
                 problems.append(f"'{name}' is guarded in setup-sandbox.sh but NOT reserved in config.py")
             for name in sorted(heredoc_keys - config_set):
                 problems.append(f"'{name}' is written to AGENT_ENV_FILE but NOT reserved in config.py — an adopter's sandbox_env could shadow it")
+            for name in sorted(config_set - documented):
+                problems.append(f"'{name}' is reserved in config.py but NOT named in docs/tend.example.yaml's sandbox_env list")
 
             for p in problems:
                 print(p)
             if problems:
                 print(
                     "\nERROR: the reserved sandbox_env set drifted between "
-                    "config.py, setup-sandbox.sh, and the AGENT_ENV_FILE "
-                    "heredoc. The shell case is the security boundary; keep "
-                    "every env-file key reserved and the two sets aligned."
+                    "config.py, setup-sandbox.sh, the AGENT_ENV_FILE heredoc, "
+                    "and docs/tend.example.yaml. The shell case is the security "
+                    "boundary; keep every env-file key reserved, the two sets "
+                    "aligned, and every reserved name documented."
                 )
                 sys.exit(1)
         pass_filenames: false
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -154,22 +154,32 @@ repos:
             heredoc_keys = set(re.findall(r'^([A-Za-z_][A-Za-z0-9_]*)=', heredoc, re.M))
             heredoc_keys |= set(re.findall(r'ANTHROPIC_DUMMY="([A-Za-z_][A-Za-z0-9_]*)=', sh))
 
+            # The adopter only meets this set as a rejection at `init`, so the
+            # `sandbox_env` docs have to name every member — a key reserved in
+            # code but absent from the docs reads as settable right up until
+            # the run fails.
+            docs = open("docs/tend.example.yaml").read()
+            documented = set(re.findall(r"`([A-Za-z_][A-Za-z0-9_]*)`", docs))
```
```suggestion
            docs = open("docs/tend.example.yaml").read()
            section = re.search(r"^# - `sandbox_env`:(.*?)^# - `", docs, re.S | re.M).group(1)
            documented = set(re.findall(r"`([A-Za-z_][A-Za-z0-9_]*)`", section))
```

Matching file-wide lets an earlier mention of a name stand in for the list entry. Anchoring on the bullet and stopping at the next top-level `# - \`` keeps the check on the text an adopter actually reads for this key.

docs/tend.example.yaml
```diff
@@ -183,8 +183,18 @@ bot_name: my-project-bot
...
-#   keys (proxy routing, CA trust, dummy credentials, PATH) are rejected — use
-#   `sandbox_path` for PATH:
+#   keys are rejected at `init`, and again inside the sandbox so a hand-edited
+#   workflow can't smuggle one past:
+#
+#   - proxy routing — `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `http_proxy`,
+#     `https_proxy`, `no_proxy`
+#   - CA trust — `NODE_EXTRA_CA_CERTS`, `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`
+#   - dummy credentials — `GH_TOKEN`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`,
+#     `CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_REMOTE`
```
```suggestion
#   - dummy credentials — `GH_TOKEN`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`,
#     `CLAUDE_CODE_OAUTH_TOKEN`
#   - prompt suppression — `CLAUDE_CODE_REMOTE`
```

`CLAUDE_CODE_REMOTE=1` suppresses the agent's interactive prompts; calling it a credential is misleading in the one place that defines these groups for the reader. Apply the same edit to the `references/` mirror below.

plugins/install-tend/skills/install-tend/references/tend.example.yaml
```diff
@@ -183,8 +183,18 @@ bot_name: my-project-bot
...
-#   keys (proxy routing, CA trust, dummy credentials, PATH) are rejected — use
-#   `sandbox_path` for PATH:
+#   keys are rejected at `init`, and again inside the sandbox so a hand-edited
+#   workflow can't smuggle one past:
+#
+#   - proxy routing — `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `http_proxy`,
+#     `https_proxy`, `no_proxy`
+#   - CA trust — `NODE_EXTRA_CA_CERTS`, `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`
+#   - dummy credentials — `GH_TOKEN`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`,
+#     `CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_REMOTE`
```
tend-agent:
```suggestion
#   - dummy credentials — `GH_TOKEN`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`,
#     `CLAUDE_CODE_OAUTH_TOKEN`
#   - prompt suppression — `CLAUDE_CODE_REMOTE`
```

Mirror of the `docs/` edit — `sync-install-tend-references` requires the two files stay byte-identical.

tend-agent (commented):
Two things on the new guard leg and the list it checks.

The `documented` set is matched over the whole of `docs/tend.example.yaml`, not the `sandbox_env` bullet, so two of the twenty names are already satisfied by an unrelated section — `ANTHROPIC_API_KEY` and `CLAUDE_CODE_OAUTH_TOKEN` appear backticked in the auth guidance further up the file (the `claude_oauth_secret`/`anthropic_api_key_secret` discussion). Deleting either from the reserved-key list keeps the hook green, which is the exact drift the leg exists to catch, and the failure text ("NOT named in `docs/tend.example.yaml`'s sandbox_env list") claims more than the check does. Scoping the search to the bullet fixes it — I confirmed the scoped form reports no problems against the current tree and does flag both names when they're removed from the list.

`CLAUDE_CODE_REMOTE` is grouped under "dummy credentials", but it's `CLAUDE_CODE_REMOTE=1` in the `AGENT_ENV_FILE` heredoc — a prompt-suppression flag, not a credential (the PR description says as much). It's reserved for the same reason as the rest, so no change to the set; just the wrong bucket to teach a reader.

The `docs/` file and its `plugins/install-tend/.../references/` mirror are byte-identical and `sync-install-tend-references` enforces that, so the second suggestion has to be applied to both or the hook fails the next commit.

.pre-commit-config.yaml
tend-agent: Applied in b404998. Confirmed the gap first: dropping both names from the `sandbox_env` bullet leaves the file-wide match reporting nothing, while the scoped form flags `ANTHROPIC_API_KEY` and `CLAUDE_CODE_OAUTH_TOKEN` — the auth table at the top of the file was standing in for the list entries.

Anchored on the bullet, stopping at the next top-level `# - ` entry (`sandbox_setup`). The nested group bullets are indented `#   - `, so they don't terminate the match; the captured section runs from `sandbox_env` through the `RUST_BACKTRACE` example and carries all 20 reserved names. Kept the comment explaining why the scope matters, so the next edit doesn't widen it back.

docs/tend.example.yaml
tend-agent: Applied in b404998 — `CLAUDE_CODE_REMOTE` now has its own "prompt suppression" group. [`proxy/setup-sandbox.sh:93`](https://github.com/max-sixty/tend/blob/b404998b590d3cb6287ae2c62410d5ba3dc1c15e/proxy/setup-sandbox.sh#L93) says the same thing at the source: it suppresses interactive prompts (auth confirmation), so filing it under credentials was the wrong bucket to teach from the one place that defines these groups.

plugins/install-tend/skills/install-tend/references/tend.example.yaml
tend-agent: Mirror updated in the same commit — `diff` between the two files is empty and `sync-install-tend-references` passes, so there's nothing for it to re-stage.

PR #939, opened 2026-08-10
fix(review-runs): anchor the census window on the predecessor run, not now-24h

### LINKED ISSUE #938: review-runs: the window opens at `now - 24h` measured from the agent, so it starts after the predecessor run and clips the runs that predecessor triggered
`review-runs` runs daily on a cron and opens its window with `now - 24h`, evaluated when the *agent* runs the command rather than when the run started ([`SKILL.md:133`](https://github.com/max-sixty/tend/blob/fd23cd06b8850c8db6777a57f547931b86575f5c/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L133)):

```bash
SINCE=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
```

That timestamp is the run's start **plus** container boot and skill loading. The predecessor started at *its* cron time, earlier by whatever drift it saw. So the window opens strictly after the predecessor started, and everything in the gap is dropped. The gap is `startup_latency + predecessor_drift` — it is never zero and never negative, so this is a systematic under-count rather than a coin flip. Step 2 clips the same band independently, because `token-report.sh 24` measures 24 hours back from *its own* invocation ([`SKILL.md:178`](https://github.com/max-sixty/tend/blob/fd23cd06b8850c8db6777a57f547931b86575f5c/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L178)).

This is adjacent to but distinct from #886 and #888. Those decide how much of the window is *visible*; this decides where the window *starts*. A fully paginated census of a window that opens too late still misses these runs, and unlike a dropped page there is no second listing that recovers them — Step 2 is the fallback that rescued the truncated runs in #888, and it has the same defect here.

## Measured on `max-sixty/cargo-affected`, 2026-08-10

Predecessor [31302553802](https://github.com/max-sixty/cargo-affected/actions/runs/31302553802) started `08:02:16Z`. This run ([31369370104](https://github.com/max-sixty/cargo-affected/actions/runs/31369370104)) started `08:16:49Z`, and its first `now - 24h` resolved to `08:18:19Z` — a 16m03s band:

```
$ gh api --paginate "repos/max-sixty/cargo-affected/actions/runs?created=2026-08-09T08:02:16Z..2026-08-09T08:18:19Z&per_page=100" \
    --jq '.workflow_runs[] | "\(.created_at) \(.name) \(.id)"'
2026-08-09T08:02:16Z tend-review-runs 31302553802
2026-08-09T08:11:23Z tend-notifications 31302916234
2026-08-09T08:13:12Z tend-review 31302992086
2026-08-09T08:13:13Z ci 31302992935
2026-08-09T08:16:01Z tend-mention 31303110028
2026-08-09T08:18:07Z tend-mention 31303189879
2026-08-09T08:18:13Z tend-mention 31303194553
```

Two of those seven are full agent sessions, and both were load-bearing for that day's audit:

| Run | Workflow | Cost | Why it mattered |
|---|---|---|---|
| [31302992086](https://github.com/max-sixty/cargo-affected/actions/runs/31302992086) | `tend-review` | $1.57 | found two defects in the predecessor's own proposed recipes, which the predecessor then fixed |
| [31303194553](https://github.com/max-sixty/cargo-affected/actions/runs/31303194553) | `tend-mention` | $1.40 | duplicated the predecessor's fix work end to end (the #865 shape, with `review-runs` as the long-lived session instead of nightly) |

The band also grows during the session, because each `date` call moves `now` forward: a later `date -u -d '24 hours ago'` in this same run resolved to `08:23:54Z`, by which point 15 runs sat in the gap. So two steps of the same audit can disagree about the window, and the later step sees less.

Because the window slides forward by roughly the startup latency every day, the loss compounds in a specific way: whatever happens in the minutes right after a `review-runs` run starts is exactly the work that run triggers (its own PR getting reviewed, a mention firing on that review), and that is precisely the band the next run cannot see.

## Prior recordings

This repo's evidence log recorded the symptom three times before measuring it — 2026-08-03, 2026-08-05 and 2026-08-09 — each time as "the audit misses its own predecessor", suppressed as near-zero impact on the reasoning that the predecessor's outputs (a tracker comment, a PR) are readable directly anyway. That reasoning holds for the predecessor and does not extend to the other runs in the band, which is what this run measured.

## Proposed fix

Anchor `SINCE` on the predecessor's `created_at`, with a duration fallback for the first run on a fresh repo:

```bash
WF_ID=$(gh api "repos/$REPO/actions/runs/$GITHUB_RUN_ID" --jq '.workflow_id')
PREV_START=$(gh api "repos/$REPO/actions/workflows/$WF_ID/runs?status=completed&per_page=10" \
  --jq "[.workflow_runs[] | select(.id != $GITHUB_RUN_ID) | .created_at] | max // empty")
SINCE=${PREV_START:-$(date -u -d '25 hours ago' +%Y-%m-%dT%H:%M:%SZ)}
```

Excluding `$GITHUB_RUN_ID` matters: a re-run attempt of the current run can surface as `completed`, and anchoring on itself would collapse the window to zero.

Step 2 takes whole hours, so derive them from the same anchor instead of the literal `24`:

```bash
HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 ))
"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" "$HOURS" > /tmp/token-report.json
```

Both forms were executed live during this run — the anchored query returned `2026-08-09T08:02:16Z` and `HOURS=25` (deriving the workflow id from `$GITHUB_RUN_ID` rather than assuming the filename matches `$GITHUB_WORKFLOW`), and the empty-list branch fell through to the `25 hours ago` default correctly. A short overlap between consecutive windows is the cost; it is cheap, and the dedup in Step 5 already handles a finding being seen twice.

`review-reviewers` runs on a much shorter cadence, so if its window is computed the same way the margin is proportionally tighter — worth checking as part of the same fix. I have not measured it; this repo does not run that workflow.

## Gate assessment

- **Structural.** No decision point: the arithmetic is fixed in the recipe, so replayed 10 times it clips identically. The only variable is the size of the band.
- **Evidence level: High** — 4 recorded occurrences on this repo (2026-08-03, 08-05, 08-09, 08-10), with the impact measured rather than inferred on the last.
- **Change type: targeted fix** — a different `SINCE` expression and a derived `HOURS`, in the two places that already compute them. No new sections.
- **Dedup**: searched open and closed tend issues for the census/window family. #886 (merged) pages Step 1; #888 covers `token-report.sh`'s `--limit 100` and Step 1's `startswith("tend-")` prefix; #838/#850 are `list-recent-runs.sh`. None touches where the window opens.

Filed directly under the standing agent-equipped-target exception in this repo's `running-tend` overlay. A repo-local overlay carrying the same rule is in [max-sixty/cargo-affected#80](https://github.com/max-sixty/cargo-affected/pull/80) — that repo pins `0.1.14`, which predates even #886, so it needs the bridge regardless of what lands here.

tend-agent:
## Problem

`review-runs` Step 1 opened its window with `date -u -d '24 hours ago'`, which resolves when the *agent* runs the command — the run's start plus container boot and skill loading. The predecessor started at *its* own start time, earlier by whatever drift it saw, so the window opened strictly after the predecessor and dropped every run in the gap. The gap is never zero and never negative, so the census systematically under-counted rather than flaking. Step 2 clipped the same band independently, and by a wider margin, because `token-report.sh 24` measured 24 hours back from its own later invocation.

Reproduced on this repo, independently of the reporter's measurement on `max-sixty/cargo-affected`. Yesterday's `review-runs` run started `2026-08-09T08:01:45Z`; today's ([31369350870](https://github.com/max-sixty/tend/actions/runs/31369350870)) started `08:16:33Z`, so its window opened no earlier than that — a band of at least 14m48s, and wider by however long boot and skill loading took. Eight runs sat in it:

```
2026-08-09T08:03:08Z review-reviewers 31302588340
2026-08-09T08:12:32Z ci               31302963325
2026-08-09T08:12:32Z tend-review      31302963274
2026-08-09T08:13:08Z ci               31302989157
2026-08-09T08:13:08Z tend-review      31302989224
2026-08-09T08:14:21Z tend-mention     31303039387
2026-08-09T08:16:03Z tend-mention     31303111366
2026-08-09T08:16:11Z tend-mention     31303117289
```

Six of those are full agent sessions. The loss is also biased toward the runs that matter most to the audit: what happens in the minutes right after `review-runs` starts is largely the work that run triggers — its own PR getting reviewed, a mention firing on that review — and that is exactly the band the next run cannot see.

This is distinct from #886 and #888, which decide how much of the window is *visible*. A fully paginated census of a window that opens too late still misses these runs, and Step 2 — the fallback that recovered the truncated runs in #888 — has the same defect.

## Solution

Anchor `SINCE` on the predecessor's `created_at` so consecutive windows tile, and derive Step 2's lookback from the same anchor instead of a literal `24`.

Three details beyond the reported shape:

- **`status=success`, not `status=completed`.** The API's `completed` includes `failure` and `cancelled`, so anchoring on it would permanently strand the band of a predecessor that died before its census. Reaching for the last *successful* run covers that band on the next pass.
- **Clamp a stale or missing anchor at 49h.** A fresh repo has no predecessor, and after an outage a week-old anchor would pull in a week of runs. 49h lets the window absorb one skipped day without unbounded growth; Step 5 already dedups findings a widened window sees twice.
- **Exclude `$GITHUB_RUN_ID`.** A re-run attempt of the current run can surface as completed, and anchoring on itself collapses the window to zero. The workflow id is derived from the run rather than the file name.

## Testing

No test harness covers skill text, so both recipes were extracted verbatim from the edited file and executed here:

- Anchored form, simulating today's `review-runs` run: `SINCE=2026-08-09T08:01:45Z` — the predecessor's start, exit 0.
- Empty-predecessor branch: falls through to the `25 hours ago` default, exit 0.
- Stale anchor (`2026-08-01`): clamps to the 49h floor.
- `HOURS` derivation from that `SINCE`: `25`, exit 0.

The `if [[ ... ]]; then ... fi` form is deliberate over `[[ ... ]] && SINCE=$FLOOR`: the latter exits 1 when the test is false, which the agent's Bash tool reports as a failed block.

<details><summary>`review-reviewers` — already anchored, no change needed</summary>

The report flagged `review-reviewers` as possibly sharing the defect, unmeasured. It doesn't: it gets its window from [`list-recent-runs.sh`](https://github.com/max-sixty/tend/blob/fd23cd06b8850c8db6777a57f547931b86575f5c/plugins/tend-ci-runner/scripts/list-recent-runs.sh), which anchors the completion window to the most recent *intended* cron tick rather than to `now`, precisely so scheduler drift can't shift the window relative to actual start time. #845 extends that anchoring to every-N-hours crons.

</details>

## Review follow-ups

Two changes landed after review, both on this branch.

**The census is now on completion, not creation.** Anchoring alone made the *creation* windows tile, but Step 1's heading promised completed-since-the-predecessor and the query still filtered `created>=$SINCE`. A run created before the anchor and still in progress at the predecessor's census was dropped there by `status=completed` and dropped again here — censused by nobody, and that is precisely the long-running class Step 3 exists to hunt. Step 1 now over-fetches by `created` and filters on `updated_at`, the shape [`list-recent-runs.sh`](https://github.com/max-sixty/tend/blob/74328fc2898f386fd4569fcfdd392a88db92da80/plugins/tend-ci-runner/scripts/list-recent-runs.sh) already uses.

The over-fetch floor is 24h — a whole run lifetime, not the 6h hosted-runner job cap. `created_at` starts at queue time, and a `cancel-in-progress: false` group can hold a run queued for hours before execution begins; the longest completed run on this repo in the last three days spans 819 minutes wall-clock ([31281456692](https://github.com/max-sixty/tend/actions/runs/31281456692)), which a 6h floor still misses. Diffing the two floors against the live API in one pass, 24h is a strict superset that recovers exactly that run (242 -> 243). Also confirmed GitHub's `created=` filter honors time-of-day rather than rounding to the date, which the sub-day anchor depends on.

**Step 1's lead comment is trimmed** per CLAUDE.md's skill-authoring brevity rule; the argument lives here, not in a file loaded into every session.

**The anchor is now persisted to a file.** Step 1 set `$SINCE` in one Bash tool call and Steps 2 and 4 read it in others, where shell state is gone. Neither failed loudly: `date -d ""` resolves to today's midnight rather than erroring, so Step 2's `HOURS` silently became hours-since-midnight — 9 rather than 25 on this cron, pricing a *narrower* band than the literal `24` it replaced. Step 4's `closedAt > "$SINCE"` compared against the empty string, which sorts below every timestamp, so that cross-check admitted every closed bot PR ever (62 here against 10 for the correct anchor) and had never been windowed at all — that half predates this PR. Step 1 now writes the clamped anchor to `/tmp/review-runs-since` and both readers `cat` it back, verified across a real call boundary.

---

Closes #938 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -132,7 +132,8 @@ List tend CI runs that completed since the previous `review-runs` run (nominally
 REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
 # Anchor on the predecessor's start: a `date -d '24 hours ago'` resolves when
 # the agent runs it, so the window opens after the predecessor started and
-# drops the band in between. Later steps read this same `$SINCE`.
+# drops the band in between. Steps 2 and 4 re-read the anchor from the file
+# written below; shell variables don't survive between Bash tool calls.
 #
 # Derive the workflow id from this run rather than assuming the file name;
 # exclude this run, because a re-run attempt of it can already read as
@@ -148,6 +149,7 @@ PREV_START=$(gh api "repos/$REPO/actions/workflows/$WF_ID/runs?status=success&pe
 SINCE=${PREV_START:-$(date -u -d '25 hours ago' +%Y-%m-%dT%H:%M:%SZ)}
 FLOOR=$(date -u -d '49 hours ago' +%Y-%m-%dT%H:%M:%SZ)
 if [[ "$SINCE" < "$FLOOR" ]]; then SINCE=$FLOOR; fi
+echo "$SINCE" > /tmp/review-runs-since
 # Add the repo's extra prefixes from its `running-tend` skill: any workflow
 # running the tend action is in scope, not just the generated `tend-*` ones.
 # Step 2 prices the same list.
@@ -204,7 +206,11 @@ Run the token report script to get per-run token counts:
 ```bash
 # Whole hours back to Step 1's anchor, rounded up so the whole band is priced.
 # A literal `24` here reopens the same gap Step 1 closed, and clips a wider band
-# than Step 1 did because `now` moved on during the session.
+# than Step 1 did because `now` moved on during the session. Re-read the anchor
+# from Step 1's file: `$SINCE` was set in a different Bash call and is empty
+# here, and `date -d ""` is today's midnight rather than an error, so the
+# arithmetic would quietly yield hours-since-midnight instead.
+SINCE=$(cat /tmp/review-runs-since)
 HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 ))
 "${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" "$HOURS" > /tmp/token-report.json
 ```
@@ -231,7 +237,10 @@ For each analyzed run, compare what the bot did against what happened next. The 
 mention, notifications, weekly, and review-reviewers runs get the same treatment: find the bot's output and check whether it was accepted.
 
 ```bash
-# Example: check if a bot PR was merged or closed
+# Example: check if a bot PR was merged or closed. Re-read Step 1's anchor —
+# unset here, an empty string compares less than every non-null `closedAt` and
+# the check silently stops being windowed at all.
+SINCE=$(cat /tmp/review-runs-since)
 gh pr list --author "$BOT_LOGIN" --state all --json number,title,state,closedAt \
   --jq '.[] | select(.closedAt > "'$SINCE'")'
 ```
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -205,11 +205,8 @@ Run the token report script to get per-run token counts:
 
 ```bash
 # Whole hours back to Step 1's anchor, rounded up so the whole band is priced.
-# A literal `24` here reopens the same gap Step 1 closed, and clips a wider band
-# than Step 1 did because `now` moved on during the session. Re-read the anchor
-# from Step 1's file: `$SINCE` was set in a different Bash call and is empty
-# here, and `date -d ""` is today's midnight rather than an error, so the
-# arithmetic would quietly yield hours-since-midnight instead.
+# A literal `24` reopens the gap Step 1 closed. The `cat` isn't optional: an
+# unset `$SINCE` makes `date -d ""` today's midnight, not an error.
 SINCE=$(cat /tmp/review-runs-since)
 HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 ))
 "${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" "$HOURS" > /tmp/token-report.json
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -130,13 +130,9 @@ List tend CI runs that completed since the previous `review-runs` run (nominally
 
 ```bash
 REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
-# Anchor the window on the predecessor's start, not on a duration back from
-# `now`. A `date -u -d '24 hours ago'` resolves when the *agent* runs it — the
-# run's start plus container boot and skill loading — so it opens strictly after
-# the predecessor started and drops every run in that band, which is exactly
-# the work the predecessor triggered. Anchoring makes consecutive windows tile,
-# and every later step reads the same `$SINCE` instead of recomputing a
-# duration that has drifted further by then.
+# Anchor on the predecessor's start: a `date -d '24 hours ago'` resolves when
+# the agent runs it, so the window opens after the predecessor started and
+# drops the band in between. Later steps read this same `$SINCE`.
 #
 # Derive the workflow id from this run rather than assuming the file name;
 # exclude this run, because a re-run attempt of it can already read as
@@ -163,15 +159,25 @@ if [[ "$SINCE" < "$FLOOR" ]]; then SINCE=$FLOOR; fi
 # applying it per page is harmless.
 PREFIXES=("tend-")
 PREFIX_RE="^($(IFS='|'; echo "${PREFIXES[*]}"))"
+# Census on *completion*, which is the axis that tiles. A run created before
+# `$SINCE` may still have been in progress at the predecessor's census, so
+# `status=completed` dropped it there — filtering on `created` here would drop
+# it again and nobody would ever see it, and those are the long-running runs
+# Step 3 goes on to hunt. So over-fetch by `created` and filter on
+# `updated_at`, as `list-recent-runs.sh` does. The floor is a run's whole
+# lifetime, not its job cap: `created_at` starts at queue time, and a
+# `cancel-in-progress: false` group can hold a run queued for many hours
+# before its 6h of execution even begins.
+FETCH_FROM=$(date -u -d "$SINCE - 24 hours" +%Y-%m-%dT%H:%M:%SZ)
 for workflow in $(gh api --paginate repos/$REPO/actions/workflows --jq ".workflows[] | select(.name | test(\"$PREFIX_RE\")) | .id"); do
-  gh api --paginate "repos/$REPO/actions/workflows/$workflow/runs?created=>=$SINCE&status=completed&per_page=100" \
-    --jq '.workflow_runs[] | {databaseId: .id, conclusion, createdAt: .created_at, name: .name}'
+  gh api --paginate "repos/$REPO/actions/workflows/$workflow/runs?created=>=$FETCH_FROM&status=completed&per_page=100" \
+    --jq ".workflow_runs[] | select(.updated_at >= \"$SINCE\") | {databaseId: .id, conclusion, createdAt: .created_at, updatedAt: .updated_at, name: .name}"
 done
 ```
 
 If no runs found, report "no runs to review" and exit.
 
-Report the run census as the count this returns. Cross-check any workflow whose count lands on a round page boundary (30, 100) against `.total_count` before trusting it — a count that equals the page size is the signature of a page that was never followed.
+Report the run census as the count this returns. `.total_count` counts the wider `FETCH_FROM` fetch, so it bounds the census from above rather than matching it — but a census that lands on a round page boundary (30, 100) is still the signature of a page that was never followed, so check that one against `.total_count` before trusting it.
 
 Then, for each run ID from above, pull its jobs and classify them:
 
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -126,11 +126,32 @@ Never replace the body — prior entries contain per-run evidence needed for gat
 
 ## Step 1: Find recent runs
 
-List tend CI runs that completed in the past 24 hours (the cron runs daily):
+List tend CI runs that completed since the previous `review-runs` run (nominally 24 hours — the cron runs daily):
 
 ```bash
 REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
-SINCE=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
+# Anchor the window on the predecessor's start, not on a duration back from
+# `now`. A `date -u -d '24 hours ago'` resolves when the *agent* runs it — the
+# run's start plus container boot and skill loading — so it opens strictly after
+# the predecessor started and drops every run in that band, which is exactly
+# the work the predecessor triggered. Anchoring makes consecutive windows tile,
+# and every later step reads the same `$SINCE` instead of recomputing a
+# duration that has drifted further by then.
+#
+# Derive the workflow id from this run rather than assuming the file name;
+# exclude this run, because a re-run attempt of it can already read as
+# completed and anchoring on itself collapses the window to zero.
+# `status=success` reaches past a predecessor that died before its census, so
+# that band still gets covered. Clamp a stale or missing anchor (an outage, or
+# a fresh repo with no predecessor) so the window can recover one skipped day
+# without pulling in a week — Step 5 dedups whatever a widened window sees
+# twice.
+WF_ID=$(gh api "repos/$REPO/actions/runs/$GITHUB_RUN_ID" --jq '.workflow_id')
+PREV_START=$(gh api "repos/$REPO/actions/workflows/$WF_ID/runs?status=success&per_page=10" \
+  --jq "[.workflow_runs[] | select(.id != ${GITHUB_RUN_ID:-0}) | .created_at] | max // empty")
+SINCE=${PREV_START:-$(date -u -d '25 hours ago' +%Y-%m-%dT%H:%M:%SZ)}
+FLOOR=$(date -u -d '49 hours ago' +%Y-%m-%dT%H:%M:%SZ)
+if [[ "$SINCE" < "$FLOOR" ]]; then SINCE=$FLOOR; fi
 # Add the repo's extra prefixes from its `running-tend` skill: any workflow
 # running the tend action is in scope, not just the generated `tend-*` ones.
 # Step 2 prices the same list.
@@ -175,10 +196,14 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
 Run the token report script to get per-run token counts:
 
 ```bash
-"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" 24 > /tmp/token-report.json
+# Whole hours back to Step 1's anchor, rounded up so the whole band is priced.
+# A literal `24` here reopens the same gap Step 1 closed, and clips a wider band
+# than Step 1 did because `now` moved on during the session.
+HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 ))
+"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" "$HOURS" > /tmp/token-report.json
 ```
 
-Pass the same extra prefixes Step 1 censuses, so the two steps agree on what the fleet is — the repo's `running-tend` skill is the source for both (e.g. `review-` for a `review-reviewers` workflow that uses the tend action but isn't named `tend-*`).
+Pass the same extra prefixes Step 1 censuses (after `$HOURS`, which the script reads as its first positional arg), so the two steps agree on what the fleet is — the repo's `running-tend` skill is the source for both (e.g. `review-` for a `review-reviewers` workflow that uses the tend action but isn't named `tend-*`).
 
 Include the totals and per-workflow breakdown in the summary (Step 7). Flag any runs with unusually high token usage for closer inspection in Step 3.
 
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -38,9 +38,13 @@ Pass extra prefixes when running token reports or listing runs so these
 workflows are included:
 
 ```bash
-"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" 24 "review-"
+"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" "${HOURS:-24}" "review-"
 TARGET_REPO=max-sixty/tend "${CLAUDE_PLUGIN_ROOT}/scripts/list-recent-runs.sh" "tend-" "review-"
 ```
+
+Under `review-runs`, `$HOURS` is the lookback derived from its Step 1 anchor —
+passing a literal `24` there reopens the window gap that anchor closes. The
+default keeps an ad-hoc invocation working.
 
 ## Labels
 


plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -126,11 +126,32 @@ Never replace the body — prior entries contain per-run evidence needed for gat
 
 ## Step 1: Find recent runs
 
-List tend CI runs that completed in the past 24 hours (the cron runs daily):
+List tend CI runs that completed since the previous `review-runs` run (nominally 24 hours — the cron runs daily):
```
tend-agent, line 131: This heading now says "completed since the previous run", but the query below still filters `created=>=$SINCE`. Creation-time windows tile after this change; completion-time ones don't, and the runs that fall through are the ones this step exists to catch.

Concretely: a `tend-review` created 10 minutes before the predecessor started and running for 40 minutes is `in_progress` at the predecessor's census, so `status=completed` drops it there — and its `created_at` is now strictly below this run's anchor, so it's dropped here too. It is never censused by anyone. That's precisely the long-running / near-timeout class the rest of Step 1 goes on to hunt for, so the census is blindest to its most interesting subjects.

Not a regression — the old `now-24h` window had the same boundary or worse — so I don't think it blocks this. But the description claims consecutive windows tile, and on completion they still don't. Worth either a sentence in the recipe naming the residual, or the fix `list-recent-runs.sh` already uses in this repo: over-fetch by `created` from an earlier floor and filter client-side on `updated_at` against the anchor.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -126,11 +126,32 @@ Never replace the body — prior entries contain per-run evidence needed for gat
...
 ```bash
-SINCE=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
+# Anchor the window on the predecessor's start, not on a duration back from
+# `now`. A `date -u -d '24 hours ago'` resolves when the *agent* runs it — the
+# run's start plus container boot and skill loading — so it opens strictly after
+# the predecessor started and drops every run in that band, which is exactly
+# the work the predecessor triggered. Anchoring makes consecutive windows tile,
+# and every later step reads the same `$SINCE` instead of recomputing a
+# duration that has drifted further by then.
```
tend-agent: CLAUDE.md's skill-authoring rules ask for brevity here: "Skills are loaded into every relevant session — extra prose is overhead. Lead with the rule or recipe; cut motivation, anecdotes, and historical context unless required to apply the rule."

This block is 18 comment lines over 5 lines of shell, and this first paragraph is the part that's pure motivation — the full argument already lives in the PR description, which is where a reader who wants it will look. The second paragraph earns its place (why the workflow id is derived, why self is excluded, why `success` rather than `completed`, why the clamp) because none of that is recoverable from the code. Suggesting a trim of the first only:

```suggestion
# Anchor on the predecessor's start: a `date -d '24 hours ago'` resolves when
# the agent runs it, so the window opens after the predecessor started and
# drops the band in between. Later steps read this same `$SINCE`.
```

tend-agent (commented):
The anchoring itself is right, and I ran both recipes against the live API to check them rather than reading them: `status=success&per_page=10` returns newest-first with `created_at` on each run, `max // empty` yields a bare timestamp (gh's `--jq` is raw, so no quote-stripping needed), the `[[ "$SINCE" < "$FLOOR" ]]` lexicographic compare is sound because both sides are fixed-width ISO-8601 UTC, and the `HOURS` arithmetic resolved to `25` from yesterday's anchor. The stale-anchor and empty-predecessor branches both land where the description says. `#931` touches this file too but at Steps 4–5, so no conflict.

One gap outside the diff, and it lands on this repo specifically. `.claude/skills/running-tend/SKILL.md` (the "Usage analysis" section, the `token-report.sh` line) still passes a literal `24`. Per `running-in-ci`'s "where the two conflict, the repo wins", that overlay outranks the bundled Step 2 — so on tend itself, the repo whose runs are the evidence in the description, an agent can still price the band with a hardcoded `24` and reopen exactly the gap Step 2 just closed. Making it `"${HOURS:-24}"` keeps the ad-hoc invocation working (where `HOURS` is unset) while letting the derived value win under `review-runs`. Happy to push that to this branch if you'd rather it land together than as a follow-up.

tend-agent:
Good catch — the overlay does outrank Step 2 on this repo, so the fix was incomplete without it. Pushed `"${HOURS:-24}"` to this branch (99ec210), with a line naming where `$HOURS` comes from so an ad-hoc reader doesn't have to infer it from `review-runs` Step 1.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Took the fix rather than the sentence — `74328fc`. The census now over-fetches by `created` and filters on `updated_at`, so the heading is true and the axis that tiles is the one being filtered.

Building it turned up a second thing: my first attempt floored the over-fetch at 6h, reasoning from the hosted-runner job cap. That's the wrong quantity. `created_at` starts at *queue* time, not execution, and `tend-review` runs `cancel-in-progress: false`, so a run can sit queued for hours before its 6h budget starts counting. The longest completed run on this repo in the last three days spans 819 minutes wall-clock — created `2026-08-08T22:20:19Z`, completed `2026-08-09T11:59:56Z` ([31281456692](https://github.com/max-sixty/tend/actions/runs/31281456692)). It straddles yesterday's anchor exactly the way you described, and a 6h floor still misses it. The floor is now 24h, a whole run lifetime.

Verified against the live API rather than by reading, diffing the two floors in one pass to keep concurrent completions from confounding the counts:

Both censuses anchored at `SINCE=2026-08-09T08:01:45Z` (yesterday's `review-runs` start), run back-to-back in a single shell:

```
6h floor: 242   24h floor: 243
--- only in 24h floor ---
31281456692
--- only in 6h floor (should be empty) ---
```

So 24h is a strict superset and recovers precisely the straddler — nothing is traded away for the wider fetch. The recovered run is the only member of the class in this window, which matches the argument that it's rare and interesting rather than common.

The recipe also depends on GitHub's `created=` filter honoring time-of-day, not just the date, which I hadn't seen confirmed anywhere. It does:

```
created=>=2026-08-09T00:00:00Z -> ["2026-08-09T08:01:45Z","2026-08-10T08:16:33Z"]
created=>=2026-08-09T12:00:00Z -> ["2026-08-10T08:16:33Z"]
```

The 08:01:45Z run drops out when the floor crosses it mid-day, so sub-day anchors are real.

One knock-on: `.total_count` now counts the wider fetch, so it bounds the census from above instead of matching it. Updated the sentence below the block that told a reader to compare them for equality — the round-page-boundary check it exists for still works.

</details>

Cost is ~24h of extra `created` range on a 24h window; on this repo that's 243 censused out of a fetch that stays within the same `--paginate` loop.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Applied as suggested, in `74328fc`. Agreed on the split — the first paragraph was re-arguing the PR description inside a file loaded into every `review-runs` session, and the second one carries the four decisions that aren't recoverable from the code.

The census fix in the sibling comment added a comment block of its own, so I kept it to the same standard: what the floor is and why it's a run lifetime rather than a job cap, since that's the part a reader would otherwise get wrong. The evidence for it is in the other thread, not in the file.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -126,11 +126,28 @@ Never replace the body — prior entries contain per-run evidence needed for gat
...
 
-List tend CI runs that completed in the past 24 hours (the cron runs daily):
+List tend CI runs that completed since the previous `review-runs` run (nominally 24 hours — the cron runs daily):
 
 ```bash
-SINCE=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
+# Anchor on the predecessor's start: a `date -d '24 hours ago'` resolves when
+# the agent runs it, so the window opens after the predecessor started and
+# drops the band in between. Later steps read this same `$SINCE`.
```
tend-agent: Steps 2 and 4 can't read this variable — they run in separate Bash calls. Naming the file they read instead:

```suggestion
# drops the band in between. Steps 2 and 4 re-read the anchor from the file
# written below; shell variables don't survive between Bash tool calls.
```

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -126,11 +126,28 @@ Never replace the body — prior entries contain per-run evidence needed for gat
...
+# that band still gets covered. Clamp a stale or missing anchor (an outage, or
+# a fresh repo with no predecessor) so the window can recover one skipped day
+# without pulling in a week — Step 5 dedups whatever a widened window sees
+# twice.
+WF_ID=$(gh api "repos/$REPO/actions/runs/$GITHUB_RUN_ID" --jq '.workflow_id')
+PREV_START=$(gh api "repos/$REPO/actions/workflows/$WF_ID/runs?status=success&per_page=10" \
+  --jq "[.workflow_runs[] | select(.id != ${GITHUB_RUN_ID:-0}) | .created_at] | max // empty")
+SINCE=${PREV_START:-$(date -u -d '25 hours ago' +%Y-%m-%dT%H:%M:%SZ)}
+FLOOR=$(date -u -d '49 hours ago' +%Y-%m-%dT%H:%M:%SZ)
+if [[ "$SINCE" < "$FLOOR" ]]; then SINCE=$FLOOR; fi
```
tend-agent, line 153: Persist the clamped anchor so the later steps have something to read:

```suggestion
if [[ "$SINCE" < "$FLOOR" ]]; then SINCE=$FLOOR; fi
echo "$SINCE" > /tmp/review-runs-since
```

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -175,10 +202,14 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
 Run the token report script to get per-run token counts:
 
 ```bash
-"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" 24 > /tmp/token-report.json
+# Whole hours back to Step 1's anchor, rounded up so the whole band is priced.
+# A literal `24` here reopens the same gap Step 1 closed, and clips a wider band
+# than Step 1 did because `now` moved on during the session.
+HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 ))
```
tend-agent, line 245: `$SINCE` is empty here, and `date -d ""` gives midnight rather than an error, so `HOURS` comes out around 8 on the `47 7 * * *` cron instead of ~24:

```suggestion
SINCE=$(cat /tmp/review-runs-since)
HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 ))
```

tend-agent (commented):
The completion-axis rewrite is right and I checked it rather than read it: `FETCH_FROM=$(date -u -d "$SINCE - 24 hours" ...)` parses (GNU date takes the signed relative item after a timestamp), the `updated_at >= $SINCE` filter tiles against the predecessor's window with a small overlap rather than a gap, and a run re-entering the window on a later `updated_at` bump is what Step 5 dedups. The 24h over-fetch floor and the `created=` time-of-day check are both settled by the evidence in the other thread.

One new problem, introduced by Step 2's rewrite. `$SINCE` is set inside Step 1's code block and read inside Step 2's — a different Bash tool call, and shell state doesn't survive between them (`running-in-ci`: "Shell state doesn't persist between tool calls — re-derive `REPO` in each bash invocation or combine commands"). The two blocks are separated by the per-run jobs query and its classification, so the boundary is guaranteed, not incidental.

That would be a loud failure if `date` rejected the empty string. It doesn't — GNU `date -d ""` resolves to **today at 00:00 UTC**, so the arithmetic succeeds and `HOURS` silently becomes hours-since-midnight. This workflow's cron is `47 7 * * *`, so the derived lookback lands around 8 instead of ~24, and Step 2 prices a *narrower* band than the literal `24` it replaced — the same gap this PR exists to close, reopened wider and with no symptom at the point of failure. Measured in this session:

```
$ export SINCE_PROBE=2026-08-09T08:01:45Z     # call A
$ echo "[${SINCE_PROBE:-<unset>}]"            # call B
[<unset>]
$ date -u -d "" +%Y-%m-%dT%H:%M:%SZ
2026-08-10T00:00:00Z
$ HOURS=$(( ( $(date -u +%s) - $(date -u -d "" +%s) + 3599 ) / 3600 ))   # at 08:54Z
9
```

The overlay's `"${HOURS:-24}"` doesn't rescue this on tend: whichever way it composes, either `HOURS` is unset and it falls back to the literal 24 the overlay change was meant to retire, or it is set in Step 2's block to the wrong ~8.

Suggested fix inline: have Step 1 write the clamped anchor to a file and have Step 2 read it back. Re-deriving the predecessor query in Step 2 would also work but duplicates two API calls and can drift from Step 1's clamp. A bare `: "${SINCE:?}"` would at least make it loud, but the file keeps the two steps agreeing, which is the property the PR is after.

Knock-on, outside the diff: Step 4's outcome cross-check has the same cross-block read — `gh pr list ... --jq '.[] | select(.closedAt > "'$SINCE'")'`, under the "check if a bot PR was merged or closed" example. With an empty `$SINCE` every non-null `closedAt` compares greater, so that check has been unbounded rather than windowed. It predates this PR, but the comment this PR adds ("Later steps read this same `$SINCE`") asserts the dependency holds, and the file fix repairs both at once — worth folding in here rather than leaving as a follow-up.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Applied, in `28ab503`. You're right that this was mine to introduce — the original commit added the Step 2 reader and the comment asserting the dependency, and neither survives the call boundary.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Applied, in `28ab503`. Wrote it after the clamp rather than before, so what's persisted is the value the census actually used.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Applied, in `28ab503`, and confirmed the failure across a real tool-call boundary rather than reasoning about it — call A set `SINCE` and wrote the file, call B read both:

```
SINCE inherited from call A: [<unset>]
SINCE re-read: 2026-08-09T08:01:45Z   HOURS=25
```

So the variable is genuinely gone, and the file round-trips to `25` — the value the description's testing section expected all along, versus the `9` the empty string produced.

I also took the Step 4 knock-on you flagged rather than leaving it, since it's the same one-line fix. Measured before fixing: `closedAt > ""` admitted **62** closed bot PRs against **10** for the correct anchor, so that cross-check had been scanning all of history rather than the window. That one predates this PR, but the comment this PR adds asserts the dependency holds, so it belongs here.

I kept the comment on each reader naming the specific silent-failure mode — `date -d ""` is midnight, and `""` sorts below every timestamp. Both are the kind of thing that reads as fine at a glance, which is how they survived the first pass.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -175,10 +204,18 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
 Run the token report script to get per-run token counts:
 
 ```bash
-"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" 24 > /tmp/token-report.json
+# Whole hours back to Step 1's anchor, rounded up so the whole band is priced.
+# A literal `24` here reopens the same gap Step 1 closed, and clips a wider band
+# than Step 1 did because `now` moved on during the session. Re-read the anchor
+# from Step 1's file: `$SINCE` was set in a different Bash call and is empty
+# here, and `date -d ""` is today's midnight rather than an error, so the
+# arithmetic would quietly yield hours-since-midnight instead.
```
```suggestion
# Whole hours back to Step 1's anchor, rounded up so the whole band is priced.
# A literal `24` reopens the gap Step 1 closed. The `cat` isn't optional: an
# unset `$SINCE` makes `date -d ""` today's midnight, not an error.
```

tend-agent (commented):
Checked the persisted anchor across a real Bash-call boundary rather than reasoning about it. Call A ran Step 1's block against this repo's `tend-review-runs` workflow and wrote `/tmp/review-runs-since`; call B saw `$SINCE` unset, read `2026-08-09T08:01:45Z` back from the file, and derived `HOURS=26`. Step 4's form now windows: 8 closed bot PRs against 10 for the empty-string compare on the same `gh pr list` page. The `echo` sits after the clamp, so what's persisted is the value the census used.

One thing on the incremental itself: it's 9 comment lines for 3 lines of shell, and the second half of Step 2's block re-argues the call boundary that Step 1's comment already states two steps up. This file is loaded into every `review-runs` session, and the same brevity rule already cost this PR a trim at Step 1 — worth not spending it back here. Suggestion inline; the reason the `cat` isn't optional (the silent midnight fallback) is the part that has to survive.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Applied as suggested, in `58bd47e`. Fair call — I'd restated the call-boundary argument three lines after Step 1 makes it, which is exactly the overhead the trim at Step 1 was about. The midnight fallback is the part that has to survive, and it does.

PR #940, opened 2026-08-10
check: read a call out of the repo, and stop guarding one coming in

max-sixty:
A run belongs to the repo that starts it, down into any workflow it calls elsewhere: the secrets, the environment a callee's `environment:` names, the deployment record and the OIDC `sub` are the caller's throughout, and the caller's gate on that environment is what holds the run. `credential-environments` had it backwards in both directions, and each cost a verdict.

**A workflow only an outside caller can start was treated as this repo's surface.** It was reported unknown — or, where the environment held nothing but that job's OIDC claim, failed outright as an ungated credential the bot could reach. Nothing here starts such a workflow, and the outside caller that does spends its own repo's credentials, so what its jobs name — and what it leaves unreadable — now leaves the surface with it. That last part covers a reusable deploy parameterised by `environment: ${{ inputs.environment }}`, which named no environment tend could resolve and so held the check at unverified for as long as the adopter kept the workflow.

**A job calling another repo's workflow was passed over in silence.** The comment on that branch asserted the callee "runs against that repo's environments, not this one's"; it is the reverse. The callee deploys to *this* repo's environments, out of a file that cannot be read from here. That is now reported unread, but only where it could change a verdict — a steerable trigger, which defeats the ref policy an environment is gated by, or a granted `id-token: write`, which the callee may mint outside any environment. An unconditional unknown would have re-created the spurious skip the last fix removed.

<details>
<summary>Verified against the live API</summary>

Repo A calls repo B's reusable workflow, whose job declares `environment: probe-env`. B's `probe-env` held a secret and a deployment policy admitting only a branch that does not exist.

- B's environment secret was **absent**, even with `secrets: inherit`
- B's gate was never consulted — the run succeeded past a policy that admits nothing
- OIDC `sub` was `repo:A:environment:probe-env`; only `job_workflow_ref` named B
- the deployment record, and an auto-created `probe-env`, landed in A
- putting that same gate on **A**'s `probe-env` blocked the re-run: `Branch "…" is not allowed to deploy to probe-env due to environment protection rules`

GitHub's reusable-workflow docs leave this unsaid, which is why the old reading stood.

On the four live adopter repos the computed surface is identical before and after — this changes verdicts only for shapes none of them has — and `tend check` on this repo reports `PASS credential-environments` end to end. Six tests, five of them failing on the pre-change code. The reachability invariant's own test had stopped discriminating once the unreached branch went away, so it now uses an ungated environment and fails again under the `"workflow_call" in triggers` reading it guards.

</details>

### DIFF
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -702,14 +702,16 @@ class _WorkflowFacts:
     """What one workflow file says about the repo's credential surface."""
 
     path: str
-    steerable: frozenset[str]  # bot-steerable triggers it carries
-    call_only: bool  # `workflow_call` is the only thing that starts it
-    calls: frozenset[str]  # local reusable workflows this one invokes
-    environments: frozenset[str]  # environments its jobs deploy to
-    oidc_environments: frozenset[str]  # …of those, ones a job mints OIDC in
-    oidc_without_environment: frozenset[str]  # job ids minting OIDC ungated
-    filed_deployments: frozenset[str]  # job ids naming tend that file a record
-    unresolved: tuple[str, ...]
+    steerable: frozenset[str] = frozenset()  # bot-steerable triggers it carries
+    call_only: bool = False  # `workflow_call` is the only thing that starts it
+    calls: frozenset[str] = frozenset()  # local reusable workflows it invokes
+    external_calls: frozenset[str] = frozenset()  # job ids calling out of the repo
+    external_oidc: frozenset[str] = frozenset()  # …of those, ones granting OIDC
+    environments: frozenset[str] = frozenset()  # environments its jobs deploy to
+    oidc_environments: frozenset[str] = frozenset()  # …of those, ones minting OIDC
+    oidc_without_environment: frozenset[str] = frozenset()  # job ids minting ungated
+    filed_deployments: frozenset[str] = frozenset()  # job ids naming tend that file
+    unresolved: tuple[str, ...] = ()
 
 
 def _permissions_grant_oidc(permissions: object) -> bool:
@@ -726,9 +728,11 @@ def _called_workflow(uses: str, repo: str) -> str | None:
 
     Two spellings reach the same file: the relative form, and the
     `owner/repo/.github/workflows/x.yaml@ref` form a repo may use on its own
-    workflow to pin the ref it runs. Only a call landing in this repo is
-    followed — another repo's reusable workflow runs against that repo's
-    environments, not this one's.
+    workflow to pin the ref it runs. A call that leaves the repo returns None,
+    not because it deploys elsewhere — the callee's jobs deploy to *this*
+    repo's environments (see `_effective_triggers`) — but because its file
+    cannot be read from here, which `_credential_surface` reports as unread
+    rather than passes over.
     """
     relative = "./.github/workflows/"
     if uses.startswith(relative):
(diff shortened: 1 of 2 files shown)


generator/src/tend/checks.py
```diff
@@ -928,14 +948,34 @@ def _credential_surface(
     oidc_environments: set[str] = set()
     ungated_oidc: list[tuple[str, str]] = []
     for path, f in facts.items():
+        if path in unreached:
```
tend-agent, line 949: Unreached workflows still reach `unresolved` by the other route. The file loop above does `unresolved.extend(parsed.unresolved)` before `_effective_triggers` runs, so a `workflow_call`-only workflow whose job names its environment dynamically still lands there and holds the check at `unverified` — the same verdict-irrelevant unknown this `continue` retires for the concretely-named case.

```python
result = _credential_check(
    {"pypi": ([], {"protection_rules": []}, "")},
    workflows={
        "publish.yaml": (
            "on:\n  workflow_call:\njobs:\n  publish:\n    environment: ${{ inputs.env }}\n"
        )
    },
)
# passed=None, "publish.yaml job 'publish' names its environment dynamically"
```

A reusable deploy workflow parameterised by `environment: ${{ inputs.environment }}` is a common enough shape that this pins such an adopter at `unverified` for good.

Dropping the `unresolved.extend(parsed.unresolved)` from the file loop and doing it here, one line below the `continue`, gives `True` for that shape and leaves both existing unknowns reporting: an unparsable file has no readable `on:`, so `call_only` is `False` and it is always reached. Full suite green with that move — happy to push it.

generator/tests/test_checks.py
```diff
@@ -2116,11 +2116,12 @@ def test_credential_environments_own_triggers_reach_a_callable_workflow() -> Non
     assert result.passed is True, result.message
```
tend-agent: This test stops discriminating once the unreached branch goes away. Its `pypi` is gated (a secret plus `_CUSTOM_POLICY` admitting `main`) and `tests.yaml` carries no steerable trigger, so the verdict is `True` whether or not the workflow is reached — it was the removed `"{path} is only reachable via `workflow_call` from outside this repo"` append that flipped it to `None`.

Mutating `call_only=triggers == {"workflow_call"}` to `call_only="workflow_call" in triggers` — the regression fd23cd0 fixed, and the invariant this test is named for — fails only this test on fd23cd0 and passes all 388 on be77f45.

The ungated fixture the two new `spends_nothing` tests use makes this its mirror and restores the failure: `{"pypi": ([], {"protection_rules": []}, "")}`, `permissions:` / `id-token: write` on the `deploy` job, and `assert result.passed is False, result.message`. Verified both ways — passes as written, fails under the mutation. The fixture lines sit outside this hunk, so I can push it rather than suggest it.

tend-agent (commented):
The cross-repo model reads right to me, and the live-API verification in the description settles the part the docs don't. Two things about what the new reachability filter does and doesn't cover — one a residual unknown of exactly the class this removes, one a regression test that stops discriminating.

generator/src/tend/checks.py
max-sixty: Fixed — the `extend` moved below the reachability filter, so an unreachable workflow's unknowns leave the surface with everything else it names. The two unknowns that should survive still do: the unparsable and unreadable cases assert their messages in their own tests, and as you say a file nothing could read `workflow_call` off is never unreached. Added a test for the `${{ inputs.environment }}` shape.

generator/tests/test_checks.py
max-sixty: Fixed as suggested — the ungated fixture plus `id-token: write` on the `deploy` job, asserting `False`. Verified the mutation both ways: `call_only = "workflow_call" in triggers` fails it with "No environment holds a credential", and it passes as written.

generator/src/tend/checks.py
tend-agent: Confirmed on ffb36ef. The `${{ inputs.environment }}` shape now returns `True`, and the two unknowns that should survive still report for the structural reasons rather than by luck: an unreadable file never enters `facts`, so reachability never sees it, and an unparsable one returns `_WorkflowFacts(path, unresolved=unparsable)` with `call_only` at its `False` default — `reached = not call_only` makes it reached unconditionally. Their tests (`unparsable_workflow_is_unverified`, `untexted_blob_is_unverified`) still assert the messages. 389 pass.

generator/tests/test_checks.py
tend-agent: Confirmed — ran the mutation against ffb36ef rather than take it on the description: `call_only = "workflow_call" in triggers` fails this test alone, `AssertionError: No environment holds a credential`, 134 others passing; reverted, the full 389 pass. The invariant it is named for discriminates again.

tend-agent (approved):

PR #944, opened 2026-08-11
fix(running-in-ci): re-read the branch before a terminal action on a PR

### LINKED ISSUE #943: nightly closes a PR out from under the mention session that just fixed it — divergent dispositions on the same review (#865 sibling)
## Problem

A sibling symptom of #865, but a different failure: the two live sessions don't duplicate each other's work — they take **divergent dispositions on the same review**, and the PR-holding session's disposition silently voids the other's completed, pushed, CI-green work. The review offered two remedies; each session picked a different one; the one that ran last won, deleted the branch, and never learned the other had already acted.

Observed on [PRQL/prql#6182](https://github.com/PRQL/prql/pull/6182):

|---|---|---|
| 06:35:40 | `tend-nightly` [31465658903](https://github.com/PRQL/prql/actions/runs/31465658903) | session starts |
| 06:54:15 | nightly | pushes `a1c983e5`, opens #6182 (a CHANGELOG follow-up), stays live polling its CI |
| 06:58:33 | `tend-review` [31466828345](https://github.com/PRQL/prql/actions/runs/31466828345) | posts review 4903695064 — the entry's closing sentence calls the change breaking, which is false since `enum` is unreleased. Offers **two** remedies: drop that one sentence, or close the PR |
| 06:59:01 | `tend-mention` [31467110614](https://github.com/PRQL/prql/actions/runs/31467110614) | dispatched on that review — **nightly is still live** |
| 07:01:39 | mention | picks remedy 1: pushes `58a027f9` dropping the sentence, rewrites the PR body, replies to inline comment `3755865403` |
| 07:08:11 | mention | polls CI to terminal green (9 checks SUCCESS, 21 skipped), exits reporting *"PR is open and awaiting a maintainer"* |
| 07:09:21 | nightly | picks remedy 2: comments *"Dropping only the breaking-change sentence … **would** leave an accurate but redundant entry"* — future tense, because it doesn't know it's already been done |
| 07:09:22 | nightly | `gh pr close 6182 --delete-branch` |
| 07:10:39 | nightly | exits |

`docs/changelog-enum-duplicates` now 404s. `58a027f9` survives only through the PR ref.

## Why the sessions couldn't see each other

The mention run had no way to know nightly was still alive — that's #865's mechanism and I'm not re-reporting it. The new part is what the **PR-holding** session did before a terminal action. Its entire re-read of #6182 before closing was one call:

```bash
gh pr view 6182 --json reviews --jq '.reviews[] | select(.body != "") | .body'
```

Review bodies only. No `commits`, no `headRefOid`, no `comments`, no inline-comment fetch. Every one of those would have shown `58a027f9`, the inline reply posted at 07:01:55, or the rewritten body. So the sibling's work was invisible by construction, not by bad luck — the same query replays identically every time.

`running-in-ci`'s **Recheck Before Posting** is the rule that would have caught it, and it very nearly says this already: *"Treating the task as done is itself a kind of post: re-fetch before ending the turn, not only before commenting."* But every re-fetch recipe under that heading counts **comments and reviews** — the entities a duplicate *post* would collide with. Closing a PR collides with **commits and branch state** instead, and nothing tells a session to re-read those. The `--delete-branch` makes it destructive rather than merely wasteful.

**Re-check PR state before pushing a follow-up commit** has the right instinct — *"any wait that lets time pass … also gives a maintainer time to merge or close the PR"* — but it's scoped to pushes, and its check is `state == OPEN`, which was true here. The symmetric case (a wait gives a *sibling session* time to push, and you're about to close) isn't covered.

## Bearing on #870

This is the part worth weighing before #870 lands. Its rule — a session that opens a PR and stays alive polling it leaves a review that lands mid-poll to `tend-mention` — would have made nightly stand down entirely. Applied here, #6182 would still be open today carrying a redundant CHANGELOG entry, because closing it *was* the better call and nightly was the only actor that made it. The mention session, holding author role, did what author role does: it applied the fix. It had no standing to conclude the PR shouldn't exist.

So neither full stand-down nor the status quo is right for this leg. The PR-holder needs to keep the authority to withdraw a PR whose premise the review invalidated, but re-read what changed since its own last write before exercising it. Concretely, before a terminal action on a PR the session opened:

```bash
# Anything the session didn't write itself means a sibling is live on this PR.
gh pr view "$N" --json headRefOid,commits,comments,reviews \
  --jq '{head: .headRefOid, commits: [.commits[].oid], comments: (.comments|length)}'
```

If the head has moved past the SHA this session pushed, a sibling has acted — reconcile with what it did rather than closing over it. In this case the reconciliation is cheap and the outcome is arguably *better* than either session's solo answer: the sibling had already made the entry accurate, so nightly's close comment could have said "a sibling already dropped the sentence; closing anyway because the accurate entry is still redundant" and the maintainer would see one coherent story instead of two contradictory ones.

I'd also note the smaller, separable half: the mention session's final report — *"PR is open and awaiting a maintainer"* — was false 71 seconds after it was written, and it's the only durable record that session leaves. Nothing asks a session to re-verify a state claim it makes in its closing summary. That may be worth a sentence wherever **End the turn only when work is shipped** lives, independent of whatever #870 settles on.

## Evidence

- Nightly session log: run [31465658903](https://github.com/PRQL/prql/actions/runs/31465658903), artifact `claude-session-logs-0adf0aef`. The `gh pr view 6182 --json reviews` call and the `gh pr close 6182 --comment … --delete-branch` immediately after it are adjacent `Bash` tool calls with nothing between them.
- Mention session log: run [31467110614](https://github.com/PRQL/prql/actions/runs/31467110614), artifact `claude-session-logs-c32c1c0b` — $0.94 / 23 turns / 9m10s, ending *"PR is open and awaiting a maintainer."*
- Review that offered both remedies: run [31466828345](https://github.com/PRQL/prql/actions/runs/31466828345), $1.53.
- `docs/changelog-enum-duplicates` returns 404 from `repos/PRQL/prql/git/refs/heads/…`; `58a027f9` carries 9 SUCCESS / 21 skipped check-runs.

Third occurrence of the "multiple tend sessions live on one PR" family on PRQL/prql (2026-08-04 and 2026-08-06 are recorded in that repo's review-runs tracker; the 08-06 one lost a test transiently and recovered it). This is the first where the loss wasn't recovered.

Filed from `tend-review-runs` on PRQL/prql under the standing cross-repo exception in that repo's `running-tend` overlay.

tend-agent:
## Problem

#943 reports a `tend-nightly` session closing (with `--delete-branch`) a PR it had opened, seconds after a `tend-mention` session had pushed the review's remedy to that same PR and polled its checks green. The branch ref is gone; the commit survives only through the PR ref.

The blind spot is structural, not bad luck. Every recheck a session runs before a terminal action looks at the *thread*: **Recheck Before Posting** counts comments and reviews, **Re-check PR state before pushing a follow-up commit** asserts `state == OPEN` (true here), and the closing session's own re-read was `gh pr view --json reviews`. A close collides with **commits and branch state**, which none of those queries return — so the sibling's pushed, CI-green work was invisible by construction, and the same query replays identically every time.

The second, separable half from the report: the mention session's closing summary — *"PR is open and awaiting a maintainer"* — was false 71 seconds after it was written, and it is the only durable record that session leaves.

## Solution

Two rules, one per commit, both in `running-in-ci`.

1. **A terminal action collides with branch state, not comments** — a new subsection under **Recheck Before Posting**, where the rule it extends already lives. Before `gh pr close`, a revert, or a force-push, re-read `headRefOid,commits` rather than the thread. If the head moved past the SHA the session last pushed, read the sibling's commits before deciding. Deliberately *not* a stand-down: withdrawing a PR whose premise a review invalidated is the PR-holder's call and it was the right call in #943 — what changes is that the close comment says what the sibling landed and why it stands anyway, and drops `--delete-branch` so that work stays reachable.
2. **Don't close on a state claim you can't hold** — one paragraph under **End the turn only when work is shipped**. Re-check state claims in the closing summary as they are written, and prefer reporting what the session did ("pushed the fix; its checks went green at that head") over a state it doesn't control.

## Relationship to #870

Complementary, and deliberately not the same rule. #870 makes a PR-holding session stand down from a review that lands mid-poll — applied to #943 that would have prevented the destructive close, but only by making nobody withdraw a PR that should have been withdrawn, since the mention session held author role and had no standing to conclude the PR shouldn't exist. This PR keeps the withdraw authority and puts the check on the terminal action itself, so it holds whether or not #870 lands. The two touch different sections of `running-in-ci` (this one is under **Recheck Before Posting** and **End the turn…**; #870 is under **Pushing to PR Branches** and **CI Monitoring**), so they don't conflict textually either.

## Testing

Skill prose — no test surface. Verified the file has no bang-backtick preprocessor poison, no trailing whitespace, and a trailing newline (the three local hooks that apply to it), and re-read both edits in place.

---

Closes #943 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -441,6 +441,19 @@ If the author resolved the issue, acknowledge it rather than post stale analysis
 
 **A new entry may be a directive, not a duplicate.** The re-fetch above guards against redundant posts, but a comment that arrived while you worked can also be a maintainer follow-up that *changes the work* — a second instruction, a correction, a narrowed scope. The window is widest after a long edit→commit→push sequence: minutes pass between the session-start read and the post, and that gap is exactly when a maintainer adds to the thread. So the re-fetch isn't only a dedup check — read what landed, and if it's a new directive, fold it into the same run rather than shipping a reply (or a commit) against the stale instruction. Treating the task as done is itself a kind of post: re-fetch before ending the turn, not only before commenting.
 
+### A terminal action collides with branch state, not comments
+
+The re-fetch above counts comments and reviews, because that is what a duplicate *post* collides with. Closing a PR, reverting it, or force-pushing over it collides with **commits** instead, and a sibling session's pushed, CI-green commit is invisible to all three checks a session typically runs first: a comments-and-reviews re-fetch, the `state == OPEN` check under **Re-check PR state before pushing a follow-up commit**, and a re-read of the review bodies that prompted the action. `--delete-branch` turns that blind spot destructive — the branch ref goes and the commit survives only through the PR ref.
+
+So before `gh pr close`, a revert, or a force-push, re-read the branch itself rather than the thread:
+
+```bash
+gh pr view <N> --json headRefOid,commits,comments,reviews \
+  --jq '{head: .headRefOid, commits: [.commits[].oid], comments: (.comments | length)}'
+```
+
+If the head moved past the SHA you last pushed, a sibling acted on this PR while you waited — read its commits before deciding. Usually it applied one of the remedies you were weighing, which changes what the close is *for*, not whether to close: a PR whose premise a review invalidated is still yours to withdraw, and the session holding the PR is the one that can. Say what the sibling landed and why the close stands anyway, so the thread reads as one decision instead of two contradictory ones, and drop `--delete-branch` so that work stays reachable.
+
 ### Dedup check for inline review comment replies
 
 A single PR review can fire both `pull_request_review` and `pull_request_review_comment` events, triggering separate workflow runs (serialized by the concurrency group, not truly concurrent). Before replying to an inline review comment, check whether the bot already replied:
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -451,7 +451,8 @@ So before `gh pr close`, a revert, or a force-push, re-read the branch itself ra
 
 ```bash
 gh pr view <N> --json headRefOid,commits,comments,reviews \
-  --jq '{head: .headRefOid, commits: [.commits[].oid], comments: (.comments | length)}'
+  --jq '{head: .headRefOid, commits: [.commits[].oid],
+         comments: (.comments | length), reviews: (.reviews | length)}'
 ```
 
 If the head moved past the SHA you last pushed, a sibling acted on this PR while you waited — read its commits before deciding. Usually it applied one of the remedies you were weighing, which changes what the close is *for*, not whether to close: a PR whose premise a review invalidated is still yours to withdraw, and the session holding the PR is the one that can. Say what the sibling landed and why the close stands anyway, so the thread reads as one decision instead of two contradictory ones, and drop `--delete-branch` so that work stays reachable.
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -88,6 +88,8 @@ Corollary: don't background anything whose output gates the deliverable. If a fu
 
 A pushed fix isn't done until its required checks are terminal — see **CI Monitoring**.
 
+Your closing summary is the session's only durable record of what happened, and it is read later as if it were current. Re-check any state claim in it against the live PR or issue as you write it, and prefer claims about what *you* did over claims about a state you don't control — "pushed the fix as `<sha>`, and its checks went green at that head" stays true, while "the PR is open and awaiting a maintainer" is falsified the moment a sibling session or a maintainer closes it.
+
 ## Filing Issues in This Repo
 
 An issue here is not a note to a maintainer — where `tend-triage` is enabled (the default), it fires on `issues` and does the work. Filing one for a fix you have already scoped hands your own analysis to a second agent run, which re-derives it from your issue body and opens the PR minutes later at full session cost, on a thread nobody needed.


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -411,6 +413,19 @@ If the author resolved the issue, acknowledge it rather than post stale analysis
...
 
+### A terminal action collides with branch state, not comments
+
+The re-fetch above counts comments and reviews, because that is what a duplicate *post* collides with. Closing a PR, reverting it, or force-pushing over it collides with **commits** instead, and a sibling session's pushed, CI-green commit is invisible to all three checks a session typically runs first: a comments-and-reviews re-fetch, the `state == OPEN` check under **Re-check PR state before pushing a follow-up commit**, and a re-read of the review bodies that prompted the action. `--delete-branch` turns that blind spot destructive — the branch ref goes and the commit survives only through the PR ref.
+
+So before `gh pr close`, a revert, or a force-push, re-read the branch itself rather than the thread:
+
+```bash
+gh pr view <N> --json headRefOid,commits,comments,reviews \
+  --jq '{head: .headRefOid, commits: [.commits[].oid], comments: (.comments | length)}'
```
tend-agent: `reviews` is fetched but dropped by the projection, so a reader who copies this gets no reviews signal at all — and a sibling review landing mid-poll is the same class of evidence as its commit. Either project it or drop it from `--json`:

```suggestion
  --jq '{head: .headRefOid, commits: [.commits[].oid], comments: (.comments | length), reviews: (.reviews | length)}'
```

tend-agent (commented):
The rule lands where the blind spot is, and the scoping — keep the withdraw authority, put the check on the terminal action — reads right against #943. One nit on the recipe.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Right — the projection came over from #943's body with the same gap. Projected it in `5187194` rather than dropping `reviews` from `--json`: a sibling review is the cheaper tell that something happened while the session waited, and the count is what the reader compares against session-start.

PR #948, opened 2026-08-11
test(notifications-check): cover the pre-check's sweep, cleanup and deferral

max-sixty:
`notifications-check.sh` became its own file so shellcheck could see it, which left its layers with no behaviour tests. This adds ten, one per belief: the shadowed-run sweep (matched on tend workflow names, so a run of the repo's own CI clears nothing), the bot's-own-closed-PR cleanup, the same-repo-only deferral window, and the four fetch paths — retry, give-up, HTML-carrying-a-200, and the Layer D recount falling back to the pre-mutation snapshot.

The fake `gh` drops threads it marked read from later listings, mirroring what the real unread listing does. Without that the recount returns the same snapshot every time and there is nothing to assert.

<details>
<summary>Mutation check — five breakages, each caught by the test that owns it</summary>

| Mutation | Test that failed |
|---|---|
| Layer C patches every PR notification | `clears_the_bots_own_closed_prs` (+ the Layer B count) |
| Layer B's name regex matches anything | `clears_what_a_recent_tend_run_covered` |
| Layer D drops the freshness cutoff | `defers_only_fresh_same_repo_work` |
| No JSON parse guard on the fetch | `tolerates_an_html_200` |
| Recount failure zeroes the count | `counts_from_the_snapshot_when_the_recount_fails` |

</details>

`_output(env, key)` replaces `_should_run` and `_count`, which were one reader keyed differently.

The shared fake `date` grew the two offsets this script asks for. Its relative-offset branches have to precede the format branch — every call also carries `+%Y-%m-%dT%H:%M:%SZ`, so matching that first would collapse both cutoffs onto one timestamp.

Also included: the example config and `SetupStep`'s docstring both still said only `tend-notifications` injects the setup `if:` guard. `tend-review` has done so since it grew its examined-HEAD gate.

### DIFF
--- a/generator/tests/test_shared_steps.py
+++ b/generator/tests/test_shared_steps.py
@@ -292,9 +292,13 @@ esac
 # The script is written for the Ubuntu runners' GNU date; macOS ships BSD
 # date, which has no `-d`. Fixed values also make the day-scoping assertions
 # deterministic: "today" is 2026-01-02.
+# The relative offsets come first: every call also carries a format string, so
+# matching that branch first would collapse them all onto one timestamp.
 FAKE_DATE = r"""#!/usr/bin/env bash
 case "$*" in
+  *"30 minutes ago"*) echo "2026-01-02T11:30:00Z" ;;
   *"20 minutes ago"*) echo "2026-01-02T11:40:00Z" ;;
+  *"10 minutes ago"*) echo "2026-01-02T11:50:00Z" ;;
   *"yesterday"*) echo "2026-01-01" ;;
   *"6 days ago"*) echo "2025-12-27" ;;
   *"%Y-%m-%dT%H:%M:%SZ"*) echo "2026-01-02T12:00:00Z" ;;
@@ -302,8 +306,8 @@ case "$*" in
 esac
 """
 
-# The preflight jitters before its check-then-act; a real sleep would add up
-# to 30s per test.
+# The preflight jitters before its check-then-act, and the notifications
+# pre-check backs off between fetch attempts; real sleeps would add up.
 FAKE_SLEEP = "#!/usr/bin/env bash\nexit 0\n"
 
 TODAY = "2026-01-02"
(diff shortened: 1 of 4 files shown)


generator/tests/test_shared_steps.py
```diff
@@ -1025,6 +1029,342 @@ def test_review_gate_fails_open_on_an_html_200(gate_env: dict[str, str]) -> None
...
+        notifications_env,
+        "NOTIFICATIONS_JSON",
+        [_notif("999", "issues", 7, NOTIF_SETTLED)],
+    )
+    notifications_env["FAIL_NOTIFS_FROM"] = "2"
+
+    result = _run_check(notifications_env)
+
+    assert result.returncode == 0, result.stderr
+    assert _count(notifications_env) == "1"
```
tend-agent: With nothing for Layers B/C to clear, the snapshot and the recount agree, so `1` is what this asserts either way — it catches a zeroed count but not a recount that quietly succeeded, and the over-count the docstring describes is never exercised. Giving Layer C one thread to clear separates them: the fallback counts 2, a live recount 1.

```suggestion
    _write_json(
        notifications_env,
        "NOTIFICATIONS_JSON",
        [
            _notif("11", "pulls", 1, NOTIF_SETTLED),
            _notif("999", "issues", 7, NOTIF_SETTLED),
        ],
    )
    # Layer C clears this one, so a recount that ran would have returned 1.
    _write_json(
        notifications_env,
        "PULLS_JSON",
        [{"number": 1, "user": {"login": "tend-agent"}, "state": "closed"}],
    )
    notifications_env["FAIL_NOTIFS_FROM"] = "2"

    result = _run_check(notifications_env)

    assert result.returncode == 0, result.stderr
    assert _marked_read(notifications_env) == ["11"]
    assert _count(notifications_env) == "2"
```

tend-agent (commented):
Verified the ten new tests pass and that each mutation in your table is actually caught — including the two I doubted, the parse guard (removing it makes `jq 'length'` abort the step under `bash -e`, so `returncode == 0` fails) and the recount fallback (removing it leaves `count=` empty). The `tend-review` doc/docstring correction matches `generate_review`, which renders `steps.gate.outputs.should_run == 'true'` into every setup step.

One gap, inline: the snapshot-fallback test asserts the count the *successful* recount would also produce, so it pins "doesn't zero out" but not "counts from the pre-mutation snapshot" — which is what its name and docstring claim, and the only part of that branch that can regress quietly.

Minor, outside the diff: `_count` is `_should_run` with a different key prefix. A shared `_output(env, key)` would serve both — happy to push that if you want it.

generator/tests/test_shared_steps.py
max-sixty: Applied, with the fixture you suggested — Layer C now has thread `11` to clear, so the fallback counts 2 where a live recount would count 1. Checked it both ways: replacing `REMAINING="$NOTIFS"` with `[]` fails it, and so does letting the recount succeed, which the old fixture couldn't tell apart.

Took the `_output(env, key)` extraction too; `_should_run` and `_count` are gone.

tend-agent (approved):

PR #951, opened 2026-08-12
fix(running-in-ci): make the rerun poll cap terminal, and print conclusions

tend-agent:
Surfaced by the nightly rolling survey of `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md`.

The CI Monitoring loop two sections up is terminal in both directions — `exit 0` once the checks are terminal, `echo` + `exit 1` when the 9-iteration budget runs out. The `gh run rerun --failed` loop below it is not: it `break`s out on success and falls off the end on exhaustion, so those two outcomes produce identical output. An agent reading that transcript has nothing to distinguish "every reran job finished" from "the budget ran out with jobs still pending", and the natural reading of the silence is the first one.

The same loop also never printed a conclusion. `pending_jobs` only asks whether each job's `.status` reached `completed`, which is true of a job that failed again — the exact outcome a rerun exists to detect. So even the success path handed the next step no signal.

This makes the loop mirror the one above: print each job's `conclusion` and `exit 0` when the last one lands, `exit 1` with a message on the cap. The trailing paragraph now says why, and carries over the "the cap is the whole poll budget, don't re-enter" rule that item 4 of the CI Monitoring list already states.

No test — the change is skill prose plus the shell recipe embedded in it, and the repo has no harness that executes recipes out of skill markdown.

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -366,12 +366,22 @@ pending_jobs() {
   echo "$n"
 }
 for i in $(seq 1 9); do
-  [ "$(pending_jobs)" -eq 0 ] && break
+  if [ "$(pending_jobs)" -eq 0 ]; then
+    # `completed` is not `success`. Print each conclusion — a rerun that failed
+    # again is the case the follow-up turns on, and the loop above only ever
+    # asked whether the jobs stopped.
+    for id in $JOB_IDS; do
+      gh api "repos/$REPO/actions/jobs/$id" --jq '"\(.conclusion)\t\(.name)"'
+    done
+    exit 0
+  fi
   sleep 60
 done
+echo "Rerun jobs still running after 9 minutes"
+exit 1
 ```
 
-As with the CI Monitoring loop above, invoke this Bash call with `timeout: 600000` (10 min) — the default 2-min Bash timeout would kill the loop early, and the 9-iteration cap is sized to fit inside the harness's 10-min Bash maximum.
+As with the CI Monitoring loop above, invoke this Bash call with `timeout: 600000` (10 min) — the default 2-min Bash timeout would kill the loop early, and the 9-iteration cap is sized to fit inside the harness's 10-min Bash maximum. Exit terminally in both directions for the same reason that loop does: a bare `break` leaves "every job finished" and "the budget ran out with jobs still pending" looking identical in the transcript, and the next step reads the silence as a green rerun. On the cap, treat it as the whole poll budget — don't re-enter — and report the still-running jobs as unverified.
 
 ## Replying to Comments
 
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -365,6 +365,7 @@ pending_jobs() {
   done
   echo "$n"
 }
+[ -n "$JOB_IDS" ] || { echo "No rerun attempt surfaced — jobs were not re-queued"; exit 1; }
 for i in $(seq 1 9); do
   if [ "$(pending_jobs)" -eq 0 ]; then
     # `completed` is not `success`. Print each conclusion — a rerun that failed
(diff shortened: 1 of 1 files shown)


tend-agent (commented):
The recipe now distinguishes "jobs finished" from "budget exhausted", but one path still lands on the ambiguous side: when `JOB_IDS` comes back empty, `pending_jobs` echoes `0` on the first iteration, the new conclusion loop iterates over nothing, and the recipe `exit 0`s with no output — which reads exactly like a clean rerun. That's not hypothetical: the comment above `sleep 10` names the case ("the next query can see only the prior `failure` rows and exit immediately"), and a 10s sleep mitigates the attempt-registration lag rather than guaranteeing it's gone. Inline suggestion below adds the guard.

Second suggestion is about skill length rather than correctness — the recipe already shows `exit 0` and `exit 1`, so the sentence explaining why it exits in both directions is motivation the reader doesn't need to apply the rule (per CLAUDE.md's "cut motivation" for skill authoring). The operative half — cap is the whole budget, report unverified — is worth keeping.

Rest of the change reads correctly to me: `--jq '"\(.conclusion)\t\(.name)"'` is valid jq interpolation, and exiting `0` on a re-failed job matches the CI Monitoring loop above, which also exits `0` on terminal-but-red and leaves the verdict to the follow-up step.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -366,12 +366,22 @@ pending_jobs() {
   echo "$n"
 }
 for i in $(seq 1 9); do
```
tend-agent, line 369: An empty `JOB_IDS` short-circuits straight past the new reporting: `pending_jobs` returns `0`, the conclusion loop prints nothing, and the recipe exits `0` — the same silent-success shape this PR is removing. Guard it so the un-surfaced-attempt case is terminal in the other direction too.

```suggestion
[ -n "$JOB_IDS" ] || { echo "No rerun attempt surfaced — jobs were not re-queued"; exit 1; }
for i in $(seq 1 9); do
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -366,12 +366,22 @@ pending_jobs() {
...
+    exit 0
+  fi
   sleep 60
 done
+echo "Rerun jobs still running after 9 minutes"
+exit 1
 ```
 
-As with the CI Monitoring loop above, invoke this Bash call with `timeout: 600000` (10 min) — the default 2-min Bash timeout would kill the loop early, and the 9-iteration cap is sized to fit inside the harness's 10-min Bash maximum.
+As with the CI Monitoring loop above, invoke this Bash call with `timeout: 600000` (10 min) — the default 2-min Bash timeout would kill the loop early, and the 9-iteration cap is sized to fit inside the harness's 10-min Bash maximum. Exit terminally in both directions for the same reason that loop does: a bare `break` leaves "every job finished" and "the budget ran out with jobs still pending" looking identical in the transcript, and the next step reads the silence as a green rerun. On the cap, treat it as the whole poll budget — don't re-enter — and report the still-running jobs as unverified.
```
tend-agent: The recipe above already shows both exits, so the middle sentence explains code the reader can see. Dropping it keeps the rule that isn't derivable from the snippet — that the cap is the whole budget and the outcome must be reported as unverified.

```suggestion
As with the CI Monitoring loop above, invoke this Bash call with `timeout: 600000` (10 min) — the default 2-min Bash timeout would kill the loop early, and the 9-iteration cap is sized to fit inside the harness's 10-min Bash maximum. On the cap, treat it as the whole poll budget — don't re-enter — and report the still-running jobs as unverified.
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied as suggested. The empty-`JOB_IDS` path was the one branch that still exited `0` with nothing printed, which is the shape this PR set out to remove — the guard now fails loudly there instead.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied. Fair — the snippet shows both exits, so the sentence restating why was motivation the reader doesn't need to act on the rule; the cap-is-the-whole-budget half stays.

PR #956, opened 2026-08-12
test: run the shell-script suite off the CI image (bash, jq), and unstale two snapshots

max-sixty:
`wt test` failed six tests on macOS: the shell-script tests replace the environment wholesale, so a bare `bash` resolved through the PATH they set for their fake binaries and reached `/bin/bash` 3.2. `compute-token-usage.sh` uses `mapfile`, a bash 4 builtin, so those tests exited 127 locally while passing on the ubuntu runners.

`BASH = shutil.which("bash")` joins `ACTION_VERSION` in `generator/tests/__init__.py` and every shell invocation in the suite uses it, `test_integration.py`'s generated-snippet run included — so the tests exercise the scripts under the bash version CI does. `_usage` was the one helper still hardcoding its PATH; it now resolves `jq` the way the three fixtures below it do, so the token-usage tests don't depend on `jq` sitting in `/usr/bin`.

Merging main surfaced two more suites of the same kind, both fixed here:

- **`jq` 1.8 removed `_nwise`**, which the outage-reconcile fake used to page comments, so it returned nothing to reconcile and `test_report_failure_reconciles_a_racing_leg` / `…_past_the_first_page` failed on any box with jq ≥ 1.8 (ubuntu ships 1.7.1). It now slices by index; an empty list still pages as `[[]]`, the shape real `gh` returns.
- **`test_workflow_with_local_setup_regtest[review]` and `[mention]` were red on main**, not just locally: [#847](https://github.com/max-sixty/tend/pull/847) recorded their snapshots against a base that predated the [#904](https://github.com/max-sixty/tend/pull/904) revert and [#849](https://github.com/max-sixty/tend/pull/849), so they carried the prose prompt and the older mention comments. Re-recorded from the current render.

429 tests pass on macOS.


tend-agent (commented):
The bash swap itself looks right — `shutil.which` resolves against the pytest process's PATH before any test replaces it, and nothing in the suite writes a fake `bash` that this would shadow.

One gap left against the macOS goal, though: `_usage` is the only helper in `test_shared_steps.py` that hardcodes its PATH instead of resolving the tool it needs, and `compute-token-usage.sh` shells out to `jq`. macOS ships no `jq`, and on Apple Silicon Homebrew installs to `/opt/homebrew/bin`, which `/usr/bin:/bin:/usr/local/bin` doesn't cover — `/usr/local/bin` is the Intel prefix, which is probably why the suite goes green on your machine. So on an arm64 Mac the token-usage tests still fail, one step later than `mapfile` did rather than not at all.

Suggestions inline. Happy to push them as a commit instead if you'd rather not hand-apply.

generator/tests/test_shared_steps.py
```diff
@@ -308,7 +309,7 @@ def _cancelled_stream(tmp_path: Path) -> Path:
 
 def _usage(tmp_path: Path, *, stream: Path | None, logs_dir: Path) -> dict[str, object]:
     result = subprocess.run(
-        ["bash", str(COMPUTE_TOKEN_USAGE)],
+        [BASH, str(COMPUTE_TOKEN_USAGE)],
         env={
             "PATH": "/usr/bin:/bin:/usr/local/bin",
```
tend-agent: `compute-token-usage.sh` runs `jq` on all three paths, and the last one — the all-zeros fallback — is the only one without a `|| echo ''`, so under `set -euo pipefail` a missing `jq` exits 127 rather than degrading. Running the script with `jq` off PATH gives `line 120: jq: command not found`, exit 127, which trips `assert result.returncode == 0` here.

The fixture two helpers down already has the idiom (`jq = shutil.which("jq")` / `assert jq, "jq is required for these tests"` / `PATH: f"{Path(jq).parent}:/usr/bin:/bin"`); `shutil` and `Path` are already imported at the top of this file.

```suggestion
def _usage(tmp_path: Path, *, stream: Path | None, logs_dir: Path) -> dict[str, object]:
    # The script shells out to jq, which macOS doesn't ship and Homebrew puts
    # outside the fixed list below on arm64.
    jq = shutil.which("jq")
    assert jq, "jq is required for these tests"
    result = subprocess.run(
        [BASH, str(COMPUTE_TOKEN_USAGE)],
        env={
            "PATH": f"{Path(jq).parent}:/usr/bin:/bin",
```

generator/tests/__init__.py
```diff
@@ -2,8 +2,14 @@
 
 from __future__ import annotations
 
+import shutil
 from importlib.metadata import version
 
+# Interpreter for the repo's shell scripts. A bare `bash` would resolve through
+# the PATH each test sets for its fake binaries, reaching macOS's /bin/bash 3.2
+# — which lacks builtins the runner's bash 5 has (`mapfile`).
+BASH = shutil.which("bash")
```
tend-agent, line 12: `which` returns `None` when the lookup misses, and there's no type checker on the suite to catch it — so a machine without `bash` on PATH gets `TypeError: expected str, bytes or os.PathLike object, not NoneType` at each of the seven call sites instead of one clear message. Same idiom the `jq` lookup in `test_shared_steps.py` already uses.

```suggestion
BASH = shutil.which("bash")
assert BASH, "bash is required for these tests"
```

generator/tests/test_shared_steps.py
max-sixty: Applied, with one correction to the diagnosis: macOS has shipped `jq` at `/usr/bin/jq` since Sequoia, and it's there on this machine (`root:wheel`, SIP-protected), which is why the suite went green — not `/usr/local/bin`. So the arm64 failure you describe doesn't reproduce here.

The change is still right, and for the reason your second paragraph gives: `_usage` was the only helper asserting where a tool lives instead of asking, and on an older macOS or any box with `jq` outside `/usr/bin` it breaks. Now `f"{Path(jq).parent}:/usr/bin:/bin"`, matching the three fixtures below.

generator/tests/__init__.py
max-sixty: Applied — `assert BASH, "bash is required for these tests"`, same shape as the `jq` lookups.

tend-agent:
Both suggestions are in on 7dda2c21, and your `jq` correction lands — I inferred the Homebrew prefix instead of checking the base system, and `/usr/bin/jq` since Sequoia explains the green run better than `/usr/local/bin` did.

The red `test` check on this head isn't from this PR. `origin/main` at f733969 fails the same two tests with nothing from this branch applied:

```
FAILED tests/test_generate.py::test_workflow_with_local_setup_regtest[review]
FAILED tests/test_generate.py::test_workflow_with_local_setup_regtest[mention]
```

Both snapshots are new files added by #847, recorded against a base that has since moved twice. The `[mention]` one predates #916 and #955, so it's missing their gate comments and the inline-comment fetch they rewrote. The `[review]` one carries `Run the /tend-ci-runner:review skill for {0}. Beyond the skills the Skill tool lists, /code-review is available too.` — the prose prompt #902 added and #904 reverted; that string exists in main's history only in those two `.out` files and in the two commits of the add-and-revert pair. Because both files were *added* rather than modified, the merge had nothing to conflict on, so the staleness only surfaced after landing.

`uv run pytest "tests/test_generate.py::test_workflow_with_local_setup_regtest" --regtest-reset` regenerates both and the suite goes green; the diff is exactly those two deltas and nothing else. main's own `ci` run for f733969 is still queued, and `tend-ci-fix` fires on a `ci` `workflow_run` concluding `failure`, so it hasn't had a chance to see this yet — leaving it to that rather than opening a competing PR. Nothing to change on this branch; it should go green once main is fixed.

Reproduced on a detached worktree at `origin/main` (f733969), no local edits: `uv run pytest tests/test_generate.py -q` → `2 failed, 127 passed`, same two ids.

Prompt-string provenance, run against main's history:

```
$ git log origin/main --oneline -S"Beyond the skills the Skill tool lists"
06414fd fix(generator): restore local setup composites before POST cleanup (#847)
494a010 revert(generator): restore the slash-command prompt, and record why (#904)
aec5b38 feat(generator): make the Claude prompt prose, so /code-review is reachable (#902)
$ git log origin/main --oneline -S"Beyond the skills the Skill tool lists" -- generator/src/tend/templates/
(no output)
```

`--regtest-reset` on just those two ids: `2 passed`, touching `[review].out` (1 line, the prompt) and `[mention].out` (78 lines, the #916/#955 comment and fetch changes).

</details>

tend-agent:
`main` is red on exactly the two snapshots this PR re-records — [run 31619590386](https://github.com/max-sixty/tend/actions/runs/31619590386) on 06414fd, `test_workflow_with_local_setup_regtest[review]` and `[mention]`, 2 failed / 427 passed. Not opening a ci-fix PR; this one already carries the fix.

The snapshots landed with #847, whose branch recorded them while #902's prose prompt was on `main`; #904 reverted that prompt and #955 added the bot-review gate to `mention.yaml.j2` after the recording, so both files drifted on two counts before #847 merged.

Verified f02bd6f against current `main` (f733969, which is this branch's merge base): `uv run pytest tests/test_generate.py -k local_setup` → 5 passed. The failure still reproduces on `main` itself.

tend-agent (commented):
The `_nwise` replacement is faithful — I checked it against `_nwise(100)` on 0/5/100/200/250-element inputs and the page shapes match, including the empty case that both render as `[[]]`. Snapshots re-record clean here and the suite is green.

One gap left against the PR's own title, though: three fixtures still pin `PATH` to `f"{bindir}:/usr/bin:/bin"` while the script under test shells out to real `jq`, so the "off the CI image (jq)" half lands for `_usage` only. Correcting my earlier framing — I said `_usage` was the only helper asserting where a tool lives rather than asking. It was the only one hardcoding the *whole* PATH; three others assert `jq`'s location just as firmly, they just prepend a `bindir` first.

- `gh_env` → `mark-notification-read.sh`, which runs `jq -r '.issue.number' "$GITHUB_EVENT_PATH"`
- `gate_env` → `review-gate.sh`, which runs `STATE=$(echo "$PR_INFO" | jq -re '.state')` on the happy path
- `test_notifications_precheck_tolerates_transient_non_json` in `test_integration.py` → `notifications-check.sh`, which runs `echo "$NOTIFS" | jq -e .`

The third is the clearest tell: `notifications_env` in `test_shared_steps.py` drives that same script and already resolves `jq`, under the comment `# Both the fake gh and the script itself shell out to jq.` So one suite covers `notifications-check.sh` both ways.

Verified rather than inferred this time — running `mark-notification-read.sh` under a PATH with the fake `gh` and coreutils but no `jq` gives `mark-notification-read.sh: line 29: jq: command not found`, exit 127. Same shape `mapfile` produced, one tool over.

The fix is the idiom already at the three fixtures below them:

```python
    jq = shutil.which("jq")
    assert jq, "jq is required for these tests"
    ...
        "PATH": f"{bindir}:{Path(jq).parent}:/usr/bin:/bin",
```

`test_integration.py` needs `import shutil` added; `test_shared_steps.py` already has both imports. None of the three lines falls inside this diff's hunks, so there's nothing to hand you as an inline suggestion — happy to push it as a commit here if you'd rather it land with the rest of the portability work than as a follow-up.

max-sixty:
Taken, and generalized rather than copied: `tests.tool_path(*fakes)` now builds the PATH — fakes, then jq's directory, then coreutils — and all six call sites use it, so the next fixture can't forget jq. It resolves per call rather than at import, so a suite that needs no jq still runs without one.

`_usage` calls it with no argument (no fakes, just the real tools); every fixture passes its `bindir`. 429 tests pass on macOS, and `dev/test.sh` is green across all four suites.

tend-agent:
`tool_path` is the better shape — the three fixtures I listed and the integration pre-check now resolve `jq` the same way the other three did, and the helper's own docstring carries the rationale, so there's one place to change rather than six. Confirmed on this runner: `uv run pytest -q` in `generator/` gives 429 passed, and `tests.tool_path` is the only PATH construction left in the suite (`grep '"PATH"' generator/tests/` returns one other hit, the `report_failure_env["PATH"].split(":")[0]` bindir lookup, which still resolves correctly since fakes stay first). Nothing outstanding from my side.

One leftover: the `# Both the fake gh and the script itself shell out to jq.` comment in `rate_limit_env` ([test_shared_steps.py:691](https://github.com/max-sixty/tend/blob/9791ba95816ee96ca9a9d6247e5068b49dfbd02a/generator/tests/test_shared_steps.py#L691)) and `notifications_env` ([:1455](https://github.com/max-sixty/tend/blob/9791ba95816ee96ca9a9d6247e5068b49dfbd02a/generator/tests/test_shared_steps.py#L1455)) annotated the lookup that moved into the helper, so both now float above an unrelated block of file setup. Cosmetic, and the docstring says the same thing in one place.

generator/tests/test_shared_steps.py
```diff
@@ -689,8 +689,6 @@ def rate_limit_env(tmp_path: Path) -> dict[str, str]:
     )
 
     # Both the fake gh and the script itself shell out to jq.
-    jq = shutil.which("jq")
-    assert jq, "jq is required for these tests"
```
tend-agent: The block this described is gone — `tool_path` resolves jq now.

```suggestion
```

generator/tests/test_shared_steps.py
```diff
@@ -1455,8 +1453,6 @@ def notifications_env(tmp_path: Path) -> dict[str, str]:
...
```
tend-agent: Same here.

```suggestion
```

tend-agent (approved):
`tool_path` closes the gap generically — all six call sites route through it, and `report_failure_env["PATH"].split(":")[0]` in `test_run_issue_reconcile_refuses_a_call_with_no_row` still lands on `bindir` because fakes stay first. Suites are green here (`test_shared_steps.py` + `test_integration.py`, 90 passed).

Two comments the refactor left behind, both inline: `# Both the fake gh and the script itself shell out to jq.` in `rate_limit_env` and `notifications_env` annotated the `shutil.which("jq")` block that moved into `tool_path`'s docstring. They now sit above the `event`/`notifications` file setup, which has nothing to do with jq.

max-sixty:
Both comments dropped — you're right that they were annotating the lookup, not the file setup they ended up above.

PR #958, opened 2026-08-12
chore: regenerate workflows with tend 0.1.15

max-sixty:
The release's last step: all eight generated workflows move from `max-sixty/tend/claude@0.1.14` to `@0.1.15`, which now exists as a tag.

The behaviour tend itself gains, having shipped it to adopters first:

- **`tend-review` gets the examined-HEAD gate** ([#903](https://github.com/max-sixty/tend/pull/903)). Concurrency drops to `cancel-in-progress: false` and the pre-check skips a queued run whose live HEAD already carries the `tend-review/<pr>` stamp, so a push mid-review queues behind the session instead of throwing away what it had read.
- **`tend-mention` stops booting sessions that can only no-op** ([#849](https://github.com/max-sixty/tend/pull/849), [#916](https://github.com/max-sixty/tend/pull/916), [#955](https://github.com/max-sixty/tend/pull/955)), and its verify gate fetches a review's inline comments when deciding engagement.
- **`tend-notifications`** carries the pre-check's setup guard, so a run that finds an empty inbox does no setup either.

Generated by `uvx tend@latest init`; the hand-maintained `review-reviewers.yaml` is restamped onto `@0.1.15` in the same branch, so `.github/workflows/` holds one action ref. `pre-commit run --all-files` (actionlint included) is clean.


tend-agent (commented):
Regeneration reproduces exactly — `uvx tend@latest init` on this tree leaves a clean worktree, the `0.1.15` tag resolves and carries `claude/action.yaml`, and every step downstream of the two new pre-check gates (`tend-review`'s `gate`, `tend-notifications`'s `check`) carries the matching `if:`, so a skip short-circuits the whole job rather than leaving a dangling checkout.

One gap, outside the diff: `review-reviewers.yaml` is still on `max-sixty/tend/claude@0.1.14` ([line 85](https://github.com/max-sixty/tend/blob/5a4bfe0e1da7701b5a27ce7041e38b5ebfa1d83d/.github/workflows/review-reviewers.yaml#L85)), so `.github/workflows/` now holds two different action refs. `init` never rewrites that file, and `running-tend` ("Nightly: restamp the hand-maintained workflow refs") asks for the restamp to ride along in the regen PR — same worktree, same commit.

It's not purely cosmetic this time: `tend-rate-limit` is introduced at `0.1.15` (`PAUSE_LABEL="tend-rate-limit"` in `shared/steps/rate-limit-preflight.sh`; the token doesn't appear anywhere under `shared/`, `claude/`, or `codex/` at `0.1.14`). This PR teaches `tend-mention` and `tend-triage` to skip issues carrying that label, so `review-reviewers` — pinned a version back — is the one workflow that can neither file the pause issue nor honor the pause the others now respect.

Happy to push the one-line restamp to this branch if you want it folded in.

max-sixty:
Folded in as `b7a4654d`: `review-reviewers.yaml:85` now reads `@0.1.15`, and `.github/workflows/` holds exactly one action ref again.

Confirmed your non-cosmetic point before taking it — `tend-rate-limit` appears nowhere under `shared/`, `claude/`, or `codex/` at the `0.1.14` tag, and at `0.1.15` it's in `rate-limit-preflight.sh` and `run-issue.sh`. So the leg that audits the others was the one that could neither file the pause issue nor honour it.

tend-agent (approved):

PR #959, opened 2026-08-12
docs: record the claude_version marker decision in CLAUDE.md

### LINKED ISSUE #927: claude_version tracks `latest`, which runs ~6 releases ahead of the installer's `stable`
`running-tend`'s weekly rule for `claude_version` says **track latest**, and names `npm view @anthropic-ai/claude-code dist-tags.latest` as the source. That's been followed consistently. The question is whether it's the channel we want, because the binary it selects runs in every consumer's CI job.

The release bucket publishes two markers, and they are not close together:

```
$ curl -fsS https://downloads.claude.ai/claude-code-releases/stable
2.1.220
$ curl -fsS https://downloads.claude.ai/claude-code-releases/latest
2.1.226
```

Six patch releases apart, as of 2026-08-09. `2.1.220` is the pin #918 is currently replacing — that is, `stable` has only just caught up to the version the *previous* weekly bump landed. The promotion lag looks to be roughly one weekly bump cycle, so tracking `latest` means adopters are more or less permanently one cycle ahead of the release Anthropic has promoted.

`claude install` takes either marker or an exact version, so all three are available to us:

```
$ claude install --help
Usage: claude install [options] [target]

Install Claude Code native build. Use [target] to specify version (stable,
latest, or specific version)
```

The pin should stay an exact version either way — reproducibility, and the action's input description already says so. The decision is only which marker we read when picking that number each week.

It genuinely cuts both ways, which is why this is a maintainer call rather than a PR:

- **Tracking `latest` gets fixes sooner, and some have been squarely ours.** In the 2.1.221–2.1.226 range alone: 2.1.225 fixed a transient 401 replacing a long-lived `CLAUDE_CODE_OAUTH_TOKEN` with a stored login's short-lived token and breaking headless sessions until restart — tend's exact auth shape, in a run that never restarts. 2.1.222 fixed the startup connectivity check hanging behind an HTTPS proxy, and every tend session starts behind `mitmdump`. Waiting a cycle for `stable` means adopters sit on those bugs for another week.
- **Tracking `stable` means adopters never run a release Anthropic hasn't promoted.** The blast radius is every consumer's job, and a bad release shows up as agent misbehaviour rather than a clean failure — the `/code-review` waiver scan (#926) is one example of a dependency that breaks silently.

No action needed on #918 either way; it followed the rule as written. Filing this so the choice gets made once rather than re-litigated at each weekly bump. Whichever way it goes, the rule in `.claude/skills/running-tend/SKILL.md` should name the marker and say why — the current "track latest" reads as a default rather than a decision.

Related: #922, on the `codex_version` rule having a confirmation step that isn't reachable from CI.

tend-agent:
#927 asked whether `claude_version` should track npm's `latest` dist-tag, given it runs several releases ahead of `stable`. Answered in [#927](https://github.com/max-sixty/tend/issues/927#issuecomment-5273490373): yes, stay on `latest`.

So this records the decision where repo rationale lives — five lines in `CLAUDE.md` under "Key details" — and leaves the weekly-bump skill with a one-line rule (`track npm's `latest` dist-tag`). Nothing about the pin itself changes: it stays an exact version, and the weekly recipe is untouched.

The one non-obvious fact, and the reason `stable` isn't the safer-looking choice it appears to be: `install.sh` validates the target against `stable|latest|X.Y.Z` and then unconditionally downloads the `latest` build as its bootstrap installer before handing off (`"$binary_path" install ${TARGET:+"$TARGET"}`). Only the agent session runs the pinned version, so a `stable` pin would not keep un-promoted code out of the job.

Closes #927.

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -186,8 +186,8 @@ shape — headless sessions losing a long-lived `CLAUDE_CODE_OAUTH_TOKEN` to a
 stored short-lived one, the startup connectivity check hanging behind an HTTPS
 proxy. `stable` would not even keep un-promoted code out of the job:
 `install.sh` always fetches the `latest` build as its bootstrap installer before
-handing off to the pinned target. Pin an exact version either way — the marker
-only picks the number.
+handing off to the pinned target — only the agent session runs the pinned
+version. Pin an exact version either way — the marker only picks the number.
 
 `mitmproxy_version` pins the process that holds the real PAT and model
 credential, so a security fix there matters here. Check anything security- or
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -157,7 +157,7 @@ both.
 
 | Pin | File | Rule |
 |---|---|---|
-| `claude_version` | `claude/action.yaml` | track latest |
+| `claude_version` | `claude/action.yaml` | track npm's `latest` dist-tag, not `stable` — see below |
 | `mitmproxy_version` | `claude/action.yaml` | track latest |
 | `uv_version` | `claude/action.yaml` | move it with `mitmproxy_version` |
 | `codex_version` | `codex/action.yaml` | track latest; the surface job confirms the bump |
@@ -178,6 +178,16 @@ target, so drift silently downgrades the model. Skim the claude-code CHANGELOG
 between the two versions for anything touching the agent paths (first-run
 onboarding, `--model` alias resolution, headless `-p` result events, Stop-hook
 behavior, slash-command or Skill-tool handling) and note it in the PR.
+
+`latest` over `stable` is a decision, not a default. Both markers exist on npm
+and in the release bucket, several releases apart (2026-08-12: `stable` 2.1.222,
+`latest` 2.1.229), and the fixes in that window have repeatedly been tend's own
+shape — headless sessions losing a long-lived `CLAUDE_CODE_OAUTH_TOKEN` to a
+stored short-lived one, the startup connectivity check hanging behind an HTTPS
+proxy. `stable` would not even keep un-promoted code out of the job:
+`install.sh` always fetches the `latest` build as its bootstrap installer before
+handing off to the pinned target. Pin an exact version either way — the marker
+only picks the number.
 
 `mitmproxy_version` pins the process that holds the real PAT and model
 credential, so a security fix there matters here. Check anything security- or
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -157,7 +157,7 @@ both.
 
 | Pin | File | Rule |
 |---|---|---|
-| `claude_version` | `claude/action.yaml` | track npm's `latest` dist-tag, not `stable` — see below |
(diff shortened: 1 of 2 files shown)


.claude/skills/running-tend/SKILL.md
```diff
@@ -179,6 +179,16 @@ between the two versions for anything touching the agent paths (first-run
...
 
+`latest` over `stable` is a decision, not a default. Both markers exist on npm
+and in the release bucket, several releases apart (2026-08-12: `stable` 2.1.222,
+`latest` 2.1.229), and the fixes in that window have repeatedly been tend's own
+shape — headless sessions losing a long-lived `CLAUDE_CODE_OAUTH_TOKEN` to a
+stored short-lived one, the startup connectivity check hanging behind an HTTPS
+proxy. `stable` would not even keep un-promoted code out of the job:
+`install.sh` always fetches the `latest` build as its bootstrap installer before
+handing off to the pinned target. Pin an exact version either way — the marker
+only picks the number.
```
tend-agent: The install.sh point is right, but as written it reads as "the pin doesn't decide what runs" — which would undercut the paragraph directly above, whose CHANGELOG skim only matters because the pinned version *is* what the agent session runs. One clause keeps both true.

```suggestion
proxy. `stable` would not even keep un-promoted code out of the job:
`install.sh` always fetches the `latest` build as its bootstrap installer before
handing off to the pinned target — only the agent session runs the pinned
version. Pin an exact version either way — the marker only picks the number.
```

tend-agent (commented):
Facts check out against the sources: npm `dist-tags` and the release bucket both read `stable` 2.1.222 / `latest` 2.1.229 today, `install.sh` (217 lines) fetches `$DOWNLOAD_BASE_URL/latest` unconditionally before `"$binary_path" install ${TARGET:+"$TARGET"}`, and both changelog entries the PR body cites are in `anthropics/claude-code` `CHANGELOG.md` at 2.1.225 and 2.1.222.

One clarity point on the install.sh sentence, inline. Not approving because it's my own PR.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied in ca252f6 — the clause is in, so the sentence now says the pin doesn't gate the *installer* download while the paragraph above still rests on the pinned version being what the agent session runs.

.claude/skills/running-tend/SKILL.md
```diff
@@ -179,6 +179,16 @@ between the two versions for anything touching the agent paths (first-run
 onboarding, `--model` alias resolution, headless `-p` result events, Stop-hook
 behavior, slash-command or Skill-tool handling) and note it in the PR.
 
+`latest` over `stable` is a decision, not a default. Both markers exist on npm
```
max-sixty: this is crazy verbose and breaks guidance around specificity in guidance

it also belongs in a CLAUDE.md, not in the `running-tend` skill I think?

tend-agent: Agreed on both counts — dropped in f493d3b.

The skill keeps a one-line rule in the table (`track npm's `latest` dist-tag`) and nothing else; the rationale is now five lines in `CLAUDE.md` under "Key details", with the dated version numbers and the changelog case history cut. Net diff against `main` is now those five lines plus the one table cell. PR title and body updated to match.

PR #960, opened 2026-08-12
skills: gate self-initiated fixes on cost, not just evidence

max-sixty:
The fleet's recent output has tilted toward machinery that prevents wasted compute — retry frameworks, skip-gates, window arithmetic — while the gates price only evidence strength, so a well-evidenced waste fix passes at any complexity. This encodes the asymmetry the repo runs on: complexity is spent on wrong outward actions (what the bot posts, approves, merges, closes), never on saving compute, where the only sanctioned fixes are simple knobs.

<details><summary>What changed, per file</summary>

- `CLAUDE.md` — the vision paragraph, plus a bound on the "gating in YAML is cheap" pattern so it can't be quoted as license for gate stacks.
- `shared/review-gates.md` — Gate 3 (cost class: wrong outward action vs wasted compute); waste supports only knob-grade fixes at any occurrence count, and aggregates escalate to the maintainer as a number; a ban on upgrading the class via hypothetical chains; the stochastic 5+ floor is not lowered by reclassification or pre-registered conditions.
- `running-in-ci` — a "Weighing a Fix" section stating the value order; skill code blocks past ~two dozen lines move to tested scripts; a 5-open-PR budget for self-initiated improvements (never holds requested work or product fixes); no stacking PRs on unmerged bot branches; PR bodies lead with two-three sentences and fold analysis into `<details>`.
- `review` — one criterion: challenge new mechanism whose only payoff is saved agent compute, scoped to the bot's own automation.
- `review-runs` / `review-reviewers` — reconcile the passages that contradicted the new gate (near-timeout "one occurrence is enough to act", the loop-filter proposal path, "put full analysis in PR description").

</details>

<details><summary>Validation</summary>

A context-blind agent applying only the edited files classified five scenarios mirroring the currently-held bot PRs (install-retry stack, survey-window ceiling, occurrence-2 skill bullet, re-anchored stale approval, false-green CI poll) the way the repo's stated values do: the first three blocked, the last two pass reshaped toward tested scripts. An adversarial pass over the full files closed the contradictions it found — the queue budget no longer holds ci-fix/triage product fixes, and the sibling skills no longer carry quotable pro-machinery text.

</details>

> _This was written by Claude Code on behalf of max-sixty_

PR #965, opened 2026-08-13
simplify: coarser mention gates, tested poll scripts, anchor-based run window

### LINKED ISSUE #829: running-in-ci: CI-monitor poll follows the PR head, so a concurrent push makes a run report another commit's green as its own

max-sixty:
Cuts the four complexity hotspots from the recent bot-PR audit back to the #960 vision — mechanism only where it prevents wrong outward actions — and moves the surviving logic under real tests. Everything loaded or executed in production shrinks (~420 lines across the template, generated workflows, and skill prose); the growth is tests for scripts that previously had none.

- **`mention.yaml.j2` verify gates: 4 → 2.** The stacked review-path skip gates collapse into two structural rules: a bot-authored review summons a session only as the reviewer→author handoff (fresh content on a PR the bot authored), and a contentless approval is terminal whoever submitted it. GitHub rejects self-approvals, so the empty-body-APPROVED gate was already subsumed by an author-keyed rule. Outward behavior is unchanged — verified case-by-case twice (by hand, and independently by the review sweep's verifier); each rendered workflow drops ~52 lines.
- **`running-in-ci`: 869 → 677 lines.** The two markdown poll recipes become tested scripts. `poll-pr-checks.sh` polls by commit OID, never the PR head, fixing the false green a concurrent push caused (closes #829) and carrying the edge cases #837's five review rounds surfaced — null merge-ref rollups, superseded check runs, STARTUP_FAILURE/ACTION_REQUIRED bucketing — so it supersedes #837. `rerun-failed-jobs.sh` owns the rerun and finds the new attempt's jobs by `run_attempt`, so a fast rerun no longer reads as "nothing re-queued" and an unregistered one can't report stale conclusions as fresh. The rarely-taken skill-PR mechanics move to `references/skill-pr-workflow.md`.
- **`list-recent-runs.sh`: 207 → 113 lines.** The completion window becomes "floor = the last successful run's start, clamped at 6h, else 6h with a coverage-gap warning". Any successful run's window opens at or before its own start, so consecutive windows overlap and never gap — which makes the cron parsing, tick tiling, dropped-tick recovery, and the schedule-event restriction all unnecessary, at any cadence. The retry wrapper goes too: a transient failure now fails loud, and the next tick's floor reaches back past the lost window. This supersedes the window-anchoring half of #845; its cadence-cut half stands alone and needs no script support under anchor-based windows.
- **`review-reviewers` Non-issues: 7 carve-outs → 3 structural rules** (designed no-ops at whatever layer catches them, designed silence, reviewer-role independence), with the closed menu folded into Gate 3's cost classification.

Also: pre-commit's shellcheck now covers `plugins/**/*.sh` (it covered none of them), and the fake-`gh` test scaffold consolidates into `generator/tests/__init__.py` (five copies → one).

## Testing

The scripts are exercised end-to-end under pytest with fake `gh`/`date`/`sleep` binaries, with the scripts' own jq filters doing the reductions: 21 cases for the poll scripts (false-green guards: >100-context page, null rollup, superseded runs, moved head), 11 for the window logic (anchoring, clamping, self-exclusion, loud failure), and the mention gate consolidation is pinned by the reworked `test_generate.py` assertions plus regenerated snapshots. Full suite green. The verify gate itself is still string-shape-tested rather than executed — extracting it to a `.sh` template like `review-gate.sh` is the natural follow-up, left out here to keep this diff reviewable.

<details><summary>Review-hardening notes (later commits)</summary>

An 8-angle verified code-review of the first commit found the substantive gaps in the new scripts themselves, all fixed and pinned by tests: the GraphQL query's one 100-node page could hide a red check past the cap (now routes to UNVERIFIED); a nullable `startedAt` could let a stale settled check run outrank its QUEUED replacement (a group with any non-terminal entry now reads pending); a transient API blip on a late poll iteration discarded the pending list the cap report names (the last usable rollup is kept); rerun discovery by job status misread fast reruns (now keyed on `run_attempt` advancing); and overlapping workflow prefixes double-counted runs (`unique_by`). The known remaining limit: latest-wins can't distinguish two independent runs of a group-less workflow on one SHA — documented in the script header and the skill, with the instruction to read the specific run's conclusion where that distinction must be exact.

A further tend-review round on the PR added: explicit `--repo`/owner arguments on the three `gh` calls that leaned on cwd remote detection (with test assertions pinning the flags), a Step 1 sentence in `review-reviewers` covering a failing `list-recent-runs.sh` call (retry once, then record a coverage gap — the run still concludes green, so the next tick anchors past the span), and the zero-checks commit named in `poll-pr-checks.sh`'s exit-2 docs.

</details>


plugins/tend-ci-runner/scripts/list-recent-runs.sh
```diff
@@ -1,207 +1,113 @@
...
-# analyzed. For non-schedule events or non-hourly crons, falls back to a
-# now-anchored 1h window.
+# Runs are fetched by *creation* time with a 2h cushion below the floor,
+# then filtered on completion (updatedAt): `gh run list --created` filters
+# by start time, and a run started before the floor may have finished
+# inside the window.
+#
+# A transient API failure fails the script loudly (`set -e`); the caller
+# treats that as "window not analyzed", never as an all-clear, and the next
+# tick's floor reaches back past the lost window.
```
tend-agent, line 26: The caller doesn't hold up its half of this contract, and dropping `gh_retry` is what makes that live.

"The next tick's floor reaches back past the lost window" holds only if *this* run's conclusion isn't `success` — the anchor query filters on `--status success`. But the script is invoked from a Bash tool call inside the agent session (`review-reviewers` Step 1, `TARGET_REPO=$ARGUMENTS ${CLAUDE_PLUGIN_ROOT}/scripts/list-recent-runs.sh`), and a `claude -p` session whose Bash call exits non-zero routinely keeps going and exits 0. The run then concludes `success`, the next tick anchors on its start, and the window is neither analyzed nor recorded as a gap — the exact false all-clear the retry wrapper was added to prevent.

Step 1 only documents the other branch: "If the script printed a `WARNING:` on stderr, the list is known-incomplete…". Nothing tells the agent what a non-zero exit means. With three attempts gone, the first transient 503 takes this path.

One sentence alongside that WARNING paragraph closes it — on a non-zero exit, retry once, and if it still fails record a coverage gap naming the span rather than an all-clear, since the floor advances either way.

plugins/tend-ci-runner/scripts/rerun-failed-jobs.sh
```diff
@@ -0,0 +1,71 @@
...
+#   1  no new attempt surfaced — the rerun did not take
+#   3  jobs still running at the poll cap — UNVERIFIED
+
+set -euo pipefail
+
+RUN_ID="$1"
+RUN_API="repos/$GITHUB_REPOSITORY/actions/runs/$RUN_ID"
+
+BASE_ATTEMPT=$(gh api "$RUN_API" --jq '.run_attempt')
+gh run rerun "$RUN_ID" --failed
```
tend-agent: Every other call in the script addresses the repo through `$GITHUB_REPOSITORY`; this one falls back to `gh`'s cwd git-remote detection (the old recipe passed `--repo "$REPO"`). From a cwd that isn't a checkout — a `/tmp` worktree the session already removed, say — it errors, `set -e` exits 1, and 1 is the documented "no new attempt surfaced — the rerun did not take" code, so the report names a cause that didn't happen.

```suggestion
gh run rerun "$RUN_ID" --failed --repo "$GITHUB_REPOSITORY"
```

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -0,0 +1,178 @@
...
+# a settled green replacement still reads red. Latest-wins assumes the later
+# entry supersedes the earlier — true for the cancellation cascade it
+# targets, not for two independent runs of a group-less workflow; where that
+# distinction must be exact, read the specific run's own conclusion.
+#
+# Exit codes:
+#   0  every gating check on <sha> settled green
+#   1  red — failing checks and their run URLs on stdout
+#   2  no usable rollup for <sha> (unresolvable OID; an ephemeral merge-ref
+#      commit, which carries none; or more contexts than the query's one
```
tend-agent: A commit that simply has no checks lands in exit 2 as well: GraphQL returns `statusCheckRollup: null` whenever a commit carries zero check runs *and* zero status contexts, not only for an unresolvable OID or a merge ref. So a push every workflow's `paths` / `paths-ignore` filter excludes burns the full ~9.5 min cap and then reports UNVERIFIED, where the markdown recipe returned green in 90s and there was nothing to gate on in the first place.

Fail-safe is the right default here, but the case is worth naming in this list — otherwise the next reader hitting exit 2 on a docs-only push reads it as a bug in the query rather than as "this commit has no checks".

tend-agent (commented):
Read the mention-gate consolidation case by case against the four gates it replaces (bot review on own PR with a body / with fresh inline comments / reply-only / empty-body approved; bot review on another author's PR; human bare approval) — outward behaviour matches on every one, and the self-approval argument that collapses the `#747` gate holds. Ran the suite locally: 458 passed, and `shellcheck -S warning` is clean on all of `plugins/**/*.sh` now that the hook covers them.

Three things, one of which I think is a real hole opened by deleting `gh_retry` — it's a caller-contract gap rather than a script bug, and the fix is one sentence in `review-reviewers` Step 1 (outside this diff), so I've left it inline rather than as a suggestion. Happy to push it if you'd like it in this PR.

plugins/tend-ci-runner/scripts/list-recent-runs.sh
max-sixty: Taken in 19770249 as one sentence beside the WARNING paragraph in Step 1: retry once on a non-zero exit, then record the window as a coverage gap — naming the why (this run still concludes green, so the next tick anchors past the span).

plugins/tend-ci-runner/scripts/rerun-failed-jobs.sh
max-sixty: Applied in 19770249, plus the same fix to `poll-pr-checks.sh`'s `gh pr view` in `head_note()` — the only other call in these scripts that fell back to cwd remote detection.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
max-sixty: Named in 19770249 — the exit-2 list now carries the zero-checks commit (paths-filtered push) alongside the merge-ref and overflow cases.

tend-agent (approved):

plugins/tend-ci-runner/scripts/list-recent-runs.sh
tend-agent: All three land as described — checked against [`1977024`](https://github.com/max-sixty/tend/commit/19770249969ba3bac055b01bea73a990242dbd72), and the 31 script tests pass locally.

One residual on the "only other call" framing: the anchor query at [`list-recent-runs.sh:65`](https://github.com/max-sixty/tend/blob/19770249969ba3bac055b01bea73a990242dbd72/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L65) is a third cwd-dependent call. It skips `repo_args` on purpose — the anchor has to come from the repo the workflow runs in, not `TARGET_REPO` — but that leaves it on `gh`'s git-remote detection, which errors from a non-checkout cwd (`failed to determine base repo: fatal: not a git repository`, confirmed by running it from `/tmp`). `$GITHUB_REPOSITORY` names exactly the repo that branch wants, and the branch is already gated on `GITHUB_WORKFLOW`, so `--repo "$GITHUB_REPOSITORY"` is the identical fix. Lower stakes than the rerun one: this exits loud and the new Step 1 sentence now tells the caller what to do with that. Consistency rather than a hole — the test env would need `GITHUB_REPOSITORY` alongside `GITHUB_WORKFLOW` for `set -u`.

Minor, on the PR's own theme: neither `--repo` addition is pinned — `test_rerun_reports_each_new_attempt_jobs_conclusion` asserts `"run rerun 9000 --failed" in calls`, which is a substring, so it passes with or without the flag.

max-sixty: Both taken in df6bdefd: the anchor query names `$GITHUB_REPOSITORY` (with the comment noting it is deliberately not `TARGET_REPO`, and `GITHUB_REPOSITORY` added to the test env), and all three `--repo` additions are now pinned — the rerun and head-re-read assertions include the flag, and a new anchor-call assertion checks it directly.

tend-agent: Confirmed against [`df6bdef`](https://github.com/max-sixty/tend/commit/df6bdefdcbcdc684982fa5962a06bf6e5cb17bc1) — both land, and the pins bite rather than just passing alongside the flag. Nothing further from me; the approval stands.

- `list-recent-runs.sh:68` now reads `gh run list --repo "$GITHUB_REPOSITORY" --workflow "$GITHUB_WORKFLOW"`, so all three cwd-dependent calls (`rerun-failed-jobs.sh:28`, `poll-pr-checks.sh:125`, this one) name the repo. The comment's `TARGET_REPO` caveat matches the code: the anchor query is the one call that deliberately doesn't take `repo_args`.
- Ran the 31 script tests locally: all pass.
- Mutation check on the new anchor assertion — dropped `--repo "$GITHUB_REPOSITORY"` from line 68 and re-ran, and `test_window_floors_at_the_anchors_start` fails with the intended message. It's filtering `GH_CALLS` on `--status success`, which uniquely identifies the anchor query, so it can't drift onto another call. The two `test_poll_scripts.py` assertions moved the flag inside the substring, so they fail the same way.
- Required checks are green at this head (`lint`, `test`, `test-codex-surface`, `test-install-tend-scripts`, `test-proxy`, `test-worker`); the pending `review` context is the sibling `tend-review` run on this push.

</details>

PR #667, opened 2026-06-08
skills(review): peek check rollup before APPROVE so visible FAILUREs aren't rubber-stamped

tend-agent:
## Summary

Add a synchronous `statusCheckRollup` peek to `plugins/tend-ci-runner/skills/review/SKILL.md` step 5 so the bot does not emit an empty-body APPROVE when a check on the head SHA has already reached terminal `FAILURE`. Step 6's "approve, then monitor CI in background, dismiss if a check fails" pattern only recovers when the agent is still alive when the failure notification arrives — sessions routinely end within minutes of the foreground `gh pr review --approve`, leaving any post-approve failure undismissed and the PR carrying a misleading APPROVED state.

## Evidence

Triggering case this hour — `max-sixty/worktrunk` PR #3015 (non-maintainer-authored, by `Xeonacid`):

- `tend-review` run [27115489508](https://github.com/max-sixty/worktrunk/actions/runs/27115489508) posted [APPROVED review 4446577248](https://github.com/max-sixty/worktrunk/pull/3015#pullrequestreview-4446577248) (empty body) at `2026-06-08T04:16:02Z` on head `338b9ec9c9661979b610668b5504280a16761707`.
- On the same head, [`codecov/patch` reached terminal `FAILURE` at `2026-06-08T04:15:23Z`](https://app.codecov.io/gh/max-sixty/worktrunk/pull/3015) — **39 s before** the APPROVE.
- worktrunk's `running-tend` overlay treats `codecov/patch` as a merge gate ("After required CI checks pass, poll `codecov/patch` — it is mandatory despite being marked non-required") and `CLAUDE.md` requires explicit user approval before merging with it failing.
- Session log: agent's pre-flight posting check verified `HEAD_SHA` + `PR_STATE` only — never queried `statusCheckRollup` — then approved. Started a background CI poll, summarized, and the session ended ~30 s later, before the background loop could surface the codecov failure for a dismissal action. The approval stands; no dismissal followed.

## Occurrences (Gate evaluation)

Cumulative across this-month and last-month evidence gists for the shape "session shorter than CI report → empty-body APPROVE → no dismissal" (or its sibling "approve while a check is already red, no synchronous peek"):

| # | PR | Run | Shape | Self-corrected? |
|---|---|---|---|---|
| 1 | worktrunk #2580 (commit `46ce764`) | 25307086342 | session ended before `codecov/patch` FAILURE reported | no |
[...]

PR #966, opened 2026-08-13
review-reviewers: pause the hourly sweep, keep it as a manual spot-check

max-sixty:
Drops the `47 * * * *` cron from `review-reviewers`, leaving `workflow_dispatch`. Its recent output had converged on maintaining its own evidence machinery, at the fleet's highest per-run cost and a seven-day quota ceiling that twice took the user-facing workflows down with it; each adopter's daily `tend-review-runs` already reviews that repo's runs from the inside and routes bundled-skill defects upstream, so this removes the direct loop rather than the only one.

The skill is scoped to match: it described itself as hourly and told the session to analyze "the past hour". It now states the window `list-recent-runs.sh` actually hands it — floor at the previous successful run of the workflow, clamped 6h back — which with no cron makes the clamp the normal case rather than an outage signal, and scopes its claims to that window. (An earlier revision of this PR described a cron-payload branch; #965 replaced that logic while this branch was open, and the text was rebased onto what merges.)

<details><summary>What the sweep caught, and when</summary>

Its unique catches — the ones no inside-out session would have found — cluster in April–June: rubber-stamped APPROVEs over visibly failing checks (#667), backslash-backtick corruption in already-published comment bodies across repos (#359, #405, #424, #562), stale PRs silently dropping mentions (#391), and the stale/re-anchored approval family (#798, #821, #884).

The last two weeks are dominated by fixes to its own machinery — evidence-gist plumbing, survey-window arithmetic, matrix branch naming, run→output attribution — while per-run cost climbed from $5.31 to $10.49. That is the shape the cost gate in #960 was written for.

</details>

Re-enable by dispatching it after a change that could move bot behavior fleet-wide: a release, a harness switch, a model bump — within the hour of that change, per the window note in the overlay.


.claude/skills/running-tend/SKILL.md
```diff
@@ -25,7 +25,12 @@ Tend has Claude-powered workflows beyond the generated `tend-*` set:
 
 | Workflow | File | Schedule | Purpose |
 |----------|------|----------|---------|
-| `review-reviewers` | `review-reviewers.yaml` | `47 * * * *` | Hourly analysis of adopter repo sessions |
+| `review-reviewers` | `review-reviewers.yaml` | manual only (paused) | Outside-in analysis of adopter repo sessions |
+
+`review-reviewers` runs only on `workflow_dispatch` — dispatch it as a
+spot-check after a release, harness switch, or model bump, not on a cadence. The
+per-repo `tend-review-runs` carries the routine loop; the workflow file's header
+explains the pause.
```
tend-agent, line 33:
```suggestion
explains the pause. Dispatch within ~6h of the change you want it to see:
`list-recent-runs.sh` anchors on the last successful run and clamps the floor at
6h, so a later dispatch analyses only the last 6h and records the rest as a
coverage gap.
```

tend-agent (commented):
Two notes, neither blocking.

The re-enable path inherits a window limit the docs don't mention. `list-recent-runs.sh` anchors on the previous *successful* run of the calling workflow and clamps that floor at 6h (`floor_cap=$((now - 21600))`), warning on stderr when the clamp bites. With the cron gone, any dispatch more than 6h after the last one hits the clamp: a spot-check run the day after a release analyses only the last 6h of the target repo, and Step 1 of the skill then tells the session to "Record a coverage gap, not an all-clear" rather than reporting on the release it was dispatched for. Inline suggestion adds a line to the overlay paragraph so whoever dispatches knows to do it promptly.

The bundled skill this workflow invokes still describes itself as hourly: `plugins/tend-ci-runner/skills/review-reviewers/SKILL.md` carries `description: Hourly outcome-based analysis of tend's CI behavior…` in its frontmatter and opens with "Analyze tend's CI behavior on the target repo over the past hour" — the past hour being the one window a dispatched run will never be looking at. (The `hourly/review-$GITHUB_RUN_ID-…` branch prefix further down is cosmetic; the frontmatter description is what shows in the skill list.) It's outside this diff — happy to push a commit rewording those two if you want it folded in here.

.claude/skills/running-tend/SKILL.md
max-sixty: Took the point, not the suggestion — the mechanism is different and the real behavior is worse.

The anchor-and-clamp block you quoted sits inside `if [ -n "$cron_minute" ]`, and `cron_minute` is set only when the event payload carries a `.schedule` matching `^([0-9]+) \* \* \* \*$`. A `workflow_dispatch` run has no `.schedule`, so it never enters that branch: no anchor, no 6h clamp, and — since both `WARNING:` lines live in there — no warning either. It falls through to the now-anchored 1-hour window the script's header documents for non-schedule events. (The anchor query also filters `--event schedule`, so dispatched runs were deliberately excluded from being anchors.)

So a spot-check dispatched the day after a release doesn't analyze 6h and flag a gap; it analyzes the last hour and reports a clean all-clear, with your second note explaining why it would sound authoritative — the skill told it to analyze "the past hour", so the narrow window reads as the intended one rather than as a shortfall.

Both are fixed in `1af440e`: the frontmatter description and the opening line drop "hourly", Step 1 now says to read the window off the script and scope every claim to its bounds, and the overlay states the one-hour reality. Widening the window on the dispatch path is the better fix and is being handled with the rest of that script's window logic, separately from this diff.

.claude/skills/running-tend/SKILL.md
```diff
@@ -25,7 +25,18 @@ Tend has Claude-powered workflows beyond the generated `tend-*` set:
...
+
+A dispatched run analyzes **the last hour** and says so: `list-recent-runs.sh`
+anchors its window to the cron tick only under an hourly cron, and every other
+trigger falls back to a now-anchored hour. So dispatch it within the hour of
+whatever you want it to see, or treat the run as a sample of that hour rather
+than a verdict on the change.
```
tend-agent: The window isn't an hour and the fallback isn't trigger-based — inside Actions `list-recent-runs.sh` always anchors on the previous successful run of the calling workflow and clamps that floor 6h back.

```suggestion
A dispatched run's window opens at the previous successful `review-reviewers`
run and is clamped 6h back (`floor_cap` in `list-recent-runs.sh`). With no cron
the clamp is the normal case: a dispatch long after the last one analyses only
the last 6h and warns on stderr that the rest is a coverage gap. Dispatch within
~6h of the change you want it to see.
```

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -220,6 +220,8 @@ If empty, record the run as all-clear per "Recording below-threshold findings" a
 
 If the script printed a `WARNING:` on stderr, the list is known-incomplete — the window was clamped, no anchor was found, or a workflow hit the fetch limit. Record a coverage gap naming the missing span instead of an all-clear, whether or not the list came back empty; the next run's floor advances past that span regardless, so an unrecorded gap is never revisited.
 
+**Read the window off the script, and report it.** The anchoring that tiles consecutive windows applies only under an hourly cron; every other trigger — `workflow_dispatch` included — takes a now-anchored 1-hour window, and takes it silently, since the coverage-gap warnings above live on the anchored path. This workflow is dispatch-only, so a run checking on a change that landed earlier than an hour ago is looking past it. State the window's bounds in your summary and scope every claim to them: "no problems in the hour to 14:05Z", never "no problems".
```
tend-agent: This contradicts the paragraph directly above it. A `workflow_dispatch` run is on the anchored path — the only branch in the script is `GITHUB_WORKFLOW` set vs unset, and it's set in the sandbox — so the clamp warning above does fire, and the window is up to 6h rather than one. "Read the window off the script" also has nothing to read unless a WARNING printed; the floor isn't echoed on the clean path.

```suggestion
**State the window you analysed.** The floor is the previous successful run of this workflow, clamped 6h back. With no cron, consecutive dispatches are usually further apart than that, so the clamp is the normal case and the `WARNING:` above names the span it dropped. Scope every claim to that window — "no problems in the 6h to 14:05Z", never "no problems" — and treat a dispatch aimed at a change older than the window as looking past it.
```

tend-agent (commented):
The new window paragraphs describe a mechanism `list-recent-runs.sh` doesn't have, and the bundled-skill one contradicts the paragraph directly above it.

The script has no trigger-aware branch — nothing in `plugins/tend-ci-runner/scripts/` reads `GITHUB_EVENT_NAME`, `schedule`, or `cron`. The single branch is on `GITHUB_WORKFLOW`: when it's set, the floor is the previous *successful* run of the calling workflow (`gh run list --workflow "$GITHUB_WORKFLOW" --status success`), clamped at `floor_cap=$((now - 21600))`; the `COMPLETED_AFTER=$((now - 3600))` past-hour path is the `else`, for running the script outside Actions. `GITHUB_WORKFLOW` is set inside the agent sandbox — I checked in this session — so a dispatched `review-reviewers` takes the anchored path exactly like the cron did. Its window is "since the last successful dispatch, floored at 6h", not one hour.

And it isn't silent: `list-recent-runs.sh` prints `WARNING: the last successful '$GITHUB_WORKFLOW' run started …, more than 6h back` whenever the clamp bites, which with the cron gone is the normal case. That's the same WARNING the paragraph above the insertion (`If the script printed a WARNING: on stderr…`) already routes to a coverage-gap record — so the new text tells the session the warnings live on a path it isn't on, when it is.

Two knock-ons if a session follows the new text as written: it would report a one-hour window when the script actually analyzed up to six, and "Read the window off the script" has nothing to read on the clean path — the floor is only ever printed inside a WARNING.

The PR description carries the same claim ("anchors its window to the cron tick only under an hourly cron"), so it's worth correcting there too.

Suggestions inline restate both paragraphs against the script's actual behaviour. I'd also note the resolved thread from the last round: the replacement wording isn't a rephrasing of the 6h point, it's a different mechanism.

.claude/skills/running-tend/SKILL.md
tend-agent: Withdrawing this suggestion — my branching was wrong, and the correction on the thread above is right.

The anchor-and-clamp block is *nested inside* the cron branch, not parallel to it: [`if [ -n "$cron_minute" ]`](https://github.com/max-sixty/tend/blob/1af440e5668d127b5002673b4052286eb0483e3d/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L89) opens at L89 and closes at L163, and the [`GITHUB_WORKFLOW` test](https://github.com/max-sixty/tend/blob/1af440e5668d127b5002673b4052286eb0483e3d/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L132) sits at L132 inside it. `cron_minute` is set only when the event payload's `.schedule` matches an `MM * * * *` cron ([L82-L87](https://github.com/max-sixty/tend/blob/1af440e5668d127b5002673b4052286eb0483e3d/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L82-L87)), and a `workflow_dispatch` payload carries no `.schedule` — so a dispatched run reaches neither the anchor, nor `floor_cap`, nor either `WARNING:`. It falls through to [L166-L169](https://github.com/max-sixty/tend/blob/1af440e5668d127b5002673b4052286eb0483e3d/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L166-L169): `COMPLETED_AFTER=$(date -d '1 hour ago' +%s)`. The script's own comment at [L116-L118](https://github.com/max-sixty/tend/blob/1af440e5668d127b5002673b4052286eb0483e3d/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L116-L118) states the same thing, which is why the anchor query also filters `--event schedule`.

So the paragraph as written in `1af440e` is accurate and this suggestion would replace it with a mechanism that never fires on this workflow. Please don't apply it; same for the companion suggestion on `plugins/tend-ci-runner/skills/review-reviewers/SKILL.md`.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: Withdrawing this too — it rests on the same wrong reading, traced on the [overlay thread](https://github.com/max-sixty/tend/pull/966#discussion_r3774852572): the anchor, the 6h clamp and both `WARNING:` lines are nested inside the `cron_minute` branch, which a `workflow_dispatch` run never enters.

So this paragraph doesn't contradict the one above it — the one above is scoped to the anchored path, which a dispatched run isn't on, and this one covers the path it is on. "Read the window off the script" is the weaker half of the point: on the dispatch path the floor is `now - 1h` with nothing echoed, so it's a fixed value the session can state rather than something to read. Not worth another round on this PR.

.claude/skills/running-tend/SKILL.md
max-sixty: Ran it rather than re-read it, because we'd each read the same lines and reached opposite conclusions.

Same script, same environment, `GITHUB_WORKFLOW=review-reviewers` exported in **both** runs — the only difference is the event payload:

| | `{"inputs":{}}` (dispatch) | `{"schedule":"47 * * * *"}` (cron) |
|---|---|---|
| `floor_cap` assignments | 0 | 1 |
| anchor `gh run list --workflow` calls | 0 | 1 |
| `WARNING:` lines on stderr | 0 | 1 |
| `COMPLETED_AFTER` assignments | 1 | 2 (default floor, then pulled back) |

The dispatch run's `COMPLETED_AFTER` came out at exactly `now - 3600`. Since `GITHUB_WORKFLOW` was set in both, it can't be what selects the branch.

The structure agrees: `if [ -n "$cron_minute" ]` opens at L89 with no indent; `if [ -n "${GITHUB_WORKFLOW:-}" ]` is indented two spaces at L132, inside it, and closes at L163; the `else` at L166 — also unindented, so it pairs with L89 — is `COMPLETED_AFTER=$(date -d '1 hour ago' +%s)`. The clamp is `floor_cap=$((intended - 21600))`, not `now - 21600` as quoted; `intended` is assigned only at L93/L95 inside the cron branch, so the clamp couldn't be reachable from outside it. The header says the same at L19-20: "For non-schedule events or non-hourly crons, falls back to a now-anchored 1h window."

Your second point stands on its own and is now fixed in `825dbee`: on the clean path nothing prints the bounds — only the clamp WARNING does — so "read the window off the script" asked for something unreadable. The instruction now says to derive it (run time, less an hour). Both paragraphs also drop the trigger framing for the payload framing, since that's what the branch actually keys on.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
max-sixty: Ran it rather than re-read it, because we'd each read the same lines and reached opposite conclusions.

Same script, same environment, `GITHUB_WORKFLOW=review-reviewers` exported in **both** runs — the only difference is the event payload:

| | `{"inputs":{}}` (dispatch) | `{"schedule":"47 * * * *"}` (cron) |
|---|---|---|
| `floor_cap` assignments | 0 | 1 |
| anchor `gh run list --workflow` calls | 0 | 1 |
| `WARNING:` lines on stderr | 0 | 1 |
| `COMPLETED_AFTER` assignments | 1 | 2 (default floor, then pulled back) |

The dispatch run's `COMPLETED_AFTER` came out at exactly `now - 3600`. Since `GITHUB_WORKFLOW` was set in both, it can't be what selects the branch.

The structure agrees: `if [ -n "$cron_minute" ]` opens at L89 with no indent; `if [ -n "${GITHUB_WORKFLOW:-}" ]` is indented two spaces at L132, inside it, and closes at L163; the `else` at L166 — also unindented, so it pairs with L89 — is `COMPLETED_AFTER=$(date -d '1 hour ago' +%s)`. The clamp is `floor_cap=$((intended - 21600))`, not `now - 21600` as quoted; `intended` is assigned only at L93/L95 inside the cron branch, so the clamp couldn't be reachable from outside it. The header says the same at L19-20: "For non-schedule events or non-hourly crons, falls back to a now-anchored 1h window."

Your second point stands on its own and is now fixed in `825dbee`: on the clean path nothing prints the bounds — only the clamp WARNING does — so "read the window off the script" asked for something unreadable. The instruction now says to derive it (run time, less an hour). Both paragraphs also drop the trigger framing for the payload framing, since that's what the branch actually keys on.

tend-agent (commented):
Both window paragraphs describe `list-recent-runs.sh` as it exists **on this branch**, not as it will exist after merge. The branch is behind `main`: #965 (`0712d15`) landed at 10:51Z, ~20 minutes before `1af440e`, and replaced the cron-payload anchoring outright. `git merge-base --is-ancestor 0712d15 825dbee` is false, and this PR doesn't touch the script, so the merge takes main's version.

In that version there is no `cron_minute`, no `intended`, and no reference to the event payload anywhere in the file — `grep -n 'GITHUB_EVENT\|schedule\|cron'` over it returns nothing. The single branch is `if [ -n "${GITHUB_WORKFLOW:-}" ]`: the floor is the previous successful run of the calling workflow, clamped at `now - 21600`, with both `WARNING:` lines on that path. `COMPLETED_AFTER=$((now - 3600))` is its `else` — for invoking the script outside Actions. The rewritten header says so directly: *"manual dispatches all need no special handling: whatever run last succeeded is the anchor"* and *"Outside GitHub Actions (no GITHUB_WORKFLOW), the window is simply the past hour"*. `generator/tests/test_list_recent_runs.py` reaches the 1h path the same way — the test that covers it is docstringed *"An ad-hoc invocation (no GITHUB_WORKFLOW) covers the past hour and never queries for an anchor"* and gets there by `del env["GITHUB_WORKFLOW"]`, not by changing a payload.

Ran it rather than only reading it, same method as last round but against the merged tree — `{"inputs":{}}` as the payload, `GITHUB_WORKFLOW=review-reviewers`, `bash -x`, at 11:28:34Z:

```
+ now=1786620514          # 11:28:34Z
+ floor_cap=1786598914    # 05:28:34Z
+ prev_start=2026-08-13T11:08:23Z
+ COMPLETED_AFTER=1786619303   # 11:08:23Z — the anchor, not now-3600 (10:28:34Z)
```

Dispatch payload, and it anchored. So after merge a dispatched run's window is "since the previous successful `review-reviewers` run, floored 6h back". With the cron gone, consecutive dispatches will usually be more than 6h apart, which makes the clamp the *normal* case and its `WARNING:` a routine one — not something that can't fire on this path. Two knock-ons for a session that follows the new text: it reports an hour when the script handed it up to six, and it's told the coverage-gap warnings live on a path it isn't on, which contradicts the paragraph immediately above the insertion in `review-reviewers/SKILL.md` — the one that routes exactly that `WARNING:` to a coverage-gap record.

This also settles the earlier round: we were reading two different files. Your trace is accurate for the branch's copy, and my withdrawal cited L89/L132/L166 in that same copy — neither of us was looking at what merges. Rebasing onto `main` is the fix; the suggestions inline are the post-rebase wording.

The PR description carries the same claim (`anchors its window to the cron tick only under an hourly cron`), so it's worth updating there too. The workflow-file header, the frontmatter description, and the table row all read correctly to me.

.claude/skills/running-tend/SKILL.md
```diff
@@ -25,7 +25,18 @@ Tend has Claude-powered workflows beyond the generated `tend-*` set:
...
+`review-reviewers` runs only on `workflow_dispatch` — dispatch it as a
+spot-check after a release, harness switch, or model bump, not on a cadence. The
+per-repo `tend-review-runs` carries the routine loop; the workflow file's header
+explains the pause.
+
+A dispatched run analyzes **the last hour only**: `list-recent-runs.sh` anchors
+its window — and warns about coverage gaps — only when the event payload carries
+an hourly cron, so every other trigger silently takes the hour ending at the run.
+Dispatch it within the hour of whatever you want it seen, and read the result as
+a sample of that hour rather than a verdict on the change.
```
```suggestion
A dispatched run's window opens at the previous successful `review-reviewers`
run, floored 6h back (`list-recent-runs.sh`; the now-anchored hour is its
fallback for being invoked outside Actions, where `GITHUB_WORKFLOW` is unset).
With no cron, dispatches are usually further apart than 6h, so the clamp is the
normal case — the run covers the last 6h and warns on stderr that the rest is a
coverage gap. Dispatch it within ~6h of whatever you want it to see.
```

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
```diff
@@ -220,6 +220,8 @@ If empty, record the run as all-clear per "Recording below-threshold findings" a
 
 If the script printed a `WARNING:` on stderr, the list is known-incomplete — the window was clamped, no anchor was found, or a workflow hit the fetch limit. Record a coverage gap naming the missing span instead of an all-clear, whether or not the list came back empty; the next run's floor advances past that span regardless, so an unrecorded gap is never revisited.
 
+**State the window you analyzed.** The anchoring that tiles consecutive windows keys on an hourly cron in the event payload; every other trigger — `workflow_dispatch` included — takes the hour ending when the script ran, and takes it silently, since both the anchor and the coverage-gap warnings above sit on the anchored path. Nothing prints the bounds on that path, so derive them (the script's run time, less an hour) rather than looking for them. Scope every claim to that hour — "no problems in the hour to 14:05Z", never "no problems" — and say so plainly when a run was dispatched to check on something that landed before it, which this workflow's dispatch-only trigger makes the common case.
```
```suggestion
**State the window you analyzed.** Inside Actions the window always opens at the previous successful run of this workflow, floored 6h back; the now-anchored hour is the fallback for invoking the script outside Actions, not for a `workflow_dispatch`. Nothing echoes the bounds on the clean path, so state the floor as the previous successful run's start. This workflow being dispatch-only, consecutive runs are usually more than 6h apart, which makes the clamp `WARNING:` above the normal case rather than an edge one. Scope every claim to that window — "no problems since 08:12Z", never "no problems" — and say so plainly when a run was dispatched to check on something that landed before the floor.
```

.claude/skills/running-tend/SKILL.md
max-sixty: You're right, and the disagreement was never about the code — it was that #965 landed 20 minutes before my first push and I never rebased. Rebased now; both paragraphs are rewritten against the merged tree in `05231ef`.

Ran the merged script both ways to confirm, with the same dispatch payload each time:

```
GITHUB_WORKFLOW=review-reviewers   COMPLETED_AFTER=1786619303   # the 11:08:23Z anchor
(unset)                            COMPLETED_AFTER=1786617380   # now - 3600
```

So the anchored path is trigger-independent and the past-hour path is the out-of-Actions fallback, exactly as you had it. Dropping `--event schedule` from the anchor query is what makes dispatch-to-dispatch anchoring work, which is also why the clamp is now the normal case rather than an outage signal.

I took the substance of both suggestions rather than the wording. The skill's note is shorter than proposed: the paragraph directly above already routes the clamp `WARNING:` to a coverage-gap record, so repeating that there would restate a rule that has a home three lines up. What was missing is only the scoping instruction, so that's all the block carries now. The overlay paragraph is close to yours.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
max-sixty: You're right, and the disagreement was never about the code — it was that #965 landed 20 minutes before my first push and I never rebased. Rebased now; both paragraphs are rewritten against the merged tree in `05231ef`.

Ran the merged script both ways to confirm, with the same dispatch payload each time:

```
GITHUB_WORKFLOW=review-reviewers   COMPLETED_AFTER=1786619303   # the 11:08:23Z anchor
(unset)                            COMPLETED_AFTER=1786617380   # now - 3600
```

So the anchored path is trigger-independent and the past-hour path is the out-of-Actions fallback, exactly as you had it. Dropping `--event schedule` from the anchor query is what makes dispatch-to-dispatch anchoring work, which is also why the clamp is now the normal case rather than an outage signal.

I took the substance of both suggestions rather than the wording. The skill's note is shorter than proposed: the paragraph directly above already routes the clamp `WARNING:` to a coverage-gap record, so repeating that there would restate a rule that has a home three lines up. What was missing is only the scoping instruction, so that's all the block carries now. The overlay paragraph is close to yours.

.claude/skills/running-tend/SKILL.md
tend-agent: Confirmed against the merged tree — the branch carries `main`'s script verbatim (`git diff origin/main...HEAD -- plugins/tend-ci-runner/scripts/list-recent-runs.sh` is empty), and its only branch is [`GITHUB_WORKFLOW` set vs unset](https://github.com/max-sixty/tend/blob/05231ef744fe5dd7f160e59e6f3ac1226d10d7cc/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L62), with the [6h floor](https://github.com/max-sixty/tend/blob/05231ef744fe5dd7f160e59e6f3ac1226d10d7cc/plugins/tend-ci-runner/scripts/list-recent-runs.sh#L73-L76) and its stderr warning inside the set arm. The overlay paragraph as written matches that; nothing outstanding here.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
tend-agent: The shorter note works — the coverage-gap routing three lines up does cover the `WARNING:` case, so scoping is the only thing that was missing.

One leftover of the same kind, outside the diff: the survey subagent's run→output map still tells it to check `tend-notifications` [in the past hour](https://github.com/max-sixty/tend/blob/05231ef744fe5dd7f160e59e6f3ac1226d10d7cc/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L273), which is the same stale hour this PR is removing everywhere else. When the floor is 6h back, a notifications run early in the window has its comments looked for in the wrong span and reads as no output — a false negative on the one workflow whose output isn't reachable from `headBranch`. `in the window from Step 1` would match the surrounding instructions.

Nit, cosmetic: the PR branch prefix a few sections down is still `hourly/review-$GITHUB_RUN_ID-…`.

tend-agent (approved):
The rebase lands the window description on what merges — I re-ran the merged tree's `list-recent-runs.sh` under a `workflow_dispatch` payload and it takes the anchored path, floor at the previous successful run clamped to `now - 21600`, so both new paragraphs match the script.

Two `past hour` references survive elsewhere in `review-reviewers/SKILL.md`, both outside this diff. The one with teeth is in Step 2's run→output mapping: the `tend-notifications` bullet reads `check for recent bot comments/issue-close events in the past hour`. Step 1 now hands the session up to 6h of runs, so a notifications run that finished four hours before the dispatch has its output outside that lookback and reads as silent. The repo-wide sweep above it doesn't cover this — it's parameterised on `$WINDOW_START`/`$WINDOW_END` and catches *uniform* emptiness, not a partial miss on the older half of the window. Under the cron the mismatch only bit on a dropped tick; dispatch-only makes it the normal case, which is the same reason the two paragraphs in this PR needed rewording.

The other is cosmetic, but it sits in the line that defines what the evidence log means — "Recording below-threshold findings" still says `For all-clear hours, record a single Low-evidence "all-clear" entry` and `confirm which hours were analyzed`.

Happy to push a commit moving both to window wording if you want it folded in here rather than left for a follow-up.

plugins/tend-ci-runner/skills/review-reviewers/SKILL.md
max-sixty: Folded in at `ae0e42e` — all three, since they're one-line edits in a file this PR already touches.

The `tend-notifications` one was worth catching: it's now "inside the window from Step 1", so the lookback follows the floor instead of a fixed hour. The evidence-log line moves from "all-clear hours" / "which hours were analyzed" to window wording, and the branch prefix is `review-reviewers/` rather than `hourly/` (nothing filters on the old prefix — checked before changing it).

`grep -n 'past hour\|hourly'` over the skill and the overlay now returns nothing; the only surviving "hourly" is in the workflow header, describing what was paused.

tend-agent: Checked all three against `ae0e42e` — the notifications bullet now reads `inside the window from Step 1`, the evidence-log line is window-shaped, and the branch prefix is `review-reviewers/`. Both of your verification claims hold independently: `grep -rn 'past hour\|hourly'` over the skill and the overlay is empty, and `grep -rn 'hourly/'` repo-wide finds nothing that filtered on the old prefix. Approval stands.

One leftover of the same family, strictly follow-up-or-ignore: `tick` survives in three places the greps didn't reach. Two are harmless synonyms for "run" ([L77](https://github.com/max-sixty/tend/blob/ae0e42e9149056708ba85bc59e29c3af53ab22cd/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L77), [L213](https://github.com/max-sixty/tend/blob/ae0e42e9149056708ba85bc59e29c3af53ab22cd/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L213)). The third carries a number that the pause invalidates: [L250](https://github.com/max-sixty/tend/blob/ae0e42e9149056708ba85bc59e29c3af53ab22cd/plugins/tend-ci-runner/skills/review-reviewers/SKILL.md#L250) justifies the `created_at` ceiling with `this skill starts 20–40 min after its tick, so on a busy repo that is most windows` — that lag was the cron's queue-to-start delay, and a dispatch starts within a minute of the button. The ceiling is still right (the session's own runtime is the gap now, and the window floor is up to 6h back rather than one), just argued from a stale figure.

Issue #967, opened 2026-08-13
review-reviewers-tracking: 2026-08


tend-agent:
Monthly tracking issue for `review-reviewers`. Per-target evidence lives in secret gists owned by the bot. A comment below is posted when each target's gist is first created.

**Do not close manually** — a new issue is created each month.

tend-agent:
Triage fired on this tracker because `review-reviewers-tracking` wasn't in the generator's `BOOKKEEPING_LABELS` — only `tend-outage` and `tend-rate-limit` were, so `tend-triage`'s `if:` let a bot-authored ledger through and booted a session with nothing to triage. #968 adds both monthly tracking labels; the Worker already treated them as bookkeeping, so this closes that drift rather than inventing a rule. `review-runs-tracking` matters beyond this repo — it comes from a bundled skill on a generated workflow, so every adopter running `tend-review-runs` was paying the same monthly boot.

Leaving this issue open and untouched otherwise. On why it exists at all: #799 was closed at 11:04 on the basis that #966 pauses the hourly sweep, but #966 is still open, so the `47 * * * *` cron ran at 11:08 and the find-or-create logic opened this one at 11:10. It'll keep reappearing each tick until #966 merges.

max-sixty:
Closing as dead bookkeeping: `review-reviewers` is paused as of #966 (manual dispatch only), so nothing appends here. This issue was recreated by the 11:08Z tick minutes after August's original tracker was closed, and the final scheduled run fired at 12:02:14Z — 90 seconds before the pause merged. The per-target evidence gists are unaffected and still readable; a future dispatched run finds or creates a tracker for its own month.

> _This was written by Claude Code on behalf of max-sixty_

PR #968, opened 2026-08-13
fix(generator): skip the monthly tracking issues in triage and mention

tend-agent:
## Problem

`BOOKKEEPING_LABELS` held `tend-outage` and `tend-rate-limit`, but not the monthly tracking issues the bundled `review-runs` and `review-reviewers` skills open. Those trackers are also bot-authored ledgers with nothing to triage — the body of each says "Do not close manually" — so `issues: opened` on one booted a full `tend-triage` session that could only no-op. #967 is the current instance: a scheduled `review-reviewers` run opened the August tracker and `tend-triage` fired on it.

This isn't a one-off. This repo has 19 open-or-closed issues carrying `review-runs-tracking` / `review-reviewers-tracking`, each of which fired a triage session at creation, and `review-runs-tracking` comes from a bundled skill on a generated workflow — so every adopter running `tend-review-runs` pays the same monthly boot. The evidence that these labels belong on the list is already in the tree: the website Worker independently enumerates them as bookkeeping ([`worker/src/index.ts`](https://github.com/max-sixty/tend/blob/9388505/worker/src/index.ts#L147-L153)) to keep them out of the `issues` activity count, while the generator's copy had drifted behind.

## Solution

Add both tracking labels to `BOOKKEEPING_LABELS`. The comment on that tuple already anticipated this ("the next such label is added in one place"): it feeds the `not_bookkeeping()` macro, so `tend-triage`'s job `if:` and `tend-mention`'s `should_run` both pick the labels up with no template change.

Not added: `nightly-cleanup`, which the Worker also lists — nothing in the tree creates that label and this repo has no such label, so it would be dead config on the generator side.

## Consequence for `tend-mention`

One list, two gates: `not_bookkeeping()` sits on [`mention.yaml.j2`](https://github.com/max-sixty/tend/blob/9388505f77c44f3454456d8bfd329c8033de185d/generator/src/tend/templates/mention.yaml.j2#L120)'s `issue_comment` branch as well as [`triage.yaml.j2`](https://github.com/max-sixty/tend/blob/9388505f77c44f3454456d8bfd329c8033de185d/generator/src/tend/templates/triage.yaml.j2#L13)'s job, so this also stops an `@bot` comment on a tracking issue from booting an immediate reply — a behavior change reaching every adopter, and a maintainer's call rather than an obvious win.

It differs from the two labels already on the list. Their rationale is the comment→run→comment loop the action's own auto-comments create, which trackers don't have, and trackers are threads maintainers are meant to read and follow up on — `review-reviewers` posts gist links there so they're discoverable. What it degrades to is latency, not silence: `tend-notifications` picks the mention up as a stale unanswered item on its next tick. The alternative is splitting the tuple into triage-only and mention-only sets, which buys back the immediate reply at the cost of the single list the tuple's comment is built around; I'd take the latency, but say so here rather than land it unflagged.

## Testing

`test_bookkeeping_labels_cover_the_monthly_trackers` asserts the rendered `tend-triage` `if:` and `tend-mention` body carry a guard for each of the four labels. It fails on `main` (`tend-triage does not skip 'review-runs-tracking'`) and passes here. The nine regtest outputs that pin the rendered `if:` are reset; full generator suite is 459 passed.

Tend's own `tend-*.yaml` still carry the two-label `if:` until the next release regenerates them from a published tag, per the pinning rule in `CLAUDE.md`.

## Scope

Refs #967 — deliberately not `Closes`, since that tracker is a live ledger `review-reviewers` appends gist links to and recreates when closed.

### DIFF
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -103,8 +103,14 @@ _JINJA.globals["openai_key_secret"] = OPENAI_KEY_SECRET
 # each row the rate-limit preflight appends is a comment, which would fire
 # tend-mention, whose handle job trips the same limit and appends another.
 # A global rather than a literal per template, so the next such label is added
-# in one place.
-BOOKKEEPING_LABELS = ("tend-outage", "tend-rate-limit")
+# in one place. The monthly `*-tracking` issues are the same shape: the bot
+# opens one, so `issues: opened` fires a triage session on its own ledger.
+BOOKKEEPING_LABELS = (
+    "tend-outage",
+    "tend-rate-limit",
+    "review-runs-tracking",
+    "review-reviewers-tracking",
+)
 _JINJA.globals["bookkeeping_labels"] = BOOKKEEPING_LABELS
 
 
--- a/generator/tests/_regtest_outputs/test_generate.test_fork_guard_rendered_shape_regtest[tend-triage.yaml].out
+++ b/generator/tests/_regtest_outputs/test_generate.test_fork_guard_rendered_shape_regtest[tend-triage.yaml].out
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   triage:
-    if: github.repository_owner == 'test-owner' && contains(github.event.issue.labels.*.name, 'tend-outage') == false && contains(github.event.issue.labels.*.name, 'tend-rate-limit') == false
+    if: github.repository_owner == 'test-owner' && contains(github.event.issue.labels.*.name, 'tend-outage') == false && contains(github.event.issue.labels.*.name, 'tend-rate-limit') == false && contains(github.event.issue.labels.*.name, 'review-runs-tracking') == false && contains(github.event.issue.labels.*.name, 'review-reviewers-tracking') == false
     runs-on: ubuntu-24.04
     environment:
       name: tend
--- a/generator/tests/_regtest_outputs/test_generate.test_sandbox_levers_regtest.out
+++ b/generator/tests/_regtest_outputs/test_generate.test_sandbox_levers_regtest.out
@@ -103,7 +103,7 @@ jobs:
       (github.event_name == 'issues' &&
         contains(github.event.issue.body, '@test-bot')) ||
       (github.event_name == 'issue_comment' &&
-        contains(github.event.issue.labels.*.name, 'tend-outage') == false && contains(github.event.issue.labels.*.name, 'tend-rate-limit') == false)
+        contains(github.event.issue.labels.*.name, 'tend-outage') == false && contains(github.event.issue.labels.*.name, 'tend-rate-limit') == false && contains(github.event.issue.labels.*.name, 'review-runs-tracking') == false && contains(github.event.issue.labels.*.name, 'review-reviewers-tracking') == false)
     runs-on: ubuntu-24.04
     environment:
       name: tend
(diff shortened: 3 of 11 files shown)


tend-agent (commented):
Both label strings match what actually gets applied: `review-runs/SKILL.md`'s `TRACKING_LABEL="review-runs-tracking"`, `review-reviewers/SKILL.md`'s `review-reviewers-tracking`, and the `init-tracking` job in `review-reviewers.yaml`. All three set the label via `gh issue create --label`, so it's present in the `issues: opened` payload and the guard actually fires — worth stating, since a label applied after creation would render the same `if:` and skip nothing. Generator suite passes here (459).

One consequence the PR body doesn't mention: `BOOKKEEPING_LABELS` feeds `not_bookkeeping()`, which gates `tend-mention`'s `issue_comment` branch as well as `tend-triage`'s job. So a maintainer commenting `@tend-agent …` on a tracking issue stops getting an immediate reply. That's a different situation from the two labels already on the list — their rationale is the comment→run→comment loop the action's own auto-comments create, which trackers don't have, and the trackers are threads maintainers are meant to read and follow up on (`review-reviewers` posts gist links there so they're discoverable). It degrades to latency rather than silence: `tend-notifications` picks the mention up as a stale unanswered item on its next tick. I'd take that trade over splitting the list into triage-only and mention-only sets, but it's a behavior change reaching every adopter and not one the PR body flags, so it's a maintainer's call rather than mine.

Aside, outside this diff: `worker/README.md` describes the `issues` bucket as dropping "four bookkeeping labels" and lists four, while `worker/src/index.ts`'s own `BOOKKEEPING_LABELS` holds five — `tend-rate-limit` is missing from the README's list. Happy to fix that separately if useful.

tend-agent:
Withdrawing this — it re-proposes a change you've already declined, and it fails the cost gate you merged yesterday. Neither fact was in the PR body, and both are more decisive than the trade-off my review raised.

#800 proposed the same skip and you closed it on 2026-08-01: "Not worth the complication ... The guard already covers this case." The mechanism here is cheaper than #800's — two strings in the existing `BOOKKEEPING_LABELS` tuple rather than a new `.config/tend.yaml` surface — but that's a discount on a decision that was already "no", not a new argument.

Then #960 (merged 2026-08-12) codified the general rule. The failure here is squarely waste-class — a no-op triage session, the gate's own example — and [Gate 3](https://github.com/max-sixty/tend/blob/9388505f77c44f3454456d8bfd329c8033de185d/plugins/tend-ci-runner/shared/review-gates.md#L41) says a waste-class failure "supports only a fix that is itself nearly free," naming "another skip-gate" as failing "at any occurrence count," judged by what the fix leaves behind rather than its line count. What this one leaves behind is a failure mode of its own, which the review on this PR found independently: `not_bookkeeping()` gates [`tend-mention`'s `issue_comment` branch](https://github.com/max-sixty/tend/blob/9388505f77c44f3454456d8bfd329c8033de185d/generator/src/tend/templates/mention.yaml.j2#L120) too, so an `@bot` comment on a tracking issue would stop getting an immediate reply — trading a real interaction for saved compute, in every adopter.

Branch kept at `fix/triage-skips-tracking-issues` (`9388505`) rather than deleted, so reopening is a click if you read the adopter-wide reach differently.

<details><summary>What I did before finding this, and the one piece I kept</summary>

Earlier in this session I updated the PR body with a "Consequence for `tend-mention`" section, in response to the review — that's why the body now flags a trade-off and then the PR closes. The prior decision surfaced afterward, on the dedup pass.

Kept separately: #969, correcting `worker/README.md`'s bookkeeping-label list, which described four labels while [`worker/src/index.ts`](https://github.com/max-sixty/tend/blob/fffe32f44000696aa576a689eb3516aeb4e04552/worker/src/index.ts#L147-L153) has held five. Docs-only and unrelated to the gate — close it too if you'd rather not have the PR.

#967 is unaffected; it stays open as the live ledger.

</details>

PR #970, opened 2026-08-13
test(mention): run the verify gate instead of matching its text

max-sixty:
tend-mention's engagement gate decides whether a comment or review summons a session — a false negative leaves someone un-answered, a false positive burns a session that can only exit silently. It lived as ~200 lines of bash inlined in `mention.yaml.j2`, where nothing shellchecked it and the tests could only assert that certain substrings appeared in the rendered YAML in a certain order. This is the follow-up #965 named and deferred to keep that diff reviewable.

The gate moves to `generator/src/tend/templates/mention-verify.sh`, the shape `review-gate.sh` and `notifications-check.sh` already use: env in, `GITHUB_OUTPUT` out, inlined verbatim into the generated workflow. The bot name arrives as `$BOT_NAME` instead of a Jinja substitution, which is what lets the script run outside the renderer.

**The logic is unchanged.** Normalizing comments, quoting and `$BOT_NAME`, the old inlined body and the new script are the same 119 lines with zero differences.

## Testing

38 tests in `generator/tests/test_mention_verify.py` run the script against a fake `gh` whose `--jq` goes through real jq, so the gate's own filters do the reductions: dispatch-payload forgery (non-numeric ids, an unknown kind, a review or comment the API doesn't have, a comment bound to another PR), the mention paths, the bot-review handoff gate, contentless approvals, and the engagement heuristics including a paginated lookup.

Seven mutations confirm they bite — dropping the fresh-inline narrowing, un-keying the handoff gate from the review author, dropping `ascii_downcase`, dropping the self-comment guard, ungating the inline fetch from the review kind, unguarding the comment fetch, and dropping an env var from the workflow each fail 1–3 tests.

`test_generate.py` keeps what generation owns and sheds seven shape tests: the `env:` block is pinned by exact value, plus a scan proving nothing the script reads is left unset — an unwired name is simply empty at runtime, so the gate would answer on a blank (a missing `COMMENT_AUTHOR` reads as "not the bot") on a green job.

Also checked: `shellcheck` covers the new file through the existing `templates/*.sh` glob, `actionlint` (which shellchecks inlined `run:` blocks) passes on the generated workflow, and the script ships in the built wheel.

<details><summary>One thing to watch</summary>

The three engagement `--jq` filters moved from single to double quotes so `$BOT_NAME` interpolates. A future filter containing `$` or a backtick would now be shell-expanded rather than passed through to jq. None does today.

</details>


generator/tests/test_mention_verify.py
```diff
@@ -0,0 +1,602 @@
...
+
+
+def test_the_words_judged_come_from_the_api_not_the_payload(
+    env: dict[str, str],
+) -> None:
+    """The body, url and timestamp all come off the fetched record — the
+    payload carries identifiers only."""
+    _review(env, body=f"@{BOT} please look")
+
+    _review(env, body=f"@{BOT} please look")
```
tend-agent: Duplicated setup call — the second `_review` re-writes the same body onto the same fixture, so it's a no-op.

```suggestion
    _review(env, body=f"@{BOT} please look")
```

generator/tests/test_mention_verify.py
```diff
@@ -0,0 +1,602 @@
...
+
+case "$*" in
+  "pr view "*)
+    emit "$(cat "$PR_AUTHOR_JSON")"
+    ;;
+  *"/reviews/"*"/comments"*)
+    emit_paged "$(cat "$INLINE_JSON")"
+    ;;
+  *"/pulls/comments/"*)
+    [ -n "${MISSING_COMMENT:-}" ] && exit 1
```
tend-agent, line 55: `MISSING_COMMENT` is never set by any test, so the script's `comment $PAYLOAD_ID not found — skipping` branch is the one dispatch-resolution failure path with no coverage — its `MISSING_REVIEW` twin has `test_a_review_the_api_does_not_have_is_refused`. Since the fake already supports it, the missing test is two lines: `env["PAYLOAD_KIND"] = "pull_request_review_comment"; env["MISSING_COMMENT"] = "1"` then assert `("false", "")`. Worth having — that branch is what stops a forged dispatch naming a comment id that doesn't exist from reaching the PR-binding check with an empty `$COMMENT`.

tend-agent (approved):
Two non-blocking test nits inline; the extraction itself checks out. I diffed the old and new rendered `run:` blocks with comments stripped and `$BOT_NAME` substituted back — code and indentation are byte-identical, the only residue being the single→double quote flip the PR body already calls out. The seven shed shape tests all have behavioural successors, and the one assertion that looks dropped (`-gt "0"` guard) is now covered better: `test_engagement_survives_a_paginated_lookup` runs `PAGES=2`, so a numeric comparison on a two-page stream trips errexit and fails the run rather than passing a grep.

generator/tests/test_mention_verify.py
max-sixty: Dropped in 8d22d56a — left over from reworking the assertion to read the outputs map.

generator/tests/test_mention_verify.py
max-sixty: Added in 8d22d56a. Worth noting the branch is outcome-equivalent to its neighbour — an empty `$COMMENT` makes the PR-binding check compare `""` against the expected URL and skip anyway — so the coverage it buys is against the fetch being left unguarded, where `bash -e` aborts the step red instead of skipping. Mutation-checked that way: removing the `if !` wrapper fails the new test.

tend-agent: Both landed — re-ran the suite at `8d22d56a`, 38 pass.

Your framing is the accurate one and mine was wrong: I said the branch stops a forged dispatch from *reaching* the PR-binding check, but under `bash -e` an unguarded `COMMENT=$(gh api ...)` never gets that far — the failed substitution aborts the step first. Reproduced it by deleting the `if !` wrapper: `test_a_comment_the_api_does_not_have_is_refused` is the only failure, and it fails on `returncode == 1` rather than on the verdict, which is exactly the red-vs-skip distinction you describe. So the test is pinning the guard's existence, not its verdict.

PR #100, opened 2026-03-29
fix(ci): add setup-uv to workflows and derive repo owner from $GITHUB_REPOSITORY

tend-agent:
## Summary

Two CI behavior fixes identified by review-reviewers analysis:

- **Add `setup-uv` to CI workflows**: The bot wastes 3-5 tool calls per session trying to locate `uv` — CLAUDE.md says `cd generator && uv run pytest` but `uv` isn't installed in CI runners. Observed 5 times across review, mention, and nightly runs (runs 23608757804, 23682034006, 23703363227, 23703434151, 23716607858). Adding `astral-sh/setup-uv@v6` as a setup step ensures `uv` is on PATH before Claude runs.
- **Derive repo owner from `$GITHUB_REPOSITORY`**: The `list-recent-runs.sh` script had a hardcoded owner assumption.

### `uv` PATH discovery waste
- **Evidence level**: Medium
- **Occurrences**: 5 (4 historical from tracking issue #12 + 1 this run)
- **Change type**: Targeted fix (config change + workflow regeneration)
- **Passes both gates**: Yes — Medium/5 meets threshold, targeted fix is proportionate

## Test plan

- [x] All 109 generator tests pass
- [ ] CI workflows regenerated successfully
- [ ] Verify `uv` is available in next CI run

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #971, opened 2026-08-13
simplify: one tested review-anchor resolver, skill depth into references

max-sixty:
Two rounds of cutting, both aimed at the same shape: logic that mattered enough to be right, living in prose that nothing checked.

## One resolver for "which review anchors this head"

A review's `.commit_id` can't be read on its own. Replying to a review thread makes GitHub wrap the reply in a synthetic zero-body review anchored at the then-current head, so the newest record routinely points at a commit nothing reviewed. And a force-push doesn't just move the head — it re-points earlier reviews' `.commit_id` at the new one, so a review of code that no longer exists reports the current head.

Every caller that reads an anchor has to discount both, and five did: three sites in `review` (step 1's anchor resolution, step 5's pre-post guard, step 6's orphan probe) and two in `weekly` (the redundant-approval guard and the stale-approval dismissal). All five were untested shell, each carrying its own paragraph of rationale — and they had already drifted, to the point that `weekly` documents its filter ordering as *deliberately* different from `review`'s.

`plugins/tend-ci-runner/scripts/bot-review-state.sh` answers the question once and emits it as JSON — the newest substantive review and whether a rewrite postdates it, whether anything genuinely anchors the current head, the orphan body a partially-failed review POST left behind, and the fresh and stale approvals. Each call site now reads one field.

The consequences of getting any of this wrong are outward, which is what earns the mechanism its place: a reply container counted as a review discards a real one after the work is done; a missing rewrite filter leaves a rebased PR carrying an approval it never earned, or makes the recovery `PUT` overwrite a published review with this run's findings.

## A head-resolution bug the extraction surfaced

Centralizing the anchor logic made a pre-existing bug visible, so it's fixed here too. `gh pr view --json commits` issues `commits(first: 100)` — unpaginated, oldest-first — so past 100 commits `.commits[-1]` is commit #100 rather than the head. Every head-keyed field then matches nothing and goes quiet rather than wrong-loud: the pre-post guard stops firing, so a fold-in re-run posts a second review on a commit it already reviewed; the 422 recovery finds no orphan and duplicates instead of editing it; `review`'s moved-head check inverts, so a push that *did* move the head reads as "HEAD hasn't moved" and the review posts against a stale commit; and `weekly`'s redundant-approval guard never matches, collecting an approval per weekly run on one commit.

Four sites now read `headRefOid`, which is the head by definition and can't truncate. The test fixture serves a deliberately wrong-last `commits` list alongside it, so reverting fails nine tests rather than passing on a short PR.

## Skill depth into `references/`

`running-in-ci` was 677 lines against the 500-line cap `authoring-skills` sets, and that skill's own remedy is splitting into `references/`, which costs nothing until read. Grounded-analysis depth, the other-repos rules, and the bar for proposing a skill PR move there; SKILL.md keeps each rule and a pointer. Every bundled skill is now under the cap.

| skill | before | after |
|---|---|---|
| `running-in-ci` | 677 | 462 |
| `review` | 533 | 474 |
| `weekly` | 98 | 78 |

## Testing

20 tests in `generator/tests/test_bot_review_state.py` run the script against a fake `gh`: reply containers vs. real reviews, rewrite timing on both sides of a review, fresh vs. stale approvals, a dismissed one that must not be re-dismissed, orphan bodies, and the repo being named on every call. Five mutations confirm they bite — counting reply containers as substantive, counting a DISMISSED review as an approval, dropping the rewrite filter from `at_head` or from `orphan_id`, and the ordering flip below each fail 1–2 tests.

The ordering subtlety is the one `weekly` had split two ways, and the single implementation settles it: take the newest approval, *then* test whether it predates the rewrite. Filtering first names a pre-rewrite approval that a later one already superseded — dismissing a review that no longer decides anything while the one that does stands untouched.

<details><summary>What this trades away</summary>

The brief was to accept modest reductions in feature precision. Only one materialized: "rewriting is authoring" is now a sentence inside `grounded-analysis.md` rather than a section with a worked example. Everything else here is behaviour-preserving — the duplication turned out to be a better trade than cutting precision.

Reading the anchor state is now up to three script calls in a review run rather than the same work inlined three times, each fetching one endpoint more than the site strictly needs. That's compute, not correctness, and the repo's own rule puts it a distant third.

</details>

### DIFF
--- a/plugins/tend-ci-runner/scripts/bot-review-state.sh
+++ b/plugins/tend-ci-runner/scripts/bot-review-state.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+# Resolves which of the bot's reviews actually anchors the PR's current head,
+# and emits the answer as one JSON object.
+#
+# Usage: bot-review-state.sh <pr-number>
+#
+# Two GitHub behaviours make `.commit_id` alone unreadable, and every caller
+# that reads a review anchor has to account for both:
+#
+#   Reply containers. Replying to a review thread (POST
+#   /pulls/{n}/comments/{id}/replies) makes GitHub wrap the reply in a synthetic
+#   zero-body COMMENTED review anchored at the then-current head. The newest
+#   review record therefore routinely points at a commit nothing reviewed. A
+#   review counts as *substantive* when it has a body, owns a top-level
+#   (non-reply) inline comment, or is an APPROVED — containers are always
+#   zero-body COMMENTED, while an empty-body approval and an empty-body review
+#   carrying real inline findings both still anchor.
+#
+#   Force-push re-anchoring. A rewrite does not just move the head: GitHub
+#   re-points earlier reviews' `.commit_id` at the NEW head. So a review of code
+#   that no longer exists reports the current head, and `.commit_id` alone
+#   cannot tell an ordinary push from a rewrite. Anything submitted before the
+#   newest `head_ref_force_pushed` is discounted.
+#
+# Output (one object, all fields always present; absent values are null or ""):
+#   head_sha            the PR's current head
+#   last_force_push_at  newest head_ref_force_pushed timestamp, "" if none
+#   last_substantive    {id, sha, state, at} — newest substantive bot review at
+#                       any anchor, null if none. `sha` is unreliable when
+#                       force_pushed_since is true; that is what the flag is for.
+#   force_pushed_since  true when a rewrite postdates last_substantive — the
+#                       commit the bot read was rewritten away, so its anchor
+#                       now names the current head and every incremental keyed
+#                       on it under-reports
+#   at_head             {id, state, at} — newest substantive bot review anchored
+#                       at head and submitted after the newest rewrite, else
+#                       null. Non-null means this head is genuinely reviewed.
+#   orphan_id           id of the newest body-bearing bot review anchored at
+#                       head post-rewrite, else null. A partially-failed review
+#                       POST persists the body and drops the inline comments;
+#                       this is the record to edit rather than duplicate.
+#   fresh_approval_sha  commit_id of the newest post-rewrite bot APPROVED, ""
+#                       if none — an approval that was genuinely earned
+#   stale_approval_id   id of the newest bot APPROVED when that newest approval
+#                       predates the rewrite, "" otherwise. Newest-then-test,
+#                       never test-then-newest: the question is whether the
+#                       approval now setting the PR's state is stale, so
+#                       filtering first would name a superseded one and leave
+#                       the live approval standing.
+#
+# env: GITHUB_REPOSITORY (optional; falls back to the checkout's remote)
+
+set -euo pipefail
+
+PR="$1"
+REPO="${GITHUB_REPOSITORY:-$(gh repo view --json nameWithOwner --jq '.nameWithOwner')}"
+BOT=$(gh api user --jq '.login')
+
+HEAD_SHA=$(gh pr view "$PR" --repo "$REPO" --json commits --jq '.commits[-1].oid')
+
+# Review ids owning at least one top-level inline comment. `gh api --paginate`
+# applies `--jq` per page, so reduce with a second jq over the merged stream
+# rather than inside the filter.
+SUBSTANTIVE=$(gh api --paginate "repos/$REPO/pulls/$PR/comments" \
+  --jq '.[] | select(.in_reply_to_id == null) | .pull_request_review_id' | jq -s 'unique')
+
+LAST_FORCE_PUSH_AT=$(gh api --paginate "repos/$REPO/issues/$PR/timeline" \
+  | jq -rs 'add | [.[] | select(.event == "head_ref_force_pushed") | .created_at] | max // ""')
+
+gh api --paginate "repos/$REPO/pulls/$PR/reviews" \
+  | jq -s --argjson sub "$SUBSTANTIVE" --arg bot "$BOT" --arg head "$HEAD_SHA" \
+      --arg fp "$LAST_FORCE_PUSH_AT" '
+    add
+    | [.[] | select(.user.login == $bot)] as $mine
+    | ($mine | map(select((.body | length) > 0 or (.id | IN($sub[])) or .state == "APPROVED"))) as $subs
+    | ($subs | last) as $lastsub
+    | ($mine | map(select(.state == "APPROVED")) | last) as $lastapp
+    | ($fp == "") as $norewrite
+    | {
+        head_sha: $head,
+        last_force_push_at: $fp,
+        last_substantive: (if $lastsub == null then null else
+          {id: $lastsub.id, sha: $lastsub.commit_id, state: $lastsub.state, at: $lastsub.submitted_at}
+        end),
+        force_pushed_since:
+          ($lastsub != null and $fp != "" and $fp > $lastsub.submitted_at),
+        at_head: ($subs
+          | map(select(.commit_id == $head and ($norewrite or .submitted_at > $fp)))
+          | last
+          | if . == null then null else {id, state, at: .submitted_at} end),
+        orphan_id: ($subs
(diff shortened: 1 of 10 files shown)


plugins/tend-ci-runner/scripts/bot-review-state.sh
```diff
@@ -0,0 +1,101 @@
...
+#
+# env: GITHUB_REPOSITORY (optional; falls back to the checkout's remote)
+
+set -euo pipefail
+
+PR="$1"
+REPO="${GITHUB_REPOSITORY:-$(gh repo view --json nameWithOwner --jq '.nameWithOwner')}"
+BOT=$(gh api user --jq '.login')
+
+HEAD_SHA=$(gh pr view "$PR" --repo "$REPO" --json commits --jq '.commits[-1].oid')
```
tend-agent: `commits(first: 100)` is unpaginated and oldest-first, so this is commit #100 rather than the head once a PR passes 100 commits — and every head-keyed field below then silently matches nothing. `headRefOid` is the head by definition.

```suggestion
HEAD_SHA=$(gh pr view "$PR" --repo "$REPO" --json headRefOid --jq '.headRefOid')
```

tend-agent (commented):
One finding, in the head resolution the new script now owns for all five call sites.

`gh pr view --json commits` is not paginated — [`api/query_builder.go`](https://github.com/cli/cli/blob/b9f99e85254629a7771a43bfa73b423d3b3e3836/api/query_builder.go#L158) issues `commits(first: 100)`, and the connection comes back oldest-first. On a PR with more than 100 commits `.commits[-1].oid` is commit #100, not the head, so `head_sha` names a mid-branch commit and every head-keyed field derived from it goes quiet rather than wrong-loud: `at_head` finds nothing (the step 5 pre-post guard stops firing, so a fold-in re-run can post a second review on a commit it already reviewed), and `orphan_id` finds nothing (a 422 recovery POSTs a duplicate instead of editing the orphan body). `weekly`'s redundant-approval guard fails the same way — its own `HEAD_SHA` never matches `fresh_approval_sha`, so approvals accumulate on one commit. `gh pr view --json headRefOid` answers directly and can't truncate.

The pattern is in three skill sites too, all inside these hunks, so I've suggested each; `test_the_head_is_the_last_commit_not_the_first` and the `COMMITS_JSON` fixture move with the script change (adding `headRefOid` alongside `commits` in the fixture keeps the fake `gh` serving both). Happy to push the lot as a commit if you'd rather not click through four.

Nothing else — the resolver's semantics match what each of the five call sites was doing, including the ordering split `weekly` had documented as deliberate, and the `references/` move preserves every section from the old `running-in-ci`.

- Ran the script live against #971 and #970 — output shape and the `fresh_approval_sha` on #970's approval both correct.
- `uv run pytest` in `generator/`: 510 passed at `ff781fa`. `ruff check`/`ruff format --check` clean at the pinned `v0.14.11`; `shellcheck` clean on the new script.
- Trailing comma in the jq object construction parses on jq 1.6 as well as the runner's 1.7, so no floor is raised there.
- Field-by-field against the five replaced snippets: `fresh_approval_sha` filter-then-last, `stale_approval_id` last-then-test, `at_head` composing the substantive + rewrite + anchor filters in the same order as the old pre-post guard, `orphan_id` adding the body-length test. `DISMISSED` dropping out of both approval fields is what keeps `weekly` from re-dismissing, and `ff781fa` pins it.
- Section-by-section diff of `running-in-ci`'s headings against the three new `references/` files: every subsection lands, and the two skills that named **Filing Issues in Other Repos** were both retargeted to **Other Repos**.

</details>

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -25,50 +25,20 @@ Load `/tend-ci-runner:running-in-ci` first — it contains CI security rules, po
 Before reading the diff, run cheap checks to avoid redundant work. Shell state doesn't persist between tool calls — re-derive `REPO` in each bash invocation or combine commands.
 
 ```bash
-REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
-BOT_LOGIN=$(gh api user --jq '.login')
 HEAD_SHA=$(gh pr view <number> --json commits --jq '.commits[-1].oid')
```
tend-agent: Same truncation as the script: past 100 commits this names commit #100, so `LAST_REVIEW_SHA == HEAD_SHA` never matches and step 9 stamps a commit that isn't the head.

```suggestion
HEAD_SHA=$(gh pr view <number> --json headRefOid --jq '.headRefOid')
```

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -241,36 +211,19 @@ gh api "repos/$REPO/issues/<number>/reactions" -f content="+1"
 Before posting, verify HEAD hasn't moved, the PR is still open, and no review was already posted for this commit:
 
 ```bash
-REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
-BOT_LOGIN=$(gh api user --jq '.login')
 read -r CURRENT_HEAD PR_STATE < <(gh pr view <number> --json commits,state \
   --jq '"\(.commits[-1].oid) \(.state)"')
```
tend-agent: Here the truncation inverts the guard: on a >100-commit PR `CURRENT_HEAD` is frozen at commit #100, so a push that *did* move the head reads as "HEAD hasn't moved" and the review posts against the stale commit.

```suggestion
read -r CURRENT_HEAD PR_STATE < <(gh pr view <number> --json headRefOid,state \
  --jq '"\(.headRefOid) \(.state)"')
```

plugins/tend-ci-runner/skills/weekly/SKILL.md
```diff
@@ -27,25 +27,15 @@ If no dependency PRs are open, note "0 dependency PRs to process" and continue t
 3. If the update is safe (patch/minor with green CI), check whether the bot has already approved this commit before approving — a dependabot PR open across multiple weekly runs (or already approved by `tend-review` on creation) would otherwise accumulate redundant approvals on the same `commit_id`:
    ```bash
    HEAD_SHA=$(gh pr view <number> --json commits --jq '.commits[-1].oid')
```
tend-agent: This is the one that compares against `fresh_approval_sha` two lines down — if it truncates, the redundant-approval guard never fires and a long-lived dependency PR collects an approval per weekly run.

```suggestion
   HEAD_SHA=$(gh pr view <number> --json headRefOid --jq '.headRefOid')
```

plugins/tend-ci-runner/scripts/bot-review-state.sh
max-sixty: Confirmed and applied in 7ba7303e. Checked both halves rather than taking it on the link: `commits(first: 100)` is line 158 of `query_builder.go` on that SHA, and `gh pr view 971 --json commits` returns oldest-first (`.commits[0]` is the older commit, `.commits[-1]` equals `headRefOid` only because this PR is short).

The fixture change is what makes it stick — `PR_HEAD_JSON` now serves `headRefOid` alongside a deliberately wrong-last `commits` list, so reverting to `.commits[-1]` fails nine tests instead of passing on a short PR. The quiet failures you named are exactly the nine: `orphan_id` and `at_head` both find nothing.

plugins/tend-ci-runner/skills/review/SKILL.md
max-sixty: Applied in 7ba7303e.

plugins/tend-ci-runner/skills/review/SKILL.md
max-sixty: Applied in 7ba7303e.

plugins/tend-ci-runner/skills/weekly/SKILL.md
max-sixty: Applied in 7ba7303e.

tend-agent (approved):

PR #974, opened 2026-08-13
feat: react 👀 when an issue or PR is opened

max-sixty:
The 👀 reaction only existed on `tend-mention`, keyed to the comment that mentioned the bot, so an opened issue or PR sat unacknowledged until the agent posted its first comment — minutes later, after the harness installs and the session boots.

A shared `react_eyes` macro now reacts on the issue or PR itself: the first step triage runs, and in review as soon as the gate admits an `opened` event (later trigger types are pushes and state changes on a PR the reaction is already on). One `gh api "repos/$REPO/issues/$NUMBER/reactions" -f content=eyes` call — the issues endpoint covers PRs too, and a repeat returns the existing reaction rather than erroring. A failure emits a `::warning::` instead of failing the job, since an acknowledgment isn't worth losing the run that does the work.

Mention's behavior is unchanged: it still reacts only to a comment that names the bot, and still removes the reaction when the session ends. A comment admitted because the bot is already in the thread gets nothing.

Verified the exact command against `tend-agent/tend-integration` (exit 0; a repeat left one reaction, no error); probe reaction deleted afterward.


README.md
```diff
@@ -80,6 +80,10 @@ file](docs/tend.example.yaml) and a repo-local `/running-tend` skill.
 | **notifications** | Every 15 minutes           | Polls GitHub notifications, responds to unhandled mentions, marks handled threads as read.                                                                  |
 | **review-runs**   | Daily                      | Reviews recent CI runs for behavioral problems and proposes skill/config improvements.                                                                      |
 
+The bot reacts 👀 when a job picks an event up: on the issue or PR when one
+is opened, and on a comment that mentions it, for as long as that session
+runs.
```
tend-agent:
```suggestion
The bot reacts 👀 when a job picks an event up: on the issue or PR when one
is opened, and — for as long as that session runs — on a comment that
mentions it.
```

tend-agent (approved):
One wording nit inline — the trailing "for as long as that session runs" reads as if it covers both cases, but the issue/PR reaction is permanent by design (as the macro comment says) while only mention's comes off at session end. Nothing blocking.

README.md
max-sixty: Right — split into two sentences so the duration only attaches to the mention case: an opened issue or PR keeps the reaction, and the one on a mentioning comment comes off when the session ends. Pushed in 8336b15c.

PR #977, opened 2026-08-13
simplify: one `uv run pytest` for every Python suite

max-sixty:
The three Python suites ran as three separate pytest invocations — `uv run pytest` in `generator/`, two `--no-project` runs with ad-hoc dependency sets for `proxy/` and the install-tend scripts — which is what `dev/test.sh` existed to sequence, along with per-suite failure accumulation, exit-code slack for filtered runs, and a conditional `npm ci`. The reason for three environments was the mitmproxy pin, since the proxy addon imports `mitmproxy.test` and has to run against the version production runs.

The repo root becomes a uv workspace with `generator/` as its only member, so one environment holds every Python test's dependencies, that mitmproxy pin included. Plain `uv run pytest` from the root then collects `generator/`, `proxy/`, and the install-tend scripts in a single session — with no pytest configuration at all, since collection from the root reaches all three and pytest's default `norecursedirs` already skips `node_modules` and dot-dirs. Same 542 tests as before, plus one new guard. `dev/test.sh` is deleted; `wt test` is four named steps in `.config/wt.toml` (pytest, then worker install, vitest, typecheck), and CI's three Python jobs collapse into `- run: uv run pytest`.

The one thing the old shape bought that this one doesn't: CI read the mitmproxy version out of `claude/action.yaml` with `yq`, so the pin existed once. A dev group can't do that, so the version is now named twice and `test_pinned_mitmproxy_matches_the_action` asserts the *installed* distribution matches the action's default — stronger than comparing the two files, since it also catches a lock that resolved elsewhere. The weekly bump rule in `running-tend` names the second pin.

Making `generator/` a workspace member moves where `uv build` writes: to the workspace root's `dist/`, not the member's. `pypi-release.yaml` now builds from the root with `uv build --package tend` (`--package` because a bare root build tries to build the virtual `tend-dev` through setuptools), and CI runs the same command — that workflow only fires on an immutable tag, so a break there costs a manual publish. Caught by the review bot, which is also the reason the CI step exists.

`wt test`'s arguments narrow pytest and nothing else, so a filtered run still pays for `worker/`; `uv run pytest -k render` is the Python half on its own. Dropping `{{ args }}` from the alias instead would make `wt` swallow the arguments silently.

<details>

- `uv run pytest` → 543 passed (510 + 23 + 9 before, plus the new pin test)
- `npm --prefix worker ci/test/run typecheck` → 31 passed, clean `tsc`
- `pre-commit run --all-files` → all hooks pass
- `uv build --package tend` from the root → `dist/` with sdist + wheel, `twine check` clean
- The pin test fails on drift (checked by editing `action.yaml`'s default)
- No ruleset required the deleted `test-install-tend-scripts` check

</details>


tend-agent (commented):
`uv build` from a workspace member writes to the **workspace root's** `dist/`, not the member's — so once `generator/` becomes a member, `generator/dist` is never created and `pypi-release.yaml` breaks. Its `Build package` step still succeeds (artifacts land in the repo root's `dist/`), but the step after it — `Check artifacts`, running `uv tool run --from twine twine check dist/*` under `working-directory: generator` — has no `dist/` to glob, hands the literal pattern to twine, and exits 1. The `upload-artifact` step below it (`path: generator/dist`) would find nothing either. Nothing on this PR catches it: that workflow only fires on a tag push, and the tag is immutable by ruleset, so recovery is a manual build-and-publish.

Verified on this branch:

```
$ cd generator && uv build
Successfully built /…/tend/dist/tend-0.1.16.tar.gz
Successfully built /…/tend/dist/tend-0.1.16-py3-none-any.whl
$ ls generator/dist
ls: cannot access 'generator/dist': No such file or directory
$ (cd /tmp/empty && uv tool run --from twine twine check dist/*); echo $?
ERROR    InvalidDistribution: Cannot find file (or expand pattern): 'dist/*'
1
```

The fix that matches the rest of the PR is to build from the root: drop both `working-directory: generator` lines, use `uv build --package tend`, and set the artifact `path: dist`. The `--package` matters — a bare `uv build` at the root tries to build `tend-dev` itself through the legacy setuptools backend and fails, which is the one case the root `pyproject.toml`'s "no build-system, so uv never builds it" doesn't cover. Happy to push that commit if you want it.

Everything else checks out: `uv run pytest` from the root collects all three suites (547 passed here), `cd generator && uv run pytest` still works off the same workspace env, `uv lock --check` is clean, and no ruleset required the deleted `test-install-tend-scripts` check.

max-sixty:
Confirmed and fixed in eb26c6e1 — thanks, that one was invisible to my own check because I built with `--out-dir` to keep the tree clean, which is exactly the flag that hides where the default lands.

Took the suggested shape: root build with `uv build --package tend`, no `working-directory`, artifact `path: dist`. Also added `uv build --package tend` as a step in CI's `test` job, since the point stands that nothing on a PR would have caught it — the release workflow only fires on a tag that's immutable by ruleset.

tend-agent (approved):

PR #978, opened 2026-08-13
dev: add Codex Cloud Worktrunk setup

max-sixty:
Add a repo-owned Codex Cloud setup command that installs pinned Worktrunk, verifies Tend's project-command approvals, syncs the generator, site, and worker dependencies, and installs `pre-commit` without installing Git hooks. Document the Cloud environment settings and validation command, expose `CLAUDE.md` to Codex through an `AGENTS.md` symlink, and include the setup script in shellcheck.

The symlink exposed unsafe fork-controlled instruction-file writes in the Codex action. Move that step into a tested script that restores files through Git without following destination symlinks, preserving trusted base-branch symlinks while preventing writes outside the checkout.

Tested with 549 Python tests, 31 worker tests, worker typecheck, `pre-commit run --all-files`, and an isolated execution of the Cloud script that installed Worktrunk 0.73.0 and reported every `.config/wt.toml` command approved.


.config/codex-cloud/environment.sh
```diff
@@ -0,0 +1,32 @@
...
+  "wt step copy-ignored",
+  "cd site && npm install --prefer-offline --no-audit --no-fund",
+  "cd site && npm run dev -- --port {{ branch | hash_port }}",
+  "lsof -ti :{{ branch | hash_port }} -sTCP:LISTEN | xargs kill 2>/dev/null || true",
+  "uv run pytest {{ args }}",
+  "npm --prefix worker ci --prefer-offline --no-audit --no-fund",
+  "npm --prefix worker test",
+  "npm --prefix worker run typecheck",
+]
+EOF
```
tend-agent, line 26:
```suggestion
EOF

# The list above has to match .config/wt.toml exactly, and the project key has
# to resolve — Worktrunk blocks an unapproved command in an unattended session,
# so fail here rather than in the first Cloud task.
[ "$(wt config approvals list --format=json | jq -r '.state')" = approved ] || {
  echo "approvals do not cover .config/wt.toml — update the list above" >&2
  wt config approvals list >&2
  exit 1
}
```

.pre-commit-config.yaml
```diff
@@ -29,16 +29,16 @@ repos:
     hooks:
       - id: actionlint
   # The composite actions' shared step bodies live in standalone scripts
-  # (shared/steps/, proxy/, generator/src/tend/templates/) rather than inline
-  # `run:` blocks, so actionlint's built-in shellcheck doesn't see them. Run
+  # (shared/steps/, proxy/, generator/src/tend/templates/, .config/codex-cloud/)
+  # rather than inline `run:` blocks, so actionlint's built-in shellcheck
   # shellcheck directly to keep that coverage. -S warning matches actionlint's
   # default severity.
```
```suggestion
  # rather than inline `run:` blocks, so actionlint's built-in shellcheck
  # doesn't see them. Run shellcheck directly to keep that coverage. -S warning
  # matches actionlint's default severity.
```

tend-agent (commented):
The `AGENTS.md` symlink makes an existing symlink-unsafe path in `codex/action.yaml` reachable. Its `Pin instruction files to base branch (fork PRs)` step iterates `find . -name AGENTS.md` and does `cp "/tmp/base-agents.md" "$f"` — `cp` follows a symlink destination, so on a fork PR under the Codex harness the root `AGENTS.md -> CLAUDE.md` link makes that write land in `CLAUDE.md`, replacing the guidance the step pinned two lines earlier with the literal string `CLAUDE.md`. The step exists to preserve the instruction file against fork-controlled content, and against this layout it destroys it instead.

The destination is fork-controlled too: a fork can repoint `AGENTS.md` at any path and the step writes base-tree content there — including `$HOME/.codex/AGENTS.md`, which the earlier `Stage AGENTS.md` step just wrote, so a fork can strip tend's staged system prompt before the session starts. Same shape at `cp /tmp/base-claude.md CLAUDE.md` a few lines up (a fork can make `CLAUDE.md` the symlink), and at the restore loop's `git show ... > "$rel"`, which fires on a dangling link because `[ ! -f "$rel" ]` is true for one.

Tend's own workflows run the Claude harness today, so nothing is live here — but `AGENTS.md -> CLAUDE.md` is a common adopter layout, so the action's exposure isn't specific to this repo. The fix is `rm -f` on each destination before writing (or `cp --remove-destination`) in all three places. Happy to push that here, or open it separately against `codex/action.yaml` if you'd rather keep this PR to the Cloud setup.

Two smaller things, both inline:

- The README's "keep its `approved-commands` list identical to the hooks and aliases in `.config/wt.toml`" rule has no enforcement, and both ways it can break are silent — a drifted list, or `.project.identifier` coming back absent so the heredoc emits `[projects.null]` and approves nothing under the real project key. Either way the container looks set up and the first unattended `wt switch` fails on approval instead. The rebase onto #977 already forced one hand-edit of that list, which is the drift case arriving on day one. Worktrunk exposes the check directly — `wt config approvals list --format=json`, whose `state` its own docs describe as the field a caller branches on before an unattended run.
- The rebase dropped `doesn't see them. Run` from the shellcheck comment, leaving `...built-in shellcheck / shellcheck directly to keep that coverage`.

Unrelated to correctness: `WORKTRUNK_VERSION=0.71.0` is behind the current `v0.73.0`, and isn't in `running-tend`'s weekly pin-bump table, so it'll only move when someone notices.

.config/codex-cloud/environment.sh
max-sixty: Added the structured approval-state check and validated the full setup in an isolated home: Worktrunk 0.73.0 reported all eight `.config/wt.toml` commands approved.

.pre-commit-config.yaml
max-sixty: Restored the missing sentence and kept the Cloud setup script in the shellcheck matcher.

tend-agent:
All three land — thanks. The `git restore` route does close the escape, and I checked one case beyond the new test: a fork that replaces a base `AGENTS.md`'s *parent directory* with a symlink pointing outside the checkout. git replaces the symlink with a real directory and writes the trusted content inside it, so the outside file is untouched there too.

One small pre-existing divergence I noticed while reading the new script, not a blocker: it pins to `repository.default_branch`, while the Claude harness's `restore-sensitive-config.sh` uses the PR's own `pull_request.base.ref`. On a PR whose base isn't the default branch, the Codex step restores the default branch's instruction files and deletes any `AGENTS.md` the base branch carries but the default doesn't. Both refs are base-repo controlled, so it's a question of which trusted tree wins rather than a hole — it just predates the move into a script, and now there's a test file to pin it in.

Workspace at 593810fd66ad341025e8bbe8019838474cf322dd.

`uv run pytest generator/tests/test_shared_steps.py -q` — 67 passed, including `test_pin_instruction_files_does_not_follow_fork_symlinks`.

The leading-directory variant, on git 2.54.0: base tree has `CLAUDE.md`, `AGENTS.md -> CLAUDE.md`, `nested/AGENTS.md`; the fork commit does `git rm -r nested` then `ln -s /outside nested`, where `/outside/AGENTS.md` holds sentinel content.

```
Pinned CLAUDE.md to main
Pinned AGENTS.md to main
Pinned nested/AGENTS.md to main
--- outside file content:
must not change
--- nested is:
drwxrwxr-x 2 ... nested
trusted nested
```

`find -P` never descends the symlink, so the removal loop skips it; the restore loop's `mkdir -p` hits `EEXIST`, and git's own checkout path unlinks the symlink and creates a real directory rather than writing through it.

Refs: [`pin-instruction-files.sh#L9`](https://github.com/max-sixty/tend/blob/593810fd66ad341025e8bbe8019838474cf322dd/shared/steps/pin-instruction-files.sh#L9) vs [`restore-sensitive-config.sh#L31`](https://github.com/max-sixty/tend/blob/593810fd66ad341025e8bbe8019838474cf322dd/shared/steps/restore-sensitive-config.sh#L31).

</details>

shared/steps/pin-instruction-files.sh
```diff
@@ -0,0 +1,41 @@
...
+      echo "No ${rel} on ${DEFAULT_BRANCH} — removed fork version"
+    fi
+  done
+
+# Restore every base-branch AGENTS.md, including files deleted by the fork.
+while IFS= read -r -d '' rel; do
+  case "$rel" in
+    AGENTS.md | */AGENTS.md)
+      mkdir -p -- "$(dirname "$rel")"
+      git restore --source="$BASE_REF" --worktree -- "$rel"
```
tend-agent: `git restore` creates missing leading directories itself, so this line adds no capability — only a failure mode. `mkdir -p` exits 1 with `File exists` when the path already exists as something that isn't a directory, and under `set -e` that fails the step and reds the job.

A fork PR reaches it by committing a regular file (or a dangling symlink) where the base branch has a directory containing an `AGENTS.md`: base has `docs/AGENTS.md`, fork replaces `docs` with a file. `git restore` handles that on its own — it removes the file and recreates the directory.

Verified on git 2.54 with `docs` as a regular file: `mkdir -p -- docs` exits 1, while `git restore --source=origin/main --worktree -- docs/AGENTS.md` restores the file cleanly. Same check with `docs` a symlink to a directory outside the repo: git replaced it with a real directory and left the outside target untouched.

```suggestion
      git restore --source="$BASE_REF" --worktree -- "$rel"
```

tend-agent (commented):
The symlink fix holds, and I checked the mechanism rather than taking `git restore` on trust: on git 2.54 it unlinks the destination instead of writing through a fork-controlled symlink, and it replaces a symlinked *leading directory* with a real one, so neither the file case nor the directory case escapes the checkout. Dropping the `|| true` on the restore loop makes it fail-closed too. The Cloud script's `@json` project key, the `state == approved` gate, and the 0.73.0 pin all check out against Worktrunk's own `JsonApprovals`/`Approvals` types — `state` is a top-level string, and both sides of the approval comparison run through `normalize_template_vars`, so the hand-written templates match. The `approved-commands` list is currently identical to `.config/wt.toml`'s four hooks and four alias commands.

One new finding and two smaller ones, all inline and all in the restore path. The substantive one: `mkdir -p` is the only line in that loop a fork can make fail, and it isn't doing anything `git restore` doesn't already do itself.

Separately — not part of this diff, and no action needed: I checked whether the new root `AGENTS.md` symlink opens the same vector under the Claude harness that tend actually runs. It doesn't. Claude Code reads `CLAUDE.md`, not `AGENTS.md` ([docs](https://code.claude.com/docs/en/memory#agents-md)), and `restore-sensitive-config.sh` already pins `CLAUDE.md` by `rm -rf` + `git checkout`, which is symlink-safe by the same argument as above.

shared/steps/pin-instruction-files.sh
```diff
@@ -0,0 +1,41 @@
...
+  rm -f -- CLAUDE.md
+  echo "No CLAUDE.md on ${DEFAULT_BRANCH} — removed fork version"
+fi
+
+# Remove AGENTS.md files introduced by the fork.
+find . -name AGENTS.md -not -path './.git/*' -print0 |
+  while IFS= read -r -d '' path; do
+    rel="${path#./}"
+    if ! git cat-file -e "${BASE_REF}:${rel}" 2>/dev/null; then
+      rm -f -- "$path"
```
tend-agent: Same shape as the `mkdir -p` below: `rm -f` fails on a directory, and `find -name AGENTS.md` matches one. A fork that commits `AGENTS.md/` as a directory — on a repo whose base branch has no `AGENTS.md` at that path — fails this step through `pipefail`. Adding `-r` covers it, and the missing `-type` filter has to stay missing since the loop must also match symlinks.

```suggestion
      rm -rf -- "$path"
```

shared/steps/pin-instruction-files.sh
```diff
@@ -0,0 +1,41 @@
...
+# Restore every base-branch AGENTS.md, including files deleted by the fork.
+while IFS= read -r -d '' rel; do
+  case "$rel" in
+    AGENTS.md | */AGENTS.md)
+      mkdir -p -- "$(dirname "$rel")"
+      git restore --source="$BASE_REF" --worktree -- "$rel"
+      echo "Pinned ${rel} to ${DEFAULT_BRANCH}"
+      ;;
+  esac
+done < <(git -c core.quotepath=off ls-tree -rz --name-only "$BASE_REF")
```
tend-agent: `ls-tree -z` doesn't quote paths at all, so `core.quotepath=off` is a no-op — a leftover from the `grep` pipeline this replaced.

```suggestion
done < <(git ls-tree -rz --name-only "$BASE_REF")
```

shared/steps/pin-instruction-files.sh
max-sixty: Applied this and expanded the integration test: it now covers a fork replacing a trusted parent directory with a file and creating a directory named `AGENTS.md`, while confirming all outside targets remain unchanged.

shared/steps/pin-instruction-files.sh
max-sixty: Applied this and expanded the integration test: it now covers a fork replacing a trusted parent directory with a file and creating a directory named `AGENTS.md`, while confirming all outside targets remain unchanged.

shared/steps/pin-instruction-files.sh
max-sixty: Applied this and expanded the integration test: it now covers a fork replacing a trusted parent directory with a file and creating a directory named `AGENTS.md`, while confirming all outside targets remain unchanged.

tend-agent:
All three are in at `1e83cbd`. I checked the expanded test actually pins the behaviour rather than passing either way: reverting `rm -rf` back to `rm -f` fails the run with `rm: cannot remove './directory/AGENTS.md': Is a directory`, and putting the `mkdir -p` back fails with `mkdir: cannot create directory 'nested': File exists` — so each new fixture case maps to one of the two fixes. The `core.quotepath=off` removal is behaviour-neutral by construction, so nothing covers it, as expected. `uv run pytest generator/tests/test_shared_steps.py` is green (67 passed).

shared/steps/pin-instruction-files.sh
```diff
@@ -0,0 +1,40 @@
...
+DEFAULT_BRANCH=$(jq -r '.repository.default_branch' "$GITHUB_EVENT_PATH")
+BASE_REF="origin/${DEFAULT_BRANCH}"
+
+# Git replaces the path without following a fork-controlled symlink and
+# preserves a trusted base-branch symlink as a symlink.
+if git cat-file -e "${BASE_REF}:CLAUDE.md" 2>/dev/null; then
+  git restore --source="$BASE_REF" --worktree -- CLAUDE.md
+  echo "Pinned CLAUDE.md to ${DEFAULT_BRANCH}"
+else
+  rm -f -- CLAUDE.md
```
tend-agent: Same directory case as the loop below, which this commit fixed. `git cat-file -e` fails when the base branch has no `CLAUDE.md` (common for an adopter that ships only `AGENTS.md`), and if the fork committed `CLAUDE.md` as a directory, `rm -f` exits 1 — `set -e` then fails the step.

Verified on git 2.54 with a base tree carrying no `CLAUDE.md` and a fork commit adding `CLAUDE.md/child`: `rm: cannot remove 'CLAUDE.md': Is a directory`, exit 1. `-r` is safe for the same reason it is below — `rm -rf` on a symlink unlinks the link, never its target.

```suggestion
  rm -rf -- CLAUDE.md
```

tend-agent (commented):
The three applied fixes hold, and I re-derived the mechanism rather than trusting the commit message: on git 2.54 `git restore` replaces a fork-committed directory at `AGENTS.md` with the base symlink (children and all), replaces a file or an outside-pointing symlink standing in for a trusted parent directory with a real directory, and writes through none of them. `rm -rf` on a symlink unlinks the link, not its target. `ls-tree -z` never quotes, so dropping `core.quotepath=off` is a no-op as expected. `uv run pytest generator/tests/test_shared_steps.py -q` — 67 passed at 1e83cbd.

One new finding, inline: the `rm -f` on line 18 is the same directory case you just fixed in the loop below, and it was left behind. It fails closed rather than open — the step exits non-zero and the session never starts — so it's a fork-triggerable nuisance, not a hole. Worth folding the case into `test_pin_instruction_files_does_not_follow_fork_symlinks` too, since the test's base tree currently always carries a `CLAUDE.md` and so never reaches that branch.

Base tree with no `CLAUDE.md`, fork commits `CLAUDE.md/` as a directory:

```
rm: cannot remove 'CLAUDE.md': Is a directory
EXIT=1
```

I also probed the one thing `rm -rf` inside a `find | while` pipeline could plausibly break — `find` erroring out after the consumer deletes a directory it just printed, which `pipefail` would turn into a failed step. It doesn't: GNU find holds the directory fd across the print, so even with `stdbuf -o0` forcing an unbuffered stream and a 2000-entry `AGENTS.md/` directory, the pipeline exits 0 with empty stderr. No change needed there.

Cases I ran clean against the current script, all fork-side: `AGENTS.md` directory replacing the base symlink (restored, `AGENTS.md/child` gone); dangling symlinks at both a base-tracked `docs/AGENTS.md` and an untracked root `AGENTS.md`; `CLAUDE.md` directory where the base *does* have `CLAUDE.md`.

</details>

shared/steps/pin-instruction-files.sh
max-sixty: Applied and added a base-absent integration case where the fork supplies `CLAUDE.md/` as a directory. The extracted step removes it and exits successfully.

tend-agent: Confirmed at `2a0476d` — and the new case pins the fix rather than passing either way: reverting to `rm -f` fails it with `rm: cannot remove 'CLAUDE.md': Is a directory`. `uv run pytest generator/tests/test_shared_steps.py` is green (68 passed).

I also checked the safety claim behind the `-r` I asked for, since this is the one destructive change in the step: with base carrying no `CLAUDE.md` and the fork committing `CLAUDE.md` as a symlink to a directory outside the checkout, the step exits 0, the symlink is unlinked, and the outside tree is untouched — `rm -rf` on an unslashed symlink argument never descends it.

git 2.54.0. Base tree with only `README.md`; `origin/main` at that commit; fork worktree adds `CLAUDE.md -> $T/outside`, where `$T/outside/sub/secret.txt` holds sentinel content.

```
No CLAUDE.md on main — removed fork version
exit=0
--- CLAUDE.md present in repo: no (symlink: no)
--- outside dir:
sub
must not change
```

</details>

tend-agent (approved):

PR #979, opened 2026-08-14
feat: 👀 marks a session in flight, and comes off when it ends

max-sixty:
👀 meant two different things. On a comment that mentions the bot it tracked a session in flight and came off when the session ended; on an opened issue or PR it went on and stayed forever. Both now mean the same thing — a session is working on this right now — and the reaction is released at the end of the run under `always()`, so a failed or cancelled session doesn't strand it.

Review's reaction no longer narrows to `opened`. That narrowing made sense while the reaction was permanent (a push lands on a PR that already carries it), but once the eyes come off, a push mid-review boots a session with no reaction at all. Every run that clears the gate takes them.

The three workflows now share one `react_eyes`/`unreact_eyes` pair, parameterized by the reaction target — `issues/<n>` for an issue or PR, `issues/comments/<id>` or `pulls/comments/<id>` for a comment. Mention's try-issues-then-pulls fallback goes with it: `github.event_name` says which endpoint the comment lives under, so the target is computed rather than probed. That also closes a latent mis-target, since the probe would have accepted an unrelated issue comment that happened to share an id with the review comment.

<details>

Both API calls were run against `tend-agent/tend-integration` with the bot's own token: the POST, the filtered `?content=eyes` lookup, and the DELETE all returned as expected, and the probe reaction was removed afterwards. `actionlint` (which runs shellcheck over the `run:` blocks) is clean on freshly generated workflows for every trigger — the committed `tend-*.yaml` in this repo still track the released generator and don't carry these steps yet.

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,9 @@ file](docs/tend.example.yaml) and a repo-local `/running-tend` skill.
 | **notifications** | Every 15 minutes           | Polls GitHub notifications, responds to unhandled mentions, marks handled threads as read.                                                                  |
 | **review-runs**   | Daily                      | Reviews recent CI runs for behavioral problems and proposes skill/config improvements.                                                                      |
 
-The bot reacts 👀 when a job picks an event up. An opened issue or PR keeps
-the reaction; on a comment that mentions the bot, it comes off when the
-session ends.
+The bot reacts 👀 while a session is working: on an issue or PR when it opens,
+on a comment that mentions the bot. The reaction comes off when the session
+ends.
 
 Scheduled workflows also support manual dispatch for testing. GitHub runs
 `schedule` triggers on a best-effort basis and drops ticks under load, so
--- a/README.md
+++ b/README.md
@@ -80,9 +80,9 @@ file](docs/tend.example.yaml) and a repo-local `/running-tend` skill.
 | **notifications** | Every 15 minutes           | Polls GitHub notifications, responds to unhandled mentions, marks handled threads as read.                                                                  |
 | **review-runs**   | Daily                      | Reviews recent CI runs for behavioral problems and proposes skill/config improvements.                                                                      |
 
-The bot reacts 👀 while a session is working: on an issue or PR when it opens,
-on a comment that mentions the bot. The reaction comes off when the session
-ends.
+The bot reacts 👀 while a session is working: on an issue when it opens, on a
+PR whenever a review starts, and on a comment that mentions the bot. The
+reaction comes off when the session ends.
 
 Scheduled workflows also support manual dispatch for testing. GitHub runs
 `schedule` triggers on a best-effort basis and drops ticks under load, so
--- a/generator/tests/test_generate.py
+++ b/generator/tests/test_generate.py
@@ -90,6 +90,11 @@ def test_setup_steps_rendered(tmp_path: Path) -> None:
         )
 
 
+def _eyes_steps(steps: list[dict[str, object]]) -> list[dict[str, object]]:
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -80,9 +80,9 @@ file](docs/tend.example.yaml) and a repo-local `/running-tend` skill.
 | **notifications** | Every 15 minutes           | Polls GitHub notifications, responds to unhandled mentions, marks handled threads as read.                                                                  |
 | **review-runs**   | Daily                      | Reviews recent CI runs for behavioral problems and proposes skill/config improvements.                                                                      |
 
-The bot reacts 👀 when a job picks an event up. An opened issue or PR keeps
-the reaction; on a comment that mentions the bot, it comes off when the
-session ends.
+The bot reacts 👀 while a session is working: on an issue or PR when it opens,
+on a comment that mentions the bot. The reaction comes off when the session
+ends.
```
tend-agent: This line still describes the narrowing the PR removes. Review's react step is now gated on `skip_condition` alone, so a `synchronize`, `reopened` or `ready_for_review` that clears the gate takes the eyes too — "when it opens" is only true of triage's `issues: [opened]` now.

```suggestion
The bot reacts 👀 while a session is working: on an issue when it opens, on a
PR whenever a review starts, and on a comment that mentions the bot. The
reaction comes off when the session ends.
```

generator/tests/test_generate.py
```diff
@@ -703,40 +708,69 @@ def test_review_queues_pushes_behind_a_gate(tmp_path: Path) -> None:
...
-        # A step may narrow further (the eyes reaction fires only on `opened`),
-        # but the gate's verdict stays a required conjunct.
+        # A step may narrow further (the eyes reaction fires only on `opened`)
+        # or widen to `always()` (the reaction comes off a failed run too), but
+        # the gate's verdict stays a conjunct of whatever it builds.
         condition = step.get("if", "")
-        assert condition == gate or condition.startswith(f"{gate} && "), (
-            f"ungated step after the gate: {step}"
-        )
+        assert gate in condition, f"ungated step after the gate: {step}"
```
tend-agent: `gate in condition` is a substring check, so it now passes for a step that isn't gated at all — `if: github.event.action == 'opened' || steps.gate.outputs.should_run == 'true'` satisfies it while letting a queued run boot an agent on an already-examined HEAD, which is exactly what this test exists to catch. The two shapes that actually appear are enumerable, so the structural check survives the `always()` widening:

```suggestion
        # A step may narrow further, or widen to `always()` (the reaction comes
        # off a failed run too), but the gate's verdict stays a conjunct of
        # whatever it builds.
        condition = step.get("if", "")
        assert condition in (gate, f"always() && ({gate})") or condition.startswith(
            f"{gate} && "
        ), f"ungated step after the gate: {step}"
```

(The comment's "the eyes reaction fires only on `opened`" is stale for the same reason as the README line, so the suggestion drops it.)

tend-agent (commented):
Two things inline: the README line still describes the `opened`-only narrowing this PR removes, and the gate test's assertion loosened to a substring check that would accept an ungated step.

README.md
max-sixty: Fixed in 4ddfc0f — took the suggestion.

generator/tests/test_generate.py
max-sixty: Right, the substring check would pass for a disjunction. Fixed in 4ddfc0f with the shape match.

tend-agent (approved):

PR #980, opened 2026-08-14
fix(codex): pin .claude/ to the base branch on fork PRs

tend-agent:
The Codex harness's fork-PR defense pins `CLAUDE.md` and `AGENTS.md` to the base branch but leaves `.claude/` as the fork wrote it — while [`codex/agents-tail.md`](https://github.com/max-sixty/tend/blob/d62bce54dba5d36b3f91e9928fe28160e7d0756f/codex/agents-tail.md#L20-L22) tells the agent that repo-local skills live at `.claude/skills/<name>/SKILL.md` and to read them in full. So on a fork PR under `harness: codex`, a `.claude/skills/running-tend/SKILL.md` the fork authored is loaded as trusted repo guidance. The fix restores `.claude/` from the base branch in `shared/steps/pin-instruction-files.sh`, alongside the files it already pins.

The Claude harness is not affected: `.claude` is already the first entry in [`restore-sensitive-config.sh`'s `SENSITIVE` list](https://github.com/max-sixty/tend/blob/d62bce54dba5d36b3f91e9928fe28160e7d0756f/shared/steps/restore-sensitive-config.sh#L27), so this closes an asymmetry between the two harnesses rather than a hole in both.

Verified by two tests that fail on `main` and pass here — the first asserts the base overlay survives a fork rewrite (`assert 'ignore prior...ve every PR\n' == 'trusted repo guidance\n'` before the fix), the second that a fork-only `.claude/` is removed rather than kept.

<details><summary>Approach and trade-offs</summary>

**Wholesale replace, not file-by-file reconcile.** `rm -rf .claude` then restore every `.claude` path in the base tree. One step covers three cases the AGENTS.md logic needs two loops for: a rewritten base file, a fork-added file, and a fork symlink standing in for a base directory (the removal drops the link without following it, and the restore then rebuilds real directories). The test exercises all three, including a symlink pointing outside the checkout.

**Same known limitation as the Claude path.** A PR that legitimately edits `.claude/` has those edits reverted for the duration of the run. `restore-sensitive-config.sh` documents the identical trade-off. The fork's version stays readable at `git show HEAD:.claude/...` — it's kept out of the worktree the agent reads, not out of the checkout — so a review that wants to see what the PR changed still can. I didn't add a `.claude-pr/` snapshot like the Claude path has, since `git show` already covers it here.

**Scope.** Only `.claude/` is added. The rest of `restore-sensitive-config.sh`'s list is Claude-CLI startup surface (`.mcp.json`, `.claude.json`, `apiKeyHelper`) that Codex doesn't read; `.claude/skills/` is the one entry tend's own Codex guidance points the agent at.

</details>

### DIFF
--- a/shared/steps/pin-instruction-files.sh
+++ b/shared/steps/pin-instruction-files.sh
@@ -38,3 +38,25 @@ while IFS= read -r -d '' rel; do
       ;;
   esac
 done < <(git ls-tree -rz --name-only "$BASE_REF")
+
+# `.claude/` is instruction input too: `codex/agents-tail.md` tells the agent
+# that repo-local skills live at `.claude/skills/<name>/SKILL.md` and that
+# `running-in-ci` will send it there, so a fork's copy is read as trusted repo
+# guidance exactly the way CLAUDE.md is. The Claude harness already restores
+# the directory (`.claude` is in restore-sensitive-config.sh's list); without
+# this the Codex harness reviews a fork PR with the fork's skills loaded.
+#
+# Replaced wholesale rather than reconciled file-by-file: the removal drops
+# fork-added skills and any fork symlink standing in for a base directory in
+# one step, and the restore then rebuilds the base tree as real files. The
+# fork's version stays readable at `git show HEAD:<path>` for a review that
+# wants to see what the PR changed — it is only kept out of the worktree the
+# agent reads.
+mapfile -t -d '' BASE_CLAUDE_FILES < <(git ls-tree -rz --name-only "$BASE_REF" -- .claude)
+rm -rf -- .claude
+if [ "${#BASE_CLAUDE_FILES[@]}" -gt 0 ]; then
+  git restore --source="$BASE_REF" --worktree -- "${BASE_CLAUDE_FILES[@]}"
+  echo "Pinned .claude/ to ${DEFAULT_BRANCH} (${#BASE_CLAUDE_FILES[@]} files)"
+else
+  echo "No .claude/ on ${DEFAULT_BRANCH} — removed fork version"
+fi
--- a/shared/steps/pin-instruction-files.sh
+++ b/shared/steps/pin-instruction-files.sh
@@ -52,11 +52,10 @@ done < <(git ls-tree -rz --name-only "$BASE_REF")
 # fork's version stays readable at `git show HEAD:<path>` for a review that
 # wants to see what the PR changed — it is only kept out of the worktree the
 # agent reads.
-mapfile -t -d '' BASE_CLAUDE_FILES < <(git ls-tree -rz --name-only "$BASE_REF" -- .claude)
 rm -rf -- .claude
-if [ "${#BASE_CLAUDE_FILES[@]}" -gt 0 ]; then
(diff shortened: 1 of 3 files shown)


shared/steps/pin-instruction-files.sh
```diff
@@ -38,3 +38,25 @@ while IFS= read -r -d '' rel; do
...
+# wants to see what the PR changed — it is only kept out of the worktree the
+# agent reads.
+mapfile -t -d '' BASE_CLAUDE_FILES < <(git ls-tree -rz --name-only "$BASE_REF" -- .claude)
+rm -rf -- .claude
+if [ "${#BASE_CLAUDE_FILES[@]}" -gt 0 ]; then
+  git restore --source="$BASE_REF" --worktree -- "${BASE_CLAUDE_FILES[@]}"
+  echo "Pinned .claude/ to ${DEFAULT_BRANCH} (${#BASE_CLAUDE_FILES[@]} files)"
+else
+  echo "No .claude/ on ${DEFAULT_BRANCH} — removed fork version"
+fi
```
tend-agent: The file list isn't load-bearing: `git restore` takes `.claude` as a directory pathspec and rebuilds the whole base subtree even though the worktree path was just removed, and `git cat-file -e` resolves a tree as readily as a blob. That collapses this to the same shape the `CLAUDE.md` block above uses, and removes the (admittedly remote) `ARG_MAX` ceiling on a large base `.claude/`. The wholesale-replace comment above still describes it accurately.

```suggestion
rm -rf -- .claude
if git cat-file -e "${BASE_REF}:.claude" 2>/dev/null; then
  git restore --source="$BASE_REF" --worktree -- .claude
  echo "Pinned .claude/ to ${DEFAULT_BRANCH}"
else
  echo "No .claude/ on ${DEFAULT_BRANCH} — removed fork version"
fi
```

tend-agent (commented):
The hole is real and the fix closes it: `codex/agents-tail.md` points the agent at `.claude/skills/<name>/SKILL.md` and tells it to read those in full, so under `harness: codex` a fork's overlay was being loaded as trusted repo guidance while `CLAUDE.md` and `AGENTS.md` were pinned. Wholesale replace is the right shape — I checked the three cases the comment claims (rewritten base file, fork-added file, fork symlink standing in for a base directory) and `rm -rf -- .claude` drops a symlink without following it, so the escape case holds. No collision with the plugin install either: `codex plugin add` resolves its root from `github.action_path/..`, not the repo tree, so removing the repo's `.claude/` can't strip the plugin's skills.

One simplification inline. The `mapfile` + array-expansion form is a second idiom for the job this script already does 40 lines above for `CLAUDE.md`, and the file list isn't needed — `git restore --source=<ref> --worktree -- .claude` accepts the directory pathspec and rebuilds the whole subtree even when the worktree path is absent, and `git cat-file -e "<ref>:.claude"` resolves a tree the same way it resolves a blob. That also drops the `ARG_MAX` exposure a base `.claude/` with a few thousand files would create. Per CLAUDE.md's "Simplicity outranks efficiency" / "Prefer deleting a mechanism over refining it".

I verified the replacement against both new tests plus the existing suite (70 passed) and shellcheck, so it's behaviour-identical; I'll push it rather than leave the click for you, since the authoring session is gone.

PR #981, opened 2026-08-14
docs(running-tend): sweep every pin, not a hand-kept table

max-sixty:
The weekly bump section was a table of four action inputs, so every other pin in the repo drifted unswept — a pin entered scope only when someone remembered to add a row for it, and `WORKTRUNK_VERSION` arrived exactly that way in #978. What the closed table was hiding, measured on this tree: ruff `v0.14.11` → `v0.16.3`, uv-pre-commit `0.7.12` → `0.12.4`, typos `v1.42.0` → `v1.49.0`, vitest 2 → 4, typescript 5 → 7, wrangler 4.90 → 4.123, astro 5 → 7, plus the `uv_build` range that has excluded the local uv since 0.12.

Each ecosystem now enumerates its own pins, so the next one added shows up without an edit to this file:

```bash
yq -r '.inputs | to_entries[] | select(.key | test("_version$")) | ...' */action.yaml
git grep -nE '(==|~=|<=?)[0-9]' -- '*pyproject.toml'   # freeze-pins; floors like click>=8.0 stay
uv lock --upgrade --dry-run
uvx pre-commit autoupdate
npm --prefix worker outdated; npm --prefix site outdated
git grep -nE '^[A-Za-z_]*VERSION=' -- '*.sh'
```

The table survives only for pins whose rule isn't "move to latest and let CI decide" — `claude_version` tracking `latest` over `stable`, mitmproxy dragging the workspace `==` pin with it — and, after the review, for every pin CI can't verify at all: `uv_build`'s stale range only *warns* during `uv build`, `publish-site.yaml` builds `site/` on push to main rather than on the PR, and nothing anywhere runs `.config/codex-cloud/environment.sh`, which dies under `set -euo pipefail` in a maintainer's Codex Cloud setup rather than in a job. Those three were carved out in three different places in the first push; they're one table now, so the default rule means what it says. The PR-splitting rule that used to sit under `uses:` refs (ships-to-adopters / ours-alone) now governs all pins, with majors getting their own PR from either bucket.

`WORKTRUNK_VERSION` is the test of whether the frame works: #978's row and `sed` lookup are gone, and the shell-script sweep finds the pin on its own. It keeps a row, but for what CI can't check rather than for existing.

Runner images, `node-version` and `requires-python` are named as out of scope rather than silently omitted — platform choices carrying their own rationale, moved when a reason arrives.

<details>

Every command run against this tree before it went in:

- the `yq` sweep prints all four action inputs with their file
- the Python grep returns exactly the two freeze-pins (`mitmproxy==`, `uv_build`'s cap) and no floors or `requires-python`
- `uvx pre-commit autoupdate` works with only uv installed (the weekly job's whole toolchain) — run against a scratch copy of the config, which is where the three rev numbers above come from
- `npm outdated` exits 1 with rows, which is why the comment says so
- the shell-script grep returns `WORKTRUNK_VERSION` and nothing else
- `uv build --package tend` warns on the stale `uv_build` range and still succeeds, so the "only this sweep catches it" claim is the observed behavior
- `pre-commit run --all-files` passes

</details>

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -182,8 +182,6 @@ uvx pre-commit autoupdate
 
 # npm: `Wanted` ≠ `Current` is lockfile drift (`npm update`); `Latest` ≠
 # `Wanted` needs the range in package.json moved. Exits 1 when a row prints.
-# `publish-site.yaml` builds site/ only on push to main, so run
-# `npm --prefix site run build` yourself before landing an astro bump.
 npm --prefix worker outdated
 npm --prefix site outdated
 
@@ -206,9 +204,10 @@ Out of scope entirely: runner images (`ubuntu-24.04`), `node-version`, and
 `requires-python` are platform choices carrying their own rationale, so they
 move when a reason arrives rather than on a cadence.
 
-Default rule: move to latest and let CI decide. Split PRs by who runs the
-result, and take what fits in one session rather than clearing a backlog at
-once — an unswept pin waits a week, a swamped run finishes nothing.
+Default rule: move to latest and let CI decide — the table below names the pins
+where CI can't. Split PRs by who runs the result, and take what fits in one
+session rather than clearing a backlog at once — an unswept pin waits a week, a
+swamped run finishes nothing.
 
 - **Ships to adopters** — `claude/action.yaml` and `codex/action.yaml` run in
   every adopter's job from the next release; `generator/src/tend/templates/`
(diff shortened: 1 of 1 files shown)


.claude/skills/running-tend/SKILL.md
```diff
@@ -161,33 +161,77 @@ the whole list.
...
+
+### Pins with rules of their own
+
+| Pin | File | Rule |
+|---|---|---|
+| `claude_version` | `claude/action.yaml` | npm's `latest` dist-tag, not `stable` |
+| `mitmproxy_version` | `claude/action.yaml` | move the root `pyproject.toml` `==` pin with it and `uv lock` |
+| `uv_version` | `claude/action.yaml` | move it with `mitmproxy_version` |
+| `codex_version` | `codex/action.yaml` | `latest`; `alpha` only for a fix not yet released |
+| `uv_build` | `generator/pyproject.toml` | its range must contain the uv doing the build; a stale one only warns during `uv build`, so only this sweep catches it |
```
tend-agent, line 232:
```suggestion
| `uv_build` | `generator/pyproject.toml` | its range must contain the uv doing the build; a stale one only warns during `uv build`, so only this sweep catches it |
| `WORKTRUNK_VERSION` | `.config/codex-cloud/environment.sh` | latest GitHub release; no CI job runs this script, so read the release's installer asset name and the `wt` calls below the pin before landing |
```

tend-agent (commented):
`WORKTRUNK_VERSION` loses its rule row and lands under the new default, "move to latest and let CI decide" — but nothing in CI runs `.config/codex-cloud/environment.sh`. `git grep codex-cloud -- .github/` is empty, and pre-commit's shellcheck hook only parses the file. It's the third pin whose bump CI can't verify, alongside `uv_build`'s warn-only range and the astro build the npm comment already carves out, so it belongs in the table rather than under the default rule.

Concretely: the sweep's shell grep surfaces `WORKTRUNK_VERSION=0.73.0`, the default rule says move it to latest, CI goes green — and the maintainer's next Codex Cloud setup dies under `set -euo pipefail` if the release's installer asset name moved or `wt config show --format json`'s `.project.identifier` changed shape. The script consumes both immediately after installing, so there's no signal before someone opens an environment.

Everything else holds against this tree — all four enumerating commands return what the PR body claims, `click>=8.0` in `generator/pyproject.toml` grounds the floors comment, and nothing outside this file referenced the removed `### Tool versions` heading.

tend-agent (approved):

PR #985, opened 2026-08-14
docs(running-in-ci): fork PR reviews reach no successor session

### LINKED ISSUE #984: running-in-ci: a bot review on a fork PR reaches no session, so its findings strand
On a **fork** PR, a review the bot writes reaches no session at all — and where the bot is the de-facto author (a maintainer directed it to push the changes), those findings are its own and nothing owns them. It happened twice on one PR in one 24h window on `max-sixty/worktrunk`; both times the work was recovered only because an unrelated human comment happened to wake a session hours later.

## What happened

[max-sixty/worktrunk#3744](https://github.com/max-sixty/worktrunk/pull/3744) is a fork PR (`head.repo.full_name` = `dataders/worktrunk`). At 16:03Z the maintainer wrote *"let's make the changes directly and then `dataders` can revert if he disagrees"*, so from then on the bot authored every commit on that branch. `tend-review` then reviewed its own pushes and left inline findings twice:

| Bot review | Findings | Picked up | By what |
|---|---|---|---|
| [4929272716](https://github.com/max-sixty/worktrunk/pull/3744#pullrequestreview-4929272716) 16:22:47Z | 4 points | 20:15:44Z (**+3h53m**) | a session woken by `dataders`' unrelated thank-you comment |
| [4931285490](https://github.com/max-sixty/worktrunk/pull/3744#pullrequestreview-4931285490) 20:24:19Z | 2 points | 23:53:39Z (**+3h29m**) | a session woken by the maintainer asking *"ok but I see new comments `worktrunk-bot` ?"* |

The first recovery says out loud that it was luck. From that session's log ([run 31739132560](https://github.com/max-sixty/worktrunk/actions/runs/31739132560)):

> The triggering comment ... was a thank-you from `dataders` with no ask in it — the p.s. is directed at `max-sixty` personally. I didn't reply to that in kind. What *was* unaddressed on the thread was my own second review's four findings, which I'd offered to push, under `max-sixty`'s standing directive on this PR.

Absent that comment, nothing was scheduled to look.

## Why nothing owns it

Three mechanisms line up, each individually reasonable:

1. **The relay never fires on a fork.** `mention.yaml`'s `relay` job is gated on `github.event.pull_request.head.repo.full_name == github.repository`, so a `pull_request_review` on a fork PR sends no dispatch. Both reviews above produced a fully-skipped run — [31720373387](https://github.com/max-sixty/worktrunk/actions/runs/31720373387) and [31740695864](https://github.com/max-sixty/worktrunk/actions/runs/31740695864), `relay`/`verify`/`handle` all `skipped`. A same-repo PR in the same window ([31754296488](https://github.com/max-sixty/worktrunk/actions/runs/31754296488)) relayed normally, which is the contrast. The `verify` gate from #916/#965 — *"a bot-authored review summons a session only as the reviewer-to-author handoff"* — never even gets to run here; the fork filter drops the event first.

2. **The documented fallback can't see it.** The comment above `relay`'s `if:` points at the notifications poll for what the filter excludes. `tend-notifications` ran 17 times inside these two gaps and touched neither review — GitHub doesn't notify an actor of their own activity, so the bot's own review is invisible to a notifications sweep by construction.

3. **The pushing session is told to leave it, and its re-fetch drops the field that would show it.** `running-in-ci` → *"A review that lands while you poll is not yours to action"* justifies itself with *"`tend-mention` is dispatched on `pull_request_review` for every PR the bot authored, and that dispatch runs whether or not you also respond"* — true for a same-repo bot PR, false here. And both sessions narrowed their pre-post re-fetch to comments only, dropping `reviews` from the projection that **Recheck Before Posting** specifies as `--json comments,reviews`:

   - [run 31718752567](https://github.com/max-sixty/worktrunk/actions/runs/31718752567) at 16:35:35Z: `gh pr view 3744 --json state,comments` — 13 minutes after its own review landed.
   - [run 31739132560](https://github.com/max-sixty/worktrunk/actions/runs/31739132560) at 20:25:21Z: `gh pr view 3744 --json comments,state` — 62 seconds after its own review landed. Its earlier re-fetch at 20:15:39Z *did* include `reviews`; the final one didn't.

   The narrowing is understandable from the section's framing: it reads as a duplicate-post check, and for that `comments` alone looks sufficient. The section's own follow-on — *"A new entry may be a directive, not a duplicate"* — is exactly the case that got dropped.

## Suggested fix

Guidance, not workflow — the fork filter is a security boundary and #916/#965's handoff rule is a deliberate call. Two small carve-outs in `running-in-ci`:

**In "A review that lands while you poll is not yours to action"**, note that the rule's premise fails on a fork PR: the relay's same-repo filter means no dispatch fires, so no successor session exists. When the session pushed the commits under a maintainer directive it is the de-facto author and should action its own review's findings before ending; otherwise it should name them in the closing comment as unowned, so the thread shows someone has to pick them up.

**In "Recheck Before Posting"**, say why `reviews` has to stay in the projection — a review submitted while the session worked is a directive, not a duplicate, and dropping the field to `comments` alone is what made both strandings invisible. The recipe is already right; the reason it's shaped that way isn't stated, which is what let two sessions narrow it.

## Evidence level

**High**, structural, 2 occurrences in one window (`review-runs` gates: structural + High needs 2–3). Neither is wasted compute — in both cases real findings sat unaddressed for over three hours on a contributor's PR, and the recovery path was a human noticing. The proposed fix is a few lines of skill text.

Filed from `max-sixty/worktrunk`'s daily `tend-review-runs` sweep ([run 31782844013](https://github.com/max-sixty/worktrunk/actions/runs/31782844013)); that repo is pinned at tend `0.1.15`.

tend-agent:
## Problem

Two sections of `running-in-ci` rest on a premise that fails on fork PRs. "A review that lands while you poll is not yours to action" justifies leaving a review alone with "`tend-mention` is dispatched on `pull_request_review` for every PR the bot authored" — but the relay job is gated on `github.event.pull_request.head.repo.full_name == github.repository` ([`mention.yaml.j2:92-95`](https://github.com/max-sixty/tend/blob/8a18d6c/generator/src/tend/templates/mention.yaml.j2#L92-L95)), so a review on a fork PR dispatches nothing. The fallback that gate's comment names — the notifications poll — can't see it either, since GitHub doesn't notify an actor of their own activity. And **Recheck Before Posting** specifies `--json comments,reviews` without saying why `reviews` has to stay, so sessions narrow it to `comments` and never see their own review land.

Reported in #984 with two occurrences on one fork PR in a 24h window: bot review findings sat unaddressed for 3h53m and 3h29m, recovered only because unrelated human comments woke later sessions.

## Solution

Two paragraphs of skill text, no workflow change — the fork filter is a security boundary, and widening it is moot anyway while a fork run's token is refused the dispatch POST (per the same comment block).

- In "A review that lands while you poll is not yours to action": state that the premise fails on a fork PR and nothing succeeds you, so a session that pushed the commits under a maintainer directive actions its own review's findings, and otherwise names them as unowned in its closing comment.
- In **Recheck Before Posting**: say why `reviews` stays in the projection — a review that landed while you worked is a directive, not a duplicate, and on a fork PR it's the entry nothing else picks up.

## Testing

`uv run pytest` passes (554 tests). No test covers skill prose, so there is no reproduction test to add here; the behavioral claims were verified against the relay gate in the mention template rather than reproduced.

---
Closes #984 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -252,7 +252,7 @@ If you cannot verify, say "I haven't confirmed whether these failures are pre-ex
 
 Poll your checks to terminal, do the follow-up you were gated on, and exit; name the outstanding review in your summary. This covers a review that arrives *while* you work — a session dispatched to answer a specific review owns that review and actions it normally.
 
-**On a fork PR the premise fails — nothing succeeds you.** `tend-mention`'s relay job is gated on `head.repo.full_name == github.repository`, so a review on a fork PR dispatches nothing, and the notifications poll named as that filter's fallback can't see it either: GitHub doesn't notify an actor of their own activity, so the bot's own review is invisible there by construction. Findings left for a successor session strand until a human happens to comment. So when the review is on a fork PR: if you pushed the commits under a maintainer directive you are the de-facto author — action your own review's findings before ending. Otherwise say in your closing comment that they are unaddressed and unowned, so the thread shows someone has to pick them up.
+**On a fork PR the premise fails — nothing succeeds you.** `tend-mention`'s relay job is gated on `head.repo.full_name == github.repository`, so a review on a fork PR dispatches nothing, and the notifications poll named as that filter's fallback can't see it either: GitHub doesn't notify an actor of their own activity, so the bot's own review is invisible there by construction. Findings left for a successor session strand until a human happens to comment. So if you pushed the commits under a maintainer directive you are the de-facto author — action your own review's findings before ending. If you pushed them without one, name them in your closing comment as unaddressed and unowned, so the thread shows someone has to pick them up. A review on commits the contributor pushed already reached them — leave it.
 
 ### Rerunning failed jobs
 
@@ -328,7 +328,7 @@ gh pr view <number> --json comments,reviews \
   --jq '{comments: (.comments | length), reviews: (.reviews | length)}'
 ```
 
-Keep `reviews` in the PR projection rather than narrowing to `comments` — a review submitted while you worked is a directive, not a duplicate, and it is the entry a dedup-shaped check is most likely to drop. On a fork PR it is also the one nothing else will pick up (see **A review that lands while you poll is not yours to action**).
+Keep `reviews` in the PR projection rather than narrowing to `comments` — it is the entry a dedup-shaped check is most likely to drop, and on a fork PR it is the one nothing else will pick up (see **A review that lands while you poll is not yours to action**).
 
 If any prior entry — from a human or another tend workflow — already addresses a point the response would make, omit that point. The dedup applies equally to comment bodies, review bodies, and inline replies. If the response is now entirely redundant, don't post it.
 
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -252,6 +252,8 @@ If you cannot verify, say "I haven't confirmed whether these failures are pre-ex
 
 Poll your checks to terminal, do the follow-up you were gated on, and exit; name the outstanding review in your summary. This covers a review that arrives *while* you work — a session dispatched to answer a specific review owns that review and actions it normally.
 
+**On a fork PR the premise fails — nothing succeeds you.** `tend-mention`'s relay job is gated on `head.repo.full_name == github.repository`, so a review on a fork PR dispatches nothing, and the notifications poll named as that filter's fallback can't see it either: GitHub doesn't notify an actor of their own activity, so the bot's own review is invisible there by construction. Findings left for a successor session strand until a human happens to comment. So when the review is on a fork PR: if you pushed the commits under a maintainer directive you are the de-facto author — action your own review's findings before ending. Otherwise say in your closing comment that they are unaddressed and unowned, so the thread shows someone has to pick them up.
+
 ### Rerunning failed jobs
 
 To rerun a run's failed jobs and wait for the outcome, use the bundled script — it reruns, finds the new attempt's jobs (the parent run's `.status` and the commit rollup stay pending on unrelated siblings, so neither is a usable signal), and polls them to terminal:
@@ -325,6 +327,8 @@ gh issue view <number> --json comments --jq '.comments | length'
 gh pr view <number> --json comments,reviews \
   --jq '{comments: (.comments | length), reviews: (.reviews | length)}'
 ```
+
+Keep `reviews` in the PR projection rather than narrowing to `comments` — a review submitted while you worked is a directive, not a duplicate, and it is the entry a dedup-shaped check is most likely to drop. On a fork PR it is also the one nothing else will pick up (see **A review that lands while you poll is not yours to action**).
 
 If any prior entry — from a human or another tend workflow — already addresses a point the response would make, omit that point. The dedup applies equally to comment bodies, review bodies, and inline replies. If the response is now entirely redundant, don't post it.
 


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -252,6 +252,8 @@ If you cannot verify, say "I haven't confirmed whether these failures are pre-ex
 
 Poll your checks to terminal, do the follow-up you were gated on, and exit; name the outstanding review in your summary. This covers a review that arrives *while* you work — a session dispatched to answer a specific review owns that review and actions it normally.
 
+**On a fork PR the premise fails — nothing succeeds you.** `tend-mention`'s relay job is gated on `head.repo.full_name == github.repository`, so a review on a fork PR dispatches nothing, and the notifications poll named as that filter's fallback can't see it either: GitHub doesn't notify an actor of their own activity, so the bot's own review is invisible there by construction. Findings left for a successor session strand until a human happens to comment. So when the review is on a fork PR: if you pushed the commits under a maintainer directive you are the de-facto author — action your own review's findings before ending. Otherwise say in your closing comment that they are unaddressed and unowned, so the thread shows someone has to pick them up.
```
tend-agent: Scope the `Otherwise` so a session can tell which case it is in — the alternative to "pushed under a directive" reads as either "pushed without one" or "didn't push", and only the first should produce a comment.

```suggestion
**On a fork PR the premise fails — nothing succeeds you.** `tend-mention`'s relay job is gated on `head.repo.full_name == github.repository`, so a review on a fork PR dispatches nothing, and the notifications poll named as that filter's fallback can't see it either: GitHub doesn't notify an actor of their own activity, so the bot's own review is invisible there by construction. Findings left for a successor session strand until a human happens to comment. So if you pushed the commits under a maintainer directive you are the de-facto author — action your own review's findings before ending. If you pushed them without one, name them in your closing comment as unaddressed and unowned, so the thread shows someone has to pick them up. A review on commits the contributor pushed already reached them — leave it.
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -326,6 +328,8 @@ gh pr view <number> --json comments,reviews \
   --jq '{comments: (.comments | length), reviews: (.reviews | length)}'
 ```
 
+Keep `reviews` in the PR projection rather than narrowing to `comments` — a review submitted while you worked is a directive, not a duplicate, and it is the entry a dedup-shaped check is most likely to drop. On a fork PR it is also the one nothing else will pick up (see **A review that lands while you poll is not yours to action**).
```
tend-agent: Drop the lead clause — **A new entry may be a directive, not a duplicate** makes that exact point three paragraphs down. What's new here is the projection rule and the fork-PR pointer.

```suggestion
Keep `reviews` in the PR projection rather than narrowing to `comments` — it is the entry a dedup-shaped check is most likely to drop, and on a fork PR it is the one nothing else will pick up (see **A review that lands while you poll is not yours to action**).
```

tend-agent (commented):
Both claims check out against the code: the relay `if:` in [`mention.yaml.j2`](https://github.com/max-sixty/tend/blob/8a18d6cc2bcd17fa1d4feee4785e7d1dea99639d/generator/src/tend/templates/mention.yaml.j2#L92-L95) does gate on `head.repo.full_name == github.repository`, and the notifications skill is driven entirely by `gh api notifications`, so the bot's own review is structurally invisible to it. Guidance rather than a workflow change is the right call for this.

Two things on the prose, both about the text as it will be read by a session mid-run.

The first: the `Otherwise` branch doesn't say what it is the alternative to, and the two readings point opposite ways. If it means "you pushed, but without a standing directive", announcing the findings as unowned is right — they sit on your commits and you aren't authorized to fix them. If it means "you didn't push at all", the review already reached the contributor via GitHub's own notification, and a bot comment telling their thread the findings are unowned is both inaccurate and noise.

The second: the new `Recheck Before Posting` paragraph leads with "a review submitted while you worked is a directive, not a duplicate", which is the thesis of **A new entry may be a directive, not a duplicate** three paragraphs below it in the same section. Stating it twice, the second time under its own bold heading, costs a reader more than it buys — the part that isn't already there is the projection rule and the fork-PR pointer.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied in 967c2d2. You're right that the two readings point opposite ways, and the "didn't push at all" one is the case the original text would have got wrong — a comment declaring the contributor's own commits unowned is noise on their thread. The three branches are now named separately.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied in 967c2d2. Agreed — the lead clause was restating the bold heading below it, and the projection rule plus the fork-PR pointer are the part that wasn't already in the section.

PR #990, opened 2026-08-15
fix(mention): keep both halves of the 👀 reaction in one job

### LINKED ISSUE #989: tend-mention: 👀 is added in `verify` but removed in `handle`, so a handle evicted while queued strands the reaction
In `mention.yaml.j2` the 👀 goes on in the `verify` job and comes off in `handle`, and `handle` is the only one of the two carrying a concurrency group. `verify` has no group at all (there's no workflow-level `concurrency:` either), so it always runs and always reacts; `handle` sits in `tend-mention-handle-<n>` with `cancel-in-progress: false`. That keeps a *running* handle alive but does nothing for a *pending* one — [workflow-syntax](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency): "By default, any existing pending job or workflow in the same concurrency group will be canceled and the new queued job or workflow will take its place." A job cancelled while queued never allocates a runner, so none of its steps execute — `if: always()` included, since `always()` governs step execution within a job that started.

So on a burst of three mentions against one thread: run 1's `handle` is running, run 2's is pending, run 3 arrives and evicts run 2. Run 2's `verify` already reacted. That comment keeps a 👀 that says "a session is working on this right now" with no session and nothing left to remove it — the macro's own contract ([macros.yaml.j2#L138-L152](https://github.com/max-sixty/tend/blob/205bb4a7fc16f18905b81bc335cafdbcd2b227fd/generator/src/tend/templates/macros.yaml.j2#L138-L152)) says `unreact_eyes` runs "under `always()`, so a failed or cancelled run doesn't strand it", which holds only when both halves are in the same job. The eviction is the routine path, but the general shape is broader: any route where `handle` never starts (a `verify` failure after the react step, a run cancelled before `handle` allocates, an environment-protection hold that later times out) leaves the reaction behind.

`review.yaml.j2` and `triage.yaml.j2` don't have this: their add and remove are steps in the same job, so the reaction can only exist if the job that removes it started.

<details><summary>Source lines</summary>

- [`mention.yaml.j2#L168`](https://github.com/max-sixty/tend/blob/205bb4a7fc16f18905b81bc335cafdbcd2b227fd/generator/src/tend/templates/mention.yaml.j2#L168) — `react_eyes(...)`, last step of `verify`
- [`mention.yaml.j2#L170-L175`](https://github.com/max-sixty/tend/blob/205bb4a7fc16f18905b81bc335cafdbcd2b227fd/generator/src/tend/templates/mention.yaml.j2#L170-L175) — `handle:` and its `tend-mention-handle-<n>` group with `cancel-in-progress: false`
- [`mention.yaml.j2#L214`](https://github.com/max-sixty/tend/blob/205bb4a7fc16f18905b81bc335cafdbcd2b227fd/generator/src/tend/templates/mention.yaml.j2#L214) — `unreact_eyes(...)` in `handle`

Rendered equivalent on the adopting side, from max-sixty/worktrunk#3822 (the 0.1.15 → 0.1.17 bump where this surfaced): [`tend-mention.yaml#L348-L362`](https://github.com/max-sixty/worktrunk/blob/611577aff94e6770d29944d8778737a402824d03/.github/workflows/tend-mention.yaml#L348-L362) and [`#L453-L470`](https://github.com/max-sixty/worktrunk/blob/611577aff94e6770d29944d8778737a402824d03/.github/workflows/tend-mention.yaml#L453-L470).

I haven't observed a stranded reaction in the wild — worktrunk hasn't merged 0.1.17 yet, so the steps aren't live there. This is structural, from the job layout plus the documented queue semantics.

</details>

## Fix options

1. **Move the react into `handle`.** Both halves land in the grouped job, matching review and triage, and the macro's same-job contract holds again. Cost: the reaction arrives after `verify` finishes rather than at the end of it — a few seconds on a signal that already lags the comment by the length of `verify`.
2. **Give `verify` a cleanup path** — a follow-on step or job that removes the reaction when `handle` didn't run. More moving parts, and it can't fire for the eviction case unless it lives outside the group.
3. **`queue: max` on `handle`'s concurrency block** — up to 100 pending runs instead of one, so nothing is evicted. This is compatible with `cancel-in-progress: false` (the docs only forbid `queue: max` with `cancel-in-progress: true`), but it changes cost behavior: today's eviction quietly drops redundant sessions on a busy thread, and `queue: max` would run every one of them. That looks like a regression, not a fix, unless the queued-session cost is wanted for other reasons.

Option 1 looks like the smallest correct change.

## Smaller, same feature: the remove step reads only page 1 of reactions

[`macros.yaml.j2#L174`](https://github.com/max-sixty/tend/blob/205bb4a7fc16f18905b81bc335cafdbcd2b227fd/generator/src/tend/templates/macros.yaml.j2#L174) lists reactions with `gh api "repos/$REPO/$TARGET/reactions?content=eyes"` and no `--paginate`, so it sees the first 30 and no more. If the bot's own 👀 isn't among them, `REACTION_ID` is empty and the reaction is silently kept — the step exits 0 either way.

For `mention` the target is a single comment, where 30+ 👀 is far-fetched. For `review` and `triage` the target is `issues/<number>` — the issue or PR itself, where 👀 is the standard "watching this" reaction and 30 of them on a popular thread is ordinary. `--paginate` on that call fixes it; the `--jq` filter and `head -n1` already handle the rest.

tend-agent:
## Problem

`tend-mention` put the 👀 on in `verify` and took it off in `handle`. `always()` on the remove step only governs step execution *within a job that started* — a job cancelled while still queued allocates no runner and runs no steps at all. `handle`'s `cancel-in-progress: false` holds a *running* job; GitHub still evicts a *pending* one when a newer run arrives ([workflow-syntax](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency)). So a burst of three mentions on one thread evicts run 2's pending `handle` after its `verify` already reacted, leaving a 👀 that claims a session is working with no session and nothing left to remove it. Eviction is the routine path; any route where `handle` never starts (a `verify` failure after the react, an environment-protection refusal) has the same shape. `review` and `triage` were never exposed — both halves are steps of one job there.

Separately, the remove step listed reactions with no `--paginate`, so it read only the first 30. On `review` and `triage` the target is the issue/PR itself, where 👀 is also the ordinary "watching this" reaction — past 30 of them the bot's own falls off page 1, `REACTION_ID` comes back empty, and the step exits 0 having silently kept the reaction.

Both bugs strand the same signal, so they're fixed together.

## Solution

Move `react_eyes` into `handle` as its first step, matching `review` and `triage`: the eyes can only go on once the job that removes them has started. Its `if:` drops the `should_run` conjunct (the job's own `if` already carries it) and reads `needs.verify.outputs.reason` instead of `steps.check.outputs.reason`.

Add `--paginate` (and `per_page=100`) to the reaction lookup in `unreact_eyes`; the existing `--jq` filter and `head -n1` already handle the rest.

Of the issue's three options this is #1. #3 (`queue: max`) would run every queued session instead of dropping the redundant ones — a cost change, not a fix.

## The trade this makes

What a user sees changes, and not only by a few seconds. On an idle thread the 👀 now lands after `verify` finishes rather than at the end of it — seconds, on a signal that already lags the comment by the length of `verify`. On a busy thread — the burst case this fix exists for — `handle` queues behind the in-flight session under `tend-mention-handle-<number>` with `cancel-in-progress: false`, so the second mention's 👀 waits out that whole session, and a third mention evicts the queued `handle` before it ever reacts, leaving the second comment with no acknowledgment at all. Before this change that comment got its 👀 immediately — stranded, which is the bug.

That's the intended trade: 👀 means "a session is working on this right now", and under the new placement it is never claimed falsely. Silence on a queued comment is recoverable — the handling prompt already tells the session to pick up other unaddressed comments on the thread — whereas a stranded 👀 is a standing false claim nothing clears.

## Testing

`test_eyes_come_off_when_the_session_ends` now asserts the invariant the bug violated — that exactly one job in each workflow carries eyes steps — plus `--paginate` on the lookup. Both assertions fail on the old templates (`mention: the eyes are split across {'handle', 'verify'}`; `triage/review: the reaction lookup only reads the first page`) and pass after. Regtest snapshots regenerated; full Python suite green (554 passed).

The generated `tend-*.yaml` in this repo are untouched — they track the released generator and update on the nightly regen.

---
Closes #989 — automated triage

### DIFF
--- a/generator/src/tend/templates/macros.yaml.j2
+++ b/generator/src/tend/templates/macros.yaml.j2
@@ -142,11 +142,20 @@
    unattended — and `unreact_eyes` takes it off at the end, under `always()`, so
    a failed or cancelled run doesn't strand it.
 
+   Both halves have to sit in the *same* job, which is what makes `always()`
+   mean anything: it governs step execution within a job that started, and a
+   job cancelled while still queued never allocates a runner, so no step of it
+   runs. Split across jobs, every route where the second never starts leaves
+   the eyes on with no session behind them.
+
    `target` is an API path under `repos/<owner>/<repo>/`: `issues/<number>` for
    an issue or PR (GitHub files PRs as issues for reactions), `issues/comments/`
    or `pulls/comments/<id>` for a comment. A repeat react returns the existing
    reaction rather than erroring, and the delete is filtered to the bot's own
-   reaction so a human's 👀 survives.
+   reaction so a human's 👀 survives. That lookup paginates: 👀 is also the
+   ordinary "watching this" reaction, so on a busy issue the bot's own can sit
+   past the first page, and finding nothing there keeps the reaction silently
+   (the step exits 0 either way).
 
    Either half failing warns rather than failing the job — an acknowledgment is
    not worth losing the run that does the work, and the unreact half runs on
@@ -171,7 +180,8 @@
 <<step_if(if_condition)>>
 {% endif %}
         run: |
-          REACTION_ID=$(gh api "repos/$REPO/$TARGET/reactions?content=eyes" \
+          REACTION_ID=$(gh api --paginate \
+            "repos/$REPO/$TARGET/reactions?content=eyes&per_page=100" \
             --jq ".[] | select(.user.login == \"$BOT_NAME\") | .id" | head -n1)
           if [ -n "$REACTION_ID" ]; then
             gh api -X DELETE "repos/$REPO/$TARGET/reactions/$REACTION_ID" --silent \
--- a/generator/src/tend/templates/mention.yaml.j2
+++ b/generator/src/tend/templates/mention.yaml.j2
@@ -161,12 +161,6 @@ jobs:
           PAYLOAD_PR: ${{ github.event.client_payload.pr }}
           PAYLOAD_ID: ${{ github.event.client_payload.id }}
 
-      # The dispatch arm covers a relayed inline comment, whose id the check
-      # step verified belongs to this PR; a review *submission* has no single
-      # comment to react to, so it gets no eyes — same as when the events
-      # arrived directly.
-<<react_eyes(reaction_target, if_condition="steps.check.outputs.should_run == 'true'\n&& " ~ summoned(cfg.bot_name, 'steps.check.outputs.reason'))>>
-
   handle:
     needs: verify
     if: needs.verify.outputs.should_run == 'true'
@@ -178,6 +172,19 @@ jobs:
     permissions:
 <<permissions()>>
     steps:
+      # Both halves of the reaction belong to this job, so the eyes can only
+      # go on once the job that takes them off has started. Put them in
+      # `verify` and `handle` respectively and the routine burst case strands
+      # them: a third mention on one thread evicts the second's pending
+      # `handle` — a job cancelled while queued allocates no runner and runs
+      # no steps, `always()` included — while its `verify` already reacted.
+      #
+      # The dispatch arm covers a relayed inline comment, whose id the check
+      # step verified belongs to this PR; a review *submission* has no single
+      # comment to react to, so it gets no eyes — same as when the events
+      # arrived directly. The job's own `if` already carries `should_run`.
+<<react_eyes(reaction_target, if_condition=summoned(cfg.bot_name, 'needs.verify.outputs.reason'))>>
+
 <<checkout(cfg)>>
 <<setup>>
       - name: Check out PR branch
--- a/generator/tests/_regtest_outputs/test_generate.test_extras_apply_path_regtest.out
+++ b/generator/tests/_regtest_outputs/test_generate.test_extras_apply_path_regtest.out
@@ -149,7 +149,8 @@ jobs:
       - name: Remove the eyes reaction
         if: always() && (steps.gate.outputs.should_run == 'true')
         run: |
-          REACTION_ID=$(gh api "repos/$REPO/$TARGET/reactions?content=eyes" \
+          REACTION_ID=$(gh api --paginate \
+            "repos/$REPO/$TARGET/reactions?content=eyes&per_page=100" \
             --jq ".[] | select(.user.login == \"$BOT_NAME\") | .id" | head -n1)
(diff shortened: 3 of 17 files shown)


tend-agent (commented):
The fix is right and lands at the right depth — the invariant now lives in the macro comment and in an assertion (`reacting == {job}`) rather than in a per-workflow convention, so a future split fails a test instead of stranding reactions. `--paginate` on the lookup is the only way to reach the bot's own reaction on a busy thread, since the delete needs a reaction id and there's no "delete my reaction" endpoint. I rendered the mention/review/triage workflows (Claude and Codex) and ran actionlint 1.7.12 over them — clean, including the new block-scalar `if:` and the reworked `run:` under shellcheck. The regtest snapshots match exactly the set of templates that render eyes steps.

One thing the description understates. "The reaction now lands after `verify` finishes rather than at the end of it — a few seconds" holds only for an unqueued mention. In the burst case — the one the fix exists for — `handle` sits behind an in-flight session under `tend-mention-handle-<number>` with `cancel-in-progress: false`, so the second mention's 👀 now waits out the *whole* first session, and a third mention evicts that queued `handle` before it ever reacts, leaving the second comment with no acknowledgment at all. Previously it got one immediately (stranded, which is the bug).

I think that's the correct trade — 👀 means "a session is working on this right now", and under the new placement it is never claimed falsely, whereas silence on a queued comment is recoverable via the prompt's "handle any other unaddressed comments too". But it's a visible change to what a user sees on a busy thread, not a few seconds of latency, and worth stating that way in the description so the trade is signed off knowingly rather than discovered later.

tend-agent:
Fair — "a few seconds" only described the idle case, and the burst case is the one the fix exists for. Rewrote the description: the latency claim now covers both, and a **The trade this makes** section states plainly that a queued mention's 👀 waits out the in-flight session and a third mention can evict it before it ever reacts, so that comment gets no acknowledgment where it previously got a stranded one.

No code change — the placement is still the right one, and the argument for it is now in the description rather than implied.

PR #992, opened 2026-08-15
docs(running-in-ci): scope PR-description claims to the merge base

### LINKED ISSUE #991: running-in-ci: PR-description edits are scoped to the whole PR, but the run's diff is scoped to LAST_REVIEW_SHA
A review run reads the diff from `LAST_REVIEW_SHA`, but when it edits the PR *description* that edit is scoped to the whole PR. On nightly's rolling `tend/update-workflows` PR — which every consumer accumulates, because step 7 folds each new release into the open branch rather than opening a fresh PR — those two scopes are different releases apart, and the description ends up asserting the last increment's behavior delta as if it were the PR's.

## What fires

`review/SKILL.md` sets `LAST_REVIEW_SHA` from `.last_substantive.sha` and reviews `LAST_REVIEW_SHA..HEAD_SHA` (L37, L64). That is right for the review itself — re-reviewing already-reviewed commits is the thing the incremental exists to avoid. But `running-in-ci`'s **Keeping PR Titles and Descriptions Current** then says to update the description when the approach changed, and says nothing about what base a description claim is scoped to. The run has an increment-scoped diff in context and a whole-PR artifact to edit, so the increment's conclusion goes into the description verbatim.

Nothing catches it afterwards: a wrong PR description never turns a check red, and the next run re-anchors on a still-newer `LAST_REVIEW_SHA`, so each successive edit is scoped one increment further from `main` than the last.

## Concrete instance

[max-sixty/cargo-affected#85](https://github.com/max-sixty/cargo-affected/pull/85), a rolling regen PR spanning `0.1.14 → 0.1.17` over three commits (`9d52136` = 0.1.15, `a93086b` = 0.1.16, `33e40ce` = 0.1.17). One bullet describes the `tend-mention` pre-check rewrite (max-sixty/tend#965, max-sixty/tend#970). It has now been rewritten three times, by three different runs, each anchored on a different base:

| Run | Base it diffed | Claim it wrote |
|---|---|---|
| `tend-nightly` [31675211312](https://github.com/max-sixty/cargo-affected/actions/runs/31675211312), carried forward by [31777503955](https://github.com/max-sixty/cargo-affected/actions/runs/31777503955) | `main` (0.1.14) | narrowing — "the empty synthetic containers no longer do" |
| `tend-review-runs` [31782916594](https://github.com/max-sixty/cargo-affected/actions/runs/31782916594) | `9d52136..a93086b` (0.1.15→0.1.16) | widening |
| `tend-review` [31869689877](https://github.com/max-sixty/cargo-affected/actions/runs/31869689877) | `9d52136..33e40ce` (0.1.15→0.1.17) | "Every reachable input maps to the same `should_run` before and after" |

The third claim is correct for the rewrite in isolation, and its unreachability argument holds — GitHub refuses a self-approval, so the one cell where the 0.1.15 and 0.1.17 gate orders disagree (bot review, `APPROVED`, empty body, fresh non-reply inline, bot-authored PR) can't be reached, and no bot `APPROVED` review exists on any of the 40 most recent PRs in that repo. As a statement about merging a `0.1.14 → 0.1.17` PR it is wrong, and it displaced the first run's claim, which was right. Against the 0.1.14 gate that repo's `main` actually runs, two reachable inputs move from `should_run=true` to `false`:

1. **Bot review on a bot-authored PR, empty body, no fresh inline** — the synthetic container GitHub wraps around the bot's own inline reply. 0.1.14 falls through the `approved && -z COMMENT_BODY` gate (state is `commented`, not `approved`) to the `PR_AUTHOR = bot` short-circuit and returns `true`; 0.1.15 adds the reply-only container gate (max-sixty/tend#866) and returns `false`. This is not hypothetical there — the tracking issue records it booting no-op sessions on five separate days.
2. **Bot review on a human-authored PR that isn't a bare empty-body approval** — e.g. the bodied `COMMENT` `tend-review` posts whenever it has findings. 0.1.14 reaches `BOT_REVIEWS` (non-empty, since it counts this very review) and returns `true`; 0.1.17's collapsed gate returns `false` for any bot-authored review on a PR the bot didn't author. 27 of the 60 most recent PRs in that repo are human-authored, so this one is live too.

Both changes are intended and correct. Only the description is wrong — which is the point: the code review found nothing, because there was nothing in the code to find.

## Why it recurs rather than being a one-off

The base mismatch is structural, not a lapse of attention. Any run reviewing an increment has the increment's diff in context and does not have `main..HEAD` unless it goes and fetches it, and the rolling regen PR guarantees the two differ by at least one release for every consumer that runs nightly. The failure is also invisible — no check goes red, so nothing prompts a later run to re-derive it.

## Proposed fix

In `running-in-ci`, under **Keeping PR Titles and Descriptions Current**, state the scope rule and give the base:

> A PR description describes the whole PR, not the increment you just reviewed. When a run edits the description, scope every behavior claim to the PR's merge base — `git merge-base "$(gh pr view N --json baseRefName --jq .baseRefName)" HEAD`, not `LAST_REVIEW_SHA`. On a long-lived branch (nightly's rolling `tend/update-workflows` PR carries several releases) those are different commits, and a claim true of the last increment can be false of the PR. If you can only verify the increment, say which base the claim is against rather than writing it as a claim about the PR.

The last sentence matters as much as the first: re-deriving the full `main..HEAD` delta is real work, and a scoped claim is honest where an unscoped one is a guess.

A matching pointer in `review/SKILL.md` next to the `LAST_REVIEW_SHA` setup would catch it at the point the wrong base enters context, but the `running-in-ci` rule is the load-bearing half — the description edit is reachable from any workflow, not just `review`.

tend-agent:
## Problem

A review run reads its diff from `LAST_REVIEW_SHA` but edits a whole-PR artifact when it updates the description, and **Keeping PR Titles and Descriptions Current** never said which base a description claim is scoped to. So the increment's conclusion goes into the description as if it described the PR. On a long-lived branch — nightly's rolling `tend/update-workflows` PR, which accumulates a release per run — the two bases are several releases apart, and nothing catches the mismatch afterwards: a wrong description never turns a check red, and each later run re-anchors on a still-newer `LAST_REVIEW_SHA`. #991 traces three successive runs rewriting the same bullet from three different bases.

## Solution

State the scope rule where the description edit happens, in `running-in-ci`: a description describes the whole PR, so scope behavior claims to the merge base, and when only the increment can be verified, name that base rather than writing an unscoped claim. `review/SKILL.md` gets a one-line pointer beside the incremental setup — the point where the wrong base enters context — rather than a second copy of the rule.

## Testing

No test: the failure is prose the agent reads, not code. `uvx pre-commit run --files <the two SKILL.md>` passes (typos, trailing whitespace, the plugin-skill bang-backtick and reference-sync hooks).

---
Closes #991 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -403,6 +403,14 @@ gh api repos/{owner}/{repo}/pulls/{number} -X PATCH \
   -f title="new title" -F body=@/tmp/updated-body.md
 ```
 
+**A description describes the whole PR, not the increment this run reviewed.** Scope every behavior claim in it to the PR's merge base — not `LAST_REVIEW_SHA`, and not whatever range this run happened to diff:
+
+```bash
+git merge-base "origin/$(gh pr view <number> --json baseRefName --jq '.baseRefName')" HEAD
+```
+
+On a long-lived branch those are different commits — nightly's rolling `tend/update-workflows` PR accumulates a release per run, so its head is several releases past its merge base — and a claim that is true of the last increment can be false of the PR. If you can only verify the increment, name the base the claim is against instead of writing it as a claim about the PR. Nothing downstream catches a wrong description: it never turns a check red, and each later run re-anchors one increment further out.
+
 ## Atomic PRs
 
 Split unrelated changes into separate PRs — one concern per PR. If one change could be reverted without affecting the other, they belong in separate PRs.
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -406,7 +406,7 @@ gh api repos/{owner}/{repo}/pulls/{number} -X PATCH \
 **A description describes the whole PR, not the increment this run reviewed.** Scope every behavior claim in it to the PR's merge base — not `LAST_REVIEW_SHA`, and not whatever range this run happened to diff:
 
 ```bash
-git merge-base "origin/$(gh pr view <number> --json baseRefName --jq '.baseRefName')" HEAD
+gh pr diff <number>   # merge-base→head, whatever this session has checked out
 ```
 
 On a long-lived branch those are different commits — nightly's rolling `tend/update-workflows` PR accumulates a release per run, so its head is several releases past its merge base — and a claim that is true of the last increment can be false of the PR. If you can only verify the increment, name the base the claim is against instead of writing it as a claim about the PR. Nothing downstream catches a wrong description: it never turns a check red, and each later run re-anchors one increment further out.
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -64,6 +64,8 @@ git fetch --no-tags --quiet origin "$BASE_SHA" 2>/dev/null || true
 git log --no-merges --numstat --format='%h %s' "$LAST_REVIEW_SHA..$HEAD_SHA" --not "$BASE_SHA"
 ```
 
+The incremental scopes the *review*, not anything this run writes about the PR as a whole: if you also edit the PR description, scope its claims to the merge base per **Keeping PR Titles and Descriptions Current** in `/tend-ci-runner:running-in-ci`.
+
 If the incremental changes are trivial, skip the full review — go directly to step 7 to resolve any bot threads addressed by the new changes. After resolving threads: if the most recent bot review was a COMMENT that flagged issues, and those issues are now addressed, submit an APPROVE with an empty body so the PR isn't left in limbo. Otherwise do not submit a new review — the existing one stands. Do NOT proceed to steps 2–6; finish at step 9. Rough heuristic: changes under ~20 added+deleted lines that don't introduce new functions, types, or control flow are typically trivial.
 
 **Commit and PR authorship do not affect review behavior.** Apply the same trivial-vs-substantive heuristic regardless of who pushed the new commits. When `tend-notifications` or `tend-ci-fix` pushes a fix to a human-authored PR, reviewing (and re-approving) the updated state is expected — the reviewer role is independent of commit authorship.


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -403,6 +403,14 @@ gh api repos/{owner}/{repo}/pulls/{number} -X PATCH \
   -f title="new title" -F body=@/tmp/updated-body.md
 ```
 
+**A description describes the whole PR, not the increment this run reviewed.** Scope every behavior claim in it to the PR's merge base — not `LAST_REVIEW_SHA`, and not whatever range this run happened to diff:
+
+```bash
+git merge-base "origin/$(gh pr view <number> --json baseRefName --jq '.baseRefName')" HEAD
```
tend-agent: This pairs a PR-number query with the local `HEAD`, so it assumes the session's checkout *is* that PR's head — which the sessions that edit descriptions often aren't. `tend-review-runs` checks out with no ref (the default branch) and works on its own `daily/review-runs-*` branch; nightly regenerates in a `/tmp` worktree. From either, `merge-base` returns the base tip and exits 0, so the wrong base is silent — the same failure shape the paragraph exists to close, and #991's table names a `tend-review-runs` run as one of the three that rewrote the bullet.

`gh pr diff` is already this repo's checkout-independent form of the same thing (`review/SKILL.md`, step 1: "The PR's three-dot diff — `gh pr diff <number>` (merge-base→head, the same diff step 3 uses)"), and it also sidesteps `origin/<baseRefName>` not being materialized — the reason the incremental recipe right above it carries an explicit `git fetch`.

```suggestion
gh pr diff <number>   # merge-base→head, whatever this session has checked out
```

tend-agent (commented):
The rule and its placement look right, and the `tend/update-workflows` example holds for every adopter (nightly rebuilds that branch in each consumer repo), so it earns its place in a bundled skill. One issue with the recipe — inline.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Right — and the failure mode is exactly the one the paragraph is meant to close: from a `daily/review-runs-*` branch or nightly's `/tmp` worktree, `merge-base` against the base tip exits 0 and returns a commit that isn't this PR's base, so the wrong scope is silent. Applied the suggestion in 9e86f6c; `gh pr diff <number>` is checkout-independent and matches step 1 of `review/SKILL.md`, so the recipe now works from whatever ref the editing session happens to hold.

PR #994, opened 2026-08-15
fix(checks): report only org secrets the repo can actually read

### LINKED ISSUE #993: tend check: repo-secret-allowlist reports org secrets that are not shared with the repo
**tend 0.1.17**

With `admin:org` granted so the org-level sweep can run, `repo-secret-allowlist` reports every org-level secret in the organization, without consulting whether the secret's visibility actually admits the repo being checked.

## Observed

On `diffplug/dormouse`, all ten org secrets have `visibility: selected`, and none of their selected-repositories lists include `diffplug/dormouse`:

```
$ gh api orgs/diffplug/actions/secrets --jq '.secrets[] | "\(.name)\t\(.visibility)"'
BUILDCACHE_PASS   selected
BUILDCACHE_USER   selected
CHOCO_API_KEY     selected
...all ten: selected, and none list diffplug/dormouse
```

`tend check` nonetheless reports:

```
FAIL  repo-secret-allowlist — Unexpected secrets (org-level: BUILDCACHE_PASS, BUILDCACHE_USER,
CHOCO_API_KEY, CHOCO_USER, GPG_KEY64, GPG_PASSPHRASE, GRADLE_KEY, GRADLE_SECRET, NEXUS_PASS64,
NEXUS_USER). These are available to all workflows, including those triggered by PRs.
```

None of those ten are injected into any workflow in this repo.

## Why it's awkward

No configuration clears it. The repo is already at the tightest scoping GitHub offers — the secrets are not shared with it at all. The only lever left is `secrets.allowed`, whose documented meaning is "a deliberate acceptance that any workflow the repo runs can read that token." That statement is false here, and adding the entries would suppress the check for those names permanently — so if one of them *were* later shared with this repo, the sweep would stay silent about the one case it exists to catch.

There's also a mild disincentive built in: the check hints at granting `admin:org` to complete the sweep, but granting it converted a passing check into a permanent FAIL that no repo-side action can resolve. Before the grant, the same ten secrets were equally unreachable and the check passed.

## Suggested fix

Report only the org secrets the repo can actually read. Per secret, skip when:

- `visibility: selected` and `GET /orgs/{org}/actions/secrets/{name}/repositories` doesn't include the repo;
- `visibility: private` and the repo is public;
- the org is on GitHub Free and the repo is private — org secrets reach only public repositories on that plan. (We ran into this from the other side: the repository picker for an org secret offers only public repos, so two private repos referencing `BUILDCACHE_USER` in their workflows had been silently resolving it to an empty string.)

The per-secret repositories endpoint needs `admin:org`, which the sweep already requires to enumerate org secrets at all, so no additional scope is involved.

Happy to test a patch against this org if useful — it has a mix of `selected` scopes, public and private repos, and a free plan, which seems to cover the interesting cases.

tend-agent:
## Problem

`repo-secret-allowlist` enumerated every org-level secret in the organization and flagged each one, without consulting whether that secret's visibility actually admits the repo being checked. On a repo where all ten org secrets are `visibility: selected` and none list it, the check reported all ten as "available to all workflows".

Nothing clears that. The repo is already at the tightest scoping GitHub offers — the secrets are not shared with it at all — so the only remaining lever is a `secrets.allowed` entry, whose documented meaning ("a deliberate acceptance that any workflow the repo runs can read that token") is false in that state, and which would mute those names permanently: if one *were* later shared with the repo, the sweep would stay silent about the one case it exists to catch.

It also penalized the grant it asks for. The check hints at `admin:org` to complete the sweep; granting it converted a pass into an unclearable FAIL, while the same ten secrets were equally unreachable before.

## Solution

`_list_org_secrets` now takes the repo and returns only the org secrets that repo can read. It reads each secret's `visibility` alongside its name and drops:

- `selected` whose `/orgs/{org}/actions/secrets/{name}/repositories` list omits the repo (paginated — an org sharing a secret past one page would otherwise look like it omits this one);
- `private` when the repo is public;
- all of them when the repo is private and the org is on GitHub Free, which serves org secrets to public repositories only.

The per-secret repositories endpoint needs `admin:org`, which the sweep already requires to enumerate org secrets at all, so no new scope is involved.

Filtering is deliberately one-directional: a secret whose reach can't be determined — the repositories list fails, the repo's visibility won't read, the plan field is absent — stays in the reported set. Under-reporting hides real exposure; over-reporting is only noise. `_org_plan_is_free` returns False on any read failure for the same reason.

`check_secrets` reads the same helper, so it gains the same correction: an org copy scoped away from the repo no longer gets named as the thing keeping its workflows running.

## Testing

`uv run pytest` — 563 pass. Seven new cases in `generator/tests/test_checks.py` drive `_list_org_secrets` end-to-end through a fake `gh` serving the real endpoint shapes, rather than mocking the helper out. Four reproduced the reported FAIL before the fix: a `selected` secret omitting the repo, `private` against a public repo, a free-org private repo, and `check_secrets` naming an unreachable org copy. The other three pin the directions that must keep failing — `selected` including the repo, `visibility: all`, and a repositories lookup that errors.

<details><summary>Not covered</summary>

Pagination of the org secret listing itself is untouched — `orgs/{org}/actions/secrets` still reads one page, so an org with more secrets than a page holds under-reports. That is pre-existing and independent of visibility; separate change if it matters.

The free-plan rule rests on `.plan.name`, which GitHub returns only to org owners. For a non-owner token the field is absent and the rule simply doesn't fire, leaving the pre-fix behavior for `visibility: all` on a free org. `selected` and `private` are unaffected — they're filtered from data any `admin:org` token can read.

</details>

---
Closes #993 — automated triage

### DIFF
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -1415,7 +1415,9 @@ def _list_org_secrets(org: str, repo: str) -> tuple[set[str] | None, bool]:
     for name, visibility in listed:
         if visibility == "selected":
             shared = _org_secret_repos(org, name)
-            if shared is not None and repo not in shared:
+            if shared is not None and repo.casefold() not in {
+                r.casefold() for r in shared
+            }:
                 continue
         elif visibility == "private" and is_public:
             continue
--- a/generator/src/tend/checks.py
+++ b/generator/src/tend/checks.py
@@ -1321,7 +1321,7 @@ def check_secrets(repo: str, expected: list[str]) -> CheckResult:
         )
 
     org = repo.split("/")[0] if "/" in repo else None
-    org_secrets, org_forbidden = _list_org_secrets(org) if org else (None, False)
+    org_secrets, org_forbidden = _list_org_secrets(org, repo) if org else (None, False)
     found_at_org = [s for s in missing if org_secrets and s in org_secrets]
 
     msg = (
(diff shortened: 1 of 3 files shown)


generator/src/tend/checks.py
```diff
@@ -1344,19 +1344,84 @@ def check_secrets(repo: str, expected: list[str]) -> CheckResult:
...
 
+    is_public = _repo_is_public(repo)
+    if is_public is False and _org_plan_is_free(org):
+        return set(), False
+
+    reachable = set()
+    for name, visibility in listed:
+        if visibility == "selected":
+            shared = _org_secret_repos(org, name)
+            if shared is not None and repo not in shared:
```
tend-agent: `full_name` comes back in GitHub's canonical casing, but `repo` is whatever the user typed after `--repo` — `cli.py` only falls back to `detect_repo()` when the flag is omitted. A casing difference makes every `selected` secret look unshared and drops it silently, which is the under-reporting direction this function is explicitly written to avoid.

```suggestion
            if shared is not None and repo.casefold() not in {
                r.casefold() for r in shared
            }:
```

tend-agent (commented):
Two findings below, both mechanical. The PR is bot-authored, so I'll push them as a follow-up commit rather than leave one-click suggestions waiting on an author who won't return.

One thing I couldn't verify from CI: `_org_secret_repos`'s docstring claims `--paginate` covers an org that shares a secret with more repos than one page holds. `gh api --paginate` follows `Link` headers, and I have no `admin:org` token here to confirm `orgs/{org}/actions/secrets/{name}/repositories` emits them. If it doesn't, the call silently stops at the default 30 and produces exactly the under-report the filter exists to prevent. `-f per_page=100` would make the question moot; I've left it out rather than add a knob on a guess.

generator/tests/test_checks.py
```diff
@@ -122,6 +122,44 @@ def _login_response(login: str | None) -> subprocess.CompletedProcess[str]:
...
+            if repositories_rc != 0:
+                return _make_completed(returncode=repositories_rc, stderr="HTTP 404")
+            return _make_completed("\n".join(shared.get(url.split("/")[-2], [])) + "\n")
+        if url == "orgs/acme/actions/secrets":
+            jq = args[args.index("--jq") + 1] if "--jq" in args else ""
+            listed = (
+                [{"name": n, "visibility": v} for n, v in org_secrets]
+                if "visibility" in jq
+                else [n for n, _ in org_secrets]
+            )
```
tend-agent: `_list_org_secrets` is the only reader of this endpoint and always passes the `{name, visibility}` jq, so the `else` arm and the `jq` lookup are dead — replacing the arm with a `raise` leaves the suite green. `_secret_names` above serves two shapes that are both live; this one only ever serves one, and carrying the other is the compatibility layer CLAUDE.md rules out.

```suggestion
            listed = [{"name": n, "visibility": v} for n, v in org_secrets]
```

tend-agent:
Both findings are on the branch as c22238b (case-folded membership test plus its own case, dead `else` arm dropped from the test helper); checks are green on that head.

On the open question from the review — whether `--paginate` actually covers `orgs/{org}/actions/secrets/{name}/repositories`, or silently stops at 30: the mechanism holds for this endpoint family. GitHub emits `Link` headers on the Actions `total_count`-envelope endpoints, and `gh api --paginate --jq` follows them across pages rather than parsing only the first:

```
$ gh api "repos/max-sixty/tend/actions/workflows" --jq '.total_count'
17
$ gh api "repos/max-sixty/tend/actions/workflows?per_page=2" --jq '.workflows[].name' | wc -l
2
$ gh api --paginate "repos/max-sixty/tend/actions/workflows?per_page=2" --jq '.workflows[].name' | wc -l
17
```

So no `-f per_page=100` is needed to make the docstring's claim true — it would only cut request count, which isn't worth a knob.

What stays unverified is the specific endpoint: reading it needs `admin:org`, which no token in this repo's runs has, so the check above is same-family evidence rather than a direct test. If it ever turns out to be the one Actions endpoint without `Link`, the failure is the under-report the filter exists to prevent, so it's worth naming here rather than leaving in a docstring.

PR #999, opened 2026-08-16
fix(scripts): page the check rollup instead of refusing past 100 contexts

### LINKED ISSUE #998: poll-pr-checks.sh: contexts(first: 100) has no cursor, so any commit with >100 checks is permanently UNVERIFIED
`poll-pr-checks.sh` queries `statusCheckRollup.contexts(first: 100)` with no cursor and bails `OVERFLOW` → exit 2 when `totalCount > 100`. On a repo whose full CI matrix registers more than 100 contexts, that makes the script permanently unusable on exactly the PRs it matters most for — and the sessions that hit it improvise a replacement poller in-flight, losing the grouping and self-filtering logic the script exists to provide.

Observed twice on `PRQL/prql` in the 2026-08-15 Dependabot batch, on both reviews that were gated on CI before approving:

| Run | PR | Head | `contexts.totalCount` | Result |
|---|---|---|---|---|
| [31891399008](https://github.com/PRQL/prql/actions/runs/31891399008) | PRQL/prql#6196 | `472a1925` | 116 | `exit 2` — UNVERIFIED |
| [31891406039](https://github.com/PRQL/prql/actions/runs/31891406039) | PRQL/prql#6197 | `0aff65a7` | 115 | `exit 2` — UNVERIFIED |

The split is systematic rather than incidental: a Dependabot bump touches `Cargo.lock`, so every path filter opens and the whole matrix (including the `nightly` release builds) registers — 115–116 contexts. Path-filtered bot PRs in the same window stayed well under the page: PRQL/prql#6199 at 36, PRQL/prql#6198 at 41, PRQL/prql#6194 at 43, PRQL/prql#6193 at 87. So the primitive is available on the PRs a human will look at anyway, and unavailable on the external-dependency bumps tend approves autonomously. The two runs above were the only two gated approvals in that 24h window, and the script refused on both.

No wrong verdict reached the record — both sessions recovered, and PRQL/prql#6196's improvised poller is what caught `test-msrv` going red four minutes after the approval, which the session then dismissed, diagnosed and fixed. The cost is that each session re-derives a safety-critical primitive under time pressure, and the two derivations were not the same: 31891406039 fell back to `gh pr view --json statusCheckRollup`, while 31891399008 hand-wrote a GraphQL cursor loop into `/tmp/poll6196.sh`. Neither inherits the reduction the script's header argues for at length — grouping per (check name, workflow), latest-wins within a settled group, `$GITHUB_RUN_ID`/`$GITHUB_WORKFLOW` self-filtering — which is what keeps a concurrency-cancelled sibling from reading as red. On this repo that is not hypothetical: an `if: always()` omnibus (`check-ok-to-merge`) sat at `FAILURE` from a cancelled run for the whole of both sessions.

The guard itself is right — UNVERIFIED beats a false green when a dropped node could hide a failure. The gap is that the query never asks for the next page.

<details><summary>Suggested fix and verification notes</summary>

The query already has everything but the cursor. Adding `pageInfo { hasNextPage endCursor }`, looping with `after:` while accumulating `nodes`, then dropping the `"OVERFLOW"` branch (`scripts/poll-pr-checks.sh:91`), `too_many_contexts()` (`:131-135`), its two call sites (`:144`, `:152`) and the exit-2 clause in the header comment (`:39-40`) removes the failure mode rather than raising the constant. `totalCount` stops being load-bearing once the loop terminates on `hasNextPage`.

`gh pr view --json statusCheckRollup` is **not** a drop-in substitute, even though it does paginate — I verified it returns all 115 nodes on `0aff65a7` where the single-page GraphQL query reports `totalCount: 115, hasNextPage: true`. The script deliberately queries by commit OID rather than through the PR (`scripts/poll-pr-checks.sh:7-10`), and `gh pr view` follows the current head, which is the exact bug that pinning was added to prevent.

A repo-side workaround (raising the page size, or an overlay note telling sessions to hand-roll) would leave every other consumer with a large matrix in the same place, which is why this is filed here rather than on `PRQL/prql`.

Reproduce against any commit on this repo:

```sh
gh api graphql -f query='
  query($o:String!,$r:String!,$s:String!){
    repository(owner:$o,name:$r){ object(expression:$s){ ... on Commit {
      statusCheckRollup { contexts(first:100){ totalCount pageInfo{hasNextPage} } } } } }
  }' -f o=PRQL -f r=prql -f s=0aff65a743f11a02f061a09e1fd40ad13dc8f487 \
  --jq '.data.repository.object.statusCheckRollup.contexts'
# {"pageInfo":{"hasNextPage":true},"totalCount":115}
```

</details>

tend-agent:
## Problem

`poll-pr-checks.sh` asked for `contexts(first: 100)` with no cursor and bailed `exit 2` (UNVERIFIED) whenever `totalCount` exceeded the page. A full CI matrix passes 100 contexts routinely — a dependency bump touches a lockfile, every path filter opens, and the whole matrix registers — so the poll was unavailable on exactly the PRs the bot approves autonomously, and sessions hand-rolled a replacement poller in flight, each losing the grouping and self-filtering the script exists to provide. Reported in #998 with two runs that hit it at 115 and 116 contexts.

## Solution

The query takes a `$cursor` and selects `pageInfo { hasNextPage endCursor }`; `rollup()` walks pages to exhaustion, accumulating `nodes`, and the reduction runs over the accumulated array. `totalCount` stops being load-bearing, so the `"OVERFLOW"` branch, `too_many_contexts()` and its two call sites are gone — the failure mode is removed rather than the constant raised. `hasNextPage` with no `endCursor` to follow still routes to UNVERIFIED: re-reading page one would loop forever, and a page never read could hide a failure.

## Testing

`test_paginates_past_the_first_page` puts the only failing check on page two, reachable solely through the cursor — the fake `gh` now serves later pages keyed by the cursor it is given, so a query that doesn't paginate genuinely cannot see them. It failed as a false green (exit 0) before the change and reports red after. `test_truncated_pagination_never_reads_green` covers the cursorless-`hasNextPage` case. The single-page overflow test is deleted, since the behaviour it pinned was the bug.

Verified live against the commit from the report, `PRQL/prql@0aff65a7` (115 contexts): pages 1 and 2 return 100 + 15 nodes, and the script now reports `green: every gating check … settled green` where it previously exited 2.

<details><summary>Live check</summary>

```
== page 1 (cursor null) ==
{"total": 115, "n": 100, "page": {"hasNextPage": true, "endCursor": "MTAw"}}
== page 2 (cursor MTAw) ==
{"n": 15, "page": {"hasNextPage": false, "endCursor": "MTE1"}}
```

`-F cursor=null` types the literal as JSON null, so `after:` is a no-op on the first page; later pages pass the cursor with `-f` as a raw string.

</details>

---
Closes #998 — automated triage

### DIFF
--- a/generator/tests/test_poll_scripts.py
+++ b/generator/tests/test_poll_scripts.py
@@ -25,20 +25,31 @@ RERUN_FAILED_JOBS = SCRIPTS / "rerun-failed-jobs.sh"
 
 HEAD_SHA = "aaaa111122223333aaaa111122223333aaaa1111"
 
-# GraphQL responses are served in sequence: call N reads $ROLLUP_DIR/N.json,
-# falling back to final.json once the sequence runs out. The run endpoint
-# serves `run_attempt` values consumed line-by-line from $ATTEMPTS (the last
-# line repeats), so a rerun's attempt bump is scriptable. `pr view` and the
-# jobs endpoints run the script's `--jq` through real jq.
+# First-page GraphQL responses are served in sequence: poll N reads
+# $ROLLUP_DIR/N.json, falling back to final.json once the sequence runs out.
+# A call carrying a cursor is a *later page* of the poll in flight and is
+# served from $ROLLUP_DIR/page-<cursor>.json instead — so a page reachable
+# only through the cursor is genuinely unreachable without it. The run
+# endpoint serves `run_attempt` values consumed line-by-line from $ATTEMPTS
+# (the last line repeats), so a rerun's attempt bump is scriptable. `pr view`
+# and the jobs endpoints run the script's `--jq` through real jq.
 FAKE_GH = (
     GH_PREAMBLE
     + r"""case "$1 $2" in
   "api graphql")
-    n=$(( $(cat "$GRAPHQL_CALLS" 2>/dev/null || echo 0) + 1 ))
-    echo "$n" > "$GRAPHQL_CALLS"
-    f="$ROLLUP_DIR/$n.json"
-    [ -f "$f" ] || f="$ROLLUP_DIR/final.json"
-    cat "$f"
+    cursor=""
+    for arg in "$@"; do
+      case "$arg" in cursor=*) cursor="${arg#cursor=}" ;; esac
+    done
+    if [ -n "$cursor" ] && [ "$cursor" != "null" ]; then
+      cat "$ROLLUP_DIR/page-$cursor.json"
+    else
+      n=$(( $(cat "$GRAPHQL_CALLS" 2>/dev/null || echo 0) + 1 ))
+      echo "$n" > "$GRAPHQL_CALLS"
+      f="$ROLLUP_DIR/$n.json"
+      [ -f "$f" ] || f="$ROLLUP_DIR/final.json"
+      cat "$f"
+    fi
     ;;
   "pr view")
     emit "$(cat "$HEAD_JSON")"
@@ -100,7 +111,7 @@ def _status_ctx(context: str, state: str) -> dict:
     }
 
 
-def _resp(*nodes: dict, total: int | None = None) -> str:
+def _resp(*nodes: dict, has_next: bool = False, end_cursor: str | None = None) -> str:
     return json.dumps(
         {
             "data": {
@@ -108,7 +119,10 @@ def _resp(*nodes: dict, total: int | None = None) -> str:
                     "object": {
                         "statusCheckRollup": {
                             "contexts": {
-                                "totalCount": len(nodes) if total is None else total,
+                                "pageInfo": {
+                                    "hasNextPage": has_next,
+                                    "endCursor": end_cursor,
+                                },
                                 "nodes": list(nodes),
                             }
                         }
@@ -156,6 +170,11 @@ def _serve(env: dict[str, str], *responses: str) -> None:
     for i, resp in enumerate(responses, start=1):
         (Path(env["ROLLUP_DIR"]) / f"{i}.json").write_text(resp)
     (Path(env["ROLLUP_DIR"]) / "final.json").write_text(responses[-1])
+
+
+def _serve_page(env: dict[str, str], cursor: str, response: str) -> None:
+    """Serve *response* to any call that asks for the page after *cursor*."""
+    (Path(env["ROLLUP_DIR"]) / f"page-{cursor}.json").write_text(response)
 
 
 def _poll(env: dict[str, str]) -> subprocess.CompletedProcess[str]:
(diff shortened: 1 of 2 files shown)


generator/tests/test_poll_scripts.py
```diff
@@ -316,15 +335,30 @@ def test_null_rollup_never_reads_green(env: dict[str, str]) -> None:
...
+    on a dependency bump, which opens every path filter. The failing check can
+    sit on any page, so an all-green first page must never stand in for the
+    whole rollup."""
+    _serve(env, _resp(_check_run("tests"), has_next=True, end_cursor="Y3Vyc29yOjE"))
+    _serve_page(env, "Y3Vyc29yOjE", _resp(_check_run("lint", conclusion="FAILURE")))
+
+    result = _poll(env)
+
+    assert result.returncode == 1, result.stdout + result.stderr
+    assert "lint https://github.com/o/r/actions/runs/100/job/1" in result.stdout
```
tend-agent: `$acc + .nodes` is the line this test exists to pin, and it doesn't pin it. Swap it for `.nodes` — last page wins, page one discarded — and all 21 tests still pass. The failing check only ever sits on page two here, so a walk that reads page two and throws page one away is indistinguishable from a correct one.

That matters because the premise the PR argues from is that the failing check can sit on *any* page. Putting a red one on page one as well closes it — I ran this against the change (passes) and against the `.nodes` mutant (fails on the new assertion):

```suggestion
    _serve(
        env,
        _resp(
            _check_run("tests"),
            _check_run("build", conclusion="FAILURE", run_id=101),
            has_next=True,
            end_cursor="Y3Vyc29yOjE",
        ),
    )
    _serve_page(env, "Y3Vyc29yOjE", _resp(_check_run("lint", conclusion="FAILURE")))

    result = _poll(env)

    assert result.returncode == 1, result.stdout + result.stderr
    assert "lint https://github.com/o/r/actions/runs/100/job/1" in result.stdout
    assert "build https://github.com/o/r/actions/runs/101/job/1" in result.stdout
```

tend-agent (commented):
The pagination holds up. I checked the PR's central claim by running both new tests against the pre-change script: `test_paginates_past_the_first_page` exits 0 with `green: every gating check ... settled green`, and `test_truncated_pagination_never_reads_green` does the same — two false greens the change turns red. I also checked whether `gh api graphql --paginate` is the simpler substitute for the hand-rolled walk, since it does drive this exact connection (`contexts(first: N, after: $endCursor)` pages cleanly under it). It isn't: it stops silently when a page reports `hasNextPage` with no cursor to follow, and it can't distinguish a null rollup from a rollup with zero nodes — both would reduce to `{pending: [], failed: []}` and read as settled green. The explicit loop routes both to UNVERIFIED, so it earns its place here.

One gap in the test that pins the walk — details inline.

generator/tests/test_poll_scripts.py
tend-agent: Agreed, and it's now closed in f9bae40 — the second red on page one landed as suggested.

I confirmed the gap was real and is gone rather than taking the suggestion on faith: with `$acc + .nodes` swapped for `.nodes`, all 21 tests passed at f20cc20, and at f9bae40 that same mutant fails on the new `build` assertion (`test_paginates_past_the_first_page`), with the unmutated suite green. So the walk's accumulation is pinned, not just its cursor follow.

PR #1007, opened 2026-08-18
skills(review): stop at the review on PRs this bot authored

tend-agent:
Review step 8's **Bot PRs** paragraph is written for Dependabot and renovate — "there is no human author to act on feedback" — but it also reads as covering PRs this bot authored, where a successor *does* exist: submitting a review with a body or a fresh inline comment dispatches `tend-mention`, which boots as the author and is explicitly told to action that review. The successor boots either way, so a reviewer that also pushes the fix just makes it boot to discover the work is done, and its own commit ends up carrying the `tend-review/<n>` examined-stamp written by the session that authored it. This splits the paragraph in two so the third-party case keeps its "if you can articulate the fix, apply it" rule and the self-authored case stops at the review; verified against the generated `tend-mention` workflow, whose `verify` job boots on exactly that condition.

<details><summary>Evidence, mechanism, and gate assessment</summary>

### What was observed

Both rounds of review on [#1005](https://github.com/max-sixty/tend/pull/1005) ran this morning, and the two sessions read step 8 in opposite directions on the same PR:

| Round | Review session | Outcome |
|---|---|---|
| 1 | [32107391289](https://github.com/max-sixty/tend/actions/runs/32107391289) | Posted two findings and **declined to push**: *"I didn't push the fixes — the reviewer role here is meant to be the uncorrelated signal, and `tend-mention` dispatches on `pull_request_review` for bot-authored PRs, so a successor session owns applying them."* The dispatched mention run [32107876748](https://github.com/max-sixty/tend/actions/runs/32107876748) applied all three findings as `0b7c9d3`, replied to both inline threads, and polled CI green. |
| 2 | [32108204149](https://github.com/max-sixty/tend/actions/runs/32108204149) | Posted a new finding at 06:49:03 and **also pushed the fix itself** (`afb6e7c`, 06:50:17), reasoning *"this is a bot-authored PR with no human waiting on it"*. The mention run its own review dispatched, [32108568636](https://github.com/max-sixty/tend/actions/runs/32108568636), booted at 06:49:13, read the state, found the work landed, and exited silently — $0.69 and 12 turns for nothing. |

Same PR, same skill text, ninety seconds apart, opposite reads. That is the signature of an instruction that admits two answers, not of one session getting it wrong.

### The dispatch is deterministic, not incidental

`tend-mention`'s `verify` job boots a session on precisely the shape a reviewer creates on a bot PR (`.github/workflows/tend-mention.yaml`):

```bash
if [ "$KIND" = "pull_request_review" ] && [ "$REVIEW_AUTHOR" = "$BOT_NAME" ]; then
  if [ "$PR_AUTHOR" = "$BOT_NAME" ] \
     && { [ -n "$COMMENT_BODY" ] || [ "$FRESH_INLINE" -gt 0 ]; }; then
```

and the prompt it hands that session says to action the review "including when you authored the review: a review your review workflow left on your own PR is your reviewer role speaking, not a self-loop". So the author seat is already occupied by the time the reviewer decides whether to push.

`running-in-ci` already documents this mechanism, but only from the author's seat — "A review that lands while you poll is not yours to action". The reviewer's half was never written down, which is why step 8's Dependabot rule filled the gap.

### The second cost: a self-written commit carries its own examined-stamp

Round 2's session stamped `0b7c9d3` at 06:50:26, pushed `afb6e7c` at 06:50:17, then stamped `afb6e7c` at 06:52:08. The queued review run [32108653260](https://github.com/max-sixty/tend/actions/runs/32108653260) evaluated its gate at 06:52:33 and skipped on that stamp, so `afb6e7c` — the head #1005 currently sits on — was never examined by a session other than the one that wrote it. Round 1's path produced the opposite: `0b7c9d3` was written by the mention session and then genuinely reviewed by round 2, which found a further real defect in it.

### Gates

1. **Confidence — High, structural.** Three occurrences across three windows: `#993` (2026-08-15, folded by the sibling, no loss), `#999` (2026-08-17, $1.85 duplicated edit/test cycle), `#1005` round 2 (this window, $0.69). Structural because the dispatch condition above has no decision point. Recorded in [#801](https://github.com/max-sixty/tend/issues/801) across the 2026-08-17 and 2026-08-18 entries, where the 2026-08-17 assessment pre-registered this exact subtraction and named the open-PR budget as the only thing holding it.
2. **Magnitude — targeted narrowing.** One paragraph split in two; no new mechanism, no new step, nothing for a session to re-derive. The added text removes an option rather than adding a procedure.
3. **Cost — waste-class, and the fix is the shape Gate 3 admits.** No occurrence left a wrong outward action; the observed cost is a wasted successor boot. The justification is not net spend — it is that the successor boots regardless, so the choice is only whether it does the work or discovers the work is done. The budget blocker has cleared: the bot's open-PR queue drained 6 → 2 in this window.

### Verification

`uvx pre-commit run --files plugins/tend-ci-runner/skills/review/SKILL.md` — all applicable hooks pass (typos, trailing whitespace, the plugin-skill bang-backtick guard, install-tend reference sync). No other file in `plugins/` or `.claude/` references the renamed **Bot PRs** heading.

</details>


tend-agent (commented):
One residual instance of the term this PR splits. Step 6's CI-failure branch — the bullet ending `apply the fix per step 8 so the PR can go green, since no author will act on the offer` — still says `On **bot PRs** (Dependabot, renovate, etc.)`, the undifferentiated heading step 8 just renamed. That's the path where the push decision actually gets made on a red PR, and a session reading it on a self-authored PR gets the pre-split answer: this is a bot PR, so apply the fix. Following its pointer to step 8 now lands on the new paragraph and self-corrects, but the PR's own premise is that the ambiguous phrasing produced opposite reads on the same PR ninety seconds apart — one instance left in place keeps that open. Matching the rename (`On **third-party bot PRs** (Dependabot, renovate, etc.)`) closes it, and the self-authored case is already covered by the same pointer.

The rest checks out against the workflow: the `verify` job's `pull_request_review` branch is exactly `PR_AUTHOR = BOT_NAME` plus a non-empty body or a fresh inline comment, with no decision point; the review-dispatch prompt does tell that session to action a review it authored ("your reviewer role speaking, not a self-loop, so action it"); and a self-pushed commit does get stamped by its own session per step 9. No other file in the repo carries the renamed heading.

I'm stopping at the review rather than pushing the one-word fix — under the rule this PR adds, the mention run this review dispatches boots as the author and can apply it.

tend-agent:
Applied in `897f2b1`. Step 6's CI-failure bullet now reads `On **third-party bot PRs** (Dependabot, renovate, etc.)`, matching the step 8 heading.

I added one clause beyond the rename. With only the rename, a session on a red self-authored PR matches neither named case in that bullet — `human-authored` no, `third-party bot` no — and the pointer to step 8 lives inside the branch it no longer matches, so the guidance it needs is one hop away with nothing pointing at it. That's the same open-to-interpretation shape the PR exists to close, so the bullet now names the third case and hands it to step 8 explicitly: `On PRs this bot authored, step 8's rule holds: the follow-up COMMENT review dispatches the author session, which applies the fix.`

<details><summary>Scope check and verification</summary>

**Other instances.** `grep -in "bot PR\|bot-authored\|bot authored" plugins/tend-ci-runner/skills/review/SKILL.md` returns four lines; two are the step 8 headings this PR already split, one is line 374 (fixed here). The fourth, [line 275](https://github.com/max-sixty/tend/blob/897f2b1e0783afe0a24406039d85da42381d30fd/plugins/tend-ci-runner/skills/review/SKILL.md#L275), is left alone deliberately — `On a bot PR where you intend to push the fix yourself (step 8), post that COMMENT before pushing` is conditional on a push decision made elsewhere, so it orders two actions rather than authorizing one. Narrowing it would be inert at best and would wrongly exclude the third-party case at worst.

`plugins/tend-ci-runner/skills/nightly/SKILL.md` uses "bot PRs" twice, but for conflict resolution across all bot-authored PRs — a different skill and a different meaning, not a push decision. Unchanged.

**Verification.** `uvx pre-commit run --files plugins/tend-ci-runner/skills/review/SKILL.md` — all applicable hooks pass (trailing whitespace, end-of-file, typos, the plugin-skill bang-backtick guard, install-tend reference sync). `uv run pytest -k "skill or plugin"` — 9 passed.

</details>

PR #1010, opened 2026-08-19
feat(install-tend): gather install preferences in one kickoff round

max-sixty:
An all-defaults install now costs one dialog. The Kickoff gathers every preference in a single `AskUserQuestion` round, and the rest of the install stops only where a step genuinely needs the user — bot-account signup, OAuth approval, the device-code approval, the push.

Installing on a repo previously took eight questions across three rounds: harness, bot name, "Ready to start?", workflow overrides, overlay contents, badge yes/no, bio stance, and the harness auth mode — each at its own step, so an agent following the skill literally interrupts even more often than one that batches them. Every one of those had a recommended option that was taken.

The round asks three questions: **Harness** (Claude/OAuth, Claude/API key, or Codex — folding in what step 7a used to ask separately), **Bot name** (availability-checked candidates, recommended first), and **Customize**, a multiselect of the four areas that have safe defaults, each option naming its default. Submitting it with nothing selected accepts every default. Answering the round is the go-ahead, so the separate "Ready to start?" confirmation is gone; a request that already specifies everything asks nothing at all.

Steps 1, 4, 5 and 10 apply their defaults silently unless their area was picked, and 7a now routes the OAuth mint path (bundled wrapper vs. paste-it-back) by checking the environment rather than asking. The badge changes from "offer to add" to added-by-default with opt-out — the one behavior change here rather than a re-arrangement of who asks what.

<details>
<summary>Supporting fixes the restructure forced, or that review surfaced</summary>

- Kickoff branches on install **completeness** rather than config presence. `.config/tend.yaml` is written in step 1 of 11, so an install interrupted anywhere after that used to resume down the "this is a change" path, which asks nothing; it now resumes through the round with the questions the config already answers dropped.
- 7a asks the auth mode lazily when no round settled it. The config records the harness, never which Claude credential, so a change flow or a resumed install reaches the mint step without an answer.
- Both nested customize multiselects listed five options against `AskUserQuestion`'s hard cap of four, so the user who opted into customizing was the one who hit a tool error. Merged to four each.
- `REPO` derivation resolves a fork clone to its root source using the generator's own recipe, and confirms with the user when that changes the name or comes back null — a maintained hard fork keeps its own name.
- Step 6 asks which email the bot account should use, step 11 states that merging is the user's call, and step 3 loses a stale `OWNER`/`REPO` placeholder and a reference to a checkbox that doesn't exist at that point.

</details>

Reviewed by an adversarial pass over the restructured flow (which verified the `AskUserQuestion` limits against the installed binary rather than from memory) and a blind walkthrough that traced an all-defaults install and a resume-after-interrupt. A design challenge weighed three alternative shapes — zero questions with the name defaulted too, a config-file-first flow, and deferring every ask to the end — and none dominates. The round is the cheapest checkpoint that still lands before the first irreversible action on the normal path; a request that specifies everything skips it by design, and reaches step 3's ruleset POST without a further prompt.


plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -29,47 +24,94 @@ a change to one, and where it exists it settles the harness: the `harness` key,
...
-override the variable with the canonical `owner/name` and continue;
-no need to touch git remotes.
+When the resolution changes the name, or returns `null` (source deleted or
+invisible to the token), confirm the target with the user before touching
+anything — a deliberately maintained hard fork keeps its own name.
+
+A config from a finished install — workflows committed, secrets set, bot
+access granted (`uvx tend@latest check` reports all three) — makes this a
+change: take the harness from the config, lay
+out only the steps the task touches, and start. The summary checklist at the
```
tend-agent: `tend check` doesn't cover the workflow files — see the review body. Scoping the claim to what it does check keeps the gate honest (and unwraps the `lay`/`out` break):

```suggestion
A config from a finished install — `tend-*.yaml` committed on the default
branch, plus secrets set and bot access granted (`uvx tend@latest check`
covers the latter two, not the workflow files) — makes this a change: take
the harness from the config, lay out only the steps the task touches, and start.
```

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -29,47 +24,94 @@ a change to one, and where it exists it settles the harness: the `harness` key,
...
+   - **Claude — API key** — a console.anthropic.com key, billed per token.
+     Fits when there's no subscription to draw on, or the user wants a
+     dedicated billing surface and per-key revocation.
+   - **Codex — OpenAI API key** — pay-per-token. The `auth.json`
+     subscription path is incompatible with tend's concurrent workflows
+     (per-call refresh-token invalidation) and is being removed. Detail in
+     ${CLAUDE_SKILL_DIR}/references/security-model.md.
+2. **Bot name** — the available candidates, recommended first. "Other"
+   takes a custom name; check its availability before using it. If all
+   candidates are taken, generate more before asking.
```
tend-agent: The options array is `.min(2).max(4)`, so the floor bites too: with two of the three candidates taken, this question ships one option and `AskUserQuestion` rejects the call — the same failure the four-option cap fix avoided, from the other end.

```suggestion
2. **Bot name** — the available candidates, recommended first. "Other"
   takes a custom name; check its availability before using it. The tool
   needs 2–4 options, so generate more candidates whenever fewer than two
   of the three come back available.
```

tend-agent (commented):
Two things in the Kickoff, both inline.

The completeness gate cites `uvx tend@latest check` as reporting workflows-committed, and it doesn't — `run_all_checks` in [`generator/src/tend/checks.py`](https://github.com/max-sixty/tend/blob/d28a3b2afde5b243a0afcb0233dea85bbad2819a/generator/src/tend/checks.py#L1694-L1709) runs branch-protection, bot-permission, environment, environment-deployments, credential-environments, secrets, harness-auth and repo-secret-allowlist, and nothing asserts `tend-*.yaml` exists. The one workflow-reading check, `check_environment_deployments`, returns `True` when the sweep finds no offenders — including when there are no tend workflows at all. That matters for exactly the case this branch is trying to catch: step 11 commits without pushing, so an install that stopped at "committed, awaiting the user's go-ahead" has every secret set, the bot a collaborator, the rulesets in place, and no `tend-*.yaml` on the default branch. `tend check` is green, the gate reads "finished install", and the resume routes down the change path — which lays out only the steps the task touches and never regenerates.

Separately, `AskUserQuestion`'s options array is `.min(2).max(4)`, not just capped at 4. The PR fixed the overflow side on both nested multiselects; question 2 can still underflow, since only the all-taken case is handled.

One observation rather than a finding: the PR body's design note says this is "the cheapest checkpoint that still lands before the first irreversible action", but the fully-specified-request path asks nothing, and the first thing that mutates the repo is step 3's ruleset POST. On a repo with no release secrets to migrate, nothing between "install tend on this repo as `foo-bot` with Codex" and a created ruleset involves the user. That may well be the intent — the user did ask for an install — but it's a different property than the one the note claims.

<details><summary>Checked and fine</summary>

- The `REPO` recipe matches `detect_canonical_owner` in `checks.py` (`.fork` → `.source`, root not parent), and `jq` indexing a missing `.source` yields `null` rather than erroring, so the stated null branch is reachable as described.
- Option counts elsewhere: Kickoff q1 is 3, q3 is 4 (3 with no README), step 1's override list is 4, step 4's overlay list is 4. All inside 2–4.
- Dropping q1 for a config-present resume is consistent with 7a's lazy auth-mode ask, since the config records the harness but not the credential.
- No dangling references to the removed "Ready to start?" step, and the README line matches step 5's new default-on behavior.

</details>

tend-agent (approved):

PR #1011, opened 2026-08-19
fix: name the real cause when an OAuth run ends empty, and keep keys out of chat

max-sixty:
Found while installing tend on another repo: the OAuth step failed four times running, and the wrapper's message pointed somewhere else every time.

The wrapper answered every empty run with the same text, telling the caller to `rerun with --code-file` — which the skill's own invocation always passes. So the diagnosis named the one thing that was not wrong, and four consecutive fifteen-minute windows read as a flag problem rather than as nobody having approved. `failure_message` now separates the ways a run ends empty: a code the authorize page refused, a child that ended before offering an authorize URL at all, a window nobody approved in, and a run with nowhere to receive a code. Each is pinned by a test.

That second case came out of review here. The read loop breaks on the deadline, on a pty the child closed, and on a child already exited, so reaching the message never implied the window had passed. A `claude` that dies early — no `setup-token` subcommand, a crash — got "no token after 900s … the authorize URL went unapproved" in a fraction of a second, naming a window that had not elapsed and a URL that was never printed, while the child's own error stayed swallowed. Same shape as the bug this PR set out to fix.

Separately, the manual OAuth path and both API-key paths had the user paste a live credential to the agent, which then passed it along on a command line. Given neither `--body` nor a pipe, `gh secret set` prompts for the value, so the user runs it themselves and the token goes from `claude setup-token` to that prompt and nowhere else. The CLI path already had this right — "Piping straight into `gh` keeps the token out of the transcript" — and the fallbacks now match it. That prompt refuses an empty submission, tested through a pty against `--no-store` with a non-empty control, so the empty-value hazard that makes the CLI path's guard load-bearing has no counterpart there.

The skill also says what an OAuth window costs now. The wrapper waits fifteen minutes, and the authorize URL logs the browser out on the way in (`claude.ai/login?reauth=1&from=logout`), so approving takes the user at the browser throughout, and an agent driving Chrome reaches that login and stops.

**Testing**: the `failure_message` branches are covered. The surrounding pty loop in `main()` is not, as before — it needs a real child process and a terminal.

### DIFF
--- a/plugins/install-tend/skills/install-tend/scripts/oauth_token.py
+++ b/plugins/install-tend/skills/install-tend/scripts/oauth_token.py
@@ -89,6 +89,35 @@ def take_controlling_tty():
     fcntl.ioctl(0, termios.TIOCSCTTY, 0)
 
 
+def failure_message(code_file, typed_at):
+    """Why no token came back, in terms of what this run actually saw.
+
+    A caller that already passed `--code-file` is told to pass it, and reads
+    that as the diagnosis, so the three ways a run ends empty are named apart:
+    a code that the page refused, a window nobody approved in, and a run with
+    nowhere to receive a code.
+    """
+    if typed_at is not None:
+        return (
+            f"Error: the code from {code_file} was typed into the prompt and no "
+            "token came back, so the authorize page rejected it. A code belongs "
+            "to the run whose challenge issued it and dies with that run; "
+            "approve the URL this run printed."
+        )
+    if code_file:
+        return (
+            f"Error: no token after {TIMEOUT_SECONDS}s. Nothing was written to "
+            f"{code_file} and the browser never came back to the CLI, so the "
+            "authorize URL went unapproved. Approving takes a person at the "
+            "browser while this runs, and a rerun issues a fresh URL that "
+            "replaces this one."
+        )
+    return (
+        "Error: no sk-ant-oat01-… token in TUI output. If the browser showed "
+        "a code to copy, rerun with --code-file and write it to that path."
+    )
+
+
 def main():
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--code-file", help="path to watch for a `code#state` string")
@@ -185,10 +214,7 @@ def main():
             token = final.group(0).decode()
 
     if not token:
-        sys.exit(
-            "Error: no sk-ant-oat01-… token in TUI output. If the browser showed "
-            "a code to copy, rerun with --code-file and write it to that path."
-        )
+        sys.exit(failure_message(args.code_file, typed_at))
     if len(token) > MAX_TOKEN_LENGTH:
         sys.exit(f"Error: extracted token has implausible length ({len(token)} chars)")
     print(token)
--- a/plugins/install-tend/skills/install-tend/scripts/oauth_token.py
+++ b/plugins/install-tend/skills/install-tend/scripts/oauth_token.py
@@ -89,13 +89,16 @@ def take_controlling_tty():
     fcntl.ioctl(0, termios.TIOCSCTTY, 0)
 
 
-def failure_message(code_file, typed_at):
+def failure_message(code_file, typed_at, announced):
     """Why no token came back, in terms of what this run actually saw.
 
     A caller that already passed `--code-file` is told to pass it, and reads
-    that as the diagnosis, so the three ways a run ends empty are named apart:
-    a code that the page refused, a window nobody approved in, and a run with
-    nowhere to receive a code.
+    that as the diagnosis, so the causes it can act on are named apart. The
+    loop ends three ways — the deadline, a pty the child closed, and a child
+    already exited — so a message may not assume the window passed either. A
+    run that never announced a URL ended inside `claude setup-token`, whose own
+    output this wrapper swallows, and is the one case where the fix is to go
+    look at that command rather than at the browser.
     """
     if typed_at is not None:
         return (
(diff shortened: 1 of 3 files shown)


plugins/install-tend/skills/install-tend/scripts/oauth_token.py
```diff
@@ -89,6 +89,35 @@ def take_controlling_tty():
...
+    if typed_at is not None:
+        return (
+            f"Error: the code from {code_file} was typed into the prompt and no "
+            "token came back, so the authorize page rejected it. A code belongs "
+            "to the run whose challenge issued it and dies with that run; "
+            "approve the URL this run printed."
+        )
+    if code_file:
+        return (
+            f"Error: no token after {TIMEOUT_SECONDS}s. Nothing was written to "
```
tend-agent: This branch fires on an early exit too, not just on the deadline — the loop breaks at `if not chunk: break` (child closed the pty) and at `if child.poll() is not None and not readable: break`:

````bash
mkdir -p /tmp/fakebin
printf '#!/bin/sh\necho "error: unknown command \\"setup-token\\"" >&2\nexit 1\n' > /tmp/fakebin/claude
chmod +x /tmp/fakebin/claude
time PATH=/tmp/fakebin:$PATH ./oauth_token.py --code-file /tmp/tend-oauth-code
# Error: no token after 900s. Nothing was written to /tmp/tend-oauth-code and the
# browser never came back to the CLI, so the authorize URL went unapproved. …
# real  0m0.038s
````

Three claims in one message that the run didn't see: 900s that didn't pass, a file whose absence wasn't the cause, and an authorize URL that was never printed. The child's `unknown command` is swallowed (the module docstring's "Nothing else from the TUI is echoed"), so the caller has nothing else to go on and reruns into the same wall — which is the loop this PR is closing.

Passing what the run actually saw separates it:

````python
def failure_message(code_file, typed_at, waited, announced):
    ...
    if not announced:
        return (
            f"Error: `claude setup-token` exited after {waited:.0f}s without printing an "
            "authorize URL, so this run never reached the browser. Run `claude setup-token` "
            "directly to see its own error."
        )
    if code_file:
        return f"Error: no token after {waited:.0f}s. …"
````

Happy to push it — function, the `main()` call site, and a test for the early-exit branch — if you'd rather not.

**Review by tend-agent — COMMENTED** [review]
One case the new message set doesn't cover, plus a note on the guard the manual path drops.

The `--code-file` branch is reachable without the 15-minute window ever elapsing: the read loop also breaks when the child closes the pty (`if not chunk: break`) and when it has already exited (`child.poll() is not None and not readable`). With a `claude` that exits on its own — an older CLI without `setup-token`, a crash, a browser approval the user denied — the run ends in ~0.04s and prints `no token after 900s … so the authorize URL went unapproved`, naming a window that didn't pass and a URL that was never printed. Since nothing from the TUI is echoed, the child's own error never reaches the caller, so that message is the only signal the caller gets. Same shape as the bug this PR fixes. Repro and a sketch inline.

Secondary, on the skill: handing the user a bare `gh secret set` is right for the token, but it drops the check the CLI path calls load-bearing a few lines up. With the agent no longer holding a value, nothing between the user's paste and step 7's own pre-check confirms one arrived — and that pre-check, like `check_secrets`, reads names only, so an empty value submitted at the prompt leaves a `CLAUDE_CODE_OAUTH_TOKEN` that every later check reads as set. `gh secret set` prints `✓ Set …` either way. I haven't tested whether gh's prompt accepts an empty submission the way empty stdin does (the skill documents the stdin case at the CLI path); if it does, `read -rs TOKEN && [ -n "$TOKEN" ] && printf '%s' "$TOKEN" | gh secret set …` keeps the token out of the transcript and the guard intact, in one command the user still runs themselves.

File: plugins/install-tend/skills/install-tend/scripts/oauth_token.py (hunk 1 above)
Comment (max-sixty) [reply plugins/install-tend/skills/install-tend/scripts/oauth_token.py]: Fixed in d24a08c — you were right, and it was the same bug this PR set out to fix. Your repro now returns in about two seconds with:

```
Error: `claude setup-token` ended without offering an authorize URL, so it failed before the browser was ever involved. Its output is not echoed here — the TUI carries the token — so run that command directly to see what it said.
```

`failure_message` takes `announced` now, and the no-URL case is checked ahead of anything that mentions the browser, so neither the window nor the authorize URL is asserted on a path where neither happened. A test pins it by asserting that message carries neither "unapproved" nor the timeout constant.

max-sixty:
On the manual path's dropped guard: I tested the part you flagged as untested, and gh's prompt refuses an empty submission, so the guard has no work to do there.

Driving `gh secret set TEND_PROMPT_PROBE --no-store` through a pty, an empty Enter left the prompt sitting untouched. A non-empty value through the same harness was masked, encrypted, printed and returned to the shell — so the empty case is the prompt refusing it, not the harness failing to drive it. `--no-store` means nothing reached the repo either way.

So the bare `gh secret set` stays. A `read -rs` guard in front of a prompt that already enforces non-empty is a second mechanism for a guarantee that already holds, and it costs the masking and the paste affordance. I noted the asymmetry in the skill instead, so the next reader doesn't have to retest it: the empty-value hazard belongs to the piped form, which is why the CLI path's guard is load-bearing and this one isn't.

tend-agent:
Both closed. The pty run answers the question I left open, and recording the asymmetry in the skill rather than adding the guard is the part that keeps the next reader from retesting it.

One check on the new branch, since it introduces a flag a message now turns on: it can't misfire the other way — reporting a run that *did* print a URL as one that never offered one. `announced` is set in the same block that prints the URL to stderr, so the claim in the message and what the caller actually saw can't diverge. And the early-exit case that survives the fix, a child that dies after the URL was surfaced (a denied approval), lands on "went unapproved", which is accurate for it; the reworded "waits up to 900s" no longer asserts the window elapsed.

tend-agent (approved):

PR #1012, opened 2026-08-19
fix(install-tend): make all-defaults one click and the OAuth mint recipe runnable

max-sixty:
Follow-ups #4/#5/#6 from #1010, none of which had been tested.

- **Kickoff question 3 becomes an accept-defaults/customize gate.** The all-defaults path answers one dialog of three single-selects, so the empty-multiselect submission — and Claude Code's "You have not answered all questions" banner it trips — can't occur. The four-area multiselect fires only on Customize…, where selection can only mean "change this area" (the old labels read as include/exclude toggles). Both surfaces list only areas still open: a request-settled area follows the request at its step, an already-applied area stays put, and the default bio is quoted rather than called "recommended" sight-unseen.
- **Step 7a's CLI mint path names its mechanism** instead of a bare "run it in the background" over a blocking `TOKEN=$(...)` snippet: launch the pipeline via the Bash tool's `run_in_background` (a foreground call traps the authorize URL in its pending result and times out before the user has anything to click), hand over the URL from the task output, and read the task's exit status as the sole success signal. Validated against a live `claude setup-token` run: the URL lands in the task output within seconds, and a failed run exits non-zero having stored nothing.
- **The overlay the skill writes gains frontmatter.** The `running-tend` overlay is loaded by skill discovery, which needs `name`/`description` frontmatter the placeholder never had — every adopter shipped an overlay the CI sessions never loaded. The requirement now covers the customized-overlay path too, and an existing overlay without the frontmatter (PRQL and numbagg ship exactly that shape) is repaired in place rather than skipped as already done.
- **Follow-through on #1011**, which rewrote the same 7a section and merged mid-branch: its start-when-at-browser gate moved ahead of the launch instruction (where a top-to-bottom reader needs it), and four inherited gaps closed — a completion signal for the user-runs-`gh secret set` paths (the step's pre-check printing SET), `gh` logged in as the maintainer on the Manual machine, step 1's registry Manual route keeping the token out of the chat transcript like 7a now does (with the repo-level delete gated on the secret appearing in the environment), and a duplicated URL in 7b.

Reviewed by an adversarial pass and a context-blind walkthrough that traced five scenarios (all-defaults, customize, no-README, resume, the OAuth mint executed literally); all findings fixed and re-verified, including a post-merge pass over the composed 7a.

**Testing**: the wrapper's unit tests pass; the SKILL.md changes are prose, covered by the blind walkthrough rather than code tests.

### DIFF
--- a/plugins/install-tend/skills/install-tend/SKILL.md
+++ b/plugins/install-tend/skills/install-tend/SKILL.md
@@ -261,7 +261,10 @@ place. Classify each remaining secret and act now — don't defer:
     at token-creation time; the user completes it in the open tab, per
     Browser sessions.
   - **Manual** — user generates the token themselves on the registry's
-    site and pastes it back.
+    site and stores it themselves: hand over the environment's
+    `gh secret set` command fully substituted. With neither `--body` nor
+    a pipe it prompts for the value, so the token never sits in the chat
+    transcript.
 
   Whichever route is chosen, include the exact token-creation URL in
   the question or option description (and in the follow-up message if
@@ -627,6 +630,10 @@ install) can't be read back — GitHub secrets are write-only — so mint the
 value into the environment per the steps below, then delete the
 repo-level copy; `tend check` flags it until deleted.
 
+Where a path below has the user run `gh secret set` themselves, the step
+finishes when its pre-check prints SET — re-run it once they say they're
+done.
+
 Branch on the harness.
 
 ### 7a. Harness = claude
@@ -702,8 +709,9 @@ as 1-year), two mint paths, routed by environment rather than asked:
 - **Manual** — when the CLI path is unavailable, the wrapper errors out,
   or the user isn't at the browser when the agent is. Hand over both
   commands, fully substituted, for them to run in their own terminal (any
-  machine with Claude Code installed; `https://claude.com/claude-code` to
-  install it), whenever suits them:
+  machine with Claude Code installed and `gh` logged in as the
+  maintainer; `https://claude.com/claude-code` to install it), whenever
+  suits them:
 
   ```bash
   claude setup-token
@@ -731,8 +739,7 @@ gh secret set ANTHROPIC_API_KEY --repo "$REPO" --env tend
 
 ### 7b. Harness = codex
 
-Codex uses `OPENAI_API_KEY` (pay-per-token, from
-`https://platform.openai.com/api-keys`). The subscription `auth.json`
+Codex uses `OPENAI_API_KEY` (pay-per-token). The subscription `auth.json`
 path is not supported — Codex rotates that refresh token on every
 API call and invalidates the prior one, so tend's concurrent
 workflows (review/mention/triage/nightly/…) would break each other's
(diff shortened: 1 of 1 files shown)


tend-agent (commented):
Two gaps about how far each fix reaches, not about the changes themselves — the diff's own claims hold up. The 7a exit-status claim in particular is exactly right: an empty `TOKEN` leaves `[ -n "$TOKEN" ]` as the last command and the script exits 1, a non-empty one hands the status straight to `gh secret set`, so 0-iff-stored is true with no `set -e` needed. The wrapper backs the rest — `TIMEOUT_SECONDS = 900`, the authorize URL goes to stderr as soon as the TUI offers it (so it shows up in the background task's output, not the captured stdout), and a stale `--code-file` is unlinked at startup.

**The overlay frontmatter fix only reaches new installs.** Of the six adopter repos in `data/consumers.json` that ship a `running-tend` overlay, two — [PRQL/prql](https://github.com/PRQL/prql/blob/main/.claude/skills/running-tend/SKILL.md) and [numbagg/numbagg](https://github.com/numbagg/numbagg/blob/main/.claude/skills/running-tend/SKILL.md) — open with an `# Running Tend…` heading and no frontmatter, so their CI sessions still never load them. Nothing repairs those: the Kickoff's "Skip steps that are already done — check each prerequisite before acting" makes step 4 a no-op on a re-run once the file exists, and `tend check` doesn't look at the overlay. One sentence in step 4's preamble would close it — an existing overlay counts as done only if it opens with that frontmatter; one without it gets the frontmatter added in place. Happy to push that commit if you want it in this PR rather than a follow-up.

**The completion signal added at step 7 doesn't cover the new step-3 Manual route.** "Where a path *below* has the user run `gh secret set` themselves…" is scoped to 7a/7b, but this PR turns step 3's registry Manual route into exactly that shape — the user now stores the token themselves instead of pasting it back. Step 3 pairs that with a destructive follow-on ("recreate it on the Environment, delete the repo-level copy"), and its Verify block checks `deployment-branch-policies`, never that the secret exists. So on this route the agent can confirm-and-delete the repo-level `PYPI_TOKEN` before the user has run their command, and the release workflow is broken until they do — the one route where the migration can drop the credential on the floor. Inline suggestion below.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -248,7 +261,10 @@ place. Classify each remaining secret and act now — don't defer:
     at token-creation time; the user completes it in the open tab, per
     Browser sessions.
   - **Manual** — user generates the token themselves on the registry's
-    site and pastes it back.
+    site and stores it themselves: hand over the environment's
+    `gh secret set` command fully substituted. With neither `--body` nor
+    a pipe it prompts for the value, so the token never sits in the chat
+    transcript.
```
```suggestion
    transcript. Don't delete the repo-level copy until
    `gh secret list --repo "$REPO" --env "$ENV" --json name` shows it —
    the write is theirs on this route, so nothing else tells the agent it
    landed.
```

max-sixty: Applied in 780eafa, matched to the bullet's wording. The overlay retrofit (existing frontmatter-less overlays repaired in place) is in the same commit.

tend-agent (approved):

PR #1014, opened 2026-08-19
skills: fetch the prior rejection before re-deriving a fix

### LINKED ISSUE #1013: nightly: Step 8 dedup reads open PRs and titles only, so a PR closed on the approach doesn't stop a repeat attempt
Nightly opened a third PR for a fix that had already been rejected twice, because the dedup that guards Step 8 only looks at **open** PRs and only at titles. The two artifacts that said "don't fix this" — a closed PR and an issue comment — are both outside what the check reads. Filed at the PRQL maintainer's request after the third attempt was closed ([PRQL/prql#6206](https://github.com/PRQL/prql/pull/6206)); PRQL runs tend 0.1.18.

## What happened

| when | what |
| --- | --- |
| 2026-08-06 | [PRQL/prql#6147](https://github.com/PRQL/prql/pull/6147) fixes duplicate `module` declarations by merging them. Closed by a maintainer on the **approach**, with the code reviewed as fine: _"I've reviewed the code itself and it's fine; the behavior change is the part that I'm unsure about."_ |
| 2026-08-08 | A nightly run files [PRQL/prql#6166](https://github.com/PRQL/prql/issues/6166) for the same bug and comments on it: _"Deliberately **not** opening a fix PR, because the direction here is already an open question in front of @kgutwin from #6147 rather than an unclaimed bug."_ That note exists to stop the next run. |
| 2026-08-18 | Nightly run [`32109908394`](https://github.com/PRQL/prql/actions/runs/32109908394) surveys the same file, finds the same bug, and opens [PRQL/prql#6206](https://github.com/PRQL/prql/pull/6206) — the same merging approach, third attempt. |
| 2026-08-19 | Closed by a maintainer: _"This is @prql-bot's third repeat at this issue. #6166 specifically mentions that the open tracking issue was created to avoid opening further PRs."_ |

## Why the guard didn't fire

Step 8 of `plugins/tend-ci-runner/skills/nightly/SKILL.md` opens with:

```bash
gh issue list --state open --limit 200 --json number,title
gh pr list --state open --limit 200 --json number,title,headRefName
```

Three things that block are all outside that projection:

1. **`--state open` excludes the one state that carries the rejection.** A closed PR on the same call site is the strongest available signal that the change was already considered and turned down — and closed is precisely the state filtered out. #6147 was invisible to the run that opened #6206.
2. **`number,title` carries no bodies or comments.** Issue #6166's title reads as an ordinary unclaimed bug (`Duplicate module definitions silently discard the earlier declaration`); the hold is in its comments. A previous run wrote that note specifically so the next run would read it, and the projection guaranteed it wouldn't.
3. **Matching is by title/number, never by the code the fix touches.** All three PRs edit the same function, `fold_module_def_stmt`. `gh pr list --state all --search "6166"` returns #6206 and #6164 but **not** #6147 — the rejected attempt predates the issue and cites no issue number. `--search "fold_module_def_stmt"` returns all three.

The pre-`gh pr create` dedup in `running-in-ci` doesn't cover this either. It does use `--state all --author "$BOT_LOGIN"`, so #6147 is inside the returned list — but the section is framed entirely around **concurrent siblings** ("A separate mention on a different issue/PR can trigger a concurrent run asking for the same fix"), and its stop condition is scope overlap. A two-week-old closed PR reads as history, not as a sibling, and nothing in the rule says to open it and find out *why* it closed. That distinction is the whole decision: a PR closed on the **code** leaves the fix available to redo properly; a PR closed on the **approach** means the semantics are an open maintainer decision, and any reimplementation of the same approach is noise however well written.

## Suggested direction

Design is yours; the three gaps above suggest:

- Step 8's PR dedup uses `--state all`, and searches by the symbol or file the fix would touch in addition to title keywords — the symbol is what survives when the issue number doesn't exist yet.
- A closed bot PR touching the same call site is a stop, not a data point: read its own comments (`gh pr view <n> --json comments,reviews`) before opening. Closed on the approach → post findings to the existing thread and leave the PR alone.
- When a finding matches an open issue, read the issue's comment **bodies**, not just its title. This is what makes a prior run's "deliberately not opening a PR" note actually reach the next run — otherwise each run re-derives the finding from scratch and the note is write-only.

This generalizes past nightly: `triage` and `ci-fix` open PRs from findings too, so a shared reference may fit better than a nightly-local rule.

<details><summary>The repo-local fix that was rejected, and why</summary>

The first attempt at this was [PRQL/prql#6212](https://github.com/PRQL/prql/pull/6212), which added a hand-maintained "parked issues" list to PRQL's `running-tend` skill. The PRQL maintainer rejected the shape: _"we would not store a list of issues in the repo itself. we have access to the issues, we can search them anytime! if there's a problem, and it's material, put an issue into tend"_ — hence this issue. A checked-in list also needs manual retirement once an issue closes, and only ever covers issues someone thought to add; the searches above cover the general case.

Verified this session: `gh pr list --state all --search "fold_module_def_stmt"` returns #6147, #6150, #6151, #6164, #6206, #6212; the same search on `6166` omits #6147. Quoted skill text is from tend-ci-runner as loaded by this run.

</details>

tend-agent:
Nightly's Step 8 dedup fetched only open PRs, and only their titles, so both places a maintainer's rejection is recorded — the closed PR that carried it, and the comments on the issue tracking it — sat outside what the check ever read. A run could re-derive a finding from the code and reopen an approach already turned down, with nothing in context to say otherwise. The searches now live in `running-in-ci` (loaded by nightly, triage, and ci-fix alike, before any `gh pr create`), matched on the symbol or path the change would edit rather than an issue number a re-derived finding doesn't have. Verified with `pre-commit run --files` on both changed files.

**This is a query fix, not a judgment rule** — worth stating up front, because #805 proposed the behavioral version of this and you closed it: _"My sense is that the models would see this principle as implicit and not require explicit guidance."_ I think that was right, and it doesn't cover this case. There, the dedup had already surfaced the rejection and the agent acted on it correctly without being told to; the rule was belt-and-suspenders. Here the rejection is never fetched at all, and no amount of inference recovers a fact that isn't in context. So what's added is three `gh` invocations and one sentence on what the result means, not a restatement of a principle the model already applies.

<details><summary>What #1013 reported, and why the existing checks miss it</summary>

Filed by PRQL's tend bot after a third repeat attempt at the same fix was closed ([PRQL/prql#6206](https://github.com/PRQL/prql/pull/6206)). The prior two artifacts both said "don't fix this": [PRQL/prql#6147](https://github.com/PRQL/prql/pull/6147), closed on the **approach** with the code reviewed as fine, and a note on [PRQL/prql#6166](https://github.com/PRQL/prql/issues/6166) written by an earlier nightly run specifically so the next run would stop.

Three gaps, all in the projection rather than the judgment:

1. `--state open` filters out the one state that carries a rejection. A closed PR on the same call site is the strongest available signal, and it's precisely what's excluded.
2. `--json number,title` carries no bodies or comments. The issue's title reads as an ordinary unclaimed bug; the hold is in its comments. A prior run's note was write-only by construction.
3. Matching by title or issue number never connects an attempt that predates the tracking issue. The reporter verified that `gh pr list --state all --search "fold_module_def_stmt"` returns all three attempts while `--search "6166"` omits the rejected one — hence searching by symbol.

The pre-`gh pr create` dedup in `running-in-ci` does use `--state all`, so the closed PR is inside the list it returns, but that section is framed entirely around concurrent siblings. A weeks-old closed PR reads as history there, and nothing said to open it and find out why it closed.

Per `/tend-ci-runner:triage`'s "Skill text fixes", I checked co-loaded skills first — no existing guidance anywhere in `plugins/` or `.claude/` covers reading a closed PR's rationale, so nothing is duplicated.

</details>

<details><summary>No regression test</summary>

Documentation-only change to two skill files; the repo has no test coverage over skill prose. The `generator/tests` hits on "skills" are regtest outputs asserting skill *names* in generated workflow YAML, unaffected here. `pre-commit run --files` on both files passes trailing-whitespace, end-of-file-fixer, typos, the bang-backtick poison guard, and the install-tend reference-sync hook.

</details>

---
Closes #1013 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -179,7 +179,7 @@ BOT_LOGIN=$(gh api user --jq '.login')
 # <symbol>: the function, file, or config key the change would edit
 gh pr list --state all --search "author:$BOT_LOGIN <symbol>" --limit 100 --json number,title,state,closedAt
 gh pr view <n> --json comments,reviews --jq '[.comments[].body, .reviews[].body]'
-gh issue view <n> --json comments --jq '.comments[].body'
+gh issue view <n> --json body,comments --jq '[.body, .comments[].body]'
 ```
 
 What you find governs: a PR closed on the **code** leaves the fix available to redo, while one closed on the **approach** means the semantics are still an open maintainer question — add findings to that thread rather than opening a second implementation of it.
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -172,10 +172,12 @@ Compare by title keywords **and** the files the new PR would modify — two conc
 
 A change a maintainer already turned down leaves its verdict in two places the checks above don't fetch: the closed PR that carried it, and the comments on the issue tracking it. Search by the symbol or path the change would edit — a finding re-derived from the code has no issue number, and an attempt predating the tracking issue cites none either — then read the closed hits and the issue bodies, not just their titles.
 
+Search, don't scan. A recency-ordered listing ages a rejection out in bot-throughput time: at a few PRs a day, any `--limit` drops it within weeks, and raising the cap only moves that boundary. A symbol match stays small however many PRs have landed since.
+
 ```bash
 BOT_LOGIN=$(gh api user --jq '.login')
 # <symbol>: the function, file, or config key the change would edit
-gh pr list --state all --search "author:$BOT_LOGIN <symbol>" --json number,title,state,closedAt
+gh pr list --state all --search "author:$BOT_LOGIN <symbol>" --limit 100 --json number,title,state,closedAt
 gh pr view <n> --json comments,reviews --jq '[.comments[].body, .reviews[].body]'
 gh issue view <n> --json comments --jq '.comments[].body'
 ```
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -168,6 +168,20 @@ gh pr list --state all --search "author:$BOT_LOGIN <issue-number>" \
 
 Compare by title keywords **and** the files the new PR would modify — two concurrent fixes for the same bug typically pick different branch names, so a branch-name match is not sufficient. If a sibling bot PR overlaps in scope — whether open, closed, or already merged — **do not create**: post a comment on the triggering thread linking the existing PR and exit.
 
+### Fetch the prior rejection before re-deriving a fix
+
+A change a maintainer already turned down leaves its verdict in two places the checks above don't fetch: the closed PR that carried it, and the comments on the issue tracking it. Search by the symbol or path the change would edit — a finding re-derived from the code has no issue number, and an attempt predating the tracking issue cites none either — then read the closed hits and the issue bodies, not just their titles.
+
+```bash
+BOT_LOGIN=$(gh api user --jq '.login')
+# <symbol>: the function, file, or config key the change would edit
+gh pr list --state all --search "author:$BOT_LOGIN <symbol>" --json number,title,state,closedAt
+gh pr view <n> --json comments,reviews --jq '[.comments[].body, .reviews[].body]'
+gh issue view <n> --json comments --jq '.comments[].body'
+```
+
+What you find governs: a PR closed on the **code** leaves the fix available to redo, while one closed on the **approach** means the semantics are still an open maintainer question — add findings to that thread rather than opening a second implementation of it.
+
 ## Pushing to PR Branches
 
 Always use `git push` without specifying a remote — `gh pr checkout` configures tracking to the correct remote, including for fork PRs. Specifying `origin` explicitly can push to the wrong place.
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -332,6 +332,8 @@ gh issue list --state open --limit 200 --json number,title
 gh pr list --state open --limit 200 --json number,title,headRefName
 ```
 
+That projection orients you; it does not clear a finding. It omits both states a prior rejection lives in — closed PRs, and the comment bodies of an open issue — so per finding, before writing code, run the searches under **Fetch the prior rejection before re-deriving a fix** in `/tend-ci-runner:running-in-ci`.
+
 The default action is a PR, not an issue. If there's a plausible fix, make it — explain uncertainty in the PR description.
 
 For each finding:


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -168,6 +168,20 @@ gh pr list --state all --search "author:$BOT_LOGIN <issue-number>" \
...
 Compare by title keywords **and** the files the new PR would modify — two concurrent fixes for the same bug typically pick different branch names, so a branch-name match is not sufficient. If a sibling bot PR overlaps in scope — whether open, closed, or already merged — **do not create**: post a comment on the triggering thread linking the existing PR and exit.
 
+### Fetch the prior rejection before re-deriving a fix
+
+A change a maintainer already turned down leaves its verdict in two places the checks above don't fetch: the closed PR that carried it, and the comments on the issue tracking it. Search by the symbol or path the change would edit — a finding re-derived from the code has no issue number, and an attempt predating the tracking issue cites none either — then read the closed hits and the issue bodies, not just their titles.
+
+```bash
+BOT_LOGIN=$(gh api user --jq '.login')
+# <symbol>: the function, file, or config key the change would edit
+gh pr list --state all --search "author:$BOT_LOGIN <symbol>" --json number,title,state,closedAt
```
tend-agent: `gh pr list` defaults to `--limit 30` (`-L, --limit int   Maximum number of items to fetch (default 30)`), so this search silently truncates at 30 — which is the exact failure mode #1013's follow-up comment identified as the mechanical cause: the dedup call ran with `--limit 30` and the rejected PR sat at index 30, one row past the cap. A symbol search is narrow enough that 30 usually suffices, but the recipe also offers `<symbol>` as "the function, file, or config key", and a *path* search in a busy repo clears 30 hits easily. The neighbouring recipe in this same section already passes `--limit 200`; matching it costs nothing and removes the one cap this section exists to stop trusting.

```suggestion
gh pr list --state all --search "author:$BOT_LOGIN <symbol>" --limit 200 --json number,title,state,closedAt
```

tend-agent (commented):
Two suggestions on the new recipe, both about the projection being wide enough to actually contain the rejection.

The framing holds up — this is a fetch gap, not a judgment rule, and the `--state all` + symbol search is the right shape. Skill-authoring rules are respected: the added text states the structural rule without citing the PRQL cases that produced it.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -168,6 +168,20 @@ gh pr list --state all --search "author:$BOT_LOGIN <issue-number>" \
...
+gh pr view <n> --json comments,reviews --jq '[.comments[].body, .reviews[].body]'
+gh issue view <n> --json comments --jq '.comments[].body'
```
tend-agent: The prose above says to read "the closed hits and the issue bodies", but this fetches only the comments — an issue whose *description* states the hold (a maintainer filing it as "tracking only, don't fix") reads as clean here. One field covers it.

```suggestion
gh issue view <n> --json body,comments --jq '[.body, .comments[].body]'
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Fixed in [`218d9b5`](https://github.com/max-sixty/tend/commit/218d9b5e2bc46ed7c4e257a57fc38d85659e3987) — the cap is now explicit, at 100 rather than 200, plus a line saying why the number isn't the real answer: a recency-ordered listing ages a rejection out at bot throughput, so any `--limit` drops it eventually and raising the cap only moves the boundary. The search term is what keeps the result set small; the explicit cap just stops the silent 30 you identified.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied in [`bc45f51`](https://github.com/max-sixty/tend/commit/bc45f514724724f6bb13fcd25fa654ebc335ce7b) — right, the recipe contradicted the prose above it. Now `--json body,comments`.

Issue #1008, opened 2026-08-18
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-08-18T16:50:55Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/32161653061) | #1005 |

This issue was created automatically. Close it once the outage is resolved.

tend-agent:
### [Run 32161653061](https://github.com/max-sixty/tend/actions/runs/32161653061)

#### handle

```
claude -p exited non-zero (exit=1): API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check https://status.claude.com. — see the session-logs artifact
```

tend-agent:
Resolved. The failure was a transient upstream `API Error: 529 Overloaded` on [run 32161653061](https://github.com/max-sixty/tend/actions/runs/32161653061), not a defect in this repo. Every tend run since has succeeded — including `tend-review` and `tend-mention` on #1005 (the PR that triggered the failed run) and the notifications sweep on a 15-minute cadence through the night.

Issue #946, opened 2026-08-11
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-08-11T15:11:08Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/31505514081) | N/A |

This issue was created automatically. Close it once the outage is resolved.

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-11T15:11:03Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/31505514081) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-11T15:11:26Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/31505514081) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-12T00:08:31Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/31548728916) | N/A |

tend-agent:
### [Run 31505514081](https://github.com/max-sixty/tend/actions/runs/31505514081)

#### review-reviewers (max-sixty/cargo-affected)

```
claude -p exited non-zero (exit=1) — see the session-logs artifact
```

#### review-reviewers (numbagg/numbagg)

```
claude -p exited non-zero (exit=1) — see the session-logs artifact
```

#### review-reviewers (max-sixty/tend)

```
claude -p exited non-zero (exit=1) — see the session-logs artifact
```

#### review-reviewers (max-sixty/worktrunk)

```
claude -p exited non-zero (exit=1) — see the session-logs artifact
```



### [Run 31548728916](https://github.com/max-sixty/tend/actions/runs/31548728916)

#### review-reviewers (numbagg/numbagg)

```
claude -p exited non-zero (exit=1) — see the session-logs artifact
```

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-08-12T06:54:34Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/31571476044) | #916 |

tend-agent:
### [Run 31571476044](https://github.com/max-sixty/tend/actions/runs/31571476044)

No failure details could be extracted.

tend-agent:
[...]

PR #1015, opened 2026-08-19
skills(ci-fix): filter the outage-escalation query by title, not label alone

tend-agent:
Most of this PR landed directly on `main` as acd9d32. Merging `main` back in leaves one change: the escalation query in `ci-fix` step 3a filters `tend-outage` issues by label and author but not by title, so it also matches the **"Bot temporarily unavailable"** issues that `shared/steps/report-failure.sh` files under the same label from the same account. This adds the title filter back and states why both filters are load-bearing.

`gh issue list` defaults to 30 results, so the outage issues don't merely dilute the result — they fill it, and the 7-day `select` downstream then counts zero prior occurrences however many real trackers exist. That silently disables the repeat-occurrence escalation the section exists to perform: an intermittent upstream regression keeps getting re-classified transient.

The two producers of `tend-outage` issues, both posting as the bot:

- `shared/steps/report-failure.sh` — `LABEL="tend-outage"`, `TITLE="Bot temporarily unavailable"` ([lines 29-30](https://github.com/max-sixty/tend/blob/acd9d32bbcb46b23c42817dc0d4922c3044142b5/shared/steps/report-failure.sh#L29-L30))
- `ci-fix` step 3a's own transient tracker — `ci-fix: transient failure on <run-id>`

Current counts in this repo, all-time, bot-authored, `tend-outage`:

```
46  Bot temporarily unavailable
 1  ci-fix: transient failure on 27215280690
```

At a 46:1 ratio the unfiltered first page is entirely outage rows.

</details>

No test — the change is skill prose. The claim it rests on is the label/title pair in `report-failure.sh`, cited above.


plugins/tend-ci-runner/skills/ci-fix/SKILL.md
```diff
@@ -101,7 +101,7 @@ gh issue list --state all --author "$BOT_LOGIN" --search "ci-fix: in:title" \
...
+If 2+ prior issues match the current failure shape within the past 7 days, escalate to durable: a fault that re-fires every 1–3 days is not transient even when individual reruns pass. Count incidents, not trackers — the same root cause taking down several jobs in one afternoon files several issues and is still one occurrence. Where the fault costs only runner time, the escalation is also subject to the evidence and remedy bar in **Weighing a Fix** (`running-in-ci`). Search for an upstream-documented workaround (`gh issue search` against the action's repo, the action's README, GitHub Community threads) and apply it. If no upstream workaround is documented, open a fix PR proposing a minimal mitigation (pin runner image, skip the affected leg, disable the relevant cache layer) and link the upstream tracking issue.
```
tend-agent: This tests the fault's cost, while `running-in-ci` tests the change's benefit. The triggering incident fails this test (the hung job left CI red) but passes `running-in-ci`'s; conversely any flake that reruns green passes this one, which is broad enough to blunt the escalation this section exists for. Matching the other file's framing resolves both directions:

```suggestion
If 2+ prior issues match the current failure shape within the past 7 days, escalate to durable: a fault that re-fires every 1–3 days is not transient even when individual reruns pass. Count incidents, not trackers — the same root cause taking down several jobs in one afternoon files several issues and is still one occurrence. Where the mitigation's only benefit is runner time — the failure reruns green, so nothing about it makes a build correct or a red branch durably green — the escalation is also subject to the evidence and remedy bar in **Weighing a Fix** (`running-in-ci`). Search for an upstream-documented workaround (`gh issue search` against the action's repo, the action's README, GitHub Community threads) and apply it. If no upstream workaround is documented, open a fix PR proposing a minimal mitigation (pin runner image, skip the affected leg, disable the relevant cache layer) and link the upstream tracking issue.
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -105,6 +105,13 @@ Your closing summary is the session's only durable record of what happened, and
...
 
+The repo's own CI runner time is the same category, not a separate one: a slow job, a burned runner-hour, a step that hung until the platform killed it. GitHub absorbs the minutes; the maintainer absorbs the machinery forever. So a change whose only benefit is runner time — nothing about it makes a build correct, or a red branch green — clears a much higher bar than a correctness fix, on two counts:
+
+- **Evidence.** One incident is not a pattern. Repeats inside a single upstream incident window — several jobs stalling on the same mirror the same afternoon, the same leg failing twice on the PR that is fixing it — are one occurrence, not three. What justifies a permanent change is a fault that keeps returning across days, after the incident that produced it closed.
+- **Remedy.** Cap it at a one-line knob: a `timeout-minutes:`, a deleted step, a pinned image. A new helper script, a retry wrapper, or a bounded-install shim threaded through several call sites is the shape to decline — when writing one, and when reviewing one. If no knob is enough, say what the waste costs on the thread and leave the decision to the maintainer.
```
tend-agent: The closing sentence restates the paragraph five lines up verbatim in substance ("otherwise note the cost where the maintainer will see it and move on"), and "when writing one, and when reviewing one" is what the paragraph two lines down says at length. CLAUDE.md's skill-authoring rule is "Be brief. Skills are loaded into every relevant session — extra prose is overhead", and this file loads in every session; the CI-flavoured examples are the part that isn't already here.

```suggestion
- **Remedy.** The same knob bar in CI terms: a `timeout-minutes:`, a deleted step, a pinned image. A new helper script, a retry wrapper, or a shim threaded through several call sites is the shape to decline.
```

tend-agent (commented):
Two wording points, both on the new guidance rather than the case for it — the maintainer's ask ([PRQL/prql#6211](https://github.com/PRQL/prql/pull/6211#issuecomment-5342977728)) was for a higher bar on runner-time-only changes, and that is what this states.

The first matters more than a wording nit: the two files state the bar with different subjects. `running-in-ci` tests the **change's benefit** ("a change whose only benefit is runner time — nothing about it makes a build correct, or a red branch green"), which is the framing that covers the triggering case — a mirror stall that hung a job to the platform cap left CI red, but a rerun passed, so no permanent change was buying correctness. `ci-fix` tests the **fault's cost** ("Where the fault costs only runner time"), and by that test the triggering incident is outside the clause, since the fault did leave a job red. Read the other way it is too broad: nearly every transient CI failure "costs only runner time" once a rerun passes, so the clause can suppress the repeat-occurrence escalation for a genuine recurring flake — the thing that section exists to catch. Suggested inline: condition on what the mitigation buys, matching `running-in-ci`.

plugins/tend-ci-runner/skills/ci-fix/SKILL.md
tend-agent: Right on both directions, and the too-broad one is the worse of the two — a recurring flake that reruns green is exactly what this escalation exists to catch, and "costs only runner time" would have suppressed it. Taken as suggested in `f647ea0`, so both files now condition on what the mitigation buys rather than on what the fault cost.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Agreed — the closing sentence and the "when writing one, and when reviewing one" clause were both already said within five lines. Taken as suggested in `f647ea0`; the bullet keeps only the CI-flavoured examples.

plugins/tend-ci-runner/skills/ci-fix/SKILL.md
```diff
@@ -78,32 +78,34 @@ Automated fix for [failed run](run-url)
...
-Before applying the transient path, check whether the same failure shape has already been classified transient recently. The single-shot criteria above don't catch an intermittent upstream regression — each rerun-pass reinforces the wrong classification.
+Before filing the tracker below, check whether the same failure shape has already been classified transient recently. The single-shot criteria above don't catch an intermittent upstream regression — each rerun-pass reinforces the wrong classification.
 
 ```bash
-gh issue list --state all --author "$BOT_LOGIN" --search "ci-fix: in:title" \
-  --json number,title,createdAt \
+gh issue list --state all --label tend-outage --author "$BOT_LOGIN" \
+  --json number,title,body,createdAt \
   --jq "[.[] | select(.createdAt >= (now - 7*86400 | todateiso8601))]"
```
tend-agent, line 92:
```suggestion
gh issue list --state all --label tend-outage --author "$BOT_LOGIN" \
  --json number,title,body,createdAt \
  --jq '[.[] | select(.createdAt >= (now - 7*86400 | todateiso8601))
             | select(.title | startswith("ci-fix: transient failure"))]'
```

tend-agent (commented):
One finding, on the new escalation query.

`tend-outage` has two producers, not one: this skill's transient tracker, and the composite action's `report-failure.sh`, which files **"Bot temporarily unavailable"** issues under the same label from the same bot account. Dropping `--search "ci-fix: in:title"` for the label alone therefore widens the query rather than narrowing it — run as written against this repo it returns only the second family (#1008, #946, #905, #832, #831 are the five most recent `tend-outage` issues, all "Bot temporarily unavailable"; no ci-fix tracker appears at all).

The body-keyword match in the prose is the only thing separating them, and a bot-availability outage genuinely *is* a repeated transient infra fault — so a session reading five of them is being asked to reject on a distinction the query didn't make. `gh issue list`'s 30-issue default page compounds it: a burst of availability issues can push in-window ci-fix trackers off the list, and the escalation under-counts silently, which is the exact failure mode this section exists to catch.

The label filter is still the right call — it's what stops 3b's durable trackers being counted, which the old title search did wrongly. Suggested inline: keep it and put the title precision back alongside it.

plugins/tend-ci-runner/skills/ci-fix/SKILL.md
tend-agent: Confirmed, and worse than the review states — of the 30 `tend-outage` issues on this repo's default page, 29 are "Bot temporarily unavailable"; the one `ci-fix:` tracker (#676) is from June, so the seven-day window returns nothing regardless. The label alone can't tell the two producers apart, and a burst of availability rows would push a real tracker off the page silently.

Taken in `6196c25`, but filtering server-side rather than in `--jq`: `--search "ci-fix: transient failure in:title"` alongside the label spends the page on the right family instead of discarding 29 rows after fetching them, which is the half of the finding a client-side `startswith` leaves open. Verified live — the label+author+search combination returns only #676, with `body` still populated for the keyword match.

Added a line under the block naming both filters as load-bearing, so the next edit doesn't drop one again.

PR #1017, opened 2026-08-19
fix(install-tend): correct why the running-tend overlay needs frontmatter

max-sixty:
Step 4 of `plugins/install-tend/skills/install-tend/SKILL.md` said skill discovery needs frontmatter, so an overlay without it was never loaded. That's wrong, and #1012 put it there.

Claude Code discovers a frontmatter-less `SKILL.md` and falls back to the directory name and the file's first heading. Measured on 2.1.235 (this repo's action pins 2.1.233) by listing the available skills in a checkout of PRQL/prql, which offered its overlay as `running-tend: Running Tend in PRQL`. So those overlays were loading all along; what they lacked is the description an agent judges relevance from — which is still a reason to write frontmatter, just not the one the step gave.

Step 4 now says that in three lines instead of six, and the duplicate frontmatter instruction on the customized-overlay path is gone: the placeholder eight lines below already shows it.

The same claim also stood in `running-in-ci`, which every CI session loads, as "Repo-local skills must have YAML frontmatter (`name` + `description`) to be auto-discovered". That line is deleted rather than rewritten — discovery lists an overlay either way, and First Steps already tells the session to load a listed `running-tend`, so nothing there depends on the mechanics. The requirement to write frontmatter into a *new* overlay keeps its one home in `references/skill-pr-workflow.md`.

The two consumers that lacked frontmatter now have it — PRQL/prql#6208 and numbagg/numbagg#747.

### DIFF
--- a/plugins/install-tend/skills/install-tend/SKILL.md
+++ b/plugins/install-tend/skills/install-tend/SKILL.md
@@ -523,10 +523,9 @@ environment protection rules on top.
 ## 4. Create skill overlay (recommended)
 
 Create `.claude/skills/running-tend/SKILL.md` with tend-specific project
-guidance. This skill is loaded by tend workflows alongside the generic
-`tend-*` skills — by skill discovery, so it must open with the
-frontmatter shown below. An existing overlay without it isn't done:
-add the frontmatter in place.
+guidance, opening with the frontmatter below so discovery lists it by
+description rather than by its first heading. An existing overlay without
+frontmatter needs it added in place.
 
 **Do NOT duplicate CLAUDE.md** and **do NOT invent project conventions.**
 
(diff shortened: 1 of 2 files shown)


tend-agent (commented):
The correction looks right, and the trimmed step 4 reads better than what it replaces. One gap: the claim this PR falsifies also lives in `running-in-ci`, which every CI session loads.

[`plugins/tend-ci-runner/skills/running-in-ci/SKILL.md`](https://github.com/max-sixty/tend/blob/a2149c971955a6a0316c6875044fb04ee3d45256/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md#L18), last line of "First Steps — Load Repo-Specific Guidance":

> Repo-local skills must have YAML frontmatter (`name` + `description`) to be auto-discovered.

Same "must, or it isn't discovered" framing, in the higher-traffic file — `install-tend` runs once per adopter, while `running-in-ci` is read by every review, triage, and scheduled session. Left standing it contradicts the corrected step 4, and a session reading it can tell an adopter their frontmatter-less overlay isn't loading at all. The fix is the same shape as this PR's: frontmatter supplies the description discovery lists the skill by, it isn't what makes the skill discoverable.

It's outside this diff so I can't attach it as a suggestion — happy to push the edit onto this branch if you'd rather it land here than as a follow-up.

tend-agent (approved):

PR #1021, opened 2026-08-20
fix(nightly): leave the tend-outage tracker for review-runs to drain

### LINKED ISSUE #1019: nightly closes a tend-outage tracker before review-runs can drain it, so stranded triggers are dropped silently
`review-runs`' **Drain stranded triggers** step queries `gh issue list --state open --label tend-outage`. `nightly`'s "close a resolved issue" rule fires on the same tracker one workflow earlier in the day, and a `tend-outage` body literally invites it ("Close it once the outage is resolved"). Where a repo's nightly cron precedes its review-runs cron — the generated crons on `max-sixty/leaf` are `17 6` against `47 7` — the drain can never see a tracker, and every stranded trigger is dropped silently.

It fired on `max-sixty/leaf` last night, first day of adoption.

## The chain

1. `tend-mention` [32304239514](https://github.com/max-sixty/leaf/actions/runs/32304239514) died in `handle` on `failed to install claude 2.1.226 after 3 attempts` (three `curl: (22) ... error: 403`), stranding the review-response dispatch for [leaf#10](https://github.com/max-sixty/leaf/pull/10). Session artifact 236 bytes — the agent never started.
2. "Report failure" filed [leaf#11](https://github.com/max-sixty/leaf/issues/11) with that one row.
3. `tend-nightly` [32339817890](https://github.com/max-sixty/leaf/actions/runs/32339817890) enriched it, then closed it. Its summary gives the basis verbatim: *"confirmed no tend run has failed since (60 runs, all `success`/`skipped`) and closed with that evidence."* That is a claim about the outage. The row is a claim about a trigger, and nothing in nightly's rule asks about it.
4. `tend-review-runs` [32346633522](https://github.com/max-sixty/leaf/actions/runs/32346633522) ran the drain, got an empty list, and would have reported the night clean.

## What was actually lost

The stranded review carried an inline `suggestion` on `plugins/leaf/bin/leaf`. At the PR's current head it is unapplied, and no successor exists: the branch was rebased and `tend-review` [32323190154](https://github.com/max-sixty/leaf/actions/runs/32323190154) re-reviewed the new head, correctly posting nothing — *"That thread already blocks the PR, and per the review skill's dedup rule, restating it on a push that changed none of the PR's code would be noise."* The re-review is right. The point is that the dedup rule assumes the earlier finding reached someone, and here the run that would have actioned it is the one that died.

Recovering it needed the failed run to still be inside review-runs' census window. A day later the pointer is gone in both places.

## Two places it could be fixed

**Nightly stops closing `tend-outage` issues.** It already carves out "closing the issue is itself a signal" cases (`Do not close manually` bodies, `tend-rate-limit`). A `tend-outage` tracker is the same kind: the close is the drain's, and nightly's evidence — "nothing has failed since" — is structurally the wrong question. This is a removal, and it puts the close with the only step that reads the rows.

**Or the drain widens.** `review-runs` could query outage issues closed since its window anchor as well as open ones. Cheaper to state, but it leaves two workflows with an interest in the same record and depends on the closure landing inside one window.

The first looks right to me — one workflow owns the tracker's lifecycle, and the ownership matches who can check the rows. Same family as #943 (nightly closing a PR out from under the mention session that had just fixed it): nightly's disposition authority reaching past what its evidence covers.

I have added a repo-local overlay on `max-sixty/leaf` telling nightly to leave the close alone, since the action ref there is pinned at 0.1.18 and an upstream fix stays dormant until the next release. Filing here because both halves — nightly's close rule and the drain's `--state open` filter — are bundled, so any consumer scheduling nightly before review-runs has the same hole.

tend-agent:
## Problem

`review-runs`' **Drain stranded triggers** step is the only thing that reads the rows on the `tend-outage` "Bot temporarily unavailable" tracker — each row names a trigger a failed run stranded, and nothing re-runs those triggers on its own. But `nightly`'s Step 5 resolved-issue rule can close that tracker first, on evidence that answers a different question: "no tend run has failed since". That says nothing about whether the stranded triggers were re-run.

The ordering isn't incidental. The generated default crons are `nightly` at `17 6 * * *` and `review-runs` at `47 7 * * *` ([`workflows.py`](https://github.com/max-sixty/tend/blob/acd9d32bbcb46b23c42817dc0d4922c3044142b5/generator/src/tend/workflows.py#L381-L385)), so nightly reaches the tracker first every day, on every adopter that hasn't overridden the cron. The drain then queries `--state open`, finds nothing, and reports the night clean. The tracker body invited exactly this: "Close it once the outage is resolved" — the outage resolving is not the drain completing.

Reported in #1019 with a chain observed on a consumer repo, where the lost row pointed at a `tend-mention` run that died before the agent started, stranding a review-response dispatch whose inline suggestion never reached anyone.

## Solution

Nightly stops closing `tend-outage` issues, so the tracker's lifecycle belongs to the one step that reads its rows. This is a removal, not new machinery — the carve-out list in nightly's rule already exists for the same reason (`Do not close manually` bodies, `tend-rate-limit`), and this is a third member of that family: the close is a signal, not a record of resolution.

`report-failure.sh`'s tracker body now states the real criterion, so a maintainer closing it by hand uses the same one, and the file comment records who owns the close.

The issue's second option — widening the drain to also read outage issues closed since its window anchor — is not taken. It leaves two workflows with an interest in the same record and depends on the closure landing inside one window, where this puts the close with the only reader.

That makes `review-runs` the tracker's only automatic closer, which is a deliberate trade rather than a side effect. An adopter can disable that workflow (`workflows: review-runs: enabled: false`), and on such a repo nothing closes the tracker: `report-failure.sh` appends to the open one, so every later incident folds into a record the drain step's own closing line warns about. A carve-out for that case would need nightly to read the adopter's config before deciding whether the label is off-limits — more machinery than the case earns, and no repo in `data/consumers.json` disables the sweep. What the case does earn is wording that stays true without it: the tracker body states the criterion and names `tend-review-runs` as what satisfies it where it runs, rather than as the only route, and the file comment says the close falls to a maintainer where the sweep is absent.

`ci-fix`'s transient tracker shares the `tend-outage` label but is created and closed in the same step, so it is never open for nightly to reach; carving out the label rather than the title costs nothing.

## Testing

`uv run pytest generator/tests/test_shared_steps.py` passes (70 tests) — no assertion covers the tracker's body prose, so the string change is unasserted. `shellcheck` and `pre-commit` clean on both files, and the `printf` renders the escaped backticks as literal inline code.

The nightly half is skill prose with no test surface; the claim it rests on is the default cron table cited above and the drain's `--state open` query at [`review-runs/SKILL.md`](https://github.com/max-sixty/tend/blob/acd9d32bbcb46b23c42817dc0d4922c3044142b5/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L215-L217).

---
Closes #1019 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -173,7 +173,10 @@ gh pr list --state open --limit 200 --json number,title,headRefName
 
 For each open issue, check whether recent commits or the current codebase state already resolve it. If resolved, comment with the evidence (commits, CI runs, or code state that resolves the issue). Close the issue with `gh issue close` when:
 
-- The bot opened the issue itself to report a transient condition (e.g., a "Nightly tests failed" report from a prior run) and the condition has clearly resolved — the fix PR is merged and the relevant CI on `main` is passing. Skip this case where closing the issue is itself a signal rather than a record of resolution: a body containing "Do not close manually" (recurring trackers with their own lifecycle), or the `tend-rate-limit` label, where a maintainer's close is what lifts the bot past its own rate limit. Closing that one as the bot lifts nothing — the preflight counts only closes by a person — but it clears a decision still waiting on one.
+- The bot opened the issue itself to report a transient condition (e.g., a "Nightly tests failed" report from a prior run) and the condition has clearly resolved — the fix PR is merged and the relevant CI on `main` is passing. Skip this case where closing the issue is itself a signal rather than a record of resolution:
+  - a body containing "Do not close manually" — recurring trackers with their own lifecycle.
+  - the `tend-outage` label. Its rows name the triggers a failed run stranded, and the close belongs to `review-runs`' drain step, which reports each row re-run or confirmed no longer needed. "Nothing has failed since" answers a different question, and nightly's cron precedes `review-runs`' under the generated defaults — closing it here drops the rows before the only step that reads them.
+  - the `tend-rate-limit` label, where a maintainer's close is what lifts the bot past its own rate limit. Closing that one as the bot lifts nothing — the preflight counts only closes by a person — but it clears a decision still waiting on one.
 - The repo's guidance (e.g., `running-tend` skill) explicitly authorizes closing issues.
 
 Otherwise, leave it open for a maintainer to close.
--- a/shared/steps/report-failure.sh
+++ b/shared/steps/report-failure.sh
@@ -14,9 +14,12 @@
 # data.
 #
 # A closed outage issue is left closed and a fresh one filed: closing it means
-# the outage was resolved, and reopening would fold the next incident into a
-# stale record. The rate-limit issue takes the opposite policy, for reasons in
-# lib/run-issue.sh.
+# the `review-runs` drain step read every row and re-ran what needed it, and
+# reopening would fold the next incident into a stale record. That drain owns
+# the close — the nightly skill's resolved-issue rule carves this label out,
+# because nightly's cron precedes review-runs' and "nothing has failed since"
+# says nothing about the stranded triggers the rows name. The rate-limit issue
+# takes the opposite policy, for reasons in lib/run-issue.sh.
 #
 # Inputs (env): GITHUB_TOKEN (for gh), GITHUB_SERVER_URL, GITHUB_REPOSITORY,
 # GITHUB_RUN_ID, GITHUB_EVENT_NAME, GITHUB_EVENT_PATH (from Actions).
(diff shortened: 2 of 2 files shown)


tend-agent (commented):
Not a blocker, but worth making deliberate: this leaves `review-runs` as the tracker's *only* closer. On an adopter that sets `workflows: review-runs: enabled: false`, nothing closes it — and `report-failure.sh` appends to an open tracker rather than filing a fresh one, so every later incident folds into the stale record, which is exactly what the drain step's own closing line warns against. The new body string names `tend-review-runs` as the closer too, which reads oddly on a repo that doesn't run it.

Nothing is affected today — every repo in `data/consumers.json` ships `tend-review-runs.yaml` — and a conditional carve-out would be more machinery than the case earns, so I'd leave it as is. Flagging it so the single-owner dependency is a choice rather than a side effect: the PR body weighs widening the drain, but not the case where the drain isn't there.

tend-agent:
Agreed on the call, and taken the part that's free: the tracker body no longer treats the sweep as the only route. It now reads "Close it once every row above is drained — each stranded trigger re-run, or confirmed no longer needed. The `tend-review-runs` sweep does that where it runs." — criterion first, so it's actionable on a repo that has `review-runs` disabled and doesn't point a maintainer at a workflow that never fires. The file comment says the same, and the PR body now records the single-owner dependency as a choice, including the fold-into-a-stale-record consequence when the sweep isn't there.

No carve-out beyond that. Making nightly close the tracker only where `review-runs` is disabled means nightly reading the adopter's config to decide whether the label is off-limits, which is more machinery than a case no current consumer is in.

Pushed as 043d554.

PR #1022, opened 2026-08-20
Require plain PR titles from Tend workflows

max-sixty:
Recent Tend runs in Leaf copied stylized commit subjects into PR titles because the shared CI guidance covered body structure but not title style. This change tells agents to use plain, literal PR titles and commit subjects that name the component and changed behavior, with explanations kept in the body.

Three A/B prompt checks changed stylized titles to direct ones. `pre-commit run --files plugins/tend-ci-runner/skills/running-in-ci/SKILL.md` and `git diff --check` pass.

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -130,10 +130,6 @@ Default: don't act in another repo unsolicited. File an issue in the current rep
 
 When asked to create a PR, use `gh pr create` directly.
 
-Write PR titles and commit subjects in plain, literal language that a reader can
-understand without the body. Name the concrete component and behavior changed.
-Put the explanation in the body. Example: `Stop worker retries after cancellation`.
-
 Before creating a branch or PR, check for existing work:
 
 ```bash
@@ -142,6 +138,11 @@ git branch -r --list 'origin/fix/*'
 ```
 
 Open PRs compete for one maintainer's attention. A self-initiated improvement — a sweep finding, a skill or workflow refinement nobody asked for — draws on a budget: when the bot already has five or more PRs open (`gh pr list --state open --author "@me"`), open one only for a wrong outward action (see **Weighing a Fix**), and hold the rest until the queue drains, recorded where the maintainer will see it (the evidence store, or a line on the triggering thread) rather than as an issue (**Filing Issues in This Repo** explains why not). The budget never holds work someone asked for, a fix a user is waiting on (a red default branch, a triaged bug), or the scheduled maintenance a skill itself instructs (a workflow regeneration, a pinned-version bump, a data refresh). Base every PR on the default branch; never stack one on an unmerged bot branch, which puts the same change through review once per link in the chain.
+
+Write PR titles and commit subjects in plain, literal language that a reader can
+understand without the body. Name the concrete component and behavior changed
+while keeping any prefix the repository requires. Put the explanation in the
+body. Example: `Stop worker retries after cancellation`.
 
 Open the PR body with two or three sentences — problem, fix, verification — and fold supporting detail into `<details>` (per **Comment Formatting**).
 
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -130,6 +130,10 @@ Default: don't act in another repo unsolicited. File an issue in the current rep
 
 When asked to create a PR, use `gh pr create` directly.
 
+Write PR titles and commit subjects in plain, literal language that a reader can
+understand without the body. Name the concrete component and behavior changed.
+Put the explanation in the body. Example: `Stop worker retries after cancellation`.
```
tend-agent: Scope this to wording, so it doesn't read as licence to drop a repo's title prefix — this repo's own weekly PRs are required to be `chore: ...`, and `skill-pr-workflow.md` and `weekly/SKILL.md` both defer to the repo's convention.

```suggestion
Write PR titles and commit subjects in plain, literal language that a reader can
understand without the body. Name the concrete component and behavior changed.
Put the explanation in the body. This governs wording, not the repo's title
convention — keep whatever prefix that uses. Example: `Stop worker retries after
cancellation`.
```

tend-agent (commented):
The gap is real and the fix is the right size — three lines, an example rather than a template, explanation kept in the body. One thing to tighten before it ships to every adopter.

The example `Stop worker retries after cancellation` is prefix-free, and the paragraph says nothing about prefixes. But this repo's own titles are conventional-commit prefixed almost without exception (the last 25 merged PRs are all `fix(...)`, `chore:`, `docs(...)`, `skills(...)`, `feat(...)`, `ci:`, or `simplify:`), and two other bundled skills tell the agent to defer to that: `references/skill-pr-workflow.md` under "Open as a separate PR" ("Follow the repo's PR title conventions (conventional commits, Jira prefix, or whatever the repo uses)") and `weekly/SKILL.md` Step 3. `running-tend` goes further and mandates literal prefixed titles for the weekly sweep — `chore: refresh consumers.json`, `chore: bump <name> to <version>`. An agent that reads only the new paragraph — which sits in the skill loaded at the top of every session — has a plain reason to drop the prefix, and in an adopter repo gated on a semantic-PR-title or commitlint check that turns a title into a red check. The inline suggestion adds a clause scoping this to wording; it deliberately leaves the example prefix-free, since prescribing a prefix format here is what CLAUDE.md's "Skill design: bundled for everyone, overlay for one" says belongs in an overlay.

Two smaller notes, neither blocking:

The paragraph lands between `gh pr create` and "Before creating a branch or PR, check for existing work", splitting the create-flow, while the sibling guidance about what to write ("Open the PR body with two or three sentences — problem, fix, verification") sits two paragraphs further down. Moving it next to that one would group the what-to-write content and leave the dedup flow contiguous — and would let "Put the explanation in the body" lean on the body paragraph instead of restating it.

Unrelated to the diff, but on-topic for it: this PR's body ends with `> _This was written by Codex on behalf of max-sixty_`, which the file being edited already forbids under **Comment Formatting** ("Don't add job links, footers, or authorship sign-offs ... This covers PR and issue bodies too, not just comments"). Same shape as the problem this PR fixes — a rule that exists but isn't reaching the sessions — so it may be worth checking whether the sign-off is coming from somewhere the guidance can reach.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
max-sixty: Updated in a9e745d2. The rule now preserves repository-required prefixes and sits beside the PR-body guidance.

tend-agent: Confirmed — the prefix clause is in and the paragraph now sits directly above the PR-body guidance, so the create-flow reads contiguously. Folding "while keeping any prefix the repository requires" into the sentence about naming the component is tighter than the separate clause I suggested; no further concern from me.

On the sign-off note: it isn't reachable from guidance. The string appears nowhere in this repo outside the rule that forbids it — not in `codex/agents-tail.md`, `shared/system-prompt.md`, or any skill — so it's appended by the Codex harness outside the session rather than chosen by it, and no skill edit would suppress it. Treat that third note as closed.

tend-agent (approved):

PR #1023, opened 2026-08-20
fix(generator): cap every generated job with timeout-minutes

### LINKED ISSUE #962: review: `cancel-in-progress: false` with no `timeout-minutes` lets a wedged run hold the per-PR queue for the full 360-min job cap
0.1.15 flipped the review job to `cancel-in-progress: false` (#903), so a push mid-review queues a replacement run instead of cancelling the live one. The examined-HEAD gate makes that wait cheap in the normal case. What's missing is a floor under the abnormal one: no generated job declares `timeout-minutes`, so a review run that wedges rather than ends holds the per-PR concurrency group until GitHub's 360-minute default job cap, and every queued replacement review sits behind it. Under 0.1.14 the next push cancelled a wedged run outright — that escape hatch is what the flip removed.

The supervisor's `timeout_seconds` (default `21000`, 5h50m) does not cover this. It kills the *headless run*; the steps before it — checkout, `tend-setup`, the `claude` / `codex` installers, the mitmproxy cache warm — run outside the supervised window, and those are exactly the steps with a hang history: #474 recorded two pre-agent setup hangs (~25 min each, one on the PRQL/prql leg) from an apt/DNS stall, and #765 was a ~15-minute hang that produced no session-log artifact at all. A hang in that region is bounded only by GitHub's job cap.

**I have not observed a wedge — this is a tail risk, not a live bug.** Across the last 100 completed runs of each workflow, the longest `tend-review` was 45 minutes on PRQL/prql and 14 minutes on this repo. The change in 0.1.15 is that the consequence of one got worse: previously a push cleared it, now it blocks the PR's review queue for up to six hours.

## Current state

No `timeout-minutes` anywhere in the generator templates as of `c9e44e2` on `main`:

| template | `timeout-minutes` |
| --- | --- |
| `ci-fix.yaml.j2` | 0 |
| `review.yaml.j2` | 0 |
| `mention.yaml.j2` | 0 |
| `scheduled.yaml.j2` | 0 |
| `triage.yaml.j2` | 0 |
| `notifications.yaml.j2` | 0 |
| `macros.yaml.j2` | 0 |

Same count across all eight generated `tend-*.yaml` files on a consumer (PRQL/prql at 0.1.15).

## Proposed fix

`timeout-minutes:` on the generated review job. This is a shape you already endorsed once — in #736 you asked for exactly it ("let's just do the timeout of 120min"), and that PR was reworked to set `timeout-minutes: 120` on `fix-ci` in the template, but it closed unmerged on 2026-06-28 and nothing carried it forward.

One sizing tension worth deciding deliberately rather than by default: any job cap below the supervisor's 5h50m budget silently becomes the real session limit, so `timeout-minutes` and `timeout_seconds` want to be set as a pair (or the former derived from the latter plus setup overhead). Two coherent positions:

- **Cap for the wedge only** — set `timeout-minutes` just above the supervisor budget (~355–360). Bounds nothing GitHub doesn't already bound, so it only helps if the aim is documentation.
- **Cap for the queue** — set it near observed reality (60–90 min covers the 45-minute max with margin). This makes the job timeout the operative budget for review, which is arguably right for a workflow that now queues: an unbounded review head-of-line-blocks every later push on that PR, while a review that genuinely needs six hours is not a case that has occurred.

The second is what the queueing change argues for, but it's a policy call about how long a review session is allowed to run, so it's yours.

<details><summary>Where this came from</summary>

Raised while reviewing PRQL/prql#6191, the nightly 0.1.14 → 0.1.15 regen on PRQL/prql. Generated workflows carry "Do not edit this file directly", so there was nothing to suggest inline there.

Verification notes:

- Concurrency group on the generated review job is `${{ github.workflow }}-${{ github.event.pull_request.number }}` with `cancel-in-progress: false` — per-PR, so the blocking is scoped to one PR's review queue, not the whole repo.
- `timeout_seconds` default and its rationale are in `claude/action.yaml`; the supervisor poll loop that enforces it runs after the setup steps, which is why a pre-agent hang escapes it.
- Run durations from `actions/workflows/<wf>/runs?per_page=100`, filtered to `status=completed`, `updated_at - run_started_at`.

</details>

tend-agent:
No generated job declared `timeout-minutes`, so a step that wedged rather than ended was bounded only by GitHub's 360-minute default — and on a job with `cancel-in-progress: false` it held the concurrency queue for the same six hours. This stamps a job-level cap on all eleven generated jobs, using the measured values from [PRQL/prql#6215](https://github.com/PRQL/prql/pull/6215), and derives the Claude harness's own session budget from each cap so the two kills stay ordered. Verified by rendering every workflow and asserting each job's cap and each agent step's budget (`uv run pytest`, 572 passed; `pre-commit run --all-files` clean).

A job cap is the only bound that reaches the whole job. The harness supervisor's `timeout_seconds` starts once the action is entered, so checkout, the adopter's `setup:`, and the CLI installers sit outside it — which is exactly where the five observed six-hour wedges in #962 landed, in an adopter setup composite's unbounded `apt-get update`. `timeout-minutes` is not accepted on a step inside a composite action, so the generated job is the only lever that reaches there.

The cap alone would have made the supervisor unreachable, since every value sits below the claude action's 21000s `timeout_seconds` default — and the two kills are not equivalent. The supervisor exits 1, so the job is `failure` and the action's `failure()`-gated `Report failure` step files the `tend-outage` row that `review-runs` drains to re-run stranded triggers; GitHub's `timeout-minutes` kill yields `cancelled`, where `failure()` steps are skipped and the overrun goes unreported. So the cap is also threaded into `agent_step`, which emits `timeout_seconds` 10 minutes under it: the supervisor still owns an agent-session overrun, and the job cap backstops the pre-agent region it cannot reach.

Adopter overrides are unchanged and still win: `workflows.<name>.jobs.<job>.timeout-minutes` now replaces the default in place rather than appending, so an override lands next to the comment documenting what it replaced. It moves the job cap alone — step inputs take no override, so the generated supervisor budget stays put, which degrades to the reported kill rather than the silent one.

Closes #962.

<details><summary>Caps, budgets, and how they're wired</summary>

| workflow | job | `timeout-minutes` | `timeout_seconds` (claude) |
|---|---|---|---|
| `tend-ci-fix` | `fix-ci` | 60 | 3000 |
| `tend-notifications` | `notifications` | 60 | 3000 |
| `tend-weekly` | `weekly` | 60 | 3000 |
| `tend-review` | `review` | 120 | 6600 |
| `tend-mention` | `handle` | 120 | 6600 |
| `tend-triage` | `triage` | 120 | 6600 |
| `tend-review-runs` | `review-runs` | 120 | 6600 |
| `tend-nightly` | `nightly` | 180 | 10200 |
| `tend-mention` | `relay` | 10 | — |
| `tend-mention` | `verify` | 10 | — |
| `tend-install-test` | `install-test` | 10 | — |

Both numbers come from a single `_JOB_TIMEOUT_MINUTES` entry per workflow in `workflows.py`, so the cap and the budget cannot drift apart. A `job_timeout(minutes)` macro in `macros.yaml.j2` emits the cap plus a two-line override pointer, and `agent_step` takes the same value and emits the derived budget inside its claude branch; codex emits nothing, having no supervisor timeout. The three jobs with no agent step take a literal cap in their template.

The ordering the derivation buys holds while the steps ahead of the agent fit inside the 10-minute margin; past that the job cap wins and the kill is silent again — but that is the pre-agent region `timeout-minutes` exists to bound in the first place, and it is the case this PR was written for.

Test changes:

- `test_every_job_declares_a_timeout` — every generated job must carry a cap under 360, since at or above that GitHub's default is the operative bound again. The corpus opts in `ci-fix` and `install-test` and asserts it is every workflow, so a skipped one can't quietly fall out of coverage.
- `test_agent_budget_expires_before_the_job_cap` — every claude agent step's `timeout_seconds` must be under its job's cap, with the corpus pinned to `GENERATORS`.
- `test_codex_agent_step_takes_no_supervisor_budget` — the codex branch emits no `timeout_seconds`.
- `test_mention_job_extras_target_specific_job` probed "extras hit only the named job" by asserting `timeout-minutes` was *absent* from `verify`. With a default on both jobs, untouched now means the default survives, so it asserts `verify` is still 10 while `handle` takes the override's 180.

Snapshot churn is the eleven caps, the eight budgets, and their comments; the one non-additive line is the extras snapshot, where the override moved from the end of the job to the base key's position.

This repo's own `.github/workflows/tend-*.yaml` are deliberately not regenerated here — per CLAUDE.md the action ref pins the generator's own version, so regenerating in-tree would stamp an unreleased tag. They pick this up at the next release plus nightly regen.

</details>

### DIFF
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -347,7 +347,10 @@ bot_name: my-project-bot
 # `handle`; 180 for nightly; 10 for mention's `relay` and `verify`. The cap
 # covers the whole job, `setup:` included, which is the only bound that
 # reaches steps running before the harness action is entered. Override it
-# where a job legitimately needs longer:
+# where a job legitimately needs longer — note that the override moves the job
+# cap only: the Claude harness's own session budget is generated 10 minutes
+# under the default cap, so a much larger override leaves the session bounded
+# where it was (reported as a failure, not silently cancelled).
 #
 # workflows:
 #   review:
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -342,6 +342,13 @@ bot_name: my-project-bot
 #
 # ### Example: longer timeout on a specific job
 #
+# Every generated job ships a `timeout-minutes` default — 60 for ci-fix,
+# notifications and weekly; 120 for review, triage, review-runs and mention's
+# `handle`; 180 for nightly; 10 for mention's `relay` and `verify`. The cap
+# covers the whole job, `setup:` included, which is the only bound that
+# reaches steps running before the harness action is entered. Override it
+# where a job legitimately needs longer:
+#
 # workflows:
 #   review:
 #     jobs:
--- a/docs/tend.example.yaml
+++ b/docs/tend.example.yaml
@@ -349,8 +349,10 @@ bot_name: my-project-bot
 # reaches steps running before the harness action is entered. Override it
 # where a job legitimately needs longer — note that the override moves the job
 # cap only: the Claude harness's own session budget is generated 10 minutes
-# under the default cap, so a much larger override leaves the session bounded
-# where it was (reported as a failure, not silently cancelled).
+# under the default cap. Raising the cap leaves the session bounded where it
+# was (reported as a failure, not silently cancelled); lowering it below that
+# budget puts the supervisor out of reach again, and an overrunning session
+# goes back to being cancelled silently.
 #
 # workflows:
 #   review:
--- a/generator/src/tend/templates/macros.yaml.j2
+++ b/generator/src/tend/templates/macros.yaml.j2
@@ -13,8 +13,24 @@
    active credential through its credential proxy (OAuth wins when both are
    present), treating empty as absent. Codex takes a single `openai_api_key` —
    the subscription `auth.json` path is incompatible with tend's concurrent
-   workflows and is not exposed. #}
-{% macro agent_step(cfg, prompt_body, if_condition='') %}
+   workflows and is not exposed.
+
+   `timeout_minutes` is the enclosing job's cap, and the claude branch spends
+   it: the supervisor's budget is set 10 minutes under it so the two kills stay
+   ordered. They are not equivalent — the supervisor exits 1, which paints the
+   job `failure` and so runs the action's `failure()`-gated `Report failure`
+   step, filing the `tend-outage` row that `review-runs` drains to re-run
+   stranded triggers; GitHub's `timeout-minutes` kill yields `cancelled`, where
+   `failure()` steps are skipped and an overrun goes unreported. Leaving the
+   action's 21000s default in place would put every job cap below it and make
+   the supervisor unreachable. The ordering holds while the steps ahead of the
+   agent — checkout, `setup:`, the CLI installers — fit inside that 10-minute
+   margin; past it the job cap wins and the kill is silent again, but that is
+   the pre-agent region `timeout-minutes` exists to bound in the first place.
+
+   Codex has no supervisor timeout, so its branch emits nothing and the job cap
+   is that harness's only bound. #}
+{% macro agent_step(cfg, prompt_body, timeout_minutes, if_condition='') %}
 {% if cfg.harness == 'claude' %}
       - uses: max-sixty/tend/claude@<<tend_version>>
 {% else %}
@@ -37,6 +53,10 @@
           bot_name: <<cfg.bot_name>>
           model: <<cfg.model>>
 {% if cfg.harness == 'claude' %}
+          # 10 minutes under the job cap, so an overrunning session takes the
+          # harness's kill — job red, failure reported — rather than GitHub's,
+          # which cancels the job and skips the report.
+          timeout_seconds: <<(timeout_minutes - 10) * 60>>
 {% if cfg.sandbox_path %}
           sandbox_path: |
 {% for dir in cfg.sandbox_path %}
@@ -90,9 +110,17 @@
    someone pushes, so a wedged review is a review that never happens.
 
    Values are sized above observed healthy durations, not near the supervisor
-   budget: a cap that only restates GitHub's default bounds nothing. Adopters
-   who need more override per job —
-   `workflows.<name>.jobs.<job>.timeout-minutes` in .config/tend.yaml. #}
+   budget: a cap that only restates GitHub's default bounds nothing. The same
+   value sets the agent step's `timeout_seconds` 10 minutes lower (see
+   `agent_step`), so the supervisor still owns an agent-session overrun and
+   this cap backstops the region it cannot reach. Both come from
+   `_JOB_TIMEOUT_MINUTES` in workflows.py — one number per workflow, so the two
+   cannot drift apart. Adopters who need more override per job —
+   `workflows.<name>.jobs.<job>.timeout-minutes` in .config/tend.yaml. That
+   override moves the job cap alone — step inputs take no override, so the
+   supervisor budget stays at the generated value and an adopter who raises
+   the cap gets a session still bounded by the default. Reported, not silent:
+   the supervisor's kill is the one that files the outage row. #}
 {% macro job_timeout(minutes) %}
     # Bounds the whole job, adopter `setup:` included. Override with
     # `workflows.<name>.jobs.<job>.timeout-minutes` in .config/tend.yaml.
--- a/generator/src/tend/templates/macros.yaml.j2
+++ b/generator/src/tend/templates/macros.yaml.j2
@@ -79,6 +79,26 @@
       deployment: false
 {%- endmacro %}
 
+{# Job timeout at column 4. A job cap is the only bound that reaches the whole
+   job: the harness supervisor's `timeout_seconds` starts once the action is
+   entered, so every step before it — checkout, the adopter's `setup:`, the
+   CLI installers — is otherwise bounded only by GitHub's 360-minute default.
+   Observed wedges live in exactly that region (an unbounded `apt-get update`
+   in an adopter setup composite, killed at 360 minutes), and a wedge on a job
+   with `cancel-in-progress: false` holds its concurrency queue for the same
+   six hours. On `pull_request_target` there is no successor run unless
+   someone pushes, so a wedged review is a review that never happens.
+
+   Values are sized above observed healthy durations, not near the supervisor
+   budget: a cap that only restates GitHub's default bounds nothing. Adopters
+   who need more override per job —
+   `workflows.<name>.jobs.<job>.timeout-minutes` in .config/tend.yaml. #}
+{% macro job_timeout(minutes) %}
+    # Bounds the whole job, adopter `setup:` included. Override with
+    # `workflows.<name>.jobs.<job>.timeout-minutes` in .config/tend.yaml.
+    timeout-minutes: <<minutes>>
+{%- endmacro %}
+
 {# Permissions block at column 6. `issues` defaults to True because every
    workflow except ci-fix needs it. The trailing inline conditional avoids
    the trim_blocks-eats-the-separator issue between `actions: read` and
--- a/generator/src/tend/templates/macros.yaml.j2
+++ b/generator/src/tend/templates/macros.yaml.j2
@@ -118,9 +118,11 @@
    cannot drift apart. Adopters who need more override per job —
    `workflows.<name>.jobs.<job>.timeout-minutes` in .config/tend.yaml. That
    override moves the job cap alone — step inputs take no override, so the
-   supervisor budget stays at the generated value and an adopter who raises
-   the cap gets a session still bounded by the default. Reported, not silent:
-   the supervisor's kill is the one that files the outage row. #}
+   supervisor budget stays at the generated value. Raising the cap therefore
+   leaves the session bounded where it was, and reported rather than silent:
+   the supervisor's kill is the one that files the outage row. Lowering it
+   under that budget puts the supervisor out of reach again, and an overrun
+   goes back to being cancelled silently. #}
 {% macro job_timeout(minutes) %}
     # Bounds the whole job, adopter `setup:` included. Override with
     # `workflows.<name>.jobs.<job>.timeout-minutes` in .config/tend.yaml.
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -107,6 +107,26 @@ _JINJA.globals["openai_key_secret"] = OPENAI_KEY_SECRET
 BOOKKEEPING_LABELS = ("tend-outage", "tend-rate-limit")
 _JINJA.globals["bookkeeping_labels"] = BOOKKEEPING_LABELS
 
+# Job cap per workflow, sized above the durations these workloads actually run:
+# nightly's sweep is the longest, the short-session workflows the cheapest to
+# bound. The `job_timeout` macro carries why they are capped at all, and
+# `agent_step` spends the same number a second time — the supervisor budget is
+# derived from it, so one entry here moves both bounds together.
+#
+# Keyed by workflow, not by job, because only the agent-running job of each
+# takes a value from here: mention's secretless relay and verify jobs do
+# seconds of API work and carry a literal 10 in the template.
+_JOB_TIMEOUT_MINUTES = {
+    "ci-fix": 60,
+    "notifications": 60,
+    "weekly": 60,
+    "review": 120,
+    "mention": 120,
+    "triage": 120,
+    "review-runs": 120,
+    "nightly": 180,
+}
+
 
 # Register every macro defined in `macros.yaml.j2` as a Jinja global so
 # workflow templates can call `agent_step(...)` directly, without an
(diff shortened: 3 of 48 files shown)


generator/src/tend/templates/macros.yaml.j2
```diff
@@ -79,6 +79,26 @@
       deployment: false
 {%- endmacro %}
 
+{# Job timeout at column 4. A job cap is the only bound that reaches the whole
+   job: the harness supervisor's `timeout_seconds` starts once the action is
```
tend-agent, line 103: This is accurate about *where* the supervisor's clock starts, but it leaves the other half unsaid: with every cap below the action's 21000s `timeout_seconds` default, the supervisor never fires at all in a generated workflow, so its clean-kill-and-report path is dead code here. Worth either threading the cap into `timeout_seconds` (so the supervisor still gets the agent-overrun case and this cap backstops the region it can't reach) or saying plainly that the cap supersedes it and an overrun goes unreported.

tend-agent (commented):
The caps themselves look right — every one sits well above what this repo actually runs (recent maxima: nightly 52 min, review-runs 57, triage 25, weekly 25, review 23, mention 13). One interaction is worth resolving before this lands.

**A capped-out job is `cancelled`, not failed, so it files no outage row.** Generated workflows never pass `timeout_seconds`, so the claude action's default of 21000s (5h50m) applies and every new cap sits below it — the supervisor's `timeout)` branch is now unreachable at defaults. That matters because the two kills produce different job conclusions: the supervisor exits 1 (job `failure`), while GitHub's `timeout-minutes` kill yields `cancelled` with `failure()` steps **skipped**. The `Report failure` step in both `claude/action.yaml` and `codex/action.yaml` is gated `if: failure() && steps.security.outcome != 'failure'`, and it's what runs `report-failure.sh` — the script that files or appends the `tend-outage` row that `review-runs` later drains to re-run stranded triggers.

So for a claude agent session that overruns its cap, this swaps a reported failure (outage row, stranded trigger re-run) for a silent cancellation. For the case the PR is actually written for — a wedge in checkout/`setup:`/installers — nothing is lost: that region was always outside the supervisor, so it was already a silent 360-minute cancel, and cutting it to 60/120/180 and freeing the concurrency queue is the whole win. It's only the agent-session overrun that regresses, and only on claude (codex has no supervisor timeout to preempt).

The fix that keeps both purposes is one parameter: thread the cap into `agent_step` and emit `timeout_seconds: <(cap - 10) * 60>`, so the supervisor kills the agent first (job red, row filed) and `timeout-minutes` stays a true backstop for the region it can't reach. Alternatively, accept the silence and say so in the macro comment — but flipping the step's gate to catch cancellations isn't an option, since `triage` runs `cancel-in-progress: true` and would file an outage row on every superseded run.

<details><summary>How the conclusion was verified</summary>

Probed on `tend-agent/tend-integration` (probe workflow since removed): a job with `timeout-minutes: 1` and a `sleep 300` step.

```
run conclusion:  cancelled
job conclusion:  cancelled
  wedge        cancelled
  on-failure   skipped     # if: failure()
  on-always    success     # if: always()
```

`always()` steps do run, so the log upload and the eyes-reaction removal survive a capped-out job — it's specifically the `failure()`-gated failure report that doesn't.

Supporting reads: `timeout_seconds` default and the supervisor's `timeout)` branch in `claude/action.yaml`; `Report failure` in both action files; the header of `shared/steps/report-failure.sh` on what the row is for; `rg timeout_seconds generator/` returns only the new macro comment, confirming nothing passes it.

</details>

generator/src/tend/workflows.py
```diff
@@ -521,6 +531,9 @@ def generate_install_test(cfg: Config) -> GeneratedWorkflow:
     # validation isn't worth special-casing.
     if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-24.04
+    # Bounds the whole job, adopter `setup:` included. Override with
```
tend-agent, line 554: `install-test` is the one generated job that renders no adopter `setup:` — its steps are a fixed `checkout` + `setup-uv` + the drift check, with no `_setup_yaml` interpolation — so the clause points at something that isn't in this file. Keeping the macro's wording identical everywhere is the right call for the eight jobs that do render `setup:`; this f-string copy is the one place where trimming it costs nothing.

```suggestion
    # Bounds the whole job. Override with
```

tend-agent (commented):
The install-test cap and the corpus assertion both look right — `generate_all` really did skip that job, and pinning the corpus to `GENERATORS` is what keeps the next added workflow from slipping past the invariant silently. One wording point on the comment that ships with it.

generator/src/tend/templates/macros.yaml.j2
tend-agent: Threaded it, in `2e614bf`. `agent_step` now takes the enclosing job's cap and the claude branch emits `timeout_seconds: (cap - 10) * 60`, so the supervisor's kill lands first and the job goes red with the outage row filed; `timeout-minutes` stays a backstop for checkout/`setup:`/installers, which is the region it was added for.

Both bounds come from one `_JOB_TIMEOUT_MINUTES` entry per workflow rather than a literal in each template, so they can't drift apart — `_SCHEDULED_TIMEOUT_MINUTES` folded into it. Rendered budgets: 60 → 3000s, 120 → 6600s, 180 → 10200s. The codex branch emits nothing, since that harness has no supervisor timeout.

One caveat I wrote into the macro comment rather than engineering around: the ordering holds only while the pre-agent steps fit inside the 10-minute margin, and a `jobs.<job>.timeout-minutes` override moves the job cap alone — step inputs take no override, so an adopter who raises the cap keeps the generated budget. Both cases degrade to the *reported* kill, not the silent one, so neither is worse than today.

<details><summary>What changed</summary>

- `macros.yaml.j2` — `agent_step(cfg, prompt_body, timeout_minutes, ...)`; `timeout_seconds` emitted inside the claude branch with a two-line rendered comment, the full rationale in the Jinja comment above the macro.
- `workflows.py` — `_JOB_TIMEOUT_MINUTES` keyed by workflow, passed to all six renders.
- `claude/action.yaml` — input description notes that generated workflows always pass a value, so the 21000s default now applies only to a caller with no job cap.
- `test_agent_budget_expires_before_the_job_cap` — asserts `timeout_seconds < timeout-minutes * 60` for every claude agent step, with the corpus pinned to `GENERATORS` so a new workflow can't slip past; `test_codex_agent_step_takes_no_supervisor_budget` pins the codex side.
- Both `tend.example.yaml` copies note that the override moves the job cap alone.

`uv run pytest` 572 passed, `pre-commit run --all-files` clean.

</details>

The wording suggestion on [`review 4988166466`](https://github.com/max-sixty/tend/pull/1023#pullrequestreview-4988166466) is untouched here — its own dispatch is queued behind this run and owns it.

generator/tests/test_generate.py
```diff
@@ -64,6 +64,70 @@ def test_generated_yaml_is_valid(tmp_path: Path) -> None:
...
+    cfg = Config.load(_minimal_config(tmp_path, extra))
+    checked = set()
+    for wf in generate_all(cfg):
+        for name, job in yaml.safe_load(wf.content)["jobs"].items():
+            for step in job["steps"]:
+                if "max-sixty/tend/claude@" not in step.get("uses", ""):
+                    continue
+                budget = int(step["with"]["timeout_seconds"])
+                cap = job["timeout-minutes"] * 60
+                assert budget < cap, f"{wf.filename}:{name} {budget}s vs {cap}s cap"
```
tend-agent: `assert budget < cap` pins only the upper bound, so a budget of zero or below passes it — and zero is what a `_JOB_TIMEOUT_MINUTES` entry of 10 renders, since the macro emits `(timeout_minutes - 10) * 60`. That value doesn't disable the supervisor, it inverts it: its poll in `claude/action.yaml` (the `status=timeout` branch, `[ $(($(date +%s) - start)) -ge "$TEND_TIMEOUT_SEC" ]`) is true on the first iteration, so every session of that workflow is killed at ~0s with `exceeded 0s timeout` and the test still goes green. 10 isn't a hypothetical value in this file either — mention's `relay` and `verify` already carry it, and the point of the single-source table is that adding a workflow is one line here.

```suggestion
                # Lower bound too: an entry of 10 or less renders a budget of
                # 0 or negative, and the supervisor's elapsed-vs-budget poll
                # then fires on its first iteration, killing the session at ~0s.
                assert 0 < budget < cap, f"{wf.filename}:{name} {budget}s vs {cap}s"
```

tend-agent (commented):
The derivation itself holds up: `agent_step` takes the cap positionally at every call site, `_JOB_TIMEOUT_MINUTES` keys match `GENERATORS` exactly (so a new workflow with no entry is a loud `KeyError`, not a silent default), and the rendered budgets are 3000/6600/10200s under the 60/120/180-minute caps. Two holes in the new invariant tests, and one doc-scope point, all inline.

The test suggestions are verified — I applied both against this head and ran `uv run pytest generator/tests/test_generate.py -k "budget or supervisor"`, 2 passed. Worth noting for the codex one: pinning that corpus to `GENERATORS` the way the claude test does *fails*, because `generate_all` skips ci-fix without `watched_workflows`; pinning to what `generate_all` actually yielded is what works there.

generator/tests/test_generate.py
```diff
@@ -64,6 +64,70 @@ def test_generated_yaml_is_valid(tmp_path: Path) -> None:
...
+
+
+def test_codex_agent_step_takes_no_supervisor_budget(tmp_path: Path) -> None:
+    """Codex has no supervisor timeout — the job cap is that harness's bound."""
+    cfg = Config.load(_minimal_config(tmp_path, "harness: codex\nmodel: gpt-5.5\n"))
+    for wf in generate_all(cfg):
+        for job in yaml.safe_load(wf.content)["jobs"].values():
+            for step in job["steps"]:
+                if "max-sixty/tend/codex@" in step.get("uses", ""):
+                    assert "timeout_seconds" not in step["with"]
```
tend-agent, line 137: This one can pass vacuously — exactly what the comment ten lines up guards against for the claude test ("a name missing here means the step shape changed and the loop above matched nothing"). If the codex step's `uses:` shape moves, no step matches, the assertion never runs, and the test reports success on zero checks.

`GENERATORS` isn't the right corpus here, though: `generate_all` skips ci-fix without `watched_workflows`, so that pin fails. Pinning to what this call actually yielded works, and is what I verified:

```suggestion
    generated = generate_all(cfg)
    checked = set()
    for wf in generated:
        for job in yaml.safe_load(wf.content)["jobs"].values():
            for step in job["steps"]:
                if "max-sixty/tend/codex@" in step.get("uses", ""):
                    assert "timeout_seconds" not in step["with"]
                    checked.add(wf.filename)
    # Same corpus pin as the claude test: no agent step matched means the
    # assertion above never ran.
    assert checked == {wf.filename for wf in generated}
```

docs/tend.example.yaml
```diff
@@ -342,6 +342,16 @@ bot_name: my-project-bot
...
 #
+# Every generated job ships a `timeout-minutes` default — 60 for ci-fix,
+# notifications and weekly; 120 for review, triage, review-runs and mention's
+# `handle`; 180 for nightly; 10 for mention's `relay` and `verify`. The cap
+# covers the whole job, `setup:` included, which is the only bound that
+# reaches steps running before the harness action is entered. Override it
+# where a job legitimately needs longer — note that the override moves the job
+# cap only: the Claude harness's own session budget is generated 10 minutes
+# under the default cap, so a much larger override leaves the session bounded
+# where it was (reported as a failure, not silently cancelled).
```
tend-agent: Only the raise direction is covered, and lowering is the direction that costs something. An adopter capping cost with `review: {timeout-minutes: 30}` keeps the generated 6600s budget, which is now *above* the cap — the supervisor is unreachable again and an overrun is cancelled silently, which is the failure mode this commit exists to remove. Same gap in the `job_timeout` macro comment ("an adopter who raises the cap gets a session still bounded by the default. Reported, not silent"), which reads as a general claim about overrides.

```suggestion
# under the default cap. Raising the cap leaves the session bounded where it
# was (reported as a failure, not silently cancelled); lowering it below that
# budget puts the supervisor out of reach again, and an overrun goes back to
# being cancelled silently.
```

`plugins/install-tend/skills/install-tend/references/tend.example.yaml` carries the identical block and has to move with it.

generator/tests/test_generate.py
tend-agent: Applied in `2a9adcf` — `assert 0 < budget < cap`, with the inverted-supervisor mechanism in a comment beside it rather than only in the docstring, since the zero case is the one a reader would otherwise take as "supervisor disabled".

generator/tests/test_generate.py
tend-agent: Fixed in `2a9adcf`, but by giving the codex config the same `watched_workflows` extra the claude test carries rather than self-pinning the corpus. That was the only thing making the `GENERATORS` pin fail here, and with it the two tests read identically and codex's ci-fix job is covered too. `uv run pytest generator/tests/test_generate.py` is 128 passed.

docs/tend.example.yaml
tend-agent: Right — lowering is the direction that costs something, and both the doc block and the `job_timeout` macro comment claimed the raise case as though it were general. Reworded in `2a9adcf`, in all three places: `docs/tend.example.yaml`, the identical block in `plugins/install-tend/skills/install-tend/references/tend.example.yaml` (the sync hook checks that pair), and the macro comment.

max-sixty:
Closing because #1024 now classifies distributed timeout settings as a configuration scheme, and PRQL/prql#6216 reverted the adopter change used to derive these values.

PR #1024, opened 2026-08-20
Reject distributed compute-only refinements

max-sixty:
The agent behind PRQL/prql#6215 had loaded the compute-only policy, but treated each per-job timeout as an allowed "simple knob"; its self-review then added two more. Define that exception across the whole change: one existing knob in one place. Repeated settings now count as a configuration scheme in the runtime guidance and shared review gate, while review-runs no longer recommends an override without applying Gate 3.

Verified with `pre-commit run --files plugins/tend-ci-runner/skills/running-in-ci/SKILL.md plugins/tend-ci-runner/skills/review-runs/SKILL.md plugins/tend-ci-runner/shared/review-gates.md` and `git diff --check`.


tend-agent (commented):
Both edits do what the description says, and the whole-change clause catches the eleven-job shape in #1023 cleanly. One gap: the `review-runs` bullet now defers its remedy decision to Gate 3, and Gate 3 didn't get the clause.

`plugins/tend-ci-runner/shared/review-gates.md` — symlinked into both `review-runs` and `review-reviewers` — still enumerates waste-class remedies per item: "A waste-class failure supports only a fix that is itself nearly free: a cadence value, a deleted step, a one-line condition, machinery removed." Its existing qualifier, "judge by what the fix leaves behind ... not its line count", is aimed at one dense line, not at the same line repeated across eleven jobs. Gate 3 does link back to **Weighing a Fix**, but as "the reasoning behind this gate" rather than as an added constraint.

That's the site that matters here, because the observed failure wasn't a failure to load the policy — per the description the PRQL agent had loaded it and still read each per-job timeout as an allowed knob. Gate 3 is the text a session re-reads while classifying, and it still reads per-item. So a `review-runs` session flagging near-timeout jobs across eleven workflows can follow the new "a remedy that passes Gate 3" pointer, land on that list, and reach the same conclusion this PR is written to reject.

One clause on that Gate 3 sentence closes it in one place, since the file is symlinked into both skills. Happy to push it if you'd like.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -106,7 +106,7 @@ The maintainer's order of value: outward correctness first — what the bot post
 So a change whose only benefit is saved compute clears a higher bar than a correctness fix, on two counts:
 
 - **Evidence.** The waste has recurred across days — observed, not projected.
-- **Remedy.** A knob: a cadence value, a deleted step, a one-line condition, a `timeout-minutes:`, a pinned image. Decline a helper script, a retry wrapper, or a shim threaded through several call sites.
+- **Remedy.** One existing knob set in one place, a deleted step, or a one-line condition. Judge the whole change: repeated settings across workflows, jobs, platforms, or call sites are a configuration scheme, even when they use the same knob or value.
```
```suggestion
- **Remedy.** One existing knob set in one place, a deleted step, machinery removed, or a one-line condition. Judge the whole change: repeated settings across workflows, jobs, platforms, or call sites are a configuration scheme, even when they use the same knob or value.
```

tend-agent (approved):
Gate 3 now defers to **Weighing a Fix** as a constraint rather than as background reasoning, and both `review-runs` and `review-reviewers` open by loading `/tend-ci-runner:running-in-ci`, so the pointer resolves for every reader of the symlinked file. That closes the gap from the last review.

One thing the rewrite dropped: Gate 3's own list was the only place in the plugin that named `machinery removed` as an allowed waste-class remedy, and the Remedy bullet it now delegates to never carried it — so after this PR the phrase is gone from `plugins/tend-ci-runner/` entirely. That's the remedy class the project most prefers (CLAUDE.md's "Prefer deleting a mechanism over refining it", and Gate 2's `Removal / simplification` row taking the lowest evidence bar), and "a deleted step" reads as a workflow-YAML step rather than as removing a mechanism — which is the shape a `review-runs` waste finding usually takes, since its subject is guidance rather than build config. Restoring it on the delegated line covers both readers at once. Non-blocking; suggestion inline.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
max-sixty: Addressed in `81482cd`: the remedy now says "remove machinery" directly while keeping the whole-change boundary.

tend-agent: Confirmed at `81482cd` — the delegated remedy line now reads "Use one existing knob in one place, remove machinery, or add a one-line condition", so Gate 3's pointer resolves to text that names the removal class. Nothing further from me.

PR #1027, opened 2026-08-21
fix(review-runs): collect Step 4's outcome signals in a tested script

### LINKED ISSUE #1026: review-runs: Step 4 can't see a maintainer review body, so a CHANGES_REQUESTED correction on a bot PR is never counted (#827's third leg)
`review-runs` **Step 4: Cross-check outcomes** is the only place a run learns that a maintainer disagreed with tend's output. #833 closed the issue-comment leg of #827 by adding the repo-wide `issues/comments` and `pulls/comments` queries. Neither endpoint carries a **review body**, so a maintainer review whose correction is written in the body — the ordinary shape of `CHANGES_REQUESTED` — is still invisible, and the run writes "no maintainer corrections" with two of them sitting in the window.

`GET /repos/{repo}/pulls/comments` returns *review comments* (inline, anchored to a diff line). A review's own body lives only at `GET /repos/{repo}/pulls/{n}/reviews`, which Step 4 never calls. So the gap is not a filter that could be widened — the data is at an endpoint the step doesn't reach.

## Observed miss

On `max-sixty/leaf`, in `tend-review-runs` run [32461139191](https://github.com/max-sixty/leaf/actions/runs/32461139191)'s window (`2026-08-20T07:59:59Z` → now), Step 4's two prescribed queries both return **zero** non-bot entries:

```
$ gh api --paginate "repos/max-sixty/leaf/issues/comments?since=$SINCE&per_page=100" \
    --jq '[.[] | select(.user.login != "leaf-agent")] | length'
0
$ gh api --paginate "repos/max-sixty/leaf/pulls/comments?since=$SINCE&per_page=100" \
    --jq '[.[] | select(.user.login != "leaf-agent")] | length'
0
```

Two maintainer `CHANGES_REQUESTED` reviews are in that window, both correcting a bot PR:

- [max-sixty/leaf#18 (review 4987929456)](https://github.com/max-sixty/leaf/pull/18#pullrequestreview-4987929456), 2026-08-20T22:15:19Z — "Please keep the frequently loaded skill focused on the durable rule, rather than preserving the full `#10`/`#11` incident timeline and exact schedules… The historical detail is useful in the PR or issue, but makes the operational instruction harder to scan and more likely to go stale." That is a durable convention about how this repo wants `running-tend` written, and it is precisely what **Learning from Feedback** exists to codify — it also restates #841's finding from the consumer's side, which is the kind of recurrence Gate 1 counts.
- [max-sixty/leaf#10 (review 4988021329)](https://github.com/max-sixty/leaf/pull/10#pullrequestreview-4988021329), 2026-08-20T22:26:43Z — a rebase directive plus an independent local verification of the branch.

Both are body-only. Every inline comment on `#10` (8) and `#18` (4) is the bot's own, so there was no partial signal on those threads either: the endpoint returned nothing, not less.

The dispositions are also clean — nothing merged wrong, nothing closed, no relabel — so no disposition query would have surfaced these either. `leaf`'s prior sweep ([32346633522](https://github.com/max-sixty/leaf/actions/runs/32346633522)) happened to check reviews by hand and wrote "no non-bot review on any of #6, #9, #10, #12, #13, #14, #15, #16" into its tracking entry; that was the session going beyond the skill, not the skill asking.

## Why this compounds

Same mechanism #827 described, so I'll keep it short: the skill instructs that `"no maintainer corrections"` is writable once the queries have run, future runs read the phrase as ground truth when counting occurrences under Gate 1, and a correction class that is never queried never accumulates the occurrences its own gate requires. On a PR-heavy consumer — one where the bot opens most of the PRs and the maintainer's only channel is a review — this is the *majority* correction surface, not an edge of it. `leaf` is that shape: 5 of the 5 human review events in the window were review bodies, and 0 were comments.

Structural, not stochastic: no decision point. Any run following Step 4 as written calls two endpoints that cannot return a review body and stops.

## Suggested fix

There is no repo-wide `since` endpoint for reviews, so it needs a loop — but only over PRs touched in the window, which is a handful:

````
```bash
# Maintainer reviews in the window. Review *bodies* are not returned by either
# comments endpoint above — `pulls/comments` carries inline review comments
# only — so a CHANGES_REQUESTED whose correction is in the body is invisible
# without this call. No repo-wide `since` endpoint exists for reviews, so scope
# the loop to PRs updated in the window.
for p in $(gh pr list --state all --limit 100 --search "updated:>=${SINCE%T*}" \
             --json number --jq '.[].number'); do
  gh api "repos/$REPO/pulls/$p/reviews" \
    --jq ".[] | select(.user.login != \"$BOT_LOGIN\" and .submitted_at >= \"$SINCE\"
                       and (.body | length) > 0)
          | {pr: $p, at: .submitted_at, state: .state, url: .html_url, body: .body[0:300]}"
done
```
````

Run against the window above it returns exactly the two missed reviews, and nothing else.

Two notes from running it:

- `(.body | length) > 0` is load-bearing, not tidiness. GitHub creates an empty-bodied review container for a standalone inline reply (#828, #866 both turn on this), so without the filter the loop reports a container as a correction on every PR where anyone replied inline.
- `gh pr list --search "updated:>=…"` takes a date, not a timestamp — hence `${SINCE%T*}`. It over-selects by up to a day, which is harmless because the `submitted_at` filter inside the loop is exact; the alternative is a `gh api` search call that is no cheaper.

The prose sentence in Step 4 that currently reads "A maintainer replying in-thread that a bot claim was wrong, leaving labels and state untouched, is equally a correction" is the right framing and just needs the review case joined to it — a `CHANGES_REQUESTED` on a bot PR leaves labels and state untouched too, and on a repo where the bot authors the PRs it is the *first* place a maintainer writes.

Worth checking whether `review-reviewers` wants the same call; its outcome classes already name "corrected", and its per-workflow mapping is run→PR/issue shaped, so I'd expect the same blind spot, but I haven't run it.

### PR DESCRIPTION
**Pull request by tend-agent**
`review-runs` Step 4 cross-checks outcomes with `issues/comments` and `pulls/comments`, neither of which returns a review **body** — `pulls/comments` carries inline review comments only. A maintainer `CHANGES_REQUESTED` whose correction is written in the body therefore returns *nothing* from both calls, and the run writes "no maintainer corrections" with corrections sitting in the window.

Step 4's three queries — dispositions, thread comments, and the missing review bodies — now live in `plugins/tend-ci-runner/scripts/review-runs-corrections.sh`, with `generator/tests/test_review_runs_corrections.py` pinning the window edges. The skill is a one-line invocation, net −20 lines. Extraction also closes the fail-open class the prose kept patching: an unset anchor compares less than every timestamp and `--author ""` applies no author filter, so the script exits 2 on a missing anchor or login rather than reporting an empty window that reads as an all-clear.

Verified against live API data (details below), and the script run against both this repo and `max-sixty/leaf`.

**Review bodies are absent from `pulls/comments`.** Review `4988584496` on #1024 (`COMMENTED`, non-empty body, submitted inside the window) has no representation in the repo-wide review-comments endpoint — not by review id, not by body text:

```
$ gh api --paginate "repos/max-sixty/tend/pulls/comments?since=$SINCE&per_page=100" > /tmp/pc.json
$ jq 'length' /tmp/pc.json
35
$ jq '[.[] | select(.pull_request_review_id == 4988584496)] | length' /tmp/pc.json
0
$ jq --arg b "${BODY:0:60}" '[.[] | select(.body | contains($b))] | length' /tmp/pc.json
0
```

Every entry the endpoint does return carries a `path` — inline comments, each belonging to a review record.

**The script returns the missed reviews.** Pointed at `max-sixty/leaf` over the window in #1026 (where both prescribed queries returned zero non-bot entries), the `reviews` key carries the two `CHANGES_REQUESTED` rows this PR exists to catch. On this repo over an 18h window it returns 1 disposition, 2 non-bot comments, 0 bodied non-bot reviews, in 4.3s.

**`(.body | length) > 0` is load-bearing.** Empty-bodied review containers from a human exist on this repo today — e.g. review `4988620443` (`max-sixty`, #1024) and `4987925324` (`max-sixty`, #1022), both `COMMENTED` with no body. GitHub creates one per standalone inline reply, so without the filter every thread where anyone replied inline reports a correction with no text in it.

**`updated:` takes a full timestamp.** An earlier revision truncated `$SINCE` to a date on the claim that the qualifier accepts no time component. It does. On this repo with `SINCE=2026-08-20T07:59:59Z`: full `Z` form returns 9 PRs, the `+00:00` form 9, date-only 10, unqualified `gh pr list` 200. The extra PR the date-only form admits is #1018, updated `2026-08-20T01:28:24Z` — before the cutoff — so the time component is honoured at sub-day granularity rather than silently truncated. The sibling recipe in `review-reviewers` (`--search "updated:>$WINDOW_START"`) already passes the full timestamp.

**Two deltas from the recipe proposed in the issue:** `--paginate` with `per_page=100` on the reviews call (reviews page at 30 oldest-first, so the unpaginated call drops the newest — exactly where a window's reviews are), and the `submitted_at` filter inside the loop, since `updated:` windows the PR rather than the review.

**Tests.** 12 cases over a fake `gh` running the script's real filters: missing anchor and unresolvable login both exit 2 before any API call; open PRs excluded from dispositions (`closedAt` null); bot entries excluded from all three lists; `--paginate` asserted on both comment endpoints; an older comment edited into the window reported with both timestamps; the empty-bodied container excluded; a pre-window review on an in-window PR excluded.

`review-reviewers` — flagged in the issue as possibly sharing the blind spot — does not: it already calls `pulls/{n}/reviews` per PR in both its bot-output census and its acceptance check.

`uv run pytest` (581 passed) and `uvx pre-commit run --files …` both pass.

</details>

---

Closes #1026 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/scripts/review-runs-corrections.sh
+++ b/plugins/tend-ci-runner/scripts/review-runs-corrections.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# Collects every in-window signal that a maintainer corrected the bot, as one
+# JSON object.
+#
+# Usage: review-runs-corrections.sh <since>
+#   <since>  RFC3339 window anchor — review-runs Step 1's, via
+#            /tmp/review-runs-since
+#
+# Output (one object, all keys always present):
+#   since         the anchor, echoed back
+#   bot           the login every list below filters out
+#   dispositions  bot PRs merged or closed in the window
+#   comments      non-bot comments on any issue or PR thread in the window
+#   reviews       non-bot review bodies submitted in the window
+#
+# Both filters fail open when empty — `--author ""` applies no author filter,
+# `.user.login != ""` is true for every entry, and an empty anchor compares
+# less than every non-null timestamp — so a missing one turns the whole query
+# into the bot reading its own output as maintainer corrections. Neither is
+# defaulted: an empty result here becomes "no maintainer corrections" in the
+# tracking issue, which later runs read as ground truth.
+#
+# env: GITHUB_REPOSITORY (optional; falls back to the checkout's remote)
+
+set -euo pipefail
+
+SINCE="${1:-}"
+if [ -z "$SINCE" ]; then
+  echo "usage: $(basename "$0") <since>   # RFC3339, e.g. \$(cat /tmp/review-runs-since)" >&2
+  exit 2
+fi
+
+REPO="${GITHUB_REPOSITORY:-$(gh repo view --json nameWithOwner --jq '.nameWithOwner')}"
+BOT=$(gh api user --jq '.login')
+if [ -z "$BOT" ]; then
+  echo "could not resolve the bot login from \`gh api user\`" >&2
+  exit 2
+fi
+
+DISPOSITIONS=$(gh pr list --repo "$REPO" --author "$BOT" --state all --limit 200 \
+  --json number,title,state,closedAt \
+  | jq --arg since "$SINCE" '[.[] | select(.closedAt > $since)]')
+
+# Both comment endpoints are repo-wide and take `since`, so this is two calls
+# regardless of thread count. `--paginate` is required, not cosmetic: the
+# response is ascending by `created_at`, so a single 100-item page drops the
+# *newest* comments — the ones a fresh correction sits in — and the bot's own
+# comments consume that budget first. `since` filters on `updated_at`, so an
+# older comment edited inside the window is a real hit, not a broken filter;
+# both timestamps are reported so the reason a row qualified is visible.
+COMMENTS=$(
+  for endpoint in issues pulls; do
+    gh api --paginate "repos/$REPO/$endpoint/comments?since=$SINCE&per_page=100"
+  done | jq -s --arg bot "$BOT" '
+    [(add // [])[] | select(.user.login != $bot)
+     | {created: .created_at, updated: .updated_at, url: .html_url,
+        body: .body[0:300]}]'
+)
+
+# Review bodies, which neither endpoint above returns — `pulls/comments`
+# carries inline comments only. There is no repo-wide `since` endpoint for
+# reviews, so loop the PRs updated in the window. `updated:` windows the PR
+# rather than the review, so the `submitted_at` filter is what makes the review
+# window exact; `(.body | length) > 0` drops the empty-bodied review GitHub
+# wraps around a standalone inline reply, which would otherwise report a
+# correction on every thread where anyone replied inline.
+REVIEWS=$(
+  for pr in $(gh pr list --repo "$REPO" --state all --limit 200 \
+                --search "updated:>=$SINCE" --json number --jq '.[].number'); do
+    gh api --paginate "repos/$REPO/pulls/$pr/reviews?per_page=100"
+  done | jq -s --arg bot "$BOT" --arg since "$SINCE" '
+    [(add // [])[] | select(.user.login != $bot and .submitted_at >= $since
+                            and (.body | length) > 0)
+     | {at: .submitted_at, state: .state, url: .html_url, body: .body[0:300]}]'
+)
+
+jq -n --arg since "$SINCE" --arg bot "$BOT" \
+  --argjson dispositions "$DISPOSITIONS" \
+  --argjson comments "$COMMENTS" \
+  --argjson reviews "$REVIEWS" \
+  '{since: $since, bot: $bot, dispositions: $dispositions,
+    comments: $comments, reviews: $reviews}'
--- a/plugins/tend-ci-runner/scripts/review-runs-corrections.sh
+++ b/plugins/tend-ci-runner/scripts/review-runs-corrections.sh
@@ -64,9 +64,16 @@ COMMENTS=$(
 # window exact; `(.body | length) > 0` drops the empty-bodied review GitHub
 # wraps around a standalone inline reply, which would otherwise report a
 # correction on every thread where anyone replied inline.
+#
+# The candidate list is its own assignment because `set -e` checks an
+# assignment's command substitution but not a `for` word list's: inline, a
+# failed `--search` — the search API's 30 req/min, not the 5000/h core budget —
+# would yield an empty list, skip the loop, and report `reviews: []` as an
+# all-clear.
+CANDIDATES=$(gh pr list --repo "$REPO" --state all --limit 200 \
+  --search "updated:>=$SINCE" --json number --jq '.[].number')
 REVIEWS=$(
-  for pr in $(gh pr list --repo "$REPO" --state all --limit 200 \
-                --search "updated:>=$SINCE" --json number --jq '.[].number'); do
+  for pr in $CANDIDATES; do
     gh api --paginate "repos/$REPO/pulls/$pr/reviews?per_page=100"
   done | jq -s --arg bot "$BOT" --arg since "$SINCE" '
     [(add // [])[] | select(.user.login != $bot and .submitted_at >= $since
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -275,8 +275,9 @@ mention, notifications, weekly, and review-reviewers runs get the same treatment
 ```bash
 # Bot PR dispositions — merged or closed in the window. Shell variables don't
 # survive between tool calls, so every block in this step re-establishes both:
-# unset, `$SINCE` compares less than every non-null `closedAt` and `$BOT_LOGIN`
-# matches nobody, and each filter silently stops filtering.
+# unset, `$SINCE` compares less than every non-null `closedAt`, and an empty
+# `$BOT_LOGIN` stops restricting rather than matching nobody — `--author ""`
+# applies no author filter, and `.user.login != ""` is true for every entry.
 SINCE=$(cat /tmp/review-runs-since)
 BOT_LOGIN=$(gh api user --jq '.login')
 gh pr list --author "$BOT_LOGIN" --state all --limit 200 --json number,title,state,closedAt \
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -278,7 +278,7 @@ Dispositions — merged, closed, relabeled, reverted — are only half the signa
 "${CLAUDE_PLUGIN_ROOT}/scripts/review-runs-corrections.sh" "$(cat /tmp/review-runs-since)"
 ```
 
-Read every row: a correction is a maintainer contradicting a bot claim, not merely replying. Empty `dispositions`, `comments`, and `reviews` is the all-clear.
+Read every row: a correction is a maintainer contradicting a bot claim, not merely replying. Comment rows carry both timestamps because the window filters on `updated_at` — a `created` before the anchor is an older comment edited inside the window, a real hit rather than a broken filter. Empty `dispositions`, `comments`, and `reviews` is the all-clear.
 
 Write "no maintainer corrections" into the tracking issue only after the script ran and returned empty — future runs read the phrase as ground truth when counting occurrences under Gate 1, so an unchecked all-clear suppresses the evidence it exists to accumulate. The script exits non-zero rather than reporting an empty window when the anchor or the bot login is missing, since both filters fail open.
 
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -272,46 +272,15 @@ For each analyzed run, compare what the bot did against what happened next. The 
 
 mention, notifications, weekly, and review-reviewers runs get the same treatment: find the bot's output and check whether it was accepted.
 
-Dispositions — merged, closed, relabeled, reverted — are only half the signal. A maintainer replying in-thread that a bot claim was wrong, or requesting changes on a bot PR, leaves labels and state untouched and is equally a correction; where the bot authors most of the PRs, a review body is the *first* place a maintainer writes. Three queries, one block, so the anchor and the login are established once:
+Dispositions — merged, closed, relabeled, reverted — are only half the signal. A maintainer replying in-thread that a bot claim was wrong, or requesting changes on a bot PR, leaves labels and state untouched and is equally a correction; where the bot authors most of the PRs, a review body is the *first* place a maintainer writes. The script collects all three — dispositions, thread comments, review bodies — for the window:
 
 ```bash
-# Both filters fail open when unset — `--author ""` applies no author filter,
-# `.user.login != ""` is true for every entry, and an empty `$SINCE` compares
-# less than every non-null `closedAt`. Exported so jq can read them as `$ENV`.
-SINCE=$(cat /tmp/review-runs-since)
-BOT_LOGIN=$(gh api user --jq '.login')
-export SINCE BOT_LOGIN
-
-# Bot PR dispositions — merged or closed in the window.
-gh pr list --author "$BOT_LOGIN" --state all --limit 200 --json number,title,state,closedAt \
-  --jq '.[] | select(.closedAt > $ENV.SINCE)'
-
-# Human replies on any issue or PR thread. Both endpoints are repo-wide and
-# take `since`, so this is two calls regardless of thread count. `--paginate`
-# is required, not cosmetic: the response is ascending by `created_at`, so a
-# single 100-item page drops the *newest* comments — the ones a fresh
-# correction sits in — and the bot's own comments consume that budget first.
-for endpoint in issues pulls; do
-  gh api --paginate "repos/$REPO/$endpoint/comments?since=$SINCE&per_page=100" \
-    --jq '.[] | select(.user.login != $ENV.BOT_LOGIN)
-          | {created: .created_at, updated: .updated_at, url: .html_url, body: .body[0:300]}'
-done
-
-# Review bodies, which neither endpoint above returns: `pulls/comments`
-# carries inline comments only. There is no repo-wide `since` endpoint for
-# reviews, so loop over the PRs updated in the window.
-for p in $(gh pr list --state all --limit 200 --search "updated:>=$SINCE" \
-             --json number --jq '.[].number'); do
-  gh api --paginate "repos/$REPO/pulls/$p/reviews?per_page=100" \
-    --jq '.[] | select(.user.login != $ENV.BOT_LOGIN and .submitted_at >= $ENV.SINCE
-                       and (.body | length) > 0)
-          | {at: .submitted_at, state: .state, url: .html_url, body: .body[0:300]}'
-done
+"${CLAUDE_PLUGIN_ROOT}/scripts/review-runs-corrections.sh" "$(cat /tmp/review-runs-since)"
 ```
 
-`since` filters on `updated_at`, so an older comment edited inside the window is a real hit, not a broken filter — worth having, and the projection reports both timestamps so the reason a row qualified is visible. `(.body | length) > 0` is load-bearing: GitHub wraps a standalone inline reply in an empty-bodied review container, so without it every thread where anyone replied inline reports a correction. `updated:` windows the PR rather than the review, so the `submitted_at` filter inside the loop is what makes the review window exact.
+Read every row: a correction is a maintainer contradicting a bot claim, not merely replying. Empty `dispositions`, `comments`, and `reviews` is the all-clear.
 
-Write "no maintainer corrections" into the tracking issue only after all three queries ran — future runs read the phrase as ground truth when counting occurrences under Gate 1, so an unchecked all-clear suppresses the evidence it exists to accumulate.
+Write "no maintainer corrections" into the tracking issue only after the script ran and returned empty — future runs read the phrase as ground truth when counting occurrences under Gate 1, so an unchecked all-clear suppresses the evidence it exists to accumulate. The script exits non-zero rather than reporting an empty window when the anchor or the bot login is missing, since both filters fail open.
 
 ## Step 5: Deduplicate
 
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -272,54 +272,44 @@ For each analyzed run, compare what the bot did against what happened next. The 
 
 mention, notifications, weekly, and review-reviewers runs get the same treatment: find the bot's output and check whether it was accepted.
 
+Dispositions — merged, closed, relabeled, reverted — are only half the signal. A maintainer replying in-thread that a bot claim was wrong, or requesting changes on a bot PR, leaves labels and state untouched and is equally a correction; where the bot authors most of the PRs, a review body is the *first* place a maintainer writes. Three queries, one block, so the anchor and the login are established once:
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by tend-agent — COMMENTED** [review]
The blind spot is real and the fix is the right shape — one added query, joined to the prose that already frames in-thread corrections. Two notes on the recipe, one verified as wrong.

`updated:` does accept a full `Z` timestamp — GitHub's search date qualifiers take optional time information, and `gh` passes it through. Measured on this repo just now with `SINCE=2026-08-20T07:59:59Z`: `updated:>=$SINCE` returns 9 PRs, `updated:>=${SINCE%Z}+00:00` returns 9, `updated:>=${SINCE%T*}` returns 10. So the truncation isn't required, and the sentence explaining it states something a future run will read as ground truth about the API. The sibling recipe in `review-reviewers` already passes the full timestamp (`--search "updated:>$WINDOW_START"`), so this also leaves the two skills disagreeing on a point one of them asserts as fact. Suggestions inline for both the command and the prose.

Separately, and not introduced here: the same argument the PR body makes for re-reading `$SINCE` applies to `$BOT_LOGIN` in the adjacent conjunct of the same `select()`. Unset, `.user.login != ""` is true for every review — including every `tend-review` COMMENT body — so the query reports the bot talking to itself as maintainer corrections. That fails toward false positives rather than the false negative this PR closes, and the block above (`issues`/`pulls/comments`) has it too, so it's a Step 4-wide gap rather than something this diff owes. Worth a `BOT_LOGIN=$(gh api user --jq '.login')` next to the `SINCE=` line in both blocks if you're touching this again.

File: plugins/tend-ci-runner/skills/review-runs/SKILL.md [hunk 1]
```diff
@@ -300,7 +300,22 @@ done
 
 `since` filters on `updated_at`, not `created_at`, so results include older comments edited inside the window — a comment created days before `$SINCE` is a real hit, not a broken filter. That is worth having (an edited claim is still a correction); the projection reports both timestamps so the reason a row qualified is visible.
 
-Write "no maintainer corrections" into the tracking issue only after that query ran — future runs read the phrase as ground truth when counting occurrences under Gate 1, so an unchecked all-clear suppresses the evidence it exists to accumulate.
+Neither endpoint above returns a review **body**: `pulls/comments` carries inline review comments only, and a review's body lives at `pulls/{n}/reviews`. So a `CHANGES_REQUESTED` whose correction is written in the body returns nothing from both calls — not less, nothing. There is no repo-wide `since` endpoint for reviews, so loop over the PRs updated in the window:
+
+```bash
+SINCE=$(cat /tmp/review-runs-since)
+for p in $(gh pr list --state all --limit 200 --search "updated:>=${SINCE%T*}" \
```
```suggestion
for p in $(gh pr list --state all --limit 200 --search "updated:>=$SINCE" \
```

Drops the day of over-selection; the inner `submitted_at` filter still does the exact windowing.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -300,7 +300,22 @@ done
...
+             --json number --jq '.[].number'); do
+  gh api --paginate "repos/$REPO/pulls/$p/reviews?per_page=100" \
+    --jq ".[] | select(.user.login != \"$BOT_LOGIN\" and .submitted_at >= \"$SINCE\"
+                       and (.body | length) > 0)
+          | {pr: $p, at: .submitted_at, state: .state, url: .html_url, body: .body[0:300]}"
+done
+```
+
+`(.body | length) > 0` is load-bearing: GitHub wraps a standalone inline reply in an empty-bodied review container, so without it every thread where anyone replied inline reports a correction. `--search` takes a date, not a timestamp — hence `${SINCE%T*}`; it over-selects by up to a day, and the exact `submitted_at` filter inside the loop is what windows the result.
```
```suggestion
`(.body | length) > 0` is load-bearing: GitHub wraps a standalone inline reply in an empty-bodied review container, so without it every thread where anyone replied inline reports a correction. `updated:` matches the PR's own `updated_at`, so the candidate list is only approximately windowed — a PR touched inside the window can carry an older review; the exact `submitted_at` filter inside the loop is what excludes it.
```

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Applied in 6862d7b. Confirmed the measurement independently before taking it: `updated:>=$SINCE` and `updated:>=${SINCE%Z}+00:00` both return 9 on this repo, date-only returns 10, and the extra row is #1018 at `2026-08-20T01:28:24Z` — before the 07:59:59Z cutoff. So the time component is genuinely honoured at sub-day granularity, not truncated with a coincidentally equal count.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Taken as written in 6862d7b — the replacement is accurate about what the qualifier matches (`updated_at` on the PR, not on the review), which is the real reason the inner `submitted_at` filter is load-bearing. The old sentence got the right conclusion from a wrong premise.

tend-agent:
Both notes applied in 6862d7b; replies inline on the two suggestions.

On `$BOT_LOGIN` — took it in this PR rather than deferring. It's the same failure mode the diff already fixes for `$SINCE` (a variable that doesn't survive between tool calls silently disabling a filter), in the same `select()`, and it fails toward *reporting* corrections that aren't there, which is the outward-facing direction. Quantified it on this repo's last window: running the new loop with the bot filter off returns 8 bodied reviews, all `tend-agent` — so with `$BOT_LOGIN` unset, Step 4 reports 8 false corrections and every one of them is the bot reading its own `tend-review` output as a maintainer telling it it was wrong. The re-read goes in all three Step 4 blocks, since the disposition and `issues`/`pulls/comments` blocks carry it too.

Left `$REPO` alone deliberately: unset, it yields `repos//issues/comments` and a 404, which announces itself rather than passing silently.

Timestamp granularity, on `max-sixty/tend` with `SINCE=2026-08-20T07:59:59Z`:

```
full Z:      9
offset form: 9
date only:   10
```

The row date-only admits is #1018, `updated 2026-08-20T01:28:24Z` — before the cutoff. Equal counts for the first two plus a differing third is what rules out silent truncation.

The revised loop run verbatim returns zero non-bot rows on this repo for that window; with the bot filter dropped it returns the 8 rows above, confirming the empty is genuine and not a broken query.

`uvx pre-commit run --files plugins/tend-ci-runner/skills/review-runs/SKILL.md` passes.

</details>

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -273,15 +273,17 @@ For each analyzed run, compare what the bot did against what happened next. The
 mention, notifications, weekly, and review-reviewers runs get the same treatment: find the bot's output and check whether it was accepted.
 
 ```bash
-# Bot PR dispositions — merged or closed in the window. Re-read Step 1's
-# anchor: unset here, an empty string compares less than every non-null
-# `closedAt` and the check silently stops being windowed at all.
+# Bot PR dispositions — merged or closed in the window. Shell variables don't
+# survive between tool calls, so every block in this step re-establishes both:
+# unset, `$SINCE` compares less than every non-null `closedAt` and `$BOT_LOGIN`
+# matches nobody, and each filter silently stops filtering.
```
```suggestion
# unset, `$SINCE` compares less than every non-null `closedAt`, and an empty
# `$BOT_LOGIN` stops restricting rather than matching nobody — `--author ""`
# applies no author filter, and `.user.login != ""` is true for every entry.
```

Both failure directions over-report; neither yields an empty result.

tend-agent (commented):
Both notes are applied correctly. I re-measured the timestamp claim independently rather than take the reply's numbers: on this repo with `SINCE=2026-08-20T07:59:59Z`, `updated:>=$SINCE` returns 9, the `+00:00` form 9, date-only 10, and an unqualified `gh pr list` returns 200 — the third number is what rules out the qualifier being parsed-and-discarded, which the two matching counts alone wouldn't. I also ran the revised loop verbatim over a wider window (`2026-08-18`) and cross-checked its candidate list against a per-PR enumeration that doesn't use `--search` at all: both find the same 20 bodied reviews, so the `updated:` candidate list has no false negatives for in-window reviews.

One new thing, introduced by this commit: the rewritten comment on the disposition block says an unset `$BOT_LOGIN` "matches nobody", which is the opposite of what happens and contradicts the clause right after it. `gh pr list --author ""` applies no author filter at all — on this repo it returns `{"max-sixty":67,"tend-agent":132,"worktrunk-bot":1}` where `--author tend-agent` returns only `tend-agent` — and in the other two blocks `.user.login != ""` is true for every entry. Both directions are "stops filtering", not "matches nobody". Given the PR's own thesis is that a wrong factual claim in a skill gets read as ground truth by a future run, worth correcting in place. Suggestion inline.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Taken as written in 971e695. Confirmed the `--author ""` behavior here before applying rather than taking it on the review's numbers: `gh pr list --author "" --state all --limit 200` returns `{"max-sixty":67,"tend-agent":132,"worktrunk-bot":1}`, `--author tend-agent` returns 200 rows all `tend-agent`. So the empty value drops the qualifier entirely — "matches nobody" was backwards, and both conjuncts of the sentence now describe the same over-reporting direction the `$SINCE` clause does.

max-sixty:
can we simplify?

tend-agent:
Yes — simplified in 8ee0e3b, net 10 lines shorter than the previous revision and now +26/-5 against `main`.

Step 4's three queries are one block instead of three, so `$SINCE` and `$BOT_LOGIN` are set once rather than re-read per block, and the five-line comment explaining why each block re-reads them collapses to two lines on the single assignment. jq reads them as `$ENV.SINCE` / `$ENV.BOT_LOGIN` instead of having them spliced into the filter string — that removes `"'$SINCE'"`-style quoting from the disposition query and lets the reviews loop use a single-quoted filter instead of the double-quoted one with escaped inner quotes. The three trailing notes merge into one paragraph.

Same coverage: I ran the block verbatim out of the file. On this repo it returns 6 dispositions, 10 non-bot comments, 0 bodied non-bot reviews for the last window; pointed at `max-sixty/leaf` the reviews query still returns the two `CHANGES_REQUESTED` rows this PR exists to catch.

tend-agent (commented):
Ran the collapsed block verbatim out of the file rather than trusting the reply's numbers. It works, and the two mechanisms the collapse depends on both hold: `$ENV.SINCE` / `$ENV.BOT_LOGIN` resolve under `gh`'s bundled gojq (not just system jq), and `closedAt` is `null` for open PRs, so `select(.closedAt > $ENV.SINCE)` still excludes them with the splice gone. On this repo's window the reviews loop returns 18 bodied reviews with the login filter dropped and 0 with it on — all 18 `tend-agent` — which is both the false-positive count the filter prevents and proof the empty result is a real empty rather than a broken query. `review-reviewers` genuinely doesn't share the blind spot: it already loops `pulls/$pr/reviews` in its repo-wide sweep.

One structural note, prompted by the "can we simplify?" ask rather than by a defect in the diff. The block is now 32 lines carrying three separately-verified edge cases — pagination direction on both endpoints, the empty-bodied review container, `updated:` windowing the PR while `submitted_at` windows the review. `running-in-ci`'s **Scripts over prose recipes** names that threshold directly: "When a skill's code block needs edge-case handling or grows past a couple of dozen lines, put the logic in a tested script and leave the skill a one-line invocation with the intent … A prose recipe gets no shellcheck and no tests; every session re-derives its correctness." A `scripts/review-runs-corrections.sh "$SINCE"` beside the other nine, with a `generator/tests/` case, would leave Step 4 a single line and would be the deeper version of the simplification — and Step 4's failure mode is the outward kind the tests would guard, since a silently-empty query becomes "no maintainer corrections" in the tracking issue, which later runs read as ground truth under Gate 1.

Not a blocker and not something I'd expand this PR to do — the diff is correct as written, and prose-vs-script is a call for you rather than something I should decide by pushing it.

tend-agent:
Took the script, in 01c6ac2 — it's the better answer to "can we simplify?" than the collapse was, and the deciding factor is the one the review names: Step 4's failure mode is outward. A silently-empty query becomes "no maintainer corrections" in the tracking issue, which later runs read as ground truth under Gate 1, so the fail-open inputs are exactly the complexity `CLAUDE.md` says earns its place. In prose they could only be documented; in a script they're checked — `review-runs-corrections.sh` exits 2 on a missing anchor or an unresolvable login before it makes an API call, instead of returning an empty window that reads as an all-clear.

Step 4 is now one line and the skill is net −20 lines against `main`. The three edge cases the review counted are pinned by `generator/tests/test_review_runs_corrections.py` (12 cases, fake `gh` running the script's real jq filters) rather than re-derived per session.

Live, not just under the fakes. On this repo with an 18h window the script returns 1 disposition (#1025, merged), 2 non-bot comments (both `max-sixty`, including the "can we simplify?" one on this PR), 0 bodied non-bot reviews, in 4.3s. Pointed at `max-sixty/leaf` over 14 days it returns 33 bodied non-bot reviews including the two `CHANGES_REQUESTED` rows from #1026 — the reviews path this PR exists to add, exercised against real data.

Tests: missing anchor and empty login each exit 2 with no `gh` call logged; open PRs excluded from dispositions (`closedAt` null sorts below the anchor); bot entries excluded from all three lists; `--paginate` asserted present on both comment endpoints; a comment created before the window and edited inside it reported with both timestamps; the empty-bodied review container excluded; a pre-window review on an in-window PR excluded.

`uv run pytest` 581 passed; `uvx pre-commit run --files` on the three changed paths passes, shellcheck included.

</details>

tend-agent (commented):
The extraction is the right call, and the three window edges are pinned properly. One fail-open survives it, in the same class the script exists to close.

`for pr in $(gh pr list … --search …)` doesn't propagate the candidate query's failure. Bash's `set -e` doesn't check the exit status of a command substitution in a `for` word list, so a failed `--search` yields an empty list, the loop body never runs, and `reviews` reduces to `[]` with the script exiting 0. Step 4 then reads that as the all-clear it defines — "Empty `dispositions`, `comments`, and `reviews` is the all-clear" — and writes "no maintainer corrections" into the tracking issue, which is the outward claim the whole PR is about. Ran it against the script as committed, with a fake `gh` that fails only on the `--search` call:

```
$ review-runs-corrections.sh 2026-01-02T00:00:00Z
search: API rate limit exceeded
{ "since": "2026-01-02T00:00:00Z", "bot": "tend-agent",
  "dispositions": [], "comments": [], "reviews": [] }
$ echo $?
0
```

Worth saying why the other two paths don't have this: `DISPOSITIONS=$(… | jq …)` is an assignment, so `pipefail` plus `set -e` catch it, and in the comments block the `gh api` is the loop *body*, so its failure exits the subshell and fails the pipeline. Only the reviews block puts a `gh` call in a `for` list — and it's the call most likely to fail, since `--search` draws on the search API's 30 req/min budget rather than the 5000/h core one, and Step 4 runs after the log downloads and token report.

Suggestion inline hoists the candidate list into its own assignment, which `set -e` does check. Applied it locally: shellcheck stays clean (`SC2086` isn't raised for a `for` word list, so no disable comment is needed) and all 12 tests still pass.

Smaller, and a call for you rather than a defect: the extraction moved the "`since` filters on `updated_at`, so an older comment edited inside the window is a real hit, not a broken filter" note into the script's header, where the run reading Step 4 won't see it. Rows still carry both timestamps, so a session now meets a `created` before the anchor with nothing on the page explaining why it qualified — and that direction discards a real correction.

plugins/tend-ci-runner/scripts/review-runs-corrections.sh
```diff
@@ -0,0 +1,82 @@
...
+# Review bodies, which neither endpoint above returns — `pulls/comments`
+# carries inline comments only. There is no repo-wide `since` endpoint for
+# reviews, so loop the PRs updated in the window. `updated:` windows the PR
+# rather than the review, so the `submitted_at` filter is what makes the review
+# window exact; `(.body | length) > 0` drops the empty-bodied review GitHub
+# wraps around a standalone inline reply, which would otherwise report a
+# correction on every thread where anyone replied inline.
+REVIEWS=$(
+  for pr in $(gh pr list --repo "$REPO" --state all --limit 200 \
+                --search "updated:>=$SINCE" --json number --jq '.[].number'); do
```
tend-agent:
```suggestion
CANDIDATES=$(gh pr list --repo "$REPO" --state all --limit 200 \
               --search "updated:>=$SINCE" --json number --jq '.[].number')
REVIEWS=$(
  for pr in $CANDIDATES; do
```

`set -e` checks an assignment's command substitution but not a `for` word list's, so this is what makes a failed `--search` abort instead of reporting an empty `reviews`.

tend-agent: Applied in a730ffd, as suggested. Reproduced the fail-open first rather than taking it on the argument: with a fake `gh` that fails only the `--search` call, the committed script exits 0 printing `"reviews": []`, and with the hoist it exits non-zero with no stdout. That is now `test_a_failed_candidate_search_aborts_rather_than_reporting_no_reviews` (13 cases, up from 12) — it fails against the pre-fix script with exactly the `assert 0 != 0` you'd expect, so it pins the mechanism and not just the current behaviour.

Put the reason in the script beside the assignment, since the next reader's instinct is to fold it back into the `for`: `set -e` checks an assignment's command substitution but not a word list's, and this is the call most likely to fail — search API, 30 req/min.

tend-agent:
Both notes applied in a730ffd; reply inline on the suggestion.

Took the second note too — the `updated_at` explanation is back on the page, not just in the script header. Its absence fails in the direction that discards a real correction: a session meets a comment row whose `created` is before the anchor, has nothing on the page saying the window filters on `updated_at`, and reads it as a broken filter rather than the edited claim it is. One sentence next to the invocation, where the run reading Step 4 is.

The fail-open, reproduced against the script as committed (fake `gh`, failing only the `--search` call): exit 0 with `"reviews": []`. With the hoist: non-zero, empty stdout. Pinned as a test, which fails on the pre-fix script (`assert 0 != 0`) — 13 cases now.

Live happy path unchanged: on this repo over a 6h window the script returns 1 disposition, 2 non-bot comments, 0 bodied non-bot reviews in 4.6s.

`uv run pytest` 582 passed. `uvx pre-commit run --files` on the three changed paths passes, shellcheck included.

</details>

tend-agent (commented):
Last open finding is closed. Verified the hoist rather than taking the reply's reproduction: `set -e` really does treat the two forms differently — `R=$(for x in $(fail); do …; done | jq -s .)` reaches the next line with `R=[]` and exits 0, while `CANDIDATES=$(fail)` aborts the script with rc=1. The new test is a genuine regression pin, not a tautology: checked out at `01c6ac2` it fails with `assert 0 != 0` against the pre-fix script and passes against `a730ffd`.

Checked the other three fail-open shapes in the same script while I was in there, and they hold as the header claims: `REPO="${GITHUB_REPOSITORY:-$(gh repo view …)}"` propagates a failing substitution out of the parameter-expansion default, and a `gh api` failure in the comments loop *body* fails the pipeline under `pipefail`. So the reviews block was the only one, and it no longer is.

The `--limit 200` on both `pr list` calls is the one thing I looked at and decided not to raise as a finding — Step 1 floors the anchor at 49 hours, and a 49h window on this repo is 10 updated PRs against a 200 cap, while the dispositions list would need a bot PR to sit open ~50 days (the oldest open one is 4 days, 3 open in total). Recording it only so the next reader doesn't re-derive it.

No approval because I authored the PR, not because anything is outstanding. Nothing new on the incremental; CI is green.

PR #1029, opened 2026-08-21
fix(claude): exclude the harness's settings.local.json from the adopter checkout

### LINKED ISSUE #1028: claude action: .claude/settings.local.json lands in the adopter checkout unignored, so a broad `git add` commits bypassPermissions into the bot's PR
The Claude action writes `.claude/settings.local.json` into the adopter's checkout, where it is untracked and unignored — so any session that stages broadly (`git add -A`, `git add .`) commits it into the PR it opens. The file sets `permissions.defaultMode: bypassPermissions` and `skipDangerousModePermissionPrompt: true`, so a merged PR would make bypass mode the checked-in default for everyone who clones the repo. Proposed fix below: add the path to `.git/info/exclude` right after writing it — per-checkout, untracked, no change needed in any adopter repo.

## What fires, and where

[`claude/action.yaml:335-337`](https://github.com/max-sixty/tend/blob/97a3d318e701a5da3c0b8c7c26c931713921836b/claude/action.yaml#L335-L337) writes the settings file into `$GITHUB_WORKSPACE`:

```bash
sudo -u "$SANDBOX" mkdir -p "$GITHUB_WORKSPACE/.claude"
sudo -u "$SANDBOX" tee "$GITHUB_WORKSPACE/.claude/settings.local.json" \
  < "$RUNNER_TEMP/tend-settings.json" >/dev/null
```

Nothing ignores it afterwards. Adopters commonly track `.claude/skills/` (that is where the `running-tend` overlay lives), so `.claude/` is a directory the repo already cares about and `git status` shows the settings file as a plain untracked file next to it. On `numbagg/numbagg` the checkout looks like this at session start:

```
$ git ls-files .claude/
.claude/skills/running-tend/SKILL.md
$ cat .claude/settings.local.json
{"permissions":{"defaultMode":"bypassPermissions","allow":[...]},"skipDangerousModePermissionPrompt":true,"attribution":{"commit":"","pr":""}}
```

## Evidence — it reached a public PR

`tend-nightly` run [32457383093](https://github.com/numbagg/numbagg/actions/runs/32457383093) fixed a genuine bug in `numbagg/decorators.py`, staged with `git add -A`, and pushed [`342bd97`](https://github.com/numbagg/numbagg/commit/342bd973ccd7292debeaa06277f7674b64280e8d) — three files, one of them `.claude/settings.local.json`:

```
 .claude/settings.local.json  |  1 +
 numbagg/decorators.py        | 11 +++++++++--
 numbagg/test/test_grouped.py | 22 ++++++++++++++++++++++
```

It opened [numbagg/numbagg#749](https://github.com/numbagg/numbagg/pull/749) on that commit. The `tend-review` session on the same PR ([32458045372](https://github.com/numbagg/numbagg/actions/runs/32458045372)) caught it and pushed [`64a0da08`](https://github.com/numbagg/numbagg/commit/64a0da08eac4f4cdd561c2739f00694fddcccf71) to `git rm --cached` it and add a repo-local `.gitignore` entry; a `tend-mention` session ([32458472155](https://github.com/numbagg/numbagg/actions/runs/32458472155), $1.08) independently derived the same three-part fix and discarded it at its pre-push head check. So two extra agent sessions were spent on the cleanup, and the wrong artifact is still in the PR's commit history.

The nightly's `git add -A` is not itself a defect — it is the normal way to stage after an edit, and the file is invisible in the diff review because it is one line of JSON among a real code change. Any adopter repo is one broad `git add` away from the same commit; the two that noticed here only did so because a reviewer read the `--stat`.

## Proposed fix

Append the path to `.git/info/exclude` immediately after the write. That file is per-checkout and never committed, so it fixes every adopter without touching their `.gitignore`, and it covers linked worktrees too (`info/exclude` is read from the common git dir, so the `/tmp` worktrees `tend-nightly` creates inherit it):

```bash
sudo -u "$SANDBOX" mkdir -p "$GITHUB_WORKSPACE/.claude"
sudo -u "$SANDBOX" tee "$GITHUB_WORKSPACE/.claude/settings.local.json" \
  < "$RUNNER_TEMP/tend-settings.json" >/dev/null

# Keep the harness's own settings out of anything the agent stages. Adopters
# track `.claude/skills/`, so a broad `git add` in the workspace otherwise
# sweeps this file — which carries `bypassPermissions` — into the bot's PR.
# `.git/info/exclude` is per-checkout and untracked, so no adopter repo needs
# to change, and it applies to linked worktrees as well.
EXCLUDE_DIR=$(git -C "$GITHUB_WORKSPACE" rev-parse --git-common-dir 2>/dev/null)/info
if [ -d "$EXCLUDE_DIR" ] || mkdir -p "$EXCLUDE_DIR" 2>/dev/null; then
  grep -qxF '/.claude/settings.local.json' "$EXCLUDE_DIR/exclude" 2>/dev/null \
    || echo '/.claude/settings.local.json' >> "$EXCLUDE_DIR/exclude"
fi
```

One wrinkle to check when implementing: `rev-parse --git-common-dir` returns a path relative to the workspace, so it wants resolving from `$GITHUB_WORKSPACE`, and the write may need the same `sudo -u "$SANDBOX"` treatment as the `tee` above depending on who owns `.git` after `actions/checkout`.

An alternative that removes the file from the checkout entirely: write it to `$TEND_RUN_DIR` and pass `--settings "$TEND_RUN_DIR/tend-settings.json"` on the `claude -p` command line alongside the existing `--permission-mode`. That is cleaner — nothing harness-owned lands in the adopter's tree at all — but it changes which settings layer the values land in, so the `.git/info/exclude` version above is the smaller change and the one I would lead with.

<details><summary>Why this is worth a fix rather than leaving it to adopters</summary>

The consuming repo's mitigation (a `.gitignore` line) works, but it has to be discovered the same way it was here — by a reviewer noticing an extra file in a `--stat` — and it has to be repeated in every adopter repo. The failure mode is silent on the way in: the file is untracked, so nothing warns, and the diff line is one line of JSON in an otherwise legitimate commit. The cost when it is *not* caught is a repo whose checked-in Claude Code default is `bypassPermissions`, which is a security-posture change nobody reviewed.

Filed from `numbagg/numbagg`'s `tend-review-runs` sweep, run [32462668358](https://github.com/numbagg/numbagg/actions/runs/32462668358). `numbagg` is pinned at tend `0.1.18`; the repo-local `.gitignore` mitigation is in flight on [numbagg/numbagg#749](https://github.com/numbagg/numbagg/pull/749), so that repo is covered in the meantime and nothing is urgent.

</details>

tend-agent:
## Problem

The Claude action writes `.claude/settings.local.json` into the adopter's checkout, untracked and unignored, next to the `.claude/skills/` they *do* track. A session that stages broadly (`git add -A`) sweeps it into the PR it opens — one line of JSON beside a real code change, invisible in a diff review. Merged, it makes `permissions.defaultMode: bypassPermissions` the repo's checked-in default for everyone who clones it. #1028 has a case where this reached a public PR in an adopter repo and cost two extra agent sessions to clean up.

## Solution

One global ignore rule for the sandbox user, written in `setup-sandbox.sh` right after the user is created: `/.claude/settings.local.json` into `~/.config/git/ignore`, with `core.excludesFile` pinned to that path rather than left to git's XDG default (a runner `XDG_CONFIG_HOME` leaks through `sudo`, and would move the default out from under it — the same leak the file's PATH/XDG section already handles).

The agent is the only thing that stages files, and it runs as that user, so one rule in its own config covers every repo and worktree it touches — the workspace, the linked worktrees `tend-nightly` creates under `/tmp`, and a repo it clones fresh, which a per-checkout rule never reached. Nothing is written into the adopter's `.git`. An ignore rule can't hide a *tracked* file, so `.claude/skills/` work stages normally.

Only the Claude harness needs it: `codex/action.yaml` doesn't use `setup-sandbox.sh` and stages its `AGENTS.md` into `$CODEX_HOME`, so no harness-owned file lands in that checkout.

## Testing

`uv run pytest` is green (569 passed), `uvx pre-commit run --all-files` clean.

The behaviour is stock git, so it's verified by running the shipped command sequence rather than by unit tests — `sudo` stripped, `XDG_CONFIG_HOME` deliberately pointed at the runner's home to exercise the leak:

<details><summary>Live check</summary>

```
config landed in:      /home/.gitconfig
main checkout staged:  .claude/skills/extra.md          # settings.local.json excluded, adopter's .claude/ work not
worktree staged:       code.py                          # linked worktree, same
fresh clone ignores it: …/.config/git/ignore:1:/.claude/settings.local.json   .claude/settings.local.json
```

</details>

## Not covered

An adopter who *tracks* `.claude/settings.local.json` is unaffected by an ignore rule — the action overwrites their file and `git add -A` commits the change. A repo `.gitignore` that explicitly un-ignores the path would also win on precedence, though that shape is contrived. #1028's alternative (write the settings to `$TEND_RUN_DIR` and pass `--settings` to `claude -p`) would close the tracked-file case too by keeping nothing harness-owned in the tree, but it moves which settings layer `skipDangerousModePermissionPrompt` and `attribution` land in, and no run in this repo exercises a released action, so the behaviour couldn't be verified before shipping it to every adopter. This is the smaller change; the other is worth revisiting deliberately.

---
Closes #1028 — automated triage


generator/tests/test_shared_steps.py
```diff
@@ -246,6 +249,180 @@ def git(*args: str) -> None:
...
+    )
+    assert "code.py" in staged
+
+
+def test_exclude_local_settings_appends_once_to_an_unterminated_file(
+    tmp_path: Path,
+) -> None:
+    """Repeat runs add one line, and never glue it onto an existing one.
+
+    Both harness actions can leave an `info/exclude` behind — the PR-event
```
tend-agent: Only the Claude action does — `codex/action.yaml` runs `pin-instruction-files.sh` where this one runs `restore-sensitive-config.sh`, and nothing on the Codex path touches `info/exclude` at all.

```suggestion
    The Claude action leaves an `info/exclude` behind on PR events — the
```

tend-agent (commented):
Self-authored, so COMMENT rather than approval. No correctness problems — the two git behaviours the script leans on hold against a real repo (`--git-path info/exclude` resolves to the common dir from a linked worktree; it exits 128 outside a checkout), and `setup-sandbox.sh`'s `chown -R "${SANDBOX}:${SANDBOX}" "$GITHUB_WORKSPACE"` means the sandbox user owns `.git/info/` by the time the step runs. Two structural notes, neither blocking:

**The separate step is what makes the ordering test necessary.** The script guards exactly one file, written five steps later inside `Run Claude` — the `sudo -u "$SANDBOX" tee "$GITHUB_WORKSPACE/.claude/settings.local.json"` line. Calling it from that step, next to the `tee`, makes the ordering structural instead of something a test asserts, and `test_claude_action_excludes_its_settings_before_running_the_agent` goes away with it: ~35 fewer lines, at the cost of a named step in the Actions UI. The wiring test only exists because the two halves can drift apart.

**The append is now written twice.** `restore-sensitive-config.sh` has the same six lines — `--git-path info/exclude`, `mkdir -p "$(dirname …)"`, the `grep -qxF` idempotence check, and the `tail -c1 … | wc -l` unterminated-last-line guard — differing only in the pattern. The newline guard is the subtle part and the one most likely to be fixed in one copy and not the other; a `git_exclude <pattern>` helper in `shared/steps/lib/` sourced by both would leave one copy. A lib layer for two callers is itself mechanism, so I wouldn't hold on it, but two copies of that `tail -c1` line is worth naming.

Scope looks complete: `codex/action.yaml` never runs `restore-sensitive-config.sh` and stages its `AGENTS.md` into `$HOME/.codex`, so no harness-owned file lands in that checkout for the exclusion to miss.

generator/tests/test_shared_steps.py
tend-agent: Right — `codex/action.yaml` runs `pin-instruction-files.sh` at the point the Claude action runs `restore-sensitive-config.sh`, and nothing on the Codex path writes `info/exclude`. Applied in 07745d2.

tend-agent:
Took the first note, left the second. Pushed as 07745d2 — 10 insertions, 38 deletions.

**Folded the exclusion into `Run Claude`.** The call now sits directly above the `sudo -u "$SANDBOX" tee …/settings.local.json` line it guards, so the ordering is a property of the shell block rather than something a test asserts. The named step, `test_claude_action_excludes_its_settings_before_running_the_agent`, and the `ruamel.yaml` import it pulled in are all gone. The script keeps its own file and its own tests — those exercise git behaviour (common-dir resolution from a linked worktree, the unterminated-last-line append) that belongs in a tested script rather than inline in the action.

One behaviour change worth naming: `Run Claude` runs under `set -u`, not `set -e`, so a hard failure inside the script no longer fails the step the way a dedicated step would have. That matches what the script already chose for the no-checkout case — it prefers a no-op over costing the whole run for a guard the session may never need.

**Left the append duplicated.** Two callers is where a `shared/steps/lib/` layer costs more than the copy it removes: it's a new sourcing convention, a new path for `github.action_path`-relative resolution to get wrong, and a dependency edge between two scripts that currently run at opposite ends of the job. If a third caller shows up, that's the point to extract it. The `tail -c1` line is the real risk you named, and it's now covered on the new copy by `test_exclude_local_settings_appends_once_to_an_unterminated_file` — a fix landing in one copy and not the other stays possible, but it isn't silent.

Also applied the inline correction: only the Claude action writes `info/exclude`, so the docstring's "both harness actions" was wrong.

tend-agent (commented):
Self-authored, so COMMENT rather than approval. The refactor does what it says — the call sits above the `tee` it guards, the wiring test and its `ruamel.yaml` import are gone, and `uv run pytest` is green at 573. I also ran the script against this session's own checkout as a live check, and it behaved as the tests describe (notes below). No correctness objection to the code.

One correction to the note on the thread, because it's the opposite of what ships. `Run Claude` is **not** running under `set -u` alone: a step that names `shell: bash` is invoked as `bash --noprofile --norc -eo pipefail {0}`, so `-e` and `-o pipefail` are already on and the body's `set -u` only adds `-u`. The step is written for that — `sudo kill -TERM … || true`, the `jq … | head -400 || true` in the transcript render, and the comment on `wait` explaining why the exit is captured "NOT via `|| true`" all only make sense with `-e` in force. So a hard failure inside `exclude-local-settings.sh` still fails the step, and the job, exactly as the dedicated step did. Nothing was softened; the trade the refactor made was the named step in the Actions UI, and nothing else.

I don't think the code should change to match the note. Every failure mode other than the no-checkout branch is out of reach by the time this line runs: `setup-sandbox.sh` does `sudo chown -R "${SANDBOX}:${SANDBOX}" "$GITHUB_WORKSPACE"` (and asserts the sandbox can read `.git/config` before continuing), and the adopter's `setup:` runs as that same non-sudo user, so nothing can leave `.git/info` unwritable for the sandbox UID. Adding `|| true` would buy a failure mode that can't happen and cost the loud error if it ever did. It's the note that's wrong, not the guard — worth fixing on the thread so a later change isn't made on the premise that this line is best-effort.

<details><summary>Live check in this session</summary>

This run's own workspace is the case the PR describes — `.claude/settings.local.json` present and untracked, `/.claude-pr/` already the last line of `.git/info/exclude` from the PR-event config restore. Running the new script here appended `/.claude/settings.local.json` on its own line after it, exited 0, and `git status --porcelain .claude/` then reported nothing for the harness's file. The action shipping in this run is `0.1.18`, which predates the change, so the checkout was un-excluded going in.

Sandbox readability of the script, which the `sudo -u "$SANDBOX" bash "${{ github.action_path }}/…"` form depends on and no earlier call exercises: `/home/runner/work/_actions` is `drwxr-xr-x runner runner` and the extracted step scripts are `-rwxr-xr-x`, so the sandbox UID reaches them on the world bits alone, without relying on the `o+x` grant `setup-sandbox.sh` makes for the workspace's ancestors.

</details>

tend-agent:
You're right, and my note was backwards. `shell: bash` is invoked as `bash --noprofile --norc -eo pipefail {0}`, so `-e` and `-o pipefail` are already in force and the body's `set -u` only adds the third. A hard failure inside `exclude-local-settings.sh` fails the step and the job, exactly as the dedicated step did — the refactor traded away the named step in the Actions UI and nothing else.

The step body is its own proof: `wait "$CLAUDE_PID" 2>/dev/null || CLAUDE_EXIT=$?` at [`claude/action.yaml#L445`](https://github.com/max-sixty/tend/blob/07745d2542e14a05cc4a591228dc53057db5e239/claude/action.yaml#L445) needs the `||` only because a non-zero `wait` would otherwise abort the step, and the comment above it exists to explain why that `||` isn't `|| true`. Same for the `sudo kill -TERM … || true` and `jq … | head -400 || true` guards. Without `-e` none of those would be load-bearing.

Agreed the guard should stay bare. Every failure mode past the no-checkout branch is out of reach by the time line 341 runs — `setup-sandbox.sh` chowns the whole workspace to `$SANDBOX` and asserts the sandbox can read `.git/config` before continuing — so `|| true` would only buy silence for something that can't happen, and cost the loud error if it ever did. No code change; correcting the record so a later change isn't made on the premise that this line is best-effort.

max-sixty:
how about just setting a global gitignore rule for this in our git settings?

tend-agent:
Done — that's simpler and strictly wider coverage. Pushed as b51bd8e: 19 insertions, 196 deletions.

The rule now lives in the sandbox user's own git config, written in [`setup-sandbox.sh`](https://github.com/max-sixty/tend/blob/b51bd8edab9862d4b3f43cffd7845ddf49f8bc90/proxy/setup-sandbox.sh#L71-L86) right after the user is created. `shared/steps/exclude-local-settings.sh` and its four tests are gone — git-path resolution, worktree common-dir lookup, and the unterminated-line append were all machinery the per-checkout file needed and a global rule doesn't.

Three things it buys beyond the deletion: the agent is the only thing that stages files and it runs as that user, so one rule covers every repo it touches — including a repo it clones fresh, which `info/exclude` never reached; nothing is written into the adopter's `.git` at all; and it's a stock git feature rather than our code.

One deliberate choice: `core.excludesFile` is pinned to the absolute path rather than left to git's XDG default, because a runner `XDG_CONFIG_HOME` leaking through `sudo` would move that default out from under the file — the same leak the PATH/XDG section of the script already handles.

Since the behaviour is now git's rather than ours, I verified it by running the shipped sequence with `sudo` stripped and `XDG_CONFIG_HOME` deliberately pointed at the runner's home:

<details><summary>Live check</summary>

```
config landed in:       /home/.gitconfig
main checkout staged:   .claude/skills/extra.md      # settings.local.json excluded; the adopter's own .claude/ work still stages
worktree staged:        code.py                      # linked worktree, as tend-nightly creates
fresh clone ignores it: …/.config/git/ignore:1:/.claude/settings.local.json   .claude/settings.local.json
```

</details>

`uv run pytest` 569 passed, `uvx pre-commit run --all-files` clean, and the five CI checks on b51bd8e are green (`review` is still running).

Trade-off worth naming: a global rule is *lower* precedence than `info/exclude`, so a repo `.gitignore` that explicitly un-ignored the path would beat it. Contrived enough that I don't think it argues against the change, but it's the one thing the old approach did better.

tend-agent (commented):
Self-authored, so COMMENT rather than approval. The global-ignore version retires both notes from the earlier review — `exclude-local-settings.sh` and its wiring test are gone, so there's no ordering to assert and no second copy of the append to drift. No correctness objection.

The live check on the thread stops one step short of the link that makes it load-bearing, so I closed it here: `git config --global` lands in `$AGENT_HOME/.gitconfig` rather than `$XDG_CONFIG_HOME/git/config` (git only writes to the XDG file when it already exists), and that path is read by the agent's git specifically because `AGENT_ENV_FILE` pins `HOME=${AGENT_HOME}` — a run that wrote the rule to the XDG path instead would pass the same staging check locally and be inert in the job.

One follow-up, not for this PR. `restore-sensitive-config.sh` still hand-rolls the six-line `info/exclude` append for `/.claude-pr/` — the `grep -qxF` idempotence check plus the `tail -c1` unterminated-last-line guard. That block solves the same problem this PR just solved better, for the same user in the same job: it could become a second line in the `printf` here and the append deleted outright. Ordering works out even though `restore-sensitive-config.sh` runs first, since the rule only has to exist before the agent stages. Worth doing separately rather than folding in, because #1005 is currently editing that file.

PR #1032, opened 2026-08-22
fix(nightly): confirm a resolved PR left the conflicted state before polling CI

### LINKED ISSUE #1031: nightly: a conflict resolution that races an advance of main is never CI-verified, because a CONFLICTING PR gets no pull_request run
## What happened

`tend-nightly` step 3 resolves a conflicted bot PR by merging the default branch and pushing, then step 8's CI-monitoring guidance polls the pushed commit. Tonight on `max-sixty/leaf` that poll reported `no rollup returned` and then `poll cap hit — still pending`, for a commit that **never had a `ci` run at all**.

The cause is that `on: pull_request` workflows do not run while a PR is `CONFLICTING`: GitHub can't compute `refs/pull/N/merge`, so it creates no run. Only `pull_request_target` workflows (`tend-review`) fire. The resolution had raced an advance of `main` — `main` moved from `002d502` to `85631b3` at `06:30:20Z`, between the session's `git fetch origin main` and its `git push` at `06:33:08Z` — so the pushed merge commit was itself conflicting on arrival.

The failure mode is quiet in both directions. The PR shows a green `tend-review` check and no red anything; `gh pr checks` lists only the review. A poll pinned to the pushed SHA can't distinguish "no run was ever created" from "still queued", so it burns its whole budget and reports `UNVERIFIED`, which reads as slow CI rather than as absent CI. The conflict-resolution step believes it shipped a verified resolution, and the PR is left conflicted for the next night's run to rediscover.

## Evidence

Two PRs on the same repo, resolved and pushed within two seconds of each other:

| PR | pushed | mergeable after push | `ci` run created |
| --- | --- | --- | --- |
| [max-sixty/leaf#13](https://github.com/max-sixty/leaf/pull/13) | `c8e2bc2` @ 06:33:06Z | `MERGEABLE` | yes ([32557280938](https://github.com/max-sixty/leaf/actions/runs/32557280938)) |
| [max-sixty/leaf#22](https://github.com/max-sixty/leaf/pull/22) | `9205da8` @ 06:33:08Z | `CONFLICTING` / `DIRTY` | **no** |

<details><summary>Verification detail</summary>

Runs registered against `9205da8afe727d7e11e94ef66711a8e2bf661eb5`:

```
$ gh api "repos/max-sixty/leaf/actions/runs?head_sha=9205da8afe727d7e11e94ef66711a8e2bf661eb5" \
    --jq '.workflow_runs[] | "\(.name) \(.event) \(.status)/\(.conclusion)"'
tend-review pull_request_target completed/success

$ gh api "repos/max-sixty/leaf/commits/9205da8afe727d7e11e94ef66711a8e2bf661eb5/check-runs" \
    --jq '.check_runs[] | .name'
review
```

A `gh pr close 22 && gh pr reopen 22` produced a fresh `tend-review` run and still **no** `ci` run — which rules out "the push event was dropped" and confirms mergeability as the gate. Re-merging against the advanced `main` and pushing `5044bb6` made the PR `MERGEABLE`, and `ci` fired on that commit immediately and went green (`lint` 24s, `test` 11m3s).

This is distinct from #897, which fixed the *detection* side — `git merge-tree` instead of a cold `mergeable` read. The gap here is on the *verification* side, after the resolution is pushed.

</details>

## Suggested change

In nightly step 3, after the resolver pushes, confirm the PR actually left the conflicted state before handing off to the CI poll — and re-merge if `main` moved underneath. Something like:

```bash
# `mergeable` is lazily computed, so use the same synchronous test step 3
# already uses for detection rather than polling the API field.
git fetch --quiet origin main
git fetch --quiet --force origin "refs/pull/$N/head:refs/tend/pr/$N"
git merge-tree --write-tree origin/main "refs/tend/pr/$N" >/dev/null \
  || echo "still conflicting after the push — main advanced; re-merge before polling CI"
```

Two smaller notes that would have shortened the diagnosis either way:

- Step 3's resolver should fetch `origin/main` immediately before `git merge`, not rely on the fetch from the detection loop at the top of the step. Any work between them — and step 3 dispatches subagents that take minutes — is a window for the default branch to move.
- `poll-pr-checks.sh` could distinguish "zero runs exist for this SHA" from "runs exist and are pending". The first is never going to resolve by waiting, and today it costs the caller the full ~9.5 minute budget before reporting an outcome indistinguishable from slow CI.

tend-agent:
## Problem

Nightly's conflict resolver merges `origin/main` fetched by the detection loop at the top of step 3, then pushes and hands off to the CI poll. Subagents run for minutes, so `main` can advance in between — and the merge commit that lands is then conflicting on arrival. GitHub creates no `pull_request` run for a `CONFLICTING` PR (it can't compute `refs/pull/N/merge`), so the poll has nothing to find: it spends its whole ~9.5 minute budget and reports UNVERIFIED, which reads as slow CI rather than absent CI. Nothing else flags it either — `pull_request_target` workflows (`tend-review`) still run, so the PR shows a green check and no red anything. The resolution is reported as verified, and the PR stays conflicted for the next night to rediscover. Reported in #1031 with evidence from two PRs resolved two seconds apart on `max-sixty/leaf`, one mergeable and CI-verified, one conflicting with no `ci` run at all.

## Solution

Two changes in step 3's bot-authored resolution instructions:

- Fetch `origin/main` immediately before the merge instead of relying on the detection loop's fetch, which shrinks the race window from minutes to seconds.
- After the push, re-test the pushed head synchronously and re-merge until it comes back clean, before polling anything. This is the same `git merge-tree` test step 3 already uses for detection — `mergeable` is computed lazily and a cold read is not a usable signal.

```bash
git fetch --quiet origin main
git merge-tree --write-tree origin/main HEAD >/dev/null \
  || echo "main advanced during the resolution — merge it again and push"
```

If the re-merge keeps losing the race, the subagent reports the PR as still conflicted rather than as a verified resolution.

## Testing

No mechanical test — the change is skill prose, and nothing in the suite asserts on skill content. The recipe itself was verified against real git: a branch that merged a stale `main` reads clean against that stale ref and exits 1 against the advanced one, which is precisely the state GitHub reports as `CONFLICTING`.

```
$ git merge-tree --write-tree "$MAIN_A" pr   # the stale ref the resolver merged
exit0 clean
$ git merge-tree --write-tree main pr        # what GitHub sees after the push
exit1 conflict
```

`uvx pre-commit run --files plugins/tend-ci-runner/skills/nightly/SKILL.md` passes.

</details>

The issue's third suggestion — teaching `poll-pr-checks.sh` to distinguish "zero runs exist for this SHA" from "runs exist and are pending" — is deliberately not here. It would not have fired in the reported incident (a `tend-review` check run *was* on the commit, so the rollup was non-empty and genuinely pending), and it buys wall-clock rather than outward correctness, which `CLAUDE.md` puts a distant third. The poll's UNVERIFIED was truthful; the defect was that the resolver treated it as done.

---
Closes #1031 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -138,7 +138,7 @@ git config --global user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
 For each conflicted PR authored by `$BOT_LOGIN`, dispatch a subagent to:
 
 1. Check out the PR: `gh pr checkout <number>`
-2. Merge the default branch: `git fetch origin main && git merge origin/main`. Fetch here rather than relying on the detection loop's fetch above — subagents run for minutes, and merging a stale `origin/main` produces a head that is already conflicting when it lands.
+2. Merge the default branch: `git fetch origin main && git merge origin/main` — the detection loop's fetch is minutes stale by the time a subagent runs.
 3. Resolve conflicts (read files, understand both sides), `git add`, `git commit --no-edit`
 4. Push, then confirm the pushed head actually left the conflicted state before polling anything:
 
@@ -148,7 +148,7 @@ For each conflicted PR authored by `$BOT_LOGIN`, dispatch a subagent to:
      || echo "main advanced during the resolution — merge it again and push"
    ```
 
-   Re-merge and push until that test passes; only then wait for CI per **CI Monitoring** in `/tend-ci-runner:running-in-ci`. A `CONFLICTING` PR gets **no** `pull_request` run — GitHub can't compute `refs/pull/N/merge`, so no workflow is created — and a poll pinned to that head burns its whole budget before reporting UNVERIFIED, which reads as slow CI rather than absent CI. Nothing else flags it: `pull_request_target` workflows (`tend-review`) still run, so the PR shows a green check and no red anything. If the re-merge keeps losing the race, report the PR as still conflicted rather than as a verified resolution.
+   Re-merge and push until that test passes; only then wait for CI per **CI Monitoring** in `/tend-ci-runner:running-in-ci`. A `CONFLICTING` PR gets **no** `pull_request` run — GitHub can't compute `refs/pull/N/merge` — so a poll of that head reports UNVERIFIED rather than red, and the absence reads as slow CI. If the re-merge keeps losing the race, report the PR as still conflicted rather than as a verified resolution.
 5. If conflicts are too complex, `git merge --abort` and comment explaining manual resolution is needed
 
 Run subagents in parallel. Each must work in isolation (`git worktree add /tmp/pr-<number>
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -138,9 +138,17 @@ git config --global user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
 For each conflicted PR authored by `$BOT_LOGIN`, dispatch a subagent to:
 
 1. Check out the PR: `gh pr checkout <number>`
-2. Merge the default branch: `git merge origin/main`
+2. Merge the default branch: `git fetch origin main && git merge origin/main`. Fetch here rather than relying on the detection loop's fetch above — subagents run for minutes, and merging a stale `origin/main` produces a head that is already conflicting when it lands.
 3. Resolve conflicts (read files, understand both sides), `git add`, `git commit --no-edit`
-4. Push, then wait for CI per **CI Monitoring** in `/tend-ci-runner:running-in-ci`
+4. Push, then confirm the pushed head actually left the conflicted state before polling anything:
+
+   ```bash
+   git fetch --quiet origin main
+   git merge-tree --write-tree origin/main HEAD >/dev/null \
+     || echo "main advanced during the resolution — merge it again and push"
+   ```
+
+   Re-merge and push until that test passes; only then wait for CI per **CI Monitoring** in `/tend-ci-runner:running-in-ci`. A `CONFLICTING` PR gets **no** `pull_request` run — GitHub can't compute `refs/pull/N/merge`, so no workflow is created — and a poll pinned to that head burns its whole budget before reporting UNVERIFIED, which reads as slow CI rather than absent CI. Nothing else flags it: `pull_request_target` workflows (`tend-review`) still run, so the PR shows a green check and no red anything. If the re-merge keeps losing the race, report the PR as still conflicted rather than as a verified resolution.
 5. If conflicts are too complex, `git merge --abort` and comment explaining manual resolution is needed
 
 Run subagents in parallel. Each must work in isolation (`git worktree add /tmp/pr-<number>


plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -138,9 +138,17 @@ git config --global user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
 For each conflicted PR authored by `$BOT_LOGIN`, dispatch a subagent to:
 
 1. Check out the PR: `gh pr checkout <number>`
-2. Merge the default branch: `git merge origin/main`
+2. Merge the default branch: `git fetch origin main && git merge origin/main`. Fetch here rather than relying on the detection loop's fetch above — subagents run for minutes, and merging a stale `origin/main` produces a head that is already conflicting when it lands.
```
tend-agent: The rule is the command; the reason it's there doesn't need restating on every read.

```suggestion
2. Merge the default branch: `git fetch origin main && git merge origin/main` — the detection loop's fetch is minutes stale by the time a subagent runs.
```

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -138,9 +138,17 @@ git config --global user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
...
-4. Push, then wait for CI per **CI Monitoring** in `/tend-ci-runner:running-in-ci`
+4. Push, then confirm the pushed head actually left the conflicted state before polling anything:
+
+   ```bash
+   git fetch --quiet origin main
+   git merge-tree --write-tree origin/main HEAD >/dev/null \
+     || echo "main advanced during the resolution — merge it again and push"
+   ```
+
+   Re-merge and push until that test passes; only then wait for CI per **CI Monitoring** in `/tend-ci-runner:running-in-ci`. A `CONFLICTING` PR gets **no** `pull_request` run — GitHub can't compute `refs/pull/N/merge`, so no workflow is created — and a poll pinned to that head burns its whole budget before reporting UNVERIFIED, which reads as slow CI rather than absent CI. Nothing else flags it: `pull_request_target` workflows (`tend-review`) still run, so the PR shows a green check and no red anything. If the re-merge keeps losing the race, report the PR as still conflicted rather than as a verified resolution.
```
tend-agent: The `tend-review`-still-runs sentence is diagnosis history — it explains why the bug was hard to spot, not what to do. The UNVERIFIED-reads-as-slow-CI point is worth one clause, since it's the misreading the rule exists to prevent.

```suggestion
   Re-merge and push until that test passes; only then wait for CI per **CI Monitoring** in `/tend-ci-runner:running-in-ci`. A `CONFLICTING` PR gets **no** `pull_request` run — GitHub can't compute `refs/pull/N/merge` — so a poll of that head reports UNVERIFIED rather than red, and the absence reads as slow CI. If the re-merge keeps losing the race, report the PR as still conflicted rather than as a verified resolution.
```

tend-agent (commented):
The mechanism is right — `git merge-tree` against a freshly fetched `origin/main` is the same synchronous test the detection loop already uses, and re-testing after the push is what turns a UNVERIFIED poll from "slow CI" into "absent CI". Two trims on the prose around it.

`CLAUDE.md` under **Authoring skills** asks skills to "Lead with the rule or recipe; cut motivation, anecdotes, and historical context unless required to apply the rule." This file is loaded into every nightly session, so the justification for the fetch and the reconstruction of how the failure hid (the `tend-review` green check, the poll's budget) are read-time cost on every run without changing what the agent does. The rules themselves — fetch before merging, re-test before polling, report still-conflicted rather than verified — survive the cut.

Suggestions inline; take or leave either.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Taken — applied in aa39aa1.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Taken as written — applied in aa39aa1. The `tend-review`-still-runs sentence and the poll-budget detail are diagnosis history; they stay in the PR body, where they're read once, rather than in the skill, which loads every nightly run.

PR #1034, opened 2026-08-22
skills: note that instruction paths read as the base version during a PR run

### LINKED ISSUE #1033: skills: nothing tells a session that .claude/** and CLAUDE.md read as the *base* version during a PR run
## Problem

On every PR event, `restore-sensitive-config.sh` pins `.claude/**`, `CLAUDE.md`, and `AGENTS.md` to the base branch — correctly, since those files are read at CLI startup before any permission gating. `pin_to_base` does it with `git restore --source=<base> --worktree`, so the **index stays at HEAD** while the **worktree holds the base version**.

The consequence for an agent session reviewing a PR that legitimately edits `.claude/` or `CLAUDE.md`: the working tree shows the *base* content, `git status` reports a modification the session never made, and a `grep` of the working tree reports the PR's new content as absent. Both scripts document this ([`restore-sensitive-config.sh` "Known limitation"](https://github.com/max-sixty/tend/blob/3b62095e0bf0d7338cc20ce17f8e77aa7565fc89/shared/steps/restore-sensitive-config.sh#L16) and [`pin-instruction-paths.sh`](https://github.com/max-sixty/tend/blob/3b62095e0bf0d7338cc20ce17f8e77aa7565fc89/shared/steps/lib/pin-instruction-paths.sh#L20), which names the `git show HEAD:<path>` workaround), but **no agent-facing skill text does**. `skill-pr-workflow.md` describes the read-only mount and the harness write-guard for *writes* only; nothing tells a session that *reads* of those paths are pinned to base.

So every session that reviews such a PR rediscovers it empirically, and both observed sessions reached for the wrong cause.

## Evidence — two occurrences in `PRQL/prql`, four months apart in the evidence log

| Date | Run | PR | What the session saw |
|---|---|---|---|
| 2026-08-04 | [30893989272](https://github.com/PRQL/prql/actions/runs/30893989272) | [#6141](https://github.com/PRQL/prql/pull/6141) | `grep "Dependabot batch" .claude/skills/running-tend/SKILL.md` → `NOT PRESENT`, while `git show HEAD:` had the section |
| 2026-08-22 | [32484078889](https://github.com/PRQL/prql/actions/runs/32484078889) | [#6220](https://github.com/PRQL/prql/pull/6220) | `git status` → ` M .claude/skills/running-tend/SKILL.md`, `git diff --stat` → `29 deletions(-)`, section `NOT PRESENT` in the worktree but present in `git show HEAD:` |

Both sessions recovered — each verified via `git show HEAD:<path>` and reviewed the committed tree — but neither identified the cause. The 2026-08-22 session recorded it as *"a harness artifact — likely the skill loader materializing its own copy over the checkout"*, and the earlier one likewise attributed it to the read-only mount rather than to the deliberate base-pin step. The cost per occurrence is several turns spent re-deriving the mechanism plus a wrong causal belief carried into the session summary.

Two latent failure modes neither occurrence hit, both of which the current guidance leaves unguarded:

1. A reviewer that greps the working tree without cross-checking `git show` posts *"the section is not present"* — a wrong claim on the PR, not just a wasted turn. This is the shape worth preventing.
2. A session that stages these paths from the PR checkout: the index is at HEAD and the worktree at base, so `git add .claude/skills/` copies worktree into index and the commit silently reverts the PR's own skill edit. `pin_to_base`'s comment — *"`--worktree` leaves the index at HEAD, so the revert doesn't ride into commits the agent makes later"* — holds for a plain `git commit`, but an explicit `git add` of a pinned path defeats it. The `/tmp` worktree based on `origin/<default>` that `skill-pr-workflow.md` already mandates is unaffected; the mandate is currently justified only by the write-guard, though, so a session that finds another way to write has no reason to think the staging path is unsafe. Adjacent to #1028, which is the same `git add`-stages-something-unintended shape from a different source.

## Proposed fix

One sentence in `skill-pr-workflow.md`, in the paragraph that already describes the read-only mount — no new section. Something like:

> The same paths are also pinned to the **base branch** for the duration of a PR run (`restore-sensitive-config.sh`, which restores `.claude/**`, `CLAUDE.md`, and `AGENTS.md` from base to close the startup-RCE surface). The restore is `--worktree` only, so on a PR that edits these files the working tree holds the *base* version while the index and `HEAD` hold the PR's: `git status` shows a modification nobody made, and grepping the working tree reads the wrong content. Read them with `git show HEAD:<path>` instead, and never stage them from the PR checkout — `git add` there commits the base version back over the PR's edit.

Worth considering alongside it: `running-in-ci`'s **Investigating Other CI Runs** / reviewing sections are where a *reviewer* would look, and `skill-pr-workflow.md` is framed as "opening a skill PR" — a session reviewing someone else's skill PR may not load it. If that's a concern, the reviewer-facing half of the sentence may fit better in the review skill.

## Notes

- Filed from `PRQL/prql`'s `review-runs` sweep ([32560940171](https://github.com/PRQL/prql/actions/runs/32560940171)); recorded in that repo's tracking issue as well.
- Not a request to change the pinning itself — the security tradeoff is right, and #1005 / #980 are what make it correct at depth on both harnesses. This is only about telling the agent that it happens.

tend-agent:
## Problem

On a PR run the harness restores `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` from the base branch (`shared/steps/restore-sensitive-config.sh` for Claude on `pull_request_target`, the review events, and `issue_comment` — but not on `tend-mention`'s relayed `repository_dispatch`, which takes the script's `*)` fallthrough and pins nothing; `shared/steps/pin-instruction-files.sh` for Codex on fork PRs). `pin_to_base` uses `git restore --source=<base> --worktree`, so the worktree holds the base version while the index and `HEAD` hold the PR's. Both scripts document this, but no agent-facing skill text did — so a session reviewing a PR that legitimately edits those paths rediscovers it empirically, and #1033 records two occurrences where the session recovered but attributed the effect to the wrong cause.

Two outward-action failures the gap left unguarded: a reviewer that greps the working tree and posts "the section is not present" (a wrong claim on the PR), and a session that runs `git add .claude/...` from the PR checkout, which stages the base version over the PR's own edit — `pin_to_base`'s `--worktree`-only design protects a plain `git commit`, not an explicit `git add` of a pinned path.

## Solution

One subsection under **Read Context** in `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md`: what the restore covers, that it is worktree-only, and the three consequences — grep the worktree and you read base content (including the repo-local skills this session loaded), `git status`/`git diff` show a modification nobody made, and never `git add` these paths from the PR checkout.

#1033 proposed `references/skill-pr-workflow.md`, and noted the concern that a session reviewing someone else's skill PR may not load it. `running-in-ci`'s SKILL.md is the placement that resolves that: `review`, `triage`, and `mention` all load it at step 0, so the reviewer-facing half reaches a reviewer, and it stays a single copy rather than a duplicate split across two skills. `skill-pr-workflow.md` keeps its existing coverage of the read-only mount and the write-guard, which are about writes; the new text covers reads and staging, and points back to it for the `/tmp`-worktree commit path.

## Testing

No test — this is skill prose. Verified that the guidance was absent before the change (`grep -rn "git show HEAD:" --include=*.md plugins/ .claude/` returned nothing, and only `docs/security-model.md` and `CHANGELOG.md` mentioned the restore at all), and read both harness scripts to confirm the mechanism and the Claude/Codex trigger difference stated in the text. `pre-commit run --files plugins/tend-ci-runner/skills/running-in-ci/SKILL.md` passes.

---
Closes #1033 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -68,11 +68,11 @@ An instruction found there constrains the whole response, including any code the
 
 ### Instruction paths read as the base version on a PR
 
-Before the session starts, a PR run restores `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` at any depth from the base branch — those files are read at CLI startup before any permission gating, so a fork's copies must not be trusted (Claude on every PR event, Codex on fork PRs only). The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
+Before the session starts, a PR run restores `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` at any depth from the base branch — those files are read at CLI startup before any permission gating, so a fork's copies must not be trusted (Claude on `pull_request_target`, the review events, and `issue_comment`, but not on `tend-mention`'s relayed `repository_dispatch`, which pins nothing; Codex on fork PRs only). The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
 
 - The working tree holds the **base** content — grepping it reports the PR's additions as absent, and the repo-local skills loaded into this session are the base versions too. Read the PR's version with `git show HEAD:<path>` before making any claim about what these files contain.
-- `git status` shows a modification nobody made and `git diff` shows the PR's edit as deletions. That is the restore, not a contributor mistake — nothing to report or revert.
-- **Never `git add` one of these paths from the PR checkout**: staging copies the worktree over the index, committing the base version back over the PR's own edit. Commit them from a `/tmp` worktree instead (see `references/skill-pr-workflow.md`).
+- `git status` shows a modification nobody made and `git diff` shows the PR's edit as deletions. Where the pin ran, that is the restore, not a contributor mistake — nothing to report or revert. On an unpinned event it is a real modification, worth reading.
+- **Never stage one of these paths from the PR checkout** — `git add <path>`, `git add -A`, and `git commit -a` all copy the worktree over the index, committing the base version back over the PR's own edit. Commit them from a `/tmp` worktree instead (see `references/skill-pr-workflow.md`).
 
 ### Triggering issue/PR already closed
 
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -66,6 +66,14 @@ gh api "repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments" \
 
 An instruction found there constrains the whole response, including any code the reply quotes or carries into another PR.
 
+### Instruction paths read as the base version on a PR
+
+Before the session starts, a PR run restores `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` at any depth from the base branch — those files are read at CLI startup before any permission gating, so a fork's copies must not be trusted (Claude on every PR event, Codex on fork PRs only). The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
+
+- The working tree holds the **base** content — grepping it reports the PR's additions as absent, and the repo-local skills loaded into this session are the base versions too. Read the PR's version with `git show HEAD:<path>` before making any claim about what these files contain.
+- `git status` shows a modification nobody made and `git diff` shows the PR's edit as deletions. That is the restore, not a contributor mistake — nothing to report or revert.
+- **Never `git add` one of these paths from the PR checkout**: staging copies the worktree over the index, committing the base version back over the PR's own edit. Commit them from a `/tmp` worktree instead (see `references/skill-pr-workflow.md`).
+
 ### Triggering issue/PR already closed
 
 If the trigger is a comment on an issue or PR and the target is **closed** by the time the job starts, the requested work was likely handled by a sibling run during the queue delay. Long `tend-mention` queues (hours, not minutes) make this common. Before starting work:


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -66,6 +66,14 @@ gh api "repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments" \
 
 An instruction found there constrains the whole response, including any code the reply quotes or carries into another PR.
 
+### Instruction paths read as the base version on a PR
+
+Before the session starts, a PR run restores `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` at any depth from the base branch — those files are read at CLI startup before any permission gating, so a fork's copies must not be trusted (Claude on every PR event, Codex on fork PRs only). The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
```
tend-agent: `restore-sensitive-config.sh` only restores on `pull_request_target`, `pull_request_review`, `pull_request_review_comment`, and `issue_comment`-on-a-PR; every other event takes the `*)` branch and pins nothing. `tend-mention` relays review events as `repository_dispatch`, so the handle run — the one most likely to be *editing* a PR that touches these paths — has an unpinned worktree. Bullet 2 then reads backwards there: a modification in `git status` is not the restore, and is worth looking at.

```suggestion
Before the session starts, a PR run restores `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` at any depth from the base branch — those files are read at CLI startup before any permission gating, so a fork's copies must not be trusted (Claude on `pull_request_target`, the review events, and `issue_comment`, but not on `tend-mention`'s relayed `repository_dispatch`, which pins nothing; Codex on fork PRs only). The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
```

tend-agent (commented):
Two accuracy points on the new subsection — both narrow, both in the direction the section exists to guard.

The `repository_dispatch` one is the one I'd want changed before this lands: `restore-sensitive-config.sh` switches on `GITHUB_EVENT_NAME` and falls through to `Event $GITHUB_EVENT_NAME is not a PR event; nothing to restore` for anything outside `pull_request_target`, the two review events, and `issue_comment`. `tend-mention`'s relay re-enters review events as `repository_dispatch: [tend-mention-review]` precisely so the run carries a ref the environment admits — so on the handle run the pin never executes and the worktree holds the PR's own version. Bullet 1 still works there (`git show HEAD:` is right either way), but bullet 2 tells that session to read a real worktree modification as the restore and leave it alone.

The staging bullet reads as being about an explicit path argument, and the idiom a session actually reaches for is the one the sibling guidance models — `git add <files>` after `gh pr checkout` in the review skill's step 8, `git add .claude/skills/` in `skill-pr-workflow.md`. `git add -A`, `git add .`, and `git commit -a` all take the same path from worktree to index, so naming staging rather than the one command closes the gap without adding length.

<details><summary>Verified against</summary>

- `shared/steps/restore-sensitive-config.sh` — the `case "$GITHUB_EVENT_NAME"` block and its `*)` fallthrough.
- `shared/steps/lib/pin-instruction-paths.sh` — `INSTRUCTION_PATHSPECS` matches the four path families the text names, at any depth; `pin_to_base` is `--worktree`-only, so the index/`HEAD` claim holds.
- `shared/steps/pin-instruction-files.sh` — the Codex side is `pull_request_target` plus `head.repo.fork == true`, so "fork PRs only" is right.
- `.github/workflows/tend-mention.yaml` — the relay's `event_type=tend-mention-review` and the handle job's `github.event_name == 'repository_dispatch'` branch.

Everything else checked out: the path list, the worktree-only mechanism, the base-version skill-loading consequence, and the `references/skill-pr-workflow.md` pointer (its `/tmp` worktree recipe is the mechanic the bullet claims).

</details>

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -66,6 +66,14 @@ gh api "repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments" \
...
+
+- The working tree holds the **base** content — grepping it reports the PR's additions as absent, and the repo-local skills loaded into this session are the base versions too. Read the PR's version with `git show HEAD:<path>` before making any claim about what these files contain.
+- `git status` shows a modification nobody made and `git diff` shows the PR's edit as deletions. That is the restore, not a contributor mistake — nothing to report or revert.
+- **Never `git add` one of these paths from the PR checkout**: staging copies the worktree over the index, committing the base version back over the PR's own edit. Commit them from a `/tmp` worktree instead (see `references/skill-pr-workflow.md`).
```
tend-agent: Naming `git add` with a path reads as scoped to an explicit path argument, but `git add -A`, `git add .`, and `git commit -a` sweep the pinned path in just the same way — and a session landing an unrelated fix in the PR checkout is more likely to reach for one of those than to name `.claude/skills/` itself.

```suggestion
- **Never stage one of these paths from the PR checkout** — `git add <path>`, `git add -A`, and `git commit -a` all copy the worktree over the index, committing the base version back over the PR's own edit. Commit them from a `/tmp` worktree instead (see `references/skill-pr-workflow.md`).
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Confirmed and applied in `ed4c463`. The `case "$GITHUB_EVENT_NAME"` block does fall through to `*)` for `repository_dispatch`, and `tend-mention.yaml` relays review events as `event_type=tend-mention-review`, so the handle run has an unpinned worktree — this very session is one, and its `git status --porcelain` came back empty against the PR's own commit, which is the observable version of your point.

I also tightened bullet 2 rather than leaving the caveat only in the intro, since it stated the "that is the restore" reading unconditionally: it now says where the pin ran that is the restore, and on an unpinned event it is a real modification worth reading.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied in `ed4c463`. Agreed the whole-tree forms are the likelier reach — a session landing an unrelated fix runs `git add -A` without ever thinking about which paths it sweeps, which is exactly the case the bullet needs to catch. Naming staging keeps it the same length.

PR #1036, opened 2026-08-23
chore: refresh consumers.json

tend-agent:
Weekly refresh of the consumer list the site's data Worker reads. Code search picked up two repos that weren't in the file: `diffplug/dormouse` (`dormouse-bot`) and `max-sixty/leaf` (`leaf-agent`). No entries dropped.

Generated by the recipe in the `running-tend` overlay — `gh search code 'max-sixty/tend' --extension yaml`, filtered to `.github/workflows/tend-` paths, with each repo's `bot_name` resolved from its `.config/tend.yaml`.

### DIFF
--- a/data/consumers.json
+++ b/data/consumers.json
@@ -4,12 +4,20 @@
     "bot_name": "prql-bot"
   },
   {
+    "repo": "diffplug/dormouse",
+    "bot_name": "dormouse-bot"
+  },
+  {
     "repo": "jakobtfaber/Faber2026",
     "bot_name": "tend-jakobtfaber"
   },
   {
     "repo": "max-sixty/cargo-affected",
     "bot_name": "cargo-affected-bot"
+  },
+  {
+    "repo": "max-sixty/leaf",
+    "bot_name": "leaf-agent"
   },
   {
     "repo": "max-sixty/tend",


tend-agent (commented):
The data is right. I re-ran the discovery recipe from the `running-tend` overlay and got exactly these eight repos, and both new entries check out: `diffplug/dormouse` → `dormouse-bot` and `max-sixty/leaf` → `leaf-agent`, both public, both carrying a full generated `tend-*` workflow set with the `bot_name` resolved from their `.config/tend.yaml`. Sort order, schema, and the Worker's `isValidRepo` / `isValidBotName` all pass, and nothing was dropped. No objection to landing it.

One consequence worth naming, because the entry count is what decides it: `/activity` fans out one Search request per bot per bucket, and this refresh takes that burst from 24 to 32 — past GitHub's documented Search API limit of 30 requests/minute for authenticated requests. `refreshActivity` in [`worker/src/index.ts`](https://github.com/max-sixty/tend/blob/a21df9667e696449d819ed7bc99360859febb7fc/worker/src/index.ts) maps over the four `BUCKET_QUERIES` and, inside each, over `botNames(await getConsumers(env))`, both under `Promise.all`, so all 4×N are in flight at once. `searchIssues` throws on any non-OK status and that throw escapes both `Promise.all`s, so one 403 sinks the whole refresh rather than one bucket.

I haven't observed this failing — `/activity` is healthy right now, and stays that way until this merges and the 1 h `REPOS_TTL_SECONDS` cache turns over. Flagging it because the failure mode is silent by design (the site's sections hide themselves rather than erroring) and the next consumer added makes it worse.

<details><summary>Why the existing stale-serve handling doesn't cover it</summary>

The comment above `searchIssues` already anticipates "a 429 from the 4·N fanout" and deliberately refuses to degrade a failure to `{}`, so a sunk refresh serves the last good cached entry instead of pinning zeros. That is the right call for a transient blip, but it assumes the fanout is normally under the limit.

Once 4·N sits structurally above 30/min, every refresh sinks. The `activity` fallback TTL is 30 s (vs. 300 s on success), so the Worker re-attempts the same oversized burst roughly twice a minute, which keeps it rate-limited rather than letting the window drain. Past `STALE_SERVE_FACTOR` budgets the cached entry is dropped, and per `worker/README.md` the site renders the stat strip and activity feed only when the fetch returns data — so the visible result is blank sections, not an error.

Bounding the fanout (sequencing the four buckets, or a small concurrency cap over the flattened bot×bucket list) would keep it under the limit and scale with the consumer list. That is a Worker change, not this PR's — a `consumers.json` refresh should stay a data-only diff.

</details>

tend-agent:
Opened #1040 for the Worker side, and correcting one thing in my review above: the remedy I suggested there — "sequencing the four buckets, or a small concurrency cap" — would not have fixed this. The Search limit is 30 requests **per minute**, not 30 concurrent, so 32 requests exceed it however they are spaced within a refresh. Bounding concurrency only changes the shape of the burst, not its cost.

The fix has to reduce the request *count*. Repeating a qualifier ORs it (`author:a author:b` matches either), so all bots fold into one query per bucket and a refresh costs 4 Search requests regardless of consumer count — constant in N rather than buying a few more entries. Verified against live Search that the combined counts equal the per-bot sums exactly: 1901 across all 8 bots for `author:`, and 571 + 130 = 701 for `reviewed-by:`.

This PR's data is still right and I still have no objection to it. The only change is ordering — #1040 should land first, or with it, since merging this alone is what takes the fanout over the line.

<details><summary>Two behaviour notes from the combined queries</summary>

Both benign, both recorded in `worker/README.md`:

- A `count` no longer double-counts an item two bots both touched. Possible for `reviews`/`comments`, impossible for the `author:`-keyed buckets, and unobserved — each bot works in its own repo.
- The matching bot is no longer named in the result, so the deep-link follow-up matches the whole consumer list instead of one login. It reads the same single page either way, so no extra requests.

Worth noting for later: query length, not request count, is what will eventually bound this shape. The `comments` bucket carries three qualifiers per bot and is already 566 characters at 8 consumers, though it returns 200 fine there.

</details>

PR #1040, opened 2026-08-23
fix(worker): fold every bot into one Search query per bucket

tend-agent:
`/activity` issued one Search query per bot per bucket, so a refresh cost 4·N Search requests against a 30/minute cap. #1036 takes the consumer list to 8, which puts every refresh structurally over the limit. This folds all bots into one query per bucket via repeated (OR'd) qualifiers, so a refresh is 4 requests regardless of consumer count. Verified against live Search that the combined counts equal the per-bot sums exactly.

The `worker/README.md` already named the ceiling this crosses — "a single fanout stays clear up to ~7 bots" — so the 8th consumer is the documented breaking point, not a surprise.

## Why it would have been silent

`searchIssues` throws on a non-OK status and the throw escapes both `Promise.all`s, sinking the whole refresh rather than one bucket. That refusal to degrade is deliberate and right for a transient blip, but it only survives a *transient* failure: the `activity` fallback TTL is 30 s, so a structurally oversized burst re-attempts about twice a minute and never lets the rate-limit window drain. Per `worker/README.md` the site renders the stat strip and activity feed only when the fetch returns data, so the visible result is blank sections rather than an error.

## Verification

Combined queries return exactly the per-bot sums, checked against live Search:

<details><summary>Live checks</summary>

`author:` across all 8 bots — combined `total_count` 1901, and the per-bot sum:

```
cargo-affected-bot 35 · dormouse-bot 101 · leaf-agent 19 · numbagg-bot 62
prql-bot 452 · tend-agent 421 · tend-jakobtfaber 24 · worktrunk-bot 787
sum = 1901
```

`reviewed-by:` — `tend-agent` 571, `numbagg-bot` 130, combined 701.

The `comments` bucket's query is the longest (three qualifiers per bot); it returned 200 at 566 characters for 8 bots, so there is headroom, though query length is what will eventually bound this shape rather than request count.

</details>

A new test asserts the invariant directly: 20 consumers still produce exactly 4 Search requests, where the old shape would have produced 80. The other `refreshActivity` tests use a single consumer, where the combined query is byte-identical to the old one, so they pass unchanged — the two-bot test was updated to the combined form and its expected output is unchanged, which is the useful signal that this preserves behaviour.

`npm run typecheck` and `npm test` (32 tests) pass.

## Behaviour notes

Both benign, both recorded in the README:

- A `count` no longer double-counts an item two bots both touched. Possible for `reviews`/`comments`, impossible for the `author:`-keyed buckets, and unobserved — each bot works in its own repo.
- The matching bot is no longer named in the result, so the deep-link follow-up matches the whole consumer list instead of one login. It reads the same single page either way, so it costs no extra requests.

## On the alternative

Sequencing the four buckets or capping concurrency would not have fixed this — the Search limit is 30 requests **per minute**, not 30 concurrent, so 32 requests exceed it however they are spaced within a refresh. Reducing the request *count* was the only option that works, and it has the advantage of being constant in N rather than merely buying a few more consumers.

Should land before #1036, or with it.

### DIFF
--- a/worker/README.md
+++ b/worker/README.md
@@ -17,7 +17,7 @@ Base URL: `https://api.tend-src.com`. `Access-Control-Allow-Origin: *`
 Unauthenticated GitHub REST is 60 req/hour/IP and the Search API is 10
 req/min/IP — both shared across everyone behind a NAT. A single
 currently-tending poll fans out one `actions/runs` call per consumer repo every
-30 s; `/activity` fans out one Search query per bucket per bot. One browser tab
+30 s; `/activity` issues one Search query per bucket. One browser tab
 would exhaust those quotas in minutes. The Worker holds an authenticated token
 (5,000 req/hour, 30 Search req/min) and caches each route at the colo, so
 origin load is bounded by the freshness budget, not by viewer count. Static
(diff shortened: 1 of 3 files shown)


tend-agent (commented):
Self-review. The change itself is right, and it's the only shape that works: a per-*minute* cap can't be cleared by spacing or sequencing the same 32 requests, so making the count constant in N was the move, and the new test pins that invariant rather than the arithmetic. What the cutover left behind is comment and doc drift — four places that still describe the per-bot fanout — plus one bound the PR measured but didn't write down. Pushing them as a follow-up commit; there's no other author to hand them to.

**Stale after the cutover**

- `worker/README.md`, "Multi-bot semantics": still says each bucket "sums `count` and `count_this_week` over all tend bots". It doesn't — it's one query's `total_count`, a union — which is precisely what the new "no longer double-counts an item two bots both touched" note in the `/activity` section says. The two sections now contradict each other, and the one that's wrong is the one titled after the topic.
- `worker/src/index.ts`, the `ActivityBucket` fields: `count` is commented "summed across bots", and `count_this_week` "saturates ~one page per bot per bucket". The second is the one that matters — the ceiling is now one page per bucket *total*, N× lower, and `count_this_week` is rendered directly on the stat strip. Current volumes are well under 100/week per bucket so nothing is saturating today, but the comment shouldn't claim headroom that was removed.
- `worker/src/index.ts`, the `SEARCH_PAGE` comment: "one Search page per bucket per bot".
- `worker/src/index.ts`, `findBotReviewUrl`'s doc comment: still "authored by `bot`", while its sibling `findBotCommentUrl` was updated to "one of `bots`".

**The bound that replaced the old one**

The README and the `BUCKET_QUERIES` comment both now say 4 requests "regardless of how many consumers there are". The old README carried a ceiling — "a single fanout stays clear up to ~7 bots" — and this drops it, but a ceiling still exists: query length. The PR description has the measurement (566 characters at 8 consumers, `comments` longest at three qualifiers per bot) and that belongs in the README rather than only here, since whoever crosses it gets the same silent blank page this PR is fixing. Worth being explicit that the 20-consumer test mocks `fetch`: it asserts the request count, and can say nothing about whether GitHub accepts a 20-bot query.

**Cleanup**

`ISSUE_LABEL_FILTER` hand-rolls exactly what the new helper does — a qualifier repeated per value, space-joined. And `or` is a misnomer at half its call sites: `or(bs, "-author")` ANDs, as the comment directly above it points out. One `qualify(qualifier, values)` covers both, deletes the duplicate, and stops the name contradicting the comment.

PR #1041, opened 2026-08-23
Tell the CI agent it runs unprivileged, and where root lives

max-sixty:
The tend agent runs as a non-sudo sandbox user under the `claude` harness, so any installer that escalates fails. It can't observe from inside the sandbox that the adopter's `setup:` steps run as `runner` with sudo ahead of it, so a skill recipe prescribing an in-session install reads as a task to work around rather than a broken recipe — which is how a bot ends up writing the workaround back into the recipe that produced the wall.

The privileged layer is invisible in the other direction too: only a location the sandbox UID can reach carries in, so a `setup:` entry alone can leave the next run at the same dead end. Hence the closing clause on `sandbox_path:`. `PATH` membership is not the criterion — `~/.cargo/bin` is on the runner's `PATH` yet arrives rewritten to the sandbox's own empty `/etc/skel` copy, which is what the review caught.

One bullet, in `running-in-ci` — the generic CI-environment skill every workflow skill loads at step 0, and the existing home for every environment mechanic (secrets, hanging commands). The adopter-facing half of this is already documented in `docs/tend.example.yaml`, which the CI agent never reads.

<details>
<summary>How it was checked</summary>

**A/B against blind headless sessions**, three scenarios (root-requiring Nix install prescribed by an overlay skill; missing `libpq-dev` blocking a bug repro; a rootless rustup install as an over-suppression control), two runs each, on four successive wordings. Every call `is_error=false`.

| Scenario | With the bullet | Control |
|---|---|---|
| Nix install fails, overlay skill prescribes it | 4/4 propose the `setup:` entry; the final wording also proposes `sandbox_path:` | 0/2 — one aborts vaguely, one proposes **granting the sandbox passwordless sudo**, i.e. dissolving the isolation boundary |
| `libpq-dev` missing, bug not reproducible without it | 4/4 propose the `setup:` entry | 0/2 — both swap in `psycopg2-binary`, masking the environment gap |
| rustup, needs no root, installs into `$HOME` | boundary read correctly — runs name root as the literal trigger and reason independently about the rest | declines on generic curl-pipe-sh caution |

An earlier wording dropped the root qualifier from its last sentence and blind sessions read the rule as unconditional; that's what the explicit re-scoping in sentence three is for.

**Adversarial and outer-loop passes** verified the claims against primary sources: `proxy/setup-sandbox.sh:35,66-67` (sandbox user created with no sudo group), `claude/action.yaml:377,389` (`sudo -u "$SANDBOX"` launch), `codex/action.yaml:43-53` (codex runs on the runner with `danger-full-access`, so the harness qualifier is right), `setup:` rendering ahead of the agent step in every generated workflow, and `proxy/setup-sandbox.sh:118-175` for the `PATH` derivation.

The outer-loop pass confirmed placement two ways: `shared/system-prompt.md` carries social rules only, with no environment mechanics, and `shared/steps/compose-system-prompt.sh:18` injects the `running-in-ci` load into every session — so the "always loaded" argument for moving it up doesn't apply. It also noted the fact is anti-distributional rather than merely weakly-known: the model's prior is that GHA steps run as passwordless-sudo `runner`.

Scoped out deliberately: enumerating all four sandbox levers. `~` in `sandbox_path:` expands to the sandbox home and so can't reach a runner-home install, and `sandbox_setup:`/`sandbox_env:` cover non-root cases this bullet excludes — a four-item menu would be wrong for some of them, and `docs/tend.example.yaml:171-219` is their canonical home. Also scoped out: general taste guidance (fail fast, don't over-engineer), which is in-distribution and ages into noise.

</details>


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -86,6 +86,7 @@ If a linked PR merged (or the triggering PR itself merged) **after the triggerin
...
 - **Hanging commands**: Never use `gh run watch` or `gh pr checks --watch` — both hang indefinitely. Poll with `gh pr checks` in a loop instead.
+- **Privileges**: Under the default `claude` harness you run as a non-sudo sandbox user, so `sudo` fails and no installer that escalates can work from inside the session. A tool that needs root belongs in the repo's `setup:` steps in `.config/tend.yaml`, which run as `runner` — with sudo — before the agent starts. When a tool you need requires root, propose that `setup:` entry rather than working around its absence, even where a skill's own recipe tells you to install it in-session. The sandbox `PATH` is derived from the runner's, so a tool that installs outside it needs a `sandbox_path:` entry in the same config.
```
tend-agent: The last sentence states the wrong criterion, and it misfires on the most likely case. Membership in the runner's `PATH` isn't what carries an entry across — `docs/tend.example.yaml` (under "Sandbox setup (Claude harness)") puts it as reachability by the sandbox UID: "a tool in a system location (`/opt/hostedtoolcache/...`, where the `setup-*` actions install) does, one in the runner's own home does not." The derivation in `proxy/setup-sandbox.sh` rewrites a leading runner-home prefix to `$AGENT_HOME` and keeps the entry only if the sandbox copy exists, so `~/.cargo/bin` and `~/.local/bin` are *on* the runner's PATH yet arrive pointing at the sandbox's own (empty) `/etc/skel` copy.

So an agent that reads this and has a `setup:` step do `cargo install cargo-nextest` concludes no `sandbox_path:` is needed — the dir is already on the runner's PATH — and the binary is still invisible in-session. Adding `sandbox_path: ["~/.cargo/bin"]` doesn't rescue it either: the `~` expands to the sandbox home, where nothing was installed. `sandbox_setup:` is the lever that puts the tool inside the sandbox.

```suggestion
- **Privileges**: Under the default `claude` harness you run as a non-sudo sandbox user, so `sudo` fails and no installer that escalates can work from inside the session. A tool that needs root belongs in the repo's `setup:` steps in `.config/tend.yaml`, which run as `runner` — with sudo — before the agent starts. When a tool you need requires root, propose that `setup:` entry rather than working around its absence, even where a skill's own recipe tells you to install it in-session. Only a location the sandbox UID can reach carries in: a system dir (`/usr/local/bin`, `/opt/hostedtoolcache/…`) does, the runner's own home does not — so a `setup:` step that installs into `~` needs a `sandbox_setup:` entry to reinstall it inside the sandbox, plus `sandbox_path:` if it lands off PATH.
```

tend-agent (commented):
One accuracy point on the new bullet — the rest reads right.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
max-sixty: Taking the criterion — you're right that `PATH` membership isn't what carries an entry across, and `~/.cargo/bin` is the case that proves it: on the runner's `PATH`, rewritten to the sandbox's own `/etc/skel` copy, empty. The last sentence now reads "The sandbox reaches system locations but never the runner's home, so if the tool lands in a system dir that isn't already on `PATH`, add a `sandbox_path:` entry pointing at it."

Leaving `sandbox_setup:` out, though. `shared/steps/sandbox-setup.sh:2-3,30` runs those commands as the non-sudo sandbox user, so it can't run a root installer — and this bullet's whole scope is tools that need root. Your `cargo install cargo-nextest` example is a rootless install, which the bullet deliberately doesn't cover: earlier wordings that reached past root got read as unconditional by blind test sessions, so the scope is load-bearing.

tend-agent: Confirmed on both counts. [`shared/steps/sandbox-setup.sh:30`](https://github.com/max-sixty/tend/blob/081ab9fae53f5380ad1e3194abf564f115fe64e6/shared/steps/sandbox-setup.sh#L30) runs the commands under `sudo -u "$SANDBOX"`, so a root installer can't work there either — `sandbox_setup:` is the wrong lever for this bullet's scope, and my suggestion conflated the rootless case with it.

The replacement sentence matches the derivation: [`proxy/setup-sandbox.sh:165-176`](https://github.com/max-sixty/tend/blob/081ab9fae53f5380ad1e3194abf564f115fe64e6/proxy/setup-sandbox.sh#L165-L176) keeps every non-runner-home PATH entry as-is (plus the guaranteed `/usr/local/bin`, `/usr/bin`, `/bin`), so a system dir that isn't already on PATH is exactly the residue `sandbox_path:` covers. No remaining concerns.

tend-agent (approved):

PR #1046, opened 2026-08-24
fix(scripts): reject an unusable SHA before the poll loop

### LINKED ISSUE #1045: poll-pr-checks.sh: an abbreviated SHA burns the whole 9-min poll budget, then head_note falsely reports the branch advanced to that same commit
`poll-pr-checks.sh` treats an abbreviated SHA as a transient API failure rather than a bad argument. The GraphQL `GitObjectID!` variable rejects anything but a 40-hex OID at coercion time, `rollup()` swallows that error with `2>/dev/null`, and the caller can't distinguish it from a blip — so the loop sleeps through all nine iterations before reporting `no rollup returned ... UNVERIFIED, not green`. On the way out, `head_note` string-compares the abbreviated argument against the full head OID and prints a **false** "branch advanced" line naming the very commit that was passed in.

Live on `main` at [`plugins/tend-ci-runner/scripts/poll-pr-checks.sh`](https://github.com/max-sixty/tend/blob/main/plugins/tend-ci-runner/scripts/poll-pr-checks.sh) — the recent pagination work in #999 didn't touch either path.

## What fires

The coercion failure is deterministic and independent of repo or commit — it never reaches the resolver:

```console
$ gh api graphql -f owner=numbagg -f name=numbagg -f oid=56ff8fc -f query='
    query($owner: String!, $name: String!, $oid: GitObjectID!) {
      repository(owner: $owner, name: $name) {
        object(oid: $oid) { ... on Commit { statusCheckRollup { contexts(first: 1) { totalCount } } } }
      }
    }'
{"errors":[{"extensions":{"value":"56ff8fc","problems":[{"path":[],"explanation":"Could not coerce value \"56ff8fc\" to GitObjectID"}]},...
```

`rollup()` maps that to empty output, which the loop reads as "transient, try again":

```bash
resp=$(gh api graphql ... 2>/dev/null) || return 0   # -> ""
...
[ -z "$cur" ] && continue                            # x9, each behind `sleep 60`
```

Nine `sleep 60`s later `R` is still empty, so the script exits 2 — a correct verdict (it fails closed), reached about nine minutes after the argument could have been rejected.

## The false note

`head_note` compares the caller's `$SHA` to the PR's current head as strings:

```bash
if [ -n "$now_oid" ] && [ "$now_oid" != "$SHA" ]; then
  echo "note: branch advanced to $now_oid — the result above is still $SHA's, the commit this run is accountable for"
fi
```

An abbreviated `$SHA` never equals the full `headRefOid`, so the branch-advanced branch always fires — even when the head is literally the same commit. This is the second symptom and the one worth more than the wasted minutes: it is machinery generating a wrong factual claim about branch state, of exactly the shape `running-in-ci` tells sessions to act on ("a sibling pushed to this branch while you waited"). A session that relayed it would post a fabricated sibling push into a PR thread.

## Evidence

Two occurrences on `numbagg/numbagg`, in different workflows nine review-runs windows apart, so it isn't one session's habit:

| When | Run | Call | Result |
|---|---|---|---|
| 2026-08-15 | [31870798743](https://github.com/numbagg/numbagg/actions/runs/31870798743) (`tend-nightly`) | `poll-pr-checks.sh 736 fc9098c` | `no rollup returned for fc9098c`; re-run with the full OID → green |
| 2026-08-24 | [32703852345](https://github.com/numbagg/numbagg/actions/runs/32703852345) (`tend-mention`) | `poll-pr-checks.sh 758 56ff8fc` | `no rollup returned for 56ff8fc` **+ the false branch-advanced note**; re-run with the full OID → green |

The 2026-08-24 call ran at `08:10:04.730Z` and returned at `08:19:07.364Z` — **9m02s**, in a 25-minute session. Its full output was:

```
no rollup returned for 56ff8fc — UNVERIFIED, not green
note: branch advanced to 56ff8fca91971ad0ba625546e5a4919fe974efe8 — the result above is still 56ff8fc's, the commit this run is accountable for
```

`56ff8fc` is the abbreviation of `56ff8fca9197…`. The branch had not advanced.

Both sessions recovered on the next call, so nothing wrong was posted either time. Recorded here as waste-class with a latent posting risk, not as a wrong outward action.

## Why guidance won't close it

`running-in-ci` already prescribes `PINNED_SHA=$(git rev-parse HEAD)`, which yields 40 chars — and both sessions had read it. What they actually did was paste a SHA they already had in hand from a commit message or a `gh` projection. An argument-shape check is the only thing that catches that, and it belongs in the script rather than in prose a session must remember.

## Proposed fix

Reject the malformed argument at entry, before any sleeping, and give it its own message so it reads as a caller bug rather than an unreachable API. This also removes the `head_note` false positive, since the only way `$SHA` can be non-full-length is the case being rejected:

```bash
PR="$1"
SHA="$2"

# GraphQL's GitObjectID! rejects an abbreviated OID at coercion time, which
# rollup() cannot distinguish from a transient failure — without this the
# loop sleeps through its whole budget before reporting a bad argument, and
# head_note's string compare then claims the branch advanced to the very
# commit that was passed in.
case "$SHA" in
  *[!0-9a-fA-F]* | "") SHA_OK=0 ;;
  *) [ "${#SHA}" -eq 40 ] && SHA_OK=1 || SHA_OK=0 ;;
esac
if [ "$SHA_OK" -ne 1 ]; then
  echo "poll-pr-checks.sh: <sha> must be a full 40-char commit OID, got '$SHA' — UNVERIFIED, not green" >&2
  exit 2
fi
```

Exit 2 keeps the existing contract ("no usable rollup for `<sha>` — UNVERIFIED, not green"), so no caller needs to change; it just arrives in milliseconds with a message that names the cause. `head_note` is deliberately not called here — there is no verdict above for it to qualify.

- Coercion error reproduced directly against `repos/numbagg/numbagg` (transcript above); it is a variable-validation failure, so it fires before any repo or commit lookup and is not rate- or permission-dependent.
- Nine iterations × `sleep 60` matches the measured 9m02s to within API overhead.
- Installed script on the consumer (tend `0.1.18`) and upstream `main` differ only in the #999 pagination rewrite; `head_note` and the missing entry guard are byte-identical in both, so this is not a stale-pin artifact.
- Prior occurrence and its gate assessment are recorded in `numbagg/numbagg` [#710](https://github.com/numbagg/numbagg/issues/710) (Run 31873379432, Finding 1), which pre-registered "if this recurs, the cheap fix is upstream — reject a non-40-char SHA with its own message rather than folding it into the no-rollup path".

</details>

tend-agent:
## Problem

`poll-pr-checks.sh` took an abbreviated SHA as far as the API. GraphQL's `GitObjectID!` rejects a non-40-hex OID at variable coercion, `rollup()` maps that errors envelope to empty output, and the loop reads empty as "transient, retry" — so a caller bug burned all nine `sleep 60` iterations before exiting 2. Worse, `head_note` string-compares the caller's `$SHA` against the PR's full `headRefOid`, so an abbreviated argument always tripped the branch-advanced line: the script printed a false claim that a sibling pushed, naming the very commit it was handed. That is the shape `running-in-ci` tells sessions to act on, so a session relaying it would post a fabricated push into a PR thread. Reported in #1045 with two occurrences on a consumer repo.

## Solution

Validate `<sha>` at entry against `^[0-9a-f]{40}$` and exit 2 with a message that names the cause instead of an unreachable API. The exit code and the `UNVERIFIED, not green` wording are unchanged, so no caller needs to change; the verdict just arrives in milliseconds. Both positional args default (`${1:-}` / `${2:-}`) so an omitted `<sha>` lands on that same guard rather than dying at `set -u` with exit 1 — the code this file documents as red, which would send the caller off to diagnose a CI failure that never happened.

Lowercase-only in the class, not `[0-9a-fA-F]`: GraphQL coerces an uppercase OID happily, so it would poll successfully and *then* mismatch `headRefOid` — which comes back lowercase — trailing every verdict with a spurious branch-advanced note pointing at its own commit. Nothing in the repo emits uppercase (`git rev-parse` and `headRefOid` are both lowercase), so rejecting it costs no real caller.

The guard narrows `head_note`'s false positives rather than eliminating them: it still fires on the other exit-2 causes, since an unresolvable OID and an ephemeral `refs/pull/N/merge` commit are both full-length and neither is the branch head. `head_note` is deliberately not called on the reject path — there is no verdict above it to qualify.

## Testing

Three guard tests in `generator/tests/test_poll_scripts.py` — abbreviated, uppercase, and omitted `<sha>` — each asserting exit 2, the cause in the message, no `branch advanced` line, and no `gh` call at all. Against the pre-fix script the abbreviated case failed with `green: every gating check on aaaa111 settled green` plus the false branch-advanced note; the fake `gh` doesn't enforce coercion, so the test reproduces the wrong-claim half directly and the guard covers both halves. Full `uv run pytest` is green (587 passed).

---

Closes #1045 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/scripts/poll-pr-checks.sh
+++ b/plugins/tend-ci-runner/scripts/poll-pr-checks.sh
@@ -43,21 +43,22 @@
 
 set -euo pipefail
 
-PR="$1"
-SHA="$2"
+# Both default rather than relying on `set -u`: an omitted <sha> is the likelier
+# arity mistake, and unbound-variable death exits 1 — the code documented above
+# as red, sending the caller off to diagnose a CI failure that never happened.
+# Defaulting routes a short call into the guard below instead.
+PR="${1:-}"
+SHA="${2:-}"
 OWNER="${GITHUB_REPOSITORY%/*}"
 
 # GraphQL's `GitObjectID!` rejects an abbreviated OID at coercion time, which
 # rollup() cannot tell from a transient failure: without this the loop sleeps
 # through its whole budget before reporting a caller bug, and head_note's
 # string compare then claims the branch advanced to the very commit that was
-# passed in. Rejecting here is also what keeps that note honest — a non-full
-# OID is the only way it can misfire.
-case "$SHA" in
-  "" | *[!0-9a-fA-F]*) SHA_OK=0 ;;
-  *) [ "${#SHA}" -eq 40 ] && SHA_OK=1 || SHA_OK=0 ;;
-esac
-if [ "$SHA_OK" -ne 1 ]; then
+# passed in. head_note can still misfire on the other exit-2 causes above — an
+# unresolvable OID and a merge-ref commit are both full-length and neither is
+# the branch head — so this narrows that note rather than fixing it.
+if [[ ! $SHA =~ ^[0-9a-fA-F]{40}$ ]]; then
   echo "poll-pr-checks.sh: <sha> must be a full 40-char commit OID, got '$SHA' — UNVERIFIED, not green" >&2
   exit 2
 fi
--- a/plugins/tend-ci-runner/scripts/poll-pr-checks.sh
+++ b/plugins/tend-ci-runner/scripts/poll-pr-checks.sh
@@ -33,7 +33,7 @@
 #   0  every gating check on <sha> settled green
 #   1  red — failing checks and their run URLs on stdout
 #   2  no usable rollup for <sha> — UNVERIFIED, not green. A <sha> that isn't
-#      a full 40-hex OID, rejected at entry; an unresolvable OID; an
+#      a full 40-char lowercase OID, rejected at entry; an unresolvable OID; an
 #      ephemeral merge-ref commit, which carries none; a commit with
 #      zero checks and zero statuses (a push every workflow's paths filter
 #      excludes — the rollup is null then too, so "nothing to gate on" is
@@ -55,11 +55,17 @@ OWNER="${GITHUB_REPOSITORY%/*}"
 # rollup() cannot tell from a transient failure: without this the loop sleeps
 # through its whole budget before reporting a caller bug, and head_note's
 # string compare then claims the branch advanced to the very commit that was
-# passed in. head_note can still misfire on the other exit-2 causes above — an
-# unresolvable OID and a merge-ref commit are both full-length and neither is
-# the branch head — so this narrows that note rather than fixing it.
-if [[ ! $SHA =~ ^[0-9a-fA-F]{40}$ ]]; then
-  echo "poll-pr-checks.sh: <sha> must be a full 40-char commit OID, got '$SHA' — UNVERIFIED, not green" >&2
+# passed in. Lowercase-only for the same reason from the other direction:
+# GraphQL coerces an uppercase OID happily, but head_note compares it against
+# `headRefOid`, which comes back lowercase — so an uppercase argument
+# mismatches its own commit and trails every verdict with that spurious note.
+# Nothing here emits uppercase (`git rev-parse` and `headRefOid` are both
+# lowercase), so rejecting it costs no real caller. head_note can still
+# misfire on the other exit-2 causes above — an unresolvable OID and a
+# merge-ref commit are both full-length and neither is the branch head — so
+# this narrows that note rather than fixing it.
+if [[ ! $SHA =~ ^[0-9a-f]{40}$ ]]; then
+  echo "poll-pr-checks.sh: <sha> must be a full 40-char lowercase commit OID, got '$SHA' — UNVERIFIED, not green" >&2
   exit 2
 fi
 NAME="${GITHUB_REPOSITORY#*/}"
--- a/plugins/tend-ci-runner/scripts/poll-pr-checks.sh
+++ b/plugins/tend-ci-runner/scripts/poll-pr-checks.sh
@@ -32,8 +32,9 @@
 # Exit codes:
 #   0  every gating check on <sha> settled green
 #   1  red — failing checks and their run URLs on stdout
-#   2  no usable rollup for <sha> — UNVERIFIED, not green. An unresolvable
-#      OID; an ephemeral merge-ref commit, which carries none; a commit with
+#   2  no usable rollup for <sha> — UNVERIFIED, not green. A <sha> that isn't
+#      a full 40-hex OID, rejected at entry; an unresolvable OID; an
+#      ephemeral merge-ref commit, which carries none; a commit with
 #      zero checks and zero statuses (a push every workflow's paths filter
 #      excludes — the rollup is null then too, so "nothing to gate on" is
 #      indistinguishable from "nothing answered"); or a page walk that could
@@ -45,6 +46,21 @@ set -euo pipefail
 PR="$1"
 SHA="$2"
 OWNER="${GITHUB_REPOSITORY%/*}"
+
+# GraphQL's `GitObjectID!` rejects an abbreviated OID at coercion time, which
+# rollup() cannot tell from a transient failure: without this the loop sleeps
+# through its whole budget before reporting a caller bug, and head_note's
+# string compare then claims the branch advanced to the very commit that was
+# passed in. Rejecting here is also what keeps that note honest — a non-full
+# OID is the only way it can misfire.
+case "$SHA" in
+  "" | *[!0-9a-fA-F]*) SHA_OK=0 ;;
+  *) [ "${#SHA}" -eq 40 ] && SHA_OK=1 || SHA_OK=0 ;;
+esac
+if [ "$SHA_OK" -ne 1 ]; then
+  echo "poll-pr-checks.sh: <sha> must be a full 40-char commit OID, got '$SHA' — UNVERIFIED, not green" >&2
+  exit 2
+fi
 NAME="${GITHUB_REPOSITORY#*/}"
 
 # One query returns check runs *and* legacy status contexts, a page at a
--- a/generator/tests/test_poll_scripts.py
+++ b/generator/tests/test_poll_scripts.py
@@ -397,6 +397,29 @@ def test_waits_out_pending_then_reports_green(env: dict[str, str]) -> None:
     assert Path(env["GRAPHQL_CALLS"]).read_text().strip() == "3"
 
 
+def test_abbreviated_sha_is_rejected_at_entry(env: dict[str, str]) -> None:
+    """GraphQL's `GitObjectID!` rejects an abbreviated OID at coercion time, and
+    rollup() cannot tell that from a transient failure — so the loop would sleep
+    through its whole budget before reporting a caller bug, and head_note's
+    string compare would then claim the branch advanced to the very commit that
+    was passed in. Reject the argument before any API call."""
+    _serve(env, _resp(_check_run("tests")))
+
+    result = subprocess.run(
+        [BASH, str(POLL_PR_CHECKS), "7", HEAD_SHA[:7]],
+        env=env,
+        capture_output=True,
+        text=True,
+    )
+    out = result.stdout + result.stderr
+
+    assert result.returncode == 2, out
+    assert "UNVERIFIED, not green" in out
+    assert HEAD_SHA[:7] in out
+    assert "branch advanced" not in out, "head_note fired on the rejected argument"
+    assert not Path(env["GH_CALLS"]).exists(), "the bad argument reached the API"
+
+
 def test_moved_head_is_reported_not_absorbed(env: dict[str, str]) -> None:
     """Another actor pushing while we poll must not retarget the verdict: the
     result stays the pinned SHA's, with the move called out."""
--- a/generator/tests/test_poll_scripts.py
+++ b/generator/tests/test_poll_scripts.py
@@ -420,6 +420,27 @@ def test_abbreviated_sha_is_rejected_at_entry(env: dict[str, str]) -> None:
     assert not Path(env["GH_CALLS"]).exists(), "the bad argument reached the API"
 
 
+def test_uppercase_sha_is_rejected_at_entry(env: dict[str, str]) -> None:
+    """GraphQL coerces an uppercase OID happily, but head_note compares it
+    against `headRefOid`, which comes back lowercase — so an uppercase argument
+    would poll its own commit successfully and then trail every verdict with a
+    spurious "branch advanced" note pointing at that same commit."""
+    _serve(env, _resp(_check_run("tests")))
+
+    result = subprocess.run(
+        [BASH, str(POLL_PR_CHECKS), "7", HEAD_SHA.upper()],
+        env=env,
+        capture_output=True,
+        text=True,
+    )
+    out = result.stdout + result.stderr
+
+    assert result.returncode == 2, out
+    assert "UNVERIFIED, not green" in out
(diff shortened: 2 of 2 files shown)


plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -45,6 +46,21 @@ set -euo pipefail
 PR="$1"
 SHA="$2"
```
tend-agent: Omitting `<sha>` dies here under `set -u` with exit **1** — the code this file documents as "red — failing checks" — so it never reaches the new guard below. Defaulting both makes the guard's `""` case the live path for a short arity instead of a near-dead branch.

```suggestion
PR="${1:-}"
SHA="${2:-}"
```

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -45,6 +46,21 @@ set -euo pipefail
...
 OWNER="${GITHUB_REPOSITORY%/*}"
+
+# GraphQL's `GitObjectID!` rejects an abbreviated OID at coercion time, which
+# rollup() cannot tell from a transient failure: without this the loop sleeps
+# through its whole budget before reporting a caller bug, and head_note's
+# string compare then claims the branch advanced to the very commit that was
+# passed in. Rejecting here is also what keeps that note honest — a non-full
+# OID is the only way it can misfire.
```
tend-agent: `head_note` also runs on the no-rollup path, where it fires on any `now_oid != SHA` — so the unresolvable OID and the `refs/pull/N/merge` commit listed in the exit-2 table above both misfire it while being full 40-hex.

```suggestion
# passed in. head_note can still misfire on the other exit-2 causes above —
# an unresolvable OID and a merge-ref commit are both full-length and neither
# is the branch head — so this narrows that note rather than fixing it.
```

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -45,6 +46,21 @@ set -euo pipefail
...
+case "$SHA" in
+  "" | *[!0-9a-fA-F]*) SHA_OK=0 ;;
+  *) [ "${#SHA}" -eq 40 ] && SHA_OK=1 || SHA_OK=0 ;;
+esac
+if [ "$SHA_OK" -ne 1 ]; then
+  echo "poll-pr-checks.sh: <sha> must be a full 40-char commit OID, got '$SHA' — UNVERIFIED, not green" >&2
+  exit 2
+fi
```
tend-agent: The script is already bash-only (`local -a cursor_arg` above), and `shared/steps/` uses `[[ ]]`, so the `SHA_OK` sentinel buys nothing over a single test.

```suggestion
if [[ ! $SHA =~ ^[0-9a-fA-F]{40}$ ]]; then
  echo "poll-pr-checks.sh: <sha> must be a full 40-char commit OID, got '$SHA' — UNVERIFIED, not green" >&2
  exit 2
fi
```

tend-agent (commented):
The guard is worth having, but as written it closes the less likely half of the problem it names, and the comment justifying it overstates what it fixes.

The `""` case in the new `case` is only reachable when a caller passes an explicit empty string. The more likely arity mistake — omitting `<sha>` — never reaches the guard: `set -u` kills the script at `SHA="$2"` and exits **1**, which this file's own exit-code table defines as *"red — failing checks and their run URLs on stdout"*. So the miscall the PR set out to catch produces the worst available outcome rather than the UNVERIFIED it was aiming for:

```
$ GITHUB_REPOSITORY=owner/repo bash poll-pr-checks.sh 7
poll-pr-checks.sh: line 47: $2: unbound variable
exit=1
```

`running-in-ci` tells the caller to read exit 1 as red and go diagnose it with `gh run view --log-failed`, so a session that fumbles the argument reports a CI failure that never happened — and if it had approved, dismisses that approval over it. Two characters (`${1:-}` / `${2:-}`) route both arity cases into the guard that already exists.

Second, `a non-full OID is the only way it can misfire` isn't right, and it's contradicted by the exit-code table eighteen lines above it. `head_note` runs on the no-rollup path too, and fires on any `now_oid != SHA` — true for an unresolvable OID and for an ephemeral `refs/pull/N/merge` commit, both of which are full 40-hex and neither of which is the branch head. This guard narrows the note's misfire set; it doesn't empty it. Worth keeping the comment honest, since the next reader deciding whether `head_note` needs guarding on the exit-2 path will take that sentence at face value.

Last one is taste: the script is already bash-only (`local -a cursor_arg`), and `shared/steps/` uses `[[ ]]` — so the `case` plus `SHA_OK` sentinel is eight lines doing what one test does.

CI is green on this head. Not approving because I authored it.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
tend-agent: Applied in c80cfb4. Both default now, so an omitted `<sha>` lands on the guard's exit 2 instead of dying at `set -u` with the code the table above defines as red. `test_omitted_sha_is_rejected_not_reported_red` pins it — it asserts exit 2 and no `unbound variable` in the output, so a later revert to `"$2"` fails rather than silently downgrading the miscall back to "red".

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
tend-agent: Right — the sentence overclaimed, and the exit-2 table directly above it lists the two counterexamples. Reworded in c80cfb4 to say the guard narrows `head_note`'s misfire set rather than emptying it, naming the unresolvable OID and the merge-ref commit as full-length cases that still trip it.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
tend-agent: Taken — collapsed to the single `[[ ]]` regex test in c80cfb4. shellcheck passes, and the two rejection tests (abbreviated and omitted) cover the same ground the sentinel did.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -42,9 +43,25 @@
...
 OWNER="${GITHUB_REPOSITORY%/*}"
+
+# GraphQL's `GitObjectID!` rejects an abbreviated OID at coercion time, which
+# rollup() cannot tell from a transient failure: without this the loop sleeps
+# through its whole budget before reporting a caller bug, and head_note's
+# string compare then claims the branch advanced to the very commit that was
+# passed in. head_note can still misfire on the other exit-2 causes above — an
+# unresolvable OID and a merge-ref commit are both full-length and neither is
+# the branch head — so this narrows that note rather than fixing it.
+if [[ ! $SHA =~ ^[0-9a-fA-F]{40}$ ]]; then
```
tend-agent: `head_note`'s compare is against `headRefOid`, which is lowercase — so an uppercase 40-hex argument passes this guard and then misfires the note on every verdict, which is the failure the comment above cites as the reason for the guard. Nothing in the repo produces uppercase, so narrowing rejects no real caller.

```suggestion
if [[ ! $SHA =~ ^[0-9a-f]{40}$ ]]; then
```

tend-agent (commented):
The three earlier findings are all closed: both args default, the comment no longer claims a non-full OID is the only misfire, and the `case`/`SHA_OK` pair is gone. Resolving those threads.

One thing the rewrite carried over. The comment immediately above the guard justifies it by `head_note`'s string compare claiming "the branch advanced to the very commit that was passed in" — but `[0-9a-fA-F]` admits a 40-char form that still does exactly that. `head_note` compares `"$now_oid" != "$SHA"` against `gh pr view --json headRefOid`, which is always lowercase, so an uppercase argument mismatches its own commit unconditionally. GraphQL almost certainly coerces `GitObjectID` case-insensitively, which makes it the worse branch of the two: the poll succeeds and every verdict ships trailed by a spurious `note: branch advanced to <the same commit>`. Nothing in the repo emits uppercase (`git rev-parse` and `headRefOid` are both lowercase), so narrowing costs no legitimate caller, and exit 2 with a precise message is the safe direction for one that somehow appears.

Alternative if you'd rather not reject a SHA GraphQL would accept: keep the class and normalize once with `SHA="${SHA,,}"` after the guard, which fixes the compare instead of forbidding the input. Either is one edit; the narrower class is the smaller one.

Not approving because I authored it. CI is green on this head (`test`, `lint`, `test-proxy`, `test-worker`, `test-codex-surface`), and `uv run pytest generator/tests/test_poll_scripts.py` passes locally at 23 tests.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
tend-agent: Taken — narrowed to `^[0-9a-f]{40}$` in a800637, plus the message ("full 40-char lowercase commit OID") and the exit-2 table entry above.

Went with the narrower class over `SHA="${SHA,,}"` for the reason you gave: it's the smaller edit, and no caller here can hit it — `git rev-parse` and `headRefOid` are both lowercase, and the `git grep` for the class turned up this one site. Exit 2 is also the direction that can't produce a wrong claim, where normalizing leaves the compare correct but the poll silent about an argument nothing should be passing.

The comment above the guard now carries the lowercase rationale rather than leaving the class looking incidental. New test `test_uppercase_sha_is_rejected_at_entry` asserts exit 2, no `branch advanced` line, and no `gh` call — the fake `gh` doesn't enforce GraphQL coercion, so against the old class it fails exactly the way described: green verdict, spurious note naming its own commit. 24 tests in `test_poll_scripts.py`, 587 in the full suite, shellcheck clean.

Also refreshed the PR title and body, which still claimed a non-full-length SHA was the only way `head_note` could misfire.

tend-agent (commented):
The uppercase finding is closed: the class is `^[0-9a-f]{40}$`, the message and the exit-2 table both say lowercase, and `test_uppercase_sha_is_rejected_at_entry` pins exit 2 with no `branch advanced` line and no `gh` call. Resolving that thread. No new findings on this push.

One thing worth recording, since it's the way this narrowing could have silently no-opped rather than anything wrong with it. A bracket range like `[a-f]` is collation-sensitive in principle — under some glibc locales `[a-z]` collates `aAbB…`, so the class would have admitted the very uppercase input the guard exists to reject, and the new test would have been the thing that caught it. It doesn't happen here: bash's `=~` goes through `regcomp`, which keeps ranges on code points rather than collation, and I checked the guard against every locale the runner carries. Uppercase is rejected in all of them, so the test isn't passing by accident of the runner's `LANG`.

<details><summary>Checks run</summary>

```
$ for L in C C.UTF-8 en_US.UTF-8; do LC_ALL=$L bash -c \
    '[[ "ABCDEF0123456789ABCDEF0123456789ABCDEF01" =~ ^[0-9a-f]{40}$ ]] && echo match || echo reject'; done
reject
reject
reject
```

Callers: `poll-pr-checks.sh` is invoked from one place, the `CI Monitoring` recipe in `running-in-ci`, which pins `git rev-parse HEAD` or `headRefOid` — both lowercase 40-hex, so the narrowing rejects no real caller. `rg '\[0-9a-f'` over `plugins/`, `shared/`, and `generator/src/` returns this guard alone, so there is no existing validator it should have reused. `uv run pytest generator/tests/test_poll_scripts.py` passes at 24; the rollup on `a800637` is green across `lint`, `test`, `test-proxy`, `test-worker`, and `test-codex-surface`.

</details>

Not approving because I authored it.

PR #1047, opened 2026-08-24
Name what the sandbox agent's PATH couldn't reach

max-sixty:
An adopter's `setup:` steps run as the runner user, and anything they install under the runner's home never reaches the sandboxed agent — the PATH derivation rewrites each runner-home entry to the sandbox's own copy and drops it when that copy doesn't exist. None of that was logged, so a consumer repo ran for ten weeks with none of the three tools its merge gate invokes: `pre-commit` exited 127, `cargo insta` answered "no such command", and the session fell back to `INSTA_UPDATE=always cargo test`, which writes snapshots rather than checking them. Sixteen of 197 bot PRs pushed a first commit with the lint job red, against 4 of 350 of the maintainer's own over the same window. Finding it took reading ~100 runs' session-log artifacts, because nothing in tend said a word.

Three things change, so the next adopter in that position learns it from tend rather than from archaeology.

**The drop is named.** `setup-sandbox.sh` logs the composed sandbox PATH — the only place the `~` expansion and the runner-home rewrite are visible. `sandbox-setup.sh` then diffs the runner's PATH against the agent's by resolvable command name and names what is missing. It runs whether or not `sandbox_setup:` is set, since an empty one is exactly the config a missing tool hides in, and it reports after the adopter's commands have had their chance to close the gap. Dir-level reporting would have missed this case: `pre-commit` sat in `~/.local/bin`, a dir that exists on both sides and is empty in the sandbox.

**The documented fix now works.** A comment told adopters that for a tool a `setup:` step installs — it named `cargo-nextest` — "`sandbox_path: ["~/.cargo/bin"]` covers them." It covers nothing: `~` expands to the *sandbox* home, and the cargo-install actions use their own roots (`taiki-e/install-action` installs to `$HOME/.install-action/bin`, `baptiste0928/cargo-install` to `$HOME/.cargo-install/<crate>/bin`). Corrected here, in `docs/tend.example.yaml`, and in the action's input descriptions. The recipe that replaces it copies self-contained binaries, reinstalls anything that resolves its libraries through `$HOME`, and asserts by *running* each tool — `command -v` passes for a copied console script whose interpreter still lives in the runner's home, and `command -v a b c` exits 0 if any one of the three resolves.

**`sandbox_path` can no longer breach the boundary it documents.** Adopter entries are prepended verbatim and never pass through the rewrite, so a runner-home entry was the one way that dir could reach the agent's PATH. Refused now, with the working fix named.

Two bugs surfaced while testing, both in code merged but not yet released:

- `git config --global` runs as the sandbox user before the workspace is made traversable, and every sudo'd command inherits the runner's cwd there. `/home/runner` is `0750`, so git's startup stat of its own cwd failed — `fatal: failed to stat`, exit 128, the sandbox setup dead before the agent launched. This would have failed **every adopter's Claude run** from the next release (`b9fddc90`, unreleased).
- The report itself could kill a run it was only meant to describe: `comm` exits 1 on input it judges unsorted, under `set -e`. Its mirror image was silent — a failed listing on the sandbox side would have named every command on the runner's PATH as missing. Both bounded, and the lister now skips a dir it can't read rather than reporting everything in it.

<details>
<summary>Validation, and what the reviews changed</summary>

`test-sandbox` is a new CI job that runs both scripts for real against the shape that hides a missing toolchain: a `setup:` step installing into the runner's home, no sandbox lever set. It asserts the sandbox PATH carries no runner-home entry, that a `~` entry expands to the front, that the report names the planted tool — and that it stays quiet about a command that *is* reachable, which is how both of the report's degraded modes would show up.

The same job ran end-to-end on a real ubuntu-24.04 runner before this PR opened (`tend-agent/tend-integration`, since cleaned up). It caught the `git config` bug on its first run. On the final code it is green: the report named exactly the planted tool and nothing else — zero false positives on a stock image — and `sandbox_setup:` closed the gap.

Three independent reviews ran over the first draft. They found that the documented recipe would have left a `pip install --user` tool broken while its check passed, that the `comm` failure modes above were reachable, and that `sandbox_path` bypassed the rewrite. A generation-time note about the risky config shape was written and then deleted: it fired on every Claude adopter with a `setup:` block and no lever, which is the default shape and usually fine, and nothing at `init` time can tell a tool-cache install from a home install. The run-time report settles it with evidence instead.

</details>


tend-agent (commented):
Diagnosis and fix both hold up — I traced the `git -C` change (git stats its cwd during `setup_git_directory_gently` whatever the subcommand, so a 0750 `/home/runner` inherited through `sudo -u` is fatal before `--global` is even read) and the `sandbox_path` refusal (verbatim prepend was the only route past the rewrite; `while … done <<<` isn't a subshell, so `exit 1` really exits). `test-sandbox` is green and the `~` expansion / no-runner-home assertions are the right two things to pin.

Three things, all on the new code:

- The refusal branch is the one new security-relevant path in `setup-sandbox.sh` and nothing exercises it — `test-sandbox` only covers the supported `~` input, and there's no Python test for either script by design. Inline below with what a step would look like; happy to push it if you'd rather not hand-write it.
- `[ -r "$dir" ]` doesn't change the listing. I checked both branches locally: an unreadable dir's glob doesn't expand, `[ -f "$dir/*" ]` is false, and the loop prints nothing — with or without the guard, silently, exit 0. The guard is fine to keep as the explicit form, but the comment's stated reason ("would list empty and report everything in it as missing") describes an outcome neither branch produces, and the under-report property it claims only holds on the runner side.
- The two lister assignments run under `set -euo pipefail` with no escape, so a non-zero exit from `sudo`/`sort` takes the step down — the `[ -z "$agent_commands" ]` guard below only catches exit-0-with-no-output. Narrow trigger, but it's the same shape as the `comm` bug you already fixed here, and the file's own "neither failure may cost more than the report itself" is the invariant it misses.

<details><summary>What I checked and didn't flag</summary>

- `command -v a b c` exits 0 if any one resolves — verified, the doc's claim is right.
- The two `tend.example.yaml` copies are byte-identical and the `.pre-commit-config.yaml` parity hook covers the pair.
- `shellcheck -S warning` is clean on both scripts (the SC2016 on `list_commands` is info-level).
- The refusal fires before the export block and before the workspace chown, so the `if: always()` ownership restore has nothing to undo — no leak from the new early exit.
- `AGENT_PATH` duplicates `PATH=` in `$AGENT_ENV_FILE`, which `sandbox-setup.sh` already reads; the second channel is two lines and cheaper to consume than parsing, so not worth collapsing.
- An init-time check for a runner-home `sandbox_path` would fire before the wasted run, but the run-time error is clear and self-hosted homes aren't `/home/runner` — the cost is compute, not a wrong outward action, so I don't think it earns the code.

</details>

.github/workflows/ci.yaml
```diff
@@ -79,6 +79,110 @@ jobs:
...
+          TEND_ANTHROPIC_OAUTH_TOKEN: dummy
+          ACTION_PATH: ${{ github.workspace }}
+          TEND_UV_DIR: ${{ runner.temp }}/tend-uv
+          UV_CACHE_DIR: ${{ runner.temp }}/tend-mitmproxy-uv
+          # The adopter lever, exercised on its supported input: a `~` entry
+          # expands to the sandbox home and wins the front of the PATH. Its
+          # unsupported input — a runner-home dir, which would put the
+          # credential-holding home on the agent's PATH — is refused by
+          # setup-sandbox.sh before the env file is written.
+          TEND_SANDBOX_PATH: "~/.cargo/bin"
```
tend-agent, line 117: The refusal added in this PR is untested — this job exercises `sandbox_path`'s supported input but never its refused one, and it's the branch that keeps the credential-holding home off the agent's PATH. It's cheap to cover: the `case` fires at line ~148 of `setup-sandbox.sh`, well before the chown and the proxy start, so a second invocation exits in under a second and touches nothing this job depends on. Something like:

```yaml
      - name: A runner-home sandbox_path entry is refused
        env:
          TEND_GH_TOKEN: dummy
          TEND_ANTHROPIC_OAUTH_TOKEN: dummy
          ACTION_PATH: ${{ github.workspace }}
          TEND_UV_DIR: ${{ runner.temp }}/tend-uv
          UV_CACHE_DIR: ${{ runner.temp }}/tend-mitmproxy-uv
        run: |
          TEND_SANDBOX_PATH="$HOME/.cargo-install/tend-probe/bin" \
            bash proxy/setup-sandbox.sh >"$RUNNER_TEMP/refused.log" 2>&1 && rc=0 || rc=$?
          [ "$rc" -eq 1 ] || { echo "::error::runner-home sandbox_path was not refused (exit $rc)"; exit 1; }
          grep -q 'under the runner.s home' "$RUNNER_TEMP/refused.log" \
            || { echo '::error::refused, but without the message that names the fix'; exit 1; }
```

Placing it last keeps it clear of the two steps that read `$AGENT_PATH` from `$GITHUB_ENV` (the early exit writes none). Say the word and I'll push it.

shared/steps/sandbox-setup.sh
```diff
@@ -1,31 +1,79 @@
...
+# until the agent hits `command not found` mid-session and works around it
+# (skipping the check, or reaching for a weaker substitute). Diffing the two
+# PATHs by resolvable command name names the gap here, after sandbox_setup has
+# had its chance to close it. Reported, not fatal: most of what a runner carries
+# is irrelevant to a given session, and only the adopter knows which tools their
+# gate needs — asserting that belongs in their own `sandbox_setup:`.
+# `-r` as well as `-d`: listing a dir needs read, executing from it needs only
+# traverse, so a 0711 PATH dir would list empty and report everything in it as
+# missing. Skipping it under-reports instead, which is the right way for a
+# diagnostic to be wrong.
```
tend-agent: The guard doesn't change what gets listed. I ran both branches against a `0311` dir: bash can't read it, so `"$dir"/*` stays literal, `[ -f ]` is false, and the loop prints nothing — no stderr, exit 0. `-r` skips the same dir to the same empty result.

And the under-report framing only holds on the runner side. On the sandbox side an unreadable-but-executable dir still yields no names, so its commands land in `missing` as false positives whether or not the guard is there.

Worth keeping the test as the explicit form, but the reason it gives is one a future reader would act on:

```suggestion
# `-r` as well as `-d` says explicitly what the glob does implicitly: a dir the
# lister can't read doesn't expand, so nothing is listed from it either way.
# On the runner side that under-reports, which is the right way for a
# diagnostic to be wrong; on the sandbox side its commands read as missing.
```

shared/steps/sandbox-setup.sh
```diff
@@ -1,31 +1,79 @@
...
+for dir in $1; do
+  { [ -d "$dir" ] && [ -r "$dir" ]; } || continue
+  for f in "$dir"/*; do
+    if [ -f "$f" ] && [ -x "$f" ]; then printf "%s\n" "${f##*/}"; fi
+  done
+done'
+# The sandbox side runs as the sandbox user, so the -x test answers the question
+# that matters: can THAT uid execute it, not does the file exist.
+runner_commands="$(bash -c "$list_commands" _ "$PATH" | sort -u)"
+agent_commands="$(sudo -u "$SANDBOX" bash -c "$list_commands" _ "$AGENT_PATH" | sort -u)"
```
tend-agent: Neither assignment has an escape, and `set -euo pipefail` is in force — a non-zero `sudo` or `sort` here fails the step, which is the thing the comment two lines down says must not happen ("neither failure may cost more than the report itself"). The `[ -z "$agent_commands" ]` branch only catches a *successful* empty listing, so it can't stand in for this.

The trigger is narrow — the lister itself always exits 0, so it takes `sudo` or `sort` failing — but it's the same shape as the `comm` exit-1 path you already closed here, and the fix is the same two characters:

```suggestion
runner_commands="$(bash -c "$list_commands" _ "$PATH" | sort -u || true)"
agent_commands="$(sudo -u "$SANDBOX" bash -c "$list_commands" _ "$AGENT_PATH" | sort -u || true)"
```

.github/workflows/ci.yaml
max-sixty: Added in 328712a6 — a second `setup-sandbox.sh` invocation with a runner-home entry, asserting non-zero exit and the error text. Placed last, since the refusal returns before the workspace chown and so disturbs nothing the earlier steps set up.

shared/steps/sandbox-setup.sh
max-sixty: Right, and I've deleted the guard rather than reworded it: the glob stays literal over an unreadable dir, so it listed nothing either way. The false positive it claimed to prevent is real and unaffected by it, which the comment now says instead of claiming a fix.

shared/steps/sandbox-setup.sh
max-sixty: Fixed in 328712a6. Both listings now carry `|| true`, and an empty agent list already falls through to the no-report path — so all three exits the diagnostic can take are inside the budget the comment above states.

tend-agent:
All three land, and `test-sandbox` is green on `328712a6` (run [32770999706](https://github.com/max-sixty/tend/actions/runs/32770999706)) — nothing outstanding from my side.

<details><summary>What I re-checked</summary>

- **Refusal test.** The grep pattern matches the message `setup-sandbox.sh` actually emits, and it's the last command under Actions' default `bash -e`, so a script that exits non-zero for some other reason still fails the step rather than passing on `rc != 0` alone. The re-invocation's pre-refusal work is idempotent — `useradd` is `id`-guarded, and the gitignore write plus `git config --global` rewrite the same values — so running it after the earlier steps disturbs nothing. Pinning `MITMPROXY_VERSION: "0"` gets past the unset check without ever reaching the install, which is 200 lines past the `exit 1`.
- **The `-r` guard.** Deleting it is better than the reword I suggested: the guard never changed the listing, and the comment now states the false positive that's actually there instead of claiming a fix for it.
- **`|| true` on both listings.** Covers the `sudo`/`sort` exit the `[ -z "$agent_commands" ]` branch couldn't, so the diagnostic's three exits are all inside the budget the comment above them states.

</details>

.github/workflows/ci.yaml
```diff
@@ -79,6 +79,135 @@ jobs:
...
+          TEND_ANTHROPIC_OAUTH_TOKEN: dummy
+          ACTION_PATH: ${{ github.workspace }}
+          TEND_UV_DIR: ${{ runner.temp }}/tend-uv
+          UV_CACHE_DIR: ${{ runner.temp }}/tend-mitmproxy-uv
+          MITMPROXY_VERSION: "0"
+        run: |
+          set -o pipefail
+          TEND_SANDBOX_PATH="$HOME/.cargo-install/tend-probe/bin" \
+            bash proxy/setup-sandbox.sh >"$RUNNER_TEMP/refused.log" 2>&1 && rc=0 || rc=$?
+          cat "$RUNNER_TEMP/refused.log"
```
tend-agent: The dump re-emits the `::error::` line the script wrote, so a passing `test-sandbox` job ships a `failure` annotation — visible on this commit's green run. Neutralize the prefix and the text still reads:

```suggestion
          sed 's/^::error::/refused: /' "$RUNNER_TEMP/refused.log"
```

The `grep -q` below still asserts against the file, so the check is unchanged.

tend-agent (approved):
All three fixes hold. I re-checked each: the refusal test's second invocation reaches the `case` before `AGENT_ENV_FILE` is assigned, so it writes no env file and the earlier steps' `$AGENT_PATH` survives; `MITMPROXY_VERSION` only has to be non-empty to clear the guard at the top, and its one real use is 200 lines past the exit. Deleting the `-r` guard rather than rewording it is the right call — and the comment now describes the false positive that is actually there. Both listings inside `|| true` closes the last of the three exits the diagnostic could take.

One thing on the new step, verified against this commit's own green `test-sandbox` run rather than inferred: it leaves a `failure`-level annotation on a passing job. `cat` re-emits the captured `::error::` line to the step's stdout, and Actions parses workflow commands from any step output regardless of where the text came from — so the run summary reads `Error: sandbox_path entry '/home/runner/.cargo-install/tend-probe/bin' is under the runner's home...` while every check is green. This repo treats an annotation as a real signal elsewhere (`rate-limit-preflight.sh` calls it "the honest signal" and builds its no-issue fallback around it), so a synthetic one on a healthy run is the kind of noise that trains a reader to skim past them. Suggestion inline keeps the log line visible and readable while stopping Actions from claiming it.

.github/workflows/ci.yaml
max-sixty: Good catch — fixed in a3e5b4ad with your suggestion. The assertion below is unchanged; only the prefix is neutralized, so the step still provokes the error and still checks for it, without claiming one to the run summary.

PR #1048, opened 2026-08-24
Make runner-home sandbox setup explicit

max-sixty:
## Summary

Replace the Claude sandbox's runner-home runtime mirror with one explicit boundary: shared system and hosted-toolcache paths cross; a runner-home path may select the same relative directory only when it was independently seeded into the sandbox home; runner-home files themselves never cross. Later home-scoped changes are installed deliberately with `sandbox_setup:`.

The previous mirror classified Python, uv, Rust, Node, Composer, Go, .NET, and SDK layouts; copied and filtered filesystem trees; retargeted symlinks; stripped metadata; and inferred `JAVA_HOME`. This change deletes that language-specific machinery. A single generic failure shim prevents a dropped home-selected command from silently falling through to a different same-named system binary, while an explicit sandbox path or `.local/bin` install shadows the shim.

The hosted sandbox fixture now tests the invariant directly: distinct runner and sandbox seed contents prove that only the sandbox-owned copy is selected; explicit and inherited checkout paths plus a non-base shared path cross; runner-home aliases do not; same-name fallback exits 127; and `sandbox_setup` closes the gap. Claude action steps receiving real credentials resolve `/usr/bin/bash` and their child commands outside adopter-controlled PATH.

The trade-off is intentional: standard setup actions that publish shared/toolcache paths and stock tools seeded into the sandbox home keep working; later changes made only under the runner home require a corresponding `sandbox_setup:` command. Tend no longer guesses which home files are tools rather than secrets.

## Validation

- `wt test` — 584 Python tests and 32 worker tests passed; typecheck passed
- hosted Ubuntu second-UID integration job in this PR


proxy/setup-sandbox.sh
```diff
@@ -226,9 +440,32 @@ AGENT_PATH="$(IFS=:; printf '%s' "${_agent_path[*]}")"
...
+    ;;
+esac
+if [ -n "${_agent_java_home}" ] && \
+   ! { [ -d "${_agent_java_home}" ] && sudo -u "$SANDBOX" test -x "${_agent_java_home}"; }; then
+  _agent_java_home=
+fi
 cat >"$AGENT_ENV_FILE" <<EOF
 HOME=${AGENT_HOME}
 PATH=${AGENT_PATH}
+${_agent_java_home:+JAVA_HOME=${_agent_java_home}}
```
tend-agent:
```suggestion
PATH=${AGENT_PATH}${_agent_java_home:+
JAVA_HOME=${_agent_java_home}}
```

The conditional has to swallow the newline rather than sit on its own line, otherwise an empty `_agent_java_home` leaves a blank line and `env "${AGENT_ENV[@]}"` exits 127 for every sandbox step. This form is compact but a bit clever — moving the assignment to an `if` after the `EOF` is the plainer equivalent.

tend-agent (commented):
One thing here will break runs; the rest are notes.

**The `JAVA_HOME` line writes a blank line when it's empty, and that kills the run.** `${_agent_java_home:+…}` on its own heredoc line expands to an empty *line*, not to nothing. Every consumer reads `$AGENT_ENV_FILE` with `mapfile -t` and splats it into `env "${AGENT_ENV[@]}"` — [`install-tend-plugins.sh`](https://github.com/max-sixty/tend/blob/1fce144dd0d9e2555a0b2687a17aad7e56dde247/shared/steps/install-tend-plugins.sh#L25-L26), [`sandbox-setup.sh`](https://github.com/max-sixty/tend/blob/1fce144dd0d9e2555a0b2687a17aad7e56dde247/shared/steps/sandbox-setup.sh#L33-L34), [`claude/action.yaml`](https://github.com/max-sixty/tend/blob/1fce144dd0d9e2555a0b2687a17aad7e56dde247/claude/action.yaml#L397-L399) — and `env` reads the empty element as the program to exec, so the run dies at "Install tend plugins (sandbox)" with `env: '': No such file or directory`, which points nowhere near the cause. The `TEND_SANDBOX_ENV` loop just below already guards this with `[ -n "$line" ] || continue`, and the block comment above the heredoc states the contract as "one NAME=VALUE per line". Inline suggestion below; appending `JAVA_HOME` after the `EOF` with an `if` reads more plainly if you prefer it. CI can't catch this — the stock image always has `java` on `PATH`, and this PR adds `setup-java` on top.

Also worth a look:

- **The `sandbox-env-reserved-parity` hook no longer sees `JAVA_HOME`.** It extracts heredoc keys with `^([A-Za-z_][A-Za-z0-9_]*)=`; the new line starts with `${`, so the key is invisible and `JAVA_HOME` is absent from the reserved `case`. The hook's own comment names exactly this drift as what it exists to fail the commit on. If adopter override via `sandbox_env:` is intended, that's a fine answer — but as a documented exemption rather than a regex miss. (The "append after `EOF`" fix keeps it outside that regex too.)
- **`test -x` on the mapped directory can pass for a runtime that can't run.** An unrecognized layout mirrors only the `PATH` directory, so a JDK at `$HOME/jdk` with `$HOME/jdk/bin` on `GITHUB_PATH` copies `bin/` and not `lib/`. `sudo -u tend-sandbox test -x .../jdk/bin/java` then succeeds, so `_unmirrored_commands` stays silent — and the new `JAVA_HOME` derivation points at `$AGENT_HOME/jdk`, whose `[ -d ] && test -x` guard passes because `mkdir -p` created that parent. The old behaviour reported `java` unreachable; this reports green and fails mid-session.
- **An empty `GITHUB_WORKSPACE` opens the boundary rather than closing it.** `"$GITHUB_WORKSPACE"/*` becomes `/*`, which in a `case` matches every absolute path, so the workspace arm swallows `/home/runner/.cargo/bin` — and `/home/runner` itself — onto the agent `PATH` unrewritten, with the `sandbox_path` refusal disabled too. `set -u` catches unset, not empty. The script hard-fails early on `TEND_GH_TOKEN` and `MITMPROXY_VERSION`; this variable now decides a boundary, and it's read long before the first `git -C "$GITHUB_WORKSPACE"` that used to be its de-facto check.
- **`_mirror_runner_tree` clears the destination before deciding it can mirror.** For a refused root that's the documented intent, but `_mirrored_roots` dedups on exact paths, so an ancestor root can re-enter and wipe a mirror an earlier `PATH` entry already built — and the entry still joins the agent `PATH`, because the `${AGENT_HOME}/*` arm skips the existence test.

**On the boundary itself** — your call, and I'd want it made deliberately rather than on momentum. The guarantee moves from "no runner-home content reaches the agent" to "runner-home content reaches the agent unless its mode is private", and the companion roots are subtrees rather than `PATH` directories: `~/.local/lib` and `~/.local/share/uv/{tools,python}` ride along with `~/.local/bin`, `~/.rustup` with `~/.cargo/bin`. Runner umask is 022, so a credential a `setup:` step leaves in one of those trees — the `direct_url.json` `pip install --user https://user:TOKEN@…` records, a `uv tool install --index-url https://user:token@…` receipt — is 0644 and crosses; CI encodes the rule by planting `~/.local/lib/tend/value` and asserting it *does*. Separately, what replaced the old rewrite is the per-language allowlist the old comment claimed not to need, and `~/.cargo/bin` and `~/.dotnet/tools` are on the stock ubuntu-24.04 `PATH`, so every adopter run copies all of `~/.rustup` and `~/.dotnet/tools` whether or not the repo is Rust or .NET — ~22 s on this PR's own green run. The seconds are cheap per CLAUDE.md; the surface and the allowlist are what gets maintained forever. `docs/security-model.md` isn't in the diff and still describes the older stance.

Blank line from the heredoc, and what the consumer pattern does with it:

```
$ _agent_java_home=; cat >/tmp/envfile <<EOF
A=1
${_agent_java_home:+JAVA_HOME=${_agent_java_home}}
B=2
EOF
$ mapfile -t E </tmp/envfile; printf '[%s]\n' "${E[@]}"
[A=1]
[]
[B=2]
$ env "${E[@]}" /bin/echo hello
env: ‘’: No such file or directory      # rc=127
```

`readlink -f /nonexistent` exits 0 and echoes the path back, so the `*/bin/java` case simply doesn't match and `_agent_java_home` stays empty — no `java` on `PATH` (a `container:` job, a self-hosted runner, a trimmed image) is enough. The second route is a JDK under the runner home in a layout the mirror refused: the `[ -d ] && test -x` guard then clears `_agent_java_home` and produces the same blank line.

Empty-`GITHUB_WORKSPACE` match:

```
$ W=""; case /home/runner/.cargo/bin in "$W"|"$W"/*) echo MATCH;; esac
MATCH
```

Mirror cost, from the `test-sandbox` job on `1fce144`:

```
20:59:34.42  mirrored 15 entries from /home/runner/.cargo/bin
20:59:49.00  mirrored 329 entries from /home/runner/.rustup        (~14.6 s)
20:59:56.18  mirrored 415 entries from /home/runner/.dotnet/tools   (~7.2 s)
```

Stock image `PATH` (`/etc/environment`, image 20260816.277.1) carries `/home/runner/.local/bin`, `/home/runner/.cargo/bin`, `/home/runner/.config/composer/vendor/bin` and `/home/runner/.dotnet/tools`; `/home/runner/.rustup` measures 624 MB on that image.

Checked and clean: the `find | tar` filter genuinely prunes non-`o+rx` directories rather than emitting empty entries, excludes special inodes, and strips setuid and file capabilities; hardlinks resolve against the extracted copies, not the runner's inodes; `$destination` can't degenerate to `$AGENT_HOME`; the `[ … ] && continue` idioms sit in the `set -e` exempt position; and an abort mid-mirror precedes both the `$GITHUB_ENV` export and the workspace chown, so there's no partial state a running agent could reach.

</details>

tend-agent (approved):

proxy/test-setup-sandbox.sh
```diff
@@ -0,0 +1,249 @@
...
+  alias="/tmp/tend-runner-home-alias-$GITHUB_RUN_ID"
+  case "$alias" in
+    "$HOME" | "$HOME"/*)
+      echo "::error::the canonical-refusal fixture is lexically under HOME"
+      exit 1
+      ;;
+  esac
+  TEND_SANDBOX_PATH="$alias" \
+    bash proxy/setup-sandbox.sh >"$RUNNER_TEMP/refused.log" 2>&1 && rc=0 || rc=$?
+  sed 's/^::error::/refused: /' "$RUNNER_TEMP/refused.log"
```
tend-agent, line 151:
```suggestion
  # Actions parses workflow commands out of any step output, so echoing the
  # captured line verbatim would annotate this passing job with the error it
  # exists to provoke.
  sed 's/^::error::/refused: /' "$RUNNER_TEMP/refused.log"
```

proxy/test-setup-sandbox.sh
```diff
@@ -0,0 +1,249 @@
...
+  setup_commands=$'printf \'#!/bin/sh\\necho private\\n\' >~/.local/bin/tend-private\nchmod +x ~/.local/bin/tend-private\ntend-private'
+  TEND_SANDBOX_SETUP="$setup_commands" \
+    bash shared/steps/sandbox-setup.sh | tee "$RUNNER_TEMP/report-after.log"
+  if grep "not the agent's:" "$RUNNER_TEMP/report-after.log" | grep -q tend-private; then
+    echo "::error::still unreachable after sandbox_setup installed it"
+    exit 1
+  fi
+}
+
+verify_refusals() {
```
tend-agent, line 143:
```suggestion
# The refusal path, the security-relevant half of the lever: a runner-home
# entry would put the credential-holding home on the agent's PATH, and adopter
# entries are the only input that never passes through the rewrite.
# setup-sandbox.sh exits before the workspace chown, so re-running it here
# disturbs nothing the steps above set up.
verify_refusals() {
```

proxy/test-setup-sandbox.sh
```diff
@@ -0,0 +1,249 @@
...
+      /home/tend-sandbox/.cargo-install/tend-probe/bin/tend-link-bypass; then
+    echo "::error::a copied symlink still reaches its runner-owned inode"
+    exit 1
+  fi
+  test -n "$(getcap "$source_bin/tend-capability")"
+  test -z "$(getcap /home/tend-sandbox/.cargo-install/tend-probe/bin/tend-capability)"
+  grep -q "runner-selected commands not mirrored into sandbox: .*tend-fallback" \
+    "$RUNNER_TEMP/setup.log"
+
+  bash shared/steps/sandbox-setup.sh | tee "$RUNNER_TEMP/report.log"
```
tend-agent, line 118:
```suggestion
  # Refused files remain visible in the diagnostic; successfully mirrored
  # commands do not.
  bash shared/steps/sandbox-setup.sh | tee "$RUNNER_TEMP/report.log"
```

proxy/test-setup-sandbox.sh
```diff
@@ -0,0 +1,249 @@
...
+  esac
+  case "$report" in
+    *tend-probe* | *tend-local* | *tend-pinned* | *" git "*)
+      echo "::error::the report named a reachable command: $report"
+      exit 1
+      ;;
+  esac
+}
+
+verify_sandbox_setup() {
```
tend-agent:
```suggestion
# The lever closes an intentional mirror refusal. Reinstall rather than copy:
# the runner source is private by definition.
verify_sandbox_setup() {
```

tend-agent (commented):
The extraction is assertion-for-assertion faithful — I diffed the old inline blocks against the script line-set and nothing was dropped, `set -euo pipefail` is a strict superset of what the steps had (`bash -e {0}` plus per-step `set -o pipefail`), and the `$'…'` `TEND_SANDBOX_SETUP` value is byte-identical to the old YAML block scalar apart from a trailing newline. `test-sandbox` is green on this head, which is the check that matters.

One thing worth restoring: four rationale comments didn't make the move, and two of them are load-bearing against a plausible future edit.

- The `sed 's/^::error::/refused: /'` line lost the note that Actions parses workflow commands out of any step output — without it that line reads as cosmetic, and "simplifying" it to `cat` annotates a *passing* job with the error it exists to provoke.
- `verify_refusals` lost the argument that re-running `setup-sandbox.sh` mid-job is safe because it exits before the workspace chown. That's the ordering invariant holding the new step in place, and it's now unwritten.

The other two (why the `verify` report checks both directions; why `verify_sandbox_setup` reinstalls rather than copies) are cheaper to lose but were the same one-line-each cost to keep. Suggestions inline for all four; the wording is lifted from the removed lines, so take or reshape as you like.

tend-agent (dismissed):

tend-agent (commented):
The boundary reads clean now — one rule, no per-language machinery, and `test-setup-sandbox.sh` asserts it directly rather than through the workflow. One thing I'd want settled before this lands, because it's a regression against `main`, not just against the intermediate commits here.

**The rewrite you deleted wasn't only reaching the runner's home — it was reaching the sandbox's own.** The dropped `case` arm mapped `${runner_home}/X` → `${AGENT_HOME}/X` and kept the entry when that destination existed. What it found there was the copy `useradd -m` makes from `/etc/skel`, and on `ubuntu-24.04` that copy is real and populated: `/etc/skel/.cargo/bin` and `/etc/skel/.dotnet/tools` both exist, so every `tend-sandbox` user starts with a working rust toolchain and the dotnet tool shims in its own home, before any adopter step has run.

So after this change a stock adopter's agent has a complete rust toolchain sitting in `/home/tend-sandbox/.cargo/bin` that nothing puts on `PATH`, and fifteen shims ahead of `/usr/bin` telling it those commands "came from the runner home and are unavailable". `sandbox_setup:` isn't a recovery for the cargo half either — `cargo install` writes to `~/.cargo/bin`, which isn't on the agent `PATH`, and `cargo` is itself a shim. Two knock-ons: `command -v cargo` now succeeds where it used to fail, so a `command -v X && X …` guard flips from a clean skip to a 127 under `sandbox_setup:`'s `bash -e`; and `sandbox-setup.sh`'s report names all sixteen on every stock run.

The rationale in the new comment — "that home can contain credentials unrelated to the selected tool" — is right about `/home/runner` and doesn't reach `/home/tend-sandbox`. Two ways out, your call:

- **Keep the `${AGENT_HOME}/${relative}` rewrite, gated on the destination already existing.** No copying, no filtering, no per-language code; it only ever names a directory the sandbox already owns. It also collapses the shim set to genuinely runtime-installed tools, which is the rare case the shim is actually for.
- **Or accept the drop and make it recoverable.** `sandbox_path: ~/.cargo/bin` still works — this PR's own test is the proof, since that entry is exactly why `cargo` is absent from the report line in the green run. But `docs/tend.example.yaml` replaced that example with `/opt/my-tool/bin`, and neither the `runner-home PATH entries unavailable` log line nor the shim's own message mentions it. The one-line fix is currently invisible to the adopter who needs it.

Smaller, independent of which route you take:

- **The `/usr/bin/bash` hardening covers one of the four steps holding the real PAT.** `Security preflight`, `Rate limit preflight` and `Restore sensitive config from base branch (PRs)` each take `github_token: ${{ inputs.github_token }}` and still carry `shell: bash` plus `run: bash "${{ github.action_path }}/…"` — two PATH lookups apiece, and `setup:` renders before the composite action in every generated workflow, so their `PATH` is adopter-mutated too. The comment's "the only one that receives both real credentials" is literally true, but it reads as "the only credential-bearing step". Worth either extending to those three or saying what makes PAT-only steps exempt.
- **The new `GITHUB_WORKSPACE` carve-outs have no accept-side test.** `verify_refusals` covers the refusal and the empty-workspace guard, but nothing exercises a `sandbox_path` entry *inside* the checkout being accepted, or a `$GITHUB_PATH` entry under the workspace crossing on the `_shared_path=1` branch — which is the one branch that skips the `sudo -u "$SANDBOX" test -x` probe. Both are new holes in a check whose whole job is to have none.
- `ln -s ../unavailable` isn't idempotent, so a second `setup-sandbox.sh` in one job dies there. Nothing reaches it today — both re-invocations in `verify_refusals` exit earlier — but `ln -sfn` costs two characters and the rest of the script is re-runnable.

Happy to push the doc example and the two test assertions if you take the second route.

Live sandbox session on the current release (`max-sixty/tend/claude@0.1.18`), i.e. the behavior this PR replaces:

```
$ ls -d /etc/skel/.cargo/bin /etc/skel/.dotnet/tools /etc/skel/.rustup
/etc/skel/.cargo/bin
/etc/skel/.dotnet/tools
/etc/skel/.rustup
$ command -v cargo rustc nbgv
/home/tend-sandbox/.cargo/bin/cargo
/home/tend-sandbox/.cargo/bin/rustc
/home/tend-sandbox/.dotnet/tools/nbgv
$ ls ~/.cargo/bin
cargo  cargo-clippy  cargo-fmt  cargo-miri  clippy-driver  rls  rust-analyzer
rust-gdb  rust-gdbgui  rust-lldb  rustc  rustdoc  rustfmt  rustup
```

`test-sandbox` on `6c56141` blocks that same set (`sudo` and `tend-probe` are the fixture's):

```
[setup-sandbox] runner-home PATH entries unavailable in sandbox: /home/runner/.cargo-install/tend-probe/bin /home/runner/.cargo/bin /home/runner/.dotnet/tools
[setup-sandbox] runner-home commands blocked from shared fallbacks: sudo tend-probe cargo cargo-clippy cargo-fmt cargo-miri clippy-driver rls rust-analyzer rust-gdb rust-gdbgui rust-lldb rustc rustdoc rustfmt rustup nbgv
[setup-sandbox] sandbox PATH: /home/tend-sandbox/.cargo/bin:/home/tend-sandbox/.local/bin:/home/tend-sandbox/.tend-blocked/bin:/opt/tend-sandbox-test-…/bin:/opt/pipx_bin:/usr/local/.ghcup/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/usr/games:/usr/local/games:/bin
[sandbox-setup] on the runner's PATH, unavailable to the agent: nbgv sudo tend-probe
```

The leading `/home/tend-sandbox/.cargo/bin` there is the test's `TEND_SANDBOX_PATH='~/.cargo/bin'`, and it's why the fourteen cargo names are shimmed but absent from the last line — the report's inner loop finds an earlier `PATH` dir that resolves them. Without that `sandbox_path:` entry, which no stock adopter has, they'd all be on it.

Checked and clean: `_add_agent_path`/`_add_blocked_command` are safe under `set -u` on bash 5.2 and return 0 on the no-match path (the append after the loop decides the status); `PATH=… type -P` is a temporary assignment on a regular builtin *and* inside a command substitution, so it can't leak; the shim's `${0##*/}` resolves through the symlink name; `_shared_path` is assigned on every branch that reaches its read (the home branch `continue`s first); the blocked-name splice can't land at index 0 since `.local/bin` is always in the prefix; `readlink -f` canonicalization closes the symlink-alias route into the runner home, which `verify_refusals` covers. `shellcheck -S warning` and `uv run pytest` (584) are clean locally, and `pre-commit`'s new `proxy/.*\.sh` glob picks up the extracted test script.

</details>

proxy/test-setup-sandbox.sh
max-sixty: Addressed in 1107c8e6 with wording adapted to the no-copy boundary: the fixture now states that dropped commands remain in the report while shared, workspace, and independently seeded sandbox-home commands do not.

proxy/test-setup-sandbox.sh
max-sixty: Addressed in 1107c8e6: the refusal helper now records both the security purpose of the check and why these re-runs stop before the ownership handoff.

proxy/test-setup-sandbox.sh
max-sixty: Addressed in 1107c8e6: the fixture again explains why the deliberate error line is rewritten before reaching Actions output.

proxy/setup-sandbox.sh
```diff
@@ -156,75 +165,123 @@ if [ -n "${TEND_SANDBOX_PATH:-}" ]; then
...
+  case "${_resolved_path}" in
+    "${GITHUB_WORKSPACE}" | "${GITHUB_WORKSPACE}"/*)
+      _d="${_resolved_path}"
+      _shared_path=1
+      ;;
+    "${runner_home}")
+      _dropped_home_path+=("${_resolved_path}")
+      _drop_home=1
+      ;;
+    "${runner_home}"/*)
```
tend-agent, line 214:
```suggestion
    # This rewrite is the whole filter for runner-home entries. Don't relax it
    # trusting the `test -x` at the shared-path gate below as a backstop: it is
    # not one. That test only rejects /home/runner paths because step 3 hasn't
    # run yet; step 3 grants o+x on every ancestor of $GITHUB_WORKSPACE (which
    # lives under the runner home), so from then on the sandbox uid can traverse
    # /home/runner and an un-rewritten entry would sail through.
    "${runner_home}"/*)
```

proxy/setup-sandbox.sh
```diff
@@ -156,75 +165,123 @@ if [ -n "${TEND_SANDBOX_PATH:-}" ]; then
...
 done
+# Put generic failure shims after adopter paths and .local/bin, but before the
+# shared runner paths. `sandbox_setup:` can therefore replace a home-selected
+# command explicitly; otherwise invoking it fails instead of changing version.
+TEND_BLOCKED_PATH=
+if [ "${#_blocked_home_command[@]}" -gt 0 ]; then
+  TEND_BLOCKED_PATH="${AGENT_HOME}/.tend-blocked/bin"
+  sudo -u "$SANDBOX" mkdir -p "$TEND_BLOCKED_PATH"
+  printf '%s\n' '#!/bin/sh' \
+    'printf "tend: %s came from the runner home and is unavailable; install it into ~/.local/bin with sandbox_setup or configure sandbox_path\n" "${0##*/}" >&2' \
```
tend-agent: The shim fires only for a command that came from a runner-home directory, and `sandbox_path` refuses exactly that directory a few dozen lines up. Qualifying it keeps the pointer without sending the reader into the refusal:

```suggestion
    'printf "tend: %s came from the runner home and is unavailable; install it into ~/.local/bin with sandbox_setup, or point sandbox_path at a copy outside the runner home\n" "${0##*/}" >&2' \
```

proxy/test-setup-sandbox.sh
```diff
@@ -0,0 +1,178 @@
...
+  echo "$workspace_path" >>"$GITHUB_PATH"
+  echo "$seeded" >>"$GITHUB_PATH"
+  echo "$bin" >>"$GITHUB_PATH"
+  ln -s "$bin" "$RUNNER_TEMP/tend-runner-home-alias"
+}
+
+setup() {
+  local agent_path path_entry
+  set_inputs
+  export TEND_SANDBOX_PATH="$GITHUB_WORKSPACE/.tend-explicit/bin"
```
tend-agent: Pairs with the assertion suggestion below — apply both, or neither. The entry on its own would sit in the PATH unexpanded without failing anything.

```suggestion
  # Two entries: the workspace path leads, and a literal `~` exercises the
  # expansion setup-sandbox does against the sandbox home.
  # shellcheck disable=SC2088
  export TEND_SANDBOX_PATH="$GITHUB_WORKSPACE/.tend-explicit/bin"$'\n~/.tend-tilde/bin'
```

proxy/test-setup-sandbox.sh
```diff
@@ -0,0 +1,178 @@
...
+    esac
+  done < <(tr : '\n' <<<"$agent_path")
+  case "$agent_path" in
+    "$GITHUB_WORKSPACE/.tend-explicit/bin":*) ;;
+    *) echo "::error::sandbox_path did not lead the PATH: $agent_path"; exit 1 ;;
+  esac
+  case ":$agent_path:" in
+    *:/home/tend-sandbox/.tend-seeded/bin:*) ;;
+    *) echo "::error::sandbox-owned counterpart omitted from PATH: $agent_path"; exit 1 ;;
+  esac
```
tend-agent, line 89:
```suggestion
  case ":$agent_path:" in
    *:/home/tend-sandbox/.tend-seeded/bin:*) ;;
    *) echo "::error::sandbox-owned counterpart omitted from PATH: $agent_path"; exit 1 ;;
  esac
  case ":$agent_path:" in
    *:/home/tend-sandbox/.tend-tilde/bin:*) ;;
    *) echo "::error::sandbox_path ~ was not expanded: $agent_path"; exit 1 ;;
  esac
```

The directory needn't exist — adopter entries are prepended verbatim, so this asserts the expansion and nothing else.

tend-agent (commented):
All four points from the last round are closed: the seeded-home rewrite is back, the `/usr/bin/bash` plus fixed-PATH pin now covers every step that takes a real credential (`gh`, `jq`, `git` and `date` are the only externals those six scripts call, and all four are in `/usr/bin`, so the narrowed PATH can't strand them), the workspace accept-side has tests on both the `sandbox_path` and the `$GITHUB_PATH` route, and `ln -sfn` is idempotent. `test-sandbox` is green on `a54f544`, whose per-entry loop is also the right shape — the old `*":$HOME/"*` substring test couldn't have coexisted with a workspace entry on the sandbox PATH, since the workspace lives under the runner home. Pinning `cleanup` to `/usr/bin/sudo` is worth having for the same reason `setup-sandbox.sh` pins its own PATH: `cleanup` is the one step that runs `if: always()`, so it is the one that can meet the planted `sudo` when `setup` died before removing it.

Three notes, none blocking.

**The rewrite is load-bearing again, and the caveat that guarded it is not.** `main` carried an explicit warning on this `case` ([`setup-sandbox.sh#L172-L181`](https://github.com/max-sixty/tend/blob/4b30b5b163e1834af2dca098e505c55dbc89ceb0/proxy/setup-sandbox.sh#L172-L181)): the `sudo -u "$SANDBOX" test -x` sitting next to the rewrite is not a backstop, it only rejects `/home/runner` paths because step 3 hasn't run yet, and once step 3 grants `o+x` on every ancestor of `$GITHUB_WORKSPACE` the sandbox uid can traverse the runner home. That warning left the file when the arm did, and this commit brings the arm back without it — so the one line that stops a runner-home path reaching the agent PATH now reads as if the guard below were a second line of defence. Suggestion inline; the wording is trimmed from the removed block.

**`sandbox_path`'s `~` expansion lost its only test.** Swapping the fixture from `'~/.cargo/bin'` to an absolute workspace path was the right move for what this commit is testing, but it also took out the only exercise of the `case "$dir" in "~"|"~/"*)` branch — nothing else in the repo sets `TEND_SANDBOX_PATH` with a tilde, and `docs/tend.example.yaml` still documents "A leading `~` expands to the sandbox home" as an adopter-facing lever. Two suggestions inline; they go together — the entry alone would pass whether or not the expansion happened.

**The shim message points at a lever that refuses the directory it is describing.** The shim only exists for a command selected from a runner-home directory, and `sandbox_path` hard-fails the job on exactly such an entry. An adopter reading "install it into `~/.local/bin` with sandbox_setup or configure sandbox_path" and reaching for the second half gets `::error::sandbox_path entry ... is under the runner's home` — a clear error, but a round-trip the message causes. `sandbox_path` is still a real answer when a copy exists somewhere off the runner home, so it's the qualifier that's missing rather than the mention.

<details><summary>What I checked and cleared</summary>

- Every step carrying `inputs.github_token`, `inputs.claude_code_oauth_token` or `inputs.anthropic_api_key` now has both the pinned shell and the `PATH=` prefix; `Run Claude` carries only the dummies, so the new comment's "a step receiving a real credential" claim is exact. `/usr/local/bin` being absent from the narrowed PATH costs nothing — none of the six scripts reaches for anything installed there.
- `_sandbox_home_path` is derived from the `readlink -f` canonical form, so the symlink-alias route into the runner home stays closed, and `_d` is never a runner-home path on the accepting branch — the boundary assertion in `setup()` (`*":$HOME/"*`) covers it.
- `_shared_path` is assigned on every branch that reaches its read; the two arms that skip it `continue` first. The accepting arm re-tests `-d` plus `test -x` a second time at the shared-path gate, which is a redundant `sudo` per home entry and not worth changing.
- Skel copy: `install -d -m 755` plus `chmod +x` gives `useradd -m` a 755 dir and 755 file to copy, so `sudo -u tend-sandbox test -x` answers true for the seeded counterpart. `cleanup` removes both.
- A command present in the runner directory but absent from the sandbox counterpart gets no shim now, and a same-named command at a different version is silently the sandbox's copy. Both match `main`, and `sandbox_setup:` for "home-scoped installs or version changes" in the example config names the second one. `sandbox-setup.sh`'s name diff still reports the first.

</details>

proxy/setup-sandbox.sh
max-sixty: Addressed in f54b1301: the runner-home rewrite now states explicitly that it is the filter and that the later executability check is not a security backstop.

proxy/setup-sandbox.sh
max-sixty: Addressed in f54b1301: the shim now qualifies that sandbox_path must point at a copy outside the runner home.

proxy/test-setup-sandbox.sh
max-sixty: Addressed in f54b1301: the paired assertion proves the literal leading tilde expands to the sandbox home.

proxy/test-setup-sandbox.sh
max-sixty: Addressed in f54b1301: the hosted fixture again configures a literal leading-tilde sandbox path.

tend-agent (approved):

PR #1049, opened 2026-08-25
test(proxy): assert --allow-hosts and the injection host sets stay in sync

tend-agent:
The proxy's credential boundary is split across two files that must agree, and nothing checks that they do: `inject_credentials.py`'s `BASIC_HOSTS`/`TOKEN_HOSTS`/`ANTHROPIC_HOSTS` decide which hosts get a real secret, while the `--allow-hosts` regex in `setup-sandbox.sh` decides which hosts mitmproxy TLS-intercepts at all. Both sides carry a "keep the two in sync" comment; neither is enforced. This adds the tests that enforce it.

They're in sync today — this is a guard against future drift, not a fix for a live break.

The failure it catches is silent and fleet-wide. A host that's in a frozenset but not in the regex is never intercepted, so its dummy credential is never swapped for the real one and every call to it 401s — while the proxy is alive, the CA is trusted, the addon's own unit tests pass, and each file reads correctly on its own. Nothing else in the repo would notice: the frozensets are covered by unit tests that never see the regex, and the regex is a string in a shell script no Python test reads.

`test_allow_hosts_regex_covers_every_injected_host` parses the flag's argument out of `setup-sandbox.sh` and asserts every injected host matches, both bare and in the `host:port` form. `host:port` is the form that matters — mitmproxy appends the port to every candidate it matches, so the bare-host assertion is extra strictness rather than a form production produces. `test_allow_hosts_regex_intercepts_nothing_extra` covers the other direction over five sampled hosts, so a widening those five catch fails — notably `objects.githubusercontent.com`, which the addon deliberately leaves out because injecting a PAT there collides with the signed-URL signature. It is a sample, not a proof: a widening that misses all five (adding `gist\.`, say) still passes.

The parsed regex is compiled with `re.IGNORECASE`, which is how mitmproxy compiles it in `next_layer._ignore_connection` — so a case-only edit to the regex behaves here the way it behaves in production. The parsing is otherwise deliberately literal (one single-quoted argument on one line, and exactly one occurrence in the file), so I noted that constraint in the comment above the flag rather than leaving the coupling implicit.

`uv run pytest` from the repo root: 593 passed. `pre-commit run` over the changed files: all hooks pass.

Mutation-tested against a temporarily edited `setup-sandbox.sh`, restoring it after each:

- dropping `uploads\.` from the regex → `AssertionError: uploads.github.com is injected but never intercepted`
- widening `raw\.githubusercontent\.com` to `[a-z]+\.githubusercontent\.com` → `AssertionError: objects.githubusercontent.com should not be intercepted`
- adding a commented `--allow-hosts '^example\.invalid$'` example above the flag → `AssertionError: expected one --allow-hosts flag in proxy/setup-sandbox.sh` (before the one-occurrence assert, the comment's value won the match and the tests validated a regex the proxy never uses)
- a case-only edit (`github\.com` → `GitHub\.com`) → passes, matching production behavior

The mitmproxy claims are read off the installed pin: `re.compile(x, re.IGNORECASE) for x in ctx.options.allow_hosts`, and `_ignore_connection` building every candidate as `f"{host}:{port}"` (peer address, server address, Host header with the port appended when absent, SNI).

No regression test in the usual sense — there's no bug to reproduce. The mutations above are the closest equivalent: each is a drift the tests would have caught and nothing else would.

Precedent for the shape: `test_pinned_mitmproxy_matches_the_action` in the same file already asserts cross-file parity between `claude/action.yaml` and the installed distribution, and the `sandbox-env-reserved-parity` pre-commit hook does the same for `config.py` vs. `setup-sandbox.sh`. A pre-commit hook would also have worked here; a test keeps it next to the addon tests that own the same boundary.

</details>

### DIFF
--- a/proxy/test_inject_credentials.py
+++ b/proxy/test_inject_credentials.py
@@ -40,9 +40,12 @@ def test_pinned_mitmproxy_matches_the_action() -> None:
 
 def _allow_hosts_regex() -> re.Pattern[str]:
     setup = (REPO_ROOT / "proxy" / "setup-sandbox.sh").read_text()
-    match = re.search(r"--allow-hosts '([^']*)'", setup)
-    assert match, "no single-quoted --allow-hosts flag in proxy/setup-sandbox.sh"
-    return re.compile(match.group(1))
+    # `[^']*` spans newlines, so a stray example in a comment would silently
+    # win the first match — require exactly one occurrence.
+    found = re.findall(r"--allow-hosts '([^']*)'", setup)
+    assert len(found) == 1, "expected one --allow-hosts flag in proxy/setup-sandbox.sh"
+    # mitmproxy compiles --allow-hosts with re.IGNORECASE; model that here.
+    return re.compile(found[0], re.IGNORECASE)
 
 
 def test_allow_hosts_regex_covers_every_injected_host() -> None:
@@ -51,9 +54,10 @@ def test_allow_hosts_regex_covers_every_injected_host() -> None:
     # intercepted, so its dummy is never swapped for the real secret and every
     # call to it 401s — with the proxy alive, the CA trusted and both files
     # looking correct in isolation. Both sides carry a "keep in sync" comment;
-    # only this asserts it. mitmproxy matches the pattern against the connection
-    # address, so check the bare host and the host:port form the regex's own
-    # optional port group exists to admit.
+    # only this asserts it. mitmproxy appends the port to every candidate it
+    # matches (peer address, Host header, SNI), so `host:port` is the form
+    # interception actually turns on — the bare host is extra strictness, not
+    # what production feeds the regex.
     allow_hosts = _allow_hosts_regex()
     for host in BASIC_HOSTS | TOKEN_HOSTS | ANTHROPIC_HOSTS:
         assert allow_hosts.search(host), f"{host} is injected but never intercepted"
--- a/proxy/test_inject_credentials.py
+++ b/proxy/test_inject_credentials.py
@@ -6,6 +6,7 @@ Run: ``uv run pytest`` from the repo root, which covers every Python suite.
 from __future__ import annotations
 
 import base64
+import re
 from importlib.metadata import version
 from pathlib import Path
 
@@ -13,7 +14,12 @@ import pytest
 from mitmproxy.test import tflow, tutils
 from ruamel.yaml import YAML
 
-from inject_credentials import CredentialInjector
+from inject_credentials import (
+    ANTHROPIC_HOSTS,
+    BASIC_HOSTS,
+    TOKEN_HOSTS,
+    CredentialInjector,
+)
 
 REPO_ROOT = Path(__file__).resolve().parents[1]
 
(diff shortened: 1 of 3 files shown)


proxy/test_inject_credentials.py
```diff
@@ -32,6 +38,51 @@ def test_pinned_mitmproxy_matches_the_action() -> None:
     assert version("mitmproxy") == action["inputs"]["mitmproxy_version"]["default"]
 
 
+def _allow_hosts_regex() -> re.Pattern[str]:
+    setup = (REPO_ROOT / "proxy" / "setup-sandbox.sh").read_text()
+    match = re.search(r"--allow-hosts '([^']*)'", setup)
+    assert match, "no single-quoted --allow-hosts flag in proxy/setup-sandbox.sh"
+    return re.compile(match.group(1))
```
tend-agent: mitmproxy compiles `--allow-hosts` with `re.IGNORECASE`, so this compiles a stricter matcher than production runs — and this file already leans on that case-insensitivity in `test_mixed_case_host_still_gets_credential` ("`--allow-hosts` matches case-insensitively, so a mixed-case host is intercepted"). A regex edit that only differs in case would fail here while working in production, or pass here while widening interception there.

While in the helper: `[^']*` spans newlines, so a stray `--allow-hosts '...'` example in a comment above the flag would silently win the first match and the tests would then validate a regex the proxy never uses (`.github/workflows/ci.yaml` already carries a line of exactly that shape). Requiring one occurrence makes that fail loudly, the same way the existing `assert match` does.

```suggestion
    # `[^']*` spans newlines, so a stray example in a comment would silently
    # win the first match — require exactly one occurrence.
    found = re.findall(r"--allow-hosts '([^']*)'", setup)
    assert len(found) == 1, "expected one --allow-hosts flag in proxy/setup-sandbox.sh"
    # mitmproxy compiles --allow-hosts with re.IGNORECASE; model that here.
    return re.compile(found[0], re.IGNORECASE)
```

proxy/test_inject_credentials.py
```diff
@@ -32,6 +38,51 @@ def test_pinned_mitmproxy_matches_the_action() -> None:
...
+
+def test_allow_hosts_regex_covers_every_injected_host() -> None:
+    # The frozensets scope injection; the --allow-hosts regex in setup-sandbox.sh
+    # scopes TLS interception. A host in a frozenset the regex misses is never
+    # intercepted, so its dummy is never swapped for the real secret and every
+    # call to it 401s — with the proxy alive, the CA trusted and both files
+    # looking correct in isolation. Both sides carry a "keep in sync" comment;
+    # only this asserts it. mitmproxy matches the pattern against the connection
+    # address, so check the bare host and the host:port form the regex's own
+    # optional port group exists to admit.
```
tend-agent: mitmproxy never matches a bare host: `_ignore_connection` appends the port to every candidate it builds — peer address, server address, the Host header (port appended when it lacks one) and the SNI. So `host:port` isn't one of two forms the regex has to admit, it's the only form; the optional port group exists because production always supplies a port, not because a bare host might arrive.

The bare assertion is fine to keep as slack, but the stated reason is what the next editor will treat as the spec — someone tightening the group to `(:[0-9]+)$` (closer to what production feeds it) would hit a spurious failure and have no way to tell it was spurious.

```suggestion
    # only this asserts it. mitmproxy appends the port to every candidate it
    # matches (peer address, Host header, SNI), so `host:port` is the form
    # interception actually turns on — the bare host is extra strictness, not
    # what production feeds the regex.
```

tend-agent (commented):
The guard is the right shape and the direction is right — `objects.githubusercontent.com` in the negative set is the case that actually matters. Two fidelity gaps in how the tests model mitmproxy's own matching, which is the thing they exist to pin down; suggestions inline.

`--allow-hosts` is matched with `re.IGNORECASE` (`mitmproxy/addons/next_layer.py`, `_ignore_connection`: `re.compile(x, re.IGNORECASE)` when the option is set, and `re.search(rex, host, re.IGNORECASE)` at match time), and every candidate it matches is built as `host:port` — `peername`, `server.address`, the Host header (port appended when absent) and the SNI all get `f"{...}:{port}"`. So the bare-host assertion tests a form production never produces, and the comment's rationale for it isn't what mitmproxy does. Both are harmless today; the cost is that the next editor reads the comment as the spec for the coupling.

One thing for the description rather than the code: `test_allow_hosts_regex_intercepts_nothing_extra` samples five hosts, so "widening the regex past what injection needs also fails" holds for the widenings those five catch, not in general — adding `gist\.` to the alternation passes every negative. Worth scoping that sentence to the sampled cases.

proxy/test_inject_credentials.py
tend-agent: Applied both in 3633fdf. Checked the claims against the installed pin before taking them: `next_layer.py` compiles `ctx.options.allow_hosts` with `re.IGNORECASE` and matches with `re.search(rex, host, re.IGNORECASE)`, so the test was compiling a stricter matcher than production.

Mutation-tested each half: a commented `--allow-hosts '^example\.invalid$'` above the flag now fails with `expected one --allow-hosts flag in proxy/setup-sandbox.sh` (before the one-occurrence assert, the comment's value won the match), and a case-only regex edit (`github\.com` → `GitHub\.com`) now passes, as it does in production.

proxy/test_inject_credentials.py
tend-agent: Applied in 3633fdf. Confirmed in `_ignore_connection`: every candidate is built as `f"{host}:{port}"` — peer address, server address, the Host header (port appended when it lacks one) and the SNI — so `host:port` is the only form the regex ever sees. Kept the bare assertion as slack with the corrected rationale, so someone tightening the port group can tell the failure is spurious. Also scoped the description's "widening past what injection needs also fails" sentence to the five sampled hosts.

PR #1053, opened 2026-08-25
fix(scripts): don't let tend-review gate the poll

### LINKED ISSUE #1050: poll-pr-checks.sh only filters the caller's own workflow, so a session that pushes to a PR burns its whole poll budget waiting on tend-review's agent job
`poll-pr-checks.sh` filters the caller's own workflow but not the *other* tend workflows, so a session that pushes to a PR spends its whole 9-minute poll budget waiting on `tend-review`'s agent job — which by construction starts at the same moment and usually runs longer than the cap.

## Mechanism

The rollup filter drops two things:

```jq
| select(.url | test($own) | not)          # this run's own check run
| select($wf == "" or .workflow != $wf)    # sibling runs of the same workflow
```

`$wf` is `$GITHUB_WORKFLOW`, so it only covers *the same* workflow. A `tend-nightly` or `tend-mention` session that pushes a commit to a PR and then polls that commit sees `tend-review / review` in the rollup as an ordinary pending check and blocks on it.

That is not an occasional collision — it is the guaranteed shape. `tend-review` fires on `synchronize`, so its `review` job is created *by the very push the poll is verifying*: the poll and the review agent start within a couple of seconds of each other. The review agent is a full Claude session; the poll cap is 9 iterations × 60 s. Observed `review` job durations on the two commits below were 9 m 11 s and 20 m 07 s, so exit 0 was unreachable in both cases.

Net effect: on any commit a tend session pushes to a PR, `poll-pr-checks.sh` reports exit 3 / `UNVERIFIED, not green` even when every repo check has settled green, and the session pays ~9 minutes of runner time and a poll round of tokens to learn nothing. `running-in-ci` then tells it not to re-enter the loop, so the CI-verification primitive is effectively unavailable on exactly the commits the bot is accountable for.

## Evidence

Both from `PRQL/prql`, 2026-08-25, on [PRQL/prql#6232](https://github.com/PRQL/prql/pull/6232).

| Session | Commit polled | Poll window | Verdict | `tend-review / review` on that commit |
|---|---|---|---|---|
| `tend-nightly` [32817217003](https://github.com/PRQL/prql/actions/runs/32817217003) | `894d2be` | 06:42:41Z → 06:51:46Z (9 m 05 s) | exit 3, sole pending check `review` | 06:42:43Z → 06:51:54Z (9 m 11 s) |
| `tend-mention` [32818775783](https://github.com/PRQL/prql/actions/runs/32818775783) | `1b40e85` | 06:57:23Z → 07:06:28Z (9 m 05 s) | exit 3, sole pending check `review` | 06:57:09Z → 07:17:16Z (20 m 07 s) |

The nightly's poll gave up **8 seconds** before the check it was waiting on completed. Timings are from the `statusCheckRollup` for each OID (`startedAt`/`completedAt` on the `CheckRun`) and from the session-log timestamps bracketing the `poll-pr-checks.sh` tool call.

In both cases `review` was the *only* entry in `pending` — every repo check was already terminal. The nightly's session ran 24 minutes total and spent 38% of it here; the mention session ran 15 m 42 s and spent 58%.

A third occurrence is recorded on 2026-08-17 in the adopter's tracking issue ([PRQL/prql#6123](https://github.com/PRQL/prql/issues/6123), Run 32008555121): *"`measure-code-cov` and tend's own `review` job were still running at the poll cap and are reported unverified, not green."*

Same reasoning applies to the other agent-job check contexts tend registers on a commit (`tend-mention / relay`, `verify`, `handle`), though those settle in seconds and have not been observed blocking.

## Proposed fix

The existing `$GITHUB_WORKFLOW` filter already encodes the principle — another tend agent's job is not a verdict on the code. Extending it to the generated fleet is one more clause in the same chain:

```jq
| select($wf == "" or .workflow != $wf)
| select(.workflow | startswith("tend-") | not)
```

Every generated workflow name carries the `tend-` prefix, and the generator owns those names, so the match is stable rather than a guess about adopter naming.

Trade-off worth deciding explicitly: this also hides a *red* `tend-review` check from the poller. That is a tend infrastructure failure rather than a code verdict, and the "Bot temporarily unavailable" outage issue already covers it — but if you'd rather keep it visible, the narrower variant is to exclude tend workflows from `pending` only and leave them in `failed`:

```jq
| {pending: [.[] | select(.status != "COMPLETED")
             | select((.workflow | startswith("tend-")) | not) | .name],
   failed:  [.[] | select(.status == "COMPLETED") | ... ]}
```

Checked against `plugins/tend-ci-runner/scripts/poll-pr-checks.sh` on `main` at the time of writing (post-#999 pagination rewrite, post-#1046 SHA guard) — the filter chain is unchanged there, so this is not already fixed past the 0.1.18 tag.

Related but distinct: #998/#999 (pagination), #1045/#1046 (unusable SHA), #42 (same-workflow deadlock, which is where the `$GITHUB_WORKFLOW` filter came from).

tend-agent:
`tend-review` starts on every push to a PR and often outlives the poller's nine-minute budget, so waiting for it makes the poll time out after the repository checks have settled. This change adds one exact workflow exemption for `tend-review`; the existing own-run and same-workflow filters remain. A rollup with no gating checks stays unverified.

Verification: `uv run pytest generator/tests/test_poll_scripts.py -q` (27 passed), plus pre-commit on the changed files.

Closes #1050


plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -144,7 +154,8 @@ rollup() {
           startedAt: ""}
        end
      | select(.url | test($own) | not)
-     | select($wf == "" or .workflow != $wf)]
+     | select($wf == "" or .workflow != $wf)
+     | select(.workflow | startswith("tend-") | not)]
```
tend-agent: `tend-install-test` matches this prefix but is a drift check, not an agent job — a red one would read as green here.

```suggestion
     | select(.workflow | (startswith("tend-") and . != "tend-install-test") | not)]
```

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -17,9 +17,19 @@
 # indistinguishable without admin scope.
 #
 # Filters out the current run's own check run ($GITHUB_RUN_ID — in flight for
-# as long as this loop runs) and sibling runs of the same workflow
+# as long as this loop runs), sibling runs of the same workflow
 # ($GITHUB_WORKFLOW — queued behind this run's concurrency group, so waiting
-# on them deadlocks). Check runs are grouped per (check name, workflow): a
+# on them deadlocks), and every other generated `tend-*` workflow. No tend
+# agent job is a verdict on the code — red ones included, which are tend
```
tend-agent: Pairs with the jq change — the exemption is every generated `tend-*` workflow *except* the drift check.

```suggestion
# on them deadlocks), and every other generated `tend-*` workflow bar
# `tend-install-test`, which verifies committed workflows against the generator
# rather than running an agent. No tend
# agent job is a verdict on the code — red ones included, which are tend
```

tend-agent (commented):
The mechanism is right and the diagnosis in #1050 holds up — `tend-review` is created by the very `synchronize` the poll verifies, so blocking on it is structural, not occasional. One clause in the same chain is the right depth, and keying it on the workflow rather than the check name is what keeps a repo's own `review` job gating. Two things below; the first is a concrete gap in the prefix match.

**`tend-install-test` is a generated `tend-*` workflow that *is* a verdict on the code.** `generate_install_test` in `generator/src/tend/workflows.py` stamps `name: tend-install-test`, and its one step runs `uvx "tend@$TEND_VERSION" init --with-install-test` then `git diff --quiet .github/workflows/` — a deterministic drift check, not an agent session. The new clause drops it, so the rationale in the header ("No tend agent job is a verdict on the code") doesn't cover the one workflow it exempts that isn't an agent job. Confirmed against the script's own jq — a single failing `install-test` check from workflow `tend-install-test` reduces to `{"pending":[],"failed":[]}`, i.e. exit 0 `green`. The window is narrow but reachable: `install-tend` is the documented path for changing an installed repo's config, it generates the file with `--with-install-test`, and on an already-installed repo `tend-review`/`tend-mention` do run on that PR — so a session that pushes to it and polls reads green with the drift check red. Inline suggestion on the jq line and the header sentence.

**Empty-after-filter now reads `green` where the header says UNVERIFIED.** The exit-code block documents "a commit with zero checks and zero statuses … the rollup is null then too, so 'nothing to gate on' is indistinguishable from 'nothing answered'" as exit 2. After this change a commit whose rollup holds *only* `tend-*` entries filters to `{"pending":[],"failed":[]}` and takes the exit-0 path with `green: every gating check on $SHA settled green` — the same epistemic state, opposite verdict. The hole predates the PR (own-run plus same-workflow could already empty the list) but this widens it from `tend-review`-as-caller to every caller: `tend-review`'s check run is guaranteed present on the commit within seconds of the push, so it's now the *tend* entry that keeps the rollup non-null while the repo's own checks may not have registered at the 60 s and 90 s samples. It fails toward a false green, which is the one direction this script is built not to fail in. Worth considering distinguishing "filtered to empty" from "nothing to filter" — if the pre-filter node list was non-empty and the post-filter list is empty, that's exit 2 rather than green. Not asking for it in this PR, but it's the case the new clause makes common.

Same pattern outside the diff: the pre-APPROVE rollup block in `plugins/tend-ci-runner/skills/review/SKILL.md` (the `FAILED=` jq under "Before APPROVE specifically") applies no own-run, own-workflow, or `tend-` filter at all, so a red `tend-mention / handle` sitting on the head commit withholds an approval for exactly the tend-outage-not-a-code-verdict reason this PR argues against. Different file, different failure (a withheld approval rather than a false green) — happy to push that as a follow-up if you want it tracked.

Testing looks right: `test_a_repo_check_named_like_tend_still_gates` is the pin that matters, since it's what stops the exemption drifting onto check-run names. `uv run pytest` is green here too (589 passed).

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
tend-agent: Applied in 35db26a, with the condition written in the same top-level `or` shape as the `$wf` clause above it rather than nested inside the `not`:

```jq
| select(.workflow == "tend-install-test"
         or (.workflow | startswith("tend-") | not))
```

Same semantics as the suggestion — keep the drift check, keep anything that isn't a `tend-` workflow, drop the rest. `test_install_test_gates_despite_the_tend_prefix` pins it: a failing `install-test` from workflow `tend-install-test` exits 1 and is named on stdout, and it reproduced your `{"pending":[],"failed":[]}` against the pre-fix script.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
tend-agent: Applied in 35db26a. I kept your framing but italicised the load-bearing word, since "no tend agent job is a verdict" is what makes the exception coherent rather than arbitrary:

> and every other generated `tend-*` workflow bar `tend-install-test`. No tend *agent* job is a verdict on the code […] `tend-install-test` is the one generated workflow that runs no agent: it regenerates the committed workflow files and diffs them, so a red one is drift in the code under poll and has to gate.

The exit-2 block and the `rollup()` header both name the new empty-after-filter case too.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -144,18 +166,22 @@ rollup() {
...
+    | if length == 0 then empty else
+      {pending: [.[] | select(.status != "COMPLETED") | .name],
        failed: [.[] | select(.status == "COMPLETED")
                 | select(.conclusion == "FAILURE" or .conclusion == "TIMED_OUT"
                          or .conclusion == "STARTUP_FAILURE"
                          or .conclusion == "ACTION_REQUIRED"
                          or .conclusion == "ERROR")
-                | "\(.name) \(.url)"]}' 2>/dev/null || true
+                | "\(.name) \(.url)"]}
+      end' 2>/dev/null || true
```
tend-agent: `select(length > 0)` is exactly `if length == 0 then empty else . end` on an array input, and keeps the object at its original indentation — the guard becomes one added line instead of a ten-line re-indent, so `git blame` on the reduction still points at the commit that wrote it. Semantics are identical; I ran the suite against this form and all 28 tests in `generator/tests/test_poll_scripts.py` pass, `test_filtering_to_empty_never_reads_green` included.

```suggestion
    | select(length > 0)
    | {pending: [.[] | select(.status != "COMPLETED") | .name],
       failed: [.[] | select(.status == "COMPLETED")
                | select(.conclusion == "FAILURE" or .conclusion == "TIMED_OUT"
                         or .conclusion == "STARTUP_FAILURE"
                         or .conclusion == "ACTION_REQUIRED"
                         or .conclusion == "ERROR")
                | "\(.name) \(.url)"]}' 2>/dev/null || true
```

tend-agent (commented):
Both findings from the last review landed and I verified them rather than taking the commit message's word: `tend-install-test` now exits 1 with the check named on stdout (`test_install_test_gates_despite_the_tend_prefix`), and an all-exempt rollup reaches exit 2 instead of the byte-identical-to-green reduction (`test_filtering_to_empty_never_reads_green`). I confirmed against `generate_install_test` in `generator/src/tend/workflows.py` that it really is the only generated workflow with no agent step — every other `name:` in `generator/src/tend/templates/` (`tend-review`, `tend-mention`, `tend-triage`, `tend-ci-fix`, `tend-notifications`, `tend-<scheduled>`) invokes the composite action. And the "read UNVERIFIED via the cap before" claim in the description checks out against the base: `483f274` has no `tend-` clause at all, so a tend-only rollup left `tend-review` pending and hit the cap. `uv run pytest generator/tests/test_poll_scripts.py` is green here (28 passed).

One simplification inline. Nothing blocking.

Two smaller notes:

**Nothing pins `tend-install-test` as the *only* agentless generated workflow.** The exception is correct today because the generator has exactly one such workflow, but that fact lives only in a comment. A second deterministic generated workflow — another drift or lint check — would silently inherit the `tend-` exemption and re-open exactly the false green this commit closes, with no test going red. A cheap pin would be a generator test asserting every `tend-*` name it stamps either invokes the composite action or is named in the script's exception. Not asking for it here; flagging it since the failure mode is a wrong outward verdict rather than wasted compute.

**Textual overlap with #1055.** Different fix, not a duplicate — it makes an OID that resolves to no object terminal instead of a retry. But it edits the same `rollup()` header block and the same exit-2 enumeration from a base that predates this PR, so whichever lands second will conflict there. Worth knowing before merge order is decided; no semantic collision that I can see, since its `NOSUCHCOMMIT` early-return sits above the filter chain this PR touches.

tend-agent:
Both review findings are applied in 35db26a and replied to inline; required checks are green on that commit.

I took the second one (empty-after-filter) into this PR rather than leaving it as a follow-up, despite the review not asking for it here. The reasoning: the exemption and the guard are one safety property, not two changes. Without the guard this PR converts a tend-only rollup from UNVERIFIED-via-the-cap into `green` — `tend-review` registers within seconds of the push and keeps the rollup non-null, so the existing exit-2 check stops firing in exactly the window it was written for. Shipping the exemption alone would have introduced a false green rather than merely leaving one in place, and that is the one direction this script exists not to fail in.

On the `review/SKILL.md` pre-APPROVE rollup: the gap is real — that `FAILED=` jq applies no own-run, own-workflow, or `tend-` filter, so a red `tend-mention / handle` on the head commit withholds an approval for the reason this PR argues against. I'm leaving it out of this PR to keep the change atomic: different file, opposite failure direction (a withheld approval is conservative; a false green is not), and it wants its own reasoning about whether an approval should be gated more strictly than a poll. Happy to open it as a separate PR — say the word rather than assuming, since it changes when the bot approves.

<details><summary>What the two clauses do now</summary>

```jq
| select($wf == "" or .workflow != $wf)
| select(.workflow == "tend-install-test"
         or (.workflow | startswith("tend-") | not))
```

`tend-install-test` is the one generated workflow with no agent step — `generate_install_test` in `generator/src/tend/workflows.py` stamps a job that runs `uvx tend@$TEND_VERSION init --with-install-test` and `git diff --quiet .github/workflows/`, so a red one is drift in the code under poll.

The reduction then emits nothing when the filtered group list is empty, so the loop keeps polling and the cap reports UNVERIFIED instead of `{pending: [], failed: []}`, which is byte-identical to settled green.

Four tests cover it, each verified failing against the pre-fix script: `test_other_tend_workflows_do_not_gate`, `test_a_repo_check_named_like_tend_still_gates`, `test_install_test_gates_despite_the_tend_prefix`, `test_filtering_to_empty_never_reads_green`.

</details>

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
tend-agent: Applied in 86c9399. Agreed on both counts — `select(length > 0)` is the same guard with the reduction left where it was, so the diff is one added line rather than a re-indent of nine. `generator/tests/test_poll_scripts.py` is green on this form here (28 passed), `test_filtering_to_empty_never_reads_green` included.

tend-agent:
The inline simplification is applied in 86c9399, and I took the first note into this PR rather than leaving it flagged.

**The pin now exists.** `test_poll_exemptions_match_the_agentless_generated_workflows` in `generator/tests/test_repo_pins.py` generates every workflow (including `--with-install-test`), scrapes the `.workflow == "tend-…"` names back out of `poll-pr-checks.sh`, and asserts the two sets are complements: a stamped `tend-*` workflow either invokes `max-sixty/tend/<harness>` or is named back in as a gating check — never both, never neither. That is the invariant this PR's exemption rests on, so it belongs with the exemption rather than after it; a test is also about as cheap as machinery gets, and the failure it prevents is a wrong outward verdict.

I verified it goes red in both directions rather than trusting that it passes: renaming the `tend-install-test` exception fails with *"runs no agent job, so the `tend-` exemption … drops it from the rollup and a red one reads as green"*, and adding `tend-review` to that list fails with the converse. `uv run pytest` is green here (592 passed), as is `pre-commit run` over both changed files.

**On #1055:** agreed it's a textual conflict and not a semantic one, and it looks one-directional — its `NOSUCHCOMMIT` early return sits above this filter chain, and this commit only touches the reduction at the bottom of `rollup()` plus the header paragraph. Whichever lands second rebases over the exit-2 enumeration and the `rollup()` header block by hand. No preference on order from here.

<details><summary>What the new test asserts, and why the invariant isn't self-evident</summary>

The exemption's rationale is "no tend *agent* job is a verdict on the code". That is a claim about what the generator stamps, not about the `tend-` prefix, and nothing enforced the two staying aligned — `generate_install_test` happens to be the only agentless generated workflow today, and that fact lived only in a comment. A second deterministic generated workflow (another drift check, a lint job) would inherit the prefix exemption silently, and the only symptom would be a caller reading `green` on a commit whose drift check was red.

Both operands are in-repo, so the test can only fail on a change to one of them — it never goes red on an upstream event.

</details>

tend-agent (commented):
The pin from the last review landed, and I checked it against the script rather than taking the commit message's word. `test_poll_exemptions_match_the_agentless_generated_workflows` goes red in both directions: naming `tend-review` back into the exemption list fails with *"runs an agent job but poll-pr-checks.sh names it back in as a gating check"*, and dropping the `tend-install-test` clause fails with the converse. Both operands are in-repo, so it can only fail on a change to one of them — that is the invariant the prefix exemption rests on, and it now has something holding it.

`select(length > 0)` is the same filter as the `if length == 0 then empty` form on an array input, and the reduction below it is untouched. `uv run pytest generator/tests/test_repo_pins.py generator/tests/test_poll_scripts.py` green here (30 passed), `pre-commit` clean over both changed files.

Incidentally the rollup on `86c9399` is the shape this PR is about: every `ci` check terminal green, `tend-review / review` the sole pending entry.

No new findings on the increment; the two notes from the last review are both closed. Can't approve my own PR, so this is the verdict — nothing blocking.

tend-agent (commented):
The base merge resolved the `test_repo_pins.py` conflict with a single blank line between the end of `test_poll_exemptions_match_the_agentless_generated_workflows` and the `ACTION_PATH_REF` comment block that follows it. `ruff-format` wants two there, so `lint` will go red on this commit:

```
$ ruff 0.14.11 format --check --diff generator/tests/test_repo_pins.py
@@ -92,6 +92,7 @@
             "as green — name it back in alongside `tend-install-test`"
         )
 
+
 # Every `${{ github.action_path }}/…` reference in the two composite actions.
1 file would be reformatted
```

Inline suggestion below. Nothing else on the increment — the resolution kept both sides of the conflict, and the invariant the new test rests on survives main's `shared/steps/` reorganisation (`generate_all`'s signature is unchanged and every `tend-*` template still stamps `uses: max-sixty/tend/`).

generator/tests/test_repo_pins.py
```diff
@@ -53,6 +57,41 @@ def test_uv_build_range_admits_the_pinned_uv() -> None:
...
+        assert runs_agent != (name in named_back_in), f"`{name}` " + (
+            "runs an agent job but poll-pr-checks.sh names it back in as a "
+            "gating check, so a tend outage on it holds up every caller's poll"
+            if runs_agent
+            else "runs no agent job, so the `tend-` exemption in "
+            "poll-pr-checks.sh drops it from the rollup and a red one reads "
+            "as green — name it back in alongside `tend-install-test`"
+        )
+
 # Every `${{ github.action_path }}/…` reference in the two composite actions.
```
tend-agent, line 56:
```suggestion

# Every `${{ github.action_path }}/…` reference in the two composite actions.
```

tend-agent: Applied in e24d129 — `lint` is green on that commit (run [33060344544](https://github.com/max-sixty/tend/actions/runs/33060344544)), along with the rest of the required checks, so the merge resolution's formatting is settled.

generator/tests/test_poll_scripts.py
```diff
@@ -308,6 +308,42 @@ def test_own_run_and_same_workflow_are_filtered(env: dict[str, str]) -> None:
...
+def test_tend_review_does_not_gate(env: dict[str, str]) -> None:
+    """`tend-review` fires on the very push this poll is verifying, so its
+    agent job is created seconds after the loop starts and routinely outlives
+    the 9-minute cap — every session that pushes to a PR would report
+    UNVERIFIED with every repo check already green."""
+    env = env | {"GITHUB_WORKFLOW": "tend-nightly"}
+    _serve(
+        env,
+        _resp(
+            _check_run("tests"),
```
tend-agent: Naming the green companion `review` too is what restores the pin `2e9407b` removed — it makes this test fail under a check-name-keyed filter (both entries drop, the rollup filters to empty, exit 2 against the `returncode == 0` below) while staying exit 0 under the workflow-keyed one it actually has.

```suggestion
            _check_run("review", workflow="ci"),
```

tend-agent (commented):
Narrowing to the one workflow is right, and I checked the premise rather than the argument: across the twelve most recent PRs the only tend check runs that land on a PR head commit are `tend-review / review` and `tend-mention`'s `relay` (one dispatch POST, seconds, 42 for 42 successful to date). `tend-mention`'s `verify` and `handle` are always `SKIPPED` on a PR head, and structurally so — `verify`'s `if:` admits only `repository_dispatch`, `issues`, and `issue_comment`, none of which carries the PR head as its run's head SHA, so the agent job that could outlive the cap never registers on the commit under poll. `tend-review` is the whole exposure, which makes the prefix rule, its hand-maintained exception, and the generator pin test machinery for a set of one. Deleting all three is the right depth. `ci` is green on `e8fd471f`, `test` included.

Two coverage holes, though, both provable by mutating the filter chain, and both closeable without adding a test.

**The exemption's keying on the workflow rather than the check-run name is no longer observed.** `2e9407b` removed the only fixture pairing a check *named* `review` with a non-exempt workflow, and the failure coverage it defers to — `test_red_names_the_failing_check_with_its_url` — uses checks named `tests` and `lint`, which can never collide with the exemption. Rewriting the clause as `select(.name != "review")` now leaves the whole file green. This script ships to every adopter, and an adopter whose own CI job is called `review` would have its red silently exempted. The inline suggestion restores the pin in place: with the green companion renamed, a name-keyed filter drops both entries, the rollup filters to empty, and the test's `returncode == 0` fails on exit 2.

**The same-workflow (`$wf`) filter now has no test at all.** `test_own_run_and_same_workflow_are_filtered` is the only test whose fixture holds an entry with `.workflow` equal to its env's `GITHUB_WORKFLOW` — `_check_run("handle", status="QUEUED", workflow="tend-review", run_id=777)` against the `env` fixture's `"GITHUB_WORKFLOW": "tend-review"`. On main that entry was removed only by `select($wf == "" or .workflow != $wf)`; this PR's new clause removes it too, so deleting the `$wf` clause outright leaves the suite green. That clause is what stops a session deadlocking on its own concurrency-queued sibling, which is the failure class #1050 is about. It's outside the diff, so two lines rather than an inline suggestion:

```python
    env = env | {"GITHUB_WORKFLOW": "tend-nightly"}
    _serve(
        env,
        _resp(
            _check_run("tests"),
            _check_run("review", status="IN_PROGRESS", run_id=555, workflow="x"),
            _check_run("handle", status="QUEUED", workflow="tend-nightly", run_id=777),
        ),
    )
```

Your call on whether either is worth keeping — but a fixture edit is about as cheap as machinery gets, and CLAUDE.md's Development section puts preventing a wrong outward action above the cost of the mechanism that prevents it. Both holes fail toward a false green, which is the direction this script exists not to fail in.

<details><summary>How the two holes were verified</summary>

`uv` isn't on the sandbox's PATH, so this is not a pytest run: I extracted the reduction from `rollup()` and ran it under `jq` over each remaining fixture, mapping the output to the exit code the loop would take (pending non-empty then cap, exit 3; failed non-empty, exit 1; both empty, exit 0; no output, exit 2). `M1` deletes `select($wf == "" or .workflow != $wf)`; `M2` replaces `select(.workflow != "tend-review")` with `select(.name != "review")`.

| fixture | asserts | base | M1 | M2 |
|---|---|---|---|---|
| `test_own_run_and_same_workflow_are_filtered` | exit 0 | exit 0 | exit 0 | exit 0 |
| `test_tend_review_does_not_gate` | exit 0 | exit 0 | exit 0 | exit 0 |
| `test_filtering_to_empty_never_reads_green` | exit 2 | exit 2 | exit 2 | exit 2 |
| `test_red_names_the_failing_check_with_its_url` | exit 1 | exit 1 | exit 1 | exit 1 |
| `test_superseded_failure_yields_to_its_replacement` | exit 1 | exit 1 | exit 1 | exit 1 |

Every remaining fixture is verdict-identical under both mutations. With the two fixes applied, `M1` moves `test_own_run_and_same_workflow_are_filtered` to exit 3 and `M2` moves `test_tend_review_does_not_gate` to exit 2 — each fails its own assertion, which is the pin.

The `tend-mention` claim above comes from the rollups of the twelve most recent PRs plus `verify`'s `if:` block in `.github/workflows/tend-mention.yaml`, and from the workflow's run history (42 `pull_request_review` runs, all `success`).

</details>

generator/tests/test_poll_scripts.py
max-sixty: Applied this fixture change and the same-workflow fixture update from the review body. Both filters now fail under the mutations the review described.

tend-agent:
Both findings were already applied in `5254ce6` before this session's edits landed, so I dropped my duplicate and pushed only what was left: `276f643` adds a sentence to each of the two docstrings saying the fixture values are load-bearing. `test_other_workflows_still_gate` was deleted in `2e9407b` for looking redundant, and after `5254ce6` the pin lives in a check name and a workflow string with nothing recording why — renaming either back leaves the suite green while a false green becomes reachable again. `ci` is green on `276f643`; `#1055` has since merged into the branch and left both docstrings intact.

<details><summary>How the two holes were re-verified</summary>

`uv` isn't on the sandbox PATH, so this isn't a pytest run — but it isn't the review's hand-extracted jq chain either: I ran the real `poll-pr-checks.sh` against a fake `gh` and `sleep` on PATH (the shape the pytest fixtures use) and read its actual exit codes.

| fixture | asserts | base | M1 (drop `$wf` clause) | M2 (`select(.name != "review")`) |
|---|---|---|---|---|
| `test_own_run_and_same_workflow_are_filtered` — before `5254ce6` | 0 | 0 | **0** | 0 |
| `test_own_run_and_same_workflow_are_filtered` — after | 0 | 0 | 3 | 0 |
| `test_tend_review_does_not_gate` — before `5254ce6` | 0 | 0 | 0 | **0** |
| `test_tend_review_does_not_gate` — after | 0 | 0 | 0 | 2 |

The bold cells are the holes; each is closed by the row below it. Deleting the own-run (`$own`) filter also still moves `test_own_run_and_same_workflow_are_filtered` to exit 3, so that pin survived the fixture edit.

</details>

PR #1055, opened 2026-08-25
fix(scripts): fail fast when a commit cannot be resolved

### LINKED ISSUE #1054: poll-pr-checks.sh: a well-formed but nonexistent OID still burns the whole poll budget and still trips head_note's false "branch advanced" — twice in one day on a consumer
`#1046`'s entry guard narrowed `head_note`'s false "branch advanced" claim to the abbreviated case and its header comment says so explicitly — "an unresolvable OID and a merge-ref commit are both full-length and neither is the branch head — so this narrows that note rather than fixing it". This is field data on the residual it names, plus a way to close it that costs no extra API call, because the unresolvable-OID case turns out to be distinguishable inside `rollup()` for free.

Two sessions on `numbagg/numbagg` hit it in the same 24h window, in two different workflows. Both passed a **well-formed 40-hex OID that does not exist** — the model extended an abbreviated SHA it had seen in `git` output instead of running `git rev-parse HEAD`, inventing characters 8 onward. Each burned the full nine-iteration budget and then printed a branch-advance that never happened.

| Run | Workflow | OID passed | Real head | Poll wall-clock |
|---|---|---|---|---|
| [32705599987](https://github.com/numbagg/numbagg/actions/runs/32705599987) | `tend-mention` (PR `numbagg/numbagg#758`) | `5a6171e0da2c8e8ac9d3b6c2c5e28e4dd8d2b1a5` | `5a6171e9a15a8a35625acdea2e998f5c08ca9205` | 9m03.7s (08:33:52.880Z → 08:42:56.612Z) |
| [32820277759](https://github.com/numbagg/numbagg/actions/runs/32820277759) | `tend-nightly` (PR `numbagg/numbagg#762`) | `bfcddcf3fd5aa4c8bbbd5f5e50e4bfb60b0d54e5` | `bfcddcf88b23b5b2929c6a5ef78947106f36f491` | 9m07.9s (07:23:34.269Z → 07:32:42.202Z) |

Both got the identical pair of lines, the second of which is false — nobody pushed, and the named commit is the session's own, the one it had meant to pass:

```
no rollup returned for 5a6171e0da2c8e8ac9d3b6c2c5e28e4dd8d2b1a5 — UNVERIFIED, not green
note: branch advanced to 5a6171e9a15a8a35625acdea2e998f5c08ca9205 — the result above is still 5a6171e0da2c8e8ac9d3b6c2c5e28e4dd8d2b1a5's, the commit this run is accountable for
```

Neither session relayed the note — both recovered with `git rev-parse HEAD` and re-polled to a true green — so nothing wrong reached either PR thread from the script itself. The caller-side fabrication is a model lapse I'm treating as stochastic and not proposing guidance for; what's actionable here is the script.

## Why the guard can't catch it, and why the loop can

`^[0-9a-f]{40}$` passes: the input is well-formed, just not a commit. GitHub answers an unresolvable OID with a **successful** response, not an errors envelope:

```
$ gh api graphql -f owner=numbagg -f name=numbagg \
    -f oid=bfcddcf3fd5aa4c8bbbd5f5e50e4bfb60b0d54e5 \
    -f query='query($owner: String!, $name: String!, $oid: GitObjectID!) {
      repository(owner: $owner, name: $name) { object(oid: $oid) { __typename } } }'
{"data":{"repository":{"object":null}}}      # exit 0

$ ... -f oid=bfcddcf88b23b5b2929c6a5ef78947106f36f491 ...
{"data":{"repository":{"object":{"__typename":"Commit"}}}}
```

So it reaches [the jq at line 119](https://github.com/max-sixty/tend/blob/483f274bfd1c77094805b03edf94346c8afd9a6c/plugins/tend-ci-runner/scripts/poll-pr-checks.sh#L119), where `.data.repository.object.statusCheckRollup.contexts` on a null `object` yields null, `select(. != null)` drops it, `page` is empty, and `rollup()` returns empty — the same value the loop reads as "transient blip, retry". Nine `sleep 60`s later it exits 2 and `head_note` compares the caller's OID against a head it was never going to equal.

The key point is that the conflation isn't forced. An errors envelope never gets this far: `gh api graphql` exits non-zero on it, and [line 117](https://github.com/max-sixty/tend/blob/483f274bfd1c77094805b03edf94346c8afd9a6c/plugins/tend-ci-runner/scripts/poll-pr-checks.sh#L117)'s `|| return 0` already returns empty for that. By the time `$resp` is in hand, `.data.repository.object == null` means exactly one thing: this OID is not an object in this repository. A real commit with no rollup is a different shape — `object` non-null, `statusCheckRollup` null — so the two don't collide.

## Suggested shape

Emit a sentinel from `rollup()` the way `"OVERFLOW"` already works, and treat it in the loop as terminal:

- in `rollup()`, on a successful `$resp` with `.data.repository.object == null`, print `"NOSUCHCOMMIT"` and return;
- in the loop, alongside the existing `[ "$cur" = '"OVERFLOW"' ] && too_many_contexts`, exit 2 with a message naming the cause (`<sha> is not a commit in this repository`) and **skip `head_note`** — with the OID unresolvable there is no "still `$SHA`'s result" to qualify, so the note has nothing true to say.

That is no extra API call and no new failure mode: it turns nine minutes of sleeping into a verdict on the first iteration, and it deletes the false-claim emission for this input class rather than narrowing it again.

One caveat worth building in: a commit pushed seconds earlier could in principle read as `object: null` through brief replication lag, and exiting at 60s would turn that into a spurious UNVERIFIED on a genuine push. Requiring two consecutive `NOSUCHCOMMIT` observations before treating it as terminal keeps that safe and still returns at ~2 minutes instead of 9.

<details><summary>Provenance</summary>

Found by `review-runs` on `numbagg/numbagg` ([run 32826311957](https://github.com/numbagg/numbagg/actions/runs/32826311957)) over the 2026-08-24T08:28Z → 2026-08-25T08:24Z window. The consumer pins `max-sixty/tend/claude@0.1.18`, whose copy of the script predates #1046 — but the diagnosis above is against `main` at `483f274b`, post-merge, and the residual is present there. Both measurements come from the session JSONL tool-call and tool-result timestamps, not from job durations. The `object: null` transcript was run against this repo at analysis time.

Prior context: #1045 (the abbreviated case, two occurrences) → #1046.

</details>

tend-agent:
A well-formed OID that names no commit currently spends the full poll budget and can emit a false branch-advanced note. The poller now validates the commit before entering the loop, retries once after ten seconds for a just-pushed commit, and returns unverified without the note if the lookup still fails. This keeps commit validation outside the rollup state machine.

Verification: `uv run pytest generator/tests/test_poll_scripts.py -q` (25 passed), plus pre-commit on the changed files.

Closes #1054


plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -172,17 +184,38 @@ head_note() {
...
+  if [ "$cur" = "NOSUCHCOMMIT" ]; then
+    absent=$((absent + 1))
+    if [ "$absent" -ge 2 ]; then
+      echo "$SHA is not a commit in $GITHUB_REPOSITORY — UNVERIFIED, not green"
+      exit 2
+    fi
+    continue
+  fi
+  absent=0
   [ -z "$cur" ] && continue
```
tend-agent: An empty `$cur` is a transient failure — no answer about the OID — but it clears `absent` before the blip check sees it, so `NOSUCHCOMMIT` → blip → `NOSUCHCOMMIT` reads as two isolated first sightings and never reaches the verdict. The run then falls out at the cap into `no rollup returned for $SHA` plus `head_note`, i.e. exactly the nine-minute wait and the false "branch advanced" note this PR removes, just via a rarer path. Swapping the two lines keeps a blip from resetting evidence it doesn't contradict:

```suggestion
  [ -z "$cur" ] && continue
  absent=0
```

tend-agent (commented):
Reviewing my own PR, so this is a COMMENT rather than an approval.

The sentinel holds up. `object: null` at exit 0 really is unambiguous — `gh api graphql` exits non-zero on an errors envelope, so `|| return 0` takes that path before the new test is reached — and it can't collide with a real commit carrying no rollup, which keeps `object` non-null. Full `uv run pytest generator/tests/test_poll_scripts.py` is green here (26 passed). Three things on the edges:

1. `absent=0` sits *above* the blip check, so an empty rollup between two absent sightings clears the counter: `NOSUCHCOMMIT` → blip → `NOSUCHCOMMIT` never reaches the verdict, and the run falls back to the cap path carrying both symptoms this PR set out to remove — nine minutes of sleeping, then `head_note`'s false "branch advanced". An empty rollup is "no answer", not evidence the commit exists, so it shouldn't reset. Moving the line one down also makes the header's claim that a merge-ref commit "remains the one exit-2 cause head_note can still misfire on" true rather than nearly true.

2. The `case` guard in the grace re-check is the one line in this diff whose loss is a **false green**, and nothing pins it. I confirmed that locally by reverting it to `[ -z "$cur" ] && continue`: a rollup carrying a `FAILURE` then exits 0, because both `[ "" -gt 0 ]` tests error out on the sentinel (`integer expression expected`) and the script falls through to `echo "green"`. A future edit that simplifies that `case` back to a `-z` test passes the whole suite today.

3. `test_unresolvable_oid_is_terminal_not_a_transient_blip` asserts `HEAD_SHA in out` and `"UNVERIFIED, not green" in out`, but the old fallback line — `no rollup returned for $SHA — UNVERIFIED, not green` — matches both, so only the call-count assert actually distinguishes the new verdict from the behaviour before the fix. Worth pinning the string the PR body says it asserts.

Pushing all three, since there's no human author to act on them.

plugins/tend-ci-runner/scripts/poll-pr-checks.sh
```diff
@@ -172,17 +184,38 @@ head_note() {
...
   [ "$(printf '%s' "$R" | jq '.pending | length')" -gt 0 ] && continue
   # A just-settled matrix can register its `if: always()` omnibus a second or
   # two later; re-check once before trusting pending == 0.
   sleep 30
   cur=$(rollup)
-  [ -z "$cur" ] && continue
+  # The sentinel is not a rollup: reaching $R it would blow up the jq below
+  # and fall through to green. It cannot follow a resolving poll in practice,
+  # so treat it as this iteration's blip rather than a second exit path.
+  case "$cur" in "" | NOSUCHCOMMIT) continue ;; esac
```
tend-agent: This guard is the one line in the diff whose loss is a **false green**, and no test pins it. Reverting it locally to `[ -z "$cur" ] && continue` and serving a settled rollup, then the sentinel, then a rollup carrying a `FAILURE`, exits **0** — both `[ "" -gt 0 ]` tests die with `integer expression expected`, the red branch is skipped, and the script reaches `echo "green"`. Adding a regression test in the follow-up commit so a later simplification back to a `-z` test can't pass the suite.

generator/tests/test_poll_scripts.py
```diff
@@ -335,6 +340,42 @@ def test_null_rollup_never_reads_green(env: dict[str, str]) -> None:
...
+    thing, so it is a verdict on the second sighting, not a retry."""
+    Path(env["HEAD_JSON"]).write_text(json.dumps({"headRefOid": "c" * 40}))
+    _serve(env, NO_SUCH_COMMIT)
+
+    result = _poll(env)
+    out = result.stdout + result.stderr
+
+    assert result.returncode == 2, out
+    assert "UNVERIFIED, not green" in out
+    assert HEAD_SHA in out
```
tend-agent: The pre-fix path prints `no rollup returned for $SHA — UNVERIFIED, not green`, which satisfies both this assert and the `"UNVERIFIED, not green"` one above it — only the call-count assert below separates the two behaviours. Pinning the verdict line itself makes the test say what the PR body claims it says:

```suggestion
    assert f"{HEAD_SHA} is not a commit in owner/repo" in out
```

tend-agent (commented):
Reviewing my own PR, so this is a COMMENT rather than an approval.

All three points from the last round landed correctly. `absent=0` now sits below the blip check, so `NOSUCHCOMMIT` → blip → `NOSUCHCOMMIT` reaches the verdict on the third poll; `test_the_sentinel_at_the_grace_recheck_never_reads_green` genuinely pins the `case` guard (reverting it to `[ -z "$cur" ] && continue` makes that test exit 0 instead of 1); and the verdict-line assert now separates the new path from the old fallback. Full `uv run pytest generator/tests/test_poll_scripts.py` is green here (28 passed), `shellcheck -S warning` is clean.

One thing the new increment introduced, and one comment that drifted with it.

1. **The verdict can now contradict a rollup this same run already read.** `absent` is reset only by a usable rollup at the *top* of an iteration — but the grace re-check one line further down sets `R` and then `continue`s on the sentinel without touching `absent`. So a green rollup followed by two sentinel answers exits 2 claiming the commit doesn't exist. Reproduced with the suite's own fixtures — `_serve(env, _resp(_check_run("tests")), NO_SUCH_COMMIT)` gives `<sha> is not a commit in owner/repo — UNVERIFIED, not green`, after the script has read and reduced that commit's check rollup. The gate outcome is UNVERIFIED either way, so nothing goes false-green; what's wrong is the claim, and it's the same shape as the false "branch advanced" note this PR exists to remove — a categorical statement about a commit, contradicted by what the run itself saw. A non-empty `R` is positive proof the OID resolves, so a later null is a blip whatever the count — which is exactly how the grace re-check already treats it. Guarding the exit on `[ -z "$R" ]` costs one condition and leaves every existing test on the same path.

2. `Two consecutive sightings rather than one` in the block comment no longer holds after the `absent=0` move — a blip between two sightings still reaches the verdict, which is the point of the fix, and the comment two lines down says so. In a file where the comments carry the design rationale, the two shouldn't read against each other.

Pushing both, plus a regression test for the first, since there's no human author to act on them.

tend-agent (commented):
Reviewing my own PR, so this is a COMMENT rather than an approval.

Both points from the last round landed correctly. The `[ -z "$R" ]` guard is genuinely pinned — reverting it to `[ "$absent" -ge 2 ]` fails exactly `test_a_rollup_already_read_outranks_a_later_absence` and nothing else — and the block comment no longer claims the two sightings must be consecutive. `uv run pytest generator/tests/test_poll_scripts.py` is green here (29 passed); CI is green on this head apart from this review's own job.

One inline suggestion on the new test, plus a note on something pre-existing that the increment now touches.

**The pre-existing part, recorded here rather than fixed here.** When the retained `$R` has `pending == []` and every later poll answers nothing, the cap report prints `poll cap hit — still pending on $SHA (UNVERIFIED, not green):` followed by an empty list — a claim contradicted by the only rollup the run holds. It is reachable on the unmodified script through empty blips at the grace re-check (`_serve(env, _resp(_check_run("tests")), NULL_ROLLUP)` gives exit 3 with nothing under the header), so this PR doesn't introduce it. What changed is that the guard routes the sentinel onto the same report, and `test_a_rollup_already_read_outranks_a_later_absence` serves precisely that fixture — so the misreport is now covered by a test that doesn't look at it. It belongs in its own change: it's about what the cap report says, not about how a null `object` is classified, and folding it in would mix the two. Verdict direction is safe either way (UNVERIFIED), so it isn't blocking.

- Guard is load-bearing and pinned: `sed`-reverting line 207 to `if [ "$absent" -ge 2 ]; then` leaves 28 passed, 1 failed — only the new test.
- New-test scenario's actual outcome: exit 3, stdout `poll cap hit — still pending on <sha> (UNVERIFIED, not green):` with an empty list, 10 GraphQL calls.
- Same output from the unmodified script via `_serve(env, _resp(_check_run("tests")), NULL_ROLLUP)`, which is what places the cap-report wording outside this PR's scope.

</details>

Pushing the inline suggestion, since there's no human author to act on it.

generator/tests/test_poll_scripts.py
```diff
@@ -335,6 +341,103 @@ def test_null_rollup_never_reads_green(env: dict[str, str]) -> None:
...
+    states the commit is not in the repository having just reported its checks,
+    the same contradicted claim the "branch advanced" note made."""
+    _serve(env, _resp(_check_run("tests")), NO_SUCH_COMMIT)
+
+    result = _poll(env)
+    out = result.stdout + result.stderr
+
+    assert "is not a commit in owner/repo" not in out, (
+        "the verdict overruled a rollup this run had already read"
+    )
```
tend-agent: This asserts only the absence of a string and nothing about the outcome, so it passes on any run that fails to produce it — a crash, a `set -e` death, a verdict re-routed somewhere else. Every other test in this file pins the return code. Exit 3 is the documented path here (a usable `$R` is retained, so the cap report owns the verdict rather than the no-rollup branch), and pinning it keeps a future change to that routing visible instead of silently satisfying the negative.

```suggestion
    )
    assert result.returncode == 3, out
```

tend-agent (commented):
Reviewing my own PR, so this is a COMMENT rather than an approval.

Nothing outstanding on `40bdae0`. The refactor deletes the sentinel rather than refining it, which is the right direction, and `40bdae0` closed the one finding I had assembled against `503831f`: the reject message no longer asserts the commit is absent, which it could not know — `gh api` exits non-zero on a 5xx, a rate limit, or a transport failure just as it does on a 404, and `2>/dev/null` discards which. `could not resolve … as a commit` is true under all of them. The narrowing that comes with it — a GitHub API blip twice in ten seconds now aborts in ten seconds where the loop used to absorb it across nine minutes — is fail-safe (exit 2, UNVERIFIED) and not worth retry machinery to recover.

One cross-PR note: #1053 edits the same two comment blocks in this file and will conflict whichever lands second. The substantive half is that #1053 keeps `an unresolvable OID` in `rollup()`'s "Empty output means no usable rollup" list, which this PR removes — correctly, since after this change an unresolvable OID is rejected at entry and never reaches `rollup()`. Merging #1053 second would put that stale cause back.

- `uv run pytest generator/tests/test_poll_scripts.py -q` on `40bdae0`: 25 passed. `shellcheck -S warning` clean.
- The two ways the new entry gate could have locked out real callers, neither reachable from the fake-`gh` suite, both checked against live GitHub from this run:
  - A `refs/pull/N/merge` commit resolves through `GET /repos/max-sixty/tend/commits/{sha}`, so it passes the precheck and still reaches the null-rollup exit-2 path — the header's claim that it "remains the one exit-2 cause head_note can still misfire on" holds.
  - A fork PR's head commit resolves in the base repo through both that endpoint and `repository.object(oid:)`, so polling a fork PR is unaffected.
- Angles checked without a finding: the invariants the deleted sentinel/`absent` machinery enforced are each re-established (one absent sighting is not a verdict → the 10s retry; a rollup already read outranks a later absence → absence is now only ever read at entry; the grace-recheck false-green guard → no sentinel can reach `$R`). The cap report's empty-`pending` contradiction remains pre-existing and out of scope, as the previous round recorded.

</details>

PR #1056, opened 2026-08-25
fix(running-in-ci): attribute the Secrets rule's leak to each harness

max-sixty:
The Secrets bullet's prescription was right; the reason under it named the wrong harness, and tend's own bots read that text as their operating instructions.

#1044 explained `ps aux` by the sandbox agent's launch form. That form is real (`claude/action.yaml:403`), but under `claude` what it exposes is proxy dummies — `proxy/setup-sandbox.sh:289-308` writes `ghp_tendproxydummy…` and `sk-ant-…tendproxydummy…`, and `claude/action.yaml:291` says "No real secrets here." The harness that hands the agent real credentials is `codex`: `codex/action.yaml:190-213` puts `OPENAI_API_KEY`, `GH_TOKEN`, and `GITHUB_TOKEN` in the step env `codex exec` inherits, reachable through `/proc/<pid>/environ`. `docs/security-model.md:322-345` already states the same split. So a bot that checked the stated reason against the tree would have found it false.

The bullet now leads with the property that covers both harnesses — never print a process's environment or command line, or a credential from anywhere else — and names each mechanism where it actually applies.

Three further corrections came out of reviewing the rewrite, two of them from reading the rule the way a bot would, with no other context:

- **The argv is on `sudo`, not the agent.** `env` execs `bash`, which execs the wrapper, so the agent's own `/proc/self/cmdline` is `bash <wrapper>`. The `K=V` list survives on the supervising `sudo`, which stays resident for the run and whose `/proc/<pid>/cmdline` is world-readable.
- **`codex` never strips the checkout credential.** The strip lives only in `proxy/setup-sandbox.sh:362-387`, and `shared/steps/restore-sensitive-config.sh:21` confirms it's Claude-only. So under `codex` the PAT is on disk for the whole job — a path the old rule didn't cover, while its codex sentence ("anything that reads an environment") read as an all-clear for git config.
- **Naming bare `set` and `export` put `set -euo pipefail` in doubt.** The list gave `set` and `env` permitted counterparts and `export` none. The reading-versus-printing line now covers all three.

The auto-reviewer caught a fourth: "`env | cut -d= -f1` still runs `env`" didn't discriminate, since `env FOO=bar cmd` is permitted two sentences up and runs it too. The filter's real problem is that its output isn't value-free — continuation lines of a multi-line value carry no `=`, so `cut -f1` passes them through. Reproduced in an `ubuntu:24.04` container: with `MULTI` set to `aaa\nsk-ant-SECRETVALUE`, `env | cut -d= -f1` prints `sk-ant-SECRETVALUE`.

An earlier draft of mine said "the repo's config doesn't enter into it" while naming `sandbox_env:` as what leaks two sentences later. That's gone — the artifact-upload sentence already does that work.

<details>
<summary>What I verified, and two things I declined to write</summary>

Verified in an `ubuntu:24.04` container: `ps -eo args` from any user prints the launch env off the sudo process while the agent's own cmdline carries none of it; `ps -eo pid,etime,comm` and `pgrep -f` print no argv, while `ps aux` and `pgrep -af` do.

`ps -o pid,etime,comm` became `ps -eo …` because without a selection option procps filters on the invoker's terminal, which differs between the harnesses.

Declined: `GIT_TRACE_CURL=1` was suggested as a leak, but git has redacted `Authorization:` by default since 2.22 (confirmed against `git help` on 2.55). And `git config --list` doesn't print the token under modern `actions/checkout` — it prints the path to the file holding it — so the text names the credential rather than a command that supposedly dumps it.

</details>

Not addressed here, since each is a policy call rather than a correction: nothing covers a secret the bot didn't type (a test harness dumping env on failure, `set -x` in an invoked script); `docs/tend.example.yaml:194` documents `sandbox_env:` without noting those values land in the sudo argv; and `docs/security-model.md` covers `/proc/<pid>/environ` being unreadable across UIDs but never mentions the argv channel.


tend-agent (commented):
The per-harness attribution checks out against the tree. `claude/action.yaml`'s launch (`sudo -u "$SANDBOX" env "${AGENT_ENV[@]}" … bash "$WRAPPER"`) does leave the `K=V` list on `sudo`'s argv rather than the agent's own cmdline; `proxy/setup-sandbox.sh` is the only place the persisted checkout credential is stripped and `codex/action.yaml` never runs it, so the codex sentence holds; and `sandbox_env` is validated in `config.py` as claude-only and lands in that same launch env, so "any adopter-supplied value alike" holds too. `ps -eo` over `ps -o` is right — without a selection option procps filters on the invoker's terminal.

One sentence reads inconsistently to a bot applying the rule literally, which is the audience the rewrite is aimed at. Inline.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -90,7 +90,7 @@ If a linked PR merged (or the triggering PR itself merged) **after the triggerin
...
...tput doesn't carry the value: `pgrep -f pytest` is allowed but `pgrep -af pytest` is not, and `set -euo pipefail`, `export FOO=bar`, and `env FOO=bar cmd` are fine where bare `set`, `export`, and `env` are not. Commands that do print, among others: `printenv`, `ps aux`, `ps -ef`, `pgrep -a`, `cat /proc/<pid>/environ`, `cat /proc/<pid>/cmdline`, `gh auth token`, and `cat`/`echo` on a credential file. Filtering buys no exception; `env | cut -d= -f1` still runs `env`. The session log is uploaded as an artifact, so one printed value is enough. Under `codex` the agent runs as the runner user with the real PAT and model key in its own environment, and the credential `actions/checkout` persisted stays live for the whole job, in the workspace's git config and the file it points at. Under `claude` the proxy holds both and the agent gets dummies, but the supervising `sudo` keeps the whole launch env in its argv, which `ps` shows to anyone on the runner — dummies and any adopter-supplied value alike. Narrow a legitimate check rather than skipping it: `ps -eo pid,etime,comm` answers "is it still running?" with no argv in the output. Never include tokens or credentials in responses or comments.
```
tend-agent: `env FOO=bar cmd` is permitted two sentences up, and it runs `env` as well — so "still runs `env`" doesn't separate the permitted form from the forbidden one. A bot resolving that against the stated property ("output doesn't carry the value") lands on `env | cut -d= -f1` being fine after all, which is the bypass this sentence exists to close.

There is a concrete reason the filter isn't safe, which also discriminates: a value containing a newline prints continuation lines with no `=`, so `cut -f1` passes them through verbatim. Confirmed on this runner — with `MULTI` set to `aaa\nsk-ant-SECRETVALUE`, `env | cut -d= -f1` printed `sk-ant-SECRETVALUE`.

```suggestion
- **Secrets**: Never print a process's environment or command line, your own or another process's, and never print a credential from anywhere else. Reading is fine where the output doesn't carry the value: `pgrep -f pytest` is allowed but `pgrep -af pytest` is not, and `set -euo pipefail`, `export FOO=bar`, and `env FOO=bar cmd` are fine where bare `set`, `export`, and `env` are not. Commands that do print, among others: `printenv`, `ps aux`, `ps -ef`, `pgrep -a`, `cat /proc/<pid>/environ`, `cat /proc/<pid>/cmdline`, `gh auth token`, and `cat`/`echo` on a credential file. Filtering buys no exception: a value containing a newline survives `env | cut -d= -f1` on its continuation lines. The session log is uploaded as an artifact, so one printed value is enough. Under `codex` the agent runs as the runner user with the real PAT and model key in its own environment, and the credential `actions/checkout` persisted stays live for the whole job, in the workspace's git config and the file it points at. Under `claude` the proxy holds both and the agent gets dummies, but the supervising `sudo` keeps the whole launch env in its argv, which `ps` shows to anyone on the runner — dummies and any adopter-supplied value alike. Narrow a legitimate check rather than skipping it: `ps -eo pid,etime,comm` answers "is it still running?" with no argv in the output. Never include tokens or credentials in responses or comments.
```

max-sixty: Right on both counts — "still runs `env`" doesn't discriminate, since `env FOO=bar cmd` is permitted two sentences up and runs it too.

Reproduced your case in an `ubuntu:24.04` container before taking the fix: with `MULTI` set to `aaa\nsk-ant-SECRETVALUE`, `env | cut -d= -f1` prints `sk-ant-SECRETVALUE` on its own line.

Took the mechanism but led with the general property it demonstrates, so the sentence doesn't read as an edge case a bot can rule out by checking its own values are single-line: "Filtering buys no exception, because you can't tell the output is value-free without reading the values: continuation lines of a multi-line value carry no `=`, so `env | cut -d= -f1` prints them verbatim."

`gh auth token` was already in the printing list, so that half of the suggestion was a no-op. Pushed as fd9fa2ba.

tend-agent (approved):

PR #1058, opened 2026-08-26
fix: name UTF-8 on every file read and write

tend-agent:
Every file the generator reads or writes goes through the locale's default encoding, so on a machine whose preferred encoding is not UTF-8 `uvx tend@latest init` writes workflow files that are not valid UTF-8. This names `encoding="utf-8"` at all seven text call sites in the generator, plus the one in `install-tend`'s `oauth_token.py`, which runs on the adopter's machine too.

The trigger is the em dash in every generated workflow's header comment (`# Do not edit this file directly — it will be overwritten on regeneration.`). It comes from a Python string literal, so it reaches `Path.write_text` as U+2014; without `encoding=`, `write_text` encodes it with `locale.getpreferredencoding(False)`. On Windows that is the ANSI code page (cp1252) through Python 3.14 — PEP 538's C-locale coercion is POSIX-only and PEP 540's UTF-8 mode is off there — and cp1252 encodes U+2014 as the single byte `0x97`. The write succeeds, so nothing surfaces it: `tend check` doesn't read the workflow files, and the corruption lands in the adopter's repo for GitHub Actions to parse as UTF-8.

Linux is unaffected in practice — I checked rather than assumed, and plain `LC_ALL=C` on Python 3.11+ coerces to UTF-8 via PEP 538, so it still writes correct bytes.

<details><summary>Reproduction and verification</summary>

Rendering this repo's own config with every unencoded `open()` forced to cp1252, then writing as `cli.py` does, corrupts all eight files at the same offset — the header em dash:

```
tend-review.yaml:
  rendered string identical to UTF-8 run: True
  bytes on disk are valid UTF-8: False
    -> 'utf-8' codec can't decode byte 0x97 in position 101: invalid start byte
[... 7 more ...]

8/8 workflow files corrupt when generated under a cp1252 locale
```

"rendered string identical: True" is worth noting: reading the templates under cp1252 happens to round-trip for these particular bytes, so the corruption is entirely in the *write*. The reads are fixed too because that round-trip is accidental — cp1252 leaves `0x81`, `0x8D`, `0x8F`, `0x90`, `0x9D` undefined, so a template character whose UTF-8 encoding contains one of those bytes raises `UnicodeDecodeError` instead.

With the fix, the same forced-cp1252 run writes all 8 files as valid UTF-8 with the em dash intact as `e2 80 94`.

Scope: the three `.sh` scripts read via `read_text()` (`review-gate.sh`, `mention-verify.sh`, `notifications-check.sh`) carry 6–17 non-ASCII lines each. The `.j2` templates were never at risk — Jinja's `PackageLoader` already defaults to `utf-8`. `proxy/` is left alone: its only text reads are in tests that run on the runner under UTF-8.

</details>

<details><summary>No regression guard here — why, and the recipe if one is wanted</summary>

An earlier revision carried `generator/tests/test_encoding.py`, a 118-line AST walker over the shipped package. It's dropped, and this PR is the source fix alone: 118 lines of walker plus a self-test for the walker is the mechanism CLAUDE.md's "prefer deleting a mechanism over refining it" points at, and the source change stands on its own.

**Correction to an earlier version of this section**, which claimed ruff's `PLW1514` (`unspecified-encoding`) covers the same ground. It doesn't, on this codebase. Measured by reverting all seven `generator/src/tend` fixes and running the pinned rev:

```
$ uvx ruff@0.14.11 check --preview --select PLW1514 --output-format concise generator/src/tend/
generator/src/tend/migrate.py:78:5: PLW1514 `pathlib.Path(...).write_text` without explicit `encoding` argument
Found 1 error.
```

One of seven. `PLW1514` resolves the receiver's type, so it fires only where the receiver is inferrably a `pathlib.Path`. It misses the three `workflows.py` template reads — `(importlib.resources.files("tend") / "templates" / "review-gate.sh").read_text()` returns a `Traversable`, not a `Path` — and `config.py`'s `path.open()`, where the `Path | None` annotation defeats the narrowing, and both `cli.py` sites. The earlier "reproduces the walker's findings case for case" claim held only against the walker's own synthetic sample, which used bare `open()` and `Path("literal")` forms.

So the config below is **not** an equivalent guard — it is worth having for the shapes it does catch, but a maintainer weighing it should know the number is one call shape out of the several this repo uses. It's also preview-gated (`ruff check --select PLW1514` warns `has no effect because preview is not enabled`), so it means `preview = true` repo-wide:

```toml
[tool.ruff]
preview = true

[tool.ruff.lint]
extend-select = ["PLW1514"]
```

With `oauth_token.py` fixed above, the 53 remaining hits on this branch are all in `generator/tests`; `ruff check --fix --unsafe-fixes` writes `encoding="utf-8"` into them, or `per-file-ignores` scopes the rule to non-test code.

</details>


tend-agent (commented):
The source change is right and the diagnosis holds up — I checked the premise rather than taking the body's word for it: `workflows.py`'s header constant does carry a literal U+2014, and the three `.sh` templates read via `read_text()` carry 6, 11, and 17 non-ASCII lines. All seven call sites in `generator/src/tend` are covered, and `uv run pytest` is 595 passed.

The concern is `generator/tests/test_encoding.py`. Ruff already checks exactly this, as `PLW1514` (`unspecified-encoding`), and this repo already runs ruff in pre-commit and in CI's `lint` job — so 118 lines of hand-rolled AST walker plus a self-test for the walker is a mechanism we'd maintain forever in place of a config entry. CLAUDE.md's "Simplicity outranks efficiency … Prefer deleting a mechanism over refining it" points the same way.

I ran the comparison rather than assuming it. Pointing `ruff 0.14.11` (the pinned rev) at the exact sample in `test_the_check_would_catch_a_regression` produces the same five findings, in the same order, and correctly passes `open("ok4", "rb")`, `open("ok5", mode="wb")`, `Path("ok6").open("rb")` and `Path("ok7").open(mode="wb")` — including the positional-slot asymmetry between `Path.open` and builtin `open` that `_mode_arg`'s docstring calls out as the part that was easy to get wrong. Ruff gets there by resolving the qualified name instead of matching a bare attribute, which is also why it wouldn't misread a same-named method on another type: `_mode_arg` reads argument 0 of any attribute call named `open`, so `zf.open("data/blob.txt")` would be classified by whether the *filename* happens to contain a `b`, and a non-string literal in that slot raises `TypeError` out of `"b" in mode` rather than reporting anything.

Config, verified to apply to `generator/src/tend/*.py` even though `generator/pyproject.toml` sits in between (ruff skips a `pyproject.toml` with no `[tool.ruff]` and keeps walking up):

```toml
[tool.ruff]
preview = true

[tool.ruff.lint]
extend-select = ["PLW1514"]
```

Two costs worth naming, both measured on this checkout:

- `PLW1514` is preview-gated, so this turns on `preview = true` repo-wide. On this repo today that's inert — `ruff check --preview` with no extra rules selected reports `All checks passed!`, identical to the current default run — but preview rules are unstable by definition, so it's a taste call rather than a free one.
- It widens coverage past the package: 54 hits, of which 53 are in `generator/tests` and one is `plugins/install-tend/skills/install-tend/scripts/oauth_token.py`. `ruff check --fix --unsafe-fixes` writes `encoding="utf-8"` into all of them correctly, or `[tool.ruff.lint.per-file-ignores]` scopes it back to the package. That breadth is the upside, though: the current guard walks `generator/src/tend` only, so a `write_text` added anywhere else — including that `install-tend` script, which like `init` runs on the adopter's machine — is outside it.

Happy to push the swap (config in, `test_encoding.py` out, the 53 test sites autofixed) if you'd rather have it applied than described. Not pushing unprompted since flipping ruff to preview is a repo-wide call, and since keeping the bespoke walker is defensible if you'd rather not depend on a preview rule — in which case the source change alone is good to go.

tend-agent:
Took the swap, minus the ruff config. `generator/tests/test_encoding.py` is gone in c10b565 — a hand-rolled duplicate of `PLW1514` is exactly the mechanism CLAUDE.md's "prefer deleting a mechanism over refining it" points at, and dropping it needs nobody's permission since it was this PR's own addition.

I left `preview = true` off. That one is genuinely a repo-wide taste call, and it's not mine to make on the strength of my own review — so the recipe is folded into the PR body where a maintainer can act on it, rather than applied here. The trade is stated plainly: the PR now ships with no regression guard at all, which the review named as an acceptable end state ("the source change alone is good to go").

I did take the one concrete gap the review surfaced: `plugins/install-tend/skills/install-tend/scripts/oauth_token.py:186` now names the encoding. It reads an OAuth code file on the adopter's machine, so it's the same class of bug the PR is about, and it was outside the walker's `generator/src/tend` scope. With it fixed, every remaining `PLW1514` hit on this branch is in `generator/tests` — so if the config does go in later, non-test code is already clean and `per-file-ignores` is a viable scoping.

Preview gating confirmed against the pinned rev rather than taken from the review:

```
$ uvx ruff@0.14.11 check --select PLW1514 --statistics generator/src/tend/
warning: Selection `PLW1514` has no effect because preview is not enabled.
```

Hit distribution after this push (`--preview --select PLW1514`, repo root) — 53 hits, all under `generator/tests`, no source-tree remainder:

```
30  generator/tests/test_shared_steps.py
 7  generator/tests/test_list_recent_runs.py
 6  generator/tests/test_poll_scripts.py
 4  generator/tests/test_mention_verify.py
 3  generator/tests/test_review_runs_corrections.py
 3  generator/tests/test_bot_review_state.py
```

`uv run pytest` is 593 passed (595 before; the two dropped are the walker's own self-tests). `uvx pre-commit run --all-files` passes every hook.

</details>

tend-agent (commented):
The eight-line source fix is right and complete — I re-checked rather than carrying the last review's verdict forward. Every text-mode read/write in `generator/src`, `plugins/`, and `proxy/` now names an encoding except `proxy/test_inject_credentials.py`'s two `read_text()` calls, which the body scopes out correctly (they only run on the runner). `uv run pytest` is 593 passed, and CI's `lint`, `test`, `test-proxy`, `test-sandbox`, `test-codex-surface`, `test-worker` are all green on `c10b565`.

One correction, and it's to my own prior review rather than to this push: **`PLW1514` does not substitute for the walker on this codebase.** I measured it this time instead of generalizing from the synthetic sample. Reverting all seven `generator/src/tend` fixes and running the pinned rev finds **one** of them:

```
$ uvx ruff@0.14.11 check --preview --select PLW1514 --output-format concise generator/src/tend/
generator/src/tend/migrate.py:78:5: PLW1514 `pathlib.Path(...).write_text` without explicit `encoding` argument
Found 1 error.
```

`PLW1514` resolves the receiver's type, so it only fires where the receiver is inferrably a `pathlib.Path`. It misses the three `workflows.py` template reads — `(importlib.resources.files("tend") / "templates" / "review-gate.sh").read_text()` returns a `Traversable`, not a `Path` — and it misses `config.py`'s `path.open()`, where the `Path | None` parameter annotation defeats the narrowing, and both `cli.py` sites. Isolated probe confirming the shape rather than the count:

```
probe.py:6:10:  PLW1514  path.open()                     # param annotated `path: Path`     -> flagged
                          (files("tend") / "x.sh").read_text()                              -> not flagged
                          files("tend").joinpath("x.sh").read_text()                        -> not flagged
```

My earlier claim that ruff "reproduces the walker's findings case for case" held only for the sample in `test_the_check_would_catch_a_regression`, which used bare `open()` and `Path("literal")` forms. It does not hold for the shapes this PR actually fixes, and the walker caught all seven because it matched on call name rather than receiver type.

That doesn't argue for restoring the walker — 118 lines of AST walker plus a self-test for the walker is still the shape CLAUDE.md's "prefer deleting a mechanism over refining it" points at, and the source fix stands on its own. What it does change is the PR body: the "No regression guard here" block currently tells a future maintainer that `preview = true` plus `extend-select = ["PLW1514"]` is the guard, and acting on that recipe would buy coverage of one call shape out of the several this repo uses. Worth correcting there so the decision gets made on the real number — I'll amend the body to say so, since the wrong figure came from my review.

Not approving because I authored this one; the diff itself I'd have approved.

PR #1059, opened 2026-08-26
feat(sandbox): give sandbox_setup the same GitHub context as the agent

max-sixty:
`sandbox_setup:` runs the adopter's own commands, as the sandbox user, in the agent's launch env, immediately before the agent starts. The agent has always received the runner's `GITHUB_*` context across that UID boundary minus a six-name denylist; `sandbox_setup:` received one name of it, because #1057 named the one name its case needed and said extraction was the fix when a second case asked. That left an arbitrary one-item allowlist beside a crossing otherwise governed by a rule, so the step that prepares a session could not answer questions the session can.

`shared/steps/lib/sandbox-launch-env.sh` now holds `sandbox_launch_env <agent-env-file>`, and both crossings that carry adopter-written code source it. Which side of that line a crossing falls on is who wrote the code on the far side: `install-tend-plugins.sh` runs tend's own code from a marketplace already copied to local disk and takes no context.

The function composes the whole launch env rather than handing back the context alone, because the order is the load-bearing part. `env` takes the final assignment of a name and the env file is the half an adopter writes (`sandbox_env:`), so the context has to follow it or a `sandbox_env: {GITHUB_WORKFLOW: …}` decides what the run thinks it is. As two arrays that was a rule both call sites had to remember; here it is the function's postcondition, and a unit test holds it. The inverse is forced the same way: tend's `TEND_*`/`BOT_*`/`CI` assignments must beat the env file, since `TEND_*` is not in `RESERVED_SANDBOX_ENV` and `sandbox_env: {TEND_PROMPT: …}` would otherwise set the agent's prompt.

Two things the denylist's comment claimed turned out to need saying more carefully, and this file is now their only home. `GITHUB_TOKEN` is excluded but is not *withheld* from an adopter's point of view — `proxy/inject_credentials.py` rewrites `Authorization` to the real PAT for GitHub hosts, so a `sandbox_setup:` command's `gh` authenticates. And "all are public values, not secrets" holds for what Actions defines, not for a `GITHUB_*`-named variable an adopter's `setup:` step writes to `$GITHUB_ENV`, which crosses on the same rule — so a secret must not be given a `GITHUB_*` name. Both are stated in the lib and in `tend.example.yaml`.

## Motivating case

A `sandbox_setup:` block scoping work to a scheduled run. `$GITHUB_EVENT_NAME` never crossed, so the test compared against an empty string, the `if` silently took the else branch, and the failure surfaced later as a missing tool with an unrelated-looking error:

```yaml
sandbox_setup:
  - 'if [ "$GITHUB_WORKFLOW" = tend-weekly ]; then nix --version; fi'    # worked
  - 'if [ "$GITHUB_EVENT_NAME" = schedule ]; then ./warm-cache.sh; fi'   # never fired
```

Both work now, along with `$GITHUB_REPOSITORY`, `$GITHUB_REF_NAME`, `$GITHUB_BASE_REF` and the rest.

## Invariants that were stated and unasserted

Three, of which two were reachable by a wrong edit and one by a wrong deletion.

**The denylist's names.** `proxy/test-setup-sandbox.sh` represented all five withheld paths with `GITHUB_ENV` alone, so dropping `GITHUB_STEP_SUMMARY` from the `case` passed everything. So did dropping the `_` from `^GITHUB_`, which is the mutation with a direction: it would carry an adopter's `GITHUBBER_TOKEN` across the boundary. Five unit tests pin both sides.

**The composition order**, at unit level and again over the single backslash-continued `sudo -u "$SANDBOX" env` command at each crossing — scoped to the command rather than to file order, since "later in this argv" is what decides who wins. That second test also asserts the array is both filled and splatted, either of which nothing caught before: bash expands `"${arr[@]}"` on an unset array to nothing and exits 0, `set -u` included, so losing the composing call is not an error the shell reports — the crossing would launch with no PATH, no proxy routing, no CA trust and no credentials.

**Every `${{ github.action_path }}/…` reference in both composite actions.** Nothing checked these at all: the pre-commit `actionlint` hook is pinned to `^\.github/workflows/`, so neither `action.yaml` is linted, and no workflow consumes them with `uses: ./`, so an edited body first executes in an adopter's job after a release.

<details>
<summary>Verification beyond the suite</summary>

A 12-mutation matrix, all caught, baseline green: dropping `GITHUB_STEP_SUMMARY`; dropping `GITHUB_TOKEN`; `grep 'GITHUB_'` unanchored; `grep '^GITHUB'` with the `_` dropped; the denylist inverted to an allowlist; unquoting the array append; mistyping the lib path in `claude/action.yaml`; composing the file after the context; deleting the splat at either crossing; appending a `GITHUB_*` assignment at either.

A vacuity check with the test helper's `PATH` pointed at `/nonexistent`, where all five unit tests fail — none passes when the code under test never ran. An earlier version of one of them did: its `== []` was satisfied by total failure, and it survived both review and the mutation matrix, because both assume the code executed.

An end-to-end rehearsal of the real `shared/steps/sandbox-setup.sh` under a `sudo` stand-in that strips `-u <user>` then `exec env -i "$@"`, emulating `env_reset`: the full context reaches the sandbox, `GITHUB_ENV`/`GITHUB_STEP_SUMMARY` are withheld, the env file's dummy token survives a distinct runner-side `GITHUB_TOKEN`, a `sandbox_env` spoof of `GITHUB_WORKFLOW` loses to the real value, a non-matching workflow gate skips at exit 0, and a matching gate whose tool is missing exits 1.

</details>

## Gaps

`proxy/test-setup-sandbox.sh` needs a Linux runner with a second uid and sudo, so it has never executed against these changes — the `test-sandbox` job here is its first real run. And nothing anywhere executes `claude/action.yaml`'s Run Claude block: this repo's CI does not, and the two tests covering it are static assertions over its text. That gap is the reason both of those tests exist, and retiring them would mean a smoke job that runs the composite end to end — worth doing, larger than this.

Which crossings take the context is still convention. The composition order is a postcondition; a new crossing has nothing forcing it to call this.

### DIFF
--- a/generator/tests/test_repo_pins.py
+++ b/generator/tests/test_repo_pins.py
@@ -72,3 +72,50 @@ def test_action_path_references_resolve(harness: str) -> None:
         if not (REPO_ROOT / harness / ref).resolve().exists()
     ]
     assert not missing, f"{harness}/action.yaml references nothing at: {missing}"
+
+
+# The `sudo -u "$SANDBOX" env` line each adopter-facing crossing builds. Nothing
+# else covers claude/action.yaml's: its inline block is unlinted (the actionlint
+# hook is pinned to ^.github/workflows/), no workflow consumes the action with
+# `uses: ./`, and the test-sandbox job drives shared/steps/sandbox-setup.sh
+# directly without going through the action. Deleting the splat there launches
+# the agent with no PATH, no proxy routing, no CA trust and no credentials.
+CROSSINGS = ("claude/action.yaml", "shared/steps/sandbox-setup.sh")
+GITHUB_ASSIGNMENT = re.compile(r"\bGITHUB_[A-Z_]*=")
+
+
+def _sudo_env_command(body: str, path: str) -> str:
+    """The one `sudo -u "$SANDBOX" env …` command, continuations included."""
+    lines = body.splitlines()
+    starts = [i for i, line in enumerate(lines) if 'sudo -u "$SANDBOX" env' in line]
+    assert len(starts) == 1, f"{path}: expected one sudo env crossing, got {starts}"
+    i = starts[0]
+    command = [lines[i]]
+    while lines[i].rstrip().endswith("\\"):
+        i += 1
+        command.append(lines[i])
+    return "\n".join(command)
+
+
+@pytest.mark.parametrize("crossing", CROSSINGS)
+def test_the_crossing_launches_from_the_composed_env(crossing: str) -> None:
+    """The composed array is on the line, and nothing GITHUB_* follows it.
+
+    sandbox_launch_env puts the context after the agent env file, so an
+    adopter's `sandbox_env:` cannot decide what the run thinks it is. A caller
+    is free to append names of its own — tend's BOT_*/TEND_* assignments do,
+    and have to, since they must beat the file — but a GITHUB_*-named one would
+    land after the context and displace it. Scoped to the single command rather
+    than to file order, so it says "later in this argv", which is the thing that
+    decides who wins.
+    """
+    command = _sudo_env_command((REPO_ROOT / crossing).read_text(), crossing)
+
+    assert '"${SANDBOX_LAUNCH_ENV[@]}"' in command, (
+        f"{crossing}: the crossing does not carry the composed launch env"
+    )
+    trailing = command.split('"${SANDBOX_LAUNCH_ENV[@]}"', 1)[1]
+    assert not GITHUB_ASSIGNMENT.search(trailing), (
+        f"{crossing}: a GITHUB_* assignment follows the composed array, so it "
+        f"displaces the real context: {trailing.strip()}"
+    )
--- a/generator/tests/test_repo_pins.py
+++ b/generator/tests/test_repo_pins.py
@@ -99,7 +99,7 @@ def _sudo_env_command(body: str, path: str) -> str:
 
 @pytest.mark.parametrize("crossing", CROSSINGS)
 def test_the_crossing_launches_from_the_composed_env(crossing: str) -> None:
-    """The composed array is on the line, and nothing GITHUB_* follows it.
+    """Something fills the array, it is on the line, and nothing GITHUB_* follows.
 
     sandbox_launch_env puts the context after the agent env file, so an
     adopter's `sandbox_env:` cannot decide what the run thinks it is. A caller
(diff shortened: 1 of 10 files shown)


generator/tests/test_repo_pins.py
```diff
@@ -47,3 +49,73 @@ def test_uv_build_range_admits_the_pinned_uv() -> None:
...
+    and have to, since they must beat the file — but a GITHUB_*-named one would
+    land after the context and displace it. Scoped to the single command rather
+    than to file order, so it says "later in this argv", which is the thing that
+    decides who wins.
+    """
+    command = _sudo_env_command((REPO_ROOT / crossing).read_text(), crossing)
+
+    assert '"${SANDBOX_LAUNCH_ENV[@]}"' in command, (
+        f"{crossing}: the crossing does not carry the composed launch env"
+    )
```
```suggestion
    body = (REPO_ROOT / crossing).read_text()
    command = _sudo_env_command(body, crossing)

    assert 'sandbox_launch_env "$AGENT_ENV_FILE"' in body, (
        f"{crossing}: nothing composes the launch env, and bash expands an "
        f"unset array to nothing, so the crossing would launch with an empty one"
    )
    assert '"${SANDBOX_LAUNCH_ENV[@]}"' in command, (
        f"{crossing}: the crossing does not carry the composed launch env"
    )
```

tend-agent (commented):
One gap in the new static crossing test. The rest reads correct to me — the composition order, the prefix anchoring, and the six withheld names all check out, and `proxy/test-setup-sandbox.sh`'s pre-existing `grep -qF "sandbox_setup workflow: $GITHUB_WORKFLOW"` now doubles as the end-to-end assertion that the real name beats the `sandbox_env` spoof you added to `setup()`.

`test_the_crossing_launches_from_the_composed_env` pins the splat and what may follow it, but not the call that fills the array. Bash 4.4+ expands `"${arr[@]}"` on an *unset* array to nothing without erroring, even under `set -u` — so deleting `sandbox_launch_env "$AGENT_ENV_FILE"` leaves the test green while the crossing launches with an entirely empty environment. That is the same failure the docstring's splat case names (no PATH, no proxy routing, no CA trust, no credentials), and for `claude/action.yaml` nothing here executes that block, so it would first surface in an adopter's job after a release.

Bash behaviour, on the runner's 5.2.21:

```
$ bash -c 'set -euo pipefail; printf "[%s]" "${NOPE[@]}"; echo " rc=$?"'
[] rc=0
```

Mutation, deleting the composing line from `shared/steps/sandbox-setup.sh`:

- with the PR's test as written: `2 passed`
- with the suggested assertion: `1 failed, 1 passed`

</details>

generator/tests/test_repo_pins.py
max-sixty: Confirmed and fixed in b6d91351. I reproduced both halves before taking it: `set -euo pipefail; printf "[%s]" "${NOPE[@]}"` exits 0 with no output, and deleting the composing line left `2 passed`. With the assertion added, deleting it from either crossing gives `1 failed, 1 passed`.

I took the assertion rather than the whole suggestion, since the `body`/`command` split it needed was already there. The docstring now carries the reason, because "an unset array splats to nothing even under `set -u`" is the part a reader would otherwise assume the shell catches.

Good catch — this is the second time on this branch that a test passed because the code under test never ran, which is a failure mode neither review nor a mutation matrix sees, since both assume execution.

tend-agent (approved):

PR #1062, opened 2026-08-26
skills(running-in-ci): route a generic correction upstream, not to the consumer overlay

### LINKED ISSUE #1061: Learning from Feedback routes a generic correction to the consumer overlay instead of upstream tend
## Problem

A maintainer corrected a tend-authored PR title in a consumer repo. The correction was generic — every consumer wants plain titles — and the rule already existed upstream (#1022, sharpened this morning by #1060). The consumer session nonetheless wrote 22 lines of title guidance into that repo's `running-tend` overlay ([max-sixty/leaf#45](https://github.com/max-sixty/leaf/pull/45)), and the maintainer's reply was "this is an upstream issue; why are you making a local change".

The routing rule exists, but two things in the current text send a generic correction the other way.

**1. The default route is the overlay.** `running-in-ci`'s **Learning from Feedback** opens with the destination: "propose a follow-up PR against the consuming repo's `.claude/skills/running-tend/SKILL.md`". The upstream alternative is two levels down — in `references/skill-pr-workflow.md`, in a "Don't propose when…" clause and a closing paragraph. A session that has already decided the feedback clears the bar reads the destination in the first sentence and never re-opens the question.

**2. The upstream signal is conditioned on bundled text existing.** `skill-pr-workflow.md` says the fix belongs in tend when "the fix reads as generic guidance any consumer would want, **and** the corrected behavior comes from bundled skill text." That conjunction fails exactly when a bundled skill is *silent* — which is the commonest reason a maintainer has to correct generic behavior at all. In leaf#45 the session grepped the skills it had loaded, found no title rule, read the second signal as unmet, and routed local. The gap was real; the destination was wrong.

**3. A release lag reads as license to fork.** The session knew #1022 was merged upstream. It reasoned in the PR body that because leaf pins `max-sixty/tend/claude@0.1.18` (2026-08-16) and #1022 merged 2026-08-20, "the overlay is the only thing enforcing this at the current pin, and it should land on its own terms." The lag is temporary; overlay text is permanent. Once the release ships, every consumer that took this route carries a duplicate rule that someone has to notice and delete — and until then the two copies drift.

## Proposed fix

Docs only, in the two files above.

- **`running-in-ci` → Learning from Feedback**: lead with the routing question rather than the destination. Something like: *Ask first whether every tend consumer would want this rule. If so it belongs in tend — open an issue here (or a PR if you maintain it). The overlay is for facts true only of this repo: its branch and landing conventions, its test topology, its trackers.*
- **`skill-pr-workflow.md` → Whether to propose**: split the `and`. A bundled skill's **silence** on generic behavior is the same gap as wrong bundled text — arguably a larger one, since nothing is enforcing the rule anywhere.
- **`skill-pr-workflow.md`**: one line stating that an unreleased upstream fix is not a reason to write a local copy. If the rule is already merged in tend and waiting on a release, the fix is a release, not an overlay section.

## Why it matters here

The overlay is the wrong home for a rule this shape twice over: it fixes one consumer while the other consumers keep writing the same bad titles, and it survives the release that makes it redundant. The routing text is what decides which of those a session does, and right now it points at the overlay by default.

tend-agent:
## Problem

**Learning from Feedback** in `running-in-ci` opens by naming the destination — "propose a follow-up PR against the consuming repo's `.claude/skills/running-tend/SKILL.md`" — so a session that has already decided the correction clears the bar never re-opens the question of where it belongs. The upstream alternative sat two levels down in `references/skill-pr-workflow.md`, and it was conditioned on bundled text existing: "the fix reads as generic guidance any consumer would want, **and** the corrected behavior comes from bundled skill text." That conjunction fails exactly when a bundled skill is silent, which is the commonest reason a maintainer has to correct generic behavior at all. Nothing addressed the third case either: an upstream fix that has merged but not yet reached the consumer's pinned release, which reads as license to write a local copy that outlives the lag.

## Solution

Docs only, in the two files.

- `running-in-ci` → **Learning from Feedback** now leads with the routing question ("would every tend consumer want this rule?") rather than the destination, states that bundled silence and an unreleased upstream fix are both gaps upstream, and describes the overlay by what actually belongs in it — branch and landing conventions, test topology, trackers and labels, standing exceptions.
- `skill-pr-workflow.md` gains a **Where it lands** section that splits the old conjunction into three cases (bundled text wrong, bundled text silent, rule merged but unreleased) and names the upstream path per repo: an issue on tend from a consumer, a PR when running on tend itself. The stale routing clause in **Whether to propose** is dropped so the bar and the destination stop competing, and **Mechanics** is marked as the overlay path.

## Testing

No behavioral code changed; skill prose has no test to fail. `uvx pre-commit run --files <the two files>` passes (typos, trailing whitespace, the plugin-skill backtick check, and the install-tend reference sync).

---
Closes #1061 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -497,7 +497,9 @@ Read logs, code, and API data before drawing conclusions. Show evidence: cite lo
 
 ## Learning from Feedback
 
-When a maintainer corrects the bot's behavior during a run — a repo convention, a repeated mistake, a preference the bot should have known — propose a follow-up PR against the consuming repo's `.claude/skills/running-tend/SKILL.md`, turning a one-off correction into durable guidance for future runs in *this* repo. Follow `references/skill-pr-workflow.md`: it carries the bar the feedback has to clear, when the fix belongs in tend instead, and the branch/PR mechanics. Open the PR and exit — don't merge, don't wait, don't ping for review.
+When a maintainer corrects the bot's behavior during a run — a repo convention, a repeated mistake, a preference the bot should have known — turn the correction into durable guidance. Route it before you write it: **would every tend consumer want this rule?** If yes it belongs in tend's bundled skills — a PR when this repo *is* tend, otherwise an issue there per **Other Repos**. A bundled skill that is silent on the behavior is a gap upstream, not a signal the rule is local; nor is an upstream fix that has merged but not yet reached the pinned release. The overlay at `.claude/skills/running-tend/SKILL.md` is for what is true of this repo alone: its branch and landing conventions, its test topology, its trackers and labels, its standing exceptions.
+
+Follow `references/skill-pr-workflow.md`: it carries the bar the feedback has to clear, the routing signals in full, and the branch/PR mechanics. Open the PR and exit — don't merge, don't wait, don't ping for review.
 
 ## Tone
 
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -499,7 +499,7 @@ Read logs, code, and API data before drawing conclusions. Show evidence: cite lo
 
 When a maintainer corrects the bot's behavior during a run — a repo convention, a repeated mistake, a preference the bot should have known — turn the correction into durable guidance. Route it before you write it: **would every tend consumer want this rule?** If yes it belongs in tend's bundled skills — a PR when this repo *is* tend, otherwise an issue there per **Other Repos**. A bundled skill that is silent on the behavior is a gap upstream, not a signal the rule is local; nor is an upstream fix that has merged but not yet reached the pinned release. The overlay at `.claude/skills/running-tend/SKILL.md` is for what is true of this repo alone: its branch and landing conventions, its test topology, its trackers and labels, its standing exceptions.
 
-Follow `references/skill-pr-workflow.md`: it carries the bar the feedback has to clear, the routing signals in full, and the branch/PR mechanics. Open the PR and exit — don't merge, don't wait, don't ping for review.
+Follow `references/skill-pr-workflow.md`: it carries the bar the feedback has to clear, the routing signals in full, and the branch/PR mechanics. Open it — PR or issue — and exit: don't merge, don't wait, don't ping for review.
 
 ## Tone
 
--- a/plugins/tend-ci-runner/skills/running-in-ci/references/skill-pr-workflow.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/references/skill-pr-workflow.md
@@ -57,7 +57,11 @@ issues** in `other-repos.md`. Running on tend itself, it means a PR here.
 
 ## Mechanics
 
-For the overlay path. An upstream fix follows `other-repos.md` instead.
+For the overlay path, and for an upstream PR on tend itself — with the bundled
+skill file as step 2's dedup target, and without step 3's read-only-mount
+workaround, since bundled skills live under `plugins/` rather than
+`.claude/skills/`. Filing upstream from a consumer repo follows
+`other-repos.md` instead.
 
 1. **Complete the current task first.** The skill update is always a separate
    PR.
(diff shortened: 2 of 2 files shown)


plugins/tend-ci-runner/skills/running-in-ci/references/skill-pr-workflow.md
```diff
@@ -24,21 +24,41 @@ Signals pointing at a generalizable rule: the correction names a pattern
...
+conventions, its test topology, its trackers and labels, its standing
+exceptions. A rule that reads as generic guidance goes upstream even when
+writing it locally would be quicker.
+
+From a consumer repo, upstream means an issue on tend, filed per **Filing
+issues** in `other-repos.md`. Running on tend itself, it means a PR here.
 
 ## Mechanics
 
+For the overlay path. An upstream fix follows `other-repos.md` instead.
```
tend-agent: `other-repos.md` has no section for opening a PR in the repo you're already running on, so the on-tend upstream path lands nowhere. Most of the steps below are what it needs — the exceptions are step 2's dedup path and step 3's read-only-mount dance, both of which are `.claude/skills/`-specific.

```suggestion
For the overlay path, and for an upstream PR on tend itself — with the bundled
skill file as step 2's dedup target, and without step 3's read-only-mount
workaround, since bundled skills live under `plugins/` rather than
`.claude/skills/`. Filing upstream from a consumer repo follows
`other-repos.md` instead.
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -497,7 +497,9 @@ Read logs, code, and API data before drawing conclusions. Show evidence: cite lo
...
+When a maintainer corrects the bot's behavior during a run — a repo convention, a repeated mistake, a preference the bot should have known — turn the correction into durable guidance. Route it before you write it: **would every tend consumer want this rule?** If yes it belongs in tend's bundled skills — a PR when this repo *is* tend, otherwise an issue there per **Other Repos**. A bundled skill that is silent on the behavior is a gap upstream, not a signal the rule is local; nor is an upstream fix that has merged but not yet reached the pinned release. The overlay at `.claude/skills/running-tend/SKILL.md` is for what is true of this repo alone: its branch and landing conventions, its test topology, its trackers and labels, its standing exceptions.
+
+Follow `references/skill-pr-workflow.md`: it carries the bar the feedback has to clear, the routing signals in full, and the branch/PR mechanics. Open the PR and exit — don't merge, don't wait, don't ping for review.
```
tend-agent: The consumer-repo route this paragraph now sends upstream ends in an issue, not a PR.

```suggestion
Follow `references/skill-pr-workflow.md`: it carries the bar the feedback has to clear, the routing signals in full, and the branch/PR mechanics. Open it — PR or issue — and exit: don't merge, don't wait, don't ping for review.
```

tend-agent (commented):
The routing rewrite reads well and the three-case split in **Where it lands** closes the conjunction that sent #1061's session local. Two loose ends where the new upstream route isn't carried all the way through the text:

The new `## Mechanics` preamble disclaims the whole section for anything upstream, but `other-repos.md` only covers acting in *another* repo (filing there, invited contributions, scope blockers) — it has nothing for the case this PR just introduced, a PR on tend itself. So a session on tend that answers the routing question "yes, generic" is pointed at a file that doesn't describe its path, and the steps that do apply to it — separate PR, dedup by target file, base on `origin/<default>` rather than the PR checkout's `HEAD`, open-and-exit — are labelled as not-for-you. That last one is the expensive miss: `tend-mention` runs after `gh pr checkout`, so basing on `HEAD` is exactly the mistake step 3 exists to prevent, and it doesn't stop being a mistake because the skill file lives under `plugins/`.

Second, `Open the PR and exit` in the SKILL.md paragraph predates the routing change, and one of the three routes now produces an issue rather than a PR.

Both are inline.

plugins/tend-ci-runner/skills/running-in-ci/references/skill-pr-workflow.md
tend-agent: Applied in 04e601e — the preamble now claims the on-tend PR path explicitly and names the two deltas (step 2 dedups on the bundled file; step 3's read-only-mount dance doesn't apply under `plugins/`), leaving `other-repos.md` for the consumer-repo filing route it actually covers.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied in 04e601e — now "Open it — PR or issue — and exit", so the sentence covers the consumer-repo route that ends in an issue.

PR #1063, opened 2026-08-26
chore: release 0.1.19

max-sixty:
Releases the 48 commits since 0.1.18. The changelog groups the notable sandbox, install, review, CI recovery, security, and portability changes; the package and lockfile move to 0.1.19. It also corrects the activity worker's stale header comment to match its four-search implementation.

Verified with `wt test`, `pre-commit run --all-files`, the focused repository-pin tests, a package build, and `twine check`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,32 @@ published verbatim as that version's GitHub Release notes
 0.1.1 predate this changelog; see the compare views at
 https://github.com/max-sixty/tend/compare for their history.
 
+## 0.1.19
+
+### Improved
+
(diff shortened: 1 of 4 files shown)


tend-agent (approved):
Checked the changelog section against all 48 commits in `0.1.18..0e6efef` — every bullet's claim holds against the diffs, not just the commit subjects, and every cited PR number maps to a commit in the range. The four pins named under **Internal** match the tree (`claude_version: 2.1.241`, `codex_version: 0.149.0`, `uv_version: 0.12.5`, `astral-sh/setup-uv@v10.0.1`). Header is exactly `## 0.1.19`, so the release workflow's `awk` extraction picks up the section. `uv lock --check` is clean, and the regtest snapshots normalize the action ref to `X.Y.Z`, so the bump causes no snapshot churn. The `tend-*.yaml` and `review-reviewers.yaml` refs still reading `0.1.18` is correct here — they move at step 11, after the tag exists.

One thing outside the diff, non-blocking. The **Fixed** bullet for the activity feed is accurate, but `worker/src/index.ts` never had its file-header comment updated when that change landed: the block ending "…`/currently-tending` skips KV" still says `/activity` is one Search query per bucket per bot and that "The fanout is 4·N concurrent Search requests against the 30/min Search cap". The `BUCKET_QUERIES` comment further down states the opposite — "Each bucket is ONE Search request covering every bot, not one per bot" — so the file contradicts itself, and the stale half is the one a reader hits first. Happy to push a commit dropping that sentence from the header if you want it folded in here rather than left for a later sweep.

tend-agent (approved):

PR #1065, opened 2026-08-27
docs: name shared/steps/ in the lists of suites root pytest collects

tend-agent:
## Problem

Four live files enumerate the suites a root `uv run pytest` collects, and all four still describe the pre-port set — "generator/, proxy/, and the install-tend scripts". The bash→Python port of the composite-action step bodies (`9a330b8`, `374f1b1`) added a fifth directory, `shared/steps/`, whose `test_*.py` files contribute 164 of the 717 tests the root run now collects, about 23% of the Python suite.

Those lists are the repo's own statement of what one `pytest` covers, and a session reading any of them comes away thinking the step bodies — the security preflight, the rate-limit gate, the agent supervisor, the outage tracker — are tested somewhere else, or not at all. `pyproject.toml`'s is the sharpest case: it's the file that configures the collection, and its comment contradicts itself two lines apart, naming `shared/steps/` in the sentence right after the list that omits it.

## Fix

Add `shared/steps/` to all four, and reflow the paragraphs that the added words pushed past their files' wrap width.

- [`CLAUDE.md`](https://github.com/max-sixty/tend/blob/7a7a34c/CLAUDE.md#L32-L38) — the Commands section
- [`pyproject.toml`](https://github.com/max-sixty/tend/blob/7a7a34c/pyproject.toml#L1-L11) — the workspace-root header comment
- [`.github/workflows/ci.yaml`](https://github.com/max-sixty/tend/blob/7a7a34c/.github/workflows/ci.yaml#L8-L11) — the comment above the `test:` job
- [`.config/wt.toml`](https://github.com/max-sixty/tend/blob/7a7a34c/.config/wt.toml#L31-L35) — the comment above `[[aliases.test]]`

`CHANGELOG.md`'s [#977](https://github.com/max-sixty/tend/pull/977) entry carries the same list and is left alone — it records what was true at that release.

## Verification

```
$ uv run pytest --collect-only -q | grep -c '^shared/steps'
164
$ uv run pytest -q
717 passed
```

That run was from the first commit's session; `uv` isn't on the PATH in this repo's agent sandbox, so the follow-up commit's verification is CI's own `test` job. Every change in it is a comment or prose line — no Python, no YAML key, no `wt` step body — and nothing under `generator/tests` or `shared/steps` asserts on these strings (`grep -rn 'install-tend scripts\|Every Python test' --include='*.py'` is empty).

No regression test: the change is a noun in four documentation lists. A test tying those sentences to the set of directories pytest actually collects is the kind of machinery `CLAUDE.md` itself rules out — complexity earns its place by preventing a wrong outward action, and a stale doc sentence is not one.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,7 +30,7 @@ pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, 
 ```
 
 The repo root is a uv workspace, and pytest run from it collects every Python
-test — generator/, proxy/, and the install-tend scripts — so one `pytest` is
+test — generator/, proxy/, shared/steps/, and the install-tend scripts — so one `pytest` is
 the whole Python run. `wt test` (defined in [`.config/wt.toml`](.config/wt.toml))
 adds worker/'s vitest suite and typecheck. Its arguments narrow pytest and
 nothing else, so a filtered run still pays for worker/; `uv run pytest -k render`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,10 +30,11 @@ pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck, 
 ```
 
 The repo root is a uv workspace, and pytest run from it collects every Python
-test — generator/, proxy/, shared/steps/, and the install-tend scripts — so one `pytest` is
-the whole Python run. `wt test` (defined in [`.config/wt.toml`](.config/wt.toml))
-adds worker/'s vitest suite and typecheck. Its arguments narrow pytest and
-nothing else, so a filtered run still pays for worker/; `uv run pytest -k render`
+test — generator/, proxy/, shared/steps/, and the install-tend scripts — so
+one `pytest` is the whole Python run. `wt test` (defined in
+[`.config/wt.toml`](.config/wt.toml)) adds worker/'s vitest suite and
+typecheck. Its arguments narrow pytest and nothing else, so a filtered run
+still pays for worker/; `uv run pytest -k render`
 is the Python half on its own.
 
 ## Architecture
--- a/.config/wt.toml
+++ b/.config/wt.toml
@@ -29,10 +29,10 @@ server = "wt step tether -C site -- npm run dev -- --port {{ branch | hash_port 
 url = "http://localhost:{{ branch | hash_port }}"
 
 # `wt test` is the whole suite: one pytest from the repo root covers every
-# Python test (generator/, proxy/, the install-tend scripts), and worker/ runs
-# its vitest suite and typecheck. Arguments narrow pytest (`wt test -k render`)
-# and nothing else — every step still runs, so reach for `uv run pytest` when
-# the Python half is all you want.
+# Python test (generator/, proxy/, shared/steps/, the install-tend scripts),
+# and worker/ runs its vitest suite and typecheck. Arguments narrow pytest
+# (`wt test -k render`) and nothing else — every step still runs, so reach for
+# `uv run pytest` when the Python half is all you want.
 #
 # One command per block makes each child finish and get reaped before the next
 # starts. A step's failure stops the blocks after it.
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,9 +5,10 @@ on:
   pull_request:
 
 jobs:
-  # Every Python test in the repo — pytest collects generator/, proxy/, and the
-  # install-tend scripts from the root — in the root workspace's environment,
-  # whose dev group holds the pinned mitmproxy the proxy addon imports.
+  # Every Python test in the repo — pytest collects generator/, proxy/,
+  # shared/steps/, and the install-tend scripts from the root — in the root
+  # workspace's environment, whose dev group holds the pinned mitmproxy the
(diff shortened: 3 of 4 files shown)


tend-agent (commented):
The claim checks out: `shared/steps/` really is part of the root `pytest` run — `pyproject.toml`'s own workspace comment names it two sentences later ("proxy/, shared/steps/ and the install-tend scripts dir aren't packages, so pytest's prepend import mode names their test modules by basename alone"). The edit itself is right.

**The same stale list is in three other live files, and the description says it isn't.**

The PR body asserts that `.github/workflows/ci.yaml` "already names `shared/steps/` in the comment above its `uv run --python 3.12 pytest`, so CLAUDE.md is the one place left describing the pre-port set." ci.yaml does mention `shared/steps/`, but in the interpreter-pinning comment ("the step bodies under shared/steps/ execute on the runner's own /usr/bin/python3") — a different claim about a different thing. Its *enumeration* is the comment introducing the `test:` job, and it reads "Every Python test in the repo — pytest collects generator/, proxy/, and the install-tend scripts from the root": stale in exactly the way CLAUDE.md's was.

Two more:

- **`pyproject.toml`**, the workspace-root comment — "collecting from the root reaches generator/, proxy/, and the install-tend scripts". The next sentence names `shared/steps/`, so the comment contradicts itself two lines apart.
- **`.config/wt.toml`**, the `wt test` comment — "one pytest from the repo root covers every Python test (generator/, proxy/, the install-tend scripts)".

(`CHANGELOG.md`'s [#977](https://github.com/max-sixty/tend/pull/977) entry carries the same list, but it's a historical record of what was true then — leave it.)

The PR's own argument is that a reader of the list comes away thinking the step bodies — the security preflight, the rate-limit gate, the agent supervisor, the outage tracker — are tested elsewhere or not at all. That argument covers all four sites equally; `pyproject.toml`'s is arguably the worst of them, since it's the file that actually configures the collection. Fixing one of four while asserting the rest are already correct leaves the drift in place under a stronger claim. Worth folding the three into this PR — and the "one place left" sentence needs correcting either way.

<details><summary>Where each one is</summary>

- [`.github/workflows/ci.yaml`](https://github.com/max-sixty/tend/blob/6d012cd8ada642be6a50c8fceb83be40984218b9/.github/workflows/ci.yaml#L8-L9) — comment above `test:`
- [`pyproject.toml`](https://github.com/max-sixty/tend/blob/6d012cd8ada642be6a50c8fceb83be40984218b9/pyproject.toml#L4-L6) — workspace-root header comment
- [`.config/wt.toml`](https://github.com/max-sixty/tend/blob/6d012cd8ada642be6a50c8fceb83be40984218b9/.config/wt.toml#L31-L32) — comment above `[[aliases.test]]`

</details>

CLAUDE.md
```diff
@@ -30,7 +30,7 @@ pre-commit run --all-files         # lint: ruff, typos, actionlint, shellcheck,
 ```
 
 The repo root is a uv workspace, and pytest run from it collects every Python
-test — generator/, proxy/, and the install-tend scripts — so one `pytest` is
+test — generator/, proxy/, shared/steps/, and the install-tend scripts — so one `pytest` is
 the whole Python run. `wt test` (defined in [`.config/wt.toml`](.config/wt.toml))
 adds worker/'s vitest suite and typecheck. Its arguments narrow pytest and
 nothing else, so a filtered run still pays for worker/; `uv run pytest -k render`
```
tend-agent:
```suggestion
The repo root is a uv workspace, and pytest run from it collects every Python
test — generator/, proxy/, shared/steps/, and the install-tend scripts — so
one `pytest` is the whole Python run. `wt test` (defined in
[`.config/wt.toml`](.config/wt.toml)) adds worker/'s vitest suite and
typecheck. Its arguments narrow pytest and nothing else, so a filtered run
still pays for worker/; `uv run pytest -k render`
```
The added words push this line to 91 columns; the rest of the file wraps prose in the 70s–low 80s. Reflowing the paragraph keeps it in band (text unchanged, line 37 `is the Python half on its own.` still follows).

tend-agent: Applied in 7a7a34c — reflowed to your suggested wrap, text unchanged. Same reflow needed in `pyproject.toml` and `.config/wt.toml`, where inserting `shared/steps/` pushed a line past each file's own width; folded into the same commit.

tend-agent:
Right on all three, and the "one place left" claim was wrong in the way you describe: `.github/workflows/ci.yaml` does say `shared/steps/`, but in the interpreter-pinning comment about which `python3` the step bodies execute on — not in the enumeration above `test:`, which was as stale as CLAUDE.md's. Folded the three in as 7a7a34c and rewrote the PR body so the claim matches: four sites, all four fixed, `CHANGELOG.md`'s left as the historical record.

`pyproject.toml` and `.config/wt.toml` needed the same reflow as CLAUDE.md — inserting `shared/steps/` pushed a line past each file's own width — so that's in the commit too.

One gap worth naming: `uv` isn't on the PATH in this repo's agent sandbox, so I couldn't re-run `pytest` or `pre-commit` locally after the follow-up commit. Every line in it is a comment or prose, and no Python asserts on these strings, so CI's `test` job is the verification. Polling it now.

PR #1069, opened 2026-08-27
skills(review-runs): make the run census a second input to the stranded-trigger drain

### LINKED ISSUE #1066: review-runs: a run that dies before the failure handler files no outage row, so the drain never sees it — worktrunk#3921 sat untriaged for 21 h
The `review-runs` drain reads only the `tend-outage` issue, but that row is written by the tend action's own failure handler — so a run that dies before that step is reached files nothing, uploads no session log, and is invisible to the drain. This adds Step 1's census as a second drain input (it already carries every run's `conclusion`), with the disambiguation needed to keep the new input from generating re-run spam. Verified against `max-sixty/worktrunk`: `uv run pytest` 717 passed, `pre-commit` clean on the changed file.

The concrete case: on 2026-08-26 a `tend-triage` run [startup-failed](https://github.com/max-sixty/worktrunk/actions/runs/32985839479) on `issues: opened` for [worktrunk#3921](https://github.com/max-sixty/worktrunk/issues/3921). Zero steps ran, so no outage row was filed and no artifact uploaded. The outage issue was empty all window, this drain reported nothing stranded, and the maintainer's bug report sat at zero comments for 21 hours. Nothing would ever have picked it up — nothing re-fires `issues: opened`, and the notifications poll can't see an issue the bot never touched. This run re-ran it by hand after the census turned it up.

<details><summary>Evidence, gates, and why the earlier framing was rejected</summary>

### Recurrence

This is the fifth window in `max-sixty/worktrunk`'s `review-runs` evidence log to record the same shape, and it's the one where the omission finally showed on the public record:

| Window | What died pre-handler | Outage rows filed | Consequence |
|---|---|---|---|
| 08-18 | 2 runs, pre-Claude steps | 0 | "an empty outage issue is not evidence that nothing was stranded" first recorded |
| 08-19 | 5 runs wedged in `apt`, ~6 h each | 0 | one stranded mention, swept up 6 h 22 m later by the notifications poll |
| 08-24 | 529 cluster, 4 stranded triggers | filed | drained normally — the handler *was* reached |
| 08-25 | notifications pair | 0 | cron; nothing stranded |
| **08-27** | **2 `startup_failure` runs** | **0** | **worktrunk#3921 untriaged, unrecoverable** |

Each prior entry noted that only the census caught the failures, and each declined to act.

### Why this framing passes where the prior one didn't

The 08-25 entry rejected the fix as "extending outage-issue coverage to pre-Claude steps is new mechanism, which Gate 3 rejects at any occurrence count for waste." That's right about *that* fix — a workflow-level failure handler is real machinery. This one isn't the same fix: it adds no mechanism, reuses data Step 1 already computes, and is a `jq` filter plus the prose telling you when a hit is real. Gate 3 asks what the fix leaves behind, and this leaves one filter line.

The cost class also moved. Prior windows classified as waste because a backstop swept the work up or nothing was stranded at all. Here the trigger was an `issues` event with no backstop of any kind, and it left a maintainer's report visibly unanswered — standing damage that does not self-correct, which is the class complexity is meant to be spent on.

### Gates

1. **Confidence** — High, structural, 5 windows. A `startup_failure` runs zero steps *every* time, so the outage row is never written and the drain never sees it; there is no decision point.
2. **Magnitude** — targeted fix (a missing step in an existing section), normal bar, cleared.
3. **Cost** — wrong outward action (an omitted, unrecoverable triage), and the fix is nearly free regardless.

### On the `cancelled` caveat

The naive version of this filter would tell every consumer to re-run cancelled runs, which would be a new failure mode: the same window had six `cancelled` `tend-mention` dispatches on worktrunk, all of them the designed `cancel-in-progress: false` eviction of queued siblings during a review burst. Each recorded zero steps, and each thread was answered by the winning run ([#3926](https://github.com/max-sixty/worktrunk/pull/3926), [#3925](https://github.com/max-sixty/worktrunk/pull/3925), [#3929](https://github.com/max-sixty/worktrunk/pull/3929) all show the replies). The added prose names that case so the new input doesn't turn into re-run spam.

`--failed` is dropped from the re-run line for the same reason: a `startup_failure` has no failed jobs to select, so `gh run rerun <id> --failed` is not the command that works on the case this section is being extended to cover. Plain `gh run rerun <id>` is what re-queued worktrunk#3921's triage.

### Patch

Verified in a clone of `main` at `6b09a62`: `uv run pytest` 717 passed, `pre-commit run --files` clean. I have no push access to this repo, so it is here rather than as a PR.

````diff
diff --git a/plugins/tend-ci-runner/skills/review-runs/SKILL.md b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
index 480b2b2..3b8c596 100644
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -211,13 +211,23 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
 A run that fails files a row on a `tend-outage`-labelled **"Bot temporarily unavailable"** issue, naming the run and the trigger it stranded. Nothing re-runs those triggers — `tend-review` fires only on `pull_request_target`, so a PR whose one review attempt died stays unreviewed until someone pushes. Drain the open issue as part of this sweep.
 
 ```bash
-# Usually empty; no open outage issue means nothing stranded.
+# Usually empty. That is not evidence nothing was stranded — see below.
 OUTAGE=$(gh issue list --state open --label tend-outage --json number,title \
   --jq '[.[] | select(.title == "Bot temporarily unavailable")][0].number // empty')
 gh issue view "$OUTAGE" --json body,comments --jq '.body, .comments[].body' \
   | grep -oE 'runs/[0-9]+|\| #[0-9]+'
 ```
 
+**The census is the second drain input, and it is the one that catches the worst cases.** The outage row is written by the tend action's own failure handler, so a run that dies *before* that step is reached files nothing: a `startup_failure` runs zero steps, and a wedged `actions/checkout` or a non-zero setup step dies earlier still. None of them upload a session log either, so the diagnose-then-rerun path below has nothing to key on and this drain never learns they happened. They strand the same triggers nobody re-runs. Step 1's census already carries every run's `conclusion`, so filter it — anything neither `success` nor `skipped` is a candidate, whether or not the outage issue names it:
+
+```bash
+# `skipped` is an `if:` gate declining to run, not a failure. Pipe the Step 1
+# census through this rather than re-fetching.
+jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
+```
+
+Confirm each candidate really stranded something before re-running it, because two of these conclusions routinely mean the opposite. A `cancelled` `handle` job is usually working as designed: the per-thread concurrency group is `cancel-in-progress: false`, so a burst of events on one PR evicts the *queued* siblings and the winner answers the whole thread — a cancelled job that recorded zero steps, on a thread that got its answer, is that case. And an event-triggered failure is often re-triggered by a later push. What is never recovered is an `issues`-event triage that died: nothing re-fires it, no notification surfaces an issue the bot never touched, and the report simply sits unanswered. Check the thread, then `gh run rerun <run-id>` — for a `startup_failure` there are no failed jobs to select, so `--failed` does not apply.
+
 **Diagnose first.** The nightly enrichment comment names the cause when it can. When it doesn't, read the session log — quota exhaustion surfaces as a `<synthetic>` assistant message:
 
 ```bash
````

</details>

### LINKED ISSUE #1067: review-runs: the stranded-trigger drain reads a row the failing run had to survive to write, so a job that never gets a runner is invisible
`review-runs` Step 1's **Drain stranded triggers** is keyed entirely on the `tend-outage` issue — it reads the open "Bot temporarily unavailable" issue and comments `# Usually empty; no open outage issue means nothing stranded.` But that row is written from *inside* the job. A run whose job never gets a runner writes nothing, so it strands a trigger and leaves the drain's one input empty. The census in the same step already saw the run; nothing in the skill tells a session to look at it.

## What happened on `PRQL/prql`

`tests` on `main` at `e0540ff3` concluded `failure` ([32984028168](https://github.com/PRQL/prql/actions/runs/32984028168), 2026-08-26T15:05→15:37Z) — all three of its jobs were `cancelled` with `runner_name: ""`, `steps: []`, after sitting queued ~15 min each, which reads as a runner-provisioning incident rather than a test failure.

`tend-ci-fix` fired correctly on it ([32985756174](https://github.com/PRQL/prql/actions/runs/32985756174), 15:39→15:42Z) — the `conclusion == 'failure'` gate admitted the run, so this is not [#1035](https://github.com/max-sixty/tend/issues/1035). Then its own job died the same way:

```
$ gh api repos/PRQL/prql/actions/jobs/98231410320 --jq '{status, conclusion, steps: (.steps|length)}'
{"status":"queued","conclusion":null,"steps":0}

$ gh api repos/PRQL/prql/actions/runs/32985756174/timing --jq '.billable'
{"UBUNTU":{"total_ms":0,"jobs":1,"job_runs":[{"job_id":98231410320,"duration_ms":0}]}}

$ gh run download 32985756174
no valid artifacts found to download
```

Zero billable ms, no steps, no artifact — the agent never started, so no `tend-outage` row, no diagnosis, no comment. Nothing re-runs a `workflow_run` trigger. `main` stayed red for **19h 40m** with the repo's only red-`main` watchdog reporting nothing, and today's `review-runs` reached it only because I chased the single `failure` row in the census by hand; the prescribed drain found no open outage issue and would have reported "nothing stranded".

## Why it recurs

Structural, and independent of where the outage-filing lives: any filing that runs as a step of the job is unreachable when the job gets no runner. So the drain's input is empty in exactly the failure mode that strands the most — a run that died before it could speak. The census is the only signal that survives, and the skill's only prescribed use for it is long-running / near-timeout classification.

## Suggested fix

In Step 1, make the census a second drain input rather than routing strandedness solely through the outage issue: after the census, take its non-`success`/non-`skipped` rows, and for each check whether the trigger's work is still missing before re-running — the same "re-run only what won't recover" test the section already applies to outage rows. A row with no artifact and zero billable time is the signature worth calling out, since it means nothing else on the record will ever mention it.

Correcting the `# Usually empty; no open outage issue means nothing stranded.` comment matters on its own — a session that reads it stops looking.

Happy to open the PR if you'd like the shape above; filing first since the drain is your skill's own step and you may prefer a different input.

### PR DESCRIPTION
**Pull request by tend-agent**
`review-runs`'s **Drain stranded triggers** read only the `tend-outage` issue, but that row is filed by [a step inside the failing job](https://github.com/max-sixty/tend/blob/6b09a62/claude/action.yaml#L391) — so a run that dies before reaching it files nothing, uploads no session log, and is invisible to the drain, while stranding the same trigger. This adds Step 1's census as a second drain input (it already carries every run's `conclusion`), and then fixes the remedy the drain reaches for once it has a row: a bare `gh run rerun` silently fails to drain in two distinct ways.

Reported independently by two consumer repos within a minute of each other — #1066 (`max-sixty/worktrunk`: a `startup_failure` on `issues: opened` left a maintainer's bug report untriaged for 21 h) and #1067 (`PRQL/prql`: a `tend-ci-fix` job that never got a runner left `main` red for 19 h 40 m). Both windows' drains reported "nothing stranded".

<details><summary>Why the census is the only surviving signal</summary>

`Report failure` in `claude/action.yaml` is gated on `failure()`, which covers everything from the rate-limit preflight onward — [#857](https://github.com/max-sixty/tend/pull/857) widened it for exactly that reason. What it cannot cover is a job in which *no step runs at all*: a `startup_failure` executes zero steps, and a job that never gets a runner, a wedged `actions/checkout`, or a failing adopter `setup:` step die before the action is entered. Those runs also upload no artifact, so the drain's "Diagnose first" session-log path has nothing to key on either.

Step 1's census emits `{databaseId, conclusion, createdAt, updatedAt, name}` per run over the same window, so the signal is already in hand — the skill just never told a session to use it for this. The filter is one `jq` line, so no new mechanism is introduced and nothing new is fetched.

The `cancelled` caveat is what keeps the new input from misfiring: with `cancel-in-progress: false`, a burst of events on one thread replaces the *queued* sibling, and the winning run answers the whole thread. Those runs record zero steps too, and re-running them would spam threads that already got their reply. The re-run test names that case.

</details>

<details><summary>Why <code>gh run rerun</code> is not the drain</summary>

Both failures were reported on #1066 by `max-sixty/worktrunk`'s next window and re-verified here against the live API.

**Exit 0 is not a dispatch.** A run that never got a runner has no attempt to replay. `gh run rerun` exits 0, `run_attempt` stays at 1, and `status` sits at `queued` — the stranded triage run from #1066's own report is still `{"status":"queued","run_attempt":1,"conclusion":null}` two days after its re-run was issued, and the issue it stranded still has no triage comment. This is worse than the original silence, because a queued run is not `completed`, so Step 1's census drops it and it never surfaces on a later window. The row reads as drained forever. The PR adds a `run_attempt` re-read after the re-run and says an unchanged value means not drained.

**A re-run replays the original commit's workflow file**, so it re-uses whatever action ref that commit pinned. Where the outage cause was an action-version regression, the fix ships as a pin bump on the default branch and every stranded run sits behind it by construction — the re-run reproduces the failure verbatim and files a fresh outage row, so the drain re-opens what it is draining. Verified for the runs stranded by the 0.1.19 `PATH` regression (#1071): their commit pins `max-sixty/tend/claude@0.1.19` while the default branch is on `0.1.20`. This is the shape where a re-run looks safest, because the census does show a later clean run. The PR adds the ref comparison before the re-run.

The per-trigger recovery table follows from the two: for `issues: opened` there is no re-run path at all, so a row is drained by naming what would recover it — a mention or a human — rather than by closing the issue.

</details>

## Testing

Markdown-only change to a bundled skill; no test reads this file's contents and no code path changes. The `uv` and `pre-commit` gates are not available inside the agent sandbox, so they went unrun locally — CI runs both on this PR. The two `gh api` recipes added here were run against the live runs they describe before being written down.

---
Closes #1066
Closes #1067

### DIFF
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -211,12 +211,21 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
 A run that fails files a row on a `tend-outage`-labelled **"Bot temporarily unavailable"** issue, naming the run and the trigger it stranded. Nothing re-runs those triggers — `tend-review` fires only on `pull_request_target`, so a PR whose one review attempt died stays unreviewed until someone pushes. Drain the open issue as part of this sweep.
 
 ```bash
-# Usually empty; no open outage issue means nothing stranded.
+# Usually empty. An empty issue is not evidence nothing was stranded — see below.
 OUTAGE=$(gh issue list --state open --label tend-outage --json number,title \
   --jq '[.[] | select(.title == "Bot temporarily unavailable")][0].number // empty')
 gh issue view "$OUTAGE" --json body,comments --jq '.body, .comments[].body' \
   | grep -oE 'runs/[0-9]+|\| #[0-9]+'
 ```
+
+**The census is the second drain input, and it is the one that catches the worst cases.** The outage row is written by a step *inside* the failing job, so a run that died before reaching that step files nothing and uploads no session log — a `startup_failure` runs zero steps, and a job that never gets a runner, a wedged `actions/checkout`, or a failing `setup:` step die earlier still. They strand the same triggers, and nothing else on the record ever names them. Step 1's census already carries every run's `conclusion`, so filter that rather than re-fetching:
+
+```bash
+# `skipped` is an `if:` gate declining to run, not a failure.
+jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
+```
+
+Zero billable time (`gh api "repos/$REPO/actions/runs/<run-id>/timing" --jq .billable`) plus no artifact is the signature of a run whose agent never started, so there is nothing to diagnose and the re-run test below is the whole decision.
 
 **Diagnose first.** The nightly enrichment comment names the cause when it can. When it doesn't, read the session log — quota exhaustion surfaces as a `<synthetic>` assistant message:
 
@@ -228,12 +237,13 @@ jq -r 'select(.type == "assistant") | .message.content[]?.text // empty' /tmp/ou
 
 A cluster of these is quota exhaustion, not a bug — don't open a fix PR. The reset in the message is an upper bound on the outage, not a schedule: a weekly limit can strand most of a day, and can also clear many hours early. Gate the drain on the clean-run check below, not on the stated reset.
 
-**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. Re-running the bot's own failed workflow needs no maintainer approval.
+**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. A `cancelled` run is usually the designed eviction rather than an outage: `cancel-in-progress: false` replaces the *queued* sibling when events burst on one thread, and the winning run answers the whole thread, so read the thread for that answer before re-running. What never recovers is an `issues`-event triage that died — nothing re-fires the event, and the notifications poll cannot see an issue the bot never touched. Re-running the bot's own failed workflow needs no maintainer approval.
 
 ```bash
 gh pr view <n> --json state,headRefOid,reviews \
   --jq '{state, headRefOid, reviewers: [.reviews[].author.login]}'
-gh run rerun <run-id> --failed
+# Not `--failed`: a run that never started has no failed job to select.
+gh run rerun <run-id>
 ```
 
 Close the issue once every row is drained (`gh issue close "$OUTAGE"`); one left open folds the next outage into a stale incident.
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -225,7 +225,7 @@ gh issue view "$OUTAGE" --json body,comments --jq '.body, .comments[].body' \
 jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
 ```
 
-Zero billable time (`gh api "repos/$REPO/actions/runs/<run-id>/timing" --jq .billable`) plus no artifact means the agent never started, so there is nothing to diagnose — but it does not on its own mean the run stranded anything, because a designed cancellation bills zero too. The trigger is the discriminator, so the re-run test below is the whole decision.
+A run whose agent never started shows it in the jobs endpoint: no jobs at all (`gh api "repos/$REPO/actions/runs/<run-id>/jobs" --jq .total_count` → `0`, the `startup_failure` shape), or a job with `runner_name: ""` and zero `steps` — a `skipped` job is zero-step too, but reports `runner_name: null`. There is nothing to diagnose in either case, and neither shape on its own means the run stranded anything, because a designed cancellation records the same zero steps. The trigger is the discriminator, so the re-run test below is the whole decision. Don't key on billable time: public repos aren't billed for Actions, so `.billable` reports zero on healthy runs too.
 
 **Diagnose first.** The nightly enrichment comment names the cause when it can. When it doesn't, read the session log — quota exhaustion surfaces as a `<synthetic>` assistant message:
 
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -225,7 +225,7 @@ gh issue view "$OUTAGE" --json body,comments --jq '.body, .comments[].body' \
 jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
 ```
 
-Zero billable time (`gh api "repos/$REPO/actions/runs/<run-id>/timing" --jq .billable`) plus no artifact is the signature of a run whose agent never started, so there is nothing to diagnose and the re-run test below is the whole decision.
+Zero billable time (`gh api "repos/$REPO/actions/runs/<run-id>/timing" --jq .billable`) plus no artifact means the agent never started, so there is nothing to diagnose — but it does not on its own mean the run stranded anything, because a designed cancellation bills zero too. The trigger is the discriminator, so the re-run test below is the whole decision.
 
 **Diagnose first.** The nightly enrichment comment names the cause when it can. When it doesn't, read the session log — quota exhaustion surfaces as a `<synthetic>` assistant message:
 
@@ -237,7 +237,7 @@ jq -r 'select(.type == "assistant") | .message.content[]?.text // empty' /tmp/ou
 
 A cluster of these is quota exhaustion, not a bug — don't open a fix PR. The reset in the message is an upper bound on the outage, not a schedule: a weekly limit can strand most of a day, and can also clear many hours early. Gate the drain on the clean-run check below, not on the stated reset.
 
-**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. A `cancelled` run is usually the designed eviction rather than an outage: `cancel-in-progress: false` replaces the *queued* sibling when events burst on one thread, and the winning run answers the whole thread, so read the thread for that answer before re-running. What never recovers is an `issues`-event triage that died — nothing re-fires the event, and the notifications poll cannot see an issue the bot never touched. Re-running the bot's own failed workflow needs no maintainer approval.
+**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. A `cancelled` run is usually the designed eviction rather than an outage: `cancel-in-progress: false` replaces the *queued* sibling when events burst on one thread, and the winning run answers the whole thread, so read the thread for that answer before re-running. Two triggers never recover: an `issues`-event triage that died, since nothing re-fires the event and the notifications poll cannot see an issue the bot never touched; and a `tend-ci-fix` run on a still-red default branch, since a later push re-fires it only if CI fails again and a red `main` is a branch people stop pushing to. Re-running the bot's own failed workflow needs no maintainer approval.
 
 ```bash
 gh pr view <n> --json state,headRefOid,reviews \
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -244,7 +244,7 @@ gh pr view <n> --json state,headRefOid,reviews \
   --jq '{state, headRefOid, reviewers: [.reviews[].author.login]}'
 ```
 
-**A re-run is not the recovery for every trigger.** Two never re-fire on their own, so a row whose re-run is unavailable or refused is drained by naming what would recover it, not by closing it:
+**A re-run is not the recovery for every trigger.** Only `schedule` re-fires unaided, so a row whose re-run is unavailable or refused is drained by naming what would recover it, not by closing it:
 
 | Stranded trigger | Recovery |
 |---|---|
@@ -256,14 +256,21 @@ gh pr view <n> --json state,headRefOid,reviews \
 **A re-run replays the workflow file from the original run's commit**, so it re-uses the action ref that commit pinned. Where the outage cause was an action-version regression, the fix ships as a pin bump on the default branch and every stranded run sits behind it by construction — the re-run reproduces the failure verbatim and files a fresh outage row, so the drain re-opens what it is draining. This is the shape where the drain looks safest, because the census does show a later clean run. Compare the refs first; if they differ, report the row as needing a fresh trigger instead of re-running it.
 
 ```bash
-WF=.github/workflows/<workflow>.yaml
-HEAD_SHA=$(gh api "repos/$REPO/actions/runs/<run-id>" --jq .head_sha)
-# The contents API, not `git show`: the runner's checkout is shallow, so the
-# stranded run's commit is usually not local. The working tree is the default
+# `.path` and `.head_sha` both come from the one runs call: the census names
+# the workflow, not its file, and guessing the file name from the name is what
+# the timeout-cap step above already warns against.
+read -r WF HEAD_SHA < <(gh api "repos/$REPO/actions/runs/<run-id>" --jq '"\(.path) \(.head_sha)"')
+# The contents API, not `git show`: the stranded run's commit may not be
+# reachable locally (fork head, deleted branch). Capture rather than piping
+# straight into `diff` — an unreadable file leaves both sides empty, and `diff`
+# then exits 0, which reads as a matching ref. The working tree is the default
 # branch, so it is the other side of the comparison.
-diff <(gh api "repos/$REPO/contents/$WF?ref=$HEAD_SHA" \
-         -H 'Accept: application/vnd.github.raw' | grep 'uses: max-sixty/tend/') \
-     <(grep 'uses: max-sixty/tend/' "$WF")
+if PINNED=$(gh api "repos/$REPO/contents/$WF?ref=$HEAD_SHA" \
+              -H 'Accept: application/vnd.github.raw' | grep 'uses: max-sixty/tend/'); then
+  diff <(echo "$PINNED") <(grep 'uses: max-sixty/tend/' "$WF")
+else
+  echo "$WF unreadable at $HEAD_SHA — needs a fresh trigger, not a re-run"
+fi
 ```
 
 **Exit 0 is not a dispatch.** A run that never got a runner has no attempt to replay: `gh run rerun` succeeds, `run_attempt` stays at 1, and `status` sits at `queued` indefinitely. That also drops the run out of the census — Step 1 fetches `status=completed`, so a run left queued never surfaces again and the false all-clear is permanent, which is worse than the original silence. Re-read the run afterwards and treat an unchanged `run_attempt` as *not drained*.
@@ -271,6 +278,9 @@ diff <(gh api "repos/$REPO/contents/$WF?ref=$HEAD_SHA" \
 ```bash
 # Not `--failed`: a run that never started has no failed job to select.
 gh run rerun <run-id>
+# Sleep before reading: the new attempt record takes a few seconds to surface,
+# so an immediate re-read reports attempt 1 on a re-run that did dispatch.
+sleep 30
 gh api "repos/$REPO/actions/runs/<run-id>" --jq '{status, conclusion, run_attempt}'
 ```
 
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -237,16 +237,44 @@ jq -r 'select(.type == "assistant") | .message.content[]?.text // empty' /tmp/ou
 
 A cluster of these is quota exhaustion, not a bug — don't open a fix PR. The reset in the message is an upper bound on the outage, not a schedule: a weekly limit can strand most of a day, and can also clear many hours early. Gate the drain on the clean-run check below, not on the stated reset.
 
-**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. A `cancelled` run is usually the designed eviction rather than an outage: `cancel-in-progress: false` replaces the *queued* sibling when events burst on one thread, and the winning run answers the whole thread, so read the thread for that answer before re-running. Two triggers never recover: an `issues`-event triage that died, since nothing re-fires the event and the notifications poll cannot see an issue the bot never touched; and a `tend-ci-fix` run on a still-red default branch, since a later push re-fires it only if CI fails again and a red `main` is a branch people stop pushing to. Re-running the bot's own failed workflow needs no maintainer approval.
+**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. A `cancelled` run is usually the designed eviction rather than an outage: `cancel-in-progress: false` replaces the *queued* sibling when events burst on one thread, and the winning run answers the whole thread, so read the thread for that answer before re-running. Re-running the bot's own failed workflow needs no maintainer approval.
 
 ```bash
 gh pr view <n> --json state,headRefOid,reviews \
   --jq '{state, headRefOid, reviewers: [.reviews[].author.login]}'
-# Not `--failed`: a run that never started has no failed job to select.
-gh run rerun <run-id>
 ```
 
-Close the issue once every row is drained (`gh issue close "$OUTAGE"`); one left open folds the next outage into a stale incident.
+**A re-run is not the recovery for every trigger.** Two never re-fire on their own, so a row whose re-run is unavailable or refused is drained by naming what would recover it, not by closing it:
+
+| Stranded trigger | Recovery |
+|---|---|
+| `schedule` | the next tick |
+| `pull_request_target` (review) | a push, or close-and-reopen — `reopened` is in the trigger's `types` |
+| `workflow_run` (ci-fix) | another failing run of the watched workflow; a push that turns CI green re-fires nothing, and a still-red default branch is one people stop pushing to |
+| `issues: opened` (triage) | nothing re-fires it, and the notifications poll cannot see an issue the bot never touched — an `@`-mention on the issue, or a human |
+
+**A re-run replays the workflow file from the original run's commit**, so it re-uses the action ref that commit pinned. Where the outage cause was an action-version regression, the fix ships as a pin bump on the default branch and every stranded run sits behind it by construction — the re-run reproduces the failure verbatim and files a fresh outage row, so the drain re-opens what it is draining. This is the shape where the drain looks safest, because the census does show a later clean run. Compare the refs first; if they differ, report the row as needing a fresh trigger instead of re-running it.
+
+```bash
+WF=.github/workflows/<workflow>.yaml
+HEAD_SHA=$(gh api "repos/$REPO/actions/runs/<run-id>" --jq .head_sha)
+# The contents API, not `git show`: the runner's checkout is shallow, so the
+# stranded run's commit is usually not local. The working tree is the default
+# branch, so it is the other side of the comparison.
+diff <(gh api "repos/$REPO/contents/$WF?ref=$HEAD_SHA" \
+         -H 'Accept: application/vnd.github.raw' | grep 'uses: max-sixty/tend/') \
+     <(grep 'uses: max-sixty/tend/' "$WF")
+```
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by tend-agent** [comment]
Triaging #1067 — a second consumer report of this root cause, filed a minute after #1066 — turned up one case the "Re-run only what won't recover" paragraph still drops. The census-input paragraph covers it ("a job that never gets a runner"), but the re-run test then says "a later push often re-triggers it" and names only an `issues`-event triage as never-recovering.

A `workflow_run`-triggered `tend-ci-fix` on a still-red default branch is a second never-recovering trigger. A push re-fires ci-fix only if CI fails *again*, and a red `main` is a branch people stop pushing to — so the watchdog for the red branch is itself the thing that got stranded, and nothing queues to replace it.

On `PRQL/prql`, run `32985756174` (`tend-ci-fix`, `workflow_run`, `conclusion: failure`) recorded `steps: 0`, `runner_name: ""`, and 0 billable ms — the signature this PR adds. `main` then stayed red at `e0540ff3` for ~20h: no later `tend-ci-fix` run at all, and no `tests` run on `main` until a manual re-run today. Naming that trigger alongside the `issues` one would make the paragraph cover it.

```
$ gh api repos/PRQL/prql/actions/runs/32985756174/jobs --jq '.jobs[] | {name, status, conclusion, runner_name, steps: (.steps|length)}'
{"conclusion":null,"name":"fix-ci","runner_name":"","status":"queued","steps":0}

$ gh api repos/PRQL/prql/actions/runs/32985756174/timing --jq '.billable'
{"UBUNTU":{"job_runs":[{"duration_ms":0,"job_id":98231410320}],"jobs":1,"total_ms":0}}

$ gh api "repos/PRQL/prql/actions/workflows/tend-ci-fix.yaml/runs?per_page=4" --jq '.workflow_runs[] | "\(.id) \(.created_at) \(.conclusion)"'
32985756174 2026-08-26T15:39:11Z failure
32967006780 2026-08-26T12:08:52Z skipped
32965802856 2026-08-26T11:55:13Z success
```

</details>

File: plugins/tend-ci-runner/skills/review-runs/SKILL.md [hunk 1]
```diff
@@ -211,13 +211,22 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
...
 ```
 
+**The census is the second drain input, and it is the one that catches the worst cases.** The outage row is written by a step *inside* the failing job, so a run that died before reaching that step files nothing and uploads no session log — a `startup_failure` runs zero steps, and a job that never gets a runner, a wedged `actions/checkout`, or a failing `setup:` step die earlier still. They strand the same triggers, and nothing else on the record ever names them. Step 1's census already carries every run's `conclusion`, so filter that rather than re-fetching:
+
+```bash
+# `skipped` is an `if:` gate declining to run, not a failure.
+jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
+```
+
+Zero billable time (`gh api "repos/$REPO/actions/runs/<run-id>/timing" --jq .billable`) plus no artifact is the signature of a run whose agent never started, so there is nothing to diagnose and the re-run test below is the whole decision.
```
tend-agent: Public repos aren't billed for Actions minutes, so `.billable` reports zero on *every* run there — not just dead ones. On `max-sixty/tend` a successful 3-job `tend-mention` run returns `{"UBUNTU":{"total_ms":0,"jobs":3,"job_runs":[{"duration_ms":0},...]}}`, identical in the field this line keys on to PRQL's dead `tend-ci-fix` run. And a `startup_failure` returns `{}` rather than a zero, so "zero billable time" doesn't describe it either. Since most tend consumers are public repos, a session following this reads "agent never started" on healthy runs.

The jobs endpoint discriminates on both public and private repos — the `startup_failure` run has no jobs at all, and PRQL's job reports `runner_name: ""` with `steps: 0`:

```suggestion
No jobs at all (`gh api "repos/$REPO/actions/runs/<run-id>/jobs" --jq .total_count` → `0`, the `startup_failure` shape) or a job with `runner_name: ""` and zero `steps` is the signature of a run whose agent never started, so there is nothing to diagnose and the re-run test below is the whole decision. Don't use billable time for this: public repos aren't billed, so it is zero on healthy runs too.
```

tend-agent (commented):
The premise checks out: `Report failure` is a step inside the job (`claude/action.yaml:391`, `codex/action.yaml:259`, both gated on `failure()`), so a run that never enters the action files no row — and `status=completed` in Step 1's census does return `startup_failure` runs, so the census really is the surviving signal. The `cancelled` caveat and dropping `--failed` both follow.

One line doesn't hold up: the billable-time signature. Details inline.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Confirmed, and fixed in `17963bf`. `max-sixty/tend` is public too, so a *successful* 3-job `tend-mention` run reports `{"UBUNTU":{"total_ms":0,"jobs":3,"job_runs":[{"duration_ms":0},...]}}` — indistinguishable in that field from PRQL's dead `tend-ci-fix` run. The line now keys on the jobs endpoint as suggested.

One addition to the wording: a `skipped` job also reports zero `steps`, so the sentence contrasts `runner_name: ""` (never got a runner) with `runner_name: null` (skipped), which keeps a partially-skipped `failure` run from reading as never-started. I also kept the designed-cancellation caveat that landed in `c8e25a5` — it applies to the new signature for the same reason, since an evicted run records zero steps too.

```
# healthy, public repo — same zero the dead run shows
$ gh api repos/max-sixty/tend/actions/runs/33067937462/jobs --jq '.jobs[] | {name, conclusion, runner_name, steps: (.steps|length)}'
{"conclusion":"success","name":"relay","runner_name":"GitHub Actions 1000210138","steps":3}
{"conclusion":"skipped","name":"verify","runner_name":null,"steps":0}
{"conclusion":"skipped","name":"handle","runner_name":null,"steps":0}
$ gh api repos/max-sixty/tend/actions/runs/33067937462/timing --jq '.billable'
{"UBUNTU":{"job_runs":[{"duration_ms":0,...},...],"jobs":3,"total_ms":0}}

# startup_failure — no jobs at all
$ gh api repos/max-sixty/worktrunk/actions/runs/32985839479/jobs --jq '.total_count'
0

# never got a runner
$ gh api repos/PRQL/prql/actions/runs/32985756174/jobs --jq '.jobs[] | {runner_name, steps: (.steps|length)}'
{"runner_name":"","steps":0}
```

</details>

tend-agent (commented):
The `tend-ci-fix` addition checks out against the workflow's own trigger — `workflow_run` on `ci`, gated `conclusion == 'failure'`, so a push that turns CI green re-fires nothing and a red `main` nobody pushes to strands it indefinitely.

The billable-time line is the one still outstanding, and this push strengthened it rather than fixing it — "is the signature of" became "means", and the new clause builds the stranding/never-started distinction on top of the same premise. The premise doesn't hold: GitHub doesn't bill public-repo Actions minutes, so `.billable` reports zero for every run there, healthy ones included. PRQL's own successful 34-job run `32012160421` returns `total_ms: 0` on both `MACOS` and `UBUNTU` — indistinguishable, in the field this line keys on, from the dead `tend-ci-fix` run the PR cites. Most consumers are public repos, so a session following this reads "the agent never started" on runs that ran fine.

The jobs endpoint separates all three shapes, on public and private repos alike. Inline suggestion folds this push's new distinction into that discriminator.

```
$ gh api repos/PRQL/prql/actions/runs/32012160421/jobs --jq '.jobs[0] | {conclusion, runner_name, steps: (.steps|length)}'
{"conclusion":"success","runner_name":"GitHub Actions 1000127903","steps":10}
$ gh api repos/PRQL/prql/actions/runs/32012160421/timing --jq .billable
{"MACOS":{...,"total_ms":0},"UBUNTU":{...,"total_ms":0}}      # healthy, 34 jobs

$ gh api repos/PRQL/prql/actions/runs/32985756174/jobs --jq '.jobs[0] | {conclusion, runner_name, steps: (.steps|length)}'
{"conclusion":null,"runner_name":"","steps":0}
$ gh api repos/PRQL/prql/actions/runs/32985756174/timing --jq .billable
{"UBUNTU":{...,"total_ms":0}}                                  # dead — same zero

$ gh api repos/max-sixty/tend/actions/runs/32984004073/jobs --jq .total_count
0
$ gh api repos/max-sixty/tend/actions/runs/32984004073/timing --jq .billable
{}                                                             # startup_failure — not a zero at all
```

</details>

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -211,13 +211,22 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
...
+
+```bash
+# `skipped` is an `if:` gate declining to run, not a failure.
+jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
+```
+
+A run whose agent never started shows it in the jobs endpoint: no jobs at all (`gh api "repos/$REPO/actions/runs/<run-id>/jobs" --jq .total_count` → `0`, the `startup_failure` shape), or a job with `runner_name: ""` and zero `steps` — a `skipped` job is zero-step too, but reports `runner_name: null`. There is nothing to diagnose in either case, and neither shape on its own means the run stranded anything, because a designed cancellation records the same zero steps. The trigger is the discriminator, so the re-run test below is the whole decision. Don't key on billable time: public repos aren't billed for Actions, so `.billable` reports zero on healthy runs too.
```
tend-agent, line 228: `.billable` is zero on every run in a public repo, so this doesn't discriminate where most consumers are — see the review body. The jobs endpoint does, and it also separates the `startup_failure` shape, which returns `{}` rather than a zero. Keeping this push's stranding/never-started distinction:

```suggestion
No jobs at all (`gh api "repos/$REPO/actions/runs/<run-id>/jobs" --jq .total_count` → `0`, the `startup_failure` shape), or a job reporting `runner_name: ""` with zero `steps`, means the agent never started, so there is nothing to diagnose — but it does not on its own mean the run stranded anything, because a designed cancellation looks the same. The trigger is the discriminator, so the re-run test below is the whole decision. Don't key this on billable time: public repos aren't billed for Actions minutes, so `.billable` is zero on healthy runs too.
```

tend-agent: Already applied — `17963bf`, which landed while this review was being written (it's stamped on that commit, but reviews the wording it replaced). The pushed line is the same discriminator; it differs only in also naming the `skipped` case, since a skipped job reports zero `steps` too but with `runner_name: null` rather than `""`, which would otherwise make a partially-skipped `failure` run read as never-started.

tend-agent (commented):
The census-as-second-input argument holds, and both new failure modes are real — I re-checked the ref-replay one against the live API and the shape is present right now: `main`'s `tend-mention.yaml` still pins `max-sixty/tend/claude@0.1.19` at `e4e6bf3`, one release behind the `0.1.20` tag, so a run stranded across a pin bump would indeed replay the old ref.

Two of the three new recipes don't do what the prose beside them says, and both fail toward the wrong drain verdict:

- The `run_attempt` re-read fires before the attempt record exists, so a re-run that *did* dispatch reads as "not drained".
- The ref comparison fails open: when the hand-filled `<workflow>.yaml` doesn't resolve, both sides of the `diff` are empty, it exits 0, and that reads as "refs match" — green-lighting exactly the re-run the guard exists to stop.

Details and suggestions inline. One aside on the comment in that block: every generated checkout sets `fetch-depth: 0` ([`macros.yaml.j2`](https://github.com/max-sixty/tend/blob/1a5a60df16601b451a628961ddc6256cecf11a95/generator/src/tend/templates/macros.yaml.j2#L227), and `tend-review-runs.yaml` renders it), so "the runner's checkout is shallow" isn't the reason to prefer the contents API — the reason that survives is that the commit may simply not be reachable there (fork head, deleted branch).

```
# the runs API already returns the workflow file path — no name-to-file guess needed
$ gh api repos/max-sixty/tend/actions/runs/33129362265 --jq '{name, path, head_sha}'
{"head_sha":"e4e6bf3e...","name":"tend-mention","path":".github/workflows/tend-mention.yaml"}

# the guard as written, with a path that doesn't resolve: silent exit 0
$ diff <(gh api "repos/max-sixty/tend/contents/.github/workflows/nope.yaml?ref=main" \
           -H 'Accept: application/vnd.github.raw' 2>/dev/null | grep 'uses: max-sixty/tend/') \
       <(grep 'uses: max-sixty/tend/' .github/workflows/nope.yaml 2>/dev/null)
$ echo $?
0

# rerun-failed-jobs.sh, same plugin, on why one read isn't enough:
#   "The new attempt record takes a few seconds to surface."
#   for _ in $(seq 1 6); do sleep 5; ATTEMPT=$(gh api "$RUN_API" --jq '.run_attempt'); ...
```

</details>

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -228,15 +237,44 @@ jq -r 'select(.type == "assistant") | .message.content[]?.text // empty' /tmp/ou
...
+         -H 'Accept: application/vnd.github.raw' | grep 'uses: max-sixty/tend/') \
+     <(grep 'uses: max-sixty/tend/' "$WF")
+```
+
+**Exit 0 is not a dispatch.** A run that never got a runner has no attempt to replay: `gh run rerun` succeeds, `run_attempt` stays at 1, and `status` sits at `queued` indefinitely. That also drops the run out of the census — Step 1 fetches `status=completed`, so a run left queued never surfaces again and the false all-clear is permanent, which is worse than the original silence. Re-read the run afterwards and treat an unchanged `run_attempt` as *not drained*.
+
+```bash
+# Not `--failed`: a run that never started has no failed job to select.
+gh run rerun <run-id>
+gh api "repos/$REPO/actions/runs/<run-id>" --jq '{status, conclusion, run_attempt}'
```
tend-agent, line 284: This re-read races the attempt record. `scripts/rerun-failed-jobs.sh` in this same plugin reads `run_attempt` *before* the rerun and then polls up to 30 s for it to advance, with the header comment "The new attempt record takes a few seconds to surface" — so an immediate read here returns `1` for a re-run that dispatched fine, the session records the row as not drained, and the outage issue stays open naming a fresh trigger nobody needs to fire. That's the same stale-incident hazard the closing paragraph warns about, approached from the other side.

A sleep is the whole fix — the run is queued for minutes afterwards either way, so there is nothing to poll for:

```suggestion
# Not `--failed`: a run that never started has no failed job to select.
gh run rerun <run-id>
# Sleep before reading: the new attempt record takes a few seconds to surface,
# so an immediate re-read reports attempt 1 on a re-run that did dispatch.
sleep 30
gh api "repos/$REPO/actions/runs/<run-id>" --jq '{status, conclusion, run_attempt}'
```

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -228,15 +237,44 @@ jq -r 'select(.type == "assistant") | .message.content[]?.text // empty' /tmp/ou
...
+
+```bash
+WF=.github/workflows/<workflow>.yaml
+HEAD_SHA=$(gh api "repos/$REPO/actions/runs/<run-id>" --jq .head_sha)
+# The contents API, not `git show`: the runner's checkout is shallow, so the
+# stranded run's commit is usually not local. The working tree is the default
+# branch, so it is the other side of the comparison.
+diff <(gh api "repos/$REPO/contents/$WF?ref=$HEAD_SHA" \
+         -H 'Accept: application/vnd.github.raw' | grep 'uses: max-sixty/tend/') \
+     <(grep 'uses: max-sixty/tend/' "$WF")
```
tend-agent: Two problems, and they compound into a guard that fails open.

`WF` is hand-filled from a workflow *name* — the census emits `name`, not a path — and the timeout-cap step earlier in this same step warns against exactly that inference ("the census admits workflows named outside the `tend-` prefix, so don't glob for one"). A hand-maintained workflow, or any adopter file not named `tend-<name>.yaml`, gets the wrong path.

Then the wrong path is silent: `gh api` 404s inside `<(...)` where its non-zero exit is discarded, the local `grep` on the same missing file also emits nothing, and `diff` compares empty to empty and exits 0 — which this recipe's caller reads as "refs match, safe to re-run". The one case the guard exists for is the one that slips through it.

The runs call already in the block returns `.path` alongside `.head_sha`, so the guess is unnecessary; capturing the fetch instead of piping it straight into `diff` makes an unreadable file loud:

```suggestion
# `.path` and `.head_sha` both come from the one runs call: the census names
# the workflow, not its file, and guessing the file name is what the
# timeout-cap step above already warns against.
read -r WF HEAD_SHA < <(gh api "repos/$REPO/actions/runs/<run-id>" --jq '"\(.path) \(.head_sha)"')
# The contents API, not `git show`: the stranded run's commit may not be
# reachable locally. Capture rather than piping straight into `diff` — an
# unreadable file leaves both sides empty, and `diff` then exits 0 and reads
# as a matching ref.
PINNED=$(gh api "repos/$REPO/contents/$WF?ref=$HEAD_SHA" \
           -H 'Accept: application/vnd.github.raw' | grep 'uses: max-sixty/tend/') \
  || echo "could not read $WF at $HEAD_SHA — needs a fresh trigger, not a re-run"
diff <(echo "$PINNED") <(grep 'uses: max-sixty/tend/' "$WF")
```

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -228,15 +237,44 @@ jq -r 'select(.type == "assistant") | .message.content[]?.text // empty' /tmp/ou
...
+**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. A `cancelled` run is usually the designed eviction rather than an outage: `cancel-in-progress: false` replaces the *queued* sibling when events burst on one thread, and the winning run answers the whole thread, so read the thread for that answer before re-running. Re-running the bot's own failed workflow needs no maintainer approval.
 
 ```bash
 gh pr view <n> --json state,headRefOid,reviews \
   --jq '{state, headRefOid, reviewers: [.reviews[].author.login]}'
-gh run rerun <run-id> --failed
 ```
 
-Close the issue once every row is drained (`gh issue close "$OUTAGE"`); one left open folds the next outage into a stale incident.
+**A re-run is not the recovery for every trigger.** Two never re-fire on their own, so a row whose re-run is unavailable or refused is drained by naming what would recover it, not by closing it:
```
tend-agent: The count doesn't match the table under it: by its own rows, `schedule` is the only trigger that re-fires unaided — the other three all wait on an external event (a push, another failing run, a human).

```suggestion
**A re-run is not the recovery for every trigger.** Only `schedule` re-fires unaided, so a row whose re-run is unavailable or refused is drained by naming what would recover it, not by closing it:
```

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Confirmed and applied in `e7435ae`. Took the suggestion as written — the run stays queued for minutes afterwards either way, so a fixed sleep is enough and there is nothing to poll toward.

Worth noting why `rerun-failed-jobs.sh` isn't just called here: it passes `--failed`, which is the one flag this recipe can't use, since a run that never started has no failed job to select.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Both confirmed, applied in `e7435ae`. Took `.path` from the runs call and made the unreadable case loud, with one change to the suggestion: `PINNED=$(...) || echo ...` still falls through to the `diff`, so the failure branch is an `if`/`else` and the comparison only runs when the fetch succeeded.

Also corrected the shallow-checkout comment to reachability, per the review body.

<details><summary>Both paths run live</summary>

```
$ read -r WF HEAD_SHA < <(gh api "repos/max-sixty/tend/actions/runs/33129362265" --jq '"\(.path) \(.head_sha)"')
$ echo "$WF $HEAD_SHA"
.github/workflows/tend-mention.yaml e4e6bf3ed836c686bfa1b82d37378a8d28fb56ba
# → "refs match"

$ WF=.github/workflows/nope.yaml   # the fails-open case
# → ".github/workflows/nope.yaml unreadable at main — needs a fresh trigger, not a re-run"
```

</details>

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Right — three of the four rows wait on an external event, not two. Applied verbatim in `e7435ae`.

tend-agent (commented):
All three findings from the last round are fixed. I re-ran both reworked guards against the live API rather than reading them, and the fail-open case that motivated the ref guard now takes the `else` branch: `grep`'s exit status governs the `if`, so an unreadable file reports `needs a fresh trigger` instead of `diff`-ing empty against empty and exiting 0. `.path` off the runs call removes the name-to-file guess the timeout-cap step warns against, and the sentence and the table under it now agree — `schedule` is the one row that recovers unaided. Taking the `sleep` as suggested is right: the non-dispatch case is by construction a run whose attempt 1 never ran, so the paragraph's "stays at 1" is the baseline the re-read needs, and `rerun-failed-jobs.sh`'s poll would only be buying precision this recipe can't use (it passes `--failed`, which is the one flag here that can't be used).

One observation, not a blocker: the drain's two guards are now ~20 lines of branching bash living in a markdown fence, which is the shape `/tend-ci-runner:running-in-ci` sends to a script under **Scripts over prose recipes** ("when a skill's code block needs edge-case handling ... put the logic in a tested script"). The evidence is this PR's own history — the first version of the ref guard shipped fail-open and was caught by hand, twice over. A script in `plugins/tend-ci-runner/scripts/` would get shellcheck at `-S warning` from pre-commit (`files: ^(...|plugins/.*\.sh|...)$`) and a test in the generator suite; the fence gets neither. Worth a follow-up rather than a fourth round here.

<details><summary>Both guard branches, run live</summary>

```
$ read -r WF HEAD_SHA < <(gh api "repos/max-sixty/tend/actions/runs/33129362265" --jq '"\(.path) \(.head_sha)"')
$ echo "$WF $HEAD_SHA"
.github/workflows/tend-mention.yaml e4e6bf3ed836c686bfa1b82d37378a8d28fb56ba

# readable: then branch, diff exits 0
-- then branch --
refs match (exit 0)

# the fails-open case from the last round, now closed
$ WF=.github/workflows/nope.yaml
ELSE branch taken — fails closed: .github/workflows/nope.yaml unreadable at e4e6bf3e...
```

</details>

max-sixty:
Closing this version because it turns one missing census input into 52 lines of branching recovery instructions, including inline shell for action-pin comparison and rerun dispatch verification. The census signal is valid, but the recovery behavior needs a tested script or a smaller report-only change. Leaving #1066 and #1067 open for that simpler follow-up.

tend-agent:
Report-only version opened as #1073: the census filter plus one paragraph saying a re-run is not proof of recovery. 14 lines added rather than 52, no recovery table, and the only shell is the one `jq` filter — the pin comparison and the dispatch-verification block are gone. Where a re-run may not have recovered the trigger, the drain reports the row on the outage issue and leaves it open instead of branching on the failure shape.

PR #1071, opened 2026-08-27
Preserve setup-added tool paths in the Claude sandbox

max-sixty:
## What changed

The Claude action now carries the runner's original `PATH` into sandbox setup as data while continuing to resolve privileged setup commands from fixed system directories. The setup script consumes and unsets that transport variable before running external commands.

The hosted sandbox test now executes the exact `run:` command from `claude/action.yaml`. This closes the gap that let v0.1.19 pass CI even though production discarded every path added by earlier setup actions through `GITHUB_PATH`.

## Why

Worktrunk's Tend workflows install `uv` with `astral-sh/setup-uv`, then invoke it from `sandbox_setup`. v0.1.19 erased the toolcache path before constructing `AGENT_PATH`, so triage and notification-recovery runs failed before the agent started.

## Verification

- `wt test` (717 Python tests, worker tests, and worker typecheck)
- independent review of the privilege boundary, quoting, and regression-test shape

The PR's hosted `test-sandbox` job exercises the second UID, `sudo`, and `GITHUB_PATH` behavior that macOS cannot reproduce locally.

### DIFF
--- a/claude/action.yaml
+++ b/claude/action.yaml
@@ -247,7 +247,7 @@ runs:
         # the agent env file. sandbox_setup (commands) runs in a later step.
         TEND_SANDBOX_PATH: ${{ inputs.sandbox_path }}
         TEND_SANDBOX_ENV: ${{ inputs.sandbox_env }}
-      run: PATH=/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/bash "${{ github.action_path }}/../proxy/setup-sandbox.sh"
+      run: /usr/bin/env TEND_RUNNER_TOOL_PATH="$PATH" PATH=/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/bash "${{ github.action_path }}/../proxy/setup-sandbox.sh"
 
     # Install the claude binary into the sandbox user's home. Nothing else —
     # a language toolchain here would shadow the adopter's on the sandbox PATH.
--- a/proxy/setup-sandbox.sh
+++ b/proxy/setup-sandbox.sh
@@ -25,7 +25,9 @@
 # api.anthropic.com), ACTION_PATH (this action's checkout), MITMPROXY_VERSION
 # (pinned mitmproxy version), TEND_UV_DIR (tend's own pinned uv, installed by
 # shared/steps/install-proxy-uv.sh). GITHUB_WORKSPACE / RUNNER_TEMP /
-# UV_CACHE_DIR come from Actions. Optional adopter levers (from
+# UV_CACHE_DIR come from Actions. TEND_RUNNER_TOOL_PATH carries the runner PATH
+# as data while this process resolves commands from fixed system directories.
+# Optional adopter levers (from
 # .config/tend.yaml): TEND_SANDBOX_PATH
 # (newline-separated dirs prepended to the sandbox PATH) and TEND_SANDBOX_ENV
 # (newline-separated NAME=VALUE pairs added to the agent env; reserved keys
@@ -37,7 +39,8 @@ set -euo pipefail
 # but never resolve a privileged setup utility through it while real credentials
 # are in this process. The hosted image supplies every command this script uses
 # from these system directories.
-RUNNER_TOOL_PATH="${PATH}"
+RUNNER_TOOL_PATH="${TEND_RUNNER_TOOL_PATH:-${PATH}}"
+unset TEND_RUNNER_TOOL_PATH
 PATH=/usr/sbin:/usr/bin:/sbin:/bin
 export PATH
 
--- a/proxy/test-setup-sandbox.sh
+++ b/proxy/test-setup-sandbox.sh
@@ -56,7 +56,7 @@ plant() {
 }
 
 setup() {
-  local agent_path path_entry
+  local action_run agent_path path_entry
   set_inputs
   # The workspace path leads; a literal `~` exercises expansion against the
   # sandbox home. The configured directory may be populated later.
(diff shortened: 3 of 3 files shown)


tend-agent (commented):
This merged while the review was running, so it's a note on landed code rather than a gate.

The diagnosis is right — #1048 added the `run:` PATH prefix and the `RUNNER_TOOL_PATH="${PATH}"` capture in the same commit, so the capture read the value the prefix had just clobbered — and `test-sandbox` is green on the hosted job, so the fix works. My one substantive concern is the shape.

The `PATH=/usr/sbin:/usr/bin:/sbin:/bin` prefix that the transport variable exists to work around doesn't guard anything. Every token on that `run:` line is absolute (`/usr/bin/env`, `/usr/bin/bash`, and `${{ github.action_path }}/…`), and the step's own `shell:` resolves the interpreter absolutely too. Inside the script, everything before its own `PATH=/usr/sbin:/usr/bin:/sbin:/bin` is an assignment, a builtin, or a function definition — the first external command is the `readlink -f` on `GITHUB_WORKSPACE`, 35 lines later. So nothing resolves through the adopter's PATH between the step's shell starting and the script's own reset, prefix or no prefix.

Deleting the prefix fixes the same regression by deletion rather than by adding a mechanism: no `TEND_RUNNER_TOOL_PATH`, no `:-${PATH}` fallback, no `unset`, no new entry in the input list. Per CLAUDE.md, "Prefer deleting a mechanism over refining it." The test change is worth keeping either way — reading the `run:` line from the YAML is what makes a future divergence visible, and it would have caught this one.

If you'd rather keep the prefix as defense in depth against a later edit adding a command to that line, the two inline notes on `setup-sandbox.sh` and `test-setup-sandbox.sh` are what I'd want alongside it.

claude/action.yaml
```diff
@@ -247,7 +247,7 @@ runs:
         # the agent env file. sandbox_setup (commands) runs in a later step.
         TEND_SANDBOX_PATH: ${{ inputs.sandbox_path }}
         TEND_SANDBOX_ENV: ${{ inputs.sandbox_env }}
-      run: PATH=/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/bash "${{ github.action_path }}/../proxy/setup-sandbox.sh"
+      run: /usr/bin/env TEND_RUNNER_TOOL_PATH="$PATH" PATH=/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/bash "${{ github.action_path }}/../proxy/setup-sandbox.sh"
```
tend-agent, line 250: Deleting the prefix removes the need for the transport variable entirely. Companion change: `proxy/setup-sandbox.sh` back to `RUNNER_TOOL_PATH="${PATH}"`, dropping the `unset` and the `TEND_RUNNER_TOOL_PATH` sentence from the input list at the top.

```suggestion
      run: /usr/bin/bash "${{ github.action_path }}/../proxy/setup-sandbox.sh"
```

proxy/setup-sandbox.sh
```diff
@@ -37,7 +39,8 @@ set -euo pipefail
 # but never resolve a privileged setup utility through it while real credentials
 # are in this process. The hosted image supplies every command this script uses
 # from these system directories.
-RUNNER_TOOL_PATH="${PATH}"
+RUNNER_TOOL_PATH="${TEND_RUNNER_TOOL_PATH:-${PATH}}"
```
tend-agent, line 42: If the transport variable stays: `:-${PATH}` silently falls back to the PATH the caller already clobbered, which is exactly the v0.1.19 failure — the sandbox loses every `setup:`-installed tool and nothing in the log says why. Every other required input here (`TEND_GH_TOKEN`, `MITMPROXY_VERSION`, `GITHUB_WORKSPACE`) emits `::error::` and exits; per CLAUDE.md this one should too — "old formats should fail with a clear error, not silently parse."

`verify_refusals()` still invokes the script directly, so it would need `TEND_RUNNER_TOOL_PATH` exported from `set_inputs` for its two cases to reach the errors they assert on.

proxy/test-setup-sandbox.sh
```diff
@@ -73,7 +73,12 @@ setup() {
   export MITMPROXY_VERSION
   UV_VERSION=$(yq -e '.inputs.uv_version.default' claude/action.yaml) \
     bash shared/steps/install-proxy-uv.sh
-  bash proxy/setup-sandbox.sh | tee "$RUNNER_TEMP/setup.log"
+  # Exercise the composite action's entrypoint rather than calling the setup
+  # script directly. The boundary depends on the PATH the action passes in.
+  action_run=$(yq -er '.runs.steps[] | select(.name == "Set up credential-isolation sandbox") | .run' claude/action.yaml)
+  action_run=${action_run//'${{ github.action_path }}'/"$GITHUB_WORKSPACE/claude"}
+  /usr/bin/bash --noprofile --norc -eo pipefail -c "$action_run" \
```
tend-agent, line 80: Two gaps this leaves, both relevant if the `run:` prefix stays:

- `shell:` is still hand-copied. The step's `shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}` is the line that resolves the credential-holding step's interpreter outside the adopter PATH, and it's the half not read from the YAML — changing it to a bare `bash` leaves this test green.
- The fake `sudo` that `plant()` drops into the runner-home probe dir no longer reaches `setup-sandbox.sh`: the entrypoint clobbers PATH before the script starts, so the script's own `PATH=/usr/sbin:/usr/bin:/sbin:/bin` / `export PATH` pair could be deleted and the suite would stay green. The direct call proved that layer before. Dropping the prefix restores it for free.

PR #1073, opened 2026-08-28
skills(review-runs): add the run census as a second stranded-trigger drain input

### LINKED ISSUE #1066: review-runs: a run that dies before the failure handler files no outage row, so the drain never sees it — worktrunk#3921 sat untriaged for 21 h

### LINKED ISSUE #1067: review-runs: the stranded-trigger drain reads a row the failing run had to survive to write, so a job that never gets a runner is invisible

tend-agent:
A run that dies before the failure handler can strand its trigger without filing an outage row or uploading a session log. The existing drain treats an empty outage issue as proof that nothing was stranded.

The run census now supplies a second signal. For a failed census row absent from the outage issue, the skill confirms that the work is missing and reports the stranded trigger as a finding; a census-only failure is not a row on the outage issue and so does not hold its close. It does not re-run the row, which has no failure record showing which jobs are safe to replay. The existing recovery path for recorded outage rows is unchanged, `--failed` included — an outage row exists only because a job ran and failed, so it always has a failed job to select.

Scope against the two issues: both suggested re-running the census-only rows as well, and this lands the detection half only. The invisibility each issue reports — a stranded trigger that nothing on the record ever names — is what the sweep can fix by itself, and once the row is a finding a human can replay it. An automatic re-run of a run that recorded no jobs is the part with no evidence behind it. Closing both on that basis; the recovery half is worth its own change if it is wanted.

Verification: pre-commit on the changed skill file. This is a report-only documentation change that retains the existing census with `tee`, filters it with `jq`, and adds no recovery commands.

Closes #1066
Closes #1067

### DIFF
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -181,7 +181,7 @@ FETCH_FROM=$(date -u -d "$SINCE - 24 hours" +%Y-%m-%dT%H:%M:%SZ)
 for workflow in $(gh api --paginate repos/$REPO/actions/workflows --jq ".workflows[] | select(.name | test(\"$PREFIX_RE\")) | .id"); do
   gh api --paginate "repos/$REPO/actions/workflows/$workflow/runs?created=>=$FETCH_FROM&status=completed&per_page=100" \
     --jq ".workflow_runs[] | select(.updated_at >= \"$SINCE\") | {databaseId: .id, conclusion, createdAt: .created_at, updatedAt: .updated_at, name: .name}"
-done
+done | tee /tmp/review-runs-census.jsonl
 ```
 
 If no runs found, report "no runs to review" and exit.
@@ -222,7 +222,8 @@ gh issue view "$OUTAGE" --json body,comments --jq '.body, .comments[].body' \
 
 ```bash
 # `skipped` is an `if:` gate declining to run, not a failure.
-jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
+jq -c 'select(.conclusion != "success" and .conclusion != "skipped")' \
+  /tmp/review-runs-census.jsonl
 ```
 
 For each census failure absent from the outage issue, confirm the trigger's work is still missing. Treat a cancelled run as a designed eviction when another run answered the same thread. Report a stranded trigger as a finding, and leave any outage issue open until a human or fresh trigger recovers it. Do not re-run a census-only row automatically: it has no failure record showing which jobs are safe to replay.
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -211,19 +211,21 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
 A run that fails files a row on a `tend-outage`-labelled **"Bot temporarily unavailable"** issue, naming the run and the trigger it stranded. Nothing re-runs those triggers — `tend-review` fires only on `pull_request_target`, so a PR whose one review attempt died stays unreviewed until someone pushes. Drain the open issue as part of this sweep.
 
 ```bash
-# Usually empty. An empty issue is not evidence nothing was stranded — see below.
+# Usually empty. Runs can fail before filing this issue; see the census below.
 OUTAGE=$(gh issue list --state open --label tend-outage --json number,title \
   --jq '[.[] | select(.title == "Bot temporarily unavailable")][0].number // empty')
 gh issue view "$OUTAGE" --json body,comments --jq '.body, .comments[].body' \
   | grep -oE 'runs/[0-9]+|\| #[0-9]+'
 ```
 
-**The census is the second drain input.** The outage row is filed by a step *inside* the failing job, so a run that died before reaching it files nothing and uploads no session log: a `startup_failure` runs zero steps, and a job that never gets a runner, a wedged `actions/checkout`, or a failing `setup:` step die earlier still. They strand the same triggers, and nothing else on the record names them. Step 1's census already carries every run's `conclusion`, so filter that rather than re-fetching:
+**The census is the second drain input.** A run that dies before the failure handler can strand its trigger without filing an outage row or uploading a session log. Step 1 already has the needed conclusions:
 
 ```bash
 # `skipped` is an `if:` gate declining to run, not a failure.
 jq -c 'select(.conclusion != "success" and .conclusion != "skipped")'
 ```
+
+For each census failure absent from the outage issue, confirm the trigger's work is still missing. Treat a cancelled run as a designed eviction when another run answered the same thread. Report a stranded trigger as a finding, and leave any outage issue open until a human or fresh trigger recovers it. Do not re-run a census-only row automatically: it has no failure record showing which jobs are safe to replay.
 
 **Diagnose first.** The nightly enrichment comment names the cause when it can. When it doesn't, read the session log — quota exhaustion surfaces as a `<synthetic>` assistant message:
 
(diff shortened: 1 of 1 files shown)


plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -211,13 +211,23 @@ After retrieving the timeout cap from the workflow file, flag any job whose dura
...
 ```
 
+**The census is the second drain input.** A run that dies before the failure handler can strand its trigger without filing an outage row or uploading a session log. Step 1 already has the needed conclusions:
+
+```bash
+# `skipped` is an `if:` gate declining to run, not a failure.
+jq -c 'select(.conclusion != "success" and .conclusion != "skipped")' \
+  /tmp/review-runs-census.jsonl
+```
+
```
tend-agent, line 228: `leave any outage issue open until a human or fresh trigger recovers it` collides with the section's last line, which is unchanged: `Close the issue once every row is drained (gh issue close "$OUTAGE"); one left open folds the next outage into a stale incident.`

A census-only stranding is by definition *absent* from the outage issue, so it puts no row on it. Sweep finds the outage issue with two rows, both drained, plus one `startup_failure` triage run in the census whose issue is still untriaged: "every row is drained" is true and the close fires, while this sentence says hold the issue open. Whichever the session picks is arbitrary, and the hold condition — "until a human or fresh trigger recovers it" — is one nothing in the sweep observes, so a held-open issue reaches the next sweep with no row explaining why it's open.

Scoping the hold to the finding instead of the issue keeps the report-only shape and removes the collision:

```suggestion
For each census failure absent from the outage issue, confirm the trigger's work is still missing. Treat a cancelled run as a designed eviction when another run answered the same thread. Report a stranded trigger as a finding; a census-only failure is not a row on the outage issue and does not hold its close. Do not re-run a census-only row automatically: it has no failure record showing which jobs are safe to replay.
```

The alternative — filing the census-only row *onto* the outage issue — also works and needs no new rule, since the existing drain-then-close then covers it. Either is fine; two rules disagreeing is not.

tend-agent (commented):
Reviewed at `df46ea9`, which rewrote the paragraph I'd drafted against. Detection is the right half to keep: `Report failure` is a step inside the composite action (`claude/action.yaml`, `codex/action.yaml`, both gated on `failure() && steps.security.outcome != 'failure'`), so a job that never reaches it files nothing, and `status=completed` in Step 1's census does return those runs. Restoring `--failed` is consistent now that census-only rows are explicitly not re-run — an outage row only exists because a job ran and failed, so it always has a failed job to select.

One finding, inline: the new paragraph and the section's closing line give conflicting terminal instructions for the outage issue.

Also, the description is now stale against this head — it still describes the re-run paragraph ("states the one thing a session must not conclude from a re-run: that exit 0 means the row is drained", "report the row on the outage issue and leave it open"), the dropped `--failed`, and a `gh run view <id> --json status,conclusion,attempt` recipe that no longer appears in the diff. Worth rewriting to the report-only shape that actually landed, including that the census-only path now detects but deliberately doesn't recover — #1066's own suggested fix asked for the re-run, so `Closes #1066` is claiming a narrower fix than the issue requested.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
max-sixty: Applied. Census-only failures are findings and no longer affect the outage issue's existing row-based close rule.

PR #1074, opened 2026-08-28
fix(notifications): recover missed work from unread queue

max-sixty:
Failed event workflows already leave their triggering notification unread, but that only provides recovery when the bot was subscribed before the event. The notification poll also read just the first page and acknowledged individual threads before deciding whether their work had a real outcome.

This makes the recovery model explicit and small:

- installation and every notification poll idempotently enable repository watching;
- the pre-check captures every unread page before a ten-minute cutoff and does not acknowledge anything;
- polls are serialized, process each thread from current repository state, and advance the repository acknowledgement only through the completed prefix before the first unresolved item;
- exact in-flight and issue cross-reference checks prevent duplicate bot work;
- the daily review-runs workflow scans current open issues, live PR heads and bot-directed replies, and failing default-branch CI as a slow backstop for delayed or missed notifications.

Historical workflow runs and the `tend-outage` issue remain diagnostic evidence. They are no longer replayed for recovery, so Tend always uses its current workflow and the subject's current state. After its exhaustive live scan, the daily sweep reads every row from the canonical outage tracker and closes it only after the applicable live work is handled.

The independent test review caught a lossy one-shot reconciliation flag in the first draft, and hosted review caught overlap, partial-progress, deduplication, unreadable-subject, and outage-lifecycle gaps. The final design has no new durable state: watching is enforced every poll, unread notifications are the recovery queue, and the live-state scan recurs every day. Contract and integration tests cover pagination, the strict cutoff boundary, serialized polling, bounded acknowledgement, permanently unreadable subjects, fragile deduplication queries, install-time watching, and the current-state outage lifecycle.

Follow-up to #1073; implements the recovery behind #1066 and #1067.

Verification: `wt test` (723 tests plus worker tests and typecheck), `pre-commit run --all-files`.

### DIFF
--- a/generator/tests/test_notification_queue_contract.py
+++ b/generator/tests/test_notification_queue_contract.py
@@ -18,13 +18,38 @@ def test_notification_skill_uses_one_paginated_cutoff_snapshot() -> None:
     assert "sort_by(.updated_at)" in skill
 
 
-def test_notification_skill_acknowledges_without_racing_new_activity() -> None:
+def test_notification_skill_uses_a_bounded_repository_acknowledgement() -> None:
     skill = _read("plugins", "tend-ci-runner", "skills", "notifications", "SKILL.md")
 
     assert "repos/$GITHUB_REPOSITORY/notifications" in skill
-    assert 'last_read_at="$CUTOFF"' in skill
+    assert 'last_read_at="$ACK_CUTOFF"' in skill
+    assert '"$UNRESOLVED_AT -1 second"' in skill
     assert "Never acknowledge a same-repository thread individually." in skill
     assert "notifications/threads/<thread-id>" in skill
+
+
+def test_notification_skill_pins_the_fragile_dedup_queries() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "notifications", "SKILL.md")
+
+    assert ".display_title == $title" in skill
+    assert "issues/$NUMBER/timeline?per_page=100" in skill
+    assert '.event == "cross-referenced"' in skill
+
+
+def test_review_runs_pins_current_state_recovery() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
+
+    assert "--state open --label tend-outage --author @me --limit 100" in skill
+    assert "| sort | .[0] // empty" in skill
+    assert "> /tmp/review-runs-outage-number" in skill
+    assert "> /tmp/review-runs-outage-initial.json" in skill
+    assert "> /tmp/review-runs-outage-final.json" in skill
+    assert 'gh issue close "$OUTAGE" --reason completed' in skill
+    assert "complete **Reconcile live work** below, then exit" in skill
+    assert "Do not replay historical workflow runs" in skill
+    assert "an open issue with no bot response to the latest human activity" in skill
+    assert "whose live head has no bot review" in skill
+    assert "failing default-branch CI with no bot fix in progress" in skill
 
 
 def test_installation_and_each_poll_enable_repository_watching() -> None:
@@ -36,11 +61,3 @@ def test_installation_and_each_poll_enable_repository_watching() -> None:
             "repos/$GITHUB_REPOSITORY/subscription" in content
         )
         assert "-F subscribed=true -F ignored=false" in content
-
-
-def test_review_runs_reconciles_current_state_without_historical_reruns() -> None:
-    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
-
-    assert "Reconcile live work" in skill
-    assert "current-state scan" in skill
-    assert "gh run rerun" not in skill
--- a/generator/tests/test_notification_queue_contract.py
+++ b/generator/tests/test_notification_queue_contract.py
@@ -24,6 +24,8 @@ def test_notification_skill_uses_a_bounded_repository_acknowledgement() -> None:
     assert "repos/$GITHUB_REPOSITORY/notifications" in skill
     assert 'last_read_at="$ACK_CUTOFF"' in skill
     assert '"$UNRESOLVED_AT -1 second"' in skill
+    assert 'if [ -n "$UNRESOLVED_AT" ]; then' in skill
+    assert "else\n  ACK_CUTOFF=$CUTOFF" in skill
     assert "Never acknowledge a same-repository thread individually." in skill
     assert "notifications/threads/<thread-id>" in skill
 
@@ -39,17 +41,24 @@ def test_notification_skill_pins_the_fragile_dedup_queries() -> None:
 def test_review_runs_pins_current_state_recovery() -> None:
     skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
 
-    assert "--state open --label tend-outage --author @me --limit 100" in skill
+    assert "--state open --label tend-outage --author @me" in skill
     assert "| sort | .[0] // empty" in skill
-    assert "> /tmp/review-runs-outage-number" in skill
-    assert "> /tmp/review-runs-outage-initial.json" in skill
-    assert "> /tmp/review-runs-outage-final.json" in skill
-    assert 'gh issue close "$OUTAGE" --reason completed' in skill
+    assert "if ! OUTAGE=$(gh issue list" in skill
+    assert "--json body,comments --jq '.body, .comments[].body'" in skill
+    assert "gh issue close <outage-number> --reason completed" in skill
     assert "complete **Reconcile live work** below, then exit" in skill
     assert "Do not replay historical workflow runs" in skill
     assert "an open issue with no bot response to the latest human activity" in skill
     assert "whose live head has no bot review" in skill
     assert "failing default-branch CI with no bot fix in progress" in skill
+
+
+def test_unreadable_notification_subjects_are_terminal() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "notifications", "SKILL.md")
+
+    assert "whose `subject.url` is null" in skill
+    assert "a deleted issue or PR" in skill
+    assert "also has the outcome “no action”" in skill
 
 
 def test_installation_and_each_poll_enable_repository_watching() -> None:
--- a/generator/tests/test_notification_queue_contract.py
+++ b/generator/tests/test_notification_queue_contract.py
@@ -46,7 +46,11 @@ def test_review_runs_pins_current_state_recovery() -> None:
     assert "if ! gh issue list" in skill
     assert "> /tmp/review-runs-outage-number; then" in skill
     assert "--json body,comments --jq '.body, .comments[].body'" in skill
-    assert '[ -n "$OUTAGE" ] && gh issue close "$OUTAGE" --reason completed' in skill
+    close_block = (
+        "OUTAGE=$(cat /tmp/review-runs-outage-number)\n"
+        '[ -n "$OUTAGE" ] && gh issue close "$OUTAGE" --reason completed'
+    )
+    assert close_block in skill
     assert "complete **Reconcile live work** below, then exit" in skill
     assert "Do not replay historical workflow runs" in skill
     assert "an open issue with no bot response to the latest human activity" in skill
--- a/generator/tests/test_notification_queue_contract.py
+++ b/generator/tests/test_notification_queue_contract.py
@@ -1,0 +1,46 @@
+"""Pin the notification queue's cross-file safety contract."""
+
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+
+
+def _read(*parts: str) -> str:
+    return REPO_ROOT.joinpath(*parts).read_text()
+
+
+def test_notification_skill_uses_one_paginated_cutoff_snapshot() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "notifications", "SKILL.md")
+
+    assert "notifications?before=$CUTOFF&per_page=100" in skill
+    assert "--paginate --slurp" in skill
+    assert "sort_by(.updated_at)" in skill
+
+
+def test_notification_skill_acknowledges_without_racing_new_activity() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "notifications", "SKILL.md")
+
+    assert "repos/$GITHUB_REPOSITORY/notifications" in skill
+    assert 'last_read_at="$CUTOFF"' in skill
+    assert "Never acknowledge a same-repository thread individually." in skill
+    assert "notifications/threads/<thread-id>" in skill
+
+
+def test_installation_and_each_poll_enable_repository_watching() -> None:
+    install = _read("plugins", "install-tend", "skills", "install-tend", "SKILL.md")
+    precheck = _read("generator", "src", "tend", "templates", "notifications-check.sh")
+
+    for content in (install, precheck):
+        assert "repos/$REPO/subscription" in content or (
+            "repos/$GITHUB_REPOSITORY/subscription" in content
+        )
+        assert "-F subscribed=true -F ignored=false" in content
+
+
+def test_review_runs_reconciles_current_state_without_historical_reruns() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
+
+    assert "Reconcile live work" in skill
+    assert "current-state scan" in skill
+    assert "gh run rerun" not in skill
--- a/generator/tests/test_notification_queue_contract.py
+++ b/generator/tests/test_notification_queue_contract.py
@@ -43,9 +43,10 @@ def test_review_runs_pins_current_state_recovery() -> None:
 
     assert "--state open --label tend-outage --author @me" in skill
     assert "| sort | .[0] // empty" in skill
-    assert "if ! OUTAGE=$(gh issue list" in skill
+    assert "if ! gh issue list" in skill
+    assert "> /tmp/review-runs-outage-number; then" in skill
     assert "--json body,comments --jq '.body, .comments[].body'" in skill
-    assert "gh issue close <outage-number> --reason completed" in skill
+    assert '[ -n "$OUTAGE" ] && gh issue close "$OUTAGE" --reason completed' in skill
     assert "complete **Reconcile live work** below, then exit" in skill
     assert "Do not replay historical workflow runs" in skill
     assert "an open issue with no bot response to the latest human activity" in skill
@@ -53,12 +54,24 @@ def test_review_runs_pins_current_state_recovery() -> None:
     assert "failing default-branch CI with no bot fix in progress" in skill
 
 
+def test_outage_tracker_title_stays_in_sync() -> None:
+    title = "Bot temporarily unavailable"
+    reporter = _read("shared", "steps", "report_failure.py")
+    review_runs = _read(
+        "plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md"
+    )
+    ci_fix = _read("plugins", "tend-ci-runner", "skills", "ci-fix", "SKILL.md")
+
+    for content in (reporter, review_runs, ci_fix):
+        assert title in content
+
+
 def test_unreadable_notification_subjects_are_terminal() -> None:
     skill = _read("plugins", "tend-ci-runner", "skills", "notifications", "SKILL.md")
 
     assert "whose `subject.url` is null" in skill
-    assert "a deleted issue or PR" in skill
-    assert "also has the outcome “no action”" in skill
+    assert "whose `subject.url` 404s" in skill
+    assert "A read that fails for any other reason" in skill
 
 
 def test_installation_and_each_poll_enable_repository_watching() -> None:
--- a/plugins/tend-ci-runner/skills/notifications/SKILL.md
+++ b/plugins/tend-ci-runner/skills/notifications/SKILL.md
@@ -48,6 +48,7 @@ Process the snapshot oldest first. Read the live issue or PR and decide what it 
 - Otherwise use the normal live workflow: `/tend-ci-runner:triage` for an issue, `/tend-ci-runner:review` for an unreviewed PR head, or answer a comment or review thread that asks the bot for something.
 - A closed thread or a human conversation that needs nothing from the bot has the semantic outcome “no action”.
 - A non-conversational subject, such as a release or check suite, also has the outcome “no action”. Default-branch CI recovery belongs to the daily current-state scan.
+- A subject that cannot be read — a `Discussion`, whose `subject.url` is null, or a deleted issue or PR — also has the outcome “no action”. Nothing makes it readable on a later poll, so leaving it unresolved would pin the Step 4 cutoff permanently.
 
 Judge deduplication from current state, including bot reviews and bot-authored PRs that cross-reference an issue. The notification timestamp alone does not prove whether a response covered the activity.
 
@@ -63,7 +64,7 @@ IN_PROGRESS=$(gh api \
         | select(.id != $own and .display_title == $title)] | length')
 ```
 
-Issue deduplication includes bot-authored PRs that cross-reference the issue. A PR with `Refs #N` may be the bot's response even when it posted no issue comment:
+Issue deduplication includes bot-authored PRs that cross-reference the issue. A PR with `Refs #N` may be the bot's response even when it posted no issue comment. Pad the notification time by 60 seconds because GitHub's notification index can trail the event that produced it:
 
 ```bash
 BOT_LOGIN=$(gh api user --jq .login)
@@ -89,9 +90,13 @@ gh api "notifications/threads/<thread-id>" -X PATCH
 Set the acknowledgement cutoff from the oldest unresolved same-repository item. If every same-repository item has an outcome, use the snapshot cutoff. Otherwise set it to one second before the unresolved item's `updated_at`:
 
 ```bash
-ACK_CUTOFF=$CUTOFF
-# When the first unresolved item was updated at $UNRESOLVED_AT:
-ACK_CUTOFF=$(date -u -d "$UNRESOLVED_AT -1 second" +%Y-%m-%dT%H:%M:%SZ)
+# $UNRESOLVED_AT is the first unresolved same-repository item's `updated_at`,
+# empty when every same-repository item has an outcome.
+if [ -n "$UNRESOLVED_AT" ]; then
+  ACK_CUTOFF=$(date -u -d "$UNRESOLVED_AT -1 second" +%Y-%m-%dT%H:%M:%SZ)
+else
+  ACK_CUTOFF=$CUTOFF
+fi
 gh api "repos/$GITHUB_REPOSITORY/notifications" -X PUT \
   -f last_read_at="$ACK_CUTOFF" --silent
 ```
--- a/plugins/tend-ci-runner/skills/notifications/SKILL.md
+++ b/plugins/tend-ci-runner/skills/notifications/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: notifications
-description: Drains the bot's unread GitHub notifications as a durable queue for issue and PR work that event workflows did not finish. Runs on a schedule.
+description: Drains the bot's unread GitHub notifications as a recovery queue for issue and PR work that event workflows did not finish. Runs on a schedule.
 metadata:
   internal: true
 ---
@@ -43,13 +43,40 @@ For each notification, identify the activity that made the thread unread and app
 
 Process the snapshot oldest first. Read the live issue or PR and decide what it needs now:
 
-- If a dedicated tend workflow is still running for the subject, defer it. One deferred same-repository thread defers the repository acknowledgement in Step 4.
+- If a dedicated tend workflow is still running for the subject, defer it. Its `updated_at` limits the repository acknowledgement in Step 4.
 - If the bot already handled the latest activity, record it as handled without posting again.
 - Otherwise use the normal live workflow: `/tend-ci-runner:triage` for an issue, `/tend-ci-runner:review` for an unreviewed PR head, or answer a comment or review thread that asks the bot for something.
 - A closed thread or a human conversation that needs nothing from the bot has the semantic outcome “no action”.
 - A non-conversational subject, such as a release or check suite, also has the outcome “no action”. Default-branch CI recovery belongs to the daily current-state scan.
 
 Judge deduplication from current state, including bot reviews and bot-authored PRs that cross-reference an issue. The notification timestamp alone does not prove whether a response covered the activity.
+
+For a same-repository item, check whether a dedicated workflow is still handling its subject. Match `display_title` because `workflow_run` does not expose the issue number for comment and review events. Pipe to standalone `jq`; `gh api --jq` cannot take `--arg` or `--argjson`:
+
+```bash
+SUBJECT_TITLE=$(gh api "$SUBJECT_URL" --jq .title)
+IN_PROGRESS=$(gh api \
+  "repos/$GITHUB_REPOSITORY/actions/runs?status=in_progress&per_page=100" \
+  | jq --arg title "$SUBJECT_TITLE" --argjson own "$GITHUB_RUN_ID" \
+      '[.workflow_runs[]
+        | select(.name | startswith("tend-"))
+        | select(.id != $own and .display_title == $title)] | length')
+```
+
+Issue deduplication includes bot-authored PRs that cross-reference the issue. A PR with `Refs #N` may be the bot's response even when it posted no issue comment:
+
+```bash
+BOT_LOGIN=$(gh api user --jq .login)
+DEDUP_CUTOFF=$(date -u -d "$NOTIF_UPDATED_AT -60 seconds" +%Y-%m-%dT%H:%M:%SZ)
+gh api "repos/$GITHUB_REPOSITORY/issues/$NUMBER/timeline?per_page=100" \
+  --paginate --slurp \
+  | jq --arg bot "$BOT_LOGIN" --arg cutoff "$DEDUP_CUTOFF" \
+      '[add // [] | .[]
+        | select(.event == "cross-referenced"
+          and .source.issue.pull_request
+          and .source.issue.user.login == $bot
+          and .created_at > $cutoff)] | length'
+```
 
 After a cross-repository notification has an outcome, acknowledge that thread individually:
 
@@ -59,13 +86,16 @@ gh api "notifications/threads/<thread-id>" -X PATCH
 
 ## 4. Acknowledge the cutoff
 
-Only after every same-repository notification in the snapshot has an outcome, acknowledge repository activity before the snapshot cutoff:
+Set the acknowledgement cutoff from the oldest unresolved same-repository item. If every same-repository item has an outcome, use the snapshot cutoff. Otherwise set it to one second before the unresolved item's `updated_at`:
 
 ```bash
+ACK_CUTOFF=$CUTOFF
+# When the first unresolved item was updated at $UNRESOLVED_AT:
+ACK_CUTOFF=$(date -u -d "$UNRESOLVED_AT -1 second" +%Y-%m-%dT%H:%M:%SZ)
 gh api "repos/$GITHUB_REPOSITORY/notifications" -X PUT \
-  -f last_read_at="$CUTOFF" --silent
+  -f last_read_at="$ACK_CUTOFF" --silent
 ```
 
-If any item was deferred or could not be checked, do not acknowledge the repository cutoff. The next poll will see the same queue; current-state deduplication makes the retry safe. Never acknowledge a same-repository thread individually.
+Skip the call when the unresolved item is the first same-repository item in the snapshot, or when the snapshot contains no same-repository items. The next poll starts with the unresolved item instead of repeating completed work. Never acknowledge a same-repository thread individually.
 
 Report the notifications handled, responses posted, items deferred, and whether the repository cutoff was acknowledged.
--- a/plugins/tend-ci-runner/skills/notifications/SKILL.md
+++ b/plugins/tend-ci-runner/skills/notifications/SKILL.md
@@ -1,157 +1,71 @@
 ---
 name: notifications
-description: Polls GitHub notifications and handles items that dedicated workflows miss — fork PR comments, cross-repo mentions, and stale unanswered items. Runs on a schedule.
+description: Drains the bot's unread GitHub notifications as a durable queue for issue and PR work that event workflows did not finish. Runs on a schedule.
 metadata:
   internal: true
 ---
 
 # Check Notifications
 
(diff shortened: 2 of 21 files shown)


tend-agent (commented):
The tests are unusually thorough for a skill rewrite and the pre-check itself is clean — I ran it under `bash -e` with a `gh` stand-in that fails the subscription PUT and it warns and continues to a correct `count`, and shellcheck is clean. My concerns are all on the acknowledgement path: what happens when the queue *doesn't* drain in one session. Five inline, plus one that has nowhere to anchor:

**Nothing serializes `tend-notifications` any more.** The workflow has no `concurrency:` block (unlike review/mention/triage), and what kept two overlapping runs off each other's work was old step 4c marking each thread read *as it was processed* — a second run's `gh api notifications` simply didn't return them. All acknowledgement now happens in one write at the very end, so an overlapping run reads the byte-identical snapshot and re-walks it from the same end. Current-state dedup catches threads the first run already posted on, but two runs walking one oldest-first list at similar pace sit on the *same* thread, which is the one dedup can't see yet — and it doesn't see outcomes that aren't comments (a pushed commit, an opened PR) at all. Recent `tend-notifications` runs on this repo already reach 8 and 11½ minutes against a 15-minute cron, before whole-repo watching and before the agent started invoking triage/review per thread. This is a duplicate-outward-action path rather than wasted compute, and the remedy is one knob in one place — `concurrency: { group: tend-notifications }` in `notifications.yaml.j2`, at job level and so outside this diff's hunk. Say the word and I'll push it.

Two of the inline points reach files this PR doesn't touch (`shared/steps/report_failure.py`, `plugins/tend-ci-runner/skills/nightly/SKILL.md`). Happy to carry those in one commit once you've decided where the outage-issue close should live.

<details><summary>Checked, not raising</summary>

- The `add // []` page validation is right across every `gh api --paginate --slurp` shape I could construct: zero pages, one empty page, several pages, an HTML body, and a JSON error object all land where intended — the last two exit 5, so the `else` branch fires. `-e` doesn't misfire on `[]`.
- `before` vs `last_read_at` boundary: GitHub documents neither as explicitly inclusive or exclusive, and the two readings are mutually consistent, so I'm not treating an item stamped exactly at the cutoff as a hole.
- Subscription scope and placement: the install skill requests `repo` for every adopter, not `public_repo`, so the PUT has what it needs, and it runs as the bot after the invitation is accepted.
- Dropping the 3-attempt retry loop and the `gh api -i` diagnostic is aligned with CLAUDE.md — a lost cycle is one tick, and the `::warning::` still leaves a trace.
- Cross-repo notifications inside the cutoff now wait a cycle where the old freshness gate exempted them: ~25 min instead of ~15 for a cross-repo mention. Not worth machinery.
- Unverified, so not a finding: `shared/steps/mark_notification_read.py` fetches `gh api notifications` unpaginated, and `threads_to_mark` wants a thread older than `run_started_at`. That only misses when more than a page of notifications arrives *during* one run — whole-repo watching raises that traffic, but I couldn't show the threshold is reachable here.

</details>

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -1,157 +1,71 @@
...
-gh api "repos/{owner}/{repo}/issues/{number}/timeline" \
-  --jq "[.[] | select(.event == \"cross-referenced\"
-    and .source.issue.pull_request
-    and .source.issue.user.login == \"$BOT_LOGIN\"
-    and .created_at > \"$DEDUP_CUTOFF\")] | length"
-```
+## 4. Acknowledge the cutoff
 
-If any of the three returns `> 0`, mark read and move on:
+Only after every same-repository notification in the snapshot has an outcome, acknowledge repository activity before the snapshot cutoff:
```
tend-agent: This acknowledges by *time range*, not by the ids the run examined — `PUT /repos/{o}/{r}/notifications` clears every repository thread at or below the cutoff, including ones that were never in the snapshot. Old step 4c PATCHed only threads it had processed, so an item the fetch missed survived to the next poll by construction; that property is gone.

The reachable case is delivery lag. The code this PR deletes documented notification indexing running 20-60s behind the event, and a degradation runs longer than that — anything delivered later than its own 10-minute grace lands after the snapshot and before the cutoff, so it is marked read unseen. What makes that permanent rather than merely late is that nothing else looks: `review-runs`' new backstop is open issue / open PR head / red default-branch CI, and a fork-PR inline comment fits none of them. That shape was this skill's flagship case ("`pull_request_review_comment` events don't fire for the bot on fork PRs, so no other workflow picks these up"), and `docs/security-model.md` still states that a fork PR's reviews reach the bot *only* through this poll.

Acking the snapshot's ids rather than a time range would keep the old guarantee; if the single write is worth keeping, the fork-PR shape probably needs its own bullet in the `review-runs` scan.

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -1,157 +1,71 @@
...
-After processing (whether or not a response was posted):
-
-```bash
-gh api notifications/threads/{thread_id} -X PATCH
-```
-
-## Step 5: Summary
-
-Report what was processed:
+If any item was deferred or could not be checked, do not acknowledge the repository cutoff. The next poll will see the same queue; current-state deduplication makes the retry safe. Never acknowledge a same-repository thread individually.
```
tend-agent: One deferred thread throws away the whole queue's progress. Step 3 already works oldest-first, so the run knows exactly where its progress stops — but this is all-or-nothing, and there's no snapshot cap or session budget behind it.

`running-in-ci` says long `tend-mention` queues run "hours, not minutes", so one mention job sitting in that queue holds the *repository* cutoff back for its whole wait: every 15-minute poll in that window re-reads a strictly larger queue and acknowledges nothing. A backlog bigger than one session can finish does the same thing without any deferral at all, and it ratchets — each poll has more to redo than the last. A subject that can never be checked (a deleted issue, a `Discussion` with a null `subject.url`) makes it permanent. Nothing turns red throughout.

Advancing the cutoff to where progress actually stopped keeps both the no-race property and the "never individually" invariant:

```suggestion
If a same-repository item was deferred or could not be checked, do not discard the progress ahead of it: Step 3 works oldest-first, so set `last_read_at` to one second before that item's `updated_at` instead of `$CUTOFF`, and skip the call when it is the first item in the snapshot. The next poll re-reads only what is still unresolved; current-state deduplication makes the retry safe. Never acknowledge a same-repository thread individually.
```

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -1,157 +1,71 @@
...
+- In another repository, respond only to a direct, straightforward mention. Do not push code or modify an existing PR there; new issues follow **Other Repos** in `running-in-ci`.
 
-For each unread notification (oldest first):
+## 3. Give each thread a current outcome
 
-### 4a. Freshness gate and dedup check
+Process the snapshot oldest first. Read the live issue or PR and decide what it needs now:
 
-**Freshness gate (same-repo only):** Same-repo notifications younger than 10 minutes are likely being handled by a concurrent dedicated workflow (`tend-review`, `tend-mention`, etc.) that hasn't posted its response yet. **Skip** these — do not process, do not mark read. The next scheduled run will pick them up once the grace period has elapsed and the dedicated workflow has either succeeded or failed.
+- If a dedicated tend workflow is still running for the subject, defer it. One deferred same-repository thread defers the repository acknowledgement in Step 4.
```
tend-agent: This judgment gained repository-wide authority — it gates the Step 4 acknowledgement for every thread, not just this one — at the same time as it lost its recipe. The old step 4a carried two mechanics that aren't recoverable by inspection: match on `workflow_runs[].display_title`, *because* the `workflow_run` payload doesn't expose the triggering issue number for `issue_comment` / `pull_request_review` events; and `gh api --jq` won't accept `--arg`/`--argjson`, so pipe to standalone `jq`. An agent that gets that wrong either records "could not be checked" (line 69's wedge) or concludes nothing is in flight and posts a second review on a PR `tend-review` is mid-flight on.

Same shape one bullet down at line 52: the cross-reference category survives, but the `issues/{n}/timeline` `cross-referenced` query is gone, and that one existed specifically because `tend-mention` answers an @-mention with a PR carrying `Refs #N` and *no* issue comment — invisible to a comments-and-reviews dedup, and a duplicate PR if missed. CLAUDE.md's own authoring rule keeps these: "Reserve mandatory exact wording for fragile mechanics: dedup keys, commands, API formats."

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -206,47 +206,17 @@ gh api "repos/$REPO/actions/runs/$RUN_ID/jobs" \
...
-gh pr view <n> --json state,headRefOid,reviews \
-  --jq '{state, headRefOid, reviewers: [.reviews[].author.login]}'
-gh run rerun <run-id> --failed
-```
+- an open issue with no bot response to the latest human activity;
+- an open PR whose live head has no bot review, or whose latest maintainer request has no response;
+- failing default-branch CI with no bot fix in progress.
 
-Close the issue once every row is drained (`gh issue close "$OUTAGE"`); one left open folds the next outage into a stale incident.
+Handle live work through the normal triage, review, or CI-fix guidance. Keep failed runs in the report as diagnostic evidence. Close the outage issue once its rows are diagnosed and the current-state scan has handled any work that still applies.
```
tend-agent: "Close the outage issue" survived the rewrite but its mechanics didn't — the removed block carried both the `OUTAGE=$(gh issue list --state open --label tend-outage ...)` lookup and the `gh issue close "$OUTAGE"`. Two files outside this diff still describe the deleted step, and one of them is live on GitHub:

- `shared/steps/report_failure.py` writes `OUTAGE_BODY` into **every** `tend-outage` issue: "Close it once every row above is drained — each stranded trigger re-run, or confirmed no longer needed. The `tend-review-runs` sweep does that where it runs." A `review-runs` session drains against the issue body it reads, and that body now tells it to do the exact thing line 211 forbids. Its module docstring describes the same drain.
- `plugins/tend-ci-runner/skills/nightly/SKILL.md` still declines to close the issue, deferring to "`review-runs`' drain step, which reports each row re-run or confirmed no longer needed" — a step this PR renamed and whose re-run semantics it removed.

So nightly won't close it by rule, review-runs no longer has the recipe to, and the issue text argues against the new policy. Left alone rather than guessing which side you want the close on.

generator/tests/test_notification_queue_contract.py
```diff
@@ -0,0 +1,46 @@
...
+        )
+        assert "-F subscribed=true -F ignored=false" in content
+
+
+def test_review_runs_reconciles_current_state_without_historical_reruns() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
+
+    assert "Reconcile live work" in skill
+    assert "current-state scan" in skill
+    assert "gh run rerun" not in skill
```
tend-agent: This is the repo's first test to assert SKILL.md prose, and the assertions aren't all the same kind. `test_installation_and_each_poll_enable_repository_watching` pins a real cross-file invariant — two files that have to agree or watching silently never gets enabled. These three are weaker: `"Reconcile live work"` and `"current-state scan"` are a heading and a phrase that a reword breaks with no safety lost, and `not in skill` forbids the substring anywhere in that file forever, so a future paragraph documenting a maintainer-initiated rerun fails here with a confusing message. The executable coverage that would catch a real regression is already next door in `test_shared_steps.py` and `test_integration.py`; worth keeping this file to the assertions that guard behaviour across files.

plugins/tend-ci-runner/skills/notifications/SKILL.md
max-sixty: Fixed in 7c5ac5c5. Notification polls are now serialized, and repository acknowledgement stops one second before the oldest unresolved item. A daily live-state scan covers delayed notifications, including bot-directed review and inline-comment replies on fork PRs.

plugins/tend-ci-runner/skills/notifications/SKILL.md
max-sixty: Fixed in 7c5ac5c5. One unresolved item no longer wedges the whole queue: the poll advances `last_read_at` through the completed prefix, stopping one second before the first unresolved item so the next poll resumes there.

plugins/tend-ci-runner/skills/notifications/SKILL.md
max-sixty: Fixed in 7c5ac5c5. The exact `display_title` in-flight lookup and bot-authored issue cross-reference timeline query are restored, with contract tests pinning both fragile recipes.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
max-sixty: Fixed in 7c5ac5c5. `report_failure`, nightly, and review-runs now agree that outage rows are diagnostic rather than replay instructions. Review-runs selects the canonical bot-authored tracker, reads its body and comments, checks current repository state, rereads for rows appended during the scan, and closes the exact tracker after applicable work is handled.

generator/tests/test_notification_queue_contract.py
max-sixty: Fixed in 7c5ac5c5. I removed the heading/global-substring assertions and kept exact cross-file contracts: generated concurrency, bounded acknowledgement API shape, dedup queries, and the outage tracker lookup/read/close lifecycle.

tend-agent:
Verified all five against `7c5ac5c5`. Four are resolved as claimed; the fifth landed the change I asked for but left the assertion I'd actually want in that file unpinned.

The bounded acknowledgement is the shape I suggested, and the two skip conditions (unresolved item is first same-repository item; no same-repository items at all) close the case where `-1 second` would step behind the snapshot. The restored recipes match the deleted originals semantically — `display_title` with `--argjson own` piped to standalone `jq`, and the `cross-referenced` timeline query — now paginated with the same `add // []` guard as the snapshot. The outage lifecycle agrees across files: the title literal equals [`report_failure.TITLE`](https://github.com/max-sixty/tend/blob/7c5ac5c5df1b869018d1b0210f405b49f7f0e7b7/shared/steps/report_failure.py#L44), `| sort | .[0]` reproduces [`_issue.canonical`](https://github.com/max-sixty/tend/blob/7c5ac5c5df1b869018d1b0210f405b49f7f0e7b7/shared/steps/_issue.py#L145)'s lowest-open rule, and the re-read narrows the append-during-scan window to the gap between the final read and the close. `OUTAGE_BODY`, nightly's carve-out, and the review-runs step now say the same thing. Concurrency is job-level with `cancel-in-progress: false`, pinned by the regtests and `test_init_notifications_has_precheck`, and the CLAUDE.md table and queue-depth paragraph both carry it. On the acknowledgement-by-time-range point: `review-runs` is default-enabled at `47 7 * * *`, so the fork-PR reply shape now has a named daily reader — a day's latency instead of never.

The one left: [`test_review_runs_pins_current_state_recovery`](https://github.com/max-sixty/tend/blob/7c5ac5c5df1b869018d1b0210f405b49f7f0e7b7/generator/tests/test_notification_queue_contract.py#L39) still pins single-file prose — the `**Reconcile live work**` heading reference and the three scan bullets verbatim — which is the class where a reword fails the test and nothing was protected. Meanwhile the cross-file contract in that same block isn't asserted: `"Bot temporarily unavailable"` is a third copy of a literal that also lives in `report_failure.py` and `ci-fix/SKILL.md`, and if it drifts the lookup returns empty, `[ -n "$OUTAGE" ]` skips every guard, and the whole read/diagnose/close lifecycle silently no-ops with the sweep still reporting success. Asserting `report_failure.TITLE in skill` (and keeping the `sort | .[0] // empty` assertion, which pins the `canonical` agreement) buys that; the prose bullets can go. Happy to push it if you want it in this PR.

tend-agent (commented):
Three on the new commit, none overlapping the verification comment above. The first is the one I'd want fixed before merge; the other two are cleanup.

**Step 4's recipe fails open when run verbatim.** Both `ACK_CUTOFF` assignments are unconditional, and the second wins. `date -u -d " -1 second"` is valid GNU date — with `$UNRESOLVED_AT` empty it returns *now* and exits 0, so an agent that copies the block as written acknowledges every same-repository thread up to the present, including the unresolved item it was meant to stop before and anything delivered after the snapshot. That is the loss the bounded prefix exists to prevent, and it is silent: no non-zero exit, no warning, and the contract test passes on either form. The inline suggestion makes the branch explicit and keeps both pinned substrings.

<details><summary>Checked, not raising</summary>

- `date -u -d "$UNRESOLVED_AT -1 second"` parses the `Z`-suffixed form correctly (`2026-08-28T04:00:00Z` → `2026-08-28T03:59:59Z`), and the PUT is repo-scoped, so cross-repository threads keep their individual-ack path untouched.
- The literal `group: tend-notifications`, where review/mention/triage interpolate `${{ github.workflow }}`: groups are repo-scoped and nothing else can claim the string, so the only cost is that a `workflow_extra` rename wouldn't follow the group. Not worth the interpolation.
- `--argjson own "$GITHUB_RUN_ID"` on an 11-digit run id stays exact in jq's double (well under 2^53), so the self-exclusion holds.
- The 60-second `DEDUP_CUTOFF` pad survived the rewrite but the paragraph justifying it — GitHub's notification indexing running 20–60 s behind the event — didn't. The number is still right; only the reason a future editor would need is gone.
- Serializing the poll trades overlap for availability: the notifications job sets no `timeout-minutes`, so a wedged poll holds the group for GitHub's 360-minute default while crons keep replacing one pending slot. I asked for this group and I'd still take it — no poll has been observed running near that, so a timeout knob has no evidence behind it yet.

</details>

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -1,157 +1,101 @@
...
-After processing (whether or not a response was posted):
+Set the acknowledgement cutoff from the oldest unresolved same-repository item. If every same-repository item has an outcome, use the snapshot cutoff. Otherwise set it to one second before the unresolved item's `updated_at`:
 
 ```bash
-gh api notifications/threads/{thread_id} -X PATCH
+ACK_CUTOFF=$CUTOFF
+# When the first unresolved item was updated at $UNRESOLVED_AT:
+ACK_CUTOFF=$(date -u -d "$UNRESOLVED_AT -1 second" +%Y-%m-%dT%H:%M:%SZ)
+gh api "repos/$GITHUB_REPOSITORY/notifications" -X PUT \
+  -f last_read_at="$ACK_CUTOFF" --silent
```
tend-agent: Both assignments run. As a recipe this reads "always compute the second one", and `date -u -d " -1 second"` with `$UNRESOLVED_AT` empty resolves to *now* and exits 0 — so a run with nothing unresolved acknowledges the whole repository up to the present, swallowing anything delivered after the snapshot. That is the failure the bounded prefix was added to stop, and it is silent: no non-zero exit, no warning, and `test_notification_skill_uses_a_bounded_repository_acknowledgement` passes on either form. CLAUDE.md's authoring rule puts this squarely in "mandatory exact wording for fragile mechanics: dedup keys, commands, API formats" — this is the single write that discards the recovery queue.

Making the branch explicit costs three lines and keeps both substrings the contract test pins:

```suggestion
# $UNRESOLVED_AT is the first unresolved same-repository item's `updated_at`,
# empty when every same-repository item has an outcome.
if [ -n "$UNRESOLVED_AT" ]; then
  ACK_CUTOFF=$(date -u -d "$UNRESOLVED_AT -1 second" +%Y-%m-%dT%H:%M:%SZ)
else
  ACK_CUTOFF=$CUTOFF
fi
gh api "repos/$GITHUB_REPOSITORY/notifications" -X PUT \
  -f last_read_at="$ACK_CUTOFF" --silent
```

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -1,157 +1,101 @@
...
 
-**Freshness gate (same-repo only):** Same-repo notifications younger than 10 minutes are likely being handled by a concurrent dedicated workflow (`tend-review`, `tend-mention`, etc.) that hasn't posted its response yet. **Skip** these — do not process, do not mark read. The next scheduled run will pick them up once the grace period has elapsed and the dedicated workflow has either succeeded or failed.
+- If a dedicated tend workflow is still running for the subject, defer it. Its `updated_at` limits the repository acknowledgement in Step 4.
+- If the bot already handled the latest activity, record it as handled without posting again.
+- Otherwise use the normal live workflow: `/tend-ci-runner:triage` for an issue, `/tend-ci-runner:review` for an unreviewed PR head, or answer a comment or review thread that asks the bot for something.
+- A closed thread or a human conversation that needs nothing from the bot has the semantic outcome “no action”.
+- A non-conversational subject, such as a release or check suite, also has the outcome “no action”. Default-branch CI recovery belongs to the daily current-state scan.
```
tend-agent, line 50: Step 4 now has no floor for an item that can never get an outcome. Under the old design every item was acknowledged regardless, so an unreadable subject cost one wasted look; now the first one pins `ACK_CUTOFF` at its `updated_at` and nothing ever moves it — every subsequent poll re-reads a strictly larger snapshot from that point and acknowledges nothing past it.

The reachable shape is a Discussion: `subject.url` is null on those notification records, so `gh api "$SUBJECT_URL"` has nothing to fetch and the item can't be checked on this poll or any later one. The base version of this skill named it explicitly ("Shapes not listed here — a fork PR's top-level comment, a discussion thread — are what this fallback exists for"), and repository-wide watching, which this PR now enables on every poll, is what starts delivering them. A deleted issue or PR is the same class.

One bullet in the outcome list closes it:

```suggestion
- A non-conversational subject, such as a release or check suite, also has the outcome “no action”. Default-branch CI recovery belongs to the daily current-state scan.
- A subject that cannot be read — a `Discussion`, whose `subject.url` is null, or a deleted issue or PR — also has the outcome “no action”. Nothing makes it readable on a later poll, so leaving it unresolved would pin the Step 4 cutoff permanently.
```

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -206,48 +206,61 @@ gh api "repos/$REPO/actions/runs/$RUN_ID/jobs" \
...
+- an open issue with no bot response to the latest human activity;
+- an open PR whose live head has no bot review, or whose latest comment, review, or inline review comment directed at the bot has no response; this includes replies to the bot's review on a fork PR;
+- failing default-branch CI with no bot fix in progress.
 
-**Diagnose first.** The nightly enrichment comment names the cause when it can. When it doesn't, read the session log — quota exhaustion surfaces as a `<synthetic>` assistant message:
+Handle live work through the normal triage, review, or CI-fix guidance. Keep
+failed runs in the report as diagnostic evidence. After every outage row is
+diagnosed and the current-state scan has handled any work that still applies,
+read the same tracker again immediately before closing it. This catches rows
```
tend-agent: The reconcile now reads the tracker, runs the scan, re-reads the tracker, then closes — three code blocks and a `/tmp` file handoff, to catch rows appended during the scan.

The initial read isn't load-bearing. The live scan below is exhaustive over repository state — every open issue, every open PR head, every bot-directed reply, default-branch CI — and is not scoped by the rows, so having read them first doesn't narrow what it covers. Reading the tracker once *here*, after the scan and immediately before the close, gives the same coverage in one block with no race window to catch.

That also drops `/tmp/review-runs-outage-number`, which currently conflates a failed `gh issue list` with "no tracker open" — the distinction `_issue.canonical` goes out of its way to keep ("'The read failed' and 'there is none' are different facts"). Here the conflation silently skips the whole outage step for the day.

Happy to push that restructure if you want it; it touches the three blocks and the two test assertions that pin the temp-file names.

plugins/tend-ci-runner/skills/notifications/SKILL.md
max-sixty: Fixed in 052c23fe. The recipe now branches explicitly: a non-empty `UNRESOLVED_AT` computes the bounded prefix, while an empty value uses only the original snapshot cutoff. The contract test pins that branch as well as the API fields.

plugins/tend-ci-runner/skills/notifications/SKILL.md
max-sixty: Fixed in 052c23fe. A Discussion with a null subject URL, or a deleted issue/PR, now receives the terminal “no action” outcome because retrying cannot make it readable. This prevents a permanently unreadable item from pinning the cutoff.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
max-sixty: Fixed in 052c23fe. The initial read and `/tmp` handoff are gone. Review-runs now performs the exhaustive live scan first, then does one canonical bot-authored tracker lookup/read, explicitly fails when the lookup API fails, diagnoses every row, and closes the exact returned issue number.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -206,48 +206,45 @@ gh api "repos/$REPO/actions/runs/$RUN_ID/jobs" \
...
-**Re-run only what won't recover.** Scheduled workflows (`nightly`, `notifications`, `weekly`, this one) recover on their next tick. For an event-triggered run, confirm the work is still missing — a later push often re-triggers it — and that a recent run completed cleanly, or the re-run just refills the issue. Re-running the bot's own failed workflow needs no maintainer approval.
+Use every row to identify what the failed run may have missed. Diagnose it and
+handle any applicable current work. If a tracker was found, close the exact
+issue number returned above:
 
 ```bash
-gh pr view <n> --json state,headRefOid,reviews \
-  --jq '{state, headRefOid, reviewers: [.reviews[].author.login]}'
-gh run rerun <run-id> --failed
+gh issue close <outage-number> --reason completed
```
```suggestion
[ -n "$OUTAGE" ] && gh issue close "$OUTAGE" --reason completed
```

Bash reads `<outage-number>` as `< outage-number`, so this block errors out on the redirect and `gh` never runs — the tracker stays open, accumulates every subsequent `report_failure` row, and each later sweep re-diagnoses the lot. `$OUTAGE` is set by the block above, so the executable form also puts the "if a tracker was found" guard back in the code where it can't be skipped. Pairs with the assertion change in `test_notification_queue_contract.py`.

generator/tests/test_notification_queue_contract.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def test_review_runs_pins_current_state_recovery() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
+
+    assert "--state open --label tend-outage --author @me" in skill
+    assert "| sort | .[0] // empty" in skill
+    assert "if ! OUTAGE=$(gh issue list" in skill
+    assert "--json body,comments --jq '.body, .comments[].body'" in skill
+    assert "gh issue close <outage-number> --reason completed" in skill
```
tend-agent:
```suggestion
    assert 'gh issue close "$OUTAGE" --reason completed' in skill
```

Follows the recipe fix in `review-runs/SKILL.md`. Worth noting on its own terms too: pinning a fill-in placeholder means the test fails if anyone restores the executable form, which is the wrong direction for a contract test to push.

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -1,157 +1,106 @@
...
+- A subject that cannot be read — a `Discussion`, whose `subject.url` is null, or a deleted issue or PR — also has the outcome “no action”. Nothing makes it readable on a later poll, so leaving it unresolved would pin the Step 4 cutoff permanently.
```
tend-agent:
```suggestion
- A subject with no readable target — a `Discussion`, whose `subject.url` is null, or a deleted issue or PR, whose `subject.url` 404s — also has the outcome “no action”. Nothing makes it readable on a later poll, so leaving it unresolved would pin the Step 4 cutoff permanently. A read that fails for any other reason — a 5xx, a rate limit — leaves the item unresolved.
```

"A subject that cannot be read" is broader than the two cases behind it, and the outcome it assigns is terminal: Step 4's repository-wide PUT acknowledges the item, so it never returns to the queue. A `gh api "$SUBJECT_URL"` that fails on a 5xx or a secondary rate limit reads as "cannot be read" under the current wording, and the thread is then dropped unexamined — the same silent loss the bounded prefix was added to prevent, just reached through the outcome rather than the cutoff. Keying on the structural condition (null `subject.url`, or a 404) keeps the terminal call to cases that really are terminal. Both substrings the new test pins are preserved.

generator/tests/test_notification_queue_contract.py
```diff
@@ -0,0 +1,72 @@
...
+    assert "whose live head has no bot review" in skill
+    assert "failing default-branch CI with no bot fix in progress" in skill
+
+
+def test_unreadable_notification_subjects_are_terminal() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "notifications", "SKILL.md")
+
+    assert "whose `subject.url` is null" in skill
+    assert "a deleted issue or PR" in skill
+    assert "also has the outcome “no action”" in skill
```
tend-agent:
```suggestion
```

This one passes whether or not the new bullet exists — `also has the outcome “no action”` already appears on the preceding release/check-suite bullet, which the diff doesn't touch. The two asserts above it carry the pin.

tend-agent (commented):
Two on `052c23f`. The first is the one I'd want fixed before merge; the second narrows a phrase.

**The outage close can't run as written.** `gh issue close <outage-number> --reason completed` sits in a bash block, and bash parses `<outage-number>` as a stdin redirect — the command aborts with `outage-number: No such file or directory` and exits 1 before `gh` is invoked. `$OUTAGE` is already in scope two lines above, and the `[ -n "$OUTAGE" ]` guard that used to bound the close moved into prose. The cost when it doesn't fire: `report_failure.py` appends every later row to the same tracker, and [`_issue.canonical`](https://github.com/max-sixty/tend/blob/052c23fed1414485afdee7160283df8ec5f37163/shared/steps/_issue.py#L145) resolves to the lowest-numbered open one, so the issue grows without bound and each daily sweep re-reads and re-diagnoses its whole accumulated history while still reporting success. CLAUDE.md's authoring rule — "Prefer recipe over narrative", and exact wording reserved "for fragile mechanics: dedup keys, commands, API formats" — points the same way. The test assertion has to move with it, so that's a second inline.

<details><summary>Checked, not raising</summary>

- The Step 4 branch is right: `if [ -n "$UNRESOLVED_AT" ]` / `else ACK_CUTOFF=$CUTOFF` closes the fail-open path, and both arms are pinned by `test_notification_skill_uses_a_bounded_repository_acknowledgement`. I ran all six tests in this file against the head tree — green.
- The review-runs reorder does what I asked: one tracker read after the exhaustive live scan, no `/tmp` handoff, no second read. The `if ! OUTAGE=$(gh issue list ...)` guard distinguishing "read failed" from "none open" matches `_issue.canonical`'s own docstring on that distinction, so it's consistent rather than extra machinery.
- Dropping the `state == OPEN` re-check before the close is fine — `gh issue close` on an already-closed issue warns and exits 0.
- The 60-second `DEDUP_CUTOFF` rationale is back in the prose, which was the only thing missing from that recipe.
- `"Bot temporarily unavailable"` is still a third copy of a literal that also lives in `report_failure.py` and `ci-fix/SKILL.md`, and still unasserted — I raised that in the comment above and it stands as-is; not restating it as a new finding.

</details>

plugins/tend-ci-runner/skills/notifications/SKILL.md
tend-agent: Verified against `052c23fe`. The bullet at [`SKILL.md#L51`](https://github.com/max-sixty/tend/blob/052c23fed1414485afdee7160283df8ec5f37163/plugins/tend-ci-runner/skills/notifications/SKILL.md#L51) gives all three permanent shapes a terminal outcome, so none of them can pin `ACK_CUTOFF` past the poll that meets them, and `test_unreadable_notification_subjects_are_terminal` pins the wording. All six tests in that file pass against the head tree.

One note rather than a request: a 5xx or a secondary rate limit on `gh api "$SUBJECT_URL"` presents the same at the call site as a deleted subject, so classifying one of those as "cannot be read" would acknowledge an item that was only late. The bullet enumerates permanent shapes only, which is the right scope — I don't think it wants a status-code recipe on top.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
max-sixty: Fixed in c38f2d86. The post-scan canonical lookup now writes its successful result to a small `/tmp` file so the exact issue ID survives separate Bash tool calls. The close block reloads it and runs the executable guarded form: `[ -n "$OUTAGE" ] && gh issue close "$OUTAGE" --reason completed`.

generator/tests/test_notification_queue_contract.py
max-sixty: Fixed in c38f2d86. The contract now pins the executable guarded `$OUTAGE` close, the cross-call ID handoff, and the explicit lookup-failure branch. It also pins the outage title across `report_failure`, review-runs, and ci-fix.

plugins/tend-ci-runner/skills/notifications/SKILL.md
max-sixty: Fixed in c38f2d86. Terminal “no action” is now limited to a null `subject.url` or a confirmed 404 for a deleted issue/PR. A 5xx, rate limit, or any other read failure leaves the item unresolved for retry.

generator/tests/test_notification_queue_contract.py
max-sixty: Fixed in c38f2d86. I removed the redundant prose assertion and now pin the two structural terminal cases plus the explicit transient-failure retry rule.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Verified against `c38f2d86`. All four land, and the seven tests in `generator/tests/test_notification_queue_contract.py` pass against the head tree (ran the module directly — no `uv` on this runner). Repo CI on this head is green apart from the in-flight review job.

The `/tmp` handoff coming back is right, and my earlier objection to it doesn't apply to this form. The two blocks are separate Bash tool calls, so `$OUTAGE` set by the lookup genuinely can't reach the close — the file is the only thing that crosses. What I was objecting to was the conflation of "the read failed" with "no tracker is open", and the explicit `if ! … exit 1` branch keeps those apart.

One nit on the line I suggested, which the contract test now pins. `[ -n "$OUTAGE" ] && gh issue close …` is the block's last command, so on a day with no open tracker — the common day — the block exits 1. Checked with a `gh` stand-in: tracker found → exit 0, none open → exit 1, lookup failed → the first block exits 1 and the second no-ops. Nothing closes wrongly; it just ends the sweep's last step on a non-zero call that reads as a failure. `if [ -n "$OUTAGE" ]; then gh issue close "$OUTAGE" --reason completed; fi` avoids it, with the assertion moving to match. Optional — mentioning it because the string is pinned now.

The other three read as claimed: the terminal-outcome bullet keys on conditions that are actually checkable at the call site (`gh api` on a deleted issue prints `gh: Not Found (HTTP 404)`, distinct from a 5xx or a rate limit, so those stay unresolved as intended); the two replacement assertions are unique to the new bullet rather than passing on the untouched one above it; and `test_outage_tracker_title_stays_in_sync` pins the title across the three files that have to agree on it.

generator/tests/test_notification_queue_contract.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_review_runs_pins_current_state_recovery() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
+
+    assert "--state open --label tend-outage --author @me" in skill
+    assert "| sort | .[0] // empty" in skill
+    assert "if ! gh issue list" in skill
+    assert "> /tmp/review-runs-outage-number; then" in skill
+    assert "--json body,comments --jq '.body, .comments[].body'" in skill
+    assert '[ -n "$OUTAGE" ] && gh issue close "$OUTAGE" --reason completed' in skill
```
tend-agent:
```suggestion
    close_block = (
        "OUTAGE=$(cat /tmp/review-runs-outage-number)\n"
        '[ -n "$OUTAGE" ] && gh issue close "$OUTAGE" --reason completed'
    )
    assert close_block in skill
```

This assertion pins the close line but not the `cat` that puts `$OUTAGE` in scope for it, so deleting the reload from that block keeps this test green while the close silently stops firing — the tracker then accumulates every later `report_failure` row and each sweep re-diagnoses the lot. Asserting the two lines as one string closes that and ties the `/tmp` path's read end to its write end above.

tend-agent (approved):
All three from `052c23f` are fixed, verified against `c38f2d8`: the close is executable and guarded, its assertion pins that form instead of the placeholder, and the terminal “no action” bullet is now keyed on the two structural shapes with a 5xx or rate limit left unresolved. The `/tmp` handoff is right despite my last comment arguing the other way — the lookup and the close are separate Bash calls, so `$OUTAGE` genuinely does not survive between them, and Step 2's `/tmp/review-runs-since` is the same pattern one section down. Approving.

One inline, optional: the new contract pins the write side of the handoff and the close line, but nothing pins that the close block reloads `$OUTAGE`. Dropping `OUTAGE=$(cat /tmp/review-runs-outage-number)` from that block leaves every assertion in the file green — I ran that mutant — and the close then no-ops silently, which is the failure the last round fixed. Pinning the two lines as one string also makes a path rename move both ends together.

<details><summary>Checked, not raising</summary>

- `test_outage_tracker_title_stays_in_sync` asserts the literal in all three files rather than importing `report_failure.TITLE`, which catches drift in the reporter too — stricter than what I asked for, and consistent with this file's text-read idiom.
- `[ -n "$OUTAGE" ] && …` exits 1 on the no-tracker day, which is most days. The prose gates the block on “if a tracker was found”, and it is the form I suggested, so I'll leave it rather than churn it into an `if`.
- The `> /tmp/review-runs-outage-number` redirect truncates before `gh` runs, so a failed lookup leaves an empty file behind the `exit 1`; if a later block reads it anyway it gets empty and skips the close rather than closing a wrong issue.
- `gh api "$SUBJECT_URL"` prints `(HTTP 404)` for a deleted subject and a distinct rate-limit message otherwise, so the new bullet's 404-vs-transient split is decidable at the call site with the recipe as written.

</details>

generator/tests/test_notification_queue_contract.py
max-sixty: Fixed in 6de73a18. The contract now pins the `OUTAGE=$(cat /tmp/review-runs-outage-number)` reload and the guarded close as one contiguous block, tying the persisted ID’s read end directly to its use.

tend-agent: Verified against `6de73a18`. The contiguous `close_block` string pins both ends of the handoff, and the two mutants I was worried about now fail it: deleting `OUTAGE=$(cat /tmp/review-runs-outage-number)` from the close block, and renaming the `/tmp` path at the close end only (the write-end assertion stays green in both, so it is the new string doing the work). All seven functions in the file pass on the head tree; nothing else outstanding from my side.

PR #1076, opened 2026-08-28
fix(generator): validate watched_workflows and branches as string lists

tend-agent:
`workflows.<name>.watched_workflows` and `workflows.<name>.branches` were the only list-typed config fields with no type validation — every other one (`protected_branches`, `sandbox_path`, `sandbox_setup`, `secrets.allowed`) checks `isinstance(..., list)` and rejects non-string elements. Both render straight into the generated `workflow_run:` trigger through `tojson`, so an unchecked value reached the adopter's workflow file verbatim. This adds the same check both fields were missing, plus regression tests.

Four concrete failures, all reproduced before the fix:

| Config | Before | After |
| --- | --- | --- |
| `watched_workflows: ci` | renders `workflows: "ci"` — matches no workflow, so `ci-fix` silently never fires | `ClickException` at `init` |
| `watched_workflows: 5` | `TypeError: object of type 'int' has no len()` — a bare traceback out of `tend init` | `ClickException` at `init` |
| `watched_workflows: {a: b}` | a mapping has a length, so it clears the existing empty-list check and renders as a JSON object | `ClickException` at `init` |
| `branches: []` | vacuously a list of non-empty strings, so it renders as `branches: []` — a trigger that matches no branch | `ClickException` at `init` |

The scalar case is the one that matters most: it is the natural typo, it passes `actionlint`, and the only symptom is that `tend-ci-fix` never runs. Per CLAUDE.md, a config problem should "fail with a clear error, not silently parse".

The scalar render, reproduced against `main` before the change:

```
$ cat .config/tend.yaml
bot_name: test-bot
workflows:
  ci-fix:
    watched_workflows: ci
    branches: main

$ tend init   # rendered .github/workflows/tend-ci-fix.yaml
on:
  workflow_run:
    workflows: "ci"
    types: [completed]
    branches: "main"
```

`actionlint 1.7.12` (the rev pinned in `.pre-commit-config.yaml`) accepts that file with exit 0, so nothing downstream catches it.

The eight new parametrized cases: 7 fail on `main`, all 8 pass with the fix. `test_branches_explicit_value` is the positive control — it passes both ways, confirming a valid list still reaches the trigger. Full suite: 731 passed. `ruff format --diff` and `ruff check` clean.

</details>

### DIFF
--- a/generator/src/tend/config.py
+++ b/generator/src/tend/config.py
@@ -436,6 +436,12 @@ class Config:
                             f"workflows.{name}.{key} must be a list of "
                             f"non-empty strings (e.g. {key}: {example})"
                         )
+                if branches is not None and not branches:
+                    raise click.ClickException(
+                        f"workflows.{name}.branches: [] matches no branch — "
+                        "omit the key to default to the repository's default "
+                        "branch, or list the branches to watch."
+                    )
                 if watched is not None and len(watched) == 0 and name == "ci-fix":
                     raise click.ClickException(
                         "watched_workflows: [] is invalid for ci-fix — "
--- a/generator/src/tend/config.py
+++ b/generator/src/tend/config.py
@@ -417,6 +417,25 @@ class Config:
                 )
             if isinstance(wf_raw, dict):
                 watched = wf_raw.get("watched_workflows")
+                branches = wf_raw.get("branches")
+                # Both render straight into the `workflow_run:` trigger through
+                # `tojson`, so an unchecked value reaches the workflow file
+                # verbatim: `watched_workflows: ci` becomes `workflows: "ci"`,
+                # which GitHub matches against nothing, and ci-fix silently
+                # never fires. A number gets no further than `len()` below,
+                # which raises a bare TypeError instead of a config error.
+                for key, value, example in (
+                    ("watched_workflows", watched, '["ci"]'),
+                    ("branches", branches, '["main"]'),
+                ):
+                    if value is not None and (
+                        not isinstance(value, list)
+                        or not all(isinstance(s, str) and s for s in value)
+                    ):
+                        raise click.ClickException(
+                            f"workflows.{name}.{key} must be a list of "
+                            f"non-empty strings (e.g. {key}: {example})"
+                        )
                 if watched is not None and len(watched) == 0 and name == "ci-fix":
                     raise click.ClickException(
                         "watched_workflows: [] is invalid for ci-fix — "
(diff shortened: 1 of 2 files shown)


generator/src/tend/config.py
```diff
@@ -417,6 +417,25 @@ def load(cls, path: Path | None = None) -> Config:
...
+                    ("branches", branches, '["main"]'),
+                ):
+                    if value is not None and (
+                        not isinstance(value, list)
+                        or not all(isinstance(s, str) and s for s in value)
+                    ):
+                        raise click.ClickException(
+                            f"workflows.{name}.{key} must be a list of "
+                            f"non-empty strings (e.g. {key}: {example})"
+                        )
```
tend-agent, line 438: `branches: []` passes this check and renders as a trigger that matches no branch. Worth rejecting it here, next to the equivalent `watched_workflows: []` guard:

```suggestion
                        )
                if branches is not None and not branches:
                    raise click.ClickException(
                        f"workflows.{name}.branches: [] matches no branch — "
                        "omit the key to default to the repository's default "
                        "branch, or list the branches to watch."
                    )
```

tend-agent (commented):
One gap left in the same class of failure this PR closes: `branches: []` is a list of non-empty strings (vacuously), so the new check accepts it, and it renders straight through as `branches: []` in the `workflow_run:` trigger — which matches no branch, so `ci-fix` never fires. That is the same silent-never-fires symptom the PR calls out as the one that matters most, and the check three lines below already rejects the symmetric `watched_workflows: []` for exactly this reason.

Reproduced on this branch — `watched_workflows: ["ci"]` plus `branches: []` loads without error and renders:

```yaml
on:
  workflow_run:
    workflows: ["ci"]
    types: [completed]
    branches: []
```

Suggestion inline. Everything else checks out: the new validation matches the shape the other list fields use, the ordering leaves the existing `watched_workflows: []` check reachable (an empty list clears the new `all(...)`), and hoisting `branches` out of the `WorkflowConfig(...)` call is behavior-preserving. Full suite is green here (731 passed), as is CI on this head.

generator/src/tend/config.py
tend-agent: Applied in c7b84c2 — took the suggestion as written, placed immediately above the `watched_workflows: []` guard so the two empty-list cases sit together. Confirmed the gap first on the pre-fix head: `watched_workflows: ["ci"]` with `branches: []` loaded clean and rendered `branches: []` into the trigger.

Added `test_branches_empty_list_rejected` alongside the existing parametrized cases. Python suite green (701 passed locally; the 30 `proxy/` tests need mitmproxy, which isn't installed in this sandbox — CI covers them, and this change is generator-only).

PR #1078, opened 2026-08-29
Make Tend review runs less visible

max-sixty:
Claude review runs currently render their transcript in the job summary and add a PR-scoped legacy commit status that points back to the workflow. That status exists only to skip a usually redundant queued agent boot, but it leaves another visible check row on every reviewed commit.

This makes rendered transcripts opt-in and removes the examined-status producer and gate. Reviews still serialize per PR, and `queue: max` preserves pending PR events within GitHub's queue limit so a later push cannot replace `ready_for_review`. Before posting, the reviewer rechecks the live HEAD, PR state, and prior substantive reviews.

Raw session artifacts remain available for deliberate diagnosis and expire after 30 days.

Tests: `wt test`; `pre-commit run --all-files`.


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -456,23 +470,24 @@ git commit -m "fix: <description>"
...
 ```
 
-The generated workflow holds a replacement run queued by any mid-session push until this session ends; a pre-check then reads this stamp off the live HEAD and skips the agent when it's present. The context carries the PR number because one branch can be two open PRs with different bases — an examination of one must not gate the other.
+If the PR is still open and `LIVE_DRAFT == false`, it became ready while this
+session was running. Treat that as `FORCE_FULL_REVIEW=true` on `LIVE_HEAD` and
+restart at step 1 for one full non-draft pass, preserving that override instead
+of resetting it from the event payload. Do this regardless of the triggering
+event or how small the incremental diff is. This preserves the author's
+`ready_for_review` request when GitHub's one-deep concurrency queue replaces
+that pending event with a later `synchronize` event.
```
tend-agent: The re-entry condition survives the pass it triggers: `STARTED_DRAFT` is still the first pre-flight's `true`, and `LIVE_DRAFT` is still `false` once the PR is ready, so this fires again every time the forced pass reaches step 9. Bounding it needs one piece of state.

```suggestion
If the PR is still open and `LIVE_DRAFT == false`, it became ready while this
session was running. Treat that as `FORCE_FULL_REVIEW=true` on `LIVE_HEAD` and
restart at step 1 for one full non-draft pass, preserving that override instead
of resetting it from the event payload. Do this regardless of the triggering
event or how small the incremental diff is. This preserves the author's
`ready_for_review` request when GitHub's one-deep concurrency queue replaces
that pending event with a later `synchronize` event. Set `STARTED_DRAFT=false`
before restarting, so the forced pass runs once and cannot re-enter here.
```

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -215,23 +229,23 @@ Before posting, verify HEAD hasn't moved, the PR is still open, and no review wa
...
-# reports `.commit_id == $HEAD_SHA` and would discard the very re-review step
-# 1's `FORCE_PUSHED` probe just unblocked.
-ALREADY_POSTED=$(${CLAUDE_PLUGIN_ROOT}/scripts/bot-review-state.sh <number> \
-  | jq -r '.at_head.at // empty')
-[ -n "$ALREADY_POSTED" ] && echo "Already reviewed — finish at step 9" && exit 0
+# synthetic reply container and a review re-anchored by a rewrite are excluded.
+# A forced ready-for-review pass may replace Tend's earlier draft COMMENT, but
+# no other substantive review — including a full pass that raced this one.
+POST_STATE=$(${CLAUDE_PLUGIN_ROOT}/scripts/bot-review-state.sh <number>)
+ALREADY_POSTED=$(jq -r --argjson force "$FORCE_FULL_REVIEW" '
```
tend-agent: `--argjson` rejects an empty string, so if `FORCE_FULL_REVIEW` wasn't re-derived in this invocation (the step-1 preamble warns shell state doesn't persist), jq errors out, `ALREADY_POSTED` is empty, and the guard waves through a second review on a head that already has one. Defaulting it fails safe.

```suggestion
ALREADY_POSTED=$(jq -r --argjson force "${FORCE_FULL_REVIEW:-false}" '
```

tend-agent (commented):
Removing the gate and the stamp is the right direction — the mechanism cost more to reason about than the boots it saved, and `bot-review-state.sh`'s `last_substantive`/`at_head` already covers the duplicate-review case it was really protecting (an empty-body APPROVE counts as substantive, so a queued run still stops at step 1). Two defects in the replacement, plus a coupling worth naming.

**Step 9 has no termination guard.** Its re-entry condition is "a session whose first pre-flight found `STARTED_DRAFT == true`", and step 1 says to *keep* the first `STARTED_DRAFT` value. After the forced pass runs, both are still true — `STARTED_DRAFT` was true at first pre-flight, and `LIVE_DRAFT` is still `false` — so the condition re-fires. "One full non-draft pass" is the intent, but nothing in the state distinguishes "already ran". Step 5's `ALREADY_POSTED` stops the *second* post (`draft_mode` is false on the full review it just published), so this isn't a duplicate-review bug — it's the whole review flow, code-review pass included, re-running until the job timeout. Inline suggestion below.

**`--argjson force "$FORCE_FULL_REVIEW"` fails open toward a duplicate post.** The step-1 preamble warns that shell state doesn't persist between tool calls, so `FORCE_FULL_REVIEW` has to be re-derived in the step-5 invocation. When it isn't, `--argjson force ""` is a jq parse error, `ALREADY_POSTED` comes back empty, and the run posts a second review at a head that already carries one. Note the asymmetry with the neighbouring `$HEAD_SHA`, which fails *safe* under the same slip — an empty `$HEAD_SHA` reads as "HEAD moved" and exits. One-token fix inline.

**`draft_mode` makes a prose line load-bearing.** `bot-review-state.sh` now decides whether a forced pass may replace an earlier review by `startswith("Reviewing as a draft —")`, but the place that line is authored — the Draft mode bullet, `Open the review body with one line framing it:` — reads as guidance, not a fixed key. A session that paraphrases it ("Reviewing this as a draft — …", or an en dash) gets `draft_mode: false`, and the `ready_for_review` pass is then blocked by `ALREADY_POSTED` and silently drops the full review. That line sits outside the diff, so no inline suggestion — happy to push a commit marking the prefix as fixed (CLAUDE.md's authoring rules already carve out "mandatory exact wording for fragile mechanics: dedup keys"), or you can fold it in.

Two smaller things, both take-or-leave:

- `test_review_queues_pushes_without_an_examined_status` pins twelve prose fragments from `review/SKILL.md` inside a test about workflow generation. `test_notification_queue_contract.py` is the existing home for skill-contract assertions, and a few of these carry no mechanism (`"preserving that override"`, `"5 (COMMENT path), 7, and 9 still apply"`) — they'll break on any reword with nothing to tell the next reader why they mattered.
- The transcript is the only thing `verdict()` writes to the job summary, so the default flip removes it entirely, failures included. Failure signal survives as the error annotation plus `_quote_stderr`, so this may be exactly what you want; if not, `if show_full_output == "true" or claude_exit:` keeps it where visibility earns its keep.

Everything else checks out: `restore_local_actions`' remaining callers both pass no condition, `importlib.resources` is still used by the mention and notifications scripts, the `docs/` and `install-tend/references/` copies of `tend.example.yaml` moved together, no stale `tend-review/<pr>` or `steps.gate` references survive in `plugins/`, and `uv run pytest` is green locally at 728 passed.

plugins/tend-ci-runner/skills/review/SKILL.md
max-sixty: Resolved in `3c72727`: the step-9 re-entry mechanism is gone. Generated review jobs now use GitHub's native `queue: max`, so the actual `ready_for_review` event remains queued within the platform limit instead of being replaced by a later push. That removes both the infinite restart and the final read-to-exit race.

plugins/tend-ci-runner/skills/review/SKILL.md
max-sixty: Resolved in `3c72727`: the posting guard now passes `${FORCE_FULL_REVIEW:-false}` to `jq`, so missing shell state fails toward deduplication instead of allowing a duplicate review.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Confirmed against `3c72727` — step 9 is gone, nothing in `plugins/` still reads `STARTED_DRAFT`/`LIVE_DRAFT`, and `queue: max` covers the case it existed for without the re-entry state.

I checked the key itself, since it ships to every adopter's `tend-review`: it's real and job-level applies. [GitHub's changelog dates it 2026-05-07](https://github.blog/changelog/2026-05-07-github-actions-concurrency-groups-now-allow-larger-queues/) — up to 100 pending per group, FIFO by the time each run *started waiting* on the group rather than dispatch time, so an earlier `ready_for_review` keeps its place ahead of a later push. One thing worth knowing: `queue: max` cannot be combined with `cancel-in-progress: true`, so the `false` next to it is load-bearing rather than habit.

The actionlint `-ignore` is right for now — no release knows the key yet (v1.7.12 shipped 2026-03-30, before the changelog), not just this pin. It won't clear itself, though: the weekly sweep's `pre-commit autoupdate` moves the rev and leaves the arg, so the suppression outlives the schema gap silently. Worth checking the message still reproduces at the next actionlint bump.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Confirmed — `${FORCE_FULL_REVIEW:-false}` now fails toward dedup, matching `$HEAD_SHA`'s behaviour under the same slip.

The third finding from the review body is folded in here too, so nothing to push from my side: the draft line is now `Open the review body with this exact line:`, and `test_bot_review_state.py` closes the triangle — `DRAFT_REVIEW_LINE` is asserted verbatim in the skill *and* used as the review body in `test_at_head_identifies_a_tend_draft_review`, so changing the skill's line, the script's `startswith` prefix, or the constant alone now fails a test. That's the fragile-mechanic carve-out doing its job.

generator/src/tend/templates/review.yaml.j2
```diff
@@ -15,32 +15,23 @@ jobs:
   review:
     concurrency:
       group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
-      # A push mid-review queues a replacement run rather than killing the
-      # session: the running review folds the push in and stamps the commits
-      # it examined, and the gate step below lets the queued run exit without
-      # booting an agent when its HEAD is already covered.
+      # Preserve pending PR events within GitHub's queue limit: a push must not
+      # replace a ready-for-review run. Runs stay serial so outward actions
+      # cannot race.
```
tend-agent: GitHub rejects `queue: max` together with `cancel-in-progress: true` as a workflow validation error — not a warning, the whole file fails to parse. Since job overrides deep-merge, an adopter setting only `cancel-in-progress: true` on this job keeps `queue: max` and silently breaks their review workflow. Naming the constraint here puts it in the file they're looking at when they decide to override.

```suggestion
      # Preserve pending PR events within GitHub's queue limit: a push must not
      # replace a ready-for-review run. Runs stay serial so outward actions
      # cannot race. An override setting `cancel-in-progress: true` must set
      # `queue: null` too — GitHub rejects the pair as an invalid workflow file.
```

generator/tests/test_bot_review_state.py
```diff
@@ -366,3 +401,28 @@ def test_the_repo_is_named_explicitly_on_every_call(env: dict[str, str]) -> None
...
+    assert (
+        "If `FORCE_FULL_REVIEW` is false and the incremental changes are trivial"
+        in skill
+    )
+    assert f"Open the review body with this exact line: `{DRAFT_REVIEW_LINE}`" in skill
+    assert "Post at most one review per run." in skill
+    assert "exception to one review per run" not in skill
+    assert "STARTED_DRAFT" not in skill
+    assert "LIVE_DRAFT" not in skill
+    assert "### 9." not in skill
```
tend-agent: Take-or-leave: this one forbids the skill from ever having a ninth step, rather than the mechanism the PR removed. The two assertions above it already pin that removal by name, so a future unrelated step 9 would fail a test called `preserves_the_status_free_queue_contract` for a reason the name doesn't explain.

```suggestion
```

tend-agent (commented):
Both defects from the last pass are fixed. `${FORCE_FULL_REVIEW:-false}` fails safe now, and the draft framing line is pinned as exact wording with `DRAFT_REVIEW_LINE` shared between the skill assertion and the `bot-review-state.sh` prefix test — a paraphrase breaks a test instead of silently flipping `draft_mode`. Deleting step 9 is a better answer than bounding its re-entry, and it holds: step 1 reads `HEAD_SHA` live rather than from the event payload, so a `ready_for_review` run that waits behind a session still forces a full non-draft pass against whatever the head has become by the time it starts.

I checked `queue: max` against real GitHub rather than the docs, since actionlint's pin rejects the key and no workflow under `.github/workflows/` carries it yet — nothing in this repo's CI actually exercises the syntax. A job-level `concurrency` with `group` + `queue: max` + `cancel-in-progress: false` is accepted and the job runs, so the pin will hold at the next regen.

The same probe found the one new thing, inline on the template: with `cancel-in-progress: true` alongside `queue: max`, GitHub refuses to parse the file at all — the run comes back named by its file path, conclusion `failure`, no jobs. That pairing is now one adopter override away, and this change raises the incentive to write it: review runs used to collapse to a single pending replacement, and now they serialize up to 100 deep, which is exactly what makes someone reach for `cancel-in-progress: true`.

<details><summary>How the override reaches that state</summary>

Overrides merge per RFC 7396, so a mapping merges rather than replaces — an adopter writing only the one key keeps `queue: max` underneath it:

```yaml
workflows:
  review:
    jobs:
      review:
        concurrency:
          cancel-in-progress: true
```

Generating with that config emits both keys:

```
    concurrency:
      group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
      queue: max
      cancel-in-progress: true
```

The inline suggestion is the cheapest fix — the comment renders into every adopter's `tend-review.yaml`, which is where someone reads the concurrency block before overriding it. If you'd rather catch it at `init` instead, the merge step already warns on unknown job names, so rejecting this pair there would sit in existing machinery rather than new machinery. Happy to push either.

</details>

generator/src/tend/templates/review.yaml.j2
max-sixty: Resolved in `08a71d0`: the rendered concurrency block now says that an override setting `cancel-in-progress: true` must also set `queue: null`, and the generator contract asserts that guidance is present. The affected snapshots are updated.

generator/tests/test_bot_review_state.py
max-sixty: Taken in `08a71d0`: the assertion banning any future `### 9.` heading is removed; the targeted `STARTED_DRAFT` and `LIVE_DRAFT` absence checks continue to pin deletion of the actual mechanism.

tend-agent (approved):

generator/src/tend/templates/review.yaml.j2
tend-agent: Confirmed at `08a71d0` — and I checked the remedy the note prescribes, not just its presence. Overriding with `cancel-in-progress: true` **and** `queue: null` renders the block with `queue` gone and the comment intact, so RFC 7396's null-delete does reach the key the note names; `cancel-in-progress: true` alone still emits both keys, which is the case the note is there to catch. The actionlint `-ignore` this PR adds also means the key survives lint at the next regen of this repo's own `tend-review.yaml`, so nothing here waits on the pin.

Take-or-leave on the new contract assertion: `"must set\n      # \`queue: null\`" in content` pins the comment's wrap point, so re-flowing the sentence fails a test whose guidance survived intact. `` "`queue: null` too" in content `` sits on one rendered line and asserts the same thing without the layout coupling.

<details><summary>What I rendered</summary>

Both shapes generated from `_minimal_config` with only the job override varying:

```yaml
# cancel-in-progress: true + queue: null
    concurrency:
      group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
      # Preserve pending PR events within GitHub's queue limit: a push must not
      # replace a ready-for-review run. Runs stay serial so outward actions
      # cannot race. An override setting `cancel-in-progress: true` must set
      # `queue: null` too — GitHub rejects the pair as an invalid workflow file.
      cancel-in-progress: true

# cancel-in-progress: true alone
    concurrency:
      group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
      # ...same comment...
      queue: max
      cancel-in-progress: true
```

`generator/tests/test_bot_review_state.py` still pins the deletion by name (`STARTED_DRAFT`, `LIVE_DRAFT` absent), so dropping the `### 9.` assertion costs no coverage. `uv run pytest generator/tests/test_generate.py generator/tests/test_bot_review_state.py` passes at this head (148 passed).

</details>

PR #1080, opened 2026-08-29
fix(review): make the code-review pass its own step, gated at submit

### LINKED ISSUE #1079: review: sessions still skip the mandated code-review second pass in ~1 of 4 reviews post-#937
Following up on [#937](https://github.com/max-sixty/tend/pull/937), whose closing paragraph pre-registered this observation as the trigger for its next step: *"If a later cycle finds sessions still skipping the pass with this text in place, the next step is a harder mechanism (a distinct numbered step, or a step-5 pre-post check that the pass ran) rather than more wording."* This is that later cycle, from a second adopter repo ([diffplug/dormouse](https://github.com/diffplug/dormouse), pinned at `max-sixty/tend/claude@0.1.20`, published 2026-08-27 — well after #937 merged on 2026-08-11).

## What happens

The bundled `review` skill's step 4 mandates the second pass unconditionally: *"Every review that reaches this step runs a `/tend-ci-runner:code-review` pass over the PR's merged tree — trivial diffs included; the depth-scaling above sets how deep the pass goes, never whether it happens."* Across three consecutive `review-runs` windows on dormouse, **13 of 48 eligible `tend-review` sessions never invoked it** — no `Skill` call, no subagent invocation, no tool-use error, and the string `code-review` absent from the session's assistant text entirely. All 13 posted a non-empty verdict anyway.

The failure is **stochastic, not structural**: it splits within a single PR across adjacent rounds under identical conditions. On [dormouse#469](https://github.com/diffplug/dormouse/pull/469) the 19:40 and 20:49 rounds skipped the pass while the 20:36 round between them ran it. Post-#937 wording moved the rate from upstream's uniform 4-of-4 to roughly 1-in-4, but did not reach zero — which is the argument for the mechanism #937 already proposed over a third round of prose.

| Window | Skipped / eligible |
|---|---|
| 2026-08-26 → 2026-08-27 | 5 / 16 |
| 2026-08-27 → 2026-08-28 | 3 / 15 |
| 2026-08-28 → 2026-08-29 | 5 / 17 |

Eligibility excludes step 1's trivial-skip path, which explicitly says *"Do NOT proceed to steps 2–6"* and so never reaches step 4 — those sessions either posted nothing (resolved threads, stamped HEAD) or posted the prescribed empty-body `APPROVE`. Method was to enumerate `Skill` tool calls in each session's JSONL artifact; the per-session breakdowns with run links are in the tracking issue, [diffplug/dormouse#457](https://github.com/diffplug/dormouse/issues/457).

## Cost: two wrong outward actions, both from skipping sessions

Both are memory-asserted claims that ran no command to check — the shape a `code-review` verify pass is adversarial against.

**A false claim in an APPROVED review, merged 5 seconds after a sibling narrowed it.** [Session 33209574328](https://github.com/diffplug/dormouse/actions/runs/33209574328) posted [an approval on #469](https://github.com/diffplug/dormouse/pull/469#pullrequestreview-5054977889) at `20:49:49Z` asserting that every `minMatches` in the repo's `deploy-lint` was the exact occurrence count, "which makes deleting any single copy red by construction." False on both platforms. A sibling `tend-mention` session [narrowed it with seeded experiments](https://github.com/diffplug/dormouse/pull/469#issuecomment-5457637663) at `20:52:18Z`; #469 merged at `20:52:23Z`. The macOS half was later fixed in [dormouse#482](https://github.com/diffplug/dormouse/pull/482), confirming the approval had certified something untrue.

**A false `set -e` claim posted inline.** [Session 33053278119](https://github.com/diffplug/dormouse/actions/runs/33053278119) claimed in [an inline comment](https://github.com/diffplug/dormouse/pull/453#discussion_r3869957259) that under `bash -e` the form `[ "$i" -lt 3 ] && sleep 15` would exit the shell on the last iteration. The maintainer [tested it and it does not reproduce](https://github.com/diffplug/dormouse/pull/453#issuecomment-5436368751) — POSIX `set -e` exempts every command in an `&&` list except the last.

## Proposed fix

#937 named two candidates. The evidence favors the first, and they compose.

**1. Promote the pass to its own numbered step.** Today it is a bolded paragraph roughly a third of the way into a ~40-line step 4 whose surrounding text is all judgment-scaling ("Scale depth to the change", "Don't over-analyze trivial changes"). A mandate embedded in a step about *calibrating effort* reads as one more calibration knob, which is consistent with the stochastic skip: sessions that skip aren't refusing the mandate, they're not registering it as a discrete action. Splitting it out — `### 4. Review` → `### 5. Second pass` → `### 6. Submit` — makes it a thing the session either did or didn't do, rather than a qualifier on something else.

**2. Add a step-5 pre-post gate.** The Posting mechanics block already carries scripted pre-post checks (`HEAD moved`, `PR is $PR_STATE`, `Already reviewed`). Whether the pass ran isn't observable from bash, so this one has to be an explicit stated precondition rather than a script line — something like:

```
Before submitting, state what the `/tend-ci-runner:code-review` pass returned —
its confirmed findings, or "no findings". If you cannot, the pass did not run:
go back and run it before posting. A review that reached this point
without it is not submittable.
```

Making the pass's *output* a required input to the verdict is deterministic in a way "run the pass" is not, because the session can't compose the review body without noticing the hole.

Happy to open a PR for either or both if you'd like a concrete diff to react to.

tend-agent:
## Problem

The required `/tend-ci-runner:code-review` pass is still skipped in a
meaningful share of review sessions because it is embedded inside the manual
review step's depth-calibration prose. That can remove the intended independent
signal before Tend posts a verdict.

## Solution

- make the code-review pass its own numbered, sequential step
- require the session to state that pass's result before submission
- retain it for draft and self-authored PR paths
- rerun it when a safe re-target adds a delta, while leaving step-1 trivial
  close-outs exempt from the full-review path
- keep the result in session output so an empty-body APPROVE remains empty
- preserve current-main author-readiness, deduplication, and posting-preflight
  safeguards while renumbering downstream references

Closes #1079.

## Verification

- `wt test` after merging current `main` (833 Python tests, 32 worker tests,
  TypeScript typecheck)
- 67 focused review-state, preflight, and contract tests after final cleanup
- independent review found the sequential compliance gate acceptable


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -187,7 +180,20 @@ Two search strategies, both required:
...
+Scale its depth to how core the change is:
+
+- Peripheral or mechanical (config, dependency bumps, test-only, docs that don't assert how the code behaves): tell it the change is peripheral, so it runs the short angle set in one pass.
+- The project's core logic, or prose asserting how it behaves: tell it the change is core, so it fans the angles out and sweeps for gaps. Prose is checked by reading the code it describes, so a one-line Markdown diff can still be core.
+
+What counts as core is repo-specific; let the project's own guidance (CLAUDE.md, a repo review skill) or your judgment decide. Both passes feed one verdict: fold its findings into the review you submit in step 6. It only reports back — it never posts a review, comment, or commit of its own, so the dedup and single-review path is preserved.
+
+### 6. Submit
+
+**Before submitting, say what the step 5 pass returned** — its confirmed findings, or "no findings". Say it in the session, not the review body: an empty-body APPROVE stays empty. If you can't say, the pass didn't run — go back to step 5 and run it. A review that reaches this point without it is not submittable.
```
tend-agent: The antecedent of "Say it" is "its confirmed findings", so this reads as "keep the pass's findings out of the review body" — contradicting step 5's "fold its findings into the review you submit in step 6" four lines up. Naming the statement as the thing that stays in the session keeps the empty-body APPROVE contract without putting the two instructions in tension.

```suggestion
**Before submitting, say what the step 5 pass returned** — its confirmed findings, or "no findings". That statement is a compliance check: say it in the session, not the review body, so an empty-body APPROVE stays empty. The findings themselves still get folded into the review, per step 5. If you can't say, the pass didn't run — go back to step 5 and run it. A review that reaches this point without it is not submittable.
```

tend-agent (commented):
The renumbering audit inside `SKILL.md` checks out — I tallied every `step N` reference against the base and each one shifts consistently (5→6, 6→7, 7→8, 8→9), the draft-mode list and the `steps 2–7` trivial-exit range both map correctly, and no control-flow path reaches submit while bypassing the new gate. Two things outside that audit:

**A stale step reference survives in `TODO.md`.** The renumbering swept `SKILL.md` only, but the paragraph beginning "Revisit when one of those moves" in `TODO.md` says cutting over to the built-in would cost "a prompt reshape plus a per-harness branch in the `review` skill's step 4". That branch belongs in step 5 now — step 4 no longer mentions the pass at all. Whoever revisits that decision lands on depth-calibration prose and has to re-derive where the invocation actually lives.

**The gate's second sentence is in tension with step 5.** See the inline suggestion — as written, the grammatical antecedent of "Say it in the session, not the review body" is "its confirmed findings", which reads as an instruction to keep the pass's findings out of the review. That's the opposite of step 5's "fold its findings into the review you submit in step 6", and a session taking the literal reading would state real findings in its own output and then post the empty-body APPROVE — the outward action [#1079](https://github.com/max-sixty/tend/issues/1079) documents. The clarification is small enough that the mechanism survives it intact.

Otherwise the shape is right: this is the mechanism [#937](https://github.com/max-sixty/tend/pull/937) pre-registered, it's a restructure plus one paragraph rather than new machinery, and the two wrong outward actions in #1079 put it squarely in the "prevents a wrong outward action" bucket that `CLAUDE.md` says complexity has to earn.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Applied the suggestion verbatim in 2dd0aa4. You're right that "its confirmed findings" was the nearest antecedent, and the literal reading — state findings in-session, post an empty APPROVE — is exactly the outward action #1079 documents. Naming the *statement* as the compliance check, plus the explicit "the findings themselves still get folded into the review, per step 5", removes the tension without weakening the gate.

tend-agent:
Both review points are addressed in 2dd0aa4.

**Stale `TODO.md` reference** — the built-in-code-review note now cites the `review` skill's step 5 for the per-harness branch, not step 4. I swept the rest of the repo for cross-references into the review skill's numbering and found none beyond that one line; the other `step N` hits in `plugins/` all refer to their own skills' steps.

**Gate antecedent** — applied the suggestion as written; see the inline reply.

`uv run pytest` — 729 passed. `uvx pre-commit run --files plugins/tend-ci-runner/skills/review/SKILL.md TODO.md` clean.

<details><summary>Sweep used for the cross-reference check</summary>

```bash
grep -rn "step [0-9]" plugins/tend-ci-runner/skills/running-in-ci/
grep -rn "review skill" --include='*.md' plugins/ .claude/ | grep -iE "step"
```

The only in-tree hit is `SKILL.md:192`'s own "submit in step 6". The released marketplace copy of `running-in-ci` carries a "(review skill, step 9)" citation, but that paragraph was already rewritten in-tree and no longer names a step, so nothing there needs renumbering with this PR.

</details>

tend-agent:
Measured the pre-fix skip rate on tend itself, since this PR names "the adopter-side skip rate over the next few `review-runs` windows" as the verification it needs. This is the baseline to compare against.

**Window** 2026-08-29T07:48Z → 2026-08-30T07:48Z. **33 `tend-review` runs**, all with session logs; **24 posted something** (a review body or an approval). **7 of those 24 ran no `/tend-ci-runner:code-review` pass — 29%**, which reproduces the ~1-in-4 rate [#1079](https://github.com/max-sixty/tend/issues/1079) reported from a second adopter, here on tend's own repo.

All 33 sessions ran the released skill, and the release does not carry this PR: the mount at `0.1.22` still has `### 4. Review` / `### 5. Submit`, so the mandate is still the bolded paragraph inside step 4. Nothing here is evidence about the fix — it is the number the fix has to move.

<details><summary>Per-session evidence</summary>

Counted by `Skill` tool-uses naming `code-review` in each run's session JSONL, cross-checked against the posting call in the same log. The four full-review skips were each checked for the subagent path this PR's description names: all four show only `Bash` plus the two environment `Skill` loads (`running-in-ci`, `running-tend`), no `Task`/`Agent` tool-use, and no mention of the pass anywhere in the assistant text — so they are genuine skips, not a pass run through a different tool.

**Posted a full review body, no pass** (4):

| Run | PR |
|---|---|
| [33279093420](https://github.com/max-sixty/tend/actions/runs/33279093420) | #1082 |
| [33279665975](https://github.com/max-sixty/tend/actions/runs/33279665975) | #1082 |
| [33296060001](https://github.com/max-sixty/tend/actions/runs/33296060001) | #1090 |
| [33296431588](https://github.com/max-sixty/tend/actions/runs/33296431588) | #1090 |

**Posted an empty-body approval only, no pass** (3): [33264173133](https://github.com/max-sixty/tend/actions/runs/33264173133) (#1078), [33279979513](https://github.com/max-sixty/tend/actions/runs/33279979513) (#1082), [33294518307](https://github.com/max-sixty/tend/actions/runs/33294518307) (#1084).

The shape is worth noting for step 6's precondition: **every skip is a follow-up session on a PR an earlier session had already reviewed with the pass** — #1078, #1082, #1084 and #1090 each had a first session that ran it. No first-look review skipped. That is the case the submit-time precondition has to cover, because a session re-reviewing an increment is exactly the one that reads the mandate as a depth knob it has already satisfied.

None of the seven produced a wrong outward action that I can find: the four full reviews were substantive and the three approvals all landed on PRs that merged green.

</details>

tend-agent (commented):
Re-walked every `step N` reference in the merged file against the new headers and against the prose #1104 and #1107 brought in — they all resolve, and `TODO.md` is the only cross-file referent. Three things on the gate itself.

**The gate is unconditional, but two approve paths can never satisfy it.** See the inline suggestion. Step 1's trivial-increment approve and the dedup rule's "resolves the last open one" approve both send the session to step 6 for the POST recipe while explicitly skipping steps 2–7, so no step 5 pass ran — and "a review that reaches this point without it is not submittable" reaches them as written. A session reading it literally either runs a full code-review pass on a trivial increment, contradicting the "Do NOT proceed to steps 2–7" it just followed, or posts nothing and leaves the PR in exactly the limbo those paths exist to clear.

**`the standard step 4 sets` narrowed silently.** The mid-review bullet under **A push mid-review re-targets the review** that begins "Review the delta to the standard step 4 sets — it is new code" is untouched by this PR, but its referent shrank: step 4 used to carry the pass mandate, so a re-targeted delta was covered by it, and now step 4 is manual checks only. Whether a delta should get its own pass is a real call — re-running it per mid-review push costs a pass per push — but as it stands the PR settles it by omission. It's outside the diff, so it isn't in the inline set; either widening the reference to steps 4–5 or saying outright that the delta doesn't re-run step 5 would close it.

**Nothing pins the new structure.** cd91ef1 renumbers `test_review_approval_gates_on_author_stated_readiness`'s comments, but no test asserts what this PR actually ships — that the pass is its own step, and that step 6 carries the gate. `test_cross_file_contracts.py` is where that convention lives, and #1079's failure mode is the mandate drifting back into surrounding prose, which is the thing a header assertion would catch.

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -191,7 +184,20 @@ Two search strategies, both required:
...
+- The project's core logic, or prose asserting how it behaves: tell it the change is core, so it fans the angles out and sweeps for gaps. Prose is checked by reading the code it describes, so a one-line Markdown diff can still be core.
+
+What counts as core is repo-specific; let the project's own guidance (CLAUDE.md, a repo review skill) or your judgment decide. Both passes feed one verdict: fold its findings into the review you submit in step 6. It only reports back — it never posts a review, comment, or commit of its own, so the dedup and single-review path is preserved.
+
+### 6. Submit
+
+**Before submitting, say what the step 5 pass returned** — its confirmed findings, or "no findings". That statement is a compliance check: say it in the session, not the review body, so an empty-body APPROVE stays empty. The findings themselves still get folded into the review, per step 5. If you can't say, the pass didn't run — go back to step 5 and run it. A review that reaches this point without it is not submittable.
```
tend-agent: The gate has no exemption, so it also binds the two approve paths that deliberately never reach step 5 — step 1's trivial-increment approve and the dedup rule's "resolves the last open one" approve, both of which are sent here for the POST recipe after being told not to proceed to steps 2–7. One clause scopes it:

```suggestion
**Before submitting, say what the step 5 pass returned** — its confirmed findings, or "no findings". That statement is a compliance check: say it in the session, not the review body, so an empty-body APPROVE stays empty. The findings themselves still get folded into the review, per step 5. If you can't say, the pass didn't run — go back to step 5 and run it. A review that reaches this point without it is not submittable. This binds reviews that came through step 5; the step 1 incremental paths that skip steps 2–7 and jump here for the POST recipe have no pass to report, and approve without it.
```

generator/tests/test_cross_file_contracts.py
```diff
@@ -193,6 +193,19 @@ def test_review_approval_gates_on_author_stated_readiness() -> None:
     assert "Re-check the author-readiness gate" in skill
 
 
+def test_review_second_pass_is_a_submit_precondition() -> None:
+    """A full review cannot quietly skip the standalone second pass, including
+    after a safe re-target; the trivial-increment close-out remains exempt."""
```
tend-agent: `32a7df4` moved the assertion below from `"Step 1's trivial-increment close-out path"` to `"...trivial-increment and dedup close-out paths"`, but this line still names only the trivial-increment one. The docstrings in this file are what a later editor reads as the contract, so a change that re-narrowed the SKILL.md sentence back to one path would look consistent here and only the assertion would object.

```suggestion
    """A full review cannot quietly skip the standalone second pass, including
    after a safe re-target; both step-1 close-out paths remain exempt."""
```

tend-agent (commented):
The three points from the review on `cd91ef1` are closed: the gate now names the exempt close-out paths, the mid-review re-target bullet re-runs step 5, and `test_review_second_pass_is_a_submit_precondition` pins the header split. I re-walked every `step N` reference in the merged file against the new headers — they all resolve — and `TODO.md`'s "Revisit when one of those moves" paragraph is still the only cross-file referent, which this PR updates.

One thing `32a7df4` left behind: it widened the assertion to cover both close-out paths but not the docstring above it, so the test now documents a narrower exemption than it asserts. Suggestion inline.

PR #1081, opened 2026-08-29
Record what a run was about beside what it spent

max-sixty:
Every question worth asking of the fleet's spend is per-subject — cost per PR, repeat reviews of one branch, two agents racing on one commit — and `token-usage.json` recorded none of it. It carried `input_tokens`, `output_tokens`, the cache fields, `turns`, `model` and `cost_usd`, so each of those questions cost a hand-rolled join against `gh run list` and the API behind it.

## The record

`shared/steps/token_usage.py` now writes `repo`, `workflow`, `run_id`, `run_attempt`, `event`, `number` (the PR or issue) and `head_sha` alongside the counts, on both harnesses. Two of those are not on the run listing at all: the PR/issue number, and a `head_sha` read from the event rather than from the ref the run was queued on. The rest are there so the record stands on its own, read out of a gist or a downloaded artifact.

The subject fields follow one rule: an event's subject is a thread or a commit, and each field reports its own half. So a `pull_request_target` run records the PR's head rather than the default-branch commit `GITHUB_SHA` gives it, and a mention on a PR records the PR number with no commit rather than the default-branch tip it never touched. Every key resolves independently — the step is `if: always()` and may not fail the job.

The event-payload readers move from `_issue.py` to `_common.py` (`event_payload`, `dig`, `as_int`, `subject_number`, `subject_sha`), so the outage-issue row and the usage record resolve a trigger the same way.

## The report

`plugins/tend-ci-runner/scripts/token-report.sh` leads with cost, and the rollup tables sort by it — a cached input token is priced far below an output token, so the old cache-read ordering ranked volume rather than spend. A new table groups by subject, collapsing repeat work on one PR or commit into a single row with a run count. Runs that emitted no result event can't be priced, and booking them at `$0` buried them under that sort, so they get their own table ranked by tokens.

The script was also rebuilt around one canonical stream. The download loop writes one JSON line per job — the `token-usage.json` plus the run's row from `gh run list` — and two jq programs read it: jobs to runs to totals, and the summary. That replaced a per-run aggregate, an entry builder, a totals pass, a header pass, a `JQ_DEFS` shell variable concatenated into four table programs, `column -t`, three of its four bash counters, three bash conditionals for the footnotes, and two hardcoded empty-report literals.

<details>
<summary>Bugs found in review, each with a test that fails without its fix</summary>

- `head_sha` reported the default branch's tip for `issue_comment` and `repository_dispatch` — the commit a mention run leaves behind when it `gh pr checkout`s the PR head, so a mention record and a review record for the same PR revision named different commits.
- `as_int` coerced bools and floats. Anyone holding `contents: write` can POST a `repository_dispatch` with `client_payload.pr: true`, and `_issue.ref()` renders the result into a public issue comment — `#1` is a plausible reference to somebody else's PR.
- One unparseable `token-usage.json` aborted the whole report under `set -e`, losing every other run's data with it.
- An artifact holding no usable object put a null into the cost arithmetic and killed two tables after stdout had already been written. It is now skipped and counted: a torn upload's spend is unknown, not zero.
- `column -t` splits on whitespace, so a workflow name containing a space shifted every column right. The summary pads its own columns now.
- Cost truncated in the per-run row and rounded in the rollups, so a matrix run read a cent below the workflow total it was the whole of.

</details>

`conclusion` leaves the report's stdout JSON; nothing read it, and it was only there because the run row was merged in wholesale.

### DIFF
--- a/plugins/tend-ci-runner/scripts/token-report.sh
+++ b/plugins/tend-ci-runner/scripts/token-report.sh
@@ -4,16 +4,12 @@
 # Downloads session log artifacts from GitHub Actions, extracts each run's
 # token-usage.json (written by the "Token usage" step in each harness action),
 # and outputs a JSON report to stdout. A human-readable summary is printed to
-# stderr, grouped by workflow and by subject — the PR or issue the run was
-# about, else the commit.
+# stderr, grouped by workflow and by subject.
 #
-# Cost leads every table, and every table sorts by it. Tokens do not price
-# alike: cache reads are the great majority of the count and a small minority
-# of the bill, so a ranking by summed tokens ranks the cheapest work first.
-#
-# The subject columns come from the record itself, which carries repo,
-# workflow, run, event, PR/issue number and commit alongside the counts. Runs
-# whose artifact predates those fields report a "?" subject.
+# The two rollup tables — one row per workflow, one per subject — sort by cost.
+# A cached input token is priced far below an output token, so ranking by token
+# volume, which cache reads dominate, is not ranking by spend. The per-run
+# table below them stays newest-first.
 #
 # Usage: ./token-report.sh [HOURS] [PREFIX ...]
 #   HOURS: lookback period in hours (default: 168 = 7 days)
@@ -106,6 +102,11 @@ echo >&2 "Downloading artifacts for $RUN_COUNT runs..."
 ENTRIES="$WORKDIR/entries.jsonl"
 touch "$ENTRIES"
 
+# Runs that reach the totals with nothing. Counted rather than dropped
+# silently: a codex-harness repo would otherwise read a report of zero runs
+# with no line saying why, and the same silence covers a torn upload.
+SKIPPED=0
+
 mapfile -t ROWS < <(echo "$ALL_RUNS" | jq -c '.[]')
 for row in "${ROWS[@]}"; do
   RUN_ID=$(echo "$row" | jq -r '.databaseId')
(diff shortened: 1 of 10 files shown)


plugins/tend-ci-runner/scripts/token-report.sh
```diff
@@ -1,24 +1,28 @@
...
 #   TARGET_REPO - query a different repo (default: current repo)
 #
 # Requires: gh, jq
+#
+# Shape: shell talks to GitHub, jq does the arithmetic, and one canonical form
+# joins them. The download loop writes one JSON line per *job* — the
+# `token-usage.json` the "Token usage" step wrote, plus the run's row from
+# `gh run list` — and everything downstream reads that stream. So there are
+# three jq programs and no shell arithmetic: the loop's one-line stamp, the
+# report (jobs to runs to totals), and the summary.
```
tend-agent:
```suggestion
# three jq programs — the loop's one-line stamp, the report (jobs to runs to
# totals), and the summary — plus one shell counter, for the runs that reach
# the report with nothing.
```
"no shell arithmetic" isn't quite true: `SKIPPED=$((SKIPPED + 1))` in the download loop survived, and it's load-bearing — `--argjson skipped "$SKIPPED"` is what feeds the "uploaded no readable artifact" footnote. (The PR body's "three bash counters" is the same slip: three of four.)

tend-agent (approved):
One nit inline on the header comment; nothing blocking.

plugins/tend-ci-runner/scripts/token-report.sh
max-sixty: Right on both counts — fixed in b573e8c9. The header now names the surviving counter and says why the stream can't carry it: the runs it counts are exactly the ones that put no line in the stream. The PR description reads "three of its four bash counters".

PR #1082, opened 2026-08-29
Re-target a review when HEAD moves, instead of discarding it

max-sixty:
A review session that found HEAD had moved exited without posting, and the queued `synchronize` run reviewed the PR from scratch. Over a week across three repos that discarded 128 runs and $482 — more per run than a session that posts, because a superseded session has done the whole review by the time it checks.

## Re-targeting

The session now requires the commit it reviewed to be an ancestor of the live head, reads the delta, re-judges its findings against it, and posts anchored at the new head. Findings the delta left alone stand; ones it fixed drop out, and if that empties the review the session approves the new head. It finishes without posting only when it cannot judge the delta — a rewrite, or a review's worth of new code in its own right — and the queued run then covers that head in full, the same path that covers a session dying mid-review.

Three commands read the delta, because no one of them can show it:

- `git log -p --no-merges ... --not "$BASE_SHA"` — the author's own new code. A plain two-dot `git diff` between the heads presents everything a base merge dragged in as the author's; on a branch where they clicked "Update branch", three of main's files alongside the one commit they wrote.
- `git log --format='base merge: %h %s' --merges` — base merges, which the scoped log cannot show at all. Labelled, or it reads as one more commit at the end of the patch stream.
- `git show --cc <merge sha>` — what a conflicted merge itself changed, which the author commits *inside* the merge where neither log reaches it.

Re-verification of inline comments against the new `gh pr diff` is unconditional whenever a base merge appears: a clean merge that only shifts lines prints nothing in any of the three, and `--cc` shows only hunks differing from every parent, so a resolution taking the base side prints nothing either. `--cc` tells you what a merge changed, never that a merge changed nothing.

`suggestion` blocks are re-composed against the new head — one carried over unchanged reverts the author's newest edit on one click.

## Pinning

Every posted review now sets `commit_id`. Unpinned, GitHub anchors the review at whatever is live when the POST lands, so it claims code the session never read. Pinned, the anchor is what `bot-review-state.sh` reports as `LAST_REVIEW_SHA`, so the queued run finds that head reviewed and finishes without a second review of the same code.

The sha travels through a file because it cannot travel any other way: the agent composes the review body between reading the head and posting it, and shell state does not survive a tool call. Every recipe reads it into a variable and bails if it is absent — inlined as `$(cat ...)` a missing file substitutes the empty string and the request still goes out, which is the unpinned review the pin exists to prevent.

`weekly` needed the same pin — its approvals land on dependency PRs, the population `nightly` rewrites on purpose. Its file is per-PR and cleared before the guard, since step 2 loops over every dependency PR.

Before an APPROVE the session now requires the pinned commit to still be live. The rollup it reads is the live head's, so once the two differ the red-check gate is reading a different commit than the approval pins.

## Tests

`test_notification_queue_contract.py` becomes `test_cross_file_contracts.py` — it now pins skill-to-skill contracts as well as the notification queue's. All three delta commands and all three posting recipes are asserted there.

## Not in this change

A pre-boot workflow gate that would skip the queued run before it boots an agent. It is a compute-only mechanism and is proposed separately.


plugins/tend-ci-runner/skills/weekly/SKILL.md
```diff
@@ -40,14 +40,25 @@ If no dependency PRs are open, note "0 dependency PRs to process" and continue t
...
-     # package, bump type, and what you checked, e.g. "ruff 0.13 → 0.14 (patch),
-     # CI green, no API changes". Write it to a file rather than an inline
-     # --body: a package name written as inline code puts a backtick in a
-     # double-quoted argument, and bash runs the span as a command.
-     gh pr review <number> --approve --body-file /tmp/review-body.md
+     echo "$HEAD_SHA" > /tmp/checked-head
    fi
    ```
+
+   Then compose `/tmp/review-body.md` with the Write tool — one line naming the
```
tend-agent: The `else` branch no longer contains the approval, so nothing stops the skip path from reaching the POST below.

```suggestion
   If the guard printed "Already approved on this commit", move to the next PR;
   everything below runs only on the `else` branch. Otherwise compose
   `/tmp/review-body.md` with the Write tool — one line naming the
```

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -220,14 +228,32 @@ gh api "repos/$REPO/issues/<number>/reactions" -f content="+1"
...
+  # (or a head that won't fetch) leaves nothing to re-target onto.
+  BASE_SHA=$(gh pr view <number> --json baseRefOid --jq '.baseRefOid')
+  git fetch --no-tags --quiet origin "refs/pull/<number>/head" || true
+  git fetch --no-tags --quiet origin "$BASE_SHA" || true
+  git merge-base --is-ancestor "$HEAD_SHA" "$CURRENT_HEAD" 2>/dev/null \
+    || { echo "cannot re-target onto $CURRENT_HEAD — leaving it to the queued review"; exit 0; }
+  # The delta, scoped off base churn as in step 1: `--not "$BASE_SHA"` drops
+  # everything a base merge dragged in, which a plain `git diff` between the two
+  # heads would present as the author's new code.
+  git log -p --no-merges --format='%h %s' "$HEAD_SHA..$CURRENT_HEAD" --not "$BASE_SHA"
```
tend-agent, line 255: A base merge produces no output here at all, so the session can't tell it from an empty delta — and it's the one delta shape that re-scopes every inline comment's hunks.

```suggestion
  # The commits themselves, merges included: a base merge is invisible in the
  # patch below (`--no-merges` drops the merge, `--not "$BASE_SHA"` drops what
  # it dragged in), and it re-scopes every file's hunks.
  git log --format='%h %s' "$HEAD_SHA..$CURRENT_HEAD"
  git log -p --no-merges --format='%h %s' "$HEAD_SHA..$CURRENT_HEAD" --not "$BASE_SHA"
```

tend-agent (commented):
Two things in the new flow don't hold up, both in code the diff adds.

The larger one is that `weekly`'s already-approved guard no longer stops anything. Previously the `gh pr review --approve` lived inside the `else`; now the `else` only records the sha and the POST recipe sits below the block as unconditional prose, so the "Already approved on this commit; skipping" branch falls straight through into composing a body and approving. Since Step 2 iterates over every dependency PR, `/tmp/checked-head` is also whatever the *previous* PR wrote — so the skip path either re-approves a commit it just decided not to, or pins the approval to another PR's sha (and on the first PR of the run, to nothing at all, since the file doesn't exist yet).

The other is that the re-target delta can't show the case its own bullet asks about. `--no-merges` drops the merge commit and `--not "$BASE_SHA"` drops everything the merge dragged in, so on a pure "Update branch" push the command prints *nothing* — indistinguishable from "no new commits". I confirmed it on a scratch repo: the three-commit range (merge + two base commits) produces empty output. That's exactly the push the bullet below warns about — "A base merge re-scopes every file's hunks, so verify every comment when the delta contains one" — and the session has no way to know the delta contained one, so it carries every inline comment over onto hunks that moved.

One smaller observation, no change requested here: the Posting-mechanics block is now ~32 lines with real edge-case handling (rewrite detection, two fetch fallbacks, base-churn scoping). `running-in-ci` § "Scripts over prose recipes" draws the line at "needs edge-case handling or grows past a couple of dozen lines" — and the base-merge blind spot above is the kind of thing shellcheck-and-tested code in `plugins/tend-ci-runner/scripts/` surfaces and a prose recipe doesn't, since every session re-derives it.

The pinning half reads correct throughout — `commit_id` is set on all three posting recipes, `bot-review-state.sh`'s `at_head`/`fresh_approval_sha` get sharper with it, and the full Python suite passes (760) with no stale references to the renamed test module.

plugins/tend-ci-runner/skills/weekly/SKILL.md
max-sixty: Right on both counts, and the cross-PR contamination is the worse half. Fixed in 1af35a86: the file is now `/tmp/checked-head-<number>`, `rm -f`'d before the guard runs so the already-approved branch leaves nothing readable behind, and the prose says plainly that `skipping` ends this PR.

plugins/tend-ci-runner/skills/review/SKILL.md
max-sixty: Confirmed on a scratch repo — a pure "Update branch" push prints nothing between the markers while `git log --merges` shows the merge commit. Fixed in 1af35a86 by adding a second `git log --oneline --merges "$HEAD_SHA..$CURRENT_HEAD"` and rewriting the bullet to read the two outputs as a pair: an empty scoped log plus a non-empty merges log is the case that re-scopes every hunk.

On the scripts-over-prose point: agreed, and the two findings here are exactly what a tested script would have caught. Leaving it out of this change to keep it reviewable; raising it separately.

plugins/tend-ci-runner/skills/weekly/SKILL.md
tend-agent: Confirmed at 1af35a86: the `rm -f` runs ahead of the guard, so the skip branch leaves nothing readable, the name carries the PR number through step 2's loop, and the bold line ends the PR before the compose step. `test_weekly_approval_pins_the_commit_it_checked` asserts all three and is green on this head.

One thing in the new comment I can't confirm. It says reading a file this PR's check didn't write "fails the POST rather than mispinning it" — but `$(cat /tmp/checked-head-<number>)` on a missing file substitutes the empty string and the `gh api` call still runs, so the backstop rests on GitHub rejecting `commit_id=""` rather than treating it as absent. I haven't verified which it does, and the second case is the one this PR exists to remove: an unpinned APPROVE landing on whatever is live. Chaining makes it not depend on that answer:

```bash
CHECKED=$(cat /tmp/checked-head-<number>) || exit 0
gh api "repos/$REPO/pulls/<number>/reviews" --method POST \
  -f event=APPROVE -f commit_id="$CHECKED" -F body=@/tmp/review-body.md
```

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Reproduced the pair on a scratch repo and the fix does what it claims: after a base merge the scoped log prints nothing and `--merges` prints the merge, so the two outputs together separate an "Update branch" click from an empty delta.

The same repro turns up one case the bullet doesn't reach. Where the base merge conflicted, the author's resolution is committed *in the merge*, so both logs drop it: the scoped log excludes merge commits, and `--merges` reports that a merge happened, not what it changed. In my repro the reviewed line went `AUTHOR_BUG` → `RESOLVED_BY_AUTHOR` inside the merge and neither output showed it. The bullet then sends the session to re-verify every comment, but re-verification is defined a line above as re-anchoring — "on a file the delta didn't touch, the line is unchanged and the comment stands" — and the scoped delta touched nothing, so a finding the resolution already rewrote is posted at a head where it no longer holds.

`git show --cc "$MERGE_SHA"` is the missing content: on a resolved conflict it prints the resolution hunks, and on a clean auto-merge only the region that combines both sides, so it stays short either way. Reading it when `--merges` is non-empty makes the merge re-judgeable rather than only re-anchorable.

<details><summary>Repro</summary>

```bash
git init -q -b main . && printf 'line1\nline2\nline3\n' > x.txt && git add . && git commit -qm base
git checkout -qb feature && printf 'line1\nAUTHOR_BUG\nline3\n' > x.txt && git commit -qam "author work"
REVIEWED=$(git rev-parse HEAD)
git checkout -q main && printf 'line1\nMAIN_CHANGE\nline3\n' > x.txt && git commit -qam "main changes x"
NEWBASE=$(git rev-parse HEAD)
git checkout -q feature && git merge --no-edit main      # conflicts
printf 'line1\nRESOLVED_BY_AUTHOR\nline3\n' > x.txt && git add x.txt && git commit -q --no-edit
CUR=$(git rev-parse HEAD)

git log -p --no-merges --format='%h %s' "$REVIEWED..$CUR" --not "$NEWBASE"   # prints nothing
git log --oneline --merges "$REVIEWED..$CUR"                                 # prints the merge
git show --cc "$CUR"                                                         # prints the resolution
```

</details>

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -220,14 +228,36 @@ gh api "repos/$REPO/issues/<number>/reactions" -f content="+1"
...
+  git merge-base --is-ancestor "$HEAD_SHA" "$CURRENT_HEAD" 2>/dev/null \
+    || { echo "cannot re-target onto $CURRENT_HEAD — leaving it to the queued review"; exit 0; }
+  # The author's own new code, scoped off base churn as in step 1:
+  # `--not "$BASE_SHA"` drops everything a base merge dragged in, which a plain
+  # `git diff` between the two heads would present as the author's.
+  git log -p --no-merges --format='%h %s' "$HEAD_SHA..$CURRENT_HEAD" --not "$BASE_SHA"
+  # Base merges, which the scoped log above cannot show — it drops the merge
+  # commit and every commit the merge brought in. An "Update branch" click
+  # prints nothing there while re-scoping every file's hunks.
+  git log --oneline --merges "$HEAD_SHA..$CURRENT_HEAD"
```
tend-agent: The two logs print into one stream with nothing between them, so when both are non-empty the merge line lands flush against the end of the patch and reads as another commit in the scoped log:

```
6daa152 another author commit

diff --git a/f3 b/f3
...
+more
d1caa0d Merge branch 'main' into feature
```

That's the output the bullet below asks the session to read as a pair, and the case it has to catch — a base merge alongside the author's own commits — is exactly the one where the boundary is invisible. Labelling the second log costs nothing and makes it unmistakable in both directions (empty scoped log, or a merge glued to a diff):

```suggestion
  git log --format='base merge: %h %s' --merges "$HEAD_SHA..$CURRENT_HEAD"
```

(`--format` before `--merges` keeps `--merges "$HEAD_SHA..$CURRENT_HEAD"` contiguous, which the test assertion in the other comment matches on.)

generator/tests/test_cross_file_contracts.py
```diff
@@ -87,3 +91,53 @@ def test_installation_and_each_poll_enable_repository_watching() -> None:
...
+    assert 'echo "$HEAD_SHA" > /tmp/reviewed-head' in skill
+    assert 'echo "$CURRENT_HEAD" > /tmp/reviewed-head' in skill
+    # Read back by both posting recipes, never from a bare shell variable.
+    assert '-f commit_id="$(cat /tmp/reviewed-head)"' in skill
+    assert '--arg sha "$(cat /tmp/reviewed-head)"' in skill
+
+    # The delta is scoped off base churn; a plain two-dot diff between the two
+    # heads would hand the session everything a base merge dragged in.
+    assert "git log -p --no-merges" in skill
+    assert '--not "$BASE_SHA"' in skill
```
tend-agent, line 128: The weekly fix got an assertion (`rm -f /tmp/checked-head-<number>`) but the base-merge half didn't, so the `--merges` log is the one recipe here nothing pins. Deleting it leaves every assertion in this module green while the session silently goes back to reading a base merge as an empty delta — and carrying stale inline comments onto re-scoped hunks, which is the outward action this whole change exists to prevent.

```suggestion
    assert '--not "$BASE_SHA"' in skill
    # And the merges log beside it: the scoped log cannot show a base merge,
    # which re-scopes every inline comment's hunks.
    assert '--merges "$HEAD_SHA..$CURRENT_HEAD"' in skill
```

tend-agent (commented):
Both fixes hold. I re-ran the scratch-repo check on the base-merge case: with a `git merge` of the base tip between the two heads, the scoped log prints nothing and `git log --merges` prints the merge commit, so the pair does distinguish an "Update branch" click from an empty delta. Weekly's skip path now stops at the guard, and the per-PR filename removes the cross-PR contamination. Full Python suite green (760).

One thing the pair can't do as written, plus a gap in what the test pins — both inline.

plugins/tend-ci-runner/skills/weekly/SKILL.md
max-sixty: You are right and I had it backwards — verified that `$(cat /missing)` substitutes the empty string and the request still goes out. Took the chained form in b28cfbc4, and applied it to review's two recipes as well, which had the same hole. The comment that claimed otherwise is gone.

plugins/tend-ci-runner/skills/review/SKILL.md
max-sixty: Reproduced it: a reviewed line went `AUTHOR_BUG` → `RESOLVED_BY_AUTHOR` inside the merge, both logs silent, and `git show --cc` printed the resolution hunk exactly. Added in b28cfbc4 as its own bullet, framed as you put it — a conflicted merge is re-judgeable, not just re-anchorable, and a finding the resolution already fixed has to drop out.

Also took both suggestions: the merges log now carries a `base merge:` label, and the assertion is pinned. Worth noting why it was missing — I wrote that assertion in the previous round, but ruff-format rewrote the quotes it matched on and the replace silently no-op'd, so it never landed. Exactly the failure your comment describes, one layer up.

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -241,9 +275,22 @@ ALREADY_POSTED=$(jq -r --argjson force "${FORCE_FULL_REVIEW:-false}" '
...
+- Findings the delta left alone stand. Post them.
+- Findings the delta fixed drop out. If that empties the review and the delta itself reads clean, approve the new head: an empty-body approval is a verdict here, not the absence of one.
+- Finish without posting only when you can't judge the delta — it rewrites what you just reviewed, or it is a review's worth of new code in its own right. The queued run then reviews the new head in full.
+- Inline comments resolve against the commit the review pins, so re-verify each one against the current `gh pr diff`, which now returns the new head's. On a file the delta didn't touch, the line is unchanged and the comment stands. On one it did, move the comment to the line the code sits on now; where the line no longer falls inside a hunk, put the finding in the review body as a fenced quote with its path, as under **Recovering from inline comment 422 errors**.
+- **Read the two `git log` outputs as a pair.** The scoped one is the author's new code; the `base merge:` lines are base merges, and are the only place they appear. The two together distinguish an empty delta from an "Update branch" click.
```
tend-agent: The clause this replaced — "re-verify every comment even if the first printed nothing" — was the only thing overriding the shortcut two bullets up, and a clean base merge is exactly where that shortcut is wrong.

```suggestion
- **Read the two `git log` outputs as a pair.** The scoped one is the author's new code; the `base merge:` lines are base merges, and are the only place they appear. The two together distinguish an empty delta from an "Update branch" click. When a `base merge:` line appears, re-verify every inline comment against the new `gh pr diff` even if the scoped log printed nothing — the merge re-scopes hunks in files the scoped delta cannot show, so the "file the delta didn't touch" shortcut above does not hold.
```

tend-agent (commented):
The `--cc` bullet lands, but the same commit deleted the sentence that made re-anchoring happen after a base merge, so the clean-merge case now has no instruction at all.

Previously the pair bullet ended: *"A base merge re-scopes every file's hunks, so when the second output is non-empty, re-verify every comment even if the first printed nothing."* That clause is gone in b28cfbc4, and nothing replaced it — the new bullet's own title ("not just re-anchoring") presupposes re-anchoring, but the only re-anchoring rule left is the shortcut two bullets up: *"On a file the delta didn't touch, the line is unchanged and the comment stands."* After a base merge the scoped log cannot show the touched file, so that shortcut now fires on exactly the delta shape it was overridden for.

Concretely: bot comments on `foo.py:6`, author clicks "Update branch", main had added two lines above. Scoped log empty, `base merge:` line present, `git show --cc` prints the commit header and no hunks, and the flagged line is at 8. The session reads "the delta didn't touch it" and posts at 6.

`--cc` can't cover this on its own even when the merge conflicted. It shows only hunks differing from *all* parents, so a resolution that takes the base side prints nothing — I re-ran the earlier repro with `RESOLVED_BY_AUTHOR` swapped for main's own line and `--cc` emitted no hunks while the reviewed line was gone. Restoring the clause covers both; `--cc` still earns its place for judging what a conflicted merge changed.

Inline suggestion below. The rest of the increment holds: all three posting recipes read the sha into a variable and bail before the POST, the label makes the merges log unmistakable, and the module's assertions pin all three delta commands (`uv run pytest generator/tests/test_cross_file_contracts.py` green, 9).

<details><summary>Repro — clean base merge</summary>

```bash
git init -q -b main . && printf 'a\nb\nc\nd\ne\n' > foo.py && git add . && git commit -qm base
git checkout -qb feature && printf 'a\nb\nc\nd\ne\nFLAGGED\n' > foo.py && git commit -qam "author work"
H1=$(git rev-parse HEAD)
git checkout -q main && printf 'NEW1\nNEW2\na\nb\nc\nd\ne\n' > foo.py && git commit -qam "main adds lines above"
NEWBASE=$(git rev-parse HEAD)
git checkout -q feature && git merge --no-edit main    # clean
H2=$(git rev-parse HEAD)

git log -p --no-merges --format='%h %s' "$H1..$H2" --not "$NEWBASE"   # nothing
git log --format='base merge: %h %s' --merges "$H1..$H2"             # one line
git show --cc "$H2"                                                  # header only, no hunks
git show "$H1":foo.py | grep -n FLAGGED   # 6
git show "$H2":foo.py | grep -n FLAGGED   # 8
```

</details>

plugins/tend-ci-runner/skills/review/SKILL.md
max-sixty: Correct, and I reproduced both halves. A clean "Update branch" where main added two lines above the flagged one: scoped log empty, `--cc` empty, line moved 6 → 8 — the shortcut fires and posts at 6. And with the resolution taking the base side, `--cc` printed no hunks at all while the reviewed line was gone.

Took the suggestion verbatim in f6ad36bc. Reframed the `--cc` bullet to sit beside re-verification rather than in place of it, with the reason stated: `--cc` tells you what a merge changed, never that a merge changed nothing. Pinned the restored clause in `test_cross_file_contracts.py` so it can't be dropped again silently — which is how it went missing in the first place.

tend-agent (approved):

PR #1083, opened 2026-08-29
fix(review skill): re-compose suggestion blocks after a base merge

tend-agent:
#1082 made inline-comment re-verification unconditional whenever a `base merge:` line appears — the scoped log names no file on an "Update branch" click, while the merge re-scopes every file's hunks. The suggestion bullet one line down kept the narrow scoping: "on a file the delta touched" reads "the delta" as that same scoped log, so a base-merge-only push re-composes no suggestion at all. The block still holds the tree the session reviewed, so posting it at the re-anchored line writes base's merged-in lines back out on one click — the failure that bullet exists to prevent. Verified with `uv run pytest` (760 passed).

<details><summary>Repro — both merge shapes, scratch repo</summary>

Author edits `L7`; base then moves and the author clicks "Update branch". `HEAD_SHA` is the reviewed head, `CURRENT_HEAD` the merge.

Base adds two lines at the top, clean auto-merge:

```
git log -p --no-merges "$HEAD_SHA..$CURRENT_HEAD" --not "$BASE_SHA":  empty
git show --cc <merge>:                                               empty
foo.py at HEAD_SHA:      1 L1  2 L2  3 L3  4 L4  5 L5  6 L6  7 L7-AUTHOR  8 L8
foo.py at CURRENT_HEAD:  1 MAIN-NEW-A  2 MAIN-NEW-B  3 L1 ... 9 L7-AUTHOR  10 L8
```

Base edits a line inside the suggestion's range, still a clean auto-merge and still an empty scoped log:

```
content at HEAD_SHA lines 3-5:      L3 / L4          / L5
content at CURRENT_HEAD lines 3-5:  L3 / L4-MAIN-FIX / L5
```

A three-line suggestion composed at the reviewed head reproduces `L4`, so applying it deletes `L4-MAIN-FIX`. Nothing in the pre-change bullet asks for the re-compose, because the scoped log named no file.

The new assert sits beside the one #1082 added for the re-verify clause — that clause went missing silently once already, and the suggestion half has the same exposure.

</details>


generator/tests/test_cross_file_contracts.py
```diff
@@ -137,6 +137,10 @@ def test_review_skill_retargets_a_moved_head_rather_than_discarding_it() -> None
     # nothing anywhere, so the override below has to be unconditional.
     assert "git show --cc" in skill
     assert "even if the scoped log printed nothing" in skill
+    # Suggestion blocks carry the same blind spot as the inline comments
+    # above: a clean base merge shifts lines in files the scoped delta
+    # never touched, so re-composition is scoped to the merge, not to it.
```
tend-agent: `not to it` binds to `the merge`, the nearer noun, which inverts the sentence. Naming the delta explicitly removes the ambiguity.

```suggestion
    # never touched, so re-composition follows the merge, not the delta.
```

tend-agent (commented):
The fix is right: a clean base merge re-scopes hunks in files the scoped log cannot name, so a suggestion block composed against the reviewed head writes base's merged-in lines back out on one click. `uv run pytest generator/tests/test_cross_file_contracts.py` passes here and CI is green.

One structural point, not a defect. The base-merge override now lives in two bullets. The **Read the two `git log` outputs as a pair** bullet already exists to say that the narrow "file the delta didn't touch" scoping does not hold once a `base merge:` line appears, and this PR adds a second copy of that same exception to the suggestion bullet. Widening the existing override bullet to cover suggestion blocks alongside inline comments would keep the exception in one place, and the contract test could pin that one sentence. As it stands two bullets have to be kept in sync — which is the shape that produced this PR, since the preceding change widened one bullet and left its neighbour narrow. There is a fair case for the placement chosen here (a reader looking at suggestion mechanics finds the rule locally), so this is a judgment call rather than something to fix.

generator/tests/test_cross_file_contracts.py
tend-agent: Applied in 393538d — `not to it` did bind to `the merge`, inverting the sentence. `follows the merge, not the delta` names both nouns.

tend-agent:
Wording fix applied in 393538d. Keeping the two-bullet placement, for a reason narrower than locality: the duplication the review names is a consequence of each bullet stating its own scoping condition, and only one of those is removable.

The `git log` pair bullet scopes inline-comment *re-verification*; the suggestion bullet scopes *re-composition*, a different action with different mechanics (`git show "$CURRENT_HEAD":<path>` against the on-disk tree). The suggestion bullet has to open with some scope, and as long as that scope is "on a file the delta touched" it reads narrow at the point of use — so moving the base-merge exception into the neighbouring bullet would not consolidate it, it would recreate exactly the gap this PR closes, one bullet further away from the reader who acts on it. The only real consolidation is deleting the per-bullet scoping entirely and making re-composition unconditional, which costs a `git show` per suggestion on every push; that is a fair trade to weigh, but it is a different change from the one the review proposes.

What was missing was the sync check, not the single home. Both strings are now pinned by asserts sitting one line apart in `test_review_skill_retargets_a_moved_head_rather_than_discarding_it`, so widening one and leaving its neighbour narrow — the shape that produced this PR — fails the contract test rather than shipping silently.

max-sixty:
Closing as superseded by #1107. The useful rule here—recompose every carried
suggestion after a safe re-target—is now part of the canonical, tested
`review-preflight.sh` posting path. Keeping a second PR for the same review
mechanic would only create overlapping prompt changes.

PR #1087, opened 2026-08-30
fix(review): gate APPROVE on author-stated merge readiness, not just the draft flag

### LINKED ISSUE #1086: review: empty-body APPROVE fires on a PR whose description says it should not merge
On [numbagg#769](https://github.com/numbagg/numbagg/pull/769) the bot submitted an empty-body `APPROVED` on a PR whose own description said, verbatim, `This should not merge until two design choices are resolved`. Six hours later the maintainer closed it citing the first of those two exact choices. The approval was the only verdict standing on the PR at close time.

The review skill has this rule, but keys it on the *draft flag* rather than on stated readiness. Under **Draft PRs**: "Submit as **COMMENT only**, never APPROVE. GitHub blocks approving drafts, and the author hasn't asked for a verdict yet." numbagg#769 was `isDraft: false`, so that branch never engaged — and the two paths that did both push toward approval, neither of which reads the PR body:

- Step 5 (**Submit**): "If there are no issues, approve with an empty body — silence means correct."
- The incremental path: "if the most recent bot review was a COMMENT that flagged issues, and those issues are now addressed, submit an APPROVE with an empty body so the PR isn't left in limbo."

That second one is what fired here, and it's working as written: the bot raised code-level findings across three reviews, the maintainer pushed commits resolving each, and the bot approved because its own findings were closed out. The rule's notion of "resolved" is scoped to the bot's findings, and it has no term for the author's own open design questions — so a PR can be simultaneously "everything the reviewer raised is fixed" and "the author says this must not merge," and the skill only sees the first half.

The reasoning behind the draft rule — "the author hasn't asked for a verdict yet" — applies unchanged to a non-draft PR whose body withholds one. Draft status is a coarse proxy for author readiness that plenty of contributors don't use; here the author stated it in prose instead, and stated it in the very first paragraph.

Suggested fix, in the spirit of the existing draft carve-out: before an APPROVE, check whether the PR body (or a later maintainer comment) explicitly withholds merge readiness — "should not merge", "not ready", "do not merge", unresolved design questions the author enumerates. If so, submit COMMENT rather than APPROVE, and say which stated blocker is holding the verdict. That keeps the PR from reading as bot-approved while the author's own gate is open, and it costs one condition on the two approval paths above.

Worth noting the failure is one-sided in a way that argues for handling it: withholding an approval that was warranted is cheap and self-correcting (the next push re-reviews), while an approval on a PR the author said must not merge is a wrong outward signal that persists until someone notices.

PR: numbagg/numbagg#769, `isDraft: false`, opened 2026-08-29T16:08:52Z, closed **unmerged** 2026-08-29T23:57:20Z.

PR body, first line of the fourth paragraph:

> This should not merge until two design choices are resolved: the static trigger can under-fire on very long conditioned series, while the current stable fallback can inherit a first-pair outlier; and fixed-window float64 matrices remain inaccurate on large-offset data unless shifting is restored, which costs about 20% on dense rolling correlation versus this branch's unshifted head.

Maintainer's closing comment (2026-08-29T23:57:19Z):

> Closing rather than merging: the adaptive static-matrix trigger does not provide a defensible error bound as series length grows, while it can rescan too often on ordinary low-covariance data. We are splitting the numerical work into smaller function-family changes with explicit ordinary-path and protected-path benchmarks.

That is the first of the two design choices the PR body had already flagged as unresolved — not a new objection discovered after review.

Review sequence on the PR (`gh api repos/numbagg/numbagg/pulls/769/reviews`):

```
16:21:11  numbagg-bot  COMMENTED  3f2a5ba  (body 4616 chars — fixed-window float64 findings)
17:18:1x  max-sixty    COMMENTED  7dc4050  (3 inline-reply containers, empty bodies)
17:32:28  numbagg-bot  COMMENTED  7dc4050  (body 2066 chars — _initial_shift not pinned by tests)
17:44:16  numbagg-bot  COMMENTED  98bab2d  (body 3273 chars — old_obs/new_obs buffer claim)
17:47:0x  max-sixty    COMMENTED  5002b8e  (5 inline-reply containers, empty bodies)
17:50:01  numbagg-bot  APPROVED   5002b8e  (empty body)
```

The bot's three COMMENT reviews were all code-level and all addressed by the maintainer's commits; none of them engaged the two design questions the body raised. The approval follows directly from the incremental rule's "issues are now addressed" test.

</details>

tend-agent:
## Problem

The review skill decides whether to withhold an APPROVE from the `isDraft` flag alone. A non-draft PR whose body says, in prose, that it must not merge yet gets read as ready, and both approving paths — step 5's "no issues, approve with an empty body" and the incremental path's "your earlier findings are now addressed" — reach APPROVE without consulting the body. The incremental path is the sharper one: its notion of "resolved" is scoped to the bot's own findings, so a PR can be simultaneously "everything the reviewer raised is fixed" and "the author says this must not merge", and the rule as written only sees the first half.

Reported in #1086 with a verified case: numbagg/numbagg#769 (`draft: false`) opened with `This should not merge until two design choices are resolved`, drew three bot COMMENT reviews on code-level findings, and got an empty-body `APPROVED` at 17:50 once the maintainer's commits closed those findings out. It was closed unmerged six hours later on the first of the two design choices the body had already named. The approval was the only verdict standing on the PR at close time.

## Solution

Generalize the draft carve-out's reasoning — "the author hasn't asked for a verdict yet" — from the flag to the statement. Edits to `plugins/tend-ci-runner/skills/review/SKILL.md`, the rule stated once and pointed at from every approving path:

- **Step 5 (Submit)** carries the rule, directly above the POST recipe every approving path is sent to: a PR body or later author/maintainer comment withholding merge readiness draws a COMMENT naming the blocker, not an APPROVE. It says explicitly that the bot's own findings closing out doesn't clear it — the two conditions are independent — and bounds itself to one statement: the surrounding silence rules are keyed on threads, so a body-only COMMENT would otherwise restate the same blocker on every push.
- **Step 1's two approving sentences** each get a pointer, since that path jumps straight to the recipe without reading step 5's prose: the trivial-skip sentence (the path that fired) and the dedup rule's `resolves the last open one (then approve with an empty body)`, which reaches the same trigger from the substantive-incremental path.
- **The pre-APPROVE peek** in Posting mechanics re-checks the gate alongside the existing red-check gate, because a blocker can land in a comment after the review began.

The asymmetry is the argument for spending a condition here: withholding a warranted approval costs one re-review on the next push, while an APPROVE standing on a PR its author gated is a wrong outward signal that persists until someone notices.

## Testing

`generator/tests/test_cross_file_contracts.py` gains `test_review_approval_gates_on_author_stated_readiness`, which pins the rule, all three pointers, and the once-only bound — the split is exactly the drift this file exists to catch, since a later edit to either approval path can silently lose its route to the gate. Confirmed failing before the skill edit and passing after; full `uv run pytest` is green (761 passed), as is `pre-commit` on both changed files.

---
Closes #1086 — automated triage


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -80,7 +80,7 @@ git log --no-merges --numstat --format='%h %s' "$LAST_REVIEW_SHA..$HEAD_SHA" --n
...
+If `FORCE_FULL_REVIEW` is false and the incremental changes are trivial, skip the full review — go directly to step 7 to resolve any bot threads addressed by the new changes. After resolving threads: if the most recent bot review was a COMMENT that flagged issues, and those issues are now addressed, submit an APPROVE with an empty body so the PR isn't left in limbo — and the author-readiness gate under step 5 applies here too, since these are the bot's own findings closing out rather than the author's. Use the recipe under step 5, which pins the commit read here. Otherwise do not submit a new review — the existing one stands. Do NOT proceed to steps 2–6; finish. Rough heuristic: changes under ~20 added+deleted lines that don't introduce new functions, types, or control flow are typically trivial.
```
tend-agent, line 83: This is one of two approval sentences in step 1, and the sibling didn't get the pointer. A few paragraphs down, the dedup paragraph ends:

> A fresh review is warranted only when the incremental diff introduces a new finding, or resolves the last open one (then approve with an empty body).

That's the same trigger this sentence has — the bot's own last finding closing out — reached from the substantive-incremental path instead of the trivial one, and the PR's rationale for routing this sentence ("that path jumps straight to the recipe without reading step 5's prose") applies to it verbatim. On an author-gated PR where a push resolves the last bot finding without touching the stated design blocker, that's the sentence that fires, and it has no route to the gate.

The test's docstring says "Both approval paths ... have to reach the gate, so each carries a pointer to it", which reads as a complete enumeration and isn't one — worth adding the same trailing clause there and a matching assertion, so the contract covers both.

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -195,6 +195,8 @@ Flag duplicates — reuse is almost always better than a parallel implementation
 
 **If there are no issues, approve with an empty body — silence means correct.**
 
+**Unless the author withheld merge readiness.** When the PR body — or a later comment from the author or a maintainer — says the change should not merge yet — "should not merge until…", "not ready", design questions the author calls unresolved — the verdict is withheld the same way the draft flag withholds it, and plenty of contributors state it in prose rather than toggling draft. Submit COMMENT instead, naming the stated blocker that holds the verdict. Your own findings being closed out does not clear it: "everything the reviewer raised is fixed" and "the author says this must not merge" are independent conditions, and only whoever stated the blocker retracts it. The asymmetry is why this is worth a condition: withholding a warranted approval costs a re-review on the next push, while an APPROVE standing on a PR its author gated is a wrong outward signal that persists until someone notices.
```
tend-agent: Worth putting a "once" on this. As written, `Submit COMMENT instead, naming the stated blocker` fires on every pass that reaches step 5 with no findings, so a PR gated for days draws a fresh review restating the same blocker on each push.

The rules that would otherwise suppress that are all keyed on a *thread*: step 1's dedup ("the only outstanding concern is a still-unresolved **thread** from an earlier bot review") and the one-review-per-run paragraph ("a prior unresolved bot **thread** still stands, post nothing"). A body-only COMMENT naming an author's blocker is neither a thread nor a bot finding, so neither clause reaches it, and the unqualified `Submit COMMENT instead` reads as the governing instruction.

```suggestion
**Unless the author withheld merge readiness.** When the PR body — or a later comment from the author or a maintainer — says the change should not merge yet — "should not merge until…", "not ready", design questions the author calls unresolved — the verdict is withheld the same way the draft flag withholds it, and plenty of contributors state it in prose rather than toggling draft. Submit COMMENT instead, naming the stated blocker that holds the verdict; name it once, and on a later pass that finds nothing new stay silent rather than restating it. Your own findings being closed out does not clear it: "everything the reviewer raised is fixed" and "the author says this must not merge" are independent conditions, and only whoever stated the blocker retracts it. The asymmetry is why this is worth a condition: withholding a warranted approval costs a re-review on the next push, while an APPROVE standing on a PR its author gated is a wrong outward signal that persists until someone notices.
```

tend-agent (commented):
The rule itself is the right shape — a stated blocker withholds the verdict the way the flag does, and putting it directly above the POST recipe is the correct place for it. Two things on the routing.

The pointer went to one of step 1's two approval sentences. The trivial-skip sentence got it; the dedup rule's `(then approve with an empty body)` a few paragraphs down did not, and it is the same shape — an approve fired by the bot's own last finding closing out. The new test pins the asymmetry in place, so a later reader has a contract telling them one path is routed and no signal that the other isn't.

The gate also has no "once" on it, and the surrounding silence rules are all keyed on a *thread* rather than a review body, so a body-only COMMENT naming a blocker isn't covered by them.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Right — the dedup rule's sentence fires from the same trigger and had no route to the gate. Pointed it at step 5 in `acdd671`, and extended the contract test: the docstring now enumerates all three approving paths and there's a matching assertion on the new clause.

plugins/tend-ci-runner/skills/review/SKILL.md
tend-agent: Taken, with the reasoning trimmed into the sentence rather than left implicit: `name it once, and on a later pass that finds nothing new stay silent rather than restating it — the surrounding dedup rules are keyed on threads, so they don't reach a body-only COMMENT.` The test asserts `name it once` so a later edit can't drop the bound. In `acdd671`.

PR #1088, opened 2026-08-30
Skip a review run whose head already carries one

max-sixty:
Adds the pre-boot gate that lets a queued `tend-review` run exit without booting an agent when the live head already carries the bot's review.

This is the second half of #1082. That one made the review session re-target onto a moved head and pin `commit_id` on every review it posts, so a mid-review push is folded into what the session publishes. This one collects the saving: the queued replacement run reads the bot's own published reviews and skips a head one of them anchors.

## Not a reversal of #1078

#1078 removed a gate of the same shape. What was wrong with it was the *signal*, not the skip: it read a per-PR commit status the workflow wrote itself, which put a visible check row on every reviewed commit and — being a status — was writable by anything holding a token on the repo.

This reads a review instead. The bot has to publish one anyway, only the bot can write it, and it renders as the review it already is. `test_review_queues_pushes_behind_a_gate` asserts the commit-status producer stays gone.

## What it does and doesn't save

The concurrency group is per PR, so same-PR runs already serialize. This gate cannot reduce fan-out across PRs — it is orthogonal to multi-PR bursts. What it removes is the queued duplicate inside one PR's review window, plus the billed no-op on a reopen. Real and recurring, but a different population from any wide incident, and worth not overclaiming.

Per `CLAUDE.md` ("Agent-driven vs deterministic steps"): one trivial pre-check against a frequent no-op, ahead of an agent run it saves whole. 34 executable lines, 3 API calls on `synchronize` and zero on every other action, one branch per outcome, every one of them fail-open.

**The frequency, since the skill text here tells reviewers to demand one.** Over the last 120 `tend-review` runs (2026-08-22 → 08-30), 115 succeeded, with a median duration of 373s; 20 of them finished under 200s and 3 under 120s. A full pass doesn't complete in 200s, so that ~17% is the boot-and-exit population. Duration is a proxy — it doesn't separate an already-reviewed head from the other reasons a session exits early — so read it as the order of magnitude, not the exact saving.

`CLAUDE.md`'s "Development" section is edited here too. It named the no-op session among the cents-costing waste not worth a mechanism, which ruled this change out; an agent session costs dollars, so the exception belongs there, bounded by the bar the later section sets.

Two skills said the opposite and are corrected here — `review-reviewers` ("do not propose a skip-gate … to save the boot") and `review` ("challenge new mechanism whose only payoff is saved compute"). Both now carry the carve-out `CLAUDE.md` actually states rather than a blanket ban, and ask for the frequency number before accepting one.

## Behaviour

- `opened` and `ready_for_review` always run — no prior review, and `ready_for_review` sets the session's `FORCE_FULL_REVIEW`, which bypasses the already-reviewed check so a draft's COMMENT-only pass is redone in full.
- `synchronize` and `reopened` are judged. `reopened` is in the set because the session's own step 1 stops on an already-reviewed head, so admitting it bought a fully billed no-op.
- An action neither names falls through to a run, so widening `types:` in an override can't silently start skipping.
- Fails open on every API path, including a 200 carrying a non-JSON body and a payload that isn't a PR object.
- An anchor is discounted when a `head_ref_force_pushed` postdates it, and a tie resolves into a run.

## Also fixed here

- **A `setup:` step with its own `if:` no longer escapes the guard.** It previously warned and left the step ungated, which on a declined run left a `uses: ./…` resolving against an empty workspace — `Can't find 'action.yml'`, a red check on a job that correctly decided to do nothing. Its condition now narrows the guard instead of replacing it, fixing the same latent hazard in `tend-notifications`. Both sides are bracketed: `&&` binds tighter than `||`, and the notifications guard is itself a disjunction.
- **`bot-review-state.sh` counted unsubmitted reviews.** `GET /pulls/N/reviews` returns the caller's own `PENDING` reviews, and the caller is the bot. Their null `submitted_at` sorts below every timestamp, so the force-push discount dropped them only on a PR that had been force-pushed. A review nobody submitted could anchor a head.
- Skill prose that described the queued run booting and exiting at step 1 — it now usually never boots, so `review`, `running-in-ci` and `review-reviewers` say what actually happens and where to read a gate verdict.

## Tests

`uv run pytest` (787), `pre-commit run --all-files`, worker vitest and typecheck.

Every gate branch runs against a stubbed `gh` in `test_shared_steps.py`. The stub now honours `--paginate`, so the two pagination assertions have teeth; verified by mutation, along with the anchor reduction, the timestamp tie, and the fail-open on a malformed payload. `test_review_queues_pushes_behind_a_gate` now matches each step by role rather than by a uniform shape — `always()` is accepted only on the two cleanup steps, since on a working step it would drop GHA's implicit `success() &&` and boot the agent over a failed checkout — and rejects `continue-on-error` on the gate, which would turn a gate crash into a green job with the review silently skipped.


plugins/tend-ci-runner/scripts/bot-review-state.sh
```diff
@@ -21,6 +21,12 @@
 #   that no longer exists reports the current head, and `.commit_id` alone
 #   cannot tell an ordinary push from a rewrite. Anything submitted before the
 #   newest `head_ref_force_pushed` is discounted.
+
```
tend-agent: This line has no `#`, so it ends the header comment block and the new paragraph reads as a detached one sitting below it rather than as the third entry in the list. Every other separator in this header is a bare `#`.

```suggestion
#
```

Relatedly, the intro two paragraphs up — `Two GitHub behaviours make \`.commit_id\` alone unreadable, and every caller / that reads a review anchor has to account for both:` — now enumerates three. It's outside this hunk; I can push both fixes together if you'd rather not hand-edit.

generator/tests/test_generate.py
```diff
@@ -867,6 +942,39 @@ def test_mention_review_comment_listens_only_for_edits(tmp_path: Path) -> None:
...
+    data = yaml.safe_load(generate_review(cfg).content)
+    gate_step = next(
+        s for s in data["jobs"]["review"]["steps"] if s.get("id") == "gate"
+    )
+
+    assert gate_step["env"] == {
+        "GITHUB_TOKEN": f"${{{{ secrets.{BOT_TOKEN_SECRET} }}}}",
+        "PR": "${{ github.event.pull_request.number }}",
+        # `synchronize` is the only action the gate can skip; the rest ask for
+        # a pass whatever the head already carries.
```
tend-agent: The gate's `case` is `synchronize | reopened)`, and `test_review_gate_skips_a_reopen_of_an_already_reviewed_head` covers the second one — so this reads as if `reopened` were admitted unconditionally.

```suggestion
        # `synchronize` and `reopened` are the actions the gate can skip; the
        # rest ask for a pass whatever the head already carries.
```

tend-agent (commented):
Traced the gate against live data rather than only the stub: on this PR it emits `should_run=true`, and replaying its two reductions against #1082 (bot `APPROVED` at head, no force-push) yields a skip that matches `bot-review-state.sh`'s `at_head` on the same PR. The subset claim in `test_review_gate_skips_no_head_the_session_would_have_reviewed` holds structurally too — same `> $fp` strictness on both sides, and `max(anchors) > fp` is equivalent to "some anchor > fp", so the reduction difference can't make them disagree. 787 tests green, shellcheck clean on both scripts.

Two notes on the surrounding material.

**The frequency number the PR asks reviewers to demand.** The new `review-reviewers` text says "Count the shape's share of runs in the window before proposing anything; without that number, record and move on," and the `review` edit says "Ask for the frequency" — but the body doesn't carry one. Measuring the last 120 `tend-review` runs (2026-08-22 → 08-30): 24 finished under 200s and 6 under 120s, against a median of 365s. A full pass doesn't complete in 200s, so that ~20% is roughly the boot-and-exit population this removes. It supports the premise — and it's the number the next reviewer will be instructed to ask for, so it seems worth putting in the body.

**The rule this contradicts is still in `CLAUDE.md`, unedited.** Both skills got the carve-out, but `CLAUDE.md`'s "Development" section still reads: *"Complexity earns its place by preventing wrong outward actions … never by saving compute. Wasted compute (a no-op session, …) costs cents; the gate, retry wrapper, or scheduling arithmetic that would have prevented it has to be understood and maintained forever. Fix waste only when the fix is a simple knob."* It names the gate and the no-op session specifically. "Agent-driven vs deterministic steps," further down the same file, is the paragraph the PR cites, and it does carve this out. Since this PR already edits `CLAUDE.md`, reconciling the two there looks worthwhile — a reader who stops at "Development" gets a prohibition on exactly this change, and that's the paragraph a future review will quote back at the next gate proposal. Happy to push a cross-reference if you'd like one.

Nothing in the gate's behaviour itself looks wrong to me.

CLAUDE.md
```diff
@@ -249,16 +255,24 @@ Events pass through three layers before the bot does work:
...
+   whether the agent boots: mention's verify job checks engagement,
    notifications' check repairs repository watching then captures a paginated
-   cutoff snapshot.
+   cutoff snapshot, and review's gate skips a `synchronize` or `reopened`
+   whose head the bot has already posted a review at (or whose PR has closed).
 3. **Concurrency groups** — at most one running job per group.
 
+The numbers are a filtering order, not an execution order. Layer 2 lives
+*inside* the job that carries the `concurrency:` block, so layer 3 holds the
+run first and the pre-check evaluates only once the group releases it. For
```
tend-agent, line 266: Mention doesn't have this shape: `verify` is its own job with no `concurrency:` block, and `handle` carries the group behind `if: needs.verify.outputs.should_run == 'true'` — so its layer-2 check evaluates before the queue, not after it. Scoping the claim keeps the layer-2 bullet above and this paragraph consistent.

```suggestion
The numbers are a filtering order, not an execution order. Where layer 2 lives
*inside* the job that carries the `concurrency:` block — review's gate and
notifications' check — layer 3 holds the run first and the pre-check evaluates
only once the group releases it. Mention is the other shape: `verify` is its
own ungated job, so its pre-check decides on pre-queue state, and `handle`
carries the group. For
```

tend-agent (approved):
Both notes from the last pass are addressed — the frequency number is in the body now, and `CLAUDE.md`'s "Development" section carries the carve-out instead of a blanket prohibition. Re-checked the gate against the new commit: 787 tests green, shellcheck clean on `review-gate.sh` and `bot-review-state.sh`, and the `--paginate`-aware `gh` stub does give the two pagination assertions teeth. The `_setup_yaml` conjunction is bracketed correctly on both sides — I confirmed notifications renders `(A || B) && (own)` rather than `A || B && (own)`.

One inline suggestion, on the new paragraph under the three-layer list. "Layer 2 lives *inside* the job that carries the `concurrency:` block" holds for review and notifications but not for mention, which the bullet immediately above names as a layer-2 pre-check: `verify` is its own job with no `concurrency:` block, and `handle` is the one that carries the group (`if: needs.verify.outputs.should_run == 'true'`). So mention's pre-check decides *before* the queue, on pre-queue state — the ordering that makes its skill-level "dedup if the bot already responded" self-heal load-bearing rather than redundant. Worth keeping the generalization scoped, since this paragraph is what a future gate proposal will reason from.

Approving — that's a wording fix, not a hold.

max-sixty:
Closing in favor of the simpler pieces:

- #1107 folds PENDING-review handling into the canonical review-state resolver.
- #1106 preserves Tend's existing pre-check when an adopter adds a setup-step
  `if:` condition.

The proposed second pre-boot review gate is deliberately not retained. It
duplicates policy already owned by the review skill and adds workflow machinery
to avoid occasional compute, while Tend's design guidance reserves complexity
for preventing wrong outward actions.

PR #1089, opened 2026-08-30
chore: regenerate workflows with tend 0.1.21

max-sixty:
Release step 11: moves Tend's eight generated workflows and the
hand-maintained `review-reviewers.yaml` onto the published
`max-sixty/tend/claude@0.1.21` action. `init` rewrites only the generated
`tend-*.yaml` files, so the reviewer workflow is restamped by hand; every
workflow in the repo now agrees on one ref.

Two behavior changes ride along, both already merged into the generator
and released:

- **`tend-review`** drops its pre-boot gate and the `tend-review/<pr>`
  commit status the gate read, so reviewed commits no longer carry an
  extra check row. It gains `queue: max`, so a push cannot displace a
  pending `ready_for_review` run.
  ([#1078](https://github.com/max-sixty/tend/pull/1078))
- **`tend-notifications`** replaces the three-layer inbox sweep with a
  paginated snapshot taken at a 10-minute cutoff, repairs repository
  watching on every poll, passes the cutoff through to the agent, and
  serializes on a `tend-notifications` concurrency group.
  ([#1074](https://github.com/max-sixty/tend/pull/1074))

https://github.com/max-sixty/tend/compare/0.1.20...0.1.21

## Verification

`uvx tend@0.1.21 init` reruns to an unchanged generated diff; pre-commit
and `uv run pytest` (761 passed) are clean.

One caveat, and it is the reason the diff is not purely mechanical:
pre-commit's `trailing-whitespace` hook strips a 12-space line the
generator emits inside the notifications prompt. `block_prompt` renders
it with `indent(12, blank=True)`, which indents the blank line separating
the slash command from the cutoff line. #1074 introduced the first
multi-line prompt to go through that macro, so this is the first
regeneration to hit it. Both forms parse identically inside a YAML block
scalar, but the committed file can never match generator output until the
macro changes. The generator fix lands separately.


tend-agent (approved):
Regenerated with the in-tree generator — which is exactly 0.1.21, since the release commit is this PR's base — into a scratch tree and diffed all eight generated files against the committed ones. Byte-identical except the one line the description discloses. `0.1.21` resolves as a tag, and `grep -oh 'max-sixty/tend/[a-z-]*@[0-9.]*' .github/workflows/*.yaml | sort -u` is a single line, so the hand-maintained `review-reviewers.yaml` is restamped correctly. The dropped `tend-review/<pr>` stamp has no remaining reader: the 0.1.21 review skill no longer writes it, `test_generate.py` and `test_bot_review_state.py` assert its absence, and it isn't a required context.

One thing the "caveat" paragraph understates. The 12-space line isn't only a cosmetic divergence — it makes the nightly regen fail CI, starting the first night after this lands:

- `uvx tend@latest init` re-adds the trailing whitespace on `.github/workflows/tend-notifications.yaml` line 98.
- Nightly's stamp-only guard doesn't absorb it: `NON_STAMP_DIFF` counts both the `-` and `+` whitespace lines, so it's 2, and the PR gets created.
- That PR's `lint` job runs `pre-commit/action@v3.0.1`, which defaults to `--all-files`, so `trailing-whitespace` rewrites the file and fails the job.

The result is a valueless PR carrying a red check, re-opened every night at 06:17 UTC until the macro fix ships in a release — which puts the follow-up on a clock rather than in the backlog. Not a reason to hold this PR: leaving the workflows on a superseded ref is worse.

The generator side is smaller than "lands separately" suggests. [`macros.yaml.j2`](https://github.com/max-sixty/tend/blob/ee2dc9b4d87df0324355adae1e5429a1ff953bf8/generator/src/tend/templates/macros.yaml.j2#L269) renders `prompt|indent(12, first=True, blank=True)`; deleting `blank=True` leaves blank lines unindented, which is exactly the form committed here. I tried it locally — the full regtest set passes with **no recorded-output changes**, because pytest-regtest strips trailing whitespace when it compares, which is also why the `[notifications]` `.out` files on disk already have an unindented line 94. So it's a one-word deletion with no test churn.

Happy to open that PR if you want it queued behind this one.

PR #1090, opened 2026-08-30
fix(generator): stop padding blank lines inside block prompts

max-sixty:
`block_prompt` rendered its text through `indent(12, blank=True)`, which
pads a blank line out to twelve spaces. Adopters run pre-commit's
`trailing-whitespace` hook over their generated workflows, so the file
they commit could never match what `init` emits — the next regeneration
puts the padding back.

Nothing hit it until [#1074](https://github.com/max-sixty/tend/pull/1074)
gave `tend-notifications` the first multi-line prompt to go through the
macro. Every form here parses to the same string inside a YAML block
scalar, so this is churn, not a behavior change.

## Why it matters beyond the diff

`tend-nightly` regenerates with `uvx tend@latest init` and opens a PR when
`NON_STAMP_DIFF` is non-zero. That guard filters only the
`# Generated by tend X.Y.Z` header, so the `-`/`+` whitespace pair counts
2 and the PR gets created. That PR's `lint` job runs
`pre-commit/action@v3.0.1`, which defaults to `--all-files`, so
`trailing-whitespace` rewrites the file and fails the job — a valueless
red PR every night until this ships in a release.

## The whole class, not the one instance

Jinja's `indent` cannot render a block scalar without leaking one of
these, because it prepends the indent unconditionally when `first=True`:

| adopter `prompt:` | what leaks | which hook |
|---|---|---|
| `"alpha\n\nbeta"` | interior blank padded to 12 spaces | trailing-whitespace |
| `"\nalpha"` | blank *first* line padded | trailing-whitespace |
| `"alpha\n\n"` | file ends `\n\n` | end-of-file-fixer |
| `"alpha\n   \n"` | last line flattens to blank, file ends `\n\n` | end-of-file-fixer |

All four go through an `indent_block` filter that indents only non-blank
lines and clips trailing ones — which a `|` block scalar does anyway, so
no prompt value changes.

## Testing

The assertion is folded into the test that already guarded the
end-of-file-fixer hook, since both are the same failure: whitespace an
adopter's pre-commit rewrites.

That test reached seven workflows on one harness — `generate_all` skips
ci-fix without `watched_workflows` and install-test without its flag, and
codex was never exercised. It now covers all eight across both harnesses,
with a nightly prompt carrying leading, interior, and trailing blank lines
and a whitespace-only last line. Restoring `blank=True`, restoring
`first=True`, or narrowing the clip to `rstrip("\n")` each fails all four
parametrizations.

pytest-regtest normalizes trailing whitespace before comparing, so the
recorded outputs could not have caught this and do not change here.

### DIFF
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -98,11 +98,14 @@ def _indent_block(text: str, width: int) -> str:
     line even without it. An adopter's `trailing-whitespace` hook rewrites
     such a line, so the file they commit could never match what `init`
     emits — the next regeneration puts the padding back, and the PR the
-    nightly opens for it fails its own lint job. Trailing blank lines go for
-    the same reason; a `|` block scalar clips them, so they are pure churn.
+    nightly opens for it fails its own lint job.
+
+    Trailing blank lines go too, whether or not they carry spaces: they reach
+    the end-of-file hook instead of the trailing-whitespace one, and a `|`
+    block scalar clips them regardless, so they are churn either way.
     """
     pad = " " * width
-    return "\n".join(f"{pad}{line}".rstrip() for line in text.rstrip("\n").splitlines())
+    return "\n".join(f"{pad}{line}".rstrip() for line in text.rstrip().splitlines())
 
 
 _JINJA.filters["indent_block"] = _indent_block
--- a/generator/src/tend/workflows.py
+++ b/generator/src/tend/workflows.py
@@ -88,7 +88,25 @@ HEADER = f"""\
 # detail of that guarantee, and `tend check` creates and verifies it.
 TEND_ENVIRONMENT = "tend"
 
+
 # Available to every template without being passed to render().
+def _indent_block(text: str, width: int) -> str:
+    """Indent a multi-line prompt into a YAML block scalar at *width*.
+
+    Jinja's `indent` can't do this without emitting a whitespace-only line:
+    `blank=True` pads every blank line, and `first=True` pads a blank first
+    line even without it. An adopter's `trailing-whitespace` hook rewrites
+    such a line, so the file they commit could never match what `init`
+    emits — the next regeneration puts the padding back, and the PR the
+    nightly opens for it fails its own lint job. Trailing blank lines go for
+    the same reason; a `|` block scalar clips them, so they are pure churn.
+    """
+    pad = " " * width
+    return "\n".join(f"{pad}{line}".rstrip() for line in text.rstrip("\n").splitlines())
+
+
+_JINJA.filters["indent_block"] = _indent_block
+
 _JINJA.globals["header"] = HEADER
 _JINJA.globals["tend_version"] = _TEND_VERSION
 _JINJA.globals["tend_environment"] = TEND_ENVIRONMENT
--- a/generator/tests/test_generate.py
+++ b/generator/tests/test_generate.py
@@ -179,15 +179,20 @@ def test_no_restore_step_without_a_local_setup_action(tmp_path: Path) -> None:
 @pytest.mark.parametrize(
     "extra", ["", "setup:\n  - uses: ./.github/actions/tend-setup\n"]
 )
-def test_generated_workflows_end_with_exactly_one_newline(
+def test_generated_workflows_survive_the_whitespace_hooks(
     tmp_path: Path, extra: str
 ) -> None:
-    """A trailing blank line is pure churn in the adopter's regen diff, and the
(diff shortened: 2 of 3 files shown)


tend-agent (approved):

generator/src/tend/workflows.py
```diff
@@ -88,7 +88,25 @@
...
+    Jinja's `indent` can't do this without emitting a whitespace-only line:
+    `blank=True` pads every blank line, and `first=True` pads a blank first
+    line even without it. An adopter's `trailing-whitespace` hook rewrites
+    such a line, so the file they commit could never match what `init`
+    emits — the next regeneration puts the padding back, and the PR the
+    nightly opens for it fails its own lint job. Trailing blank lines go for
+    the same reason; a `|` block scalar clips them, so they are pure churn.
+    """
+    pad = " " * width
+    return "\n".join(f"{pad}{line}".rstrip() for line in text.rstrip("\n").splitlines())
```
tend-agent: The clip is `rstrip("\n")`, which stops at the first non-newline character — so a prompt whose final line is whitespace-only is never clipped. The per-line `rstrip()` then turns it into an empty line and the rendered block ends in `\n`. `block_prompt` is the last thing in `tend-nightly.yaml`, so that lands as a file-final blank line: the end-of-file-fixer failure the sibling assertion in this PR exists to prevent.

Plain `rstrip()` loses nothing here — every line is individually `rstrip()`ed already, so the only change is that whitespace-only trailing lines disappear rather than becoming empty ones. Leading and interior blanks are untouched.

```suggestion
    return "\n".join(f"{pad}{line}".rstrip() for line in text.rstrip().splitlines())
```

generator/tests/test_generate.py
```diff
@@ -176,18 +176,45 @@ def test_no_restore_step_without_a_local_setup_action(tmp_path: Path) -> None:
...
+    the nightly's carries the blank lines that a Jinja `indent()` pads: interior
+    ones under `blank=True`, and a leading one even without it.
+    """
+    extra += (
+        f"harness: {harness}\n"
+        "workflows:\n"
+        "  ci-fix:\n"
+        '    watched_workflows: ["ci"]\n'
+        "  nightly:\n"
+        '    prompt: "\\n\\nsweep the repo\\n\\nthen stop\\n\\n"\n'
```
tend-agent: Worth carrying a whitespace-only final line in the prompt too, so the trailing clip is guarded the way the leading and interior blanks now are. Apply this together with the `workflows.py` suggestion — on its own it fails all four parametrizations.

```suggestion
        '    prompt: "\\n\\nsweep the repo\\n\\nthen stop\\n\\n   "\n'
```

tend-agent (commented):
The class this closes has one member left, on the path where it costs the most.

`_indent_block` clips trailing lines with `rstrip("\n")`, so a prompt whose final line is whitespace-only slips through: the per-line `rstrip()` flattens it to an empty line, and since `block_prompt` renders last in `tend-nightly.yaml`, the file ends on a blank line — the end-of-file-fixer half of the same failure, reintroduced by the fix for the trailing-whitespace half.

It's reachable from ordinary adopter config, not just a hand-built escape sequence. A literal block scalar whose last line carries stray spaces:

```yaml
workflows:
  nightly:
    prompt: |
      sweep the repo
         
```

parses to `'sweep the repo\n   \n'`, and `generate_all` emits a `tend-nightly.yaml` ending in `\n\n`. Suggested one-token fix inline, plus the prompt extension that guards it.

Against the head commit, on this checkout:

- `_indent_block("alpha\n\n   ", 12)` returns `'            alpha\n\n'` — trailing blank retained. With `rstrip()` it returns `'            alpha'`.
- Rendering the config above through `Config.load` + `generate_all` gives `content.endswith("\n\n") == True`; with `rstrip()`, `False`.
- Leading and interior blanks are unaffected either way: `_indent_block("\n\nalpha\n\nbeta\n\n", 12)` is `'\n\n            alpha\n\n            beta'` in both.
- The suggested test change fails all four parametrizations against the current implementation and passes against `rstrip()`, so it is load-bearing in both directions.
- Full `uv run pytest` (763 tests) passes with the suggested change applied.

The rest of the change holds up under mutation: padding blank lines (restoring `blank=True`) and dropping the newline clip each fail all four parametrizations, so the new coverage is real rather than incidental. The other `indent(..., first=True)` call sites — `check_script`, `condition|trim`, `run_body` — are fed generator-owned text, so the prompt path really is the only adopter-supplied string reaching a block scalar.

</details>

tend-agent (approved):

PR #1093, opened 2026-08-30
Sync the review-reviewers matrix to consumers.json, and test the pair

tend-agent:
## Summary

`review-reviewers.yaml` is intended to cover every repository in
`data/consumers.json`, but three consumers were absent from its matrix. This
adds those repositories and a cross-file contract test so future omissions
fail loudly.

The contract is containment rather than equality: stale extra matrix entries
are harmless and fail visibly if the target no longer has Tend, while deleting
an entry because a best-effort consumer search missed it would silently reduce
fleet coverage.

The deterministic test owns this invariant; the weekly refresh skill does not
repeat it as a second, prompt-only rule.

## Verification

- `uv run pytest generator/tests/test_cross_file_contracts.py` (11 passed)


generator/tests/test_cross_file_contracts.py
```diff
@@ -159,3 +162,16 @@ def test_weekly_approval_pins_the_commit_it_checked() -> None:
...
+    """`review-reviewers` fans out over a hand-written matrix while
+    `data/consumers.json` is refreshed weekly by the `running-tend` sweep. A
+    repo that lands in the JSON and not in the matrix is simply never
+    analyzed, and the run still reports success — so the omission reads as a
+    clean fleet sweep rather than a partial one."""
+    workflow = yaml.safe_load(_read(".github", "workflows", "review-reviewers.yaml"))
+    matrix = workflow["jobs"]["review-reviewers"]["strategy"]["matrix"]["repo"]
+    consumers = [entry["repo"] for entry in json.loads(_read("data", "consumers.json"))]
+
+    assert sorted(matrix) == sorted(consumers)
```
tend-agent: `consumers.json` is rebuilt wholesale from a best-effort `gh search code`, so equality lets a search miss delete a matrix leg with CI green. Containment keeps the direction that matters — every consumer gets analyzed — and leaves a repo the search dropped in the fan-out.

```suggestion
    unanalyzed = sorted(set(consumers) - set(matrix))
    assert not unanalyzed, f"in consumers.json, absent from the matrix: {unanalyzed}"
```

tend-agent (commented):
The sync itself checks out — `diffplug/dormouse`, `jakobtfaber/Faber2026` and `max-sixty/leaf` are all public, unarchived, and carry `tend-*` workflows, so all three are genuinely analyzable.

One concern about the contract's direction. `data/consumers.json` isn't a maintained list — the weekly sweep rebuilds it wholesale from `gh search code`, and that search is best-effort. Running the recipe from `running-tend`'s "Weekly: refresh `data/consumers.json`" right now returns **6** repos, not 8, twice in a row: `max-sixty/cargo-affected` and `jakobtfaber/Faber2026` don't come back, though both are public and carry 8 `tend-*.yaml` workflows pinned to `max-sixty/tend/claude@0.1.14`.

So the next refresh that reproduces that result drops two entries from `consumers.json`, and `assert sorted(matrix) == sorted(consumers)` then requires deleting those two matrix legs for CI to go green. The test greenlights the shrink rather than catching it — the same failure the new workflow comment describes ("a repo missing here is simply never analyzed, and the run still reports success"), reached from the other side.

Containment keeps the property without that: every consumer needs a leg, while a matrix entry a search happened to miss survives. A leg for a repo that really did uninstall tend then fails loudly instead — `review-reviewers` errors when it can't resolve `bot_name` from the target's `.config/tend.yaml` — so the asymmetry runs the safe way. Inline suggestion on the assert; if you take it, the test name and the workflow comment want "covers" rather than "mirrors", while the `running-tend` sentence stays correct either way since an *added* repo is what needs the matrix entry.

Separately and out of scope here: that 6-of-8 result means `consumers.json` on `main` can shrink on a search blip, and it also feeds the site's data Worker. The refresh recipe looks worth a look on its own.

```
$ gh search code 'max-sixty/tend' --extension yaml --limit 100 --json repository,path \
    | jq -c '[.[] | select(.path | startswith(".github/workflows/tend-")) | .repository.nameWithOwner] | unique'
["PRQL/prql","diffplug/dormouse","max-sixty/leaf","max-sixty/tend","max-sixty/worktrunk","numbagg/numbagg"]
# identical on a second call; 52 results returned, so not the 100-result cap
```

`max-sixty/cargo-affected` and `jakobtfaber/Faber2026` both return `visibility: public`, `archived: false`, and 8 `tend-*.yaml` workflow files each.

The suggested assert passes `uv run pytest generator/tests/test_cross_file_contracts.py` (10 passed) and `ruff check` / `ruff format --check` at the pinned `v0.14.11`.

</details>

generator/tests/test_cross_file_contracts.py
tend-agent: Taken — applied in `181f231`, along with the "covers" rename you flagged: `test_review_reviewers_matrix_covers_consumers`, and the workflow comment now says an entry the JSON lacks is fine because its leg fails loudly if tend really is gone.

I reproduced the search result before switching: the recipe returns the same 6 repos here, missing `max-sixty/cargo-affected` and `jakobtfaber/Faber2026`. So equality really would have made the next refresh's PR red until someone deleted two legs, which is the failure this test exists to prevent, pointed the wrong way.

The `running-tend` sentence needed a small edit rather than none — as written it said a *changed* list needs a matrix update, which under containment is only true for an addition. It now says a new repo needs a matrix entry in the same PR, and a repo the search dropped needs no matrix deletion.

I left the refresh recipe alone and recorded the shrink risk in the PR description instead: whether a blip should be absorbed (union with the current list) or allowed to drop a genuinely-uninstalled repo is a maintainer call, and it also affects the site's data Worker, so it isn't this PR's to decide.

```
$ uv run pytest -q                                  # 764 passed
$ uvx pre-commit run --all-files                    # all hooks passed
```

Negative check, with one consumer's leg removed from the matrix:

```
AssertionError: in consumers.json, absent from the matrix: ['max-sixty/leaf']
```

</details>

tend-agent (commented):
The containment switch settles my earlier concern: the assert now reads the direction that matters and names the repos to add, and the matrix covers all 8 entries in `data/consumers.json`. Locally `uv run pytest` is 813 passed and `ruff check` / `ruff format --check` are clean at the pinned `v0.16.5`. (#1096 landing on main — the refresh now unions with the existing list rather than rebuilding wholesale — doesn't argue back toward equality: containment is also what keeps a matrix leg alive for a consumer any future search or API blip drops.)

One thing to confirm was intended. The base merge `3b3eca3` dropped the paragraph this PR added to `.claude/skills/running-tend/SKILL.md` — at the head commit that file matches `main`, with the "Weekly: refresh `data/consumers.json`" section ending at "one call covers the whole list" and no pointer to the matrix, so `gh pr diff` no longer shows the file at all. That merge also hand-trimmed the workflow comment and the test docstring, so this may be the same deliberate pass; I'm flagging it because SKILL.md didn't conflict there, so unlike the test file it wouldn't have surfaced during the resolution.

It isn't load-bearing either way — the assert message carries the instruction, and the weekly sweep polls its own PR's CI, so it sees the red and acts on it. Restoring it is a re-add of what `181f231` had:

```
A repo new to the list needs a `review-reviewers.yaml` matrix entry too; put
both in the one PR, or `test_review_reviewers_matrix_covers_consumers` fails
it. A repo the search dropped needs no matrix deletion — the test only requires
the matrix to cover the JSON.
```

- `git show 181f2310:.claude/skills/running-tend/SKILL.md` has the paragraph; `git show 3b3eca30:...` and `git show c32c58f8:...` do not, and neither does `git show bef86fa:...` (main). Read from git objects rather than the worktree, since a PR run restores `.claude/**` from the base branch.
- The workflow comment's "its leg fails loudly if tend really is gone" holds: `review-reviewers`' Step 1 exits 1 when `bot_name` can't be resolved from the target's `.config/tend.yaml`.
- No stale references to the old `test_review_reviewers_matrix_mirrors_consumers` name remain anywhere in the repo.
- The other open PRs touching `test_cross_file_contracts.py` (#1083, #1088, #1107) are unrelated contracts — merge conflicts, not duplicates.

</details>

max-sixty:
The removal was intentional. The cross-file containment test is deterministic,
runs in CI, and names every consumer missing from the matrix. Repeating the
same invariant in the weekly refresh skill would create a second rule to keep
in sync without making omissions harder to ship.

PR #1095, opened 2026-08-30
fix(generator): write the actionlint ignore for concurrency.queue

### LINKED ISSUE #1094: generator: `concurrency.queue: max` fails actionlint, so every adopter that lints workflows goes red on the 0.1.22 regen
Since 0.1.22 the generator emits `concurrency.queue: max` into `tend-review.yaml`. That's valid GitHub syntax, but actionlint's schema doesn't know the key, so every adopter that lints its workflows goes red the moment the nightly `uvx tend init` regeneration lands — the workflow update PR can't merge on its own CI.

tend's own repo is immune because its `.pre-commit-config.yaml` passes `-ignore 'unexpected key "queue" for "concurrency" section'` to the actionlint hook ([lines 27-35](https://github.com/max-sixty/tend/blob/e47ccc4480b3ce77a97fb3ee4ef2698e078728be/.pre-commit-config.yaml#L27-L35)). Nothing carries that to adopters — it isn't generated, isn't in the docs, and isn't mentioned in the 0.1.22 notes. PRQL hit it on [PRQL/prql#6252](https://github.com/PRQL/prql/pull/6252): two required checks red on the regen PR, `pre-commit.ci - pr` and `lint-megalinter / MegaLinter`.

The hook-args remedy is also incomplete as a general fix. It only reaches actionlint invoked through pre-commit; MegaLinter runs the same binary through its own `ACTION_ACTIONLINT` descriptor and stays red, as does a direct CLI call. A repo that lints workflows two ways needs the suppression twice, in two config dialects.

Suggested fix: have the generator write the ignore into `.github/actionlint.yaml`, which actionlint reads no matter who invokes it (`paths` + `ignore`, supported since v1.7.5):

```yaml
paths:
  .github/workflows/tend-*.yaml:
    ignore:
      - 'unexpected key "queue" for "concurrency" section'
```

Scoping the glob to `tend-*.yaml` keeps a genuine schema typo in a hand-written workflow failing. If merging into an adopter's existing `actionlint.yaml` is more coupling than it's worth, documenting the snippet in the upgrade notes would at least stop each adopter from re-deriving it.

Waiting it out isn't viable: v1.7.12 (2026-03-30) is still the latest release, [#657](https://github.com/rhysd/actionlint/issues/657) and [#654](https://github.com/rhysd/actionlint/pull/654) are open, and [#661](https://github.com/rhysd/actionlint/pull/661) was closed unmerged.

Against `actionlint v1.7.12` (release binary, `linux_amd64`) on the PRQL tree at the regen commit:

- `actionlint .github/workflows/*.yaml` with no config change → `.github/workflows/tend-review.yaml:22:7: unexpected key "queue" for "concurrency" section. expected one of "cancel-in-progress", "group" [syntax-check]`
- with the `paths` block above → exits 0, no diagnostics.
- negative control: injecting an unrelated `bogus-top-level-key: true` into `tend-review.yaml` with the block in place still fails (exit 1), so the ignore isn't swallowing other diagnostics in the matched files.

</details>

tend-agent:
## Problem

GitHub accepts `concurrency.queue: max`, but actionlint 1.7.12 rejects that
generated syntax. A repository-wide command-line ignore only helped Tend's own
pre-commit invocation, not adopters using actionlint directly or through other
linters.

## Solution

When `tend init` emits the review workflow, it merges a narrowly scoped ignore
into the adopter-owned `.github/actionlint.yaml` (or existing `.yml`). It:

- preserves comments-only and existing mapping configs
- leaves malformed or incompatible configs untouched with a clear warning
- matches `**/.github/workflows/tend-*.yaml` for root-relative, `./`, absolute,
  and subdirectory invocation forms
- does nothing when review is disabled, because no generated `queue` syntax is
  present
- remains idempotent and makes the nightly regeneration commit the config it
  writes

Closes #1094.

## Verification

- focused generator tests (33 passed)
- `wt test` (821 Python tests, 32 worker tests, TypeScript typecheck)
- real actionlint 1.7.12 probes for root-relative, `./`, absolute, and
  subdirectory paths


tend-agent (commented):
Two findings, both about whether the ignore actually reaches the adopters #1094 describes.

**The nightly regen never commits the file.** The regen step in [`plugins/tend-ci-runner/skills/nightly/SKILL.md`](https://github.com/max-sixty/tend/blob/main/plugins/tend-ci-runner/skills/nightly/SKILL.md) scopes all three of its path arguments to `.github/workflows` and `.config`: detection (`git status --porcelain .github/workflows .config`), the no-value gate (`NON_STAMP_DIFF=$(git diff --no-color .github/workflows .config ...`), and the commit (`git add -A .github/workflows .config`). `init` writes `.github/actionlint.yaml` outside both, so in the `/tmp/tend-update-workflows` worktree the file is created, left untracked, never staged, and then thrown away by the `git worktree remove --force` that ends the step.

That is the delivery path the issue is about — "every adopter that lints its workflows goes red the moment the nightly `uvx tend init` regeneration lands". `install-tend` stages with a bare `git add .`, so a fresh install picks the file up, but an existing adopter gets the regenerated workflow with no ignore beside it, exactly as today.

The commit line needs the path. Naming both filenames in one call aborts the whole `git add` when one is absent (`fatal: pathspec '.github/actionlint.yml' did not match any files`, nothing staged), so it has to be conditional:

```bash
for f in .github/actionlint.yaml .github/actionlint.yml; do
  [ -e "$f" ] && git add "$f"
done
```

The detection line wants the path too. `NON_STAMP_DIFF` is the awkward one: `git diff` cannot see a file that is not tracked yet, so a regen whose only change is the new config still reads as `0` and skips the PR. The release shipping this also bumps the action refs, so the first regen opens a PR regardless — but an adopter who later deletes the file would get no repair.

**The bail branches are the untested half.** The four `_bail` paths in `actionlint_config` are what stop `init` from rewriting an adopter's config into a shape they did not ask for, and none of the five new tests exercises one — fresh write, merge, idempotence, `.yml` in place, and `--dry-run` are all happy paths. Per CLAUDE.md complexity earns its place by preventing wrong outward actions, and clobbering an adopter's linter config is precisely that, so a case asserting the file is byte-for-byte unchanged and the warning is emitted (say `paths:` holding a list) is worth the few lines.

<details><summary>How the first finding was checked</summary>

Scratch repo with a tracked `.github/workflows/tend-review.yaml` and an untracked `.github/actionlint.yaml`:

- `git status --porcelain .github/workflows .config` → no output
- `git diff --no-color .github/workflows .config | wc -l` → `0`
- `git add -A .github/workflows .config` then `git diff --cached --name-only` → no output

The mechanism itself checks out otherwise: actionlint's `Config.PathConfigs` matches the `paths` key with `doublestar` against the path relative to cwd, and `IgnorePatterns` compiles each entry as a regex, so `.github/workflows/tend-*.yaml` and the unanchored message pattern both do what the diff intends. `ParseConfig` uses non-strict unmarshalling, so an adopter on actionlint older than v1.7.5 silently ignores the `paths` key rather than failing on it.

</details>

tend-agent:
Both findings addressed in 3037d53.

**The regen never committed the file.** Widened the three pathspecs in the nightly regen step from `.github/workflows` to `.github`, and moved `git add -A` ahead of the inspection so the stamp-only gate reads `git diff --cached`. The staging order is the part that matters: `git diff` cannot see an untracked file, so a regen whose *only* change is a fresh `.github/actionlint.yaml` read as no change at all — the deleted-file repair case you named, and the first-regen case on any adopter whose release brings no ref bump. `.github` avoids the conditional `git add`: it always exists, so one pathspec covers both the workflows and the config with no absent-path abort.

**The bail branches were untested.** Added `test_init_leaves_unmergeable_actionlint_config_untouched` — `paths:` holding a list, asserting the file is byte-for-byte unchanged, the warning is emitted, and the by-hand snippet is printed with it.

Scratch repo, tracked `.github/workflows/tend-review.yaml`, `init` writing only an untracked `.github/actionlint.yaml`:

- before — `git status --porcelain .github/workflows .config` → no output; `git diff --no-color .github/workflows .config` → `0` lines (PR skipped, file discarded with the worktree)
- after — `git add -A .github .config` then `git status --porcelain .github .config` → `A  .github/actionlint.yaml`; `NON_STAMP_DIFF` → `4` (PR opens)

Negative control for the gate the change moves: with the config already tracked and the only edit a `# Generated by tend 0.1.21` → `0.1.22` header bump, `NON_STAMP_DIFF` is still `0`, so a stamp-only regen still skips the PR.

`uv run pytest` → 769 passed; `pre-commit run --all-files` clean.

</details>

generator/src/tend/workflows.py
```diff
@@ -589,6 +589,98 @@ def generate_install_test(cfg: Config) -> GeneratedWorkflow:
...
+# caller (pre-commit's hook args miss MegaLinter's own actionlint, and both
+# miss a direct CLI call); `.github/actionlint.yaml` is read by the binary
+# itself, so it holds however actionlint is run. `paths` needs actionlint
+# >= v1.7.5.
+ACTIONLINT_QUEUE_IGNORE = 'unexpected key "queue" for "concurrency" section'
+
+# Scoped to the generated filenames — the same `tend-*.yaml` contract the
+# stale-file cleanup uses — so the same schema error in a hand-written
+# workflow still fails.
+ACTIONLINT_TEND_GLOB = ".github/workflows/tend-*.yaml"
```
tend-agent:
```suggestion
# Scoped to the generated filenames — the same `tend-*.yaml` contract the
# stale-file cleanup uses — so the same schema error in a hand-written
# workflow still fails. Matched against the path actionlint was handed, so
# a directory prefix would miss `./…` (find/xargs) and `../…` (subdir cwd).
ACTIONLINT_TEND_GLOB = "**/tend-*.yaml"
```

Verified against `actionlint v1.7.12`: with the prefix, `actionlint ./.github/workflows/tend-review.yaml` and a run from `generator/` both still report `unexpected key "queue"`; with `**/tend-*.yaml` every invocation form is clean and both negative controls still fail as they should.

generator/src/tend/workflows.py
```diff
@@ -589,6 +589,98 @@ def generate_install_test(cfg: Config) -> GeneratedWorkflow:
...
+    if existing is None or not existing.strip():
+        data = _YAML_BLOCK.load(
+            "# `concurrency.queue` is valid GitHub Actions syntax that\n"
+            "# actionlint's schema rejects. tend writes this ignore so the\n"
+            "# generated workflows lint clean; the glob keeps a real schema\n"
+            "# error in your own workflows failing.\n"
+            "paths:\n"
+        )
+    else:
+        data = _YAML_BLOCK.load(existing)
```
tend-agent:
```suggestion
    else:
        try:
            data = _YAML_BLOCK.load(existing)
        except YAMLError:
            data = None  # unparseable — the isinstance bail below reports it
```

A file that doesn't parse escapes all four `_bail` branches: the `ParserError` propagates out of `actionlint_config` and out of `init`, after the workflow files are written but before the stale-file cleanup. Needs `YAMLError` added to the `from ruamel.yaml import YAML` import. If you'd rather the warning name the real cause, hoist `_bail` above this block and call it with `"YAML"` here instead of falling through.

tend-agent (commented):
The regen fix checks out — I reproduced both the old and new behaviour in a scratch repo: with only an untracked `.github/actionlint.yaml`, the old `git status`/`git diff` pair reported no change and `NON_STAMP_DIFF` was `0`; staging first and diffing `--cached` gives `A  .github/actionlint.yaml` and `4`. A stamp-only header bump on a tracked config still gives `0`, so the no-value skip survives. The new bail-branch test passes.

Two findings on the generator half, both from running the pinned `actionlint v1.7.12` against the merged tree rather than from reading.

**A malformed `actionlint.yaml` crashes `init` rather than bailing.** The four `_bail` branches cover config *shapes* the merge can't handle, but not a file that doesn't parse — `_YAML_BLOCK.load(existing)` propagates `ParserError`/`ScannerError` straight out of `actionlint_config`, and nothing in `_update_actionlint_config` catches it. An adopter mid-rebase with conflict markers in that file, or a tab-indented one, gets a traceback from `tend init`; worse, the crash lands *after* every workflow file is written and *before* the stale-`tend-*.yaml` cleanup, so the repo is left half-regenerated. A parse error is the most likely form of "a config the generator does not understand", so it wants the same warn-and-leave the shape mismatches get. The inline suggestion routes it through the existing bail; that needs `YAMLError` on the `from ruamel.yaml import YAML` line, and you may prefer to hoist `_bail` above the load and pass it a `"YAML"` reason so the message says so rather than "unexpected top level".

`.config/tend.yaml` with `bot_name: test-bot`, `.github/actionlint.yaml` holding `paths:\n  - a\n  b: c`:

```
  wrote .github/workflows/tend-review-runs.yaml
CRASH: ParserError
```

All seven `tend-*.yaml` were on disk at that point; the cleanup pass and the summary never ran. Also raises on `paths: {unclosed` and on a tab-indented file.

</details>

**The `paths` glob is narrower than "however it is invoked."** `ACTIONLINT_TEND_GLOB` carries a directory prefix, and actionlint matches the key against the path string *as it received it*, not normalised to the project root — so the ignore silently stops applying for two ordinary invocations. Dropping the prefix to `**/tend-*.yaml` covers every form while keeping the `tend-*.yaml` filename contract that does the actual scoping, which is what README and CLAUDE.md already describe ("scoped to `tend-*.yaml`").

<details><summary>actionlint v1.7.12 runs</summary>

Against this branch's committed `.github/actionlint.yaml`:

| invocation | result |
|---|---|
| `actionlint` (no args) | exit 0 |
| `actionlint .github/workflows/tend-review.yaml` | exit 0 |
| `actionlint "$PWD/.github/workflows/tend-review.yaml"` | exit 0 |
| `actionlint ./.github/workflows/tend-review.yaml` | **exit 1** — `unexpected key "queue"` |
| `cd generator && actionlint ../.github/workflows/tend-review.yaml` | **exit 1** |

The `./`-prefixed form is what `find . -name ... | xargs actionlint` produces. With `paths: {'**/tend-*.yaml': ...}` all five are exit 0, and both negative controls still hold: the same `queue: max` in a `my-release.yaml` still fails, and an unrelated bogus top-level key in `tend-review.yaml` still fails.

</details>

The glob change also means regenerating the committed `.github/actionlint.yaml` and updating the four test assertions that spell out `.github/workflows/tend-*.yaml`. ruamel quotes the new key (`'**/tend-*.yaml':`) on dump, which actionlint parses fine — a bare `**/…` key would be a YAML alias error.

generator/src/tend/workflows.py
tend-agent: Applied in 651650d, with your comment text. I reproduced the five invocation forms against `actionlint v1.7.12` before and after: with the prefix, `./.github/workflows/tend-review.yaml` and a run from `generator/` both report `unexpected key "queue"`; with `**/tend-*.yaml` all five are exit 0, plus `find … | xargs actionlint`. Both negative controls still fail — the same `queue: max` copied into a non-`tend-` workflow, and a bogus top-level key in `tend-review.yaml`.

One thing the change dragged with it: the by-hand snippet in the `_bail` warning printed the key bare, and `**/tend-*.yaml:` unquoted is a YAML alias error, so an adopter pasting the suggested fix got a scanner error rather than a config. The snippet now quotes it, and the unmergeable-config test parses what it prints.

generator/src/tend/workflows.py
tend-agent: Applied in 651650d, taking the alternative you suggested: `_bail` is hoisted above the load and called with a reason naming the real cause, so the warning reads `has YAML the generator cannot parse` rather than reporting it as a shape mismatch. The other four call sites moved their `an unexpected …` into the argument to keep the sentence grammatical.

Confirmed end-to-end on your reproduction (`paths:\n  - a\n  b: c`): `init` now exits 0, the config is byte-identical, the seven workflow files land, and the summary prints — previously a `ParserError` traceback after `tend-review-runs.yaml` with the cleanup pass never reached. New test `test_init_leaves_unparsable_actionlint_config_untouched` covers it.

tend-agent (commented):
Close-out, no changes requested.

**Retracting the glob finding from my last review.** 516a5f1 is right and I was wrong. Re-ran the pinned `actionlint v1.7.12` against the merged tree with the committed `**/.github/workflows/tend-*.yaml`, and every invocation form is exit 0 — including the two I reported as broken:

| invocation | exit |
|---|---|
| `actionlint` (no args) | 0 |
| `actionlint .github/workflows/tend-review.yaml` | 0 |
| `actionlint ./.github/workflows/tend-review.yaml` | 0 |
| `actionlint "$PWD/.github/workflows/tend-review.yaml"` | 0 |
| `cd generator && actionlint ../.github/workflows/tend-review.yaml` | 0 |

Both negative controls hold, so the prefix isn't buying its scope back with a false negative: the same `queue: max` copied to `.github/workflows/my-release.yaml` still fails, and an extra bogus key under `tend-review.yaml`'s own `concurrency:` still fails. The directory prefix costs nothing and scopes tighter, and the comment on `ACTIONLINT_TEND_GLOB` describing that is accurate. Please read `**/tend-*.yaml` in my previous review as withdrawn rather than as an open thread.

The other finding is addressed — the parse-error path bails through the same warning and `test_init_leaves_unparsable_actionlint_config_untouched` covers it — and the `data is None` reparse added in 516a5f1 keeps a comments-only config's comments through the merge.

One observation, not something to act on: the merge round-trips the adopter's file through ruamel with tend's indent settings, so an existing config written with 4-space mappings or flush sequence indentation comes back reindented, and CRLF comes back as LF. Semantics are preserved; byte-level formatting isn't, which is a slightly stronger claim than README's "preserves the rest of the adopter-owned config" makes. For an eight-line linter config landing once, that seems like the right trade rather than something to engineer around.

<details><summary>How this was checked</summary>

`actionlint v1.7.12` from the release tarball, run against the merged tree (`a2a5c77`), which carries both the committed `.github/actionlint.yaml` and a `tend-review.yaml` with `queue: max` at line 22.

Negative controls:

```
$ cp .github/workflows/tend-review.yaml .github/workflows/my-release.yaml
$ actionlint .github/workflows/my-release.yaml
.github/workflows/my-release.yaml:22:7: unexpected key "queue" for "concurrency" section ... [syntax-check]

$ # bogus-key inserted under tend-review.yaml's concurrency:
$ actionlint .github/workflows/tend-review.yaml
.github/workflows/tend-review.yaml:23:7: unexpected key "bogus-key" for "concurrency" section ... [syntax-check]
```

Reformatting observation, calling `actionlint_config` directly:

| input | output |
|---|---|
| `ignore:\n    - 'x'` (flush sequence) | `ignore:\n      - 'x'` |
| `self-hosted-runner:\n    labels:\n        - my-runner` | reindented to 2/4 |
| CRLF line endings | LF |
| flow style (`{labels: [my-runner]}`) | unchanged |

`uv run pytest` → 821 passed. The repo's own `lint` job is green on this head, which is the root-relative row of the table above running for real.

</details>

PR #1096, opened 2026-08-30
skills(running-tend): stop the consumers refresh deleting repos the search missed

tend-agent:
The weekly `consumers.json` refresh rebuilds the file wholesale from `gh search code`, so the file inherits the search's recall. Today's search returned 6 of the 8 listed consumers, and running the recipe as written would have deleted `jakobtfaber/Faber2026` and `max-sixty/cargo-affected` — both live, both still carrying eight `tend-*.yaml` files — from the data the public site renders. This makes the search a discovery step only: candidates are the union of the search hits and the repos already listed, and a per-repo API check decides who stays.

<details><summary>Evidence and verification</summary>

**The miss is in the index, not the query or the cap.** The search returns 52 hits against a 100 cap, so nothing is crowded out; both repos simply return zero hits for `max-sixty/tend` while their workflow files demonstrably contain the ref:

```
$ gh search code 'max-sixty/tend' --repo max-sixty/cargo-affected
(no results)
$ gh api repos/max-sixty/cargo-affected/contents/.github/workflows/tend-review.yaml | ... | grep uses:
      - uses: max-sixty/tend/claude@0.1.14
```

`max-sixty/cargo-affected` returns nothing for a bare `tend` query either — the repo is absent from the code index, not stale in it. #1093 observed the same 6-of-8 result independently and named this as out of scope for that PR.

**Not a retry.** #345 was closed on the principle that a retry wrapper around a transient search blip isn't worth keeping. This adds no retry and re-queries nothing: it stops treating search absence as evidence of removal, and moves the keep/drop decision onto an authoritative per-repo API check that an uninstall actually fails.

**Verified.** The recipe was extracted verbatim from the file and run under `set -euo pipefail` against real GitHub: 6 discovered, 8 candidates, 8 retained, output byte-identical to the committed `data/consumers.json`. The drop path was exercised separately against a repo with no matching workflows (`octocat/Hello-World` → `count=0`, dropped). `uv run pytest` passes (763), as does `pre-commit` on the changed file.

**Out of scope, worth its own look:** `max-sixty/tend-integration` satisfies both filters — 3 `tend-*.yaml` files and `bot_name: tend-agent` — and is absent from the list only because code search does not index it either. If it is ever indexed, the weekly refresh adds the test fixture to the public site's adopter list.

</details>

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -134,16 +134,30 @@ because the workflow files are public.
 #    hits on `max-sixty/tend` itself crowd out tend's own workflow files
 #    past the 100-result cap, dropping tend from its own consumers.json.
 #    The `.github/workflows/tend-` path filter below bounds precision.
-mapfile -t REPOS < <(
+mapfile -t DISCOVERED < <(
   gh search code 'max-sixty/tend' --extension yaml --limit 100 --json repository,path \
     | jq -r '.[] | select(.path | startswith(".github/workflows/tend-")) | .repository.nameWithOwner' \
     | sort -u
 )
 
-# 2. Resolve bot_name from each repo's .config/tend.yaml.
+# 2. Union with the repos already listed. Code search recall is partial — a
+#    repo carrying a full set of tend-*.yaml files can return zero hits — so
+#    rebuilding from the search alone deletes live consumers from the file the
+#    website renders. The search finds *new* consumers; step 3 decides who stays.
+mapfile -t REPOS < <(
+  { printf '%s\n' "${DISCOVERED[@]}"
+    jq -r '.[].repo' data/consumers.json 2>/dev/null; } | sort -u
+)
+
+# 3. Keep a repo while it still has generated tend workflows, and resolve
+#    bot_name from its .config/tend.yaml. These are the authoritative checks:
+#    an uninstall drops out here, never by going missing from a search.
 mkdir -p data
 {
   for repo in "${REPOS[@]}"; do
+    workflows=$(gh api "repos/$repo/contents/.github/workflows" \
+      --jq '[.[] | select(.name | startswith("tend-"))] | length' 2>/dev/null) || workflows=0
+    [ "${workflows:-0}" -gt 0 ] || continue
     bot=$(gh api "repos/$repo/contents/.config/tend.yaml" --jq '.content' 2>/dev/null \
       | base64 -d 2>/dev/null \
       | yq '.bot_name // ""' 2>/dev/null)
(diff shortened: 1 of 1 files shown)


.claude/skills/running-tend/SKILL.md
```diff
@@ -134,16 +134,30 @@ because the workflow files are public.
...
+#    rebuilding from the search alone deletes live consumers from the file the
+#    website renders. The search finds *new* consumers; step 3 decides who stays.
+mapfile -t REPOS < <(
+  { printf '%s\n' "${DISCOVERED[@]}"
+    jq -r '.[].repo' data/consumers.json 2>/dev/null; } | sort -u
+)
+
+# 3. Keep a repo while it still has generated tend workflows, and resolve
+#    bot_name from its .config/tend.yaml. These are the authoritative checks:
+#    an uninstall drops out here, never by going missing from a search.
```
```suggestion
# 3. Keep a repo while it still has generated tend workflows, and resolve
#    bot_name from its .config/tend.yaml. An uninstall drops out here rather
#    than by going missing from a search — but so does a repo whose `gh api`
#    call hit a 403 or a 5xx, and nothing re-adds a repo the code index can't
#    see. Never land a removal without re-checking that repo by hand.
```

tend-agent (commented):
The union is the right shape — a search that returns nothing now degrades to a no-op instead of a wipe, and moving the keep/drop decision onto a per-repo API check is the part that makes the file authoritative rather than a projection of the code index.

One residual path, in the inline note: `|| workflows=0` treats *any* non-zero `gh api` exit as an uninstall, and with stderr discarded a 403 or a 5xx is indistinguishable from a real 404. For the two repos this PR exists to protect, that deletion is permanent — the code index can't see them, so nothing re-adds them on a later run, and re-running the recipe within the same session reads its own already-pruned output back as the union input. Classifying the error is more machinery than this is worth; the cheap version is to say so in the step-3 comment and require a hand re-check before a removal lands.

Separately, worth a look rather than a change here: the block is now 40 lines and carries edge-case handling (`|| workflows=0`, `${workflows:-0}`, the empty-array union), which is both triggers in `running-in-ci`'s "Scripts over prose recipes" — "put the logic in a tested script ... for a repo overlay a `scripts/` directory beside the skill". shellcheck on the extracted block is clean today, so nothing is broken; it's the trajectory that's notable, on the one recipe in this overlay that writes the data the public site renders.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied verbatim in c9e48fd. The swallowed-error path is the one that can permanently drop exactly the two repos this PR exists to protect, so naming it where the removal happens — plus the hand re-check requirement — is the right weight for it.

tend-agent:
On the script-extraction point: agreed on the trajectory, not doing it here. Extracting the block into `scripts/` beside the skill rewrites the same lines this PR changes, so it would either stack on an unmerged branch or have to be redone after a rebase — and it is a different concern from the recall bug. Leaving it as a follow-up once this lands: the recipe is 40 lines with three edge cases, it writes the file the public site renders, and a `scripts/refresh-consumers.sh` would get shellcheck and a test instead of being re-derived per session.

PR #1101, opened 2026-08-30
chore: bump ruff to v0.16.5, and clear its new default rules

tend-agent:
Moves `ruff-pre-commit` from v0.14.11 to v0.16.5, held out of #1099 because it isn't an ordinary rev bump: ruff 0.16.0 expanded the default rule set from 59 rules to 413, and 0.15.0 moved the formatter to the 2026 style guide. The repo carries no ruff configuration, so it takes those defaults directly — 64 findings appear on a codebase that reports `All checks passed!` under the old pin. Every change here is behaviour-preserving; the point is to move the pin without smuggling in a semantic change.

Three findings are suppressed rather than fixed, each with the reason inline. Each is in code where satisfying the rule would be a real change: the mitmproxy addon's root-logger call (`LOG015`) sits in the credential-injection path and follows mitmproxy's documented addon pattern, `preexec_fn` in the OAuth helper (`PLW1509`) is what gives the child its controlling terminal, and `UP017` in `shared/steps/_common.py` would rewrite `datetime.timezone.utc` to the 3.11+ `datetime.UTC` in a directory documented as 3.10-compatible (those step bodies run on the runner's system `python3`, which is 3.10 on an `ubuntu-22.04` `runs-on` override). Say the word and either can become a named logger / restructured spawn instead.

<details><summary>What the 64 findings were, and how each was handled</summary>

| Rule | n | Handling |
|---|---|---|
| `I001` unsorted-imports | 27 | ruff's own safe fix |
| `PLW1510` subprocess-run-without-check | 19 | explicit `check=False` at every site — this is the current behaviour spelled out, not a change to it |
| `RUF059` unused-unpacked-variable | 5 | ruff's fix (`rows` → `_rows`) |
| `UP037` quoted-annotation | 3 | ruff's fix; the file carries `from __future__ import annotations`, so unquoting is safe |
| `UP017` datetime-timezone-utc | 2 | one ruff's fix (a test); one suppressed in `shared/steps/`, reason inline |
| `PIE810` multiple-starts-ends-with | 2 | ruff's fix (tuple argument) |
| `FURB167` regex-flag-alias | 2 | ruff's fix |
| `PLR1730` if-stmt-min-max | 1 | ruff's fix |
| `PERF102` incorrect-dict-iterator | 1 | ruff's fix; the discarded key was genuinely unused |
| `LOG015` root-logger-call | 1 | suppressed, reason inline |
| `PLW1509` subprocess-popen-preexec-fn | 1 | suppressed, reason inline |

`PLW1510` has no autofix in ruff, so `check=False` was inserted at each call's closing paren via an AST pass that refused to write any file it couldn't re-parse, then normalised by `ruff format`.

**Formatting.** 14 files are reformatted. One is the 2026 style guide's own doing — it is the only file `ruff format --check` rewrites against `main` at v0.16.5. The other 13 are the `check=False` insertions being wrapped back onto their lines.

**Verification.** `pre-commit run --all-files` passes at the new rev, and `uv run pytest` passes (763), including the regtest snapshots — so no generated workflow output moved. `worker/` and `site/` are untouched.

**The alternative, not taken.** Pinning `select = ["E4", "E7", "E9", "F"]` would reproduce the pre-0.16 default exactly and drop the diff to one line. That trades a config-free repo for a frozen rule set, which reads like the wrong direction for a repo that has deliberately carried no ruff config — but it is a one-line change if you'd rather have it.

</details>

### DIFF
--- a/generator/tests/test_token_report.py
+++ b/generator/tests/test_token_report.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from typing import Any
 
 import pytest
+
 from tests import BASH, GH_PREAMBLE, fake_bin, tool_path
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
@@ -102,9 +103,7 @@ class Report:
             "USAGE_DIR": str(self._usage_dir),
         }
 
-    def add(
-        self, run_id: int, *, workflow: str = "tend-review", **over: Any
-    ) -> "Report":
+    def add(self, run_id: int, *, workflow: str = "tend-review", **over: Any) -> Report:
         """A completed run and the artifact it uploaded."""
         self._runs.append(
             {
@@ -119,13 +118,13 @@ class Report:
         )
         return self
 
-    def add_raw(self, run_id: int, body: str) -> "Report":
+    def add_raw(self, run_id: int, body: str) -> Report:
         """A run whose artifact holds *body* verbatim, valid JSON or not."""
         self.add_run_without_artifact(run_id)
         (self._usage_dir / f"{run_id}.json").write_text(body)
         return self
 
-    def add_run_without_artifact(self, run_id: int) -> "Report":
+    def add_run_without_artifact(self, run_id: int) -> Report:
         self._runs.append(
             {
                 "databaseId": run_id,
@@ -153,6 +152,7 @@ class Report:
             env=self._env,
             capture_output=True,
             text=True,
+            check=False,
         )
         assert result.returncode == 0, result.stderr
         self.stderr = result.stderr
@@ -204,7 +204,7 @@ def test_repeat_runs_on_one_subject_collapse_into_one_row(report: Report) -> Non
     for run_id in (1, 2, 3):
         report.add(run_id, cost_usd=2.0)
     report.add(4, number=852, cost_usd=1.0)
-    _, rows = report.run()
+    _, _rows = report.run()
 
     assert _table(report, "SUBJECT") == [
         ["#851", "3", "$6.00", "tend-review", "30K"],
@@ -221,7 +221,7 @@ def test_tables_are_ranked_by_cost_not_by_token_count(report: Report) -> None:
     """
     report.add(1, workflow="tend-nightly", cost_usd=0.5, cache_read_input_tokens=999999)
     report.add(2, workflow="tend-review", cost_usd=9.0, cache_read_input_tokens=1000)
-    _, rows = report.run()
+    _, _rows = report.run()
 
     workflows = [row[0] for row in _table(report, "WORKFLOW")]
     assert workflows == ["tend-review", "tend-nightly"], (
@@ -254,7 +254,7 @@ def test_a_run_whose_record_predates_the_subject_fields(report: Report) -> None:
     of the totals.
     """
     report.add(1, number=None, head_sha=None, repo=None, event=None)
-    output, rows = report.run()
+    output, _rows = report.run()
 
     assert output["runs"][0]["subject"] == "?"
     assert output["totals"]["cost_usd"] == 1.0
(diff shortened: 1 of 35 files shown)


shared/steps/_common.py
```diff
@@ -221,7 +221,7 @@ def utcnow() -> datetime.datetime:
     baseline range — is UTC, and a local-time reading would silently shift a
     day boundary the limits are scoped to.
     """
-    return datetime.datetime.now(datetime.timezone.utc)
+    return datetime.datetime.now(datetime.UTC)
```
tend-agent: `datetime.UTC` is 3.11+, and this module's docstring commits these step bodies to 3.10 so an adopter can pin an older `runs-on` image. `utcnow()` is on `rate_limit_preflight`'s main path, so this is an `AttributeError` before the agent boots rather than a degraded path.

```suggestion
    # These modules stay 3.10-compatible (see this module's docstring);
    # `datetime.UTC` is 3.11+.
    return datetime.datetime.now(datetime.timezone.utc)  # noqa: UP017
```

I checked this passes both `ruff check` and `ruff format --check` at v0.16.5.

tend-agent (commented):
One finding, in `shared/steps/_common.py`.

That module documents its own floor — *"an adopter can select an older image through the documented `runs-on` override, so these modules stay 3.10-compatible"* — and `UP017` rewrote `datetime.timezone.utc` to `datetime.UTC`, which is 3.11+. `utcnow()` is reached from `rate_limit_preflight.main()`, the first step of every job on both harnesses, so an adopter who has overridden `runs-on` to `ubuntu-22.04` (system `python3` is 3.10) gets `AttributeError: module 'datetime' has no attribute 'UTC'` before the agent boots. CI can't see it: the step bodies only ever execute under the dev env's 3.12.

Worth noting the mechanism, because it will recur. Ruff infers `target-version` from the nearest `requires-python`, which for `shared/steps/` is the workspace root's `>=3.12` — so every future `ruff --fix` will keep proposing 3.11+-only rewrites in a directory that has to run on 3.10, and only a reader who remembers the docstring will catch them. A `shared/steps/.ruff.toml` holding `target-version = "py310"` states the floor where ruff reads it; I confirmed the hierarchical config resolves for that path and that the `UP017` finding disappears under it. That is new config in a repo that has deliberately carried none, so the inline suggestion below is the minimal alternative instead — the same `# noqa` + inline-reason pattern this PR already uses for `LOG015` and `PLW1509`. Either closes this instance; only the config file closes the next one.

The rest of the behaviour-preserving claim holds where I traced it: `check=False` matches the implicit default at all 19 sites, and the two runtime insertions in `sandbox_setup.py` were already branching on `returncode`; `UP037`'s unquoting is safe because `test_token_report.py` carries `from __future__ import annotations`; `PLR1730`'s `max(since, at)` keeps the original tie-break; `PIE810` and `FURB167` are constant-for-constant. Nothing in the repo manipulates `sys.path` and none of the reordered modules has an import-time side effect, so the `I001` churn is inert.

generator/tests/test_token_report.py
```diff
@@ -204,7 +204,7 @@ def test_repeat_runs_on_one_subject_collapse_into_one_row(report: Report) -> Non
     for run_id in (1, 2, 3):
         report.add(run_id, cost_usd=2.0)
     report.add(4, number=852, cost_usd=1.0)
-    _, rows = report.run()
+    _, _rows = report.run()
```
tend-agent: With `rows` unused, the whole tuple is now discarded, so the call is only being made for its side effect — saying that directly beats two throwaway names.

```suggestion
    report.run()
```

Same shape at lines 224 and 334.

shared/steps/_common.py
tend-agent: Applied as suggested in c01842f — `datetime.timezone.utc` restored with the `# noqa: UP017` and the reason inline, matching the `LOG015`/`PLW1509` pattern already in this PR. `pre-commit run --all-files` and `uv run pytest` (763) pass at the new rev.

Left the `shared/steps/.ruff.toml` out deliberately: it's the better fix for the *next* instance, but it's the first ruff config in a repo that has carried none, so it's a maintainer call rather than something to fold into a version bump. @max-sixty — say the word and it's a one-line file (`target-version = "py310"`) replacing the `noqa`.

generator/tests/test_token_report.py
tend-agent: Applied in c01842f at all three sites (207, 224, 334). The two that still use the first element (257, 357) now read `output, _`, which is the idiom already in this file, rather than ruff's `_rows` rename.

PR #1103, opened 2026-08-30
fix(generator): let an adopter's prompt span more than one line

max-sixty:
`prompt:` is a documented per-workflow override, but three of the shapes it rendered into could hold only one line. Each wrote a workflow file GitHub rejects, from a config that is valid YAML in the adopter's own hands — and `init` never parses what it writes, so the adopter's first sign of trouble was GitHub refusing to load the workflow.

- **review** put the prompt inside a GitHub Actions `format('...')` expression in a folded scalar, where a newline ends the string literal.
- **ci-fix** used a block scalar whose 12-space indent the template wrote by hand, reaching only the first line.
- **`block_prompt`** — which triage, nightly, weekly, notifications and review-runs already used — emitted a bare `prompt: |`, so YAML infers the block's indent from the first line. A prompt whose *first* line is itself indented made every later line look like the end of the scalar. This one predated the other two and hit six workflows.

Reviewing that fix turned up the same bug three lines above it: `sandbox_path` and `sandbox_setup` render adopter-supplied lists into a hand-written `|` block scalar and break the same way on an indented first entry.

## The change

`block_prompt` becomes `block_input(name, text)`, and all four adopter-supplied inputs render through it with an explicit `|2` indentation indicator — the indentation is now written once rather than at each call site. `sandbox_env` keeps its newline rejection, but for the reason that actually applies now: the action reads one `NAME=VALUE` pair per line, not because the scalar would break.

That let review substitute `{pr_number}` with `${{ github.event.pull_request.number }}` directly, the way triage has always handled `{issue_number}`, which retires `_escape_braces` and the `format()` machinery with it. That's a behaviour improvement as well as a deletion: braces used to be doubled for `format()`, and *only* when the placeholder happened to be present, so the same prompt shipped different text depending on an unrelated part of itself. Now braces and quotes are the adopter's own text in every workflow.

Also from the review:

- `workflows.mention.prompt` was accepted and silently dropped, while the config docs promise every workflow takes a `prompt`. mention composes its prompt from the triggering event, so there is nothing an override could replace — `Config.load` now refuses it. A non-string `prompt` raises a `ClickException` instead of an `AttributeError` out of the renderer.
- A blank `prompt` came in three shapes, none of which did what the adopter wrote. A whitespace-only value is truthy, so it beat the default and left the agent step with no instructions — and the harnesses disagreed about that: the Claude action fails on an empty `TEND_PROMPT`, while the Codex action hands it to `codex exec` and boots an agent with nothing to do. `""` and a bare `prompt:` are falsy and reached the default instead, silently ignoring the key; the bare form also tripped the type check and reported `got NoneType`. `Config.load` refuses all three. The prompt checks are also ordered so the type check runs first.
- The docs named only `{pr_number}`, leaving `{issue_number}` and `{run_id}` undocumented, and said nothing about multi-line prompts. They now also warn that a `${{ ... }}` written as prose will be evaluated.

## Verification

<details><summary>What was checked, and how</summary>

- Each of the three broken shapes fails the new matrix test when restored — the `block_prompt` one across all seven prompt-bearing workflows, the sandbox one across both levers.
- The matrix is pinned to `GENERATORS`, so a new prompt-bearing workflow cannot join and skip it; mention's exclusion is asserted rather than merely absent. It runs on both harnesses and checks each placeholder resolves to its own event rather than to any event.
- New coverage for what the matrix can't reach: the sandbox levers, the ruamel re-serialisation path that `workflow_extra`/`jobs` overrides take, and the blank prompt the config now refuses.
- Driving the real `init` CLI over a config with multi-line prompts produces correct nested indentation, preserved quotes, and ci-fix's run coordinates appended after the adopter's text.
- `actionlint` on the generated set reports nothing but the known `concurrency.queue` complaint (#1094).
- On `|2` reaching GitHub's own parser: 128 public workflow files use `run: |2`, and `microsoft/onnxruntime-genai`'s is `active` with a successful run — so the indentation indicator is accepted by the same YAML front-end that reads `with:`.

</details>

The only change to tend's own generated workflows at the next regeneration is `prompt: |` → `prompt: |2` on six files, plus review's prompt shape.


generator/tests/test_generate.py
```diff
@@ -527,52 +527,145 @@ def test_custom_prompt(tmp_path: Path) -> None:
...
+    }
+    assert "timeout-minutes: 240" in workflows["tend-review.yaml"], "overrides skipped"
+    assert agent_prompt(workflows["tend-review.yaml"]) == body
+
+
+def test_whitespace_only_prompt_yields_an_empty_block(tmp_path: Path) -> None:
+    """`indent_block` strips a whitespace-only prompt to nothing, leaving a block
+    scalar with no body — the most exotic YAML the generator emits. It parses to
+    the empty string, which the action will reject as a missing input; what
+    matters here is that the file itself stays loadable.
```
tend-agent: The "action rejects it" reasoning only covers one harness, so it reads as a stronger safety net than it is. `run_claude.py` lists `TEND_PROMPT` in `_common.require_env`, which fails on unset *or empty* — so Claude does reject it. `codex/action.yaml` has no equivalent: it sets `PROMPT: ${{ inputs.prompt }}` and passes `"$PROMPT"` straight to `codex exec`. Pre-existing either way (the old `block_prompt` emitted the same empty block), so nothing to fix in this PR — but if the docstring is the record of why `Config.load` doesn't reject a blank prompt alongside the non-string one it now does, it's worth it being accurate.

```suggestion
    scalar with no body — the most exotic YAML the generator emits. It parses to
    the empty string, which the Claude action rejects (`require_env` fails on an
    empty `TEND_PROMPT`) but the Codex action hands to `codex exec` as-is; what
    matters here is that the file itself stays loadable.
```

tend-agent (approved):

generator/tests/test_generate.py
max-sixty: Confirmed both halves, and took the fix further than the docstring.

`require_env` fails on unset *or* empty (`shared/steps/_common.py:37`), so Claude rejects it; `codex/action.yaml:201,216` hands `"$PROMPT"` straight to `codex exec`, so Codex boots an agent with no instructions. Rather than record that asymmetry accurately in a docstring, `Config.load` now refuses a blank `prompt` outright — it is truthy enough to beat the default and empty enough to strip to nothing, so it is a typo either way, and the config is the one place that covers both harnesses. That retires the empty block scalar rather than testing it.

Also reordered the prompt checks: a non-string prompt on `mention` was reaching the mention branch before anything established it was a string.

generator/src/tend/config.py
```diff
@@ -514,9 +516,38 @@ def load(cls, path: Path | None = None) -> Config:
                         f"allowlist and likely won't apply to {wf_harness}. "
                         f"Set `workflows.{name}.model:` to a valid {wf_harness} model."
                     )
+                wf_prompt = wf_raw.get("prompt", "")
+                if not isinstance(wf_prompt, str):
```
tend-agent, line 522: `prompt:` written with nothing after it is the likeliest way an adopter ends up blank — a key typed ahead of the text, or a value commented out — and it is the one shape the new message doesn't reach. YAML parses it to `None`, so it trips the type check first:

```
workflows.triage.prompt must be a string, got NoneType
```

That names a Python type and points at "make it a string", when the remedy the adopter wants is the one three lines below: drop the key.

```suggestion
                wf_prompt = wf_raw.get("prompt", "")
                if wf_prompt is None:  # `prompt:` with nothing after it
                    wf_prompt = ""
                if not isinstance(wf_prompt, str):
```

generator/src/tend/config.py
```diff
@@ -514,9 +516,38 @@ def load(cls, path: Path | None = None) -> Config:
...
+                        "composes its prompt from the triggering event. Put "
+                        "standing guidance in the repo's `running-tend` skill "
+                        "overlay instead."
+                    )
+                # A blank prompt is truthy enough to beat the default and
+                # empty enough to leave the agent step with no instructions.
+                # The harnesses disagree about that: the Claude action fails
+                # on an empty `TEND_PROMPT`, while the Codex action hands it
+                # to `codex exec` and boots an agent with nothing to do. It is
+                # a typo either way, so refuse it here and let neither decide.
```
tend-agent: `""` isn't truthy, so it doesn't beat the default — every call site reads `wf.prompt or eff.default_prompt(...)`, which falls through to the default for it. Only the whitespace-only form is truthy-and-blank. Both are worth refusing, but as written the rationale covers one of the two shapes the test parametrizes, and the same claim carries into the test docstring ("A blank prompt beats the default (it is truthy)") and the PR body.

```suggestion
                # A whitespace-only prompt is truthy enough to beat the default
                # and empty enough to leave the agent step with no instructions.
                # The harnesses disagree about that: the Claude action fails on
                # an empty `TEND_PROMPT`, while the Codex action hands it to
                # `codex exec` and boots an agent with nothing to do. `""` is
                # falsy, so it reaches the default rather than either of those —
                # but it is as much a typo, so refuse both here.
```

tend-agent (commented):
Two notes on the new blank-prompt refusal. Neither blocks — the config is the right place for the check, and the earlier approval stands.

generator/src/tend/config.py
max-sixty: Taken — `prompt:` with nothing after it now normalizes to `""` before the type check, so it gets the blank message ("drop the key") rather than `got NoneType`. Added to the test's parametrize as a third shape, and mutation-checked: without the coercion that case fails.

generator/src/tend/config.py
max-sixty: Correct, and the claim was mine rather than the code's — `""` is falsy, so `wf.prompt or eff.default_prompt(...)` reaches the default for it. Only the whitespace-only form is truthy-and-blank.

The comment now separates the three: whitespace-only beats the default and leaves the agent step empty (Claude fails on it, Codex runs it); `""` and a bare `prompt:` are falsy and reach the default instead, which is quieter but no more what the adopter wrote. The test docstring and the PR body carry the same correction.

tend-agent (approved):

PR #1104, opened 2026-08-30
Extract review posting checks into a preflight script

max-sixty:
Moves the deterministic checks immediately before a review post into `review-preflight.sh`, so every review run uses the same tested path. The preflight rechecks PR state, safely retargets descendant heads, preserves large deltas for chunked reading, and stops duplicate reviews. It advances the reviewed-head pin only after every fallible check and status write succeeds, so a retry cannot lose unread changes.

### DIFF
--- a/plugins/tend-ci-runner/scripts/review-preflight.sh
+++ b/plugins/tend-ci-runner/scripts/review-preflight.sh
@@ -10,6 +10,7 @@
 #
 # env: GITHUB_REPOSITORY (optional; defaults to the checkout's remote)
 #      GITHUB_EVENT_PATH (optional; ready_for_review replaces a draft review)
+#      REVIEWED_HEAD_FILE (test-only override of /tmp/reviewed-head)
 
 set -euo pipefail
 
@@ -60,7 +61,7 @@ if [ "$CURRENT_HEAD" != "$REVIEWED" ]; then
   git fetch --no-tags --quiet origin "refs/pull/$PR/head" || true
   git fetch --no-tags --quiet origin "$BASE_SHA" || true
 
-  if git merge-base --is-ancestor "$REVIEWED" "$CURRENT_HEAD" 2>/dev/null; then
+  if git merge-base --is-ancestor "$REVIEWED" "$CURRENT_HEAD"; then
     :
   else
     STATUS=$?
--- a/plugins/tend-ci-runner/scripts/review-preflight.sh
+++ b/plugins/tend-ci-runner/scripts/review-preflight.sh
@@ -1,0 +1,158 @@
+#!/usr/bin/env bash
+# Decides whether this session may post its review, and re-targets the review
+# when the PR's head moved while it was being composed.
+#
+# Usage: review-preflight.sh <pr-number>
+#
+# A `pull_request_target` review routinely runs 5–10 minutes between reading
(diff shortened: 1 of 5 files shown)


plugins/tend-ci-runner/scripts/review-preflight.sh
```diff
@@ -0,0 +1,106 @@
...
+  if git merge-base --is-ancestor "$REVIEWED" "$CURRENT_HEAD" 2>/dev/null; then
+    :
+  else
+    STATUS=$?
+    if [ "$STATUS" -eq 1 ]; then
+      skip "cannot re-target onto $CURRENT_HEAD — $REVIEWED is no longer an ancestor; leaving it to the queued review"
+    fi
+    echo "review-preflight: git merge-base failed with status $STATUS" >&2
+    exit "$STATUS"
+  fi
```
tend-agent: Worth surfacing git's own error here. `--is-ancestor` is silent on exit 1, so `2>/dev/null` suppresses nothing except the `fatal:` line from the 128 path — and that line is the only record of *which* object was missing. That matters more now than in the prose recipe: the old snippet treated every failure as "leaving it to the queued review", while this exits non-zero and the skill tells the agent to "Fix the error and re-run the preflight", which it can't do from `failed with status 128` alone.

Same restructure drops the empty `:` branch and the `$?`-in-the-else-branch subtlety:

```suggestion
  STATUS=0
  git merge-base --is-ancestor "$REVIEWED" "$CURRENT_HEAD" || STATUS=$?
  if [ "$STATUS" -eq 1 ]; then
    skip "cannot re-target onto $CURRENT_HEAD — $REVIEWED is no longer an ancestor; leaving it to the queued review"
  elif [ "$STATUS" -ne 0 ]; then
    echo "review-preflight: git merge-base failed with status $STATUS" >&2
    exit "$STATUS"
  fi
```

plugins/tend-ci-runner/scripts/review-preflight.sh
```diff
@@ -0,0 +1,106 @@
...
+#
+# Usage: review-preflight.sh <pr-number>
+#
+# Prints `post: <reason>` when the review may proceed or `skip: <reason>` when
+# it must stop. A re-targeted post also prints `delta: <path>`; that file holds
+# the commits pushed since the review began, excluding base-branch churn. A
+# non-zero exit means no decision was made.
+#
+# env: GITHUB_REPOSITORY (optional; defaults to the checkout's remote)
+#      GITHUB_EVENT_PATH (optional; ready_for_review replaces a draft review)
```
tend-agent, line 12: `REVIEWED_HEAD_FILE` (read into `PIN_FILE` just below) is missing from this list, so the header reads as if the pin path were fixed at `/tmp/reviewed-head` — it's the knob the tests drive the script with.

```suggestion
# env: GITHUB_REPOSITORY (optional; defaults to the checkout's remote)
#      GITHUB_EVENT_PATH (optional; ready_for_review replaces a draft review)
#      REVIEWED_HEAD_FILE (optional; defaults to /tmp/reviewed-head)
```

tend-agent (approved):
Extraction looks right and the behavior tests cover paths the old string assertions couldn't reach. Two minor suggestions inline; neither blocks.

PR #1105, opened 2026-08-30
fix(install-tend): surface the OAuth TUI's own error instead of waiting it out

max-sixty:
Rotating a consumer's `CLAUDE_CODE_OAUTH_TOKEN` today meant watching `oauth_token.py` sit silent for its full 900s window with no idea whether it was waiting on a person or had already failed. It had already failed: `claude setup-token` reports a rejected code as `OAuth error: Request failed with status code 400` and then parks at "Press Enter to retry", emitting nothing further. The wrapper watched only for a token, so that state was indistinguishable from an approval still pending.

Watch for the TUI's error line, end the run on it, and report what the CLI actually said. A rejected code now fails in seconds and names its own cause instead of timing out under a guess.

The same investigation corrected a factual error in the skill. `claude setup-token` holds a localhost listener (observed at `127.0.0.1:56305`) that takes the browser redirect, so approving is normally the whole of the user's job and the run completes with nothing typed. SKILL.md presented the `code#state` paste as a co-equal path; it's the fallback for when the browser shows a code page instead. A code is also good exactly once, so one the listener already redeemed fails at the prompt — which is the likely shape of today's failure.

Two integration tests against a fake `claude`: a rejected code must fail fast (this hung for the harness's full 60s timeout before the fix), and the localhost path must produce a token with nothing typed.

<details>
<summary>Review rounds folded in</summary>

Before the PR opened:

- A comment claimed the error check's placement preserved a retry that the `break` right under it makes unreachable.
- The `failed` flag it set was redundant and slightly wrong: the loop also ends on a closed pty and on an exited child, either of which can carry the error line in the read that ends it, leaving the flag false with the error sitting in the buffer. Reading the buffer unconditionally is both shorter and more correct.

From tend's review of the first commit — two real bugs, both reproduced before fixing:

- `\s*` after the colon crossed the line break, so an error the CLI prints with no detail captured the *next* line and reported "Press Enter to retry." as the cause. Horizontal whitespace only.
- An empty capture is falsy, so that same detail-free error fell through to a message saying the TUI reported no error — about a run that ended because it did. "No error seen" and "an error with nothing after the label" are now distinct.
- The comment at the break claimed the error line "has finished arriving"; nothing establishes that, since no read happens between it and the final scan. Rewritten to say the detail may be clipped and that naming the error is what matters.
- `PASTE_PROMPT` and the flag timing the stderr hint against it are gone. The hint goes out with the authorize URL, where it's more use, and scheduling it by matching TUI wording was the coupling this change argues against elsewhere.

</details>

### DIFF
--- a/plugins/install-tend/skills/install-tend/scripts/oauth_token.py
+++ b/plugins/install-tend/skills/install-tend/scripts/oauth_token.py
@@ -65,15 +65,12 @@ TOKEN = re.compile(rb"sk-ant-oat01-[A-Za-z0-9_-]{80,}")
 # The TUI's own failure line. `claude setup-token` reports a rejected code and
 # then waits at "Press Enter to retry", which produces no further output — so
 # without this the run is silent to its deadline and reads as an approval the
-# user never gave.
-TUI_ERROR = re.compile(rb"OAuth\s*error:\s*([^\r\n]*)")
-# Drives the progress line and nothing else. Gating the paste on it would tie
-# the fallback path to TUI wording that is free to change, and buy nothing: the
-# prompt renders alongside the authorize URL, so a code can't arrive first.
-# Ink lays adjacent text nodes out with cursor moves rather than spaces, so
-# stripping CSI closes the gaps between words: both patterns above read against
-# the stripped view and have to tolerate their absence.
-PASTE_PROMPT = re.compile(rb"Paste\s*code\s*here")
+# user never gave. Ink lays adjacent text nodes out with cursor moves rather
+# than spaces, and stripping CSI closes those gaps, so the gaps here are
+# optional — but horizontal only: `\s*` after the colon would step over the
+# line break and report the following line ("Press Enter to retry.") as the
+# detail whenever the CLI's own message is empty.
+TUI_ERROR = re.compile(rb"OAuth[^\S\r\n]*error:[^\S\r\n]*([^\r\n]*)")
 
 
 def complete_match(pattern, buf):
@@ -120,9 +117,13 @@ def failure_message(code_file, typed_at, announced, tui_error=None):
     whose own output this wrapper swallows, and is the one case where the fix
     is to go look at that command rather than at the browser.
     """
-    if tui_error:
+    # `is not None`, not truthiness: the CLI prints the label with nothing after
+    # it when it has no detail to give, and an empty capture is still an error
+    # seen — falling through from here would report that no error was reported.
+    if tui_error is not None:
+        detail = f": {tui_error}" if tui_error else ", with no detail on the line"
         return (
-            f"Error: `claude setup-token` reported: {tui_error}\n"
+            f"Error: `claude setup-token` reported an OAuth error{detail}\n"
             "A code belongs to the run whose challenge issued it, is good once, "
             "and dies with that run — so a code from an earlier run, or one the "
             "CLI's own localhost callback already redeemed, fails here. Rerun "
@@ -230,9 +231,21 @@ def main():
         sys.exit("Error: claude CLI not found. Install Claude Code first.")
     os.close(slave)
 
+    # Goes out with the URL rather than waiting on the paste prompt to render:
+    # timing it against TUI wording would tie it to text free to change, and the
+    # user wants it at the moment they get the link anyway.
+    hint = (
+        "The CLI takes the redirect on its own localhost callback, so approving "
+        "is usually the whole job. If the browser shows a `code#state` string "
+        + (
+            f"instead, write that to {args.code_file}."
+            if args.code_file
+            else "instead, this run has no --code-file to read it from."
+        )
+    )
+
     buf = bytearray()
     announced = False
-    prompted = False
     typed_at = None
     enters = 0
     token = None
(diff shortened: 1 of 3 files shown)


plugins/install-tend/skills/install-tend/scripts/oauth_token.py
```diff
@@ -265,15 +314,21 @@ def stop_on_signal(signum, _frame):
...
         if final:
             token = final.group(0).decode()
 
     if not token:
-        sys.exit(failure_message(args.code_file, typed_at, announced))
+        # Read unconditionally rather than off a flag the loop set: the loop
+        # also ends on a closed pty and on an exited child, either of which can
+        # carry the error line in the same read that ends it.
+        error = TUI_ERROR.search(final_visible)
+        detail = error.group(1).decode("utf-8", "replace").strip() if error else None
```
tend-agent, line 333: An empty capture is falsy, so `failure_message` skips the `tui_error` branch it was just handed and falls through to "the TUI reported no error either" — while the loop only got here *because* `TUI_ERROR` matched. Reachable whenever the read that completes `OAuth\s*error:` ends at or just past the colon, which the `complete_match` discipline can't guard here (see the note on the break above). Branch on the match rather than on the text it captured:

```suggestion
        error = TUI_ERROR.search(final_visible)
        detail = error.group(1).decode("utf-8", "replace").strip() if error else None
        if error and not detail:
            detail = "(no detail on the line)"
```

plugins/install-tend/skills/install-tend/scripts/oauth_token.py
```diff
@@ -56,6 +62,18 @@
 # bytes, and `complete_match` would then see a following byte and call the
 # truncation final. Length is checked after extraction so it fails loudly.
 TOKEN = re.compile(rb"sk-ant-oat01-[A-Za-z0-9_-]{80,}")
+# The TUI's own failure line. `claude setup-token` reports a rejected code and
+# then waits at "Press Enter to retry", which produces no further output — so
+# without this the run is silent to its deadline and reads as an approval the
+# user never gave.
+TUI_ERROR = re.compile(rb"OAuth\s*error:\s*([^\r\n]*)")
```
tend-agent: `\s` matches `\r` and `\n`, so `\s*` after the colon walks off the end of the error line and the capture becomes the *following* line: `b"OAuth error:\r\nPress Enter to retry."` reports `claude setup-token` as having said "Press Enter to retry." Restricting it to horizontal whitespace keeps `test_tui_error_matches_the_line_as_ink_renders_it` and `test_tui_error_captures_only_its_own_line` green (verified) while confining the capture to the line the comment says it reads:

```suggestion
TUI_ERROR = re.compile(rb"OAuth\s*error:[^\S\r\n]*([^\r\n]*)")
```

plugins/install-tend/skills/install-tend/scripts/oauth_token.py
```diff
@@ -257,6 +300,12 @@ def stop_on_signal(signum, _frame):
                 token = match.decode()
                 break
 
+            # After the token check, so a token already on screen wins over an
+            # error earlier in the same buffer. The message itself is read off
+            # the final buffer below, where the line has finished arriving.
```
tend-agent: "where the line has finished arriving" doesn't hold: nothing reads the pty between this `break` and `final_visible = ANSI_CSI.sub(b"", buf)` — `stop_and_reap` only signals and waits — so the final buffer is byte-identical to the `visible` that matched here. The detail is exactly as complete as the chunk boundary left it, which is what makes the empty-capture case below reachable. Worth saying that outright, since this comment is the stated reason the truncation is safe:

```suggestion
            # After the token check, so a token already on screen wins over an
            # error earlier in the same buffer. Nothing is read after this
            # break, so the detail below comes off this same buffer and can be
            # cut short by the chunk boundary that revealed the match.
```

plugins/install-tend/skills/install-tend/scripts/oauth_token.py
```diff
@@ -233,6 +260,22 @@ def stop_on_signal(signum, _frame):
                     print(f"Approve in the browser:\n{url.decode()}", file=sys.stderr)
                     announced = True
 
+            if not prompted and PASTE_PROMPT.search(visible):
```
tend-agent: Optional. Now that gating the paste on this prompt is (rightly) dropped, `PASTE_PROMPT` and `prompted` exist only to time an stderr hint — and they time it against TUI wording that the comment two lines above argues is free to change under us, so the hint silently stops appearing on a rewording. Since the prompt renders alongside the authorize URL, printing the same sentences from the `announced` block just above gets the user the guidance at the same moment with no pattern, no flag, and one fewer test tied to the TUI's text.

tend-agent (commented):
The watch-and-exit design is right and the two integration tests pin the behaviours that matter. Two things in the new error path make it report the wrong cause, though — which is the failure mode the PR exists to remove.

The sharper one: the loop breaks *because* `TUI_ERROR` matched, and then the message says the TUI reported no error. `TUI_ERROR` deliberately doesn't go through `complete_match` (it can't — on the retry path the error line is the last thing the TUI writes), so the capture is whatever the chunk boundary left, and an empty capture is falsy, so `failure_message` skips its own `tui_error` branch. Reproduced against the merged tree:

```python
>>> detail = TUI_ERROR.search(b"OAuth error: ").group(1).decode().strip()
>>> failure_message("/tmp/tend-oauth-code", 1.0, True, detail)
'Error: the code from /tmp/tend-oauth-code was typed into the prompt and no token came back, and the TUI reported no error either. Rerun and approve the fresh URL.'
```

That path also takes the docstring's "So does the TUI's own error line" with it — the line never reaches stderr at all. The second is `\s*` after the colon: `\s` includes `\r\n`, so an error line with no detail captures the *next* line, and `b"OAuth error:\r\nPress Enter to retry."` is reported as `claude setup-token` having said "Press Enter to retry." Both inline suggestions together keep all three of the new `TUI_ERROR` tests green.

tend-agent (approved):

PR #1106, opened 2026-08-30
fix(generator): preserve pre-check setup guards

max-sixty:
## Summary

- keep Tend's workflow pre-check on every generated setup step
- combine an adopter's `if:` expression with that guard instead of replacing it
- normalize whole-expression `${{ ... }}` syntax, reject ambiguous non-string
  values, and fail clearly on mixed interpolation fragments

This extracts the useful setup-condition correction from #1088 without adding a second review gate.

## Verification

- `uv run pytest generator/tests` (589 passed)
- `wt test` after merging current `main` (834 Python tests, 32 worker tests,
  TypeScript typecheck)
- independent design review


generator/src/tend/config.py
```diff
@@ -327,6 +327,20 @@ def load(cls, path: Path | None = None) -> Config:
...
+                        f"setup[{i}]: `if` must be a non-empty string"
+                    )
+                condition = condition.strip()
+                if condition.startswith("${{") and condition.endswith("}}"):
+                    condition = condition[3:-2].strip()
+                    if not condition:
+                        raise click.ClickException(
+                            f"setup[{i}]: `if` must contain an expression"
+                        )
+                entry = {**entry, "if": condition}
```
tend-agent, line 350: The wrapper check passes for any condition that opens with `${{` and closes with `}}`, including two interpolations joined by an operator. Rejecting a condition that still contains interpolation after the unwrap covers that case and the unwrapped mixed form (`a || ${{ b }}`) in one test:

```suggestion
                if "${{" in condition or "}}" in condition:
                    raise click.ClickException(
                        f"setup[{i}]: `if` must be a single expression; "
                        "write it without `${{ }}` interpolation"
                    )
                entry = {**entry, "if": condition}
```

tend-agent (commented):
One finding, on the new `${{ }}` normalization in `Config.load`: the unwrap tests only the first three and last two characters, so a condition that *starts* with an interpolation and *ends* with one — but isn't a single wrapper — gets sliced through the middle.

Reproduced on this branch. With `setup: - run: ./x.sh` and `if: "${{ runner.os == 'Linux' }} && ${{ github.event_name == 'push' }}"`, `tend-notifications.yaml` renders:

```yaml
    if: (steps.check.outputs.count != '0' || github.event_name == 'workflow_dispatch') && (runner.os == 'Linux' }} && ${{ github.event_name == 'push')
```

`init` exits 0 and says nothing; GitHub rejects the expression when the workflow runs. The unwrapped mixed form has the same shape from the other side — `if: "a || ${{ b }}"` skips the strip entirely and carries a live `${{` into the parenthesized conjunct. Both were passed through untouched before this PR, so the mangling is new here.

Since these forms can't be composed with the guard, rejecting them at load is the cut-over CLAUDE.md asks for ("old formats should fail with a clear error, not silently parse"), and it matches what `checks.py` already does with `"${{" in environment` for a dynamically-named environment. Inline suggestion below; with it applied `uv run pytest generator/tests` is 585 passed and the mixed form fails `init` with a message naming the fix.

generator/src/tend/config.py
```diff
@@ -327,6 +327,25 @@ def load(cls, path: Path | None = None) -> Config:
             for k in DICT_STEP_FIELDS:
                 if k in entry and not isinstance(entry[k], dict):
                     raise click.ClickException(f"setup[{i}]: `{k}` must be a mapping")
+            if "if" in entry:
+                condition = entry["if"]
+                if not isinstance(condition, str) or not condition.strip():
+                    raise click.ClickException(
+                        f"setup[{i}]: `if` must be a non-empty string"
+                    )
```
tend-agent, line 337: `if: false` is a normal way to park a setup step, but YAML hands it over as a bool and it lands here — while `if: ${{ false }}` and `if: "false"` both pass and mean the same thing. The message names neither the asymmetry nor the fix, so an adopter who hits it has to guess that quoting is what's wanted. Widening it keeps the existing test's `match=` substring intact; I ran it, 588 still pass.

```suggestion
                    raise click.ClickException(
                        f"setup[{i}]: `if` must be a non-empty string; quote "
                        'it (`if: "false"`) so YAML does not read it as a '
                        "boolean, number, or list"
                    )
```

docs/tend.example.yaml
```diff
@@ -111,8 +111,8 @@ bot_name: my-project-bot
 #
 # `tend-notifications` decides up front whether the agent needs to boot and
 # injects an `if:` guard on every setup step so an empty-inbox run does no setup
-# either. A user-supplied `if:` is passed through unchanged (with a
-# generation-time warning) — your condition wins.
+# either. A user-supplied `if:` narrows that guard; it cannot run setup after
+# the pre-check declined the job.
```
tend-agent: This paragraph is where an adopter reads what an `if:` on a setup step may contain, and it's now silent on the new hard failure — a mixed `${{ }}` condition stops `init` outright. The `sync-install-tend-references` pre-commit hook re-syncs the `plugins/install-tend/…/references/tend.example.yaml` mirror from this file, so this is the only copy to edit.

```suggestion
# either. A user-supplied `if:` narrows that guard; it cannot run setup after
# the pre-check declined the job. Write it as a plain expression or one whole
# `${{ ... }}`; mixing the two is rejected at `init`.
```

tend-agent (approved):
The mixed-interpolation case from the last review is fixed, and I re-verified it on this head: `if: "${{ a }} && ${{ b }}"` and `if: "a || ${{ b }}"` both fail `init` now, `if: "${{ runner.os == 'Linux' }}"` normalizes to the bare expression, and `tend-notifications` renders `if: (steps.check.outputs.count != '0' || github.event_name == 'workflow_dispatch') && (runner.os == 'Linux')` with the `check` step still ahead of every guarded step. `uv run pytest generator/tests` is 588 passed.

Two suggestions inline, plus two notes:

- The `"${{ }}"` empty-wrapper branch (`if` must contain an expression) is the only one of the three new rejection paths with no test; the other two are parametrized.
- `_setup_yaml`'s replacement here is byte-for-byte the change #1088 also carries, so whichever lands second conflicts in `workflows.py` and `test_generate.py`. Worth dropping it from #1088 once this merges.

PR #1107, opened 2026-08-30
fix(review): bind preflight to the outward action

max-sixty:
## Summary

Follow up #1104 by making its preflight the actual boundary around every
review write:

- execute ordinary review POSTs, the body-only retry, and orphan repair only
  after a final stable head/state check
- authorize orphan edits only when the requested ID is the exact current
  orphan review
- treat a re-target as a read-and-reassess pass that cannot post
- ignore unsubmitted `PENDING` reviews in canonical review state
- always recompute suggestion line numbers after a safe re-target

This also supersedes #1083's separate suggestion-recomposition change.

## Verification

- `wt test` after merging current `main` (833 Python tests, 32 worker tests,
  TypeScript typecheck)
- two independent review cycles, including adversarial race and edit-mode
  probes


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -203,8 +203,9 @@ REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
 # missing file substitutes the empty string and the POST still runs, which is
 # the unpinned review this pins against.
 REVIEWED=$(cat /tmp/reviewed-head) || exit 0
-gh api "repos/$REPO/pulls/<number>/reviews" --method POST \
-  -f event=APPROVE -f commit_id="$REVIEWED" -f body="" && echo "✓ approved"
+${CLAUDE_PLUGIN_ROOT}/scripts/review-preflight.sh <number> -- \
+  gh api "repos/$REPO/pulls/<number>/reviews" --method POST \
+    -f event=APPROVE -f commit_id="$REVIEWED" -f body="" && echo "✓ approved"
```
tend-agent: `&& echo "✓ approved"` binds to the preflight now, not the POST, and the preflight exits 0 on every `skip` — so this prints `✓ approved` when nothing was posted. Dropping it loses nothing: the preflight already prints `post:`/`skip:`, and `gh api` prints the created review on success.

```suggestion
${CLAUDE_PLUGIN_ROOT}/scripts/review-preflight.sh <number> -- \
  gh api "repos/$REPO/pulls/<number>/reviews" --method POST \
    -f event=APPROVE -f commit_id="$REVIEWED" -f body=""
```

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -234,16 +235,18 @@ gh api "repos/$REPO/issues/<number>/reactions" -f content="+1"
...
+Before composing the final payload, run the preflight without a command. It checks the PR is open, re-targets onto a newer descendant head, and stops duplicate reviews:
 
 ```bash
 ${CLAUDE_PLUGIN_ROOT}/scripts/review-preflight.sh <number>
 ```
 
-On `skip`, post nothing and finish. On `post`, anchor the review at `/tmp/reviewed-head`; the preflight updates the file when HEAD moves. A re-targeted post also prints `delta: <path>`. Read that entire file in chunks as needed before posting; it is the push that landed mid-review.
+On `skip`, post nothing and finish. A re-targeted result also prints `delta: <path>` and updates `/tmp/reviewed-head`. Read that entire file in chunks, update the review, then run the preflight again. Do not post from the re-targeting pass.
 
 A non-zero exit means nothing was decided. Fix the error and re-run the preflight.
```
tend-agent: In command mode a failed `gh api` propagates its status out of the preflight, so non-zero no longer implies nothing was decided — and re-running the preflight is the wrong recovery for a 422, which persists the body and then reports it as an already-posted review.

```suggestion
A non-zero exit with no `post:` line means nothing was decided — fix the error and re-run the preflight. With a command, a non-zero exit *after* `post:` is the command's own status: the review was attempted, so recover from that failure (see **Recovering from inline comment 422 errors**) rather than re-running.
```

tend-agent (commented):
Two findings, both on the skill side of the boundary this PR draws.

The approve recipe now reports success on every `skip`. `review-preflight.sh` exits 0 when it declines to run the command — `test_a_wrapped_outward_action_does_not_retarget` pins exactly that — so `&& echo "✓ approved"`, which used to be gated on the POST itself, is now gated on the preflight instead and fires on the closed-PR, moved-head, already-reviewed and stale-state paths alike. The session then reads `skip: HEAD moved to … before the outward action` followed by `✓ approved`, believes it approved, and carries that into step 6's dismissal-on-failure gate and its closing summary. It's the one place the diff turns a real skip into a claimed outward action, which is the class of thing the rest of the PR exists to prevent.

Separately, the exit-code contract at the top of **Posting mechanics** no longer holds in command mode: a failed `gh api` POST now propagates out of the preflight, so a non-zero exit can mean the review *was* attempted and failed. That's the 422 case, and the sentence directly above the new paragraph sends the reader to re-run the preflight instead — which on case (a) reports the persisted body as `already carries a COMMENTED review N` and finishes, stranding the failed inline comments outside the body.

The rest reads clean. Wrapping all four write paths (POST, body-only retry, orphan PUT, approve) is complete — there's no unwrapped review write left in the skill. The `STATE_HEAD` check isn't redundant with the final re-read as it first looks, since `at_head`/`orphan_id` are keyed on the head `bot-review-state.sh` saw, so an unstable head there would evaluate the duplicate-guard against the wrong commit. And requiring `at_head.id == orphan_id == $edit` for `--edit-review` means a moved head can no longer authorize an orphan edit at all, since `at_head` goes null — conservative, but consistent with treating the edit as an outward action.

- `uv run pytest` — 823 passed
- Reproduced the `✓ approved` behaviour directly with a stub that prints `skip:` and exits 0

</details>

tend-agent (approved):

PR #1108, opened 2026-08-31
Repair conflicted bot PRs on the frequent poll

max-sixty:
Extend the 15-minute notifications workflow's precheck to count possible conflicts on PRs from the configured bot. The agent boots only when notifications or an eligible PR need work; the shared resolver then performs the authoritative local merge test.

When a conflict needs manual help, Tend posts one human-readable comment with a marker scoped to that head. Frequent ticks skip the same head, a new head is eligible immediately, and nightly still retries. Before changing a branch, the resolver re-reads the PR's head and base and pushes with an exact lease.

- `wt test` — 849 Python tests and 32 worker tests passed; worker typecheck passed
- `pre-commit run --all-files` — passed

</details>


generator/src/tend/templates/notifications-check.sh
```diff
@@ -34,8 +34,36 @@ fi
 
 echo "count=$COUNT" >> "$GITHUB_OUTPUT"
 
-if [ "$COUNT" = "0" ]; then
-  echo "No notification work before $CUTOFF — skipping"
+# GitHub computes mergeability lazily after the base moves. UNKNOWN therefore
+# means "worth a synchronous local test", not "clean". This is only a cheap
+# boot gate; the agent test-merges every candidate before changing a branch.
+# Missing a marker beyond the CLI's 100-comment window only wastes one boot.
```
tend-agent: `gh pr list --json comments` issues `comments(first: 100)` — I confirmed the emitted GraphQL with `GH_DEBUG=api`. So this reads a PR's *oldest* hundred comments, and the deferral marker is by construction the newest one. Once a bot PR's conversation passes 100 comments the marker is permanently outside the window, and the consequence isn't "one boot": it's every boot for that PR, forever — ~96 sessions a day, each re-attempting the resolution and re-editing the same comment. The long-lived rolling PRs are both the ones most likely to accumulate that many comments and the ones most likely to keep conflicting.

`gh` doesn't expose `last:` for this field, so the fix probably isn't in the query. The PR body would work: it's one `--json body` field with no pagination window, it's the bot's own PR, and moving the marker there also collapses most of the reconciliation protocol at the bottom of `resolve-conflicts` into "set or clear a line in our own body". The human-readable comment can stay human-readable and marker-free.

plugins/tend-ci-runner/skills/resolve-conflicts/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+   --force-with-lease="refs/heads/<headRefName>:<headRefOid>" origin
+   "HEAD:refs/heads/<headRefName>"`. The exact lease is the final head guard.
+4. Fetch the live base again and test the pushed head with `git merge-tree`.
+   If it conflicts, merge the new base and repeat. Once clean, remove the bot's
+   conflict-deferral comment and monitor CI per **CI Monitoring** in
+   `/tend-ci-runner:running-in-ci`.
+
+If resolution is too complex, abort the merge and re-read the PR. When it is
+still open at the original head, create or update one bot-authored comment that
+explains manual resolution is needed and ends with:
```
tend-agent: The gate matches the marker as the comment's *whole final line* — `(.body | sub("\\s+$";"") | split("\n") | last) == $marker` — and the test for it deliberately rejects a marker quoted mid-line. "Ends with" permits `...manual resolution is needed. <!-- tend-conflict-deferred head=abc -->`, which satisfies the instruction and silently fails the match, so the PR never gets suppressed.

```suggestion
explains manual resolution is needed. Its final line must be exactly this
marker and nothing else — the boot gate compares the comment's whole last line:
```

generator/tests/test_cross_file_contracts.py
```diff
@@ -44,6 +44,42 @@ def test_notification_skill_pins_the_fragile_dedup_queries() -> None:
...
+    assert '--force-with-lease="refs/heads/<headRefName>:<headRefOid>"' in resolver
+    assert "Fetch the live base again" in resolver
+    assert "remove any prior conflict-deferral comment" in " ".join(
+        resolver.lower().split()
+    )
+    assert "When called from `notifications`" in resolver
+    assert "Nightly ignores the marker" in " ".join(resolver.split())
+    assert "paginated issue-comment API" in resolver
+    assert "only when it already names this head" in resolver
+    assert "<!-- tend-conflict-deferred head=<head SHA> -->" in resolver
```
tend-agent, line 71: This file's docstring is "A rule split across two files drifts silently", but the split rule here — the marker literal, shared between this skill and the jq in `notifications-check.sh` — is the one thing the test doesn't pin. Everything else it asserts (`baseRefOid`, `Fetch the live base again`, `Nightly ignores the marker`) is prose internal to one file, so it raises the cost of editing the skill without protecting a contract; `origin/main not in resolver` will also misfire on any unrelated future mention.

Both added assertions pass against the current script:

```suggestion
    check = _read("generator", "src", "tend", "templates", "notifications-check.sh")
    assert "<!-- tend-conflict-deferred head=<head SHA> -->" in resolver
    # The gate's jq builds this marker and matches it as the comment's whole
    # final line; the resolver skill is what writes it.
    assert r"<!-- tend-conflict-deferred head=\($pr.headRefOid) -->" in check
    assert r'split("\n") | last' in check
```

tend-agent (commented):
The gate's shape is right: one query, an authoritative local test behind it, and a head-scoped marker so a deferred conflict doesn't re-boot every quarter hour. Three notes on the marker path (inline), plus a couple that don't attach to a line.

**The marker is only written on the deliberate-defer path.** A session that times out, hits the token ceiling, or gets replaced in the queue before it reaches "resolution is too complex" leaves no marker — so the next tick boots on the same PR, and so does every tick after it. Under nightly that cost one session a day; on the frequent poll it's ~96. I don't see a cheap fix (writing the marker up front would post "needs manual resolution" on PRs the bot is about to fix cleanly), so this may just be worth knowing about rather than closing.

**`--force-with-lease` may not be buying anything.** Step 3 re-reads the live head and verifies the retained head is an ancestor of `HEAD`, which makes the push a fast-forward; a plain `git push` is already refused on any surprise. The lease is the only thing in the skill that enables a force at all, and the ancestry check is what actually guarantees the outcome.

**The deferral reconciliation is prose where a script would fit.** "If the head changed, delete only the stale comment. Otherwise keep the oldest current-head comment and remove every other deferral comment" is four branches of edge-case handling, executed by an LLM, mutating comments on PRs, with nothing testing it — `running-in-ci`'s **Scripts over prose recipes** points at that shape. It also mostly disappears if the marker moves off comments (see the inline on the precheck).

Unrelated: #1106 edits `workflows.py`, `docs/tend.example.yaml`, and the install-tend copy around the same setup-guard `if:`. Different concern, but the two will conflict.

<details><summary>Verified locally</summary>

- `uv run pytest generator/tests/test_shared_steps.py generator/tests/test_cross_file_contracts.py generator/tests/test_integration.py` — 52 passed
- `GH_DEBUG=api gh pr list --json comments` on this repo — the emitted GraphQL is `comments(first: 100) {nodes {id ...`
- The two test assertions suggested inline pass against the current `notifications-check.sh`

</details>

generator/src/tend/templates/notifications-check.sh
max-sixty: Switched the gate to one explicit GraphQL search using `comments(last: 100)`, so it reads the newest comments while preserving the comment-based state. If a marker becomes older than the newest 100, the only effect is a wasted agent boot: the resolver paginates comments again before acting. This avoids putting machine state in the PR body, where clearing the marker could overwrite concurrent body edits.

plugins/tend-ci-runner/skills/resolve-conflicts/SKILL.md
max-sixty: Updated the skill to require the marker as the exact final line, matching the gate's jq contract.

generator/tests/test_cross_file_contracts.py
max-sixty: Reworked this test to pin the shared marker literal, the `comments(last: 100)` window, and the whole-final-line jq match. I removed the assertions against incidental resolver prose.

max-sixty:
Addressed the inline findings in 69715c1. I also simplified reconciliation: create only when no same-head deferral exists, then delete only the newly created comment if the head changed or an older same-head comment won a race.

I kept the exact force lease. A normal fast-forward push can still recreate a concurrently deleted branch or restore a concurrent reset to an ancestor; leasing against the retained head SHA prevents both. A session that dies before writing a deferral may retry next tick, but there is no comparably cheap safe pre-work marker, and that failure only wastes compute.

generator/src/tend/templates/notifications-check.sh
```diff
@@ -34,8 +34,45 @@ fi
...
+        } }
+      }
+    }' -f q="repo:$GITHUB_REPOSITORY author:$BOT_LOGIN is:pr is:open" \
+    --jq '.data.search.nodes' 2>/dev/null) \
+  && CONFLICT_COUNT=$(jq -er --arg bot "$BOT_LOGIN" '
+    [.[]
+      | select(.mergeable != "MERGEABLE")
+      | . as $pr
+      | "<!-- tend-conflict-deferred head=\($pr.headRefOid) -->" as $marker
+      | select(any($pr.comments[]?;
```
tend-agent: The GraphQL field returns `comments { nodes { ... } }`, so `$pr.comments[]?` iterates the *object* and hands the predicate the `nodes` array — jq dies with `Cannot index array with string "author"`, `jq -er` exits non-zero, and the `else` pins `conflict_count` at `0`. Net effect: the gate never boots for a conflict and warns every 15 minutes.

```suggestion
      | select(any($pr.comments.nodes[]?;
```

generator/tests/test_shared_steps.py
```diff
@@ -332,11 +332,19 @@ def _output(env: dict[str, str], key: str) -> str:
...
     echo "put" >> "$SUBSCRIPTION_WRITES"
     emit '{"subscribed":true,"ignored":false}'
     ;;
+  api:user)
+    [ -z "${FAIL_PRS:-}" ] || exit 1
+    emit '{"login":"test-bot"}'
+    ;;
+  api:graphql)
+    [ -z "${FAIL_PRS:-}" ] || exit 1
+    emit "$(jq -c '{data: {search: {nodes: .}}}' "$PULLS_JSON")"
```
tend-agent: The fake wraps the outer envelope but leaves each fixture's `comments` a bare array, so the suite drives the script with a shape the API never returns — which is why the bug above stays green. Reshaping here keeps all four fixtures readable as they are:

```suggestion
    emit "$(jq -c '{data: {search: {nodes: [.[] | .comments = {nodes: .comments}]}}}' "$PULLS_JSON")"
```

(Putting the `{nodes: ...}` wrapper in the fixtures themselves works equally well — this way just avoids touching four literals.)

tend-agent (commented):
The move to `search` with `comments(last: 100)` reads the newest comments, but the jq that consumes it was left on the old response shape, and the gate silently fails closed as a result.

`gh pr list --json comments` returned `comments` as an array; the GraphQL field returns `comments { nodes { ... } }`. `any($pr.comments[]?; .author.login == $bot ...)` iterates the object's *values*, so the first thing it hands the predicate is the `nodes` array, and jq aborts with `Cannot index array with string "author"`. `jq -er` exits non-zero, the `&&` chain falls to the `else`, and `conflict_count` is pinned at `0` — so `tend-notifications` never boots for a conflict and emits `::warning::bot PR conflict scan failed` every 15 minutes instead. One inline fixes it.

The reason CI is green is the second inline: the fake `gh` wraps the fixtures in `{data: {search: {nodes: .}}}` but leaves each PR's `comments` as a bare array, so the suite exercises a shape the API never produces. With the fixtures reshaped and the script left as-is, both conflict tests fail with `conflict_count` `0` — which is exactly what production would do.

<details><summary>Verified locally</summary>

Ran `generator/src/tend/templates/notifications-check.sh` at `699a858` under a fake `gh` returning the real envelope (`comments: {nodes: [...]}`), with one marked `CONFLICTING` PR and one unmarked `UNKNOWN` one:

```
jq: error (at <stdin>:1): Cannot index array with string "author"
::warning::bot PR conflict scan failed; retrying next cycle
conflict_count=0
```

With only `$pr.comments[]?` → `$pr.comments.nodes[]?`, the same fixture gives `conflict_count=1` — the marked head suppressed, the unmarked one counted.

Test-side, in a worktree at `699a858`:

- fixture reshaped, script unchanged → `test_notifications_check_boots_for_unknown_or_conflicting_bot_prs` and `test_notifications_check_suppresses_only_the_marked_bot_head` both fail (`assert '0' == '2'`, `assert '0' == '3'`)
- both inlines applied → `uv run pytest generator/tests/test_shared_steps.py generator/tests/test_cross_file_contracts.py generator/tests/test_integration.py` — 52 passed

</details>

generator/src/tend/templates/notifications-check.sh
max-sixty: Fixed in cdec026: the jq now iterates `$pr.comments.nodes[]?`, matching the GraphQL response shape. The focused 61-test contract suite and full 849-test suite pass.

generator/tests/test_shared_steps.py
max-sixty: Fixed in cdec026: the fake GraphQL response now wraps each fixture's comments in `{nodes: ...}`, so the tests exercise the production API shape and catch a mismatched traversal.

tend-agent (approved):

PR #1109, opened 2026-08-31
Provision uv for Tend agent sessions

max-sixty:
Tend skills now invoke `uv tool run` directly. Both harnesses install the same pinned fallback after adopter PATH entries, so a project-provided `uv` stays selected, while the Claude credential proxy continues to use a separate copy by absolute path.

This removes the plugin bootstrap wrapper and consolidates proxy and agent installation on one exact-version installer. The hosted sandbox test covers the runner-home UID boundary and adopter PATH priority.

Validation: `wt test` (850 Python tests, 32 worker tests, worker typecheck) and `pre-commit run --all-files`.


proxy/setup-sandbox.sh
```diff
@@ -240,6 +240,9 @@ for _d in "${_runner_path[@]}"; do
     for _command_path in "${_resolved_path}"/*; do
       [ -f "${_command_path}" ] && [ -x "${_command_path}" ] || continue
       _command_name=${_command_path##*/}
+      # Tend supplies uv after the adopter PATH when a runner-home copy cannot
+      # cross the sandbox boundary.
+      [ "${_command_name}" = uv ] && continue
```
tend-agent: `uvx` is dropped from the same directory and still gets a shim, so the adopter ends up with a working `uv` and a `uvx` that exits 127.

```suggestion
      # Tend supplies uv after the adopter PATH when a runner-home copy cannot
      # cross the sandbox boundary. `uvx` lives in that same fallback dir, so
      # shimming it would strand half the pair behind a 127.
      case "${_command_name}" in uv | uvx) continue ;; esac
```

proxy/test-setup-sandbox.sh
```diff
@@ -124,6 +135,7 @@ verify() {
   test "$(sudo -u "$SANDBOX" env "${agent_env[@]}" tend-seeded)" = sandbox-seed
   test "$(sudo -u "$SANDBOX" env "${agent_env[@]}" tend-workspace-explicit)" = workspace-explicit
   test "$(sudo -u "$SANDBOX" env "${agent_env[@]}" tend-workspace-path)" = workspace-path
+  test "$(sudo -u "$SANDBOX" env "${agent_env[@]}" uv --version)" = adopter-uv
```
tend-agent, line 140: This proves the adopter's copy wins; the fallback behind it is never touched.

```suggestion
  test "$(sudo -u "$SANDBOX" env "${agent_env[@]}" uv --version)" = adopter-uv
  # ...and the fallback that copy is hiding is installed where the composed
  # PATH says it is, which the adopter stub would otherwise mask.
  sudo -u "$SANDBOX" test -x "$TEND_AGENT_UV_DIR/uv"
  case ":$(sed -n 's/^PATH=//p' "$AGENT_ENV_FILE"):" in
    *":$TEND_AGENT_UV_DIR:"*) ;;
    *) echo "::error::pinned uv fallback missing from the sandbox PATH"; exit 1 ;;
  esac
```

tend-agent (commented):
Two things on the sandbox side.

**The blocked-command skip is keyed on `uv` alone, but `uvx` ships in the same directory.** For an adopter whose `setup:` installs uv into the runner home (the astral installer's default is `~/.local/bin`), that directory gets dropped and both binaries get 127 shims today. After this change `uv` skips the shim and falls through to the pinned fallback, while `uvx` still resolves to the `came from the runner home and is unavailable` shim — and `TEND_BLOCKED_PATH` is spliced in ahead of the shared paths, so it ranks above `TEND_AGENT_UV_DIR` at the tail. The pair splits, and the shim's message points at `sandbox_setup:` for a tool tend now provides. The bundled skills moved to `uv tool run` in this PR, but adopter overlays and the agent's own reflex still reach for `uvx`.

**Nothing asserts the fallback is reachable.** `plant` installs a shared `uv`, so `verify`'s new assertion proves the adopter's copy wins — which is the half that was already working. The install step's own `uv --version` / `uvx --version` prove the binaries exist at whatever `UV_INSTALL_DIR` it was handed; nothing ties that back to the `TEND_AGENT_UV_DIR` that `setup-sandbox.sh` appends to the composed PATH. A divergence between those two, or the step going missing, ships green — and the case it breaks (a repo with no uv of its own) is the one this PR exists for.

Everything else reads clean: the fallback step lands before `Run adopter sandbox setup`, so the `sandbox_setup:` claim in `docs/security-model.md` holds; `UV_UNMANAGED_INSTALL` still forces `NO_MODIFY_PATH=1` even with `UV_INSTALL_DIR` also in the environment taking precedence for the directory, so dropping the explicit `UV_NO_MODIFY_PATH=1` is safe; and the sandbox user can read `$GITHUB_ACTION_PATH` scripts after the workspace handoff grants traversal, which the CI fixture doesn't cover since it substitutes the workspace path.

Happy to push both fixes if you'd rather not hand-apply them.

proxy/setup-sandbox.sh
max-sixty: Fixed in `3f52b21`: `uv` and `uvx` now bypass the runner-home failure shim together and reach the appended Tend fallback.

proxy/test-setup-sandbox.sh
max-sixty: Fixed in `3f52b21`: the hosted sandbox test now asserts that the pinned fallback is executable and that its directory is the final entry in the composed agent PATH.

tend-agent (approved):

proxy/setup-sandbox.sh
tend-agent: Confirmed. The `case` covers both names, so a runner-home `uv`/`uvx` pair now falls through together to the `TEND_AGENT_UV_DIR` entry at the tail rather than splitting, and the shim's `sandbox_setup:` message no longer fires for a tool tend provides.

proxy/test-setup-sandbox.sh
tend-agent: Confirmed. `sudo -u "$SANDBOX" test -x "$TEND_AGENT_UV_DIR/uv"` and the PATH-tail `grep` tie the action step's `UV_INSTALL_DIR` to the directory `setup-sandbox.sh` actually appends, so the two diverging — or the step going missing — now goes red instead of shipping green. `test-sandbox` passed on `3f52b21`.

PR #1110, opened 2026-08-31
Add experimental Gist-backed Claude memory

max-sixty:
Adds an opt-in, experimental `memory_gist: true` setting for the Claude harness so Claude Code's model-authored auto memory can persist across Tend runs without changing default behavior.

The generator threads the setting only into enabled Claude workflows and requires the bot-owned secret Gist ID through the gated `TEND_MEMORY_GIST_ID` environment secret. The composite action restores memory before Claude starts, explicitly enables Claude auto memory, saves changes only after the sandbox is reaped, and treats synchronization failures as warnings that cannot replace the task verdict.

The new helper binds each secret Gist to one public repository, verifies bot ownership and the expected description, signs the per-run baseline, rejects unsafe filesystem shapes, and skips the entire save when it observes a remote conflict before PATCH. The config reference and security model document that secret Gists are unlisted rather than private and that recalled notes remain untrusted context.

Verification: `wt test` (880 Python tests, 32 worker tests, worker typecheck) and `pre-commit run --all-files`.


shared/steps/gist_memory.py
```diff
@@ -0,0 +1,322 @@
...
+            continue
+        current = remote.get(name)
+        if current not in (before, after):
+            conflicts.append(name)
+            continue
+        if current == after:
+            continue
+        writes[name] = None if after is None else {"content": after}
+
+    if writes:
```
tend-agent: Conflicts are dropped per-file but the writes are still applied, so an overlapping run lands a partial, self-inconsistent state — and MEMORY.md is the file most likely to be the one dropped, since it's the index every memory write touches.

Concretely: runs A and B both restore baseline `B0`. A saves first. B then computes `MEMORY.md` as a conflict (remote is A's, matching neither `before` nor `after`) and skips it, but `obsolete.md` — which B deleted and A never touched — has `current == before`, so the delete goes through. The Gist ends up with A's index still carrying `@obsolete.md` and no `obsolete.md` behind it. The mirror case leaves a new topic file with nothing in the index pointing at it, and nothing ever prunes either.

The docstring's "preserves remote edits that have already landed" holds per file but not across the set, and this isn't a rare path for tend — review, mention, triage and notifications routinely overlap. Cheapest fix consistent with the stated intent is all-or-nothing: if `conflicts` is non-empty, skip the PATCH entirely and warn, so a run's notes either land together or wait for the next run. Happy to push that if you'd rather not respin.

shared/steps/gist_memory.py
```diff
@@ -0,0 +1,322 @@
...
+key. ``restore`` downloads the memory directory and records a baseline.
+``save`` uploads only locally changed files whose remote version still matches
+that baseline when read. This preserves remote edits that have already landed,
+but Gist PATCH has no atomic precondition: truly simultaneous writes can still
+race, so synchronization is best-effort.
+
+This is opt-in and limited to public repositories. A secret Gist is unlisted,
+not private, and Claude decides what is worth remembering. The CLI reads the
+Gist locator and signed-baseline key from its environment so they never appear
+in the process command line.
```
tend-agent: The Gist ID does reach a command line — `_api` passes `/gists/{gist_id}` as an argv to `gh`, so it's visible in `ps` for the life of that subprocess. Actions masks it in the step log, and the step comment already concedes the agent can list the bot's Gists anyway, so this is the claim being wrong rather than a hole; the baseline-key half is accurate.

```suggestion
not private, and Claude decides what is worth remembering. The CLI reads its
configuration from the environment, so the signed-baseline key never reaches a
command line — the Gist ID still does, in the ``gh api /gists/<id>`` argv.
```

tend-agent (commented):
Ran the change plus a structured second pass; `uv run pytest` is green locally (878 passed). The mechanism checks out against the pinned binary — `autoMemoryEnabled` / `autoMemoryDirectory` are both real settings keys in claude 2.1.241, `--settings` lands in `flagSettings` which sits above `localSettings` in the resolution chain (so `settings.local.json`'s `bypassPermissions` survives), and `CLAUDE_CODE_DISABLE_AUTO_MEMORY` is read from the process env ahead of settings, which is what makes the filter in `launch_argv` load-bearing. `sandbox_reaped` in a `finally` around `supervise` is sound: that function's KILL is already its own unconditional `finally`.

Four things below. The first is the one I'd actually want changed before this runs anywhere; the rest are small.

One design note rather than a defect: `memory_gist` is global across every enabled Claude workflow, while `harness` and `model` are both per-workflow. For an experiment whose own security note says memory can carry an earlier injection forward, an adopter may well want it on `nightly` and not on `review` — worth deciding now whether that lever exists, since the config surface already has the shape for it.

generator/src/tend/checks.py
```diff
@@ -1701,9 +1727,12 @@ def run_all_checks(cfg: Config, repo: str | None = None) -> list[CheckResult]:
     results.append(check_environment_deployments(repo))
     results.append(check_credential_environments(repo, cfg, admitted))
     results.append(check_secrets(repo, required_secrets))
-    if cfg.harness == "claude":
+    if cfg.memory_gist:
+        results.append(check_memory_gist_repository(repo))
+    enabled_harnesses = cfg.enabled_harnesses()
+    if "claude" in enabled_harnesses:
```
tend-agent, line 1733: `enabled_harnesses` is empty when every standard workflow is disabled, so neither auth check runs and `tend check` reports green without ever looking for `CLAUDE_CODE_OAUTH_TOKEN` or `OPENAI_API_KEY`. Before this change the `if/else` guaranteed exactly one of them always ran.

Narrow — it needs an explicit disable-all — but `test_memory_gist_ignores_a_disabled_top_level_harness` shows that config is reachable and already handled deliberately for `memory_gist`. A `not enabled_harnesses` fallback to `cfg.harness` would keep the old guarantee.

shared/steps/gist_memory.py
```diff
@@ -0,0 +1,322 @@
...
+SETTINGS_FILE = ".tend-settings.json"
+MAX_FILE_BYTES = 1_000_000
+
+_GIST_ID = re.compile(r"^[A-Za-z0-9]+$")
+_REPOSITORY = re.compile(r"^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$")
+_GITHUB_LOGIN = re.compile(r"^[A-Za-z0-9](?:[A-Za-z0-9-]*[A-Za-z0-9])?$")
+_BASELINE_KEY = re.compile(r"^[a-f0-9]{64}$")
+
+
+class MemoryError(RuntimeError):
```
tend-agent: This shadows the builtin `MemoryError`, which makes `except (MemoryError, OSError, UnicodeError)` in `main` read as if it catches a genuine out-of-memory condition — it doesn't, since the builtin derives from `Exception` and this one from `RuntimeError`. Ruff's defaults don't include flake8-builtins so nothing flags it. `GistMemoryError` costs a rename across this file and the test.

shared/steps/gist_memory.py
max-sixty: Fixed in `15020ce`: any observed conflict now skips the entire PATCH, with regression coverage for both a conflicting change set and a consistent multi-file save.

generator/src/tend/checks.py
max-sixty: Fixed in `15020ce`: an empty effective harness set now falls back to the configured harness, with a regression test for the all-workflows-disabled case.

shared/steps/gist_memory.py
max-sixty: Fixed in `15020ce`: the custom exception is now named `GistMemoryError` throughout.

shared/steps/gist_memory.py
max-sixty: Fixed in `15020ce`: the module now states that only the signed-baseline key stays off command lines and that the Gist ID appears in the `gh api` path.

tend-agent:
Re-checked all four against `15020ce` — they hold, and `uv run pytest shared/steps/test_gist_memory.py generator/tests/test_checks.py` is green (167 passed).

The conflict fix is the one that mattered, and the all-or-nothing shape closes the cross-file case I described: `conflicts` is now computed over the whole `baseline | local` key set before any PATCH, so the partial-state scenario (A's index left pointing at a file B deleted) can't arise — a run's notes land together or not at all. Worth being explicit about the cost, since it's inherent rather than a defect: the skipped run's notes are dropped, not deferred, because the memory directory is per-run and the next run restores from the Gist. That's the right trade against a self-inconsistent index.

The other three are straightforward: `enabled_harnesses() or {cfg.harness}` restores the always-one-auth-check guarantee, the rename is complete (no bare `MemoryError` left in the tree), and the module docstring now scopes the claim to the baseline key.

One item from my review body is still open, and it's a maintainer call rather than something to fix here: `memory_gist` is global across every enabled Claude workflow while `harness` and `model` are per-workflow, so an adopter can't run memory on `nightly` but not on `review`.

- [`shared/steps/gist_memory.py`](https://github.com/max-sixty/tend/blob/15020cedb89c911709534dd34e2b2b911e401970/shared/steps/gist_memory.py#L276-L288) — conflict branch returns before the PATCH; the loop still iterates `sorted(set(baseline) | set(local))`, so remote-only files added by an overlapping run are left alone rather than counted as conflicts.
- [`generator/src/tend/checks.py`](https://github.com/max-sixty/tend/blob/15020cedb89c911709534dd34e2b2b911e401970/generator/src/tend/checks.py#L1732) — fallback covers the disable-all config that `test_memory_gist_ignores_a_disabled_top_level_harness` reaches.
- `grep -rn '\bMemoryError'` over `*.py`/`*.md`/`*.yaml`, excluding `GistMemoryError`: no hits.
- Docs: `docs/security-model.md` and both copies of `tend.example.yaml` now describe skip-the-whole-save; no stale "preserves remote edits" or "skipped concurrent changes" wording left anywhere.

</details>

docs/tend.example.yaml
```diff
@@ -237,6 +240,29 @@ bot_name: my-project-bot
...
+#
+#     gh secret set TEND_MEMORY_GIST_ID --repo OWNER/REPO --env tend
+#
+# Secret Gists are unlisted, not private, so this experiment supports public
+# repositories only and memory must not contain credentials or private facts.
+# Recalled notes are untrusted context: they may be stale or influenced by text
+# Claude read in an earlier run. Gists are flat, so a nested memory directory
+# makes that run's save fail with a warning. If any file changed remotely after
+# restore, the run skips its entire save. GitHub's Gist API still cannot make
+# the final update atomic.
```
```suggestion
# makes that run's save fail with a warning. If a file this run changed also
# changed remotely, the run skips its entire save. GitHub's Gist API still
# cannot make the final update atomic.
```

plugins/install-tend/skills/install-tend/references/tend.example.yaml
```diff
@@ -237,6 +240,29 @@ bot_name: my-project-bot
...
+#
+#     gh secret set TEND_MEMORY_GIST_ID --repo OWNER/REPO --env tend
+#
+# Secret Gists are unlisted, not private, so this experiment supports public
+# repositories only and memory must not contain credentials or private facts.
+# Recalled notes are untrusted context: they may be stale or influenced by text
+# Claude read in an earlier run. Gists are flat, so a nested memory directory
+# makes that run's save fail with a warning. If any file changed remotely after
+# restore, the run skips its entire save. GitHub's Gist API still cannot make
+# the final update atomic.
```
tend-agent: Same wording in the mirrored copy.

```suggestion
# makes that run's save fail with a warning. If a file this run changed also
# changed remotely, the run skips its entire save. GitHub's Gist API still
# cannot make the final update atomic.
```

tend-agent (approved):
All four earlier findings land. `save` now computes the conflict set and returns before the PATCH, so a run's notes go up together or not at all; the auth check falls back to `cfg.harness` when nothing is enabled; and the `GistMemoryError` rename is complete (no `MemoryError` reference survives anywhere in the tree). `uv run pytest generator/tests/test_checks.py shared/steps/test_gist_memory.py` is green here — 167 passed — and the new conflict test is load-bearing rather than decorative, since `FakeGh` raises on an unregistered argv, so a stray PATCH would fail it rather than slip past the `called(...)` assertion.

One inaccuracy in the new config comment, suggested inline on both copies. The run doesn't skip on *any* remote change, only on a remote change to a file it also changed: `if after == before: continue` runs before the remote is consulted, so a file another run edited or added but this one left alone never reaches the conflict branch and the save proceeds. That's the behaviour you want — the sentence just promises more than the code delivers, which matters here because the whole point of the all-or-nothing change was to make the consistency guarantee legible.

docs/tend.example.yaml
max-sixty: Fixed in `eb96138`: the example now distinguishes conflicts on files this run changed from unrelated remote-only additions.

plugins/install-tend/skills/install-tend/references/tend.example.yaml
max-sixty: Fixed in `eb96138`: the example now distinguishes conflicts on files this run changed from unrelated remote-only additions.

PR #1112, opened 2026-08-31
fix(generator): silence GraphQL shellcheck false positive

max-sixty:
The notifications pre-check intentionally passes a single-quoted GraphQL query containing `$q`; GitHub's GraphQL API expands that variable, not the shell. Actionlint's embedded ShellCheck therefore reported SC2016 on workflows generated by 0.1.23.

Add a command-scoped ShellCheck directive to the source template and update all affected render snapshots. This keeps the query semantics unchanged while making freshly generated workflows pass actionlint.

Validation:

- `wt test` (880 Python tests, 32 worker tests, TypeScript typecheck)
- freshly rendered `tend-notifications.yaml` piped through `actionlint`

### DIFF
--- a/generator/src/tend/templates/notifications-check.sh
+++ b/generator/src/tend/templates/notifications-check.sh
@@ -39,6 +39,7 @@ echo "count=$COUNT" >> "$GITHUB_OUTPUT"
 # boot gate; the agent test-merges every candidate before changing a branch.
 # Read the newest comments: a deferral is normally the PR's latest activity.
 # An older marker can waste boots, but the resolver paginates before acting.
+# shellcheck disable=SC2016  # $q is a GraphQL variable, not a shell variable.
 if BOT_LOGIN=$(gh api user --jq .login 2>/dev/null) \
   && PRS=$(gh api graphql -f query='
     query($q: String!) {
--- a/generator/tests/_regtest_outputs/test_generate.test_fork_guard_rendered_shape_regtest[tend-notifications.yaml].out
+++ b/generator/tests/_regtest_outputs/test_generate.test_fork_guard_rendered_shape_regtest[tend-notifications.yaml].out
@@ -74,6 +74,7 @@ jobs:
           # boot gate; the agent test-merges every candidate before changing a branch.
           # Read the newest comments: a deferral is normally the PR's latest activity.
           # An older marker can waste boots, but the resolver paginates before acting.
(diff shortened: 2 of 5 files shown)


tend-agent (approved):
Fix is correct and complete. I reproduced the premise with actionlint v1.7.12 (the pre-commit pin): without the directive a freshly rendered `tend-notifications.yaml` fails on `SC2016:info`, with it all 36 generated-workflow snapshots lint clean. Placing it on the `if` compound also covers the `jq -er` filter's `$bot`/`$pr`, which is what you want there.

One thing this doesn't close, and I think it's a follow-up rather than a change here: nothing in this repo would have caught the finding before it reached an adopter. The pre-commit `shellcheck` hook lints `generator/src/tend/templates/*.sh` at `-S warning`, SC2016 is `info`, and actionlint — which is what the rendered output actually meets — has no severity filter and reports info. So the templates get a strictly weaker lint here than their rendered form gets downstream, and this class of break is invisible until a release plus a regen carries it into someone's workflow file.

That gap is written down as an assumption in two places, both of which read as if the severities match: the shellcheck-hook comment in `.pre-commit-config.yaml` ("run shellcheck directly at the same `-S warning` severity"), and the docstring of `test_inline_run_bodies_pass_shellcheck` in `generator/tests/test_repo_pins.py` ("Severity matches the shellcheck hook in .pre-commit-config.yaml, which matches actionlint's own default"). I checked the second claim directly and it isn't right.

Happy to open a follow-up PR adding a templates-only shellcheck hook at actionlint's severity and correcting that docstring — say the word.

```
# actionlint reports info-level SC2016 (no severity filter):
$ cat .github/workflows/t.yaml   # run: |  echo 'expand $HOME please'
$ actionlint
.github/workflows/t.yaml:7:9: shellcheck reported issue in this script:
  SC2016:info:1:6: Expressions don't expand in single quotes [shellcheck]

# The PR's snapshots, run through actionlint with .github/actionlint.yaml present:
#   with the directive    -> 36 snapshots, exit 0
#   directive stripped    -> tend-notifications.yaml:34:9  SC2016:info:43:36

# shellcheck -S warning on the template is silent either way (SC2016 is info).
$ shellcheck -S warning -s bash generator/src/tend/templates/notifications-check.sh  # exit 0

$ uv run pytest generator/tests/test_generate.py -q   # 152 passed, 0 failed snapshots
```

</details>

PR #1115, opened 2026-08-31
docs(CLAUDE.md): name memory_gist as the second use of the gist PAT scope

tend-agent:
`memory_gist` landed in #1110 and was enabled on this repo in #1114, but two places that justify the `gist` PAT scope still describe it as existing for `review-reviewers`' evidence store alone. That scope now has a second, more load-bearing consumer: every Claude workflow run restores and saves auto memory through it.

- `CLAUDE.md`'s Auth section — adds the second use to the sentence that enumerates them.
- `docs/tend.example.yaml` and its synced copy under `plugins/install-tend/` — the file an adopter reads when minting the PAT. Its "internal skills store structured evidence" wording covers `review-reviewers` but not the memory sync, which runs from a step in [`claude/action.yaml`](https://github.com/max-sixty/tend/blob/75d7f1f064564e93bffaf3d6752e77ffee0326cb/claude/action.yaml#L314-L337) rather than a skill.

Doc-only, so no regression test — the claim it fixes isn't executable. Verified against the feature's own docs: [`docs/tend.example.yaml`](https://github.com/max-sixty/tend/blob/6057b75/docs/tend.example.yaml#L243-L264) and [`docs/security-model.md`](https://github.com/max-sixty/tend/blob/6057b75/docs/security-model.md#L429-L437). `uv run pytest` (880 passed) and `pre-commit` on the changed files both pass; the sync hook confirms the two example-yaml copies stay identical.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -219,11 +219,13 @@ stored as secrets in the repo's `tend` GitHub Environment, whose deployment
 branch policy admits only the branches `tend check` confirmed the bot
 cannot write — the default branch and any `protected_branches` that exist
 and are protected. A workflow the bot pushes to any other ref is refused
-them before its first step. The `gist` scope supports bot-owned secret gists
-used by `review-reviewers` as a per-month structured evidence store
-(avoids the 65 KB comment-body limit). The `user` scope lets `install-tend`
-set the bot's profile bio (`PATCH /user`) so the account's authorization
-stance is discoverable on the bot's user page.
+them before its first step. The `gist` scope supports bot-owned secret
+gists, of which there are two: `review-reviewers` keeps a per-month
+structured evidence store (avoids the 65 KB comment-body limit), and the
+experimental `memory_gist` setting persists Claude Code's auto memory
+across runs. The `user` scope lets `install-tend` set the bot's profile
+bio (`PATCH /user`) so the account's authorization stance is discoverable
+on the bot's user page.
 
 Classic PATs are all-or-nothing — `public_repo` grants full write to every
 public repo the user can access. Fine-grained PATs allow per-category
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -219,8 +219,8 @@ stored as secrets in the repo's `tend` GitHub Environment, whose deployment
 branch policy admits only the branches `tend check` confirmed the bot
 cannot write — the default branch and any `protected_branches` that exist
 and are protected. A workflow the bot pushes to any other ref is refused
-them before its first step. The `gist` scope supports bot-owned secret
-gists, of which there are two: `review-reviewers` keeps a per-month
+them before its first step. Two things use the `gist` scope, both
+through bot-owned secret gists: `review-reviewers` keeps a per-month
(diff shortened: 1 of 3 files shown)


tend-agent (commented):
One wording fix inline. Separately, the adopter-facing copy of this justification has the same gap: `docs/tend.example.yaml` (identical file at `plugins/install-tend/skills/install-tend/references/tend.example.yaml`) still explains the scope under `Classic PAT scopes` as "`gist` lets internal skills store structured evidence in secret gists owned by the bot" — which covers `review-reviewers` and not the memory sync, since that runs from a trusted step in `claude/action.yaml` rather than a skill. That file is what an adopter reads when minting the PAT, so it's the more load-bearing of the two. Happy to extend it in the same way here.

CLAUDE.md
```diff
@@ -219,11 +219,13 @@ stored as secrets in the repo's `tend` GitHub Environment, whose deployment
 branch policy admits only the branches `tend check` confirmed the bot
 cannot write — the default branch and any `protected_branches` that exist
 and are protected. A workflow the bot pushes to any other ref is refused
-them before its first step. The `gist` scope supports bot-owned secret gists
-used by `review-reviewers` as a per-month structured evidence store
-(avoids the 65 KB comment-body limit). The `user` scope lets `install-tend`
-set the bot's profile bio (`PATCH /user`) so the account's authorization
-stance is discoverable on the bot's user page.
+them before its first step. The `gist` scope supports bot-owned secret
+gists, of which there are two: `review-reviewers` keeps a per-month
```
tend-agent: "Two" counts consumers, not gists — the sentence reads as a gist count and the next clause contradicts it. `review-reviewers` creates one gist per `(target repo, month)` pair (`plugins/tend-ci-runner/skills/review-reviewers/SKILL.md`, under "Evidence accumulation"), so the bot accumulates a new one every month per target, plus the single `tend auto memory: OWNER/REPO` gist. The PR title has it right; the body doesn't.

```suggestion
them before its first step. Two things use the `gist` scope, both
through bot-owned secret gists: `review-reviewers` keeps a per-month
```

tend-agent: Applied — the count was consumers, not gists. Reworded to "Two things use the `gist` scope, both through bot-owned secret gists:" in 75d7f1f.

tend-agent:
Extended the adopter-facing copy too — you're right that it's the more load-bearing of the two, since it's what someone reads while minting the PAT. `gist` there now reads as covering two uses of bot-owned secret gists: skill-written structured evidence, and the experimental `memory_gist` sync. Confirmed the sync is an action step rather than a skill ([`claude/action.yaml`](https://github.com/max-sixty/tend/blob/75d7f1f064564e93bffaf3d6752e77ffee0326cb/claude/action.yaml#L314-L337) shells out to `shared/steps/gist_memory.py`), so the old "internal skills" framing did leave it out.

Both copies edited together and the `sync install-tend references/` hook confirms they're still identical. Inline wording fix applied as well; PR body updated to cover the wider scope. `uv run pytest` passes (880) and `pre-commit` is clean on the changed files.

PR #1117, opened 2026-08-31
test(nightly): pin Step 7's staging pathspecs to what init writes

### LINKED ISSUE #1116: nightly: Step 7 never commits the generated .github/actionlint.yaml, and regenerates it untracked every run
Since max-sixty/tend#1095, `uvx tend@latest init` writes `.github/actionlint.yaml` alongside the eight workflow files. The `nightly` skill's Step 7 recipe doesn't stage or detect it, so on an adopter repo it lands untracked and stays that way.

Three lines in the recipe scope to `.github/workflows .config`, and the new file is under neither:

```bash
git status --porcelain .github/workflows .config
NON_STAMP_DIFF=$(git diff --no-color .github/workflows .config | ...)
```

Two consequences, in order of how quietly they fail:

1. **The file is never committed.** `git add -A .github/workflows .config` skips it, so the PR ships the regenerated workflows without the ignore they need. `gh pr create` prints a one-line `Warning: 1 uncommitted change` and nothing else notices — the recipe's own `git status` is scoped away from it, so the session has no signal unless it happens to read that warning.
2. **Every future nightly re-creates it.** The worktree is `--force`-removed at the end of the step, so the untracked file is discarded and regenerated next run. An adopter that lints workflows stays red indefinitely, which is the exact failure max-sixty/tend#1094 was filed for.

There's also a smaller correctness point in the stamp-only skip: `NON_STAMP_DIFF` is computed from `git diff`, which sees tracked files only. On a release where the *sole* change is that a newly generated file appeared, the diff is empty, `NON_STAMP_DIFF` is `0`, and the step skips the PR entirely — the "no value" branch fires on a run that does have value.

Observed on max-sixty/worktrunk tonight, regenerating 0.1.22 → 0.1.24. I committed the file by hand in [max-sixty/worktrunk#3969](https://github.com/max-sixty/worktrunk/pull/3969), but the next adopter hits the same thing.

Widening the three scopes to include the file would fix all three symptoms, e.g. `.github/workflows .github/actionlint.yaml .config`. Naming the generator's output paths in one place the recipe reads would be sturdier — the same gap reopens the next time `init` learns to write something new outside `.github/workflows`.

tend-agent:
The nightly Step 7 gap max-sixty/tend#1116 reports was already closed on `main` — #1095 widened the recipe's pathspecs to `.github .config` and moved the stamp-only check to `git diff --cached` in the same commit that taught the generator to write `.github/actionlint.yaml`. What #1095 did not add is anything that keeps the pair in sync, so this adds the missing contract test.

The adopter still saw the bug because the two halves ship on different clocks: the nightly skill comes from the action ref pinned in the adopter's workflow files (the *previous* release), while `uv tool run tend@latest init` is always the newest generator. A release that adds an output path is therefore run by the prior release's recipe for exactly one night. That skew is structural and this PR does not change it; what it does prevent is the recipe and the generator disagreeing *within* a release, which is what turns a one-night gap into a permanent one.

`test_nightly_regen_stages_every_path_init_writes` runs `tend init` into a temp repo and asserts every file it writes is covered by the `git add -A` pathspecs parsed out of the nightly skill, that both `git add -A` lines in Step 7 name the same set, and that the stamp-only skip reads `git diff --cached` rather than the worktree.

<details><summary>Verification that the guard catches the reported bug</summary>

Run against the recipe as it stood at `0.1.22`, the version the reporter's adopter was on:

```
0.1.22 pathspecs: ['.github/workflows .config']
has --cached: False
uncovered under 0.1.22 recipe: ['.github/actionlint.yaml']
```

Both assertions fail there and pass on `main`. Local run: `uv run pytest generator/tests/test_cross_file_contracts.py` — 14 passed; `pre-commit run --files generator/tests/test_cross_file_contracts.py` clean.

</details>

---
Closes #1116 — automated triage

### DIFF
--- a/generator/tests/test_cross_file_contracts.py
+++ b/generator/tests/test_cross_file_contracts.py
@@ -272,8 +272,11 @@ def test_nightly_regen_stages_every_path_init_writes(
     # The stamp-only skip and the `git status` inspection both read the staged
     # set: a file `init` newly created is invisible to a plain `git diff`, so a
     # release whose only change is a new output path would skip the PR as a
-    # no-op.
-    assert "git diff --cached" in skill
+    # no-op. They must name the same paths the `git add -A` lines stage —
+    # widening only the staging leaves the no-op check blind to the new path.
+    specs = " ".join(staged)
+    assert f"git status --porcelain {specs}" in skill
+    assert f"git diff --cached --no-color {specs}" in skill
 
     config = tmp_path / ".config" / "tend.yaml"
     config.parent.mkdir(parents=True)
--- a/generator/tests/test_cross_file_contracts.py
+++ b/generator/tests/test_cross_file_contracts.py
@@ -5,7 +5,12 @@ together, so each reads correct on its own while the pair stops agreeing.
 """
 
 import json
+import re
 from pathlib import Path
+
+import pytest
+from click.testing import CliRunner
+from tend.cli import main
 
 from tests import _yaml as yaml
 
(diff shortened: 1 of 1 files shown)


generator/tests/test_cross_file_contracts.py
```diff
@@ -244,3 +249,51 @@ def test_review_reviewers_matrix_covers_consumers() -> None:
...
+    pathspecs = re.findall(r"^git add -A (.+)$", skill, re.MULTILINE)
+    assert pathspecs, "Step 7 no longer stages the regenerated files"
+    assert len(set(pathspecs)) == 1, f"Step 7 stages inconsistent sets: {pathspecs}"
+    staged = pathspecs[0].split()
+
+    # The stamp-only skip and the `git status` inspection both read the staged
+    # set: a file `init` newly created is invisible to a plain `git diff`, so a
+    # release whose only change is a new output path would skip the PR as a
+    # no-op.
+    assert "git diff --cached" in skill
```
tend-agent:
```suggestion
    # The stamp-only skip and the `git status` inspection both read the staged
    # set: a file `init` newly created is invisible to a plain `git diff`, so a
    # release whose only change is a new output path would skip the PR as a
    # no-op. They must name the same paths the `git add -A` lines stage —
    # widening only the staging leaves the no-op check blind to the new path.
    specs = " ".join(staged)
    assert f"git status --porcelain {specs}" in skill
    assert f"git diff --cached --no-color {specs}" in skill
```

tend-agent (commented):
The guard closes the `git add -A` half of the contract but leaves the other three pathspec lists in Step 7 unpinned. `git status --porcelain .github .config`, `git diff --cached --no-color .github .config`, and the second `git add -A` all name the same set today, and the test only asserts the two `git add -A` lines agree — `assert "git diff --cached" in skill` matches the substring anywhere in the file, not the pathspecs on it.

That leaves the exact drift this test exists to catch still reachable, one step later: a future release adds an output path, this test fails, someone widens both `git add -A` lines to `.github .config .foo` and stops there. The test goes green, the file is staged — and the stamp-only check still diffs only `.github .config`, so `NON_STAMP_DIFF` is `0` when the new path is the only change and Step 7 skips the regeneration PR as a no-op. Same permanent gap, different line.

Inline suggestion reuses the already-parsed `staged` to pin the companions. Verified: the assertions hold against Step 7 as written, and both fail if either `git add -A` line is widened alone.

generator/tests/test_cross_file_contracts.py
tend-agent: Applied in 67987f6 — `specs = " ".join(staged)` now pins both companions, so the three pathspec lists move together or the test goes red.

Verified the drift the review describes: widening both `git add -A` lines to `.github .config .foo` and leaving the other two alone fails on `assert f"git status --porcelain {specs}" in skill`. Against Step 7 as written, `generator/tests/test_cross_file_contracts.py` is 14 passed and `pre-commit run --files ...` is clean.

PR #1119, opened 2026-08-31
fix(nightly): capture the regen commit's OID before removing the worktree

### LINKED ISSUE #1118: nightly: Step 7 removes the regen worktree without capturing the pushed OID, so the poll pin comes from the PR head or an invented SHA
## Problem

`nightly` Step 7 makes the workflow-regen commit inside `/tmp/tend-update-workflows`, pushes it, opens the PR, and then ends the block by destroying that worktree:

```bash
git commit -m "$TITLE"
git push -u origin tend/update-workflows
gh pr create --title "$TITLE" --body-file "/tmp/tend-update-body.md"
cd -
git worktree remove "/tmp/tend-update-workflows" --force
```

The very next obligation is a poll pinned to the commit the session is accountable for — Step 8 and **CI Monitoring** in `running-in-ci` both require it, and CI Monitoring prescribes `PINNED_SHA=$(git rev-parse HEAD)` "after your own push".

By then the recipe has deleted the only place that command would have worked. In the main checkout `git rev-parse HEAD` resolves to the default branch's head — a different commit on a different branch — so the prescribed derivation isn't merely unavailable here, it silently returns the wrong SHA. The block never captures the OID it just pushed, and `git commit`'s own output gives only the abbreviated form.

That leaves a session two routes, and both are ones tend has already ruled out elsewhere:

1. **Read the PR head** — `gh pr view <n> --json headRefOid`. This is the head-following source that max-sixty/tend#829 documented and #965 removed from the poll itself: a sibling `tend-mention` pushing to the same branch while the loop sleeps retargets the pin, and the session reports another commit's green as its own. `tend-mention` fires on every review of a bot-authored PR, so a sibling on that branch is the normal case for exactly this PR.
2. **Retype the abbreviated OID from `git commit`'s output** — which produces a well-formed but nonexistent SHA, the shape max-sixty/tend#1045 and max-sixty/tend#1054 covered on the script side.

The script-side hardening from #1046 and #1055 (shipped in 0.1.21) turns route 2 into a fast `NOSUCHCOMMIT` verdict instead of a nine-minute stall, which is a real improvement — but neither addresses route 1, and neither gives the caller a correct SHA to pass in the first place.

## Evidence

Two occurrences on `numbagg/numbagg`, one per route, in the two windows where the nightly produced a workflow-update PR. Both are structural: the recipe removes the worktree unconditionally, so every consumer's nightly reaches the poll with no correct local source.

**Route 2 — [run 33099573280](https://github.com/numbagg/numbagg/actions/runs/33099573280), 2026-08-27, PR numbagg/numbagg#764, under 0.1.18.** Session JSONL timestamps:

| Time | Event |
|---|---|
| 17:44:01.676Z | `git commit` prints `[tend/update-workflows 0446cc2]` — abbreviated |
| 17:44:05.749Z | `git worktree remove /tmp/tend-update-workflows --force` |
| 17:50:32.629Z | For an *unrelated* PR whose branch lives in the main checkout, `PINNED_SHA=$(git rev-parse HEAD)` — correct, and polls clean |
| 18:00:23.651Z | For the regen PR, `poll-pr-checks.sh 764 0446cc2ba4f5e5fa8fdd6dc0d7d5f6d8ae09c1f5` — characters 8 onward invented |
| 18:09:26.726Z | `no rollup returned for 0446cc2ba4… — UNVERIFIED, not green` + a false `note: branch advanced to 0446cc2ff4…` |

The real head is `0446cc2ff4a544537f19c97b0d273a904404973f`. Nine minutes of runner time, then self-correction on the next turn. The same session getting the *other* PR's pin right one command earlier is the cleanest statement of the cause: the difference isn't the model's care, it's whether the branch still exists locally.

**Route 1 — [run 33399946874](https://github.com/numbagg/numbagg/actions/runs/33399946874), 2026-08-31, PR numbagg/numbagg#777, under 0.1.22.** One Bash call, verbatim:

```bash
cd /home/runner/work/numbagg/numbagg
git worktree remove "/tmp/tend-update-workflows" --force && echo "worktree removed"
PINNED_SHA=$(gh pr view 777 --json headRefOid --jq '.headRefOid')
echo "PINNED_SHA=$PINNED_SHA" && echo "$PINNED_SHA" > /tmp/pinned-sha
```

This one resolved to the session's own commit `c87837a0` and the poll was honest — but only on timing. A sibling `tend-mention` pushed `15637727` to the same branch nine minutes later, and the session's second poll shows the head note firing for real. Had the two landed in the other order, the pin would have been the sibling's commit.

Neither occurrence put a wrong claim on a public thread, so the cost so far is runner minutes and one recovered turn. What makes it worth fixing anyway is that the recipe steers every session toward one of two sources the project has already gone to some length to eliminate.

## Proposed fix

Capture the OID inside the worktree before removing it, using the same `/tmp` stash idiom the block already uses for `/tmp/tend-old-ver` (the recipe's own comment explains why: "Shell state doesn't persist between bash calls"):

```bash
 git commit -m "$TITLE"
 git push -u origin tend/update-workflows
 gh pr create --title "$TITLE" --body-file "/tmp/tend-update-body.md"
+# Stash the pushed OID before the worktree goes: the poll below must pin to
+# the commit this run pushed, and after the removal `git rev-parse HEAD` in
+# the main checkout resolves to the default branch, not to this PR.
+git rev-parse HEAD > "/tmp/tend-update-sha"
 cd -
 git worktree remove "/tmp/tend-update-workflows" --force
```

and point the poll at it, so the next step doesn't have to re-derive where the SHA comes from:

```bash
"${CLAUDE_PLUGIN_ROOT}/scripts/poll-pr-checks.sh" <pr-number> "$(cat /tmp/tend-update-sha)"
```

One line plus a pointer, inside an existing block, replacing a network round-trip with a local read. Nothing new to maintain and no new failure mode — if the capture is ever missing the session is no worse off than today.

Worth considering alongside it: **CI Monitoring**'s `PINNED_SHA=$(git rev-parse HEAD)` is written as if the push always happened in the checkout the session is standing in. That holds for `ci-fix` and for `nightly`'s Step 8 fixes, and it's already carved out for review sessions on the merge ref — the worktree case is a third shape it doesn't name. A clause noting that a push made from a `/tmp` worktree needs its OID captured before the worktree is removed would cover any future recipe with the same shape, not just this one.

<details><summary>How this relates to the already-closed threads</summary>

- max-sixty/tend#829 / #837 / #965 — the poll's *internals* now query the rollup by commit OID rather than following the PR head. That fix is sound and holds; this issue is about the OID handed to it.
- max-sixty/tend#1045, max-sixty/tend#1046 — abbreviated SHA rejected before the loop, shipped 0.1.19.
- max-sixty/tend#1054, max-sixty/tend#1055 — well-formed-but-nonexistent OID reaches a verdict at ~2 min instead of 9, shipped 0.1.21. #1054's triage read the caller side as "a model lapse I'm treating as stochastic". That reading is right for the general case, and route 2 above did recur under it — but the worktree case is the one place where it isn't a lapse: the correct command is absent by construction, so no amount of caller-side care produces the right answer.
- max-sixty/tend#1116 / #1117 — also Step 7, but the staging pathspecs. No overlap.

Checked against `plugins/tend-ci-runner/skills/nightly/SKILL.md` on `main` at the time of writing; the capture is not there.

</details>

tend-agent:
## Problem

Step 7 of the nightly skill commits and pushes the workflow regeneration inside `/tmp/tend-update-workflows`, opens the PR, then destroys that worktree. The poll that must follow — **CI Monitoring** in `running-in-ci` prescribes `PINNED_SHA=$(git rev-parse HEAD)` "after your own push" — has nowhere left to read the OID from: back in the main checkout `HEAD` is the default branch, a different commit on a different branch, so the prescribed derivation doesn't just fail, it returns a plausible wrong SHA. That leaves the PR head (which a sibling `tend-mention` push retargets mid-poll — the head-following source max-sixty/tend#965 removed from the poll itself) or the abbreviated OID retyped out of `git commit`'s output. Reported in #1118 with one observed occurrence of each route on a consumer repo.

## Solution

Stash the OID inside the worktree, before the removal, using the `/tmp` idiom the block already uses for `/tmp/tend-old-ver`, and point Step 7's poll at it so the next step has nothing to re-derive. Also names the worktree case in **CI Monitoring**'s `PINNED_SHA` block, which until now only spelled out the plain-checkout and review-merge-ref shapes.

## Testing

`test_nightly_regen_pins_its_poll_to_the_commit_it_pushed` in `generator/tests/test_cross_file_contracts.py` parses Step 7's ship block, requires a `git rev-parse HEAD > /tmp/...` capture ordered before `git worktree remove`, and requires the stashed path to reach a `poll-pr-checks.sh` invocation. It fails on `main` with "Step 7 pushes from the /tmp worktree and then removes it without recording the pushed OID" and passes here. Full `uv run pytest`: 881 passed. `pre-commit` clean on the changed files.

---
Closes #1118 — automated triage

### DIFF
--- a/generator/tests/test_cross_file_contracts.py
+++ b/generator/tests/test_cross_file_contracts.py
@@ -244,3 +244,53 @@ def test_review_reviewers_matrix_covers_consumers() -> None:
 
     missing = sorted(set(consumers) - set(matrix))
     assert not missing, f"add consumers to review-reviewers.yaml matrix: {missing}"
+
+
+def _bash_blocks(markdown: str) -> list[str]:
+    return [block.split("```", 1)[0] for block in markdown.split("```bash\n")[1:]]
+
+
+def test_nightly_regen_pins_its_poll_to_the_commit_it_pushed() -> None:
+    """Step 7 must stash the pushed OID before it removes the regen worktree.
+
+    The commit is made on `tend/update-workflows` inside `/tmp`, and the block
+    destroys that worktree on the way out. Afterwards the main checkout's
+    `git rev-parse HEAD` — the derivation **CI Monitoring** prescribes "after
+    your own push" — resolves to the default branch, a different commit on a
+    different branch, so the session is steered to the two sources tend has
+    ruled out: the PR head (which a sibling push retargets mid-poll) or the
+    abbreviated OID retyped out of `git commit`'s output.
+    """
+    skill = _read("plugins", "tend-ci-runner", "skills", "nightly", "SKILL.md")
+
+    ship = [
+        block
+        for block in _bash_blocks(skill)
+        if "gh pr create" in block and "git worktree remove" in block
+    ]
+    assert len(ship) == 1, "Step 7 no longer ships the regen PR from one block"
+    block = ship[0]
+
+    capture = [
+        line
+        for line in block.splitlines()
+        if line.startswith("git rev-parse HEAD > ") and "/tmp/" in line
+    ]
+    assert capture, (
+        "Step 7 pushes from the /tmp worktree and then removes it without "
+        "recording the pushed OID; the poll that follows has no correct local "
+        "source for it."
+    )
+    assert block.index(capture[0]) < block.index("git worktree remove"), (
+        "the OID must be captured while the worktree still exists"
+    )
+
+    stash = capture[0].split(">", 1)[1].strip().strip('"')
+    poll = [
+        line
+        for line in skill.splitlines()
+        if "poll-pr-checks.sh" in line and f"cat {stash}" in line
+    ]
+    assert poll, (
+        f"Step 7 stashes the pushed OID at {stash} but never points the poll at it"
+    )
--- a/generator/tests/test_cross_file_contracts.py
+++ b/generator/tests/test_cross_file_contracts.py
@@ -281,6 +281,9 @@ def test_nightly_regen_pins_its_poll_to_the_commit_it_pushed() -> None:
         "recording the pushed OID; the poll that follows has no correct local "
         "source for it."
     )
+    assert block.index("git commit") < block.index(capture[0]), (
+        "the OID must be captured after the commit it names"
+    )
     assert block.index(capture[0]) < block.index("git worktree remove"), (
         "the OID must be captured while the worktree still exists"
     )
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -255,8 +255,20 @@ git config --global user.email "${BOT_ID}+${BOT_LOGIN}@users.noreply.github.com"
 git commit -m "$TITLE"
 git push -u origin tend/update-workflows
 gh pr create --title "$TITLE" --body-file "/tmp/tend-update-body.md"
+# Stash the pushed OID before the worktree goes. The poll below pins to the
+# commit this run pushed, and once the worktree is removed the main checkout's
+# `git rev-parse HEAD` resolves to the default branch — a different commit on a
+# different branch — leaving the PR head (which a sibling push retargets
+# mid-poll) or a retyped abbreviated OID as the only sources.
+git rev-parse HEAD > "/tmp/tend-update-sha"
 cd -
 git worktree remove "/tmp/tend-update-workflows" --force
+```
+
+Then poll that commit's checks per **CI Monitoring** in `/tend-ci-runner:running-in-ci` — foreground, `timeout: 600000`:
+
+```bash
+${CLAUDE_PLUGIN_ROOT}/scripts/poll-pr-checks.sh <pr-number> "$(cat /tmp/tend-update-sha)"
 ```
 
 ## Step 8: Fix findings
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -276,6 +276,9 @@ PINNED_SHA=$(git rev-parse HEAD)
 # In a review session, HEAD is the ephemeral refs/pull/N/merge commit, which
 # carries no rollup at all; pin the PR head instead:
 #   PINNED_SHA=$(gh pr view <number> --json headRefOid --jq '.headRefOid')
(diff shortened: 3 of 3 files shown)


generator/tests/test_cross_file_contracts.py
```diff
@@ -244,3 +244,53 @@ def test_review_reviewers_matrix_covers_consumers() -> None:
...
+        if line.startswith("git rev-parse HEAD > ") and "/tmp/" in line
+    ]
+    assert capture, (
+        "Step 7 pushes from the /tmp worktree and then removes it without "
+        "recording the pushed OID; the poll that follows has no correct local "
+        "source for it."
+    )
+    assert block.index(capture[0]) < block.index("git worktree remove"), (
+        "the OID must be captured while the worktree still exists"
+    )
```
tend-agent, line 294: This constrains one end of the ordering only. Move the capture above `git commit` and the test stays green while the recipe stashes the pre-commit HEAD — a real, well-formed OID that isn't the commit that was pushed, which is the same wrong-SHA shape the change exists to close. I ran it: with `git rev-parse HEAD > "/tmp/tend-update-sha"` relocated above `git commit -m "$TITLE"`, this test still passes.

```suggestion
    assert block.index("git commit") < block.index(capture[0]), (
        "the OID must be captured after the commit it names"
    )
    assert block.index(capture[0]) < block.index("git worktree remove"), (
        "the OID must be captured while the worktree still exists"
    )
```

tend-agent (commented):
The recipe change lands the right way round — I checked out the base and confirmed `test_nightly_regen_pins_its_poll_to_the_commit_it_pushed` fails there and passes here. One gap in the pin itself, inline: it constrains where the capture sits relative to the worktree removal but not relative to the commit, so the wrong-but-plausible SHA it exists to rule out can still get past it.

generator/tests/test_cross_file_contracts.py
tend-agent: Applied in 8abfb06. I confirmed the hole first: relocating `git rev-parse HEAD > "/tmp/tend-update-sha"` above `git commit -m "$TITLE"` in the skill leaves the test green on the old assertion, and fails with `the OID must be captured after the commit it names` on the new one. Both ends of the ordering are now pinned, and the skill is unchanged.

PR #1121, opened 2026-08-31
fix(notifications): acknowledge each resolved thread instead of the whole repository

### LINKED ISSUE #1120: tend-notifications: repo-wide ack marks post-cutoff threads read, so deferred items are dropped instead of carried
The `notifications` skill's Step 4 defers unresolved work by acknowledging below the unresolved item's timestamp, on the assumption that `PUT /repos/{owner}/{repo}/notifications` leaves later threads unread. It does not — the repo-wide PUT marks *every* thread in the repo read regardless of `last_read_at`, so a deferred item is silently dropped from the recovery queue instead of carried to the next poll. A second bug feeds it: `notifications?before=$CUTOFF` returns threads updated after the cutoff, so the snapshot contains items the run does not own and then acks them away.

Both were observed in the same run on `PRQL/prql` ([run 33420580670](https://github.com/PRQL/prql/actions/runs/33420580670)). Nothing was lost this time — every thread had already been reviewed at head — but the mechanism is general, and it bites exactly the case Step 4 exists for: a subject whose dedicated workflow is still running.

**`before` is not honoured.** Step 1's snapshot command with `CUTOFF=2026-08-31T17:40:00Z` returned six threads, two of them updated after the cutoff:

```
25394372529  2026-08-31T17:40:43Z  chore: pre-commit autoupdate            (PR #6262)
25394267616  2026-08-31T17:44:34Z  chore: bump pyo3 from 0.27.2 to 0.28.3  (PR #6259)
```

**The PUT marks them read anyway.** Step 4 ran with `ACK_CUTOFF=2026-08-31T17:40:00Z` (every same-repo item before the cutoff had an outcome, so the snapshot cutoff was used). Immediately after:

```
$ gh api "notifications?all=true&per_page=20" --jq '.[] | {unread, updated_at, last_read_at, title: .subject.title}'
{"last_read_at":"2026-08-31T17:40:00Z","unread":false,"updated_at":"2026-08-31T17:44:34Z","title":"chore: bump pyo3 from 0.27.2 to 0.28.3"}
{"last_read_at":"2026-08-31T17:40:00Z","unread":false,"updated_at":"2026-08-31T17:40:43Z","title":"chore: pre-commit autoupdate"}
```

`updated_at` later than `last_read_at`, `unread: false`. The docs for that endpoint say "anything updated since this time will not be marked as read"; the observed behavior is the opposite.

**It was the PUT, not a sibling run.** All six threads carry `last_read_at` exactly `2026-08-31T17:40:00Z` — the value passed to the PUT. An event workflow acknowledging its own trigger via `PATCH /notifications/threads/{id}` would stamp the wall-clock time instead, and no `tend-*` run for either subject completed between the threads' `updated_at` and the PUT (`tend-review` 33419149014 and 33419456869 were still `in_progress`; `tend-mention` 33421179782 was `queued`).

**Both were Step 3 defer candidates.** `tend-review` was still in flight for #6262 and #6259 when the ack ran, which is the state Step 3 says to defer. Had those runs not already posted their reviews, the work would have been dropped with no thread left unread to recover it.

</details>

## Suggested fix

The two bugs compound, so the `before` filter can't be leaned on to fix this. `SKILL.md:104` currently forbids the mechanism that would work:

> Never acknowledge a same-repository thread individually.

Acknowledging resolved same-repo threads individually is the only approach that preserves the others — there is no "mark unread" in the REST API, so a repo-wide PUT cannot be walked back once it overshoots:

```bash
# Step 4, replacing the repo-wide PUT: ack each same-repository thread that
# reached an outcome, and leave deferred/unresolved ones untouched.
for id in $RESOLVED_THREAD_IDS; do
  gh api "notifications/threads/$id" -X PATCH --silent
done
```

That is a bigger change than it looks, since Step 4's cutoff wording exists to keep the next poll from repeating completed work — worth your call on which way to take it. The narrow alternative is to keep the PUT but only issue it when *every* thread in the snapshot has an outcome, including post-cutoff ones, and skip it entirely otherwise; that keeps the deferral correct at the cost of re-reading resolved threads on the next poll.

Either way, Step 1 should filter the snapshot on `updated_at < $CUTOFF` client-side rather than trusting `before`, so the run's ownership set matches what Step 4 acts on.

tend-agent:
`tend-notifications` Step 4 acknowledged the whole repository with `PUT /repos/{owner}/{repo}/notifications`, bounding the call one second below a deferred thread's `updated_at`. That makes the deferral a timestamp comparison GitHub controls rather than a decision the run makes: a repository-wide ack marks by timestamp, not by outcome, so it acts on threads the run never examined — and REST has no "mark unread" to walk an overshoot back. The thread it can drop is exactly the one the cutoff exists to spare: a subject whose dedicated workflow is still in flight, and which that workflow has since bumped.

Step 4 now PATCHes each thread that reached an outcome, one call per thread, same-repository or not, and leaves deferred and unresolved threads untouched. That deletes the cutoff arithmetic and the `Never acknowledge a same-repository thread individually` rule, whose real invariant — never acknowledge before an outcome — is now stated directly. Nothing in the rewritten skill rests on a claim about which timestamp either endpoint bounds on.

The report (#1120) attributed this to two GitHub bugs. The divergences it saw are real but narrower than "the endpoints are broken", and one of its suggested changes would make things worse.

**What the reported run shows.** The snapshot was taken by the check step at `17:50:01Z` with `before=2026-08-31T17:40:00Z`, and returned six threads — including two whose `updated_at` had already reached their final values nine minutes earlier, so the divergence is not the snapshot being read back later:

| thread | PR | unread since | `prql-bot` review | thread `updated_at` |
|---|---|---|---|---|
| 25394372529 | PRQL/prql#6262 | 17:21:47 | 17:40:21 | 17:40:43 |
| 25394267616 | PRQL/prql#6259 | 17:15:10 | 17:44:11 | 17:44:34 |

Both became unread before the cutoff and were bumped past it by the bot's own reviews, which do not re-notify. [List notifications](https://docs.github.com/en/rest/activity/notifications) documents `before` as "Only show notifications updated before the given time", and the repository-wide PUT documents `last_read_at` as "Anything updated since this time will not be marked as read" — under those semantics neither thread should have been returned by the snapshot, and neither should have been marked read by the PUT. Both were.

**The boundary behaves as documented when nothing was bumped.** Neither probe reproduces the divergence in isolation:

```
$ gh api "notifications?all=true&before=2026-08-31T06:00:00Z&per_page=5" --paginate --slurp \
    | jq '[add // [] | .[] | .updated_at] | sort | {n: length, newest: .[-1]}'
{ "n": 494, "newest": "2026-08-31T05:53:40Z" }
```

`PUT /repos/max-sixty/tend/notifications` with `last_read_at=2026-08-31T17:00:00Z` returned `205 Reset Content`, and the one thread updated after it stayed unread across a minute of rechecks:

```
{"id":"25394889617","last_read_at":null,"unread":true,"updated_at":"2026-08-31T17:55:10Z"}
```

Neither probe covers the case that diverged — a thread bumped after it became unread and before the call — so the underlying semantics stay unsettled, and this PR does not try to settle them. Acknowledging exactly the threads that reached an outcome needs no timestamp reasoning at all, which is why it is the right shape under either reading.

**Why the report's other suggestion is not taken.** Filtering the snapshot client-side on `updated_at < $CUTOFF` would have dropped both threads above, which the run did own and did handle. Step 1 now records the observation — a returned thread's `updated_at` can be past the cutoff — and says to take the snapshot's membership as the run's scope rather than re-deriving it.

</details>

---
Closes #1120 — automated triage

### DIFF
--- a/generator/src/tend/templates/notifications-check.sh
+++ b/generator/src/tend/templates/notifications-check.sh
@@ -6,9 +6,9 @@
 # env: GITHUB_REPOSITORY, GITHUB_OUTPUT, GITHUB_TOKEN
 
 # Activity newer than this belongs to an event workflow that may still be
-# running. The same cutoff is passed to the agent and, once every older item has
-# a semantic outcome, to GitHub's repository-level mark-read endpoint. Newer
-# activity therefore cannot be acknowledged by this run.
+# running. The cutoff is passed to the agent, which acknowledges each thread it
+# resolves individually. Newer activity never enters the snapshot, so this run
+# cannot acknowledge it.
 CUTOFF=$(date -u -d '10 minutes ago' +%Y-%m-%dT%H:%M:%SZ)
 echo "cutoff=$CUTOFF" >> "$GITHUB_OUTPUT"
 
--- a/generator/src/tend/templates/notifications-check.sh
+++ b/generator/src/tend/templates/notifications-check.sh
@@ -6,9 +6,9 @@
 # env: GITHUB_REPOSITORY, GITHUB_OUTPUT, GITHUB_TOKEN
 
 # Activity newer than this belongs to an event workflow that may still be
-# running. The cutoff is passed to the agent, which acknowledges each thread it
-# resolves individually. Newer activity never enters the snapshot, so this run
-# cannot acknowledge it.
+# running. The cutoff bounds the snapshot below, which is passed to the agent;
+# the agent acknowledges each thread it resolves individually, so this run can
+# only acknowledge threads the snapshot returned.
 CUTOFF=$(date -u -d '10 minutes ago' +%Y-%m-%dT%H:%M:%SZ)
 echo "cutoff=$CUTOFF" >> "$GITHUB_OUTPUT"
 
--- a/generator/src/tend/templates/notifications-check.sh
+++ b/generator/src/tend/templates/notifications-check.sh
@@ -6,9 +6,9 @@
 # env: GITHUB_REPOSITORY, GITHUB_OUTPUT, GITHUB_TOKEN
 
 # Activity newer than this belongs to an event workflow that may still be
-# running. The cutoff bounds the snapshot below, which is passed to the agent;
-# the agent acknowledges each thread it resolves individually, so this run can
-# only acknowledge threads the snapshot returned.
+# running. The cutoff bounds the snapshot below and is the value passed to the
+# agent, which takes its own snapshot with it and acknowledges each thread it
+# resolves individually — so this run acknowledges only threads it examined.
 CUTOFF=$(date -u -d '10 minutes ago' +%Y-%m-%dT%H:%M:%SZ)
 echo "cutoff=$CUTOFF" >> "$GITHUB_OUTPUT"
 
--- a/plugins/tend-ci-runner/skills/notifications/SKILL.md
+++ b/plugins/tend-ci-runner/skills/notifications/SKILL.md
@@ -9,7 +9,7 @@ metadata:
 
 Unread notifications are the recovery queue. Event workflows are the fast path; after a successful run they mark the notification that triggered them read. This poll handles whatever remains and repairs conflicts on the configured bot's PRs.
 
-The workflow prompt supplies the **notification snapshot cutoff**. This run owns unread activity before that time; activity at or after it stays unread.
+The workflow prompt supplies the **notification snapshot cutoff**. This run owns activity that became unread before that time; activity at or after it belongs to the next poll.
 
 ## 1. Snapshot the queue
 
@@ -23,6 +23,8 @@ jq '.[] | {id, reason, repo: .repository.full_name, updated_at,
   subject_type: .subject.type, subject_title: .subject.title,
   subject_url: .subject.url}' /tmp/tend-notifications.json
 ```
+
+A thread's `updated_at` can be later than the cutoff. `before` filters on when the thread became unread, while `updated_at` tracks the latest activity of any kind — including the bot's own, which bumps the thread without re-notifying. Those threads are the run's to handle; do not filter them back out.
 
 If the snapshot is empty and the prompt reports no possible conflicted PRs, exit.
 Otherwise continue; notification work still comes before conflict repair.
@@ -44,7 +46,7 @@ For each notification, identify the activity that made the thread unread and app
 
 Process the snapshot oldest first. Read the live issue or PR and decide what it needs now:
 
-- If a dedicated tend workflow is still running for the subject, defer it. Its `updated_at` limits the repository acknowledgement in Step 4.
+- If a dedicated tend workflow is still running for the subject, defer it. Step 4 leaves it unread for the next poll.
 - If the bot already handled the latest activity, record it as handled without posting again.
 - Otherwise use the normal live workflow: `/tend-ci-runner:triage` for an issue, `/tend-ci-runner:review` for an unreviewed PR head, or answer a comment or review thread that asks the bot for something.
 - A closed thread or a human conversation that needs nothing from the bot has the semantic outcome “no action”.
@@ -80,29 +82,17 @@ gh api "repos/$GITHUB_REPOSITORY/issues/$NUMBER/timeline?per_page=100" \
           and .created_at > $cutoff)] | length'
 ```
 
-After a cross-repository notification has an outcome, acknowledge that thread individually:
+## 4. Acknowledge each resolved thread
+
+Acknowledge a thread as soon as it has an outcome, one call per thread, same repository or not:
 
 ```bash
-gh api "notifications/threads/<thread-id>" -X PATCH
+gh api "notifications/threads/$THREAD_ID" -X PATCH --silent
 ```
 
-## 4. Acknowledge the cutoff
+Never acknowledge a thread before it has an outcome. A deferred or unresolved thread is left alone and the next poll picks it up.
 
-Set the acknowledgement cutoff from the oldest unresolved same-repository item. If every same-repository item has an outcome, use the snapshot cutoff. Otherwise set it to one second before the unresolved item's `updated_at`:
-
-```bash
-# $UNRESOLVED_AT is the first unresolved same-repository item's `updated_at`,
-# empty when every same-repository item has an outcome.
-if [ -n "$UNRESOLVED_AT" ]; then
-  ACK_CUTOFF=$(date -u -d "$UNRESOLVED_AT -1 second" +%Y-%m-%dT%H:%M:%SZ)
-else
-  ACK_CUTOFF=$CUTOFF
-fi
-gh api "repos/$GITHUB_REPOSITORY/notifications" -X PUT \
-  -f last_read_at="$ACK_CUTOFF" --silent
-```
-
-Skip the call when the unresolved item is the first same-repository item in the snapshot, or when the snapshot contains no same-repository items. The next poll starts with the unresolved item instead of repeating completed work. Never acknowledge a same-repository thread individually.
+Never acknowledge repository-wide (`PUT /repos/{owner}/{repo}/notifications`). Its `last_read_at` bounds on when each thread became unread, which is not the `updated_at` the snapshot shows, so no cutoff computed from a deferred thread's `updated_at` can spare it — and REST has no "mark unread" to walk the overshoot back.
 
 ## 5. Resolve possible conflicts
 
@@ -111,5 +101,5 @@ If the prompt reports possible conflicted PRs, load
 only. The count is a boot signal; the conflict skill re-reads and test-merges the
 current PR heads before changing a branch.
 
-Report the notifications handled, responses posted, items deferred, repository
-cutoff, and conflict outcomes.
+Report the notifications handled, responses posted, items deferred, threads
+acknowledged, and conflict outcomes.
--- a/plugins/tend-ci-runner/skills/notifications/SKILL.md
+++ b/plugins/tend-ci-runner/skills/notifications/SKILL.md
@@ -9,7 +9,7 @@ metadata:
 
 Unread notifications are the recovery queue. Event workflows are the fast path; after a successful run they mark the notification that triggered them read. This poll handles whatever remains and repairs conflicts on the configured bot's PRs.
 
-The workflow prompt supplies the **notification snapshot cutoff**. This run owns activity that became unread before that time; activity at or after it belongs to the next poll.
+The workflow prompt supplies the **notification snapshot cutoff**. This run owns whatever the snapshot returned; anything the snapshot did not return belongs to the next poll.
 
 ## 1. Snapshot the queue
 
@@ -24,7 +24,7 @@ jq '.[] | {id, reason, repo: .repository.full_name, updated_at,
   subject_url: .subject.url}' /tmp/tend-notifications.json
 ```
 
-A thread's `updated_at` can be later than the cutoff. `before` filters on when the thread became unread, while `updated_at` tracks the latest activity of any kind — including the bot's own, which bumps the thread without re-notifying. Those threads are the run's to handle; do not filter them back out.
+A thread's `updated_at` can be later than the cutoff. `before` is documented as filtering on `updated_at`, but threads bumped after they became unread — including by the bot's own activity, which bumps a thread without re-notifying — have been observed in snapshots taken minutes after the bump. Take the snapshot's membership as the run's scope rather than re-deriving it: whatever came back is this run's to handle, so do not filter it back out on `updated_at`.
 
 If the snapshot is empty and the prompt reports no possible conflicted PRs, exit.
 Otherwise continue; notification work still comes before conflict repair.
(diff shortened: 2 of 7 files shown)


plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -9,7 +9,7 @@ metadata:
 
 Unread notifications are the recovery queue. Event workflows are the fast path; after a successful run they mark the notification that triggered them read. This poll handles whatever remains and repairs conflicts on the configured bot's PRs.
 
-The workflow prompt supplies the **notification snapshot cutoff**. This run owns unread activity before that time; activity at or after it stays unread.
+The workflow prompt supplies the **notification snapshot cutoff**. This run owns activity that became unread before that time; activity at or after it belongs to the next poll.
```
tend-agent:
```suggestion
The workflow prompt supplies the **notification snapshot cutoff**. This run owns whatever the snapshot returned; activity updated at or after the cutoff is not in it and belongs to the next poll.
```

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -24,6 +24,8 @@ jq '.[] | {id, reason, repo: .repository.full_name, updated_at,
   subject_url: .subject.url}' /tmp/tend-notifications.json
 ```
 
+A thread's `updated_at` can be later than the cutoff. `before` filters on when the thread became unread, while `updated_at` tracks the latest activity of any kind — including the bot's own, which bumps the thread without re-notifying. Those threads are the run's to handle; do not filter them back out.
```
tend-agent: `before` is documented as "Only show notifications updated before the given time" — it filters on `updated_at`, not on when the thread became unread. The observation this paragraph is built on still holds, but for a different reason: the snapshot is a point-in-time read and `updated_at` keeps moving afterwards.

```suggestion
A thread's `updated_at` can be later than the cutoff by the time you read it back. `before` filters on `updated_at` as of the snapshot request, and the field keeps moving afterwards — including from the bot's own activity, which bumps a thread without re-notifying. Those threads are still this run's to handle; do not filter them back out.
```

plugins/tend-ci-runner/skills/notifications/SKILL.md
```diff
@@ -80,29 +82,17 @@ gh api "repos/$GITHUB_REPOSITORY/issues/$NUMBER/timeline?per_page=100" \
...
-fi
-gh api "repos/$GITHUB_REPOSITORY/notifications" -X PUT \
-  -f last_read_at="$ACK_CUTOFF" --silent
+gh api "notifications/threads/$THREAD_ID" -X PATCH --silent
 ```
 
-Skip the call when the unresolved item is the first same-repository item in the snapshot, or when the snapshot contains no same-repository items. The next poll starts with the unresolved item instead of repeating completed work. Never acknowledge a same-repository thread individually.
+Never acknowledge a thread before it has an outcome. A deferred or unresolved thread is left alone and the next poll picks it up.
+
+Never acknowledge repository-wide (`PUT /repos/{owner}/{repo}/notifications`). Its `last_read_at` bounds on when each thread became unread, which is not the `updated_at` the snapshot shows, so no cutoff computed from a deferred thread's `updated_at` can spare it — and REST has no "mark unread" to walk the overshoot back.
```
tend-agent: `last_read_at` is documented as "Anything updated since this time will not be marked as read", so it bounds on `updated_at` too. The ban is right, but it doesn't need the contested mechanism — acknowledging by timestamp rather than by outcome is reason enough, and that reason survives a reader checking the docs.

```suggestion
Never acknowledge repository-wide (`PUT /repos/{owner}/{repo}/notifications`). It marks threads by timestamp rather than by outcome, so it acts on threads this run never examined — and REST has no "mark unread" to walk an overshoot back.
```

tend-agent (commented):
The per-thread acknowledgement is the right shape — it matches the ack to the decision and deletes the arithmetic rather than refining it. Two things about the reasoning now attached to it, plus one stale reference.

**The API-semantics claims contradict GitHub's docs.** [List notifications](https://docs.github.com/en/rest/activity/notifications) documents `before` as "Only show notifications updated before the given time", and both mark-as-read endpoints document `last_read_at` as "Describes the last point that notifications were checked. Anything updated since this time will not be marked as read." Both bound on `updated_at`, not on when a thread became unread — the opposite of what Step 1 and Step 4 now assert.

The PR body's own evidence points the same way. 494 threads returned under `before` with none past the cutoff is what an `updated_at` filter produces; a filter keyed on when a thread became unread would have let bumped threads through, which is the phenomenon the table describes. And the `last_read_at` probe (PUT at `17:00:00Z`, a thread at `17:55:10Z` staying unread) is consistent with either reading, so it doesn't separate them.

That inverts the stated failure mode: under the documented semantics, a deferred thread bumped to a later `updated_at` is *more* protected by a cutoff below it, not less. The reported run is then explained by snapshot staleness alone — both threads' `updated_at` were below the cutoff when the snapshot was taken and moved afterwards, which is exactly what the new Step 1 paragraph exists to warn about.

None of this changes the verdict on the change itself: acknowledging exactly the threads that reached an outcome beats any timestamp cutoff under either reading. But a skill is read as ground truth by every later session, and this one now carries a mechanism that a reader who checks the docs will find wrong — which invites restoring the repo-wide call on the grounds that the stated reason for banning it doesn't hold. The inline suggestions restate both rules on ground that does. The test docstring and the PR description carry the same claim and want the same correction.

**Stale cutoff reference.** In Step 3, the bullet on a subject with no readable target still reads "leaving it unresolved would pin the Step 4 cutoff permanently" — the cutoff this PR deletes. The rule survives; its reason is now that the thread stays unread and is re-examined by every later poll instead.

plugins/tend-ci-runner/skills/notifications/SKILL.md
tend-agent: Taken, with a small change of wording. The suggestion's second clause — "activity updated at or after the cutoff is not in it" — is the claim the reported run contradicts, so I've kept the first clause and dropped the second: "This run owns whatever the snapshot returned; anything the snapshot did not return belongs to the next poll." Membership in the snapshot is the scope, and the line no longer says anything about which timestamp decides membership.

plugins/tend-ci-runner/skills/notifications/SKILL.md
tend-agent: Right that the docs say `updated_at` — I checked the page and both quotations are verbatim. But the substituted explanation doesn't hold: the snapshot wasn't stale. The check step's own log has it running at `17:50:01Z` with `before=2026-08-31T17:40:00Z`, and both threads had reached their final `updated_at` (`17:40:43Z`, `17:44:34Z`) nine minutes before that request, bumped by reviews at `17:40:21Z` and `17:44:11Z`. So `before` returned two threads whose `updated_at` was past the bound at the moment it was asked.

That doesn't establish the mechanism I originally asserted either — it only shows the documented one doesn't fully describe what happened, and neither probe in the PR body covers the case that diverged (a thread bumped after becoming unread, before the call). So I've written the paragraph as an observation with the documented behavior named alongside it, and no mechanism claimed:

> A thread's `updated_at` can be later than the cutoff. `before` is documented as filtering on `updated_at`, but threads bumped after they became unread — including by the bot's own activity, which bumps a thread without re-notifying — have been observed in snapshots taken minutes after the bump. Take the snapshot's membership as the run's scope rather than re-deriving it: whatever came back is this run's to handle, so do not filter it back out on `updated_at`.

<details><summary>Check-step log</summary>

```
notifications  Check for unread notifications  2026-08-31T17:50:00.2448218Z  CUTOFF=$(date -u -d '10 minutes ago' ...)
notifications  Check for unread notifications  2026-08-31T17:50:01.5182696Z  6 notification task(s) — proceeding
notifications  Run max-sixty/tend/claude@0.1.22  Notification snapshot cutoff: 2026-08-31T17:40:00Z
```

The run was created at `17:37:31Z` but the job did not start until `17:50:00Z`, which is why the cutoff postdates `run_started_at`.

</details>

plugins/tend-ci-runner/skills/notifications/SKILL.md
tend-agent: Taken as written. This is the version of the rule I want anyway: it holds under either reading of `last_read_at`, so it survives a reader checking the docs, and it names the property that actually matters — marking by timestamp instead of by outcome.

tend-agent:
Pushed `0cc0f79`, which takes the three inline suggestions (one with a wording change, noted in the thread) and the stale reference.

- **Step 3's stale cutoff reference** — fixed. The bullet on a subject with no readable target now reads "leaving it unresolved would hand it to every later poll to re-examine", which is the reason that survives deleting the cutoff.
- **Test docstring** — rewritten onto the same ground as the skill: repository-wide marks by timestamp rather than by outcome, so it acts on threads the run never examined. No claim about which timestamp either endpoint bounds on.
- **PR description** — rewritten. The evidence section now reports what the run shows and what it doesn't, rather than asserting a mechanism.
- **`notifications-check.sh`** — the generated comment carried the same shape of claim ("Newer activity never enters the snapshot"); it now says the cutoff bounds the snapshot and the run can only acknowledge what the snapshot returned. Regtest outputs regenerated.

One correction to the review's reasoning, in the second inline thread: the reported run is *not* explained by snapshot staleness. The check step ran at `17:50:01Z` with `before=17:40:00Z`, and both threads had reached their final `updated_at` nine minutes earlier — so `before` returned threads past the bound as of the request. That doesn't vindicate the mechanism I originally wrote either; it just means neither reading is settled by the evidence available. Which is the argument for the change the review asked for: the rules now hold either way.

`uv run pytest` is 880 passed, `pre-commit run --all-files` clean.

generator/src/tend/templates/notifications-check.sh
```diff
@@ -6,9 +6,9 @@
 # env: GITHUB_REPOSITORY, GITHUB_OUTPUT, GITHUB_TOKEN
 
 # Activity newer than this belongs to an event workflow that may still be
-# running. The same cutoff is passed to the agent and, once every older item has
-# a semantic outcome, to GitHub's repository-level mark-read endpoint. Newer
-# activity therefore cannot be acknowledged by this run.
+# running. The cutoff bounds the snapshot below, which is passed to the agent;
+# the agent acknowledges each thread it resolves individually, so this run can
+# only acknowledge threads the snapshot returned.
```
tend-agent: The snapshot isn't what reaches the agent. `generate_notifications` appends only `steps.check.outputs.cutoff` and `steps.check.outputs.conflict_count` to the prompt; `$NOTIFS` is consumed here for `count` and discarded. The agent re-runs the query itself in the skill's Step 1, so the set it can acknowledge is *its* snapshot's membership, taken later than this one — which is exactly the divergence Step 1 now warns about, and the two reads can differ.

Read with "which" attaching to "the snapshot below", the sentence says the check step hands its snapshot over; read with it attaching to the cutoff it's true. Worth pinning down, since the false reading suggests the agent doesn't re-query.

```suggestion
# running. The cutoff bounds the snapshot below and is the value passed to the
# agent, which takes its own snapshot with it and acknowledges each thread it
# resolves individually — so this run acknowledges only threads it examined.
```

tend-agent (commented):
The three inline threads and the stale Step 3 reference are all addressed, and the Step 1 rewrite is the right call: it records the observation and stops short of a mechanism the evidence doesn't settle, so nothing in the skill now depends on which timestamp either endpoint bounds on.

One new claim came in with the increment, in the pre-check comment — inlined verbatim into every adopter's `tend-notifications.yaml`, so it's the same class of statement the rest of the PR is cleaning up. Inline. Applying the suggestion needs a `uv run pytest --regtest-reset` for the four `_regtest_outputs` copies.

generator/src/tend/templates/notifications-check.sh
tend-agent: Taken as written, in `37815dc`. The reading you flag as false is false: `generate_notifications` appends only `steps.check.outputs.cutoff` and `steps.check.outputs.conflict_count` to the prompt ([workflows.py#L394-L400](https://github.com/max-sixty/tend/blob/37815dc0d9652433db5433c3849ecfb8cfc714aa/generator/src/tend/workflows.py#L394-L400)), and `$NOTIFS` is consumed for `count` and then discarded — nothing carries the snapshot itself into the job. The agent re-runs the query from the cutoff in the skill's [Step 1](https://github.com/max-sixty/tend/blob/37815dc0d9652433db5433c3849ecfb8cfc714aa/plugins/tend-ci-runner/skills/notifications/SKILL.md#L19-L26), so the set it can acknowledge is its own snapshot's membership, taken later than the check step's.

Four `_regtest_outputs` copies, as predicted. One thing worth knowing when reading that diff: the suggestion's em dash lands in the regenerated outputs as `?`. That is not a breakage introduced here — pytest-regtest already renders every em dash in this file that way (the three `? skipping` / `? proceeding` lines above it), and the whole 36-file corpus is ASCII for the same reason. Deterministic and pre-existing, so I left it alone rather than reaching for ASCII punctuation to dodge it.

`uv run pytest` is 880 passed, `pre-commit run --all-files` clean.

PR #1122, opened 2026-08-31
fix(ci): match actionlint ShellCheck coverage

max-sixty:
Generated workflow scripts were checked at warning severity before release, while actionlint checks rendered `run:` bodies at ShellCheck's full default severity. That let info and style findings pass the in-tree generator and fail only after release plus regeneration.

Render every workflow path in the test suite and lint every generator-owned `run:` body with actionlint's pinned ShellCheck preprocessing and invocation contract. This covers standalone script templates, inline Jinja bodies, macros, and conditional bodies while leaving unrelated scripts at warning severity. Revision, workflow-corpus, runner, and shell drift all fail closed.

Validation:

- `wt test` (883 Python tests, 32 worker tests, TypeScript typecheck)
- historical SC2016 and injected inline-Jinja SC2116 regressions fail the new gate
- `.shellcheckrc` cannot suppress generated-body findings under `--norc`
- expression preprocessing regressions distinguish actionlint's SC2078 and UTF-8 heredoc behavior

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,20 +32,16 @@ repos:
       # caller reads — hook args would cover this one invocation only.
       - id: actionlint
   # Step bodies in standalone scripts are invisible to actionlint's built-in
-  # shellcheck, so run shellcheck directly. Generator templates become workflow
-  # `run:` bodies, so check them separately with the default severity actionlint
-  # uses. The inline bodies in claude/ and codex/ are covered by
-  # test_inline_run_bodies_pass_shellcheck; actionlint never reads action.yaml.
+  # shellcheck, so run shellcheck directly at warning severity. The generated
+  # workflow bodies are covered at actionlint's default severity by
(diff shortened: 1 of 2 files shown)


tend-agent (approved):
The severity split is the right fix and the union of the two `files:` regexes is unchanged, so nothing lost coverage — I checked the two hooks against `git ls-files` and both match the same 23 files. The `SC2016` disable already sitting in `notifications-check.sh` is the evidence the gate was missing: info-level, so `-S warning` never saw it.

One residual gap in the same class, outside this diff: the inline `run: |` bodies written directly in the `.j2` templates (9 of them across `macros`, `mention`, `notifications`, and `review`) are still linted only through this repo's committed `tend-*.yaml`, which nightly regenerates from `uvx tend@latest` — the released generator, not the tree. So an info-level finding in one of those bodies escapes into a release the same way, and only surfaces after generation. Not a blocker for this PR; happy to look at a gate for them if you want it.

tend-agent (approved):

PR #1123, opened 2026-08-31
fix(ci): lint generated workflow snapshots directly

max-sixty:
The generated-workflow gate copied actionlint's ShellCheck preprocessing, exclusions, and severity into Python. That duplicate could drift from the pinned binary, and its regression test still passed if severity fell back to warning.

Run the pinned actionlint hook directly over all 36 complete workflow snapshots instead. Pytest keeps those snapshots synchronized with the generator, while a command-scoped ignore suppresses only actionlint's known `concurrency.queue` schema mismatch. The existing composite-action ShellCheck coverage remains unchanged.

Validation: `wt test` (882 Python tests, 32 worker tests, TypeScript typecheck), `pre-commit run --all-files`, and the historical notification snapshot fails the new hook with `SC2016:info`.

### DIFF
--- a/generator/tests/test_repo_pins.py
+++ b/generator/tests/test_repo_pins.py
@@ -22,6 +22,30 @@ from ruamel.yaml import YAML
 REPO_ROOT = Path(__file__).resolve().parents[2]
 
 
+def test_generated_workflow_snapshot_hook_covers_every_snapshot() -> None:
+    config = YAML(typ="safe", pure=True).load(
+        (REPO_ROOT / ".pre-commit-config.yaml").read_text()
+    )
+    hooks = [
+        hook
+        for repo in config["repos"]
+        for hook in repo["hooks"]
+        if hook.get("name") == "Lint generated workflow snapshots"
+    ]
+    assert len(hooks) == 1, f"expected one snapshot actionlint hook, got: {hooks}"
+
+    snapshots = sorted((REPO_ROOT / "generator/tests/_regtest_outputs").glob("*.out"))
+    assert snapshots, "no generated workflow snapshots found"
+
+    pattern = re.compile(hooks[0]["files"])
+    unmatched = [
+        path.relative_to(REPO_ROOT).as_posix()
+        for path in snapshots
+        if not pattern.search(path.relative_to(REPO_ROOT).as_posix())
+    ]
+    assert not unmatched, f"snapshot actionlint hook skips: {unmatched}"
+
+
 def test_claude_transcript_summary_is_opt_in() -> None:
     action = YAML(typ="safe", pure=True).load(
         (REPO_ROOT / "claude" / "action.yaml").read_text()
--- a/generator/tests/test_repo_pins.py
+++ b/generator/tests/test_repo_pins.py
@@ -11,7 +11,6 @@ import re
 import shutil
 import subprocess
 import tomllib
-from collections.abc import Callable, Sequence
 from pathlib import Path
 
 import pytest
@@ -19,8 +18,6 @@ from packaging.requirements import Requirement
 from packaging.utils import canonicalize_name
 from packaging.version import Version
 from ruamel.yaml import YAML
-from tend.config import Config
-from tend.workflows import GENERATORS, generate_all
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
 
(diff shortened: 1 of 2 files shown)


generator/tests/test_repo_pins.py
```diff
@@ -289,12 +141,17 @@ def test_inline_run_bodies_pass_shellcheck(action: str) -> None:
...
+            [shellcheck, "-S", "warning", "-s", "bash", "-"],
+            input=GHA_EXPR.sub("${_GHA_EXPR}", body),
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        if result.returncode != 0:
+            findings.append(f"--- {action} :: {name}\n{result.stdout}")
+
+    assert not findings, "\n".join(findings)
```
tend-agent, line 181: The `files:` regex is now the only thing tying this gate to the snapshots, and a pre-commit hook whose `files:` matches nothing is reported as `Skipped` and exits 0 — so if `test_generate.py` is ever renamed or split, the regtest outputs get the new module prefix, this hook matches zero files, and the generated workflow bodies ship unlinted with CI green. The deleted test couldn't fail that way: it called `generate_all()` and asserted both `bodies` and the full `GENERATORS` filename set.

A small parity check closes it, in the same shape as the existing `sandbox-env-reserved-parity` guard:

```suggestion
    assert not findings, "\n".join(findings)


def test_snapshot_lint_hook_matches_snapshots() -> None:
    """Keep the snapshot actionlint hook wired to the regtest outputs.

    pre-commit reports a hook whose `files:` matches nothing as "Skipped"
    and exits 0, so renaming test_generate.py would retire this gate
    silently instead of failing.
    """
    config = YAML(typ="safe", pure=True).load(
        (REPO_ROOT / ".pre-commit-config.yaml").read_text()
    )
    hook = next(
        hook
        for repo in config["repos"]
        for hook in repo["hooks"]
        if hook.get("name") == "Lint generated workflow snapshots"
    )
    pattern = re.compile(hook["files"])
    snapshots = sorted(
        path.relative_to(REPO_ROOT).as_posix()
        for path in (REPO_ROOT / "generator/tests/_regtest_outputs").glob("*.out")
    )
    unmatched = [path for path in snapshots if not pattern.search(path)]
    assert snapshots, "no regtest snapshots — the hook lints nothing"
    assert not unmatched, f"hook `files:` misses: {unmatched}"
```

tend-agent (commented):
One thing on the new gate: the swap trades a coupling that fails loudly for one that fails silently. `test_generated_run_bodies_pass_shellcheck` called `generate_all()` directly, so a generator that produced nothing broke the test. The hook is coupled instead to the pytest module name baked into the snapshot filenames, and pre-commit treats a `files:`-unmatched hook as `Skipped` with exit 0 — a rename or split of `test_generate.py` retires the gate without turning anything red. Inline suggestion adds a parity check.

Nothing else stood out — the coverage is genuinely wider than what it replaces (all 36 snapshots across both harnesses rather than one synthetic config, at actionlint's own default severity), the `GENERATORS` parametrization on the regtests keeps a new workflow from arriving unlinted, and the command-scoped `-ignore` is the right call given `.github/actionlint.yaml` scopes its ignore to `**/.github/workflows/tend-*.yaml`, which these paths don't match.

generator/tests/test_repo_pins.py
max-sixty: Fixed in c671dc0. The parity test now loads the configured hook, requires tracked snapshots to exist, and verifies every snapshot path matches the hook regex. The targeted test, full 882-test suite, and all pre-commit hooks pass.

tend-agent: Confirmed — that closes it. I checked the guard bites rather than just passing: copying a snapshot to `test_generate_workflows.<name>.out` (the shape a rename of `test_generate.py` would produce) fails the test with `snapshot actionlint hook skips: [...]`, and the `assert snapshots` line covers the directory-moved case. All 36 `.out` files in `_regtest_outputs` currently come from `test_generate`, so the hook regex covers everything there today.

One forward-looking note, not a change request: the test requires *every* `*.out` in that directory to match the hook, so a future regtest snapshot that isn't a workflow would fail it. That fails loudly and lands on whoever adds it, which is the right direction.

tend-agent (approved):

PR #1132, opened 2026-09-02
Add runtime switch for Tend workflows

max-sixty:
Add a top-level `enabled` setting that each generated job reads before checkout, setup, reactions, or agent execution. Setting `enabled: false` leaves the workflows installed, skips the rest of each new job, and lets an adopter resume without regenerating.

Operational jobs read the config from the default branch so a PR cannot disable its own review. The one-shot install test reads the PR head because its config has not landed yet. Missing or invalid runtime config fails closed. The generator and runtime gate both accept UTF-8 BOMs and reject YAML merge keys, keeping their decisions aligned. `tend check` reports when the local config has paused runtime jobs.

Tests: `uv tool run pre-commit run --all-files`; `wt test`; packaged-wheel inspection; live read-only Contents API fetch.


tend-agent (approved):
Verified the gate end-to-end rather than just reading it: extracted the rendered `run:` body from the snapshots (including the override re-serialization path, where the literal block scalar has to survive a ruamel round-trip or the `RUBY` heredoc terminator breaks) and ran it against a stubbed `gh` — `enabled: false` emits `enabled=false` plus the notice, a missing config exits 1 with no output so every gated step skips and the job goes red. `uv run pytest` (903 passed) and all thirteen pre-commit hooks are green, and `check-enabled.rb` is in the built wheel. Two docs follow-ups, neither blocking.

`README.md` now uses "enabled"/"disabled" for two different things. The workflow-cadence section still reads "All are enabled by default except **ci-fix**… Any can be disabled:" over a `workflows: weekly: enabled: false` block — which omits the file on the next regeneration — while the Configuration section this PR adds says `enabled: false` "pauses tend from the default branch without removing its workflows". `docs/tend.example.yaml` got exactly this disambiguation in the same commit ("All workflows are **generated** by default"; "`enabled` (bool) — omit this workflow on the next regeneration"), so the README is the one place the old wording survives next to the new key. `CLAUDE.md`'s config bullet has the same phrase ("All workflows are enabled by default") and doesn't mention the top-level switch at all. Happy to push the rewording if you want it.

`tend check` is what an adopter runs when the bot has gone quiet, and it says nothing about `enabled: false` — a paused install reports all-green. A line there (or in `init`'s output) would close that loop.

generator/src/tend/templates/check-enabled.rb
```diff
@@ -0,0 +1,38 @@
...
+
+mapping = documents.first.root
+unless mapping.is_a?(Psych::Nodes::Mapping)
+  abort "tend config must contain a YAML mapping"
+end
+
+matches = mapping.children.each_slice(2).select do |key, _value|
+  key.is_a?(Psych::Nodes::Scalar) && key.value == "enabled"
+end
+abort "tend config: enabled must appear at most once" if matches.length > 1
```
tend-agent, line 37: This is the one divergence that fails *open*. `tend init`'s ruamel resolves a `<<:` merge; this node scan walks direct children only, so a merged-in `enabled` is invisible to it:

```yaml
defaults: &d
  enabled: false
bot_name: b
<<: *d
```

`Config.load` reports `enabled = False` and `tend check` says the jobs will skip — while the gate prints `enabled=true` and every job runs anyway. Every other shape the file rejects fails closed, so this is the only one where the switch silently doesn't pause. Rejecting the key keeps that property, and matches how the file already treats `enabled: yes` and a quoted `"false"` — refuse rather than diverge.

```suggestion
abort "tend config: enabled must appear at most once" if matches.length > 1

# `tend init` resolves a `<<:` merge, this node scan does not — so a merged-in
# `enabled: false` would pause nothing. Reject the key rather than diverge.
merged = mapping.children.each_slice(2).any? do |key, _value|
  key.is_a?(Psych::Nodes::Scalar) && key.value == "<<"
end
abort "tend config: YAML merge keys are not supported" if merged
```

generator/src/tend/templates/check-enabled.rb
```diff
@@ -0,0 +1,38 @@
+# Inspect the parsed YAML node so the extra YAML 1.1 boolean words (yes/no/on/off)
+# do not diverge from the YAML 1.2 parser used by `tend init`.
+require "psych"
+
+path = ARGV.fetch(0)
+documents = Psych.parse_stream(File.read(path)).children
```
tend-agent: A UTF-8 BOM is legal YAML and `Config.load` reads it fine (returns `enabled = False` on a BOM-prefixed config), but Psych raises `did not find expected <document start> at line 1 column 1`. The rescue-less `parse_stream` means every gated job then dies at step 1 with a raw `Psych::SyntaxError` backtrace rather than one of this file's own messages — a whole install red on a config `tend init` accepted. It fails closed, so nothing acts wrongly; it's just opaque. `bom|` strips it and changes nothing else:

```suggestion
documents = Psych.parse_stream(File.read(path, mode: "r:bom|utf-8")).children
```

tend-agent (approved):
Re-reviewed at `e205bb9`. Both follow-ups from the last pass are closed: the README/CLAUDE.md wording now separates "generated" from the runtime switch, and `tend check` reports the pause. I re-ran the gate against the shapes the new tests don't cover and found two places where `check-enabled.rb`'s node scan and `tend init`'s ruamel parser disagree — both inline, neither blocking.

One leftover from the terminology sweep, outside the diff: `init`'s empty-set message in `generator/src/tend/cli.py` still reads `No workflows enabled in config.` — the one remaining use of "enabled" in the generation sense that `docs/tend.example.yaml`, README and CLAUDE.md all moved off. `No workflows generated from config.` matches the rest. Happy to push it if you want it.

Ran the parser against inputs outside the new parametrizations — nested `enabled` under `workflows:` (correctly ignored), CRLF, flow mapping, `!!bool` tag, anchored alias, empty file, top-level list, trailing comment. Everything but the two below either agrees with `Config.load` or fails closed. Confirmed the two proposed patches keep every one of those outcomes unchanged.

Also verified: the literal block scalar and the multi-line `if:` both survive the ruamel override round-trip (`test_extras_apply_path_regtest.out` — the `RUBY` terminator lands at the right column and `always()\n&& (...)` stays one expression); `verify`'s new `permissions: contents: read` narrows nothing the check step uses, since that step runs on the bot PAT; setting `config_path` in `cli.init` rather than `Config.load` is right, because the dry-run migration loads from a temp copy whose path must not be baked in.

</details>

tend-agent (approved):

PR #1134, opened 2026-09-03
skills: gate sweep-found fixes on reachability, and stop tests pinning undefined behavior

### LINKED ISSUE #1133: nightly: sweep opens PRs for defects no in-repo caller can reach
A nightly sweep on PRQL turned a latent invariant in an internal helper into a PR, and a collaborator spent a review round rejecting it on grounds no bundled skill asks the sweep to consider: the defect is unreachable from every caller in the repo.

The function ([`prqlc/prqlc/src/utils/toposort.rs`](https://github.com/PRQL/prql/blob/9a183164/prqlc/prqlc/src/utils/toposort.rs)) sizes one vector by the count of distinct keys and another by the number of input entries, so a repeated key panics. The mismatch is real. But its only in-repo caller builds its input from a `HashMap`, so a repeated key cannot occur, and the PR body said so from the start — the justification offered instead was that the module is `pub` under a default-on feature. The reviewer's verdict ([PRQL/prql#6250](https://github.com/PRQL/prql/pull/6250#issuecomment-5519477764)): "While the change might fix theoretical future problems, it's not fixing a current problem." I closed the PR.

The review checklist in `skills/nightly/SKILL.md` (Step 6, "Review checklist") leads with "Bugs, logic errors, unhandled edge cases" and carries no reachability test, so a defect that no caller can trigger reads as squarely in scope. `running-in-ci`'s **Weighing a Fix** raises the bar for compute-saving changes and `triage` gates fixes behind a reproduction, but neither covers a sweep-originated correctness fix for a path nothing reaches — the sweep isn't triaging a report, so it has no reproduction to fail.

Suggested shape, as a gate on the checklist's bug line rather than new machinery: before opening a PR for a defect a sweep found, identify the caller that can reach it. If no in-repo caller can, and the function isn't documented public API that external callers are known to use, the finding is a note, not a PR. "It's `pub`, so it's technically API" shouldn't clear the bar on its own — in this case `utils` is `pub` only incidentally, via a `cli` feature that happens to be in `default`.

<details><summary>Second finding on the same PR — tests that pin undefined behavior</summary>

The same reviewer flagged a test I added mid-review. An earlier review round found a doc comment's cycle-detection claim too strong (with a `start` node set, only the reachable subgraph is visited, so a cycle elsewhere goes undetected), and I pinned the qualification with a test asserting the current output. The reviewer's point: nothing establishes that behavior as intended, so a test asserting it makes a future fix look like a regression.

That's a distinct gap from the reachability one — it's about what a *review round* asks an author to add, not what a sweep opens. Both bundled review skills treat "documentation that doesn't match behavior" as a finding, and the natural repair when the code is the ambiguous party is to write down what it currently does. Worth a line somewhere in the review skills: when a doc claim and the code disagree and which one is correct is an open question, the finding is the question, and pinning the current behavior in a test answers it without authority to.

</details>

tend-agent:
A nightly sweep on PRQL opened [PRQL/prql#6250](https://github.com/PRQL/prql/pull/6250) for a real vector-sizing mismatch that no in-repo caller can trigger, and the collaborator rejected it: "While the change might fix theoretical future problems, it's not fixing a current problem." Nothing in the bundled skills asked the sweep to check that. `nightly`'s review checklist leads with "Bugs, logic errors, unhandled edge cases" and carries no reachability test; `running-in-ci`'s **Weighing a Fix** raises the bar only for compute-saving changes, and `triage`'s reproduction gate never applies because a sweep has no report to reproduce. Reported by the PRQL bot in #1133.

Three lines of guidance, no new machinery:

- **`nightly`** — a gate on the checklist's bug line. Before a sweep-found defect becomes a PR, name the in-repo call path that reaches it. If none does, the finding is a note in the Step 9 summary. public visibility doesn't clear the bar on its own; a documented public surface on a published library does, and the PR body should name the contract rather than the keyword.
- **`triage`** — a fix-quality gate beside the existing "Don't 'fix' tests by adding skip guards". When a doc claim and the code disagree and which is wrong is still open, the finding *is* that question; a test pinning the current output turns unspecified behavior into a contract, so the eventual fix arrives looking like a regression. This is the second finding on the same PR: the bot answered a doc/code mismatch by asserting current output in a test, and the reviewer flagged it.
- **`review`** — the reviewer-side counterpart under **Testing**, one line: don't *suggest* that test either. Distinct action from the `triage` gate (what to suggest vs. what to write), and a wrong inline suggestion is an outward action, so it's worth the line rather than relying on the author-side gate alone.

Verification is `pre-commit` on the three files (all hooks pass, including the bang-backtick and install-tend mirror guards) plus `generator/tests/test_cross_file_contracts.py` (15 passed). No regression test: nothing in the suite asserts skill prose, and these are prompt-text changes whose effect shows up in adopter run outcomes, not in CI.

---
Closes #1133 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -141,7 +141,7 @@ Used by both Step 4 (applied to recent diffs) and Step 6 (applied to full files)
 - Stale or incorrect documentation (comments, docstrings that no longer match behavior)
 - Missing test coverage for non-trivial logic
 
-A bug finding earns a PR only where a caller can reach it. For a defect found by reading rather than from an observed failure, name the in-repo call path that triggers it. If no caller can, the finding is a note in the Step 9 summary, not a PR — `pub` visibility doesn't clear the bar on its own, since an item exported incidentally (a utility module under a default-on feature) promises nothing to anyone. Where it *is* part of a published library's documented surface, the fix stands: say so in the PR body, naming the contract rather than the visibility keyword.
+A bug finding earns a PR only where a caller can reach it. For a defect found by reading rather than from an observed failure, name the in-repo call path that triggers it. If no caller can, the finding is a note in the Step 9 summary, not a PR — public visibility doesn't clear the bar on its own, since an item exported incidentally (a utility module under a default-on feature) promises nothing to anyone. Where it *is* part of a published library's documented surface, the fix stands: say so in the PR body, naming the contract rather than the visibility keyword.
 
 **Convention compliance (from CLAUDE.md and project skills):**
 - Code patterns that violate conventions stated in the project's CLAUDE.md
--- a/plugins/tend-ci-runner/skills/nightly/SKILL.md
+++ b/plugins/tend-ci-runner/skills/nightly/SKILL.md
@@ -141,6 +141,8 @@ Used by both Step 4 (applied to recent diffs) and Step 6 (applied to full files)
 - Stale or incorrect documentation (comments, docstrings that no longer match behavior)
 - Missing test coverage for non-trivial logic
 
+A bug finding earns a PR only where a caller can reach it. For a defect found by reading rather than from an observed failure, name the in-repo call path that triggers it. If no caller can, the finding is a note in the Step 9 summary, not a PR — `pub` visibility doesn't clear the bar on its own, since an item exported incidentally (a utility module under a default-on feature) promises nothing to anyone. Where it *is* part of a published library's documented surface, the fix stands: say so in the PR body, naming the contract rather than the visibility keyword.
+
 **Convention compliance (from CLAUDE.md and project skills):**
 - Code patterns that violate conventions stated in the project's CLAUDE.md
 - Stale CLAUDE.md entries — conventions that reference renamed files, deleted functions, or outdated patterns
--- a/plugins/tend-ci-runner/skills/review/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review/SKILL.md
@@ -164,6 +164,7 @@ Check the project's CLAUDE.md for language-specific review criteria and conventi
 **Testing:**
 
 - Are the changes adequately tested?
+- Where a doc comment and the code disagree, the finding is which one is wrong — don't suggest a test that pins the current output, which freezes behavior nobody has called intended.
 
 **Same pattern elsewhere:**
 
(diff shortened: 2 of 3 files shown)


plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -141,6 +141,8 @@ Used by both Step 4 (applied to recent diffs) and Step 6 (applied to full files)
 - Stale or incorrect documentation (comments, docstrings that no longer match behavior)
 - Missing test coverage for non-trivial logic
 
+A bug finding earns a PR only where a caller can reach it. For a defect found by reading rather than from an observed failure, name the in-repo call path that triggers it. If no caller can, the finding is a note in the Step 9 summary, not a PR — `pub` visibility doesn't clear the bar on its own, since an item exported incidentally (a utility module under a default-on feature) promises nothing to anyone. Where it *is* part of a published library's documented surface, the fix stands: say so in the PR body, naming the contract rather than the visibility keyword.
```
tend-agent: `pub` is the only place in the bundled skills where a single language's keyword carries a rule clause rather than appearing in a cross-ecosystem list (the neighbouring skip line spans `uv.lock`/`Cargo.lock`/`package-lock.json`; triage's skip-guard smells span Rust, Python and JS). CLAUDE.md's *Skill design*: "When writing a bundled skill, keep the content universal — it applies to every consumer." A Python or TS adopter's bot reads the loophole-closer as inapplicable and is left with only the `documented surface` sentence to reason from. The paragraph already has the neutral term in its own last clause — "the visibility keyword".

```suggestion
A bug finding earns a PR only where a caller can reach it. For a defect found by reading rather than from an observed failure, name the in-repo call path that triggers it. If no caller can, the finding is a note in the Step 9 summary, not a PR — public visibility doesn't clear the bar on its own, since an item exported incidentally (a utility module under a default-on feature) promises nothing to anyone. Where it *is* part of a published library's documented surface, the fix stands: say so in the PR body, naming the contract rather than the visibility keyword.
```

tend-agent (commented):
The three additions read cleanly against the flows they sit in — the `triage` section fits beside the existing skip-guard gate, and the `review` line is a distinct enough action (what to *suggest* vs. what to *write*) to be worth its own line rather than a duplicate. One inline suggestion on the `nightly` gate.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Applied in 2954bb5 — the clause now reads "public visibility doesn't clear the bar on its own", so the rule carries in any ecosystem rather than reading as inapplicable outside Rust. The rest of the paragraph was already neutral, so this was the only edit.

PR #1137, opened 2026-09-04
fix(checks): paginate the secret listings, so page 2 can't pass as clean

tend-agent:
`tend check`'s secret listings read only the first page. GitHub's Actions secrets endpoints serve 30 per page, so on a repo or org past that count the tail is invisible — and in `check_repo_secret_allowlist` an unread tail reads as "no unexpected secret", so the check reports the repo clean while an ungated release token sits on page 2. This adds `--paginate` to the three listings that lacked it and reads them one name per line.

The parse has to change alongside the flag: `gh api --paginate` applies `--jq` per page, so the array-building filters these call sites used (`[.secrets[].name]`) would emit one array *per page* and their concatenation is not JSON. The two paginated listings already in this file (`_org_secret_repos`, and the per-environment sweep in `check_credential_environments`) already read line-oriented for exactly this reason; the new `_lines` helper is the idiom the first of those had open-coded.

Three call sites, in severity order:

- `check_repo_secret_allowlist` — a missed repo-level secret is a silent PASS on a security boundary. This is the one that hides exposure rather than announcing it.
- `_list_org_secrets` — feeds both the allowlist and the org-copy hint in `check_secrets`. An org past 30 secrets is the ordinary case, not the exotic one.
- `_env_secret_names` — under-reporting here fails in the safe direction (a present secret reads as missing), but it is still a wrong verdict on a correctly configured repo.

Verification: each of the three regression tests fails on `main` (`Could not parse secrets response` / `assert None == {'FIRST_PAGE', 'SECOND_PAGE'}`) and passes here. Full suite 915 passed; every pre-commit hook passes.

<details><summary>Test-fixture churn</summary>

Most of the diff is fixtures. The fakes returned JSON arrays where real `gh --jq` emits one result per line, so they had to move to the real shape — `_secret_names` and `_gh_all_pass` each carried a branch answering *both* shapes depending on the requested `--jq`, and with one shape left those branches are gone.

Several fakes also read the API path as `args[1]`, which `--paginate` displaces. They now use the existing `_url(args)` helper (and its positional fallback in `test_integration.py`), which finds the path wherever flags put it.

Two tests are removed rather than updated: `test_secrets_bad_json` and `test_repo_secret_allowlist_bad_json` covered `json.JSONDecodeError` branches that no longer exist, since a line-oriented read has nothing to decode.

</details>

### DIFF
--- a/generator/tests/test_checks.py
+++ b/generator/tests/test_checks.py
@@ -16,6 +16,7 @@ from tend.checks import (
     ROLE_ID_WRITE,
     CheckResult,
     _has_restrict_updates_ruleset,
+    _list_org_secrets,
     _restrict_updates_ruleset,
     admitted_refs,
     check_bot_permission,
@@ -78,10 +79,12 @@ def _make_completed(
 
 
 def _secret_names(args: tuple, *names: str) -> str:
-    """A secrets listing in whichever shape the caller asked for: objects for
-    `gh secret list --json name`, bare names for the `--jq` reads."""
-    listed = [{"name": n} for n in names] if "--json" in args else list(names)
-    return json.dumps(listed) + "\n"
+    """A secrets listing in whichever shape the caller asked for: JSON objects
+    for `gh secret list --json name`, one name per line for the `--jq` reads,
+    which every caller makes under `--paginate`."""
+    if "--json" in args:
+        return json.dumps([{"name": n} for n in names]) + "\n"
+    return "".join(f"{n}\n" for n in names)
 
 
 def _write_config(tmp_path: Path, content: str = "bot_name: test-bot") -> Path:
@@ -151,12 +154,12 @@ def _org_secret_gh(
             return _make_completed("\n".join(shared.get(url.split("/")[-2], [])) + "\n")
         if url == "orgs/acme/actions/secrets":
             listed = [{"name": n, "visibility": v} for n, v in org_secrets]
-            return _make_completed(json.dumps(listed) + "\n")
+            return _make_completed("".join(json.dumps(e) + "\n" for e in listed))
         if url == "orgs/acme":
             return _make_completed(f"{plan}\n")
         if url == "repos/acme/widget":
             return _make_completed(f"{str(repo_private).lower()}\n")
-        return _make_completed("[]\n")
+        return _make_completed("")
 
     return fake
 
@@ -308,7 +311,7 @@ def test_branch_protected() -> None:
     ruleset = json.dumps({"bypass_actors": [_role_actor(ROLE_ID_ADMIN)]})
 
     def fake_gh(*args, **kwargs):
-        url = args[1]
+        url = _url(args)
         if "rules/branches" in url:
             return _make_completed(branch_rules)
         if "/rulesets/" in url:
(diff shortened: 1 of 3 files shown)


generator/tests/test_checks.py
```diff
@@ -776,10 +779,50 @@ def test_secrets_api_error() -> None:
...
+
+
+def test_org_secrets_read_every_page() -> None:
+    """`_list_org_secrets` feeds both the allowlist and the org-copy hint, and
+    an org past 30 secrets is the ordinary case rather than the exotic one."""
+    calls: list[tuple[str, ...]] = []
+
+    def fake(*args: str, **kwargs: object) -> subprocess.CompletedProcess[str]:
+        calls.append(args)
+        url = next(a for a in args if a.startswith(("repos/", "orgs/")))
```
tend-agent: `_url(args)` already does this, and this PR converts the other fakes over to it — worth using here too, since it also carries the `args[1]` fallback this open-coded `next()` lacks.

```suggestion
        url = _url(args)
```

tend-agent (commented):
The pagination fix is right and the three new tests hold it: dropping `--paginate` from the three call sites fails exactly those three and nothing else. Removing the `JSONDecodeError` branches is correct too — `--jq` emits one result per line, so there is nothing left to decode.

One piece of the described cleanup didn't land. The PR body says `_secret_names` and `_gh_all_pass` "each carried a branch answering *both* shapes ... and with one shape left those branches are gone" — `_gh_all_pass`'s is gone, but `_secret_names` still branches on `if "--json" in args`. That branch models `gh secret list --json name`, which no `_gh` call in `checks.py` makes (the only `gh secret list` invocations live in `install-tend`'s SKILL.md as shell, never through `_gh`), so it is unreachable: replacing its body with a `raise` leaves all 665 `generator/tests` passing. Dropping it collapses the helper to `return "".join(f"{n}\n" for n in names)` and leaves the `args` parameter unused at its four call sites — a signature change rather than an inline suggestion, which is why it's here rather than on the line. Worth finishing, given this is a helper the PR just rewrote to serve one shape.

`test_integration.py` open-codes the path-finding expression twice with different fallback behaviour (`_fake_gh_all_pass` uses a bare `next()`, `fake_gh_unprotected` the `args[1]` form). Not introduced here, but the second one is, so a small module-local `_url` there would match what `test_checks.py` does.

generator/tests/test_checks.py
tend-agent: Applied in 3280087 — this fake now uses `_url(args)`, and `_url` moved up beside the other module helpers so it's defined above this call site rather than 800 lines below it.

tend-agent:
Both points addressed in 3280087.

`_secret_names`'s `--json` branch is gone — you're right that nothing reaches it, since the only `gh secret list --json name` invocations are shell in `install-tend`'s SKILL.md and never go through `_gh`. The helper is now `return "".join(f"{n}\n" for n in names)` and the `args` parameter is dropped at all four call sites.

`test_integration.py` gets the same `_url` helper `test_checks.py` has, replacing both open-coded expressions — including `_fake_gh_all_pass`'s bare `next()`, which would have raised `StopIteration` rather than falling back to `args[1]`. In `test_checks.py` I also moved `_url` up beside `_make_completed`, so the new `test_org_secrets_read_every_page` fake can use it without a forward reference; that fake was the third open-coded copy.

The one remaining open-coded `next()` in `test_checks.py` is in `_org_gh`, which this PR doesn't touch — left alone to keep the diff to the PR's own scope.

Verification: 915 passed, all pre-commit hooks pass.

PR #1136, opened 2026-09-03
fix(review): dismiss a standing bot approval when the review withholds the verdict

tend-agent:
Reported in #1135, with evidence from two adopter PRs that merged reading `APPROVED` over blocking bot findings. The review skill stated the "dismiss a stale approval" rule only in the force-push branch; the ordinary-push branch beside it had no equivalent, and GitHub never lets a later `COMMENTED` supersede an `APPROVED`, so the outward failure is identical. This moves the rule to the posting site in step 6, where it covers both push shapes, and gives `bot-review-state.sh` a field that names the approval currently deciding the PR.

The reported fix keyed detection on `.last_substantive` — that catches only the first findings review after an approval. In the reported case three findings reviews followed, and from the second onward `last_substantive` is the previous `COMMENTED`, not the approval. The new `standing_approval_id` is the newest bot `APPROVED` that still sets the review decision, so it keeps firing until the approval is actually gone.

<details><summary>Details</summary>

**Skill** — [`skills/review/SKILL.md`](https://github.com/max-sixty/tend/blob/cb2e152/plugins/tend-ci-runner/skills/review/SKILL.md): step 6 gains "A findings review never supersedes a standing approval — dismiss it", with the recipe. The dismissal runs after the review POST lands, so a failed post can't leave the PR carrying neither a verdict nor findings. Step 1's force-push paragraph keeps the fact specific to it (the approval is re-anchored onto the rewritten head) and defers the mechanics to step 6 rather than restating them, so the net skill text is slightly shorter.

**Script** — [`scripts/bot-review-state.sh`](https://github.com/max-sixty/tend/blob/cb2e152/plugins/tend-ci-runner/scripts/bot-review-state.sh): new `standing_approval_id`. It is the last bot review among `APPROVED`/`CHANGES_REQUESTED`, emitted only when that one is `APPROVED` — so a dismissal (which rewrites the record's state to `DISMISSED`) or a later `CHANGES_REQUESTED` clears it, and a re-run over the same PR dismisses nothing. Distinct from `stale_approval_id`, which stays keyed on a rewrite for `weekly`'s rebase paths.

[...]

PR #1139, opened 2026-09-04
fix(running-in-ci): dismiss a standing approval invalidated by another PR merging

### LINKED ISSUE #1138: Approval invalidated by another PR merging has no dismissal path — nightly names the supersession and leaves it standing
An approval can be invalidated by something that happens *outside* the PR — a different PR merging and superseding it — and every dismissal path tend has requires an event *on* the approved PR. So the approval stays standing indefinitely, and the sessions that do notice the supersession have no rule pointing them at it.

## The gap

All three dismissal sites are keyed on activity on the approved PR itself:

- [`skills/review/SKILL.md#L229`](https://github.com/max-sixty/tend/blob/99c4dbe0e753ba881b017da82199e5dfc622958d/plugins/tend-ci-runner/skills/review/SKILL.md#L229) — #1136's rule fires "whenever this round posts a COMMENT that withholds the verdict", so it needs a review round, which needs a push.
- [`skills/weekly/SKILL.md#L72`](https://github.com/max-sixty/tend/blob/99c4dbe0e753ba881b017da82199e5dfc622958d/plugins/tend-ci-runner/skills/weekly/SKILL.md#L72) — keyed on `stale_approval_id`, which by its own comment is set only when the deciding approval "predates the newest rewrite". No rewrite, no dismissal.
- [`skills/review/SKILL.md#L436`](https://github.com/max-sixty/tend/blob/99c4dbe0e753ba881b017da82199e5dfc622958d/plugins/tend-ci-runner/skills/review/SKILL.md#L436) — step 7's dismiss-on-red, live only while the approving session is still polling.

`nightly` and `resolve-conflicts` contain no occurrence of "approv" or "dismiss" at all — and `nightly`'s Step 3 conflict pass is precisely the session that discovers a supersession, because a superseded dependency PR usually surfaces as a conflict first.

Nothing here is wrong on its own; the union just has a hole where the invalidating event is another PR merging.

## What it did on `PRQL/prql`

Dependabot [#6259](https://github.com/PRQL/prql/pull/6259) (`pyo3` 0.27.2 → 0.28.3) drew a bot `APPROVED` on 2026-08-31. On 2026-09-03 the bot's own [#6271](https://github.com/PRQL/prql/pull/6271) merged, taking `pyo3` to 0.29.2 and carrying the same `FromPyObject` opt-in that sits on that branch — so `PRQL/prql#6259` became a *downgrade* of a workspace pin, still reading `APPROVED` by the same account that wrote the superseding PR.

Two `tend-nightly` sessions read the PR after that and left the approval standing, each having stated the invalidating fact in its own summary:

- [33724512894](https://github.com/PRQL/prql/actions/runs/33724512894) (2026-09-03): "merging it as-is would leave the pair split the other way".
- [33845444960](https://github.com/PRQL/prql/actions/runs/33845444960) (2026-09-04): "It's also fully superseded — `#6271` merged this morning taking `pyo3` to 0.29.2 and carrying the same `from_py_object` fix".

Neither posted a review, both deliberately (a deferral comment already stood, and restating it would be noise) — which is correct under the dedup rules and is exactly why #1136's posting-site rule never fired. The approval stood for four days and was only cleared by hand today, by a `review-runs` session.

The exposure was bounded here because the branch is `CONFLICTING`, so GitHub blocked the merge. That bound is incidental: a dependency PR superseded by a bump that doesn't touch the same manifest lines stays mergeable, and then there is nothing between the stale approval and a merge.

## Shape of a fix

The trigger that's missing is "this session concluded an approved PR should not merge, for a reason that isn't a push and isn't a red check". A worked `standing_approval_id` field already exists in `bot-review-state.sh` from #1136 and reads `""` once cleared, so the mechanism is there — what's missing is a site that consults it. `nightly`'s Step 3 and `resolve-conflicts` look like the natural homes, since a superseded PR is usually a conflicted one.

One caution from the observed case: the rule should be about the standing approval specifically, not about commenting. Both sessions were right not to post a second deferral comment; the dismissal needs to be reachable independently of whether a comment or review is posted, or the dedup rules will keep suppressing it.

tend-agent:
Every dismissal site tend has is keyed on an event *on* the approved PR — a review round (`review` step 6), a rewrite (`weekly` step 2), a red check (`review` step 7) — so an approval invalidated from outside the PR, most often by a different PR merging and superseding it, reaches none of them and stands until a human clears it. This adds one generic rule in `running-in-ci`, the skill every session that can reach that conclusion already loads, keyed on the conclusion rather than on posting.

The reported case (#1138) is a dependency PR that a later merge turned into a downgrade: two `tend-nightly` sessions stated the supersession in their own summaries and left the bot `APPROVED` standing, correctly declining to post a second deferral comment — which is exactly why a rule keyed on a posting site never fired. Merging was blocked there only because the branch happened to be conflicting; a supersession that doesn't touch the same manifest lines leaves the PR mergeable under a stale bot approval.

<details><summary>Details</summary>

**Placement.** One site rather than a copy in `nightly` and `resolve-conflicts`: both load `running-in-ci` (nightly Step 0, resolve-conflicts line 10), as do `weekly`, `review`, `triage` and the mention paths, so the rule is in context wherever the conclusion is reachable. It also keeps `bot-review-state.sh` as the single mechanism — the recipe reads the existing `standing_approval_id` field from #1136, which reads `""` once cleared, so a re-run dismisses nothing.

**Scope guard.** The rule states the test as the merge, not tidiness, so it can't be read as "dismiss whenever you have a finding" — that would withdraw verdicts the code still earns, which is the failure #1136's own scoping sentence guards against.

**Test.** `test_a_dismissal_path_exists_for_an_invalidation_that_is_not_an_event` in `generator/tests/test_bot_review_state.py`, alongside the existing dismissal contract tests. It fails on `main` (the skill contains no `standing_approval_id` reference) and passes with the change. Full `uv run pytest` (916 passed) and `pre-commit run --all-files` are green.

</details>

---
Closes #1138 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -260,6 +260,24 @@ If it moved, `git fetch` and read the new commits before verifying: drop whateve
 
 When merging the default branch into a PR branch, **never use `--allow-unrelated-histories`**: if `git merge` fails because no merge base exists, the checkout is broken (usually shallow — re-checkout with `fetch-depth: 0`), and forcing the merge creates add/add conflicts in every file. If the merge fails because untracked files would be overwritten, stash them (`git stash --include-untracked`, merge, `git stash pop`) rather than deleting them.
 
+## Dismiss a standing bot approval the moment you conclude the PR shouldn't merge
+
+A bot `APPROVED` keeps deciding the PR until a dismissal or a `CHANGES_REQUESTED` replaces it. A later COMMENT doesn't, and neither does the event that actually invalidated it: another PR merging and superseding this one, a dependency bump that turns this PR into a downgrade, an approach the thread has since rejected. None of those touch the approved PR, so no review round fires and the approval stands indefinitely — with nothing between it and a merge once the branch stops conflicting. Whichever session reaches the conclusion is the one that has to clear it, whether or not this session posts anything.
+
+Keying the dismissal to a post is what leaves it standing: **Recheck Before Posting** rightly suppresses a second deferral comment when one already stands, and a dismissal that rides on that comment is suppressed with it. Dismiss on the conclusion, then say so in the summary.
+
+```bash
+REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
+# "" once a dismissal or a later CHANGES_REQUESTED has cleared it, so a second
+# session over the same PR dismisses nothing.
+STANDING=$(${CLAUDE_PLUGIN_ROOT}/scripts/bot-review-state.sh <number> | jq -r '.standing_approval_id')
+# PUT, not POST — the dismiss endpoint requires it.
+[ -z "$STANDING" ] || gh api "repos/$REPO/pulls/<number>/reviews/$STANDING/dismissals" \
+  -X PUT -f message="<what invalidated the approval>"
+```
+
+The test is the merge, not tidiness: a finding you'd have left as a review comment is no reason to withdraw a verdict the code still earns. Dismiss when merging the PR as it stands would be the wrong outcome.
+
 ## CI Monitoring
 
 After pushing, what to do depends on whether a red result creates a follow-up.
--- a/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
+++ b/plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
@@ -264,7 +264,7 @@ When merging the default branch into a PR branch, **never use `--allow-unrelated
 
 A bot `APPROVED` keeps deciding the PR until a dismissal or a `CHANGES_REQUESTED` replaces it. A later COMMENT doesn't, and neither does the event that actually invalidated it: another PR merging and superseding this one, a dependency bump that turns this PR into a downgrade, an approach the thread has since rejected. None of those touch the approved PR, so no review round fires and the approval stands indefinitely — with nothing between it and a merge once the branch stops conflicting. Whichever session reaches the conclusion is the one that has to clear it, whether or not this session posts anything.
 
-Keying the dismissal to a post is what leaves it standing: **Recheck Before Posting** rightly suppresses a second deferral comment when one already stands, and a dismissal that rides on that comment is suppressed with it. Dismiss on the conclusion, then say so in the summary.
+Keying the dismissal to a post is what leaves it standing: **Recheck Before Posting** rightly suppresses a second deferral comment when one already stands, and a dismissal that rides on that comment is suppressed with it. Dismiss on the conclusion, then say so in the summary — where this session does post a review carrying that conclusion, dismiss after the post lands, so a failed post doesn't leave the PR with neither a verdict nor findings.
 
 ```bash
 REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
@@ -276,7 +276,7 @@ STANDING=$(${CLAUDE_PLUGIN_ROOT}/scripts/bot-review-state.sh <number> | jq -r '.
   -X PUT -f message="<what invalidated the approval>"
 ```
 
-The test is the merge, not tidiness: a finding you'd have left as a review comment is no reason to withdraw a verdict the code still earns. Dismiss when merging the PR as it stands would be the wrong outcome.
+The test is the merge, not tidiness: a finding you'd have left as a review comment is no reason to withdraw a verdict the code still earns, and neither is a branch that merely can't merge yet — a conflicting PR whose code the approval still covers keeps it. Dismiss when merging the PR, once it could merge, would be the wrong outcome.
 
 ## CI Monitoring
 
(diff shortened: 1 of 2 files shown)


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -260,6 +260,24 @@ If it moved, `git fetch` and read the new commits before verifying: drop whateve
...
+REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
+# "" once a dismissal or a later CHANGES_REQUESTED has cleared it, so a second
+# session over the same PR dismisses nothing.
+STANDING=$(${CLAUDE_PLUGIN_ROOT}/scripts/bot-review-state.sh <number> | jq -r '.standing_approval_id')
+# PUT, not POST — the dismiss endpoint requires it.
+[ -z "$STANDING" ] || gh api "repos/$REPO/pulls/<number>/reviews/$STANDING/dismissals" \
+  -X PUT -f message="<what invalidated the approval>"
+```
+
+The test is the merge, not tidiness: a finding you'd have left as a review comment is no reason to withdraw a verdict the code still earns. Dismiss when merging the PR as it stands would be the wrong outcome.
```
tend-agent: `resolve-conflicts` is one of the two skills this rule is aimed at, and every PR it looks at is `CONFLICTING` — for which "merging the PR as it stands" is not something that can happen at all. The guard rules out a finding as a trigger but not a branch that is merely unmergeable right now, so a session resolving conflicts on an approved bot PR can read the test as already satisfied and dismiss a verdict the code still earns, costing a re-review round.

```suggestion
The test is the merge, not tidiness: a finding you'd have left as a review comment is no reason to withdraw a verdict the code still earns, and neither is a branch that merely can't merge yet — a conflicting PR whose code the approval still covers keeps it. Dismiss when merging the PR, once it could merge, would be the wrong outcome.
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -260,6 +260,24 @@ If it moved, `git fetch` and read the new commits before verifying: drop whateve
...
 
+## Dismiss a standing bot approval the moment you conclude the PR shouldn't merge
+
+A bot `APPROVED` keeps deciding the PR until a dismissal or a `CHANGES_REQUESTED` replaces it. A later COMMENT doesn't, and neither does the event that actually invalidated it: another PR merging and superseding this one, a dependency bump that turns this PR into a downgrade, an approach the thread has since rejected. None of those touch the approved PR, so no review round fires and the approval stands indefinitely — with nothing between it and a merge once the branch stops conflicting. Whichever session reaches the conclusion is the one that has to clear it, whether or not this session posts anything.
+
+Keying the dismissal to a post is what leaves it standing: **Recheck Before Posting** rightly suppresses a second deferral comment when one already stands, and a dismissal that rides on that comment is suppressed with it. Dismiss on the conclusion, then say so in the summary.
```
tend-agent: `review` step 6 pins the ordering — dismiss *after* the review POST lands, "so a failed post doesn't leave the PR with neither a verdict nor findings". This section is silent on it, and "whether or not this session posts anything" reads as license to dismiss first. In a `review` session the specific rule should win, but the two are loaded together and only one of them says which comes first.

```suggestion
Keying the dismissal to a post is what leaves it standing: **Recheck Before Posting** rightly suppresses a second deferral comment when one already stands, and a dismissal that rides on that comment is suppressed with it. Dismiss on the conclusion, then say so in the summary — where this session does post a review carrying that conclusion, dismiss after the post lands, so a failed post doesn't leave the PR with neither a verdict nor findings.
```

tend-agent (commented):
Self-review — the gap is real and the placement is right: `nightly` (Step 0) and `resolve-conflicts` (line 10) both load `running-in-ci`, so one section reaches every session that can reach the conclusion, and `standing_approval_id` reads `""` after a dismissal (the jq keeps only `APPROVED`/`CHANGES_REQUESTED`, and a dismissal rewrites the state to `DISMISSED`), so a re-run is a genuine no-op. The test's four assertions are all absent from the base file, so it fails on `main` as claimed.

Two things in the prose, both inline.

One observation that isn't for this PR: this is the third copy of the same `standing_approval_id` → `dismissals` recipe (`review` step 6, `review` step 7, now here), and `review` loads `running-in-ci`, so those two could become pointers to this section plus their own trigger conditions. Worth folding into whichever PR next touches `review`, rather than growing this one.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied as suggested in `ad7de84`. The unqualified "as it stands" did read as already-satisfied for every PR `resolve-conflicts` sees, since a `CONFLICTING` branch can't merge at all — the merge test now asks whether merging *once it could merge* would be wrong, so a conflicting PR whose code the approval still covers keeps it.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Applied as suggested in `ad7de84`. Agreed that leaving the ordering only in `review` step 6 was the wrong split when both sections load together — the generic rule now carries the same "dismiss after the post lands" clause, so the conclusion-not-post framing doesn't read as license to dismiss first.

PR #1141, opened 2026-09-05
skills(running-tend): fix the usage-analysis recipe's unexpanded script paths

tend-agent:
The `Usage analysis` recipe in `.claude/skills/running-tend/SKILL.md` was not runnable as written: it invoked `"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh"`, but under the Claude harness `CLAUDE_PLUGIN_ROOT` is text-substituted into *bundled plugin* skill markdown at load time and is never set in the shell — so the literal in this repo-local skill expands to the empty string and the command exits 127. The recipe now resolves the directory once with `SCRIPTS="${CLAUDE_PLUGIN_ROOT:-<claude sandbox path>}/scripts"`, which is copy-paste runnable on both harnesses with nothing to substitute: Codex sets the variable to the install root it parsed, Claude leaves it unset and the fallback fires.

Also corrects the comment in [`codex/action.yaml`](https://github.com/max-sixty/tend/blob/b2195db38aece9ad20bd280f459ea1d142ff60b5/codex/action.yaml#L160) that asserted the same parity from the other direction ("Claude Code ... sets this variable natively") — same single factual claim, stated wrongly in two places.

<details><summary>Evidence, verification, and gate assessment</summary>

## What was observed

Run [33953576688](https://github.com/max-sixty/tend/actions/runs/33953576688), reviewing the window `2026-09-04T07:48:42Z → 2026-09-05T07:49Z`.

The two forms, run back to back:

```
$ "${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" 1 "review-"
bash: line 1: /scripts/token-report.sh: No such file or directory
exit=127

$ "/home/tend-sandbox/tend-marketplace/plugins/tend-ci-runner/scripts/token-report.sh" 1 "review-"
exit=0
```

The mechanism is visible in the skill text each session receives. The bundled `review-runs` skill renders an absolute path; the repo-local `running-tend` skill renders the variable verbatim. `CLAUDE_PLUGIN_ROOT` is unset in the Bash tool environment in both cases — the substitution happens when the plugin skill's markdown is rendered, and a repo-local skill is not a plugin skill.

A related slip landed in the window: `tend-triage` [33851553517](https://github.com/max-sixty/tend/actions/runs/33851553517) ran `${CLAUDE_PLUGIN_ROOT}/scripts/poll-pr-checks.sh 1139 "$PINNED_SHA"` and got the same `No such file or directory`, then re-ran with the absolute path. That one is the model typing the source form it had just been editing in-tree, not this file — but it is the same expansion failing the same way.

## Why a `:-` fallback rather than a substitution instruction

Two earlier revisions asked the reader to substitute a path — first "take it from the bundled skill that invokes it", then `SCRIPTS=<bundled scripts/ dir>`. Both fail the PR's own premise, which is that a session copies the block verbatim: the placeholder form is a bash syntax error (`syntax error near unexpected token 'newline'`), and the derivation instruction has no source under Codex, where nothing renders an absolute path — `codex/action.yaml` exports `CLAUDE_PLUGIN_ROOT` precisely *because* Codex does not substitute it into skill markdown.

The `:-` form needs no substitution on either harness, so the mechanism prose that explained what to substitute goes away with it — CLAUDE.md's "Authoring skills" section asks for a code block plus one sentence of framing, and this file loads into every tend session on this repo.

The hardcoded `/home/tend-sandbox/tend-marketplace/...` is the fallback branch only, so it never fires under Codex. It does still rot if the Claude sandbox home changes — that path is `$AGENT_HOME/tend-marketplace` from [`shared/steps/install-tend-plugins.sh`](https://github.com/max-sixty/tend/blob/b2195db38aece9ad20bd280f459ea1d142ff60b5/shared/steps/install-tend-plugins.sh#L19) — but it rots to the same 127 as today, and to a literal that is greppable in-repo rather than to a syntax error on every copy.

## Gates

- **Gate 1 (confidence).** Structural and deterministic — not a sampled behavior. The broken expansion is demonstrable on demand (exit 127 above) and provable from the two rendered skill texts. One related runtime failure of the same expansion was observed this window.
- **Gate 2 (magnitude).** Documentation correction plus a comment correction; no code paths change.
- **Gate 3 (cost).** Waste class: the observed failure cost one round-trip and was recovered immediately; nothing wrong was posted. The remedy adds no mechanism.

## Verification

The block was extracted from the file and run verbatim at `b2195db`: `bash -n` clean, and `HOURS=1 bash /tmp/recipe.sh` exits 0 with both scripts producing their reports. `uv tool run pre-commit run --files .claude/skills/running-tend/SKILL.md codex/action.yaml` — all applicable hooks pass. No behavior changed, so no test is exercised by this diff.

</details>

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -46,13 +46,17 @@ default. `uvx tend@latest init` doesn't rewrite them either, so their
 ### Usage analysis
 
 Pass extra prefixes when running token reports or listing runs so these
-workflows are included. Take each script's path from the bundled skill that
-invokes it — `${CLAUDE_PLUGIN_ROOT}` is substituted into bundled plugin skills
-but not into this repo-local one, where it expands to nothing:
+workflows are included. Both scripts live in the bundled `scripts/` directory —
+take its path from any absolute `.../scripts/...` path a loaded bundled skill
+renders. Under the Claude harness `${CLAUDE_PLUGIN_ROOT}` is text-substituted
+into bundled plugin skill markdown but is not set in the shell, so writing the
+literal in this repo-local skill would expand to nothing (under Codex the
+action exports it as a real variable, so there it would work).
 
 ```bash
-token-report.sh "${HOURS:-24}" "review-"
-TARGET_REPO=max-sixty/tend list-recent-runs.sh "tend-" "review-"
+SCRIPTS=<bundled scripts/ dir>
+"$SCRIPTS/token-report.sh" "${HOURS:-24}" "review-"
+TARGET_REPO=max-sixty/tend "$SCRIPTS/list-recent-runs.sh" "tend-" "review-"
 ```
 
 Under `review-runs`, `$HOURS` is the lookback derived from its Step 1 anchor —
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -46,11 +46,13 @@ default. `uvx tend@latest init` doesn't rewrite them either, so their
 ### Usage analysis
 
 Pass extra prefixes when running token reports or listing runs so these
-workflows are included:
+workflows are included. Take each script's path from the bundled skill that
+invokes it — `${CLAUDE_PLUGIN_ROOT}` is substituted into bundled plugin skills
+but not into this repo-local one, where it expands to nothing:
 
 ```bash
-"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" "${HOURS:-24}" "review-"
-TARGET_REPO=max-sixty/tend "${CLAUDE_PLUGIN_ROOT}/scripts/list-recent-runs.sh" "tend-" "review-"
+token-report.sh "${HOURS:-24}" "review-"
+TARGET_REPO=max-sixty/tend list-recent-runs.sh "tend-" "review-"
 ```
 
 Under `review-runs`, `$HOURS` is the lookback derived from its Step 1 anchor —
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -46,15 +46,11 @@ default. `uvx tend@latest init` doesn't rewrite them either, so their
 ### Usage analysis
 
 Pass extra prefixes when running token reports or listing runs so these
-workflows are included. Both scripts live in the bundled `scripts/` directory —
-take its path from any absolute `.../scripts/...` path a loaded bundled skill
-renders. Under the Claude harness `${CLAUDE_PLUGIN_ROOT}` is text-substituted
-into bundled plugin skill markdown but is not set in the shell, so writing the
-literal in this repo-local skill would expand to nothing (under Codex the
-action exports it as a real variable, so there it would work).
+workflows are included:
 
 ```bash
-SCRIPTS=<bundled scripts/ dir>
+# Claude leaves this unset in the shell; Codex exports it.
+SCRIPTS="${CLAUDE_PLUGIN_ROOT:-/home/tend-sandbox/tend-marketplace/plugins/tend-ci-runner}/scripts"
 "$SCRIPTS/token-report.sh" "${HOURS:-24}" "review-"
 TARGET_REPO=max-sixty/tend "$SCRIPTS/list-recent-runs.sh" "tend-" "review-"
 ```
--- a/codex/action.yaml
+++ b/codex/action.yaml
@@ -158,9 +158,11 @@ runs:
     # become discoverable by `codex exec`.
     #
     # We export `CLAUDE_PLUGIN_ROOT` to the install location so skills can
-    # resolve `${CLAUDE_PLUGIN_ROOT}/scripts/...` paths the same way they do
-    # under Claude Code (which sets this variable natively). One variable
-    # name across harnesses — skills stay harness-agnostic.
+    # resolve `${CLAUDE_PLUGIN_ROOT}/scripts/...` paths. Claude Code reaches the
+    # same place by a different route — it text-substitutes the variable into
+    # bundled plugin skill markdown at load time rather than setting it in the
+    # shell, so under Claude the literal resolves only inside a bundled skill.
+    # One variable name across harnesses — skills stay harness-agnostic.
     - name: Install tend-ci-runner plugin
       shell: bash
       run: |


.claude/skills/running-tend/SKILL.md
```diff
@@ -46,11 +46,13 @@ default. `uvx tend@latest init` doesn't rewrite them either, so their
 ### Usage analysis
 
 Pass extra prefixes when running token reports or listing runs so these
-workflows are included:
+workflows are included. Take each script's path from the bundled skill that
+invokes it — `${CLAUDE_PLUGIN_ROOT}` is substituted into bundled plugin skills
+but not into this repo-local one, where it expands to nothing:
```
tend-agent: Scopes the mechanism claim to the Claude harness (under Codex the action exports the variable, so it does expand), and points at the shared `scripts/` directory so the second line of the block resolves too — `review-reviewers` is the only bundled skill that invokes `list-recent-runs.sh`, and it isn't loaded in a `review-runs` session.

```suggestion
workflows are included. Prefix each script below with the `scripts/` directory
the bundled skills use — under the Claude harness `${CLAUDE_PLUGIN_ROOT}` is
text-substituted into bundled plugin skill markdown but is not set in the
shell, so the literal in this repo-local skill expands to nothing. Any
absolute `.../scripts/...` path a loaded bundled skill renders gives you the
directory; both scripts live in it:
```

tend-agent (commented):
The diagnosis holds — I reproduced it in this session: `CLAUDE_PLUGIN_ROOT` is unset in the Bash environment under the Claude harness, and `"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" 1 "review-"` exits 127. Two things about the replacement text, plus one observation about where the fix could live.

**The claim is harness-specific, but written as general.** `codex/action.yaml`'s "Install tend-ci-runner plugin" step parses the install root and writes `CLAUDE_PLUGIN_ROOT=$PLUGIN_ROOT` into `$GITHUB_ENV`, so under Codex it *is* a real environment variable and the old form expands correctly — including in a repo-local skill, which `codex/agents-tail.md` tells Codex to read ("Repo-local skills live under `.claude/skills/<name>/SKILL.md`"). Since tend supports per-workflow `harness:` overrides, this overlay can be read by a Codex session, where "it expands to nothing" is false. The inline suggestion scopes the sentence to the Claude harness.

**"From the bundled skill that invokes it" doesn't resolve for `list-recent-runs.sh`.** The only bundled skill that invokes it is `review-reviewers`, and that skill isn't loaded in a `review-runs` session — which is the session the next paragraph is written for. `token-report.sh` is fine (`review-runs` renders it), but the second line of the block has no source. Cheapest fix is to point at the *directory* rather than at a per-script skill: both scripts sit in the same `scripts/`, so one rendered absolute path resolves both. That also removes the "bare command name that isn't on `PATH`" shape — `token-report.sh …` copied verbatim still exits 127, just with a different message.

**Worth considering instead: fix it in the Claude action.** Under Claude the plugin root is deterministic — `$AGENT_HOME/tend-marketplace/plugins/tend-ci-runner`, from `shared/steps/install-tend-plugins.sh` — so the action could append `CLAUDE_PLUGIN_ROOT` to the agent env file the way `codex/action.yaml` writes it to `$GITHUB_ENV`. That would make the idiom work in *every* adopter's repo-local skill rather than documenting around it once here, and it would cover the related runtime slip cited in the PR body (a session typing `${CLAUDE_PLUGIN_ROOT}/scripts/poll-pr-checks.sh` into Bash and getting 127). It's action machinery rather than a deletion, so it's a maintainer call — but the comment in `codex/action.yaml` already asserts the parity it would create ("the same way they do under Claude Code (which sets this variable natively)"), and that assertion is currently wrong for the shell environment. Whichever way that goes, that comment is worth correcting.

Not approving: I authored this PR, so this is a self-review.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied in `cfcaef3`, with the wording adjusted so the code block carries the directory rather than the prose alone — the suggestion's block still left bare `token-report.sh` in it, which exits 127 the same way when copied. The block now assigns `SCRIPTS=<bundled scripts/ dir>` and prefixes both invocations with it, so the "bare name not on `PATH`" shape is gone rather than described.

Both other points are in the same commit: the mechanism claim is now scoped to the Claude harness, and `list-recent-runs.sh` resolves from the directory instead of from a skill that a `review-runs` session never loads.

tend-agent:
On the third point — setting `CLAUDE_PLUGIN_ROOT` in the Claude action's agent env rather than documenting around it — I've left it out of this PR and am recording the trade-off here rather than filing it, since it's a maintainer call on action machinery and this PR is a documentation correction.

It is feasible and small: the Claude harness passes the agent's environment through `$AGENT_ENV_FILE`, so one appended line in [`shared/steps/install-tend-plugins.sh`](https://github.com/max-sixty/tend/blob/cfcaef36c4c9f52c1db28640599c84e05cf9355c/shared/steps/install-tend-plugins.sh#L19) pointing at `$AGENT_HOME/tend-marketplace/plugins/tend-ci-runner` would do it, and it would make the idiom work in every adopter's repo-local skill instead of being documented around once here.

The argument against is that `CLAUDE_PLUGIN_ROOT` is a *per-plugin* value in Claude Code's rendering, and the Claude harness installs two plugins (`install-tend` and `tend-ci-runner`). A single shell variable has to pick one, so exporting it bakes in a one-plugin assumption that is right today only because nothing in `install-tend` uses the variable — the parity would be with Codex's own single-plugin export, not with Claude Code's semantics. Weighed against the observed cost (a 127, recovered in the same turn, nothing wrong posted), that reads like machinery this repo's bar doesn't want. Worth revisiting if `install-tend` ever grows a script reference, since that's the point where the ambiguity stops being latent.

The half of that point that was unambiguously a defect — the `codex/action.yaml` comment asserting Claude "sets this variable natively" — is corrected in `cfcaef3`.

.claude/skills/running-tend/SKILL.md
```diff
@@ -46,11 +46,17 @@ default. `uvx tend@latest init` doesn't rewrite them either, so their
 ### Usage analysis
 
 Pass extra prefixes when running token reports or listing runs so these
-workflows are included:
+workflows are included. Both scripts live in the bundled `scripts/` directory —
+take its path from any absolute `.../scripts/...` path a loaded bundled skill
+renders. Under the Claude harness `${CLAUDE_PLUGIN_ROOT}` is text-substituted
+into bundled plugin skill markdown but is not set in the shell, so writing the
+literal in this repo-local skill would expand to nothing (under Codex the
+action exports it as a real variable, so there it would work).
```
tend-agent: The mechanism explanation goes away if the block resolves the path itself (see the suggestion below). Folding it back to the one-sentence framing keeps this file to the recipe, per CLAUDE.md's "Authoring skills" guidance.

```suggestion
Pass extra prefixes when running token reports or listing runs so these
workflows are included:
```

.claude/skills/running-tend/SKILL.md
```diff
@@ -46,11 +46,17 @@ default. `uvx tend@latest init` doesn't rewrite them either, so their
...
 
 ```bash
-"${CLAUDE_PLUGIN_ROOT}/scripts/token-report.sh" "${HOURS:-24}" "review-"
-TARGET_REPO=max-sixty/tend "${CLAUDE_PLUGIN_ROOT}/scripts/list-recent-runs.sh" "tend-" "review-"
+SCRIPTS=<bundled scripts/ dir>
```
tend-agent: `SCRIPTS=<bundled scripts/ dir>` is a bash syntax error when copied verbatim (`syntax error near unexpected token 'newline'`), so the recipe still isn't runnable — and the prose that tells you what to substitute has no source under Codex, where nothing renders an absolute path. A `:-` fallback resolves on both harnesses with nothing to substitute: Codex sets the variable to the install root it parsed, Claude leaves it unset and the default fires.

```suggestion
# Claude leaves this unset in the shell; Codex exports it.
SCRIPTS="${CLAUDE_PLUGIN_ROOT:-/home/tend-sandbox/tend-marketplace/plugins/tend-ci-runner}/scripts"
```

tend-agent (commented):
The three points from the previous review are addressed, and the `codex/action.yaml` comment is now accurate — I re-verified the mechanism in this session: bundled skills render absolute `.../tend-ci-runner/scripts/...` paths, this repo-local skill renders the literal, and `CLAUDE_PLUGIN_ROOT` is unset in the Bash environment. Two things about the replacement recipe, which one change closes.

**The block still isn't runnable as written.** `SCRIPTS=<bundled scripts/ dir>` is a bash syntax error, not merely an unresolved path — I ran it: `syntax error near unexpected token 'newline'`. The PR's premise is that a session copied the recipe verbatim and got a 127; the replacement fails at the same point, just with a different message. That is the shape the previous review flagged in the bare-command-name revision, and it survived the fix.

**The derivation instruction has no source under Codex.** "Take its path from any absolute `.../scripts/...` path a loaded bundled skill renders" only holds under Claude. `codex/action.yaml` exports `CLAUDE_PLUGIN_ROOT` into `$GITHUB_ENV` *because* Codex doesn't text-substitute it into skill markdown — so under Codex no bundled skill renders an absolute path, and there is nothing to take. The paragraph notes Codex would expand the literal, but the block no longer uses the literal, so the one harness where the old form worked now has no working form either.

Both dissolve into one line: `SCRIPTS="${CLAUDE_PLUGIN_ROOT:-<claude path>}/scripts"`. Under Codex the variable is set and it resolves to the parsed install root; under Claude it is unset and the fallback fires — I checked in this session that both scripts are executable at that path. That makes the block copy-paste runnable on both harnesses, which in turn makes the six added lines of mechanism prose unnecessary: CLAUDE.md's "Authoring skills" section asks for brevity and "a code block plus a one-sentence framing" over a multi-paragraph explanation, and this file loads into every tend session on this repo.

The PR body argues against hardcoding `/home/tend-sandbox/...` because it would be "wrong under a harness switch". As a `:-` fallback it isn't: Codex sets the variable, so the fallback never fires there. It does still rot if the Claude sandbox home changes — but to the same 127 as today, and to something greppable in-repo, which beats a syntax error on every copy.

Not approving: I authored this PR, so this is a self-review.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied in `b2195db` — the framing is back to the one sentence, and the six lines of mechanism prose are gone with it. They only existed to tell the reader what to substitute, and the block no longer asks for a substitution.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied in `b2195db`. Both objections land: the placeholder is a syntax error when copied, and the derivation instruction genuinely has no source under Codex — `codex/action.yaml` exports the variable *because* Codex doesn't substitute it into skill markdown, so there is no rendered absolute path to take it from there.

Verified the `:-` form in this session rather than reasoning it through: `CLAUDE_PLUGIN_ROOT` is unset in this shell, the fallback resolves to `/home/tend-sandbox/tend-marketplace/plugins/tend-ci-runner/scripts`, and both scripts are executable there. Then extracted the block from the file and ran it verbatim — `bash -n` clean, `HOURS=1` run exits 0 with both reports printing. That is the check the two prior revisions skipped.

On the hardcoded path: agreed it isn't the harness-switch hazard the PR body claimed, since Codex sets the variable and the fallback never fires there. I've rewritten that paragraph rather than leaving the old argument standing against the new code.

PR #1143, opened 2026-09-05
fix(running-in-ci): resolve the SHAs in a body before posting it

### LINKED ISSUE #1142: running-in-ci: the pre-post link scan checks that a #L link is SHA-pinned, not that the SHA exists
`running-in-ci`'s **Comment Formatting** prescribes a pre-post scan over the composed body: "confirm every owner matches `$GITHUB_REPOSITORY`, **and** every URL with a `#L<n>` anchor is SHA-pinned. A `blob/main/...#L<n>` hit is the link-rot shape". That catches an *un-pinned* link (the shape [#658](https://github.com/max-sixty/tend/pull/658) added) but not a *pinned-to-nothing* one: a hand-typed 40-hex string satisfies "is SHA-pinned" whether or not the object exists, so a fabricated OID passes the scan and ships a permalink that 404s.

This just happened on numbagg for the second time, and this time nothing in the session caught it — the dead link sat live on a public PR body until a later sweep repaired it.

## What happened

`tend-nightly` [33961296576](https://github.com/numbagg/numbagg/actions/runs/33961296576) opened [numbagg/numbagg#783](https://github.com/numbagg/numbagg/pull/783) with a provenance link in the body:

```
([decorators.py#L305-L331](https://github.com/numbagg/numbagg/blob/b375b9d2a1a2b98f6f2b7cef0d0ac66ba5d51d29/numbagg/decorators.py#L305-L331))
```

`main` was `b375b9d4b09af41bb5a8041b26bd681607782f23`. The two diverge at character 8. `GET /repos/numbagg/numbagg/commits/b375b9d2a1a2b98f6f2b7cef0d0ac66ba5d51d29` returns `422 No commit found for SHA`, so the link was dead from the moment the PR was created.

The provenance is unambiguous in the session JSONL. `grep -o 'b375b9d[0-9a-f]*'` over the log returns `b375b9d` eleven times — every one from `git log --oneline` output — and the 40-hex `b375b9d2a1a2b98f6f2b7cef0d0ac66ba5d51d29` exactly once, in the heredoc that composed the PR body. The session never ran `git rev-parse HEAD`; it saw the abbreviated form and extended it. The line range itself is correct at the real SHA, so only the identity of the commit was invented.

## Why the existing guidance does not cover it

Two rules point at this and both stop one step short:

- **Comment Formatting**'s pre-post scan tests the *shape* (`blob/main/...` vs. a hex string), not resolvability.
- The same section says "Get the SHA with `git rev-parse HEAD` before composing the link" — correct, and unheeded here, which is the stochastic half. A prose restatement adds nothing a session that already read it would gain; that is why numbagg's sweeps have declined to add repo-local prose for five prior occurrences.

The gap is that there is no *check* between the invented SHA and the post. Everything else in that scan is mechanical and cheap; this one is missing.

## Proposed remedy

Extend the pre-post scan with an existence check on any 40-hex SHA in the body — one `gh api` call per distinct SHA, and bodies carry one or two:

```bash
grep -oE 'blob/[0-9a-f]{40}' /tmp/comment-body.md | cut -d/ -f2 | sort -u | while read -r s; do
  gh api "repos/$GITHUB_REPOSITORY/commits/$s" --jq '.sha' >/dev/null 2>&1 \
    || echo "DEAD SHA in body: $s"
done
```

A cheaper variant that avoids the API call is to prescribe capturing the SHA into a shell variable and interpolating it into the body (`SHA=$(git rev-parse HEAD)`), so the string is never typed — but that only closes the `HEAD` case, while the existence check closes the class regardless of where the SHA came from.

## Evidence

Six occurrences of the narrow "invent the tail of an abbreviated SHA" shape on numbagg, recorded across its review-runs tracking issues. Five reached only `poll-pr-checks.sh`, where [#1046](https://github.com/max-sixty/tend/pull/1046) and [#1054](https://github.com/max-sixty/tend/issues/1054) turned a nine-minute silent stall into a first-iteration verdict that names the cause — that path is fixed and has not regressed. Two reached the public record instead, where no such sentinel exists:

- numbagg/numbagg#762 (2026-08): `blob/bfcddcfa/...` against the real `bfcddcf8`, live ~9m40s. The session found it only because a failed poll happened to send it to `git rev-parse HEAD`.
- numbagg/numbagg#783 (today): the case above. **Nothing sent this session to `git rev-parse HEAD`** — it opened two PRs, polled both with correctly-sourced SHAs, and finished clean. The poll-side sentinel is exactly what was missing, and the body-side has no equivalent.

So the two halves are not independent: the first public-record occurrence was caught by the poll path as a side effect, and now that the poll path resolves a good SHA on the first iteration, that accidental backstop is gone. The failure is a wrong outward action — a citation that resolves to nothing, in the artifact a maintainer reads to decide whether to merge.

```
$ gh api repos/numbagg/numbagg/commits/b375b9d2a1a2b98f6f2b7cef0d0ac66ba5d51d29 --jq '.sha'
gh: No commit found for SHA: b375b9d2a1a2b98f6f2b7cef0d0ac66ba5d51d29 (HTTP 422)

$ git rev-parse origin/main
b375b9d4b09af41bb5a8041b26bd681607782f23

$ grep -o 'b375b9d[0-9a-f]*' <session.jsonl> | sort | uniq -c
     11 b375b9d
      1 b375b9d2a1a2b98f6f2b7cef0d0ac66ba5d51d29
```

`L305-L331` at the real SHA is `ndmove.__call__` through its `len(axis) > 1` guard — the range the body describes, so the citation was right about everything except which commit it pointed at.

The body has since been corrected in place on numbagg/numbagg#783, so the link there now resolves; the run artifact and the session log preserve the original.

</details>

tend-agent:
## Problem

`running-in-ci`'s pre-post scan checked that a `#L` link was *SHA-pinned*, not that the SHA *existed*. A hand-typed 40-hex string satisfies the shape either way, so a fabricated OID — the model extending an abbreviation it saw in `git log` output rather than running `git rev-parse HEAD` — passed the scan and shipped a permalink that 404s from the moment it was posted. #1142 documents two occurrences that reached a public PR body, the second with nothing in the session to catch it: the poll-side sentinel that accidentally backstopped the first now resolves a good SHA on the first iteration, so it never sends the session to `git rev-parse HEAD` any more.

## Solution

`plugins/tend-ci-runner/scripts/check-body-links.sh <body-file>` resolves every 40-hex SHA in a composed body against the API — one call per distinct (repo, SHA) pair, and bodies carry one or two — and reports any `#L` anchor pinned to a branch or an abbreviation. A hand-typed owner falls out of the same call, since `repos/<wrong-owner>/<repo>/commits/<sha>` does not resolve either. Exit 1 with one problem per line, exit 0 clean.

The Comment Formatting section now invokes it in place of the by-eye scan. The owner rule stays as prose for the links the script cannot check (issue/PR links carry no SHA to resolve).

Deliberately reported rather than swallowed: a repo the token cannot read, and a commit this session has made but not yet pushed, both read as unresolvable. The second is correct — the link would 404 for a reader — so the skill says to run the check after the push.

## Testing

`generator/tests/test_check_body_links.py` (10 tests) drives the script against a fake `gh` that resolves exactly one (repo, SHA) pair: a fabricated OID and a wrong owner are reported, a resolvable permalink and a file-level branch link pass, a branch- or abbreviation-pinned `#L` anchor is reported without an API call, a repeated permalink costs one call, and neither markdown delimiters (`)`, backticks) nor sentence punctuation after a bare URL leak into the parsed ref.

Full `uv run pytest` is green (925 passed), as is `pre-commit run --all-files` — shellcheck at warning severity covers `plugins/**/*.sh`.

---
Closes #1142 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/scripts/check-body-links.sh
+++ b/plugins/tend-ci-runner/scripts/check-body-links.sh
@@ -58,6 +58,9 @@ while IFS= read -r url; do
   # https://github.com/OWNER/REPO/blob/REF/path -> fields 4, 5, 7
   read -r slug ref <<<"$(printf '%s\n' "$url" | awk -F/ '{print $4 "/" $5, $7}')"
   ref=${ref%%#*}
+  # A bare URL in prose ends at the sentence, so trailing punctuation lands in
+  # the ref of a `commit/<sha>` link and would skip the resolve entirely.
+  ref=${ref%[.,;:!?]}
   if [[ "$ref" =~ ^[0-9a-f]{40}$ ]]; then
     printf '%s %s\n' "$slug" "$ref" >>"$shas"
   elif [[ "$url" == *"#L"* ]]; then
--- a/plugins/tend-ci-runner/scripts/check-body-links.sh
+++ b/plugins/tend-ci-runner/scripts/check-body-links.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+# Checks the GitHub links in a composed body before it is posted.
+#
+# Usage: check-body-links.sh <body-file>
+#
+# Two failure shapes, both of which ship a citation that resolves to nothing or
+# to different code than the text describes, in the artifact a maintainer reads
+# to decide whether to merge:
+#
+#   Fabricated SHA. The model sees an abbreviated OID in `git log` output and
+#   extends it to 40 characters instead of running `git rev-parse HEAD`. The
+#   result is well-formed, so any check that tests the *shape* of a permalink
+#   passes it, and the link 404s from the moment it is posted. Only resolving
+#   the OID separates a real commit from an invented one.
+#
+#   Un-pinned line anchor. `blob/main/...#L42` stays valid while the lines
+#   underneath it move, so the link silently comes to point at other code. A
+#   ref that is not a full 40-hex OID cannot pin a line.
+#
+# One API call per distinct (repo, SHA) pair, and bodies carry one or two. A
+# hand-typed owner falls out of the same call: `repos/<wrong-owner>/<repo>/
+# commits/<sha>` does not resolve either.
+#
+# A repo the token cannot read reports the same way a dead SHA does, so the
+# message names both readings rather than asserting the link is dead. A commit
+# this session made but has not pushed yet reports the same way too, and there
+# it is right: the link would 404 for a reader. Run the check after the push.
+#
+# Exit codes:
+#   0  every GitHub link in the body checks out
+#   1  problems found, one per line on stdout
+#   2  usage error
+
+set -euo pipefail
+
+if [ "$#" -ne 1 ]; then
+  echo "usage: check-body-links.sh <body-file>" >&2
+  exit 2
+fi
+
+body="$1"
+if [ ! -f "$body" ]; then
+  echo "check-body-links.sh: no such file: $body" >&2
+  exit 2
+fi
+
+# A URL ends at the delimiters that surround it in markdown: `)` closes a link,
+# `]`, quotes, backticks, angle brackets and whitespace end a bare one. The
+# leading `]` inside the negated class is literal.
+url_re='https?://github\.com/[A-Za-z0-9._-]+/[A-Za-z0-9._-]+/(blob|blame|tree|commit|commits|raw)/[^][)("<>`'"'"'[:space:]]+'
+
+problems=$(mktemp)
+shas=$(mktemp)
+trap 'rm -f "$problems" "$shas"' EXIT
+
+while IFS= read -r url; do
+  [ -n "$url" ] || continue
+  # https://github.com/OWNER/REPO/blob/REF/path -> fields 4, 5, 7
+  read -r nwo ref <<<"$(printf '%s\n' "$url" | awk -F/ '{print $4 "/" $5, $7}')"
+  ref=${ref%%#*}
+  if [[ "$ref" =~ ^[0-9a-f]{40}$ ]]; then
+    printf '%s %s\n' "$nwo" "$ref" >>"$shas"
+  elif [[ "$url" == *"#L"* ]]; then
+    printf 'un-pinned line link: ref `%s` is not a full commit SHA, so the lines it points at can move — %s\n' \
+      "$ref" "$url" >>"$problems"
+  fi
+done < <(grep -oE "$url_re" "$body" || true)
+
+while read -r nwo sha; do
+  [ -n "$sha" ] || continue
+  gh api "repos/$nwo/commits/$sha" --jq '.sha' >/dev/null 2>&1 && continue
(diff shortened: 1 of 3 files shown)


tend-agent (commented):
One gap in the new gate. The ref is taken as awk field 7, so on a `blob/<sha>/path` link — the shape #1142 documents — trailing prose punctuation lands in the *path* and the SHA still resolves. On a `commit/<sha>` or `tree/<sha>` link the SHA *is* field 7, so a sentence-ending `.` or `,` makes the ref `<sha>.`, which fails the 40-hex test, carries no `#L`, and falls through both branches silently. A fabricated OID in that shape passes the check with exit 0.

Bare commit URLs in prose are a normal body shape ("pushed as https://github.com/…/commit/<sha>."), and `)`-closed markdown links are already handled by the exclusion class, so the one-line trim below closes the remaining case. I ran it against the nine tests in this PR and they stay green.

<details><summary>Repro at 36baf7b</summary>

```
$ printf 'Pushed as https://github.com/max-sixty/tend/commit/0123456789abcdefdeadbeefdeadbeefdeadbeef.\n' > /tmp/t3.md
$ bash check-body-links.sh /tmp/t3.md; echo "exit=$?"
exit=0

$ # same body without the trailing period
exit=1
unresolvable SHA 0123456789abcdefdeadbeefdeadbeefdeadbeef in max-sixty/tend — the commit does not exist …
```

With the suggested trim, both report. Worth a test alongside `test_markdown_delimiters_do_not_leak_into_the_ref`, which covers the `)` and backtick delimiters but not sentence punctuation.

</details>

plugins/tend-ci-runner/scripts/check-body-links.sh
```diff
@@ -0,0 +1,79 @@
...
+
+problems=$(mktemp)
+shas=$(mktemp)
+trap 'rm -f "$problems" "$shas"' EXIT
+
+while IFS= read -r url; do
+  [ -n "$url" ] || continue
+  # https://github.com/OWNER/REPO/blob/REF/path -> fields 4, 5, 7
+  read -r slug ref <<<"$(printf '%s\n' "$url" | awk -F/ '{print $4 "/" $5, $7}')"
+  ref=${ref%%#*}
```
tend-agent, line 60:
```suggestion
  ref=${ref%%#*}
  # A bare URL in prose ends at the sentence, so trailing punctuation lands in
  # the ref of a `commit/<sha>` link and would skip the resolve entirely.
  ref=${ref%[.,;:!?]}
```

tend-agent: Applied in 14b1463 — the trim goes in right after the `#` strip, so a bare `commit/<sha>.` in prose now resolves the SHA instead of falling through.

Added `test_sentence_punctuation_does_not_skip_the_resolve` alongside the delimiter test: the fabricated OID with a trailing `.` is reported, and a real one followed by `,` still resolves (so the trim doesn't cost a good link its check). Full `uv run pytest` is green at 925, and `pre-commit run --all-files` passes.

PR #1149, opened 2026-09-05
fix(logs): copy the agent's session tree out as root instead of opening it up

max-sixty:
The token-usage step consolidates the agent's session JSONL so it can be uploaded as an artifact. To read it, the step granted itself access to the source, as root, in a tree the sandbox user owns:

```
sudo chmod a+x     $AGENT_HOME $AGENT_HOME/.claude
sudo chmod -R a+rX $AGENT_HOME/.claude/projects
```

`chmod` follows a symlinked argument, and the agent owns both `.claude` and `projects`, so replacing either with a link points an `a+rX` at whatever tree the agent names. Nothing checked for one. The `cp -a` that follows has the same shape from the other side: it reproduces whatever it finds, and `upload-artifact` then reads the result as the runner, which can read the proxy's CA private key and its own `/proc/<pid>/environ`.

This copies as root and chowns the result to the runner instead. That needs no permission change on the agent's side, so the escalation primitive is gone rather than guarded.

Four things then have to hold for the copy that replaces it, and the agent chooses the input to each:

- **Neither boundary is a symlink.** `cp -a` follows a symlinked argument, so the session directory and the dot-directory above it are both checked. The check asks root when it cannot answer itself: `Path.is_symlink` reports an unreachable path as "not a symlink", and the sandbox home is the agent's to `chmod` shut, so trusting the runner's own `lstat` would wave through exactly the link the check exists to catch. It falls back to `sudo test`, and an answer neither can give reads as unsafe.
- **Nothing is alive to change what was checked.** A check holds only while nothing can act after it, so the copy waits on `sandbox_reaped` — the output the memory save already keys on, for the same reason and against a read that isn't even privileged.
- **The copied modes are the runner's.** `cp -a` preserves the agent's directory modes and `chown` leaves them alone, and deleting an entry needs write on its parent. Without normalising them, a `0555` directory keeps what it holds (and raises out of an `if: always()` step) and a `0000` one hides it.
- **Only regular files and directories reach the upload.** The sweep tested for a symlink while its docstring named the invariant that matters — the agent writes only regular files there. A `mkfifo` rode the copy out, and since `upload-artifact` streams every entry that isn't a directory, opening a FIFO with no writer hangs the upload to the job timeout. Testing the mode covers links, FIFOs and sockets at once.

Only the Claude harness is affected: `codex_step` reads the runner's own `$HOME/.codex`, with no privileged copy.

`docs/security-model.md` gains a paragraph on this boundary — it enumerates the others and had nothing on the session-log copy.

Three docstrings said `tend-review` runs with `cancel-in-progress: true`. It runs with `false`; `tend-triage` is the one that cancels in progress, so it's the one that makes a result-less session routine.

<details>
<summary>Test coverage</summary>

Eight tests, each failing against the current `main` and passing with the change:

- a symlinked `.claude/projects` copies nothing
- a symlinked `.claude` copies nothing (the boundary the old code had no notion of)
- a `.claude` the runner cannot stat is re-asked as root rather than answered `False`
- an absent `AGENT_HOME` runs no privileged command at all
- an unreaped sandbox copies nothing
- a link planted inside the session dir never reaches the uploaded artifact
- nor does one inside a `0555` directory, which without the mode fix both survives and raises
- nor does a FIFO

</details>

### DIFF
--- a/docs/security-model.md
+++ b/docs/security-model.md
@@ -340,6 +340,14 @@ made as the sandbox user with `sandbox_setup:`. A generic failure shim keeps a
 dropped home-selected command from silently falling through to a different
 same-named system tool.
 
+**Session-log upload.** The token-usage step uploads the agent's session JSONL
+as an artifact, so the runner has to read a tree the sandbox user owns. It
+copies as root and chowns the result rather than making the source readable,
+since a `chmod -R` aimed through a symlink the agent planted would grant read on
+whatever tree it named. Both the session directory and the dot-directory above
+it are refused if either is a symlink, and links inside the copy are deleted
+before `upload-artifact` resolves them as the runner.
+
 The Codex harness (`codex/action.yaml`) still passes both the PAT and the model
 auth directly to the agent. The merge restriction and the environment gate
 remain the load-bearing boundaries regardless of harness.
--- a/docs/security-model.md
+++ b/docs/security-model.md
@@ -344,9 +344,16 @@ same-named system tool.
 as an artifact, so the runner has to read a tree the sandbox user owns. It
 copies as root and chowns the result rather than making the source readable,
 since a `chmod -R` aimed through a symlink the agent planted would grant read on
-whatever tree it named. Both the session directory and the dot-directory above
-it are refused if either is a symlink, and links inside the copy are deleted
-before `upload-artifact` resolves them as the runner.
+whatever tree it named.
+
+The agent chooses the input to every check that follows. The copy waits for the
+supervisor to reap every sandbox process, so nothing is left alive to change
+what was checked. The session directory and the dot-directory above it are
+refused if either is a symlink. The copied modes are reset to the runner's,
+since deleting an entry needs write on its parent and those modes came from the
+agent. Every entry that is not a regular file or a directory is deleted before
+`upload-artifact` reads it: a symlink it would otherwise resolve as the runner,
+a FIFO it would block on until the job times out.
 
 The Codex harness (`codex/action.yaml`) still passes both the PAT and the model
 auth directly to the agent. The merge restriction and the environment gate
--- a/shared/steps/token_usage.py
+++ b/shared/steps/token_usage.py
@@ -39,9 +39,9 @@ Claude's accounting has three paths, tried in order:
    ``usage.*`` and ``num_turns`` are per-event while ``total_cost_usd`` is
    cumulative, so the per-event fields are summed and cost taken from the last.
 2. No result event: the session JSONL this step has just consolidated. A
-   cancelled session never emits a result, and ``tend-review`` runs with
+   cancelled session never emits a result, and ``tend-triage`` runs with
    ``cancel-in-progress: true``, so this is routine rather than exotic — the
-   run may have done dozens of turns and already posted its review.
+   run may have done dozens of turns and already posted its comment.
 3. Neither: the agent never ran (a preflight failure, say). That run genuinely
    cost nothing, so it reports a real zero rather than flagging an unknown.
 
@@ -204,13 +204,25 @@ def codex_step(model: str) -> tuple[dict[str, Any], Path]:
 def consolidate_logs(logs_dir: Path, runner_temp: Path) -> None:
     """Copy the agent's session JSONL and stderr log into a runner-owned dir.
 
+    The copy waits on ``SANDBOX_REAPED``, the supervisor's record that every
+    sandbox process is dead. :func:`copy_agent_tree` checks the tree it is about
+    to read, and a surviving agent could swap a session dir for a symlink
+    between that check and the root copy that follows. The memory save keys on
+    the same output for the same reason, and that read is unprivileged where
+    this one runs as root.
+
+    It is set from a ``finally``, so only a supervisor killed outright — a
+    second signal during an escalating cancel — leaves it unset. That run
+    reports its usage from the stream-json and uploads no session logs.
+
     ``AGENT_HOME`` is unset when sandbox setup died before exporting it; the
     placeholder names nothing, which :func:`copy_agent_tree` reads as the same
     nothing-to-copy outcome as a run whose agent never started.
     """
     logs_dir.mkdir(parents=True, exist_ok=True)
     agent_home = Path(os.environ.get("AGENT_HOME") or "/nonexistent")
-    copy_agent_tree(agent_home / ".claude" / "projects", logs_dir)
+    if os.environ.get("SANDBOX_REAPED") == "true":
+        copy_agent_tree(agent_home / ".claude" / "projects", logs_dir)
     best_effort("cp", "-a", str(runner_temp / "tend-claude-stderr.log"), f"{logs_dir}/")
 
 
--- a/shared/steps/token_usage.py
+++ b/shared/steps/token_usage.py
@@ -239,11 +239,12 @@ def copy_agent_tree(source: Path, destination: Path) -> None:
     - **No symlink at either boundary.** ``cp -a`` follows a symlinked
       *argument*, so the session dir and the dot-dir holding it are both
       checked; see :func:`safe_agent_directory`.
-    - **No symlink inside.** ``cp -a`` copies nested links as links, and they
-      are deleted before anything reads the copy; see :func:`drop_symlinks`.
+    - **Only regular files inside.** ``cp -a`` reproduces a nested symlink or
+      FIFO as one, and both are removed before anything reads the copy; see
+      :func:`drop_non_regular_files`.
       ``cp -a`` also preserves the agent's directory modes and ``chown`` leaves
       them alone, so the modes are normalised too: unlinking needs write on the
-      parent, and a ``0555`` directory would keep its link (and raise), a
+      parent, and a ``0555`` directory would keep what it holds (and raise), a
       ``0000`` one hide it. That chmod is nothing like the one this replaces —
       it lands on a destination the runner made and now owns, so there is
       nothing for the agent to aim it at.
@@ -260,7 +261,7 @@ def copy_agent_tree(source: Path, destination: Path) -> None:
     best_effort("sudo", "cp", "-a", f"{source}/.", f"{destination}/")
     best_effort("sudo", "chown", "-R", f"{os.getuid()}:{os.getgid()}", str(destination))
     best_effort("sudo", "chmod", "-R", "u+rwX", str(destination))
-    drop_symlinks(destination)
+    drop_non_regular_files(destination)
 
 
 def safe_agent_directory(path: Path) -> bool:
@@ -295,16 +296,20 @@ def root_test(flag: str, path: Path) -> bool:
     )
 
 
-def drop_symlinks(logs_dir: Path) -> None:
-    """Remove every symlink from the dir about to be uploaded.
+def drop_non_regular_files(logs_dir: Path) -> None:
+    """Remove everything but files and directories from the dir to be uploaded.
 
-    ``cp -a`` copies links as links and ``upload-artifact`` resolves them as
-    the runner — which can read the proxy's CA private key and its own
-    ``/proc/<pid>/environ``. The agent writes only regular files here, so a
-    link is it reaching for something else.
+    The agent writes only regular files here, so anything else is it reaching
+    for something other than its own session log, and ``cp -a`` reproduces
+    whatever it finds. A symlink stays a symlink, and ``upload-artifact``
+    resolves it as the runner — which can read the proxy's CA private key and
+    its own ``/proc/<pid>/environ``. A FIFO stays a FIFO, and the upload
+    streams every entry that isn't a directory, so opening one with no writer
+    blocks until the job times out. One test on the mode covers both.
     """
     for path in list(logs_dir.rglob("*")):
-        if path.is_symlink():
+        mode = path.lstat().st_mode
+        if not (stat.S_ISREG(mode) or stat.S_ISDIR(mode)):
             path.unlink()
 
 
--- a/shared/steps/token_usage.py
+++ b/shared/steps/token_usage.py
@@ -241,6 +241,12 @@ def copy_agent_tree(source: Path, destination: Path) -> None:
       checked; see :func:`safe_agent_directory`.
     - **No symlink inside.** ``cp -a`` copies nested links as links, and they
       are deleted before anything reads the copy; see :func:`drop_symlinks`.
+      ``cp -a`` also preserves the agent's directory modes and ``chown`` leaves
+      them alone, so the modes are normalised too: unlinking needs write on the
+      parent, and a ``0555`` directory would keep its link (and raise), a
+      ``0000`` one hide it. That chmod is nothing like the one this replaces —
+      it lands on a destination the runner made and now owns, so there is
+      nothing for the agent to aim it at.
 
     A boundary that fails the check copies nothing: an empty artifact, the same
     thing a run whose agent never started leaves behind. Telling the two apart
@@ -253,6 +259,7 @@ def copy_agent_tree(source: Path, destination: Path) -> None:
     destination.mkdir(parents=True, exist_ok=True)
     best_effort("sudo", "cp", "-a", f"{source}/.", f"{destination}/")
     best_effort("sudo", "chown", "-R", f"{os.getuid()}:{os.getgid()}", str(destination))
+    best_effort("sudo", "chmod", "-R", "u+rwX", str(destination))
     drop_symlinks(destination)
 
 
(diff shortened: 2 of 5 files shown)


shared/steps/token_usage.py
```diff
@@ -203,24 +204,95 @@ def codex_step(model: str) -> tuple[dict[str, Any], Path]:
...
+    A boundary that fails the check copies nothing: an empty artifact, the same
+    thing a run whose agent never started leaves behind. Telling the two apart
+    is not worth failing an ``if: always()`` step over a directory the runner
+    may simply be unable to see.
+    """
+    for boundary in (source.parent, source):
+        if not safe_agent_directory(boundary):
+            return
+    destination.mkdir(parents=True, exist_ok=True)
+    best_effort("sudo", "cp", "-a", f"{source}/.", f"{destination}/")
```
tend-agent, line 261: The window between `safe_agent_directory` and this line is closed only by the agent being dead, and the step doesn't check that it is. `Token usage` is `if: always()` with no gate, while the memory-save step above it carries `steps.claude.outputs.sandbox_reaped == 'true'` for exactly this reason — `run_claude.py`'s comment on that output says it exists "so it never reads agent-owned files while a sandbox process survives" — and that read is unprivileged where this one is root.

`sandbox_reaped` is set from a `finally`, so the gap is a supervisor that never reached it: the SIGKILL after an escalating cancel, which `raise_on_cancel`'s own docstring contemplates ("a second signal during the reap ends the process outright"). A surviving sandbox process could then swap `projects` for a link between the check and this `cp`.

Narrow, and I don't think it's worth new machinery — the step has to keep reporting usage whether or not the reap landed. But the docstring above states the reap as the premise, so it seems worth either keying the copy on the output the neighbouring step already reads, or saying in the docstring why this one doesn't need it.

tend-agent (commented):
The copy-don't-grant shape is the right fix — it removes the primitive rather than guarding it — and the boundary handling holds up: I confirmed `cp -a src/. dst/` does resolve a symlinked `src` (so the check is load-bearing), `chown -R` refuses `--dereference` without `-H`/`-L` so it acts on the links themselves, and `pathlib.rglob("*")` returns hidden entries and descends hidden directories without following symlinked ones, so `drop_symlinks` covers a `.leak` too. One observation inline on the privileged copy.

Separately, outside this diff: the module docstring's path-2 explanation (the paragraph ending "so this is routine rather than exotic") says `tend-review` runs with `cancel-in-progress: true`. `review.yaml.j2` sets `cancel-in-progress: false`, with a comment explaining why; `triage.yaml.j2` is the one that's `true`. Pre-existing, and it's the docstring for the accounting path this step feeds — happy to push the correction here if you'd rather not carry it separately.

shared/steps/token_usage.py
max-sixty: Both taken, in 16cc9fa0.

The race is real, and the reap is the right key for it — the premise the docstring states isn't guaranteed, so stating it and not enforcing it was the wrong half of your two options. `Token usage` stays `if: always()` and keeps reporting usage; only the copy waits on `steps.claude.outputs.sandbox_reaped`, which is now wired into the step's env. A run whose supervisor was killed outright reports from the stream-json and uploads no session logs, which is the right trade when something may still be alive in the sandbox.

On the docstring: confirmed. `review.yaml.j2` sets `cancel-in-progress: false` with a comment on why, and `triage.yaml.j2` is the `true` one — so `tend-triage` is what makes a result-less session routine. Corrected here rather than separately, since it's the docstring for the path this step feeds.

tend-agent: Verified on 16cc9fa. The gate is wired on the Claude step's env only, and `consolidate_logs` is reached only from `claude_step`, so the Codex path — which reads the runner's own `$HOME/.codex` — doesn't lose its logs to a variable nothing sets there. `sandbox_reaped` is published from the `finally`, so a timed-out run still uploads. And the templates match the corrected docstring: `review.yaml.j2` is `false`, `triage.yaml.j2` is `true`. `test_token_usage.py` passes locally, 16 tests.

One neighbour I checked while confirming: `drop_symlinks` runs inside `copy_agent_tree`, so the stderr `cp -a` one line below it lands in the uploaded dir unscanned. Not a hole — `tend-claude-stderr.log` sits in `RUNNER_TEMP`, which `setup-sandbox.sh` notes is runner-owned and not the sandbox's to write, so nothing can make it a link. Worth saying only because it's the same shape as the copy above it.

shared/steps/token_usage.py
```diff
@@ -203,24 +204,107 @@ def codex_step(model: str) -> tuple[dict[str, Any], Path]:
...
+    best_effort("sudo", "chown", "-R", f"{os.getuid()}:{os.getgid()}", str(destination))
```
tend-agent, line 262: `cp -a` preserves the source's modes and `chown -R` doesn't touch them, so the agent chooses the mode of every directory in the copy — including one the runner cannot write. `unlink` needs write on the *parent*, so a link dropped into a `0555` directory survives `drop_symlinks`:

```
mkdir -p ~/.claude/projects/p
ln -s <ca-key> ~/.claude/projects/p/leak.pem
chmod 555 ~/.claude/projects/p
```

Reproduced against this branch: `drop_symlinks` raises `PermissionError: .../logs/p/leak.pem` and the link is still there. Two consequences, the second being the one this PR exists to prevent:

- `_common.run` catches only `CalledProcessError`, so the traceback exits the `if: always()` accounting step non-zero — no `token-usage.json`, no `usage` output, no summary.
- `Upload session logs` is `if: always()` as well, so it runs anyway and resolves the surviving link as the runner.

Normalising the modes as root right after the chown closes both, and can't be aimed anywhere: the destination is runner-owned by then, and `chmod -R` neither traverses nor dereferences symlinks (checked — the link target's mode is unchanged). It also covers the `0000` directory, where `rglob` yields nothing and the link survives silently.

Takes `best_effort`'s docstring with it — "a copy or a chown" becomes a chmod too.

```suggestion
    best_effort("sudo", "chown", "-R", f"{os.getuid()}:{os.getgid()}", str(destination))
    # `cp -a` preserves the agent's modes, and a directory the runner cannot
    # write is a link `drop_symlinks` cannot unlink.
    best_effort("sudo", "chmod", "-R", "u+rwX", str(destination))
```

tend-agent (commented):
The reap gate is the right key: `sandbox_reaped` is set from `supervise`'s `finally`, so the only run that loses its session logs is one whose supervisor was killed outright — and that is exactly the run whose boundary check couldn't be trusted. Docstring correction confirmed against `triage.yaml.j2`.

One thing I missed last pass, on `drop_symlinks`: it's the last guard before the upload, and a directory mode the agent chooses defeats it. Inline, with a suggestion.

shared/steps/token_usage.py
max-sixty: Taken as suggested, in ba016abc — I reproduced both halves against the branch first.

`0555`: `drop_symlinks` raises `PermissionError` and `leak.pem` is still in the copy. `0000`: no exception at all, `rglob` yields nothing inside, and the link stays — the quieter one, and the one I'd have missed without the mode you named. `chmod -R u+rwX` on the destination clears both, and the link target's mode is unchanged after it, so nothing is being followed.

The test goes through `consolidate_logs` with a `0555` project directory, so it fails without the chmod in both of the ways you describe. `0000` needs a root `cp -a` to set up faithfully — the test user can't read its own `0000` directory, so nothing copies — so that case is named in the test's docstring rather than pinned; it's the same cause and the same one-line fix.

`best_effort`'s docstring now says "a copy, a chown, or a chmod".

tend-agent: Confirmed on ba016ab. I reproduced both directory modes standalone and then ran the suite against the branch with and without the new line.

`0555`: without [`best_effort("sudo", "chmod", "-R", "u+rwX", ...)`](https://github.com/max-sixty/tend/blob/ba016abc974491c18369a1ef641e98de8b53474e/shared/steps/token_usage.py#L262) the new test fails with exactly the shape the commit message describes — `PermissionError: [Errno 13] Permission denied: .../logs/project/leak.pem` out of `Path.unlink`, so the link is still in the tree the upload step then reads. With the line, 17 tests pass. `0000`: `rglob` yields nothing under the directory before the chmod and the link is unlinked after it, so the quiet variant closes on the same line.

No dereference: after `chmod -R u+rwX` over a tree holding a link to a `0600` file outside it, the target is still `0600` and the copied directory is `755`. That matches `chmod`'s documented physical traversal, and the ordering is the part that makes it moot anyway — the chmod lands after the `chown`, on a destination the runner made and owns, so there is no agent-chosen path in its argument.

Agreed on leaving `0000` to the docstring — it needs a root `cp -a` to stage faithfully, and it is the same parent-write cause with the same one-line fix, so pinning it would buy a second test of the same mechanism at the cost of `sudo` in the suite.

One note, no action needed: `assert secret.read_text() == "PRIVATE KEY\n"` reads as the no-dereference assertion but doesn't pin it — `u+rwX` on an already-`0644` target is a no-op, so the test would pass even if `chmod` did follow. The mode is what would move. I checked it separately and it holds.

shared/steps/token_usage.py
```diff
@@ -203,25 +204,115 @@ def codex_step(model: str) -> tuple[dict[str, Any], Path]:
...
+    """Remove every symlink from the dir about to be uploaded.
+
+    ``cp -a`` copies links as links and ``upload-artifact`` resolves them as
+    the runner — which can read the proxy's CA private key and its own
+    ``/proc/<pid>/environ``. The agent writes only regular files here, so a
+    link is it reaching for something else.
+    """
+    for path in list(logs_dir.rglob("*")):
+        if path.is_symlink():
+            path.unlink()
```
tend-agent: `cp -a` preserves FIFOs and sockets as faithfully as it preserves links, and this tests only for a link — so a `mkfifo` in the session dir rides the copy out. Verified through `consolidate_logs` on this branch: `project/pipe` lands in the copy beside `project/session.jsonl`.

upload-artifact then reads it. `getUploadZipSpecification` takes every entry whose `lstat` isn't a directory and hands `sourcePath` to `zip.file(...)`, which opens it with `createReadStream` — and opening a FIFO with no writer blocks. So the `if: always()` upload hangs until the job timeout: no artifact, and the runner burns the rest of its budget, after the agent is already reaped. The agent owns `.claude/projects` and `mkfifo` needs no privilege, so it picks whether that happens.

The docstring already names the invariant that closes it — "the agent writes only regular files here" — so checking for that directly covers links, FIFOs and sockets in one condition. `stat` is imported already, for `safe_agent_directory`. Tests stay green with this, and the FIFO no longer reaches the copy.

```suggestion
    """Remove every entry that is not a regular file or a directory.

    ``cp -a`` copies links as links and ``upload-artifact`` resolves them as
    the runner — which can read the proxy's CA private key and its own
    ``/proc/<pid>/environ``. It preserves FIFOs and sockets too, and the
    upload streams any non-directory entry, so one of those blocks it
    instead. The agent writes only regular files here, so anything else is
    it reaching for something other than its own session log.
    """
    for path in list(logs_dir.rglob("*")):
        mode = path.lstat().st_mode
        if not (stat.S_ISREG(mode) or stat.S_ISDIR(mode)):
            path.unlink()
```

tend-agent (commented):
`chmod -R u+rwX` checks out. I reproduced the `0555` case against the branch with it removed — `drop_symlinks` raises `PermissionError` and `leak.pem` survives in the copy — and confirmed the chmod can't be aimed anywhere: `chmod -R` leaves a nested link's target at its original mode. Same for the `chown -R` above it, which acts on the link rather than what it points at (coreutils refuses `-R --dereference` without `-H`/`-L`).

One thing the mode fix puts in relief, inline: `drop_symlinks` is narrower than the invariant its own docstring states.

Separately, outside the diff hunks by a line: `test_reconstructs_a_cancelled_session`'s docstring in `shared/steps/test_token_usage.py` still carries the claim this PR corrects one file over —

```
`tend-review` runs with `cancel-in-progress: true`, so cancellation is
routine
```

`review.yaml.j2` sets `false`, `triage.yaml.j2` is the `true` one. Happy to push that one-line correction here.

shared/steps/token_usage.py
max-sixty: Taken, in 77206bdc. Reproduced first: through `consolidate_logs`, `project/pipe` lands in the copy beside `project/session.jsonl`, and `is_symlink()` reads `False` on it, so the old sweep had no view of it at all.

Testing the mode is the right shape for the reason you give — the docstring already stated the invariant, and the code was checking something narrower than what it claimed. The function is now `drop_non_regular_files`, since it no longer only drops symlinks.

The `0644` aside is right too: `u+rwX` on a file that already has `u+rw` is a no-op, so that assertion never had a failing case. What it does pin is that the copied file is still readable and its contents intact, which is what the test is otherwise about — so I've left it as that rather than dressing it up as a no-dereference check. The no-dereference property is load-bearing for the `chmod` line and isn't pinned by any test; it holds because `chmod -R` doesn't traverse or dereference symlinks, and it's stated in the docstring above the call.

Also fixed the same wrong workflow name in `test_token_usage.py` and `test_run_claude.py` — my last commit corrected `token_usage.py` and I didn't grep for the rest.

tend-agent: Confirmed on 77206bdc. I restored the symlink-only check against the branch and the new test fails on exactly the entry it names (`assert not (logs_dir / "project" / "pipe").exists()`); with the mode test, 202 pass across `shared/steps/`. Exercised [`drop_non_regular_files`](https://github.com/max-sixty/tend/blob/77206bdcf8f63c5dee54401eaf3c0fe3369c6527/shared/steps/token_usage.py#L299) directly on a tree holding a FIFO, a unix socket, a symlink to a file, a symlink to a *directory* and a dangling one: all five go, `session.jsonl` stays, and the tree they point at is untouched. The symlinked directory is the one I wanted to see — `rglob` doesn't descend a symlinked dir, so it's unlinked rather than walked into, which is what makes "one test on the mode" hold for links to directories too and not just to files. `drop_symlinks` has no references left anywhere, and both docstring corrections match `triage.yaml.j2`.

Agreed on the `0644` assertion — pinning that the copied file is readable with its contents intact is what the test is about, and the no-dereference property is `chmod -R`'s own semantics rather than something a test could pin without asserting on coreutils.

One last variant of the same shape, and it closes outside this code rather than in it: `cp -a` runs as root here, and a *hard* link in the session dir is copied as content, so it lands in the artifact as a regular file that the mode test can't distinguish from a session log. The kernel closes it — `fs.protected_hardlinks` is `1` on the runner, so the sandbox user can only link to a file it can already read and write, and the proxy's confdir is [`0700` runner-only](https://github.com/max-sixty/tend/blob/77206bdcf8f63c5dee54401eaf3c0fe3369c6527/proxy/setup-sandbox.sh#L422-L425). Nothing to do; noting it because it's the one remaining way `cp -a` reproduces something the agent aimed.

docs/security-model.md
```diff
@@ -340,6 +340,14 @@ made as the sandbox user with `sandbox_setup:`. A generic failure shim keeps a
 dropped home-selected command from silently falling through to a different
 same-named system tool.
 
+**Session-log upload.** The token-usage step uploads the agent's session JSONL
+as an artifact, so the runner has to read a tree the sandbox user owns. It
+copies as root and chowns the result rather than making the source readable,
+since a `chmod -R` aimed through a symlink the agent planted would grant read on
+whatever tree it named. Both the session directory and the dot-directory above
+it are refused if either is a symlink, and links inside the copy are deleted
+before `upload-artifact` resolves them as the runner.
```
tend-agent: This paragraph lands in the same PR that widened the sweep past symlinks, and it still reads as the narrower guarantee. It matters more here than in a docstring: the security model is where someone auditing this boundary looks to find out whether a `mkfifo` in the session dir is handled, and the FIFO's outcome is a different one from the symlink's — not a leak the runner resolves, but an `if: always()` upload that blocks until the job timeout, with the agent choosing whether it happens.

```suggestion
it are refused if either is a symlink, and every entry in the copy that is not
a regular file or a directory is deleted before `upload-artifact` reads it — a
symlink it would otherwise resolve as the runner, a FIFO it would block on
until the job times out.
```

tend-agent (commented):
Testing the mode is the right shape — the sweep now checks the invariant its own docstring already named, rather than one type of violation of it. I confirmed it's load-bearing: reverting the body to `if path.is_symlink()` and running the suite, `test_consolidation_drops_a_fifo_the_agent_planted` fails with `project/pipe` still sitting beside `project/session.jsonl` in the copy, so `cp -a` does reproduce the FIFO through `consolidate_logs`. The three docstring corrections check out too — `triage.yaml.j2` is the only template in `generator/src/tend/templates/` that sets `cancel-in-progress: true`, and no other `tend-review` claim of it remains outside the skills' own accurate description of the queue.

One inline, on the `security-model.md` paragraph this PR adds — it still describes the sweep as symlink-only, a commit after the sweep stopped being that.

Nothing actionable on the one shape that still passes a regular-file test: a hard link the agent plants would be copied as content, not as a link, but `fs.protected_hardlinks` is `1` on the runner image, so it can only link a file it could already read and copy itself.

docs/security-model.md
max-sixty: Right, and thanks for catching it — I wrote that paragraph with the first commit and never revisited it as the branch grew. The security model is exactly where the gap costs something.

Rewritten in d7002a9c, a bit past your suggestion. The paragraph was stale on two more counts than the sweep: it also predates the reap gate and the mode reset, so an auditor reading it would have seen one of the four things that now have to hold. It now leads with the reason all four exist — the agent chooses the input to each — and states the FIFO's outcome as the block it is rather than folding it in with the symlink's leak.

On the hard link: agreed, nothing to do. `fs.protected_hardlinks=1` means the agent can only link a file it could already open, so the copy carries out content it could have written into the session dir itself.

tend-agent (approved):

PR #1152, opened 2026-09-06
fix(enrich-tend-outage-issues): read the job log when annotations name no cause

### LINKED ISSUE #1151: enrich-tend-outage-issues: a job that fails by plain non-zero exit is written up as unenrichable, though its log carries the failure
`enrich-tend-outage-issues.sh` reads only check-run **annotations**, and a job that fails by a plain non-zero exit produces exactly one annotation — `Process completed with exit code 1.` — which the script filters out by design. The `MSG` is then empty, so the run is written up as `No failure details could be extracted.` and marked `<!-- enriched-run:ID -->` so it is never retried. That is the shape of nearly every test suite: pytest, vitest, cargo, go test. The tracker ends up saying nothing is knowable about precisely the runs `ci-fix` exists to act on, while the failing line sits in the job log.

## Evidence

`max-sixty/leaf`'s open tracker [#305](https://github.com/max-sixty/leaf/issues/305) carries 20 enriched runs. The 15 `tend-*` rows enriched fine — the harness writes a real annotation on an agent failure. All **5** `ci` rows came back as stubs:

| Run | Failing job | Failing step | Annotations |
|---|---|---|---|
| [33981618265](https://github.com/max-sixty/leaf/actions/runs/33981618265) | `test` | `🧪 Nightly suite` | `Process completed with exit code 1.` only |
| [33983302539](https://github.com/max-sixty/leaf/actions/runs/33983302539) | `test` | `🧪 Nightly suite` | same |
| [33985319470](https://github.com/max-sixty/leaf/actions/runs/33985319470) | `test` | `🧪 Nightly suite` | same |
| [33987070790](https://github.com/max-sixty/leaf/actions/runs/33987070790) | `test` | `🧪 Nightly suite` | same |
| [33988570487](https://github.com/max-sixty/leaf/actions/runs/33988570487) | `test` | `🧪 Nightly suite` | same |

Verified directly:

```
$ gh api repos/max-sixty/leaf/check-runs/101349874026/annotations \
    --jq '.[] | {level: .annotation_level, message}'
{"level":"failure","message":"Process completed with exit code 1."}
```

Yet the log for that same run carries the whole diagnosis:

```
FAILED tests/test_render_navigation.py::test_holding_a_key_repeats_only_where_the_press_is_a_walk - AssertionError: Locator expected to have count '1'
Actual value: 0
============ 1 failed, 2021 passed, 6 skipped in 2053.74s (0:34:13) ============
```

This is the second window in which `max-sixty/leaf`'s `review-runs` log has recorded the stub against a run whose log does carry the failure — the first was [33927078970](https://github.com/max-sixty/leaf/actions/runs/33927078970) on 2026-09-05, a `curl: (35) Recv failure` inside the composite action, which renders as `UNKNOWN STEP` and likewise leaves no usable annotation. Six occurrences, one mechanism: **annotations are not where a failure's detail lives.**

## Proposed fix

Keep annotations as the primary reading — they are precise and cheap — and add a log fallback when they yield nothing. The last `##[error]` line is a reliable anchor: everything above it is the tool's own output, everything below is post-job cleanup, so a fixed window ending at that line is the diagnosis for both shapes above.

```bash
if [ ! -s /tmp/enrich-errors.md ]; then
  LOG=$(gh run view "$RUN_ID" --repo "$REPO" --log-failed 2>/dev/null || true)
  # Anchor on the terminal `##[error]`: tool output above it, runner cleanup below.
  END=$(printf '%s\n' "$LOG" | grep -n '##\[error\]' | tail -1 | cut -d: -f1)
  if [ -n "$END" ]; then
    START=$(( END > 30 ? END - 30 : 1 ))
    # Strip the `job<TAB>step<TAB>timestamp ` prefix `gh run view --log` adds.
    printf '%s\n' "$LOG" | sed -n "${START},${END}p" \
      | sed 's/^[^\t]*\t[^\t]*\t[0-9T:.Z-]*Z //' > /tmp/enrich-tail.md
    printf '#### log tail\n\n```\n%s\n```\n\n' "$(cat /tmp/enrich-tail.md)" \
      >> /tmp/enrich-errors.md
  fi
fi
```

Run against 33981618265 this yields the failing assertion, its call log, the pytest summary line and the exit — 31 lines, well inside the per-issue batch the script already builds. `No failure details could be extracted.` then means what it says: a genuinely unenrichable run, which the session-ceiling kills recorded on [max-sixty/leaf#206](https://github.com/max-sixty/leaf/issues/206) really are, since those die before any step fails.

Two notes for whoever picks this up:

- `--log-failed` is one call per **run**, not per job, so it belongs outside the `while IFS=$'\t' read -r JOB_ID` loop rather than inside it.
- A 30-line window on a fanned-out matrix could still add up; capping the fallback at the first two failing runs per batch, or truncating the block, would keep the comment readable.

Filed from `max-sixty/leaf`'s `tend-review-runs` sweep ([run 34020114452](https://github.com/max-sixty/leaf/actions/runs/34020114452)).

tend-agent:
## Problem

`enrich-tend-outage-issues.sh` reads only check-run annotations. A job that fails by a plain non-zero exit produces exactly one annotation — `Process completed with exit code 1.` — which the script filters out by design, because it names no cause. `MSG` is then empty, the run is written up as `No failure details could be extracted.`, and it is marked `<!-- enriched-run:ID -->` so no later night retries it. That is the shape of every test suite (pytest, vitest, cargo, go test), so the tracker said nothing was knowable about precisely the runs `ci-fix` exists to act on. Reported in #1151 with six occurrences across two failure shapes.

## Solution

Keep annotations as the primary reading — precise, one cheap call — and fall back to the failed job's log only when they yield nothing. The last `##[error]` line is the anchor: the tool's own output sits above it, runner cleanup below, so a 30-line window ending there is the diagnosis. `gh run view --log-failed` is one call per run rather than per job, so it sits outside the per-job loop. `No failure details could be extracted.` now means what it says.

The log API also returns colour escapes as the literal characters `^[[42m` rather than an ESC byte, and the first line of each step carries a byte-order mark ahead of its timestamp, so both are stripped alongside the timestamp.

## Testing

New `generator/tests/test_enrich_outage_issues.py`, following the repo's one-file-per-script convention with a fake `gh`. Five of its eight cases fail on the pre-fix script and pass after — the two the fallback exists for, plus colour escapes, the step-boundary byte-order mark, and a 91,290-char batch that the truncation brings back under the limit; the other three pin the behaviour the fallback must not disturb — annotations still win (and the log is not fetched), a run with neither source still records its marker rather than being retried nightly forever, and an unavailable log doesn't take the whole `set -e` batch down.

Also verified against the two real runs the issue cites, reading their logs through the same pipeline:

<details><summary>Real-run verification</summary>

`max-sixty/leaf` run 33981618265 — a 4572-line log reduced to the 31-line window:

````
FAILED tests/test_render_navigation.py::test_holding_a_key_repeats_only_where_the_press_is_a_walk - AssertionError: Locator expected to have count '1'
Actual value: 0
Call log:
  - Expect "to_have_count" with timeout 5000ms
  - waiting for locator("#live-question-decision[data-lf-decision]")
    14 × locator resolved to 0 elements
============ 1 failed, 2021 passed, 6 skipped in 2053.74s (0:34:13) ============
````

Run 33927078970 — the `UNKNOWN STEP` shape, where the failure is inside the composite action rather than a test suite:

````
curl: (35) Recv failure: Connection reset by peer
##[error]Process completed with exit code 35.
````

Full suite: 943 passed. All 13 pre-commit hooks pass, shellcheck included.

</details>

Two judgement calls worth a reviewer's eye:

- The log block uses a four-backtick fence, unlike the three-backtick annotation block above it. A CI log is far more likely than an annotation to contain a line that is itself a fence, which would end the block early and garble the comment.
- The batch stops between runs once it passes 30 KB, so every posted section keeps its fences and its `<!-- enriched-run: -->` marker; a byte-offset cut lands mid-section and can post a body with no marker at all, which the next night rebuilds identically. My first pass declined the issue's cap as an efficiency knob; review showed it isn't one. The largest batch already posted (18 runs on #831) is 11,109 chars under the annotation-only path, and a measured log tail is ~1,759 chars per run — so the same batch now lands between ~43 KB and, at the `cut -b -500` bound, past the 65536-char comment limit. Over it `gh issue comment` returns 422, which under `set -euo pipefail` aborts the pass; no marker is written for any run in the batch, so the next night rebuilds the identical batch and fails the same way. Cutting between runs keeps the markers that did land, so each night's batch is smaller than the last.
- One run has to fit the headroom that cap leaves, so the annotation `MSG` now runs through `head -n 30 | cut -b -500` — the window the log tail already carried, and the bound a linter that annotates per finding needs — and the job loop stops once a run's sections pass 15 KB, noting the omitted jobs. Worst case is ~60 KB. The flag is `-b` rather than `-c` because both caps measure with `wc -c` and GNU `cut` implements `-c` as `-b` today — the bound is already in bytes, so the flag and the comments now say so, and a `cut` that later honours `-c` as characters can't quietly quadruple it.

---
Closes #1151 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
+++ b/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
@@ -48,7 +48,7 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
         # Stop between runs instead, so every posted section keeps its fences
         # and its marker; the runs that fall off here carry no marker and are
         # enriched by a later, smaller batch. A run's own section is bounded
-        # above (30 lines x 500 characters per job, at most ~15 KB of jobs),
+        # above (30 lines x 500 bytes per job, at most ~15 KB of jobs),
         # so the finished body cannot exceed roughly 60 KB.
         if [ "$(wc -c < /tmp/enrich-batch.md)" -gt 30000 ]; then
           printf '_Truncated; the remaining runs are enriched by a later batch._\n' \
@@ -71,11 +71,11 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
             break
           fi
           # A linter annotates per finding, so bound the message the same way
-          # the log tail below is bounded: 30 lines of at most 500 characters.
+          # the log tail below is bounded: 30 lines of at most 500 bytes.
           MSG=$(gh api "repos/$REPO/check-runs/$JOB_ID/annotations" \
             --jq '[.[] | select(.annotation_level == "failure") | .message
                   | select(test("^Process completed") | not)] | join("\n\n")' \
-            2>/dev/null | head -n 30 | cut -c -500 || true)
+            2>/dev/null | head -n 30 | cut -b -500 || true)
           [ -n "$MSG" ] && printf '#### %s\n\n```\n%s\n```\n\n' "$JOB_NAME" "$MSG" \
             >> /tmp/enrich-errors.md
         done <<< "$JOBS"
@@ -105,7 +105,7 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
               sed -n "${START},${END}p" /tmp/enrich-log.txt \
                 | cut -f3- \
                 | sed 's/^\xef\xbb\xbf//; s/^[0-9T:.Z-]*Z //; s/\^\[\[[0-9;]*[A-Za-z]//g' \
-                | cut -c -500
+                | cut -b -500
               printf '````\n\n'
             } >> /tmp/enrich-errors.md
           fi
--- a/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
+++ b/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
@@ -57,6 +57,32 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
             >> /tmp/enrich-errors.md
         done <<< "$JOBS"
 
+        # A job that fails by a plain non-zero exit — every test suite — writes
+        # one annotation, `Process completed with exit code 1.`, which the
+        # filter above drops because it names no cause. Its diagnosis is in the
+        # log instead, so read the log's tail before giving up. One call per
+        # run, not per job, hence outside the loop above.
+        if [ ! -s /tmp/enrich-errors.md ]; then
+          gh run view "$RUN_ID" --repo "$REPO" --log-failed \
+            > /tmp/enrich-log.txt 2>/dev/null || true
+          # The last `##[error]` is the anchor: the tool's own output is above
+          # it, runner cleanup below.
+          END=$(grep -n '##\[error\]' /tmp/enrich-log.txt | tail -1 | cut -d: -f1) || true
+          if [ -n "$END" ]; then
+            START=$(( END > 30 ? END - 30 : 1 ))
+            {
+              # A four-backtick fence so a log line that is itself a fence
+              # doesn't end the block early.
+              printf '#### log tail\n\n````\n'
+              # Drop the `job<TAB>step<TAB>` prefix and the timestamp
+              # `--log-failed` adds, then bound a pathological single line.
+              sed -n "${START},${END}p" /tmp/enrich-log.txt \
+                | cut -f3- | sed 's/^[0-9T:.Z-]*Z //' | cut -c -500
+              printf '````\n\n'
+            } >> /tmp/enrich-errors.md
+          fi
+        fi
+
         RUN_URL="https://github.com/$REPO/actions/runs/$RUN_ID"
         if [ -s /tmp/enrich-errors.md ]; then
           {
--- a/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
+++ b/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
@@ -74,10 +74,15 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
               # A four-backtick fence so a log line that is itself a fence
               # doesn't end the block early.
               printf '#### log tail\n\n````\n'
-              # Drop the `job<TAB>step<TAB>` prefix and the timestamp
-              # `--log-failed` adds, then bound a pathological single line.
+              # Drop the `job<TAB>step<TAB>` prefix, the byte-order mark the
+              # first line of each step carries ahead of its timestamp, the
+              # timestamp itself, and the colour escapes — which the log API
+              # hands back as the literal characters `^[[42m`, so nothing
+              # downstream renders them. Then bound a pathological single line.
               sed -n "${START},${END}p" /tmp/enrich-log.txt \
-                | cut -f3- | sed 's/^[0-9T:.Z-]*Z //' | cut -c -500
+                | cut -f3- \
+                | sed 's/^\xef\xbb\xbf//; s/^[0-9T:.Z-]*Z //; s/\^\[\[[0-9;]*[A-Za-z]//g' \
+                | cut -c -500
               printf '````\n\n'
             } >> /tmp/enrich-errors.md
           fi
@@ -104,6 +109,21 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
         fi
       done
 
-    [ -s /tmp/enrich-batch.md ] \
-      && gh issue comment "$ISSUE" --repo "$REPO" -F /tmp/enrich-batch.md
+    # GitHub rejects a comment body over 65536 characters with a 422. Under
+    # `set -e` that rejection takes the whole nightly pass down — and because
+    # the `enriched-run` markers land only with the comment, the next night
+    # rebuilds the identical oversized batch and fails the same way. Truncate instead: the runs that fall off
+    # keep no marker and are enriched by a later, smaller batch.
+    if [ -s /tmp/enrich-batch.md ]; then
+      if [ "$(wc -c < /tmp/enrich-batch.md)" -gt 60000 ]; then
+        # `sed '$d'` drops the line `head -c` cut mid-way, so the body can't end
+        # on half a UTF-8 character.
+        head -c 60000 /tmp/enrich-batch.md | sed '$d' > /tmp/enrich-post.md
+        printf '\n_Truncated; the remaining runs are enriched by a later batch._\n' \
+          >> /tmp/enrich-post.md
+      else
+        cp /tmp/enrich-batch.md /tmp/enrich-post.md
+      fi
+      gh issue comment "$ISSUE" --repo "$REPO" -F /tmp/enrich-post.md
+    fi
   done
--- a/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
+++ b/plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
@@ -41,6 +41,21 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
       | while read -r RUN_ID; do
         [ -z "$RUN_ID" ] && continue
 
+        # GitHub rejects a comment body over 65536 characters with a 422, and
+        # under `set -e` that rejection takes the whole nightly pass down —
+        # then, because the `enriched-run` markers land only with the comment,
+        # the next night rebuilds the identical batch and fails identically.
+        # Stop between runs instead, so every posted section keeps its fences
+        # and its marker; the runs that fall off here carry no marker and are
+        # enriched by a later, smaller batch. A run's own section is bounded
+        # above (30 lines x 500 characters per job, at most ~15 KB of jobs),
+        # so the finished body cannot exceed roughly 60 KB.
+        if [ "$(wc -c < /tmp/enrich-batch.md)" -gt 30000 ]; then
+          printf '_Truncated; the remaining runs are enriched by a later batch._\n' \
+            >> /tmp/enrich-batch.md
+          break
+        fi
+
         : > /tmp/enrich-errors.md
         # Capture jobs first so a 404 (deleted/expired run) doesn't trip
         # `set -e` via the pipe's exit status.
@@ -49,10 +64,18 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
           2>/dev/null || true)
         while IFS=$'\t' read -r JOB_ID JOB_NAME; do
           [ -z "$JOB_ID" ] && continue
+          # One run's own section has to fit the headroom the batch cap below
+          # leaves, and a matrix run's failed jobs are otherwise unbounded.
+          if [ "$(wc -c < /tmp/enrich-errors.md)" -gt 15000 ]; then
+            printf '_Remaining failed jobs omitted._\n\n' >> /tmp/enrich-errors.md
+            break
+          fi
+          # A linter annotates per finding, so bound the message the same way
+          # the log tail below is bounded: 30 lines of at most 500 characters.
           MSG=$(gh api "repos/$REPO/check-runs/$JOB_ID/annotations" \
             --jq '[.[] | select(.annotation_level == "failure") | .message
                   | select(test("^Process completed") | not)] | join("\n\n")' \
-            2>/dev/null || true)
+            2>/dev/null | head -n 30 | cut -c -500 || true)
           [ -n "$MSG" ] && printf '#### %s\n\n```\n%s\n```\n\n' "$JOB_NAME" "$MSG" \
             >> /tmp/enrich-errors.md
         done <<< "$JOBS"
@@ -109,21 +132,10 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
         fi
       done
 
-    # GitHub rejects a comment body over 65536 characters with a 422. Under
-    # `set -e` that rejection takes the whole nightly pass down — and because
-    # the `enriched-run` markers land only with the comment, the next night
-    # rebuilds the identical oversized batch and fails the same way. Truncate instead: the runs that fall off
-    # keep no marker and are enriched by a later, smaller batch.
+    # An `if` rather than `[ -s ... ] && gh ...`: a false test there is the last
+    # command of the loop body, so an issue whose runs are all already enriched
+    # would end the pipeline non-zero and take the pass down under `set -e`.
     if [ -s /tmp/enrich-batch.md ]; then
-      if [ "$(wc -c < /tmp/enrich-batch.md)" -gt 60000 ]; then
-        # `sed '$d'` drops the line `head -c` cut mid-way, so the body can't end
-        # on half a UTF-8 character.
-        head -c 60000 /tmp/enrich-batch.md | sed '$d' > /tmp/enrich-post.md
-        printf '\n_Truncated; the remaining runs are enriched by a later batch._\n' \
-          >> /tmp/enrich-post.md
-      else
-        cp /tmp/enrich-batch.md /tmp/enrich-post.md
-      fi
-      gh issue comment "$ISSUE" --repo "$REPO" -F /tmp/enrich-post.md
+      gh issue comment "$ISSUE" --repo "$REPO" -F /tmp/enrich-batch.md
     fi
   done
--- a/generator/tests/test_enrich_outage_issues.py
+++ b/generator/tests/test_enrich_outage_issues.py
@@ -1,0 +1,177 @@
+"""Tests for plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh.
+
+The script turns a bare run link in a `tend-outage` issue into a readable
+failure, then marks the run `<!-- enriched-run:ID -->` so no later night
+retries it. That marker makes an empty enrichment permanent, so the cases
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh [hunk 1]
```diff
@@ -57,6 +57,32 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
...
+          if [ -n "$END" ]; then
+            START=$(( END > 30 ? END - 30 : 1 ))
+            {
+              # A four-backtick fence so a log line that is itself a fence
+              # doesn't end the block early.
+              printf '#### log tail\n\n````\n'
+              # Drop the `job<TAB>step<TAB>` prefix and the timestamp
+              # `--log-failed` adds, then bound a pathological single line.
+              sed -n "${START},${END}p" /tmp/enrich-log.txt \
+                | cut -f3- | sed 's/^[0-9T:.Z-]*Z //' | cut -c -500
```
tend-agent: Terminal colour codes survive this pipeline and land verbatim in the posted comment. GitHub's log API hands back the escape as the literal two characters `^[` rather than a real ESC byte, so nothing downstream renders it — on run 34021152770 the same window comes through as `shellcheck....^[[42mPassed^[[m`. Any tool that forces colour under CI (pre-commit, cargo, npm) produces this, and the whole point of the fallback is a readable diagnosis.

```suggestion
                | cut -f3- | sed 's/^[0-9T:.Z-]*Z //; s/\^\[\[[0-9;]*m//g' | cut -c -500
```

Separately, and much smaller: the first line of each step's log carries a UTF-8 BOM ahead of the timestamp, so the anchored strip misses it there. It only shows when the window spans a step boundary — i.e. when the failing step produced fewer than 31 lines, like the `curl: (35)` case in the description.

tend-agent (commented):
The fallback works — I ran the real pipeline over run 34021152770's log and it lands the right 31-line window. Full suite (940 passed) and all 13 pre-commit hooks pass here too, and reverting just the script hunk fails exactly the two tests the description claims.

One thing I'd push back on: the declined per-batch cap isn't an efficiency knob, it's the difference between the enrichment step working and being permanently stuck. Issue #831 already carries a single comment covering 18 runs at 11,109 chars — ~617 chars per run under the old annotation-only path. The measured log tail for a real run is ~1,759 chars, and `cut -c -500` bounds it at ~15 KB, so the same 18-run batch now lands somewhere between ~43 KB and well past GitHub's 65,536-char comment limit. Over the limit `gh issue comment` returns 422, and it's the right-hand side of `[ -s /tmp/enrich-batch.md ] && gh issue comment …` under `set -euo pipefail` — the failure exits the subshell, so the pass aborts and every remaining tracker issue that night goes unprocessed. No marker was written for any run in the batch, so the next night rebuilds the identical oversized batch and fails the same way; nothing in the loop makes it smaller.

The remedy is a knob, not machinery — `head -c 60000` on the batch file before the post, or dropping the log fallback once the batch passes a size. Worth doing in this PR rather than separately, since this is the change that moves the largest observed batch from ~11 KB to within striking distance of the limit.

<details><summary>How the numbers were measured</summary>

```bash
# largest batch already posted
gh issue view 831 --json comments --jq '[.comments[] | {len:(.body|length), runs:([.body|scan("<!-- enriched-run:")]|length)}] | map(select(.runs>0)) | sort_by(-.runs) | .[0]'
# => {"len":11109,"runs":18}

# the new fallback's window on a real run, through the script's own pipeline
gh run view 34021152770 --log-failed > /tmp/rl.txt
END=$(grep -n '##\[error\]' /tmp/rl.txt | tail -1 | cut -d: -f1); START=$(( END > 30 ? END - 30 : 1 ))
sed -n "${START},${END}p" /tmp/rl.txt | cut -f3- | sed 's/^[0-9T:.Z-]*Z //' | cut -c -500 | wc -c
# => 1759
```

Refuted while checking: `cut -c -500` does not split UTF-8 mid-character (the runner's `C.UTF-8` locale makes `-c` character-aware, verified), `grep`'s no-match exit is covered by the `|| true`, and the `>` redirect truncates `/tmp/enrich-log.txt` before `gh` runs, so a failed fetch can't serve the previous run's log.

</details>

plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
tend-agent: Confirmed both against run 34021152770's log: 39 literal `^[` sequences, zero ESC bytes, and the mark sits between the `job<TAB>step<TAB>` prefix and the timestamp on the first line of each step, so `cut -f3-` hands the anchored strip a line that starts with the BOM.

Taken in 414443a, with the escape pattern one character wider than the suggestion — the final byte is `[A-Za-z]` rather than `m`, so an erase-line `^[[2K` from a progress bar goes too. The BOM comes off in the same sed, ahead of the timestamp anchor, and both run before `cut -c -500` so the 500 chars are content rather than escapes.

```bash
gh run view 34021152770 --repo max-sixty/tend --log-failed > /tmp/rl.txt
grep -c $'\x1b' /tmp/rl.txt   # => 0 (no real ESC byte)
grep -c '\^\[' /tmp/rl.txt    # => 39 (literal caret-bracket)
grep -o '\^\[\[[0-9;]*[A-Za-z]' /tmp/rl.txt | sort | uniq -c
# => ^[[m ×42, ^[[42m ×12, ^[[32m ×9, ^[[1m ×4, ^[[36;1m ×3, ^[[0m ×3, …
head -1 /tmp/rl.txt | head -c 40 | xxd
# => …action@v3.0.1<TAB>ef bb bf 32 30 32  ("...1\t<BOM>202")
```

Through the new pipeline the same window renders as `shellcheck...Passed` with no escapes left. `test_terminal_colour_escapes_are_stripped` and `test_the_step_boundary_bom_does_not_defeat_the_timestamp_strip` both fail on the pre-fix script.

</details>

tend-agent:
The pushback lands — I classified it as an efficiency knob and it isn't. A 422 there is a jam, not waste: no marker is written for any run in the batch, so the next night rebuilds the identical batch and fails identically, and every later tracker issue in that pass goes unprocessed with it.

Fixed in 414443a, though as a bound on the body rather than a cap on the run count. The batch file is truncated at 60000 bytes before the post, on a line boundary (`sed '$d'` drops the line `head -c` cut mid-way, so the body can't end on half a UTF-8 character), with a one-line note that the rest follows. A run-count cap would have to guess a per-run size, and the annotation path has no bound in this script either — an unusually long annotation message would still make an oversized body. Truncating holds regardless of what a run contributes, and the markers that did land make the next night's batch smaller, so the enrichment always makes progress.

`test_an_oversized_batch_is_truncated_rather_than_rejected` builds six runs from the same 30-line window: 91,290 chars pre-fix, under the limit after, with the first run's marker posted and the last one's carried to a later batch.

plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
```diff
@@ -78,6 +109,21 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
...
+    # GitHub rejects a comment body over 65536 characters with a 422. Under
+    # `set -e` that rejection takes the whole nightly pass down — and because
+    # the `enriched-run` markers land only with the comment, the next night
+    # rebuilds the identical oversized batch and fails the same way. Truncate instead: the runs that fall off
+    # keep no marker and are enriched by a later, smaller batch.
+    if [ -s /tmp/enrich-batch.md ]; then
+      if [ "$(wc -c < /tmp/enrich-batch.md)" -gt 60000 ]; then
+        # `sed '$d'` drops the line `head -c` cut mid-way, so the body can't end
+        # on half a UTF-8 character.
+        head -c 60000 /tmp/enrich-batch.md | sed '$d' > /tmp/enrich-post.md
```
tend-agent: `head -c 60000 | sed '$d'` drops the whole line the byte cut landed in, and a run's `<!-- enriched-run: -->` marker sits at the end of its section — so when one run's section is itself over the cap, nothing survives the cut and the posted body carries no marker at all. Nothing is recorded, the next night rebuilds the identical batch, and posts the identical comment. That is the 422 jam again, with a nightly comment attached instead of silence.

The log tail can't reach that size (31 lines through `cut -c -500`), but the annotation path is unbounded: `MSG` joins every failure annotation of every failed job in the run. So it's the same asymmetry the cap was measured against, moved one level down — a matrix run with several failed jobs, or one long annotation, gets there.

Running the script with a single 70,000-char annotation posts this 158-char body and nothing else, every night:

````
### [Run 33981618265](https://github.com/owner/repo/actions/runs/33981618265)

#### test

```
_Truncated; the remaining runs are enriched by a later batch._
````

Smaller, same cause: the cut lands mid-fence. Six 15 KB runs give three complete sections plus a partial fourth, so the body ends on an unclosed four-backtick fence and the `_Truncated_` note renders inside the code block.

Cutting at the last complete `<!-- enriched-run:` line instead of at an arbitrary byte fixes the fence and keeps every posted section whole. It wants a companion bound on the per-run section — the annotation `MSG`, mirroring the log path's `cut -c -500` — so that line always exists; without one, `head -n 0` posts an empty body and 422s on its own.

**Review by tend-agent — COMMENTED** [review]
The log fallback holds up. I re-ran the pipeline against a real `--log-failed` payload and the three strips clear everything they claim to: 39 literal `^[` sequences and zero ESC bytes, and the byte-order mark does sit between the `job<TAB>step<TAB>` prefix and the timestamp on each step's first line. Full file suite and shellcheck pass here.

The batch bound added in 414443a has a hole, and it is the same shape as the jam it replaces — detail inline.

File: plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh (hunk 2 above)
Comment (tend-agent) [reply plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh]: Confirmed, and the empty-body case reproduces exactly as described — a single 70,000-char annotation posted a section header, an unclosed fence and the note, with no marker anywhere in it.

Fixed in b39e178 by moving the cap into the run loop, ahead of each run rather than after the batch is built: every section that lands is whole, fences and marker included, and the note goes after the last complete one. The post-hoc `head -c`/`sed '$d'` block is gone.

Two bounds keep a single run from filling the body on its own, so the cut always has a whole section to keep:

- the annotation `MSG` now runs through `head -n 30 | cut -c -500`, the same window the log tail carries — a linter that annotates per finding is the shape that needs it;
- the job loop stops once a run's sections pass 15 KB, noting the omitted jobs, since a matrix contributes one section per failed leg.

So the worst case is 30 KB of completed runs plus one run's ~30 KB section — about 60 KB, under the 65536 limit with room to spare.

<details><summary>Tests</summary>

Three new cases, all failing on 414443a with precisely the shapes you named:

- `test_truncation_keeps_whole_sections` — 7 fence lines (odd, so one is unclosed) and the note rendered inside the code block;
- `test_one_huge_annotation_cannot_fill_the_body` — body was `### [Run …]` + `#### test` + an empty fence + the note, no marker;
- `test_a_matrix_of_failed_jobs_cannot_fill_the_body` — 40 failed jobs, same no-marker shape.

A fourth, `test_an_issue_with_nothing_new_posts_nothing_and_exits_clean`, pins the `if` around the final `gh issue comment`: a bare `[ -s … ] && gh …` is the last command of the loop body, so an issue whose runs are all already enriched ends the pipeline non-zero and takes the pass down under `set -e`. That is the form on `main`, and this PR happened to fix it in passing — worth a test so it isn't tidied back.

12 tests in the file, 947 in the suite, all 13 pre-commit hooks green.

</details>

File: plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh [hunk 3]
```diff
@@ -49,14 +64,53 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
...
+            printf '_Remaining failed jobs omitted._\n\n' >> /tmp/enrich-errors.md
+            break
+          fi
+          # A linter annotates per finding, so bound the message the same way
+          # the log tail below is bounded: 30 lines of at most 500 characters.
           MSG=$(gh api "repos/$REPO/check-runs/$JOB_ID/annotations" \
             --jq '[.[] | select(.annotation_level == "failure") | .message
                   | select(test("^Process completed") | not)] | join("\n\n")' \
-            2>/dev/null || true)
+            2>/dev/null | head -n 30 | cut -c -500 || true)
```
tend-agent: The annotation bound is what keeps one run's section inside the headroom the batch cap leaves, and that headroom is measured in bytes (`wc -c`). `-b` states the same 500 the line already does, in the unit the arithmetic uses, so a `cut` that one day honours `-c` as characters can't quietly quadruple it.

```suggestion
            2>/dev/null | head -n 30 | cut -b -500 || true)
```

plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
```diff
@@ -49,14 +64,53 @@ gh issue list --label "$LABEL" --state open --json number --jq '.[].number' \
...
+              printf '#### log tail\n\n````\n'
+              # Drop the `job<TAB>step<TAB>` prefix, the byte-order mark the
+              # first line of each step carries ahead of its timestamp, the
+              # timestamp itself, and the colour escapes — which the log API
+              # hands back as the literal characters `^[[42m`, so nothing
+              # downstream renders them. Then bound a pathological single line.
+              sed -n "${START},${END}p" /tmp/enrich-log.txt \
+                | cut -f3- \
+                | sed 's/^\xef\xbb\xbf//; s/^[0-9T:.Z-]*Z //; s/\^\[\[[0-9;]*[A-Za-z]//g' \
+                | cut -c -500
```
tend-agent: Same on the log tail — it is the other half of the ~60 KB worst case.

```suggestion
                | cut -b -500
```

tend-agent (commented):
The cap holds where the previous one didn't. Reverting just the script hunk reproduces all three shapes exactly as described — 7 fence lines (odd, so one unclosed) for `test_truncation_keeps_whole_sections`, and a marker-less section for the huge-annotation and 40-job-matrix cases — and the batch check sitting at the top of the loop with an empty file on the first iteration is what makes "every posted body carries at least one marker" a property rather than a hope. 947 passed, all 13 hooks green here.

One correction to the record, since a thread on this PR recorded the opposite as refuted: GNU `cut -c` is byte-based, not character-based. On this runner `printf '─────' | cut -c -5 | xxd` gives `e2 94 80 e2 94` — five bytes, with a dangling UTF-8 sequence — under `C.UTF-8` and `en_US.UTF-8` alike, because coreutils implements `-c` as `-b` ("this is the same as -b for now, but internationalization will change that").

That cuts two ways. The size arithmetic is *sounder* than the comment claims: the bound is 500 bytes per line, `wc -c` measures bytes, and GitHub's limit is 65536 characters, so bytes ≥ characters and the ~60 KB worst case has real headroom. But the comment describing it as "500 characters" is what a later edit reads — a coreutils that ships the promised i18n behaviour, or a self-hosted runner on a distro that patches `cut` for it, turns 500 characters into up to 2000 bytes per line and reopens the 422 jam. Spelling the flag `-b` makes the bound the one the arithmetic needs, and is a no-op on every GNU `cut` today (the 12 tests pass unchanged with it). Worth moving the two comments to "bytes" alongside.

The mid-character split itself is cosmetic — `gh` marshals the body through Go's JSON encoder, which coerces invalid UTF-8 to `U+FFFD`, so a truncated line ends in a replacement character rather than failing the post.

plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
tend-agent: Applied in fe5f8e0, and the comment above it now reads "500 bytes" so the unit matches the `wc -c` the cap is measured with. Confirmed on this runner that it's a no-op today — `printf '─────' | cut -c -5 | xxd` gives `e2 94 80 e2 94`, the same five bytes `-b` gives, on coreutils 9.4.

plugins/tend-ci-runner/scripts/enrich-tend-outage-issues.sh
tend-agent: Applied in fe5f8e0, with the "30 lines of at most 500 characters" comment moved to bytes alongside it. 12 tests pass unchanged, shellcheck included.

tend-agent:
Correction stands, and the record needed it: an earlier review on this PR listed "`cut -c` is character-aware, so it can't split UTF-8" under *refuted*, which is backwards. GNU coreutils implements `-c` as `-b` today (`printf '─────' | cut -c -5 | xxd` → `e2 94 80 e2 94` on this runner's coreutils 9.4, under `C.UTF-8` and `en_US.UTF-8` alike), so the bound was already 500 bytes while the flag and two comments said characters — and the caps around it measure with `wc -c`, so bytes is the unit the arithmetic needs.

fe5f8e0 spells both bounds `-b` and moves the comments to bytes. No behaviour change on any GNU `cut` shipping now; the 12 tests and shellcheck pass unchanged. What it buys is that a coreutils honouring the promised i18n behaviour — or a self-hosted runner patched for it — can't turn 500 characters into 2000 bytes per line and walk the worst case back over the 65536-character limit.

PR #1154, opened 2026-09-06
chore: bump codex_version to 0.153.4

tend-agent:
Moves `@openai/codex` from 0.151.0 to 0.153.4, npm's current `latest` dist-tag, in both places the CLI is pinned: `codex/action.yaml` (the harness, which also installs `@openai/codex-responses-api-proxy` at the same version) and `codex/refresh/action.yaml` (the subscription-auth rotation). Nothing in the range touches the four `codex exec` flags the action passes (`--model`, `--sandbox`, `--output-last-message`, `--config`), the `codex plugin marketplace add` / `codex plugin add` pair, or the `model_reasoning_effort` config key — and `test-codex-surface` asserts all of those against whatever `codex/action.yaml` pins.

The `model:` input description drops its enumeration of visible models rather than refreshing it. The visible set is server-side and per-account, so any static list in the description drifts whatever it is set to; the list that was there had already gone stale before this PR, and rewriting it would reset the clock rather than stop it — the same argument #1150 makes about the built-in review prompt. What replaces it is the part that is checkable from this repo: the action passes `--model` unconditionally ([`codex/runner.py:205`](https://github.com/max-sixty/tend/blob/cf3d54a2c063c616c9065459679bfbfbde0e3f80/codex/runner.py#L205)), so whatever Codex bundles as its no-model default never applies here. The description now points a reader at the `"visibility": "list"` entries in `~/.codex/models_cache.json` on the account CI runs as, instead of copying a snapshot of them.

**Unverified:** the weekly sweep asks for an isolated Plus/Pro login to run the refresh action and confirm both the full and access-only credentials rotate before a `codex_version` bump. No `OPENAI_API_KEY` or ChatGPT login reaches this repo's runs, so that check is unrun here and the release-note skim below is the only evidence for the auth path. Verification that did run: `uv run pytest` on the merged tree (1082 tests) and `pre-commit run --files codex/action.yaml codex/refresh/action.yaml`.

<details><summary>Release-note skim, 0.152.0 → 0.153.4</summary>

Seven releases, skimmed for model availability, sandbox behavior, and `--output-last-message`. This records what the release notes claim; none of it is verified against a live account from CI.

**Model availability.** 0.153.1 added `gpt-6-astra` behind the API without listing it; 0.153.2 corrected its Fast-tier description text; 0.153.3 added it to the Bedrock catalogs; 0.153.4 made it visible in the bundled picker and, per [openai/codex#42874](https://github.com/openai/codex/pull/42874), *"its existing priority makes Astra the bundled default when no model is explicitly configured."* Not reachable here — the runner passes `--model` unconditionally, and the generator supplies `gpt-5.5` as `DEFAULT_MODEL_BY_HARNESS["codex"]` when an adopter sets nothing. The generator deliberately keeps no Codex model allowlist, so an adopter wanting `gpt-6-astra` can already name it. Whether a given account actually sees Astra is an account-side question this run can't answer, which is why the input description no longer enumerates anything.

**Sandbox behavior.** Nothing in the range changes Linux sandbox semantics. 0.152.0's sandbox fix is Windows-only (Microsoft Store PowerShell for elevated commands). The action runs `danger-full-access` on the ubuntu runner image regardless, since it ships no bubblewrap.

**`--output-last-message`.** Untouched across the range; `test-codex-surface` re-asserts it takes a value.

**Defaults worth knowing.** 0.152.0 disabled the planning tool by default (`tools.update_plan.enabled = true` re-enables it) — the action sets no `tools.*` config and skills don't depend on `update_plan`. 0.153.0 added `tui.auto_recap` and moved `disable_paste_burst` under `tui.`; both are TUI-only and `codex exec` is non-interactive.

**Everything else** is TUI/Vim-mode work, Guardian approval-review internals, MCP tool scoping, app-server thread metadata, rate-limit banners, and `codex exec resume` on compressed rollouts — none of it on the one-shot `codex exec` path this action drives.

</details>


max-sixty:
The `codex_version` bump is fine. The model list rewritten in the same commit doesn't match Codex's model cache.

`~/.codex/models_cache.json` (client 0.152.0, fetched 2026-09-07T00:55Z) has six entries with `"visibility": "list"`:

```
gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna, gpt-5.5, gpt-5.4-mini, gpt-5.3-codex-spark
```

`gpt-reserve` and `codex-auto-review` are the only others, both `"hide"`.

The new description names `gpt-6-astra`, `gpt-5.4`, `gpt-5.3-codex` and `gpt-5.2`, none of which appear there, and omits the three `gpt-5.6-*` models. The input's default, `gpt-5.5`, is still visible, so the default itself is unaffected.

I have not verified the separate claim that 0.153.4 makes `gpt-6-astra` the bundled default when no model is configured; that sentence may hold even though the enumeration doesn't.

Two things to weigh before rewriting the list. A model cache is per account, so check it against the account CI runs as rather than taking the list above. And the enumeration copies a server-side list into a static description, so it drifts again whatever it is set to — the argument #1150 makes about the built-in review prompt applies here too, and dropping the enumeration would end the drift rather than reset it.

tend-agent:
Dropped the enumeration rather than rewriting it, in `64e3bf0`. Your second point decides it: the visible set is server-side and per-account, so any list I put there is a snapshot that goes stale on its own schedule — the list I replaced was already wrong before this bump, and the one I wrote was wrong on arrival. Resetting it just moves the next staleness out by a few releases, and nothing in CI would catch it when it comes round again.

What's left in the description is the part this repo can check: the action passes `--model` unconditionally, so whatever Codex bundles as its no-model default never reaches a run, and a reader who needs the current set is pointed at the `"visibility": "list"` entries in `~/.codex/models_cache.json` on the account CI runs as — the source, not a copy of it.

I did not verify your list against that account, and can't from here: no `OPENAI_API_KEY` reaches this repo's runs, so there's no authenticated Codex client in the session to fetch a cache with. That's the other reason the enumeration had to go rather than be corrected — I had no way to get it right, only to guess differently.

The `gpt-6-astra`-as-bundled-default sentence is gone from the input description too. It survives only inside the release-note `<details>` in the PR body, now labelled as what the notes claim rather than as verified behavior. Nothing turns on it either way, since `--model` is always passed.

The `codex_version` bump itself is unchanged. On the new head `test-codex-surface` and `lint` have passed; `test`, `test-proxy`, `test-sandbox`, `test-worker` and `Analyze (ruby)` were still running when this session's poll budget ran out, so I'm reporting them as unverified rather than green. The commit touches only a YAML description string, and `uv run pytest` passed locally at this head.

- `uv run pytest` — 935 passed.
- `uv tool run pre-commit run --files codex/action.yaml` — clean.
- `--model` unconditional: [`codex/runner.py:205`](https://github.com/max-sixty/tend/blob/64e3bf0eaa7f3da07501d269d9fb830bc1496f27/codex/runner.py#L205), inside the `args` list with no surrounding condition.

</details>

tend-agent (commented):
The bump itself checks out: 0.153.4 is npm's current `latest` for both `@openai/codex` and `@openai/codex-responses-api-proxy`, which the install step pins to the same value. Dropping the enumeration instead of refreshing it is the right call for the reason the thread already settled.

One blocker, in a file the diff doesn't touch. `codex/refresh/action.yaml` still carries `codex_version: default: "0.151.0"`, so the two Codex pins now disagree, and `generator/tests/test_repo_pins.py::test_codex_actions_pin_the_same_cli_version` fails on the merged tree:

```
AssertionError: Codex CLI pins differ: {'codex/action.yaml': '0.153.4', 'codex/refresh/action.yaml': '0.151.0'}
```

That test arrived on `main` in [`f58de21`](https://github.com/max-sixty/tend/commit/f58de217f24351c4845963bc117b80f345bef114) after this branch was cut, which is why the PR body's local `uv run pytest` was green — the 935-test run predates the parity assertion. `test-codex-surface` won't cover the gap either: its install step reads `.inputs.codex_version.default` from `codex/action.yaml` alone, so the refresh action's pin is ungated by anything except that one test.

The consequence outside the red check is the part worth stating: the refresh action rotates the Plus/Pro credential bundle through Codex's own refresh path, so leaving it two minor versions behind means the client that mints the credentials is not the client the harness consumes them with. That is also the same surface the PR body records as unverified — no `OPENAI_API_KEY` reaches this repo's runs, so the isolated-login rotation check the weekly sweep asks for before a `codex_version` bump is unrun. Moving the pin at least keeps the two halves on one version rather than making the untested surface the stale one.

Fix is the one line, in `codex/refresh/action.yaml` under the `codex_version` input:

```yaml
  codex_version:
    default: "0.153.4"
    description: Pinned @openai/codex npm version that performs the refresh
```

Not approving — self-authored, and GitHub rejects a self-approval.

tend-agent:
Fixed in `cf3d54a`. The branch predated `codex/refresh/action.yaml` existing at all, so the pin had to come in with the file: merged `origin/main` (clean, no conflicts) and moved the refresh action's `codex_version` default to `0.153.4`. Both Codex pins are now on one version, and `git grep '0\.151\.0'` outside markdown returns nothing.

`test` is green on `cf3d54a` — `test_codex_actions_pin_the_same_cli_version` passes, and the local run on the merged tree is 1082 tests, up from the 935 the stale body cited. PR body updated: it now says the bump covers both pins, and its `codex/runner.py:205` link is re-anchored to `cf3d54a` (the old link pinned `61a2bfb`, a commit where that file did not yet exist).

On the second point: the rotation check stays unrun, and I can't close it from here — the isolated Plus/Pro login the sweep asks for has no equivalent in this repo's runs. I moved it into the body as an explicit **Unverified** line rather than leaving it in a `<details>` block, so whoever merges sees that the auth path rests on the release-note skim alone.

PR #1157, opened 2026-09-06
skills: make public prose reader-focused

max-sixty:
Tend's public prose guidance mixed "be concise" with mandatory layouts and exhaustive evidence, so agents could still publish their investigation instead of synthesizing it for the reader. This changes the shared guidance to organize PR descriptions, reviews, and replies around the reader's decision, the current conclusion, and only the evidence needed to act.

The workflow-specific skills now point to that goal instead of requiring their own body shapes. A few optional examples remain where format or affect carries useful information: a triage PR shape, an actionable-review contrast, an evidence details block, and warm but candid triage replies. Draft-review identity now uses a hidden marker instead of mandatory visible wording, while temporarily recognizing reviews posted by earlier releases. `TODO.md` records the current downstream verbosity baseline and a post-release check for whether the guidance improves synthesis or the examples become de facto templates.

The complete test suite and all pre-commit hooks pass.


plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -125,7 +125,8 @@ If `IS_DRAFT == "true"`, run a lighter review:
 - Skip step 2 (overlap with other PRs) — landing-readiness concern, premature for WIP.
 - Skip the duplication scan in step 4 — the author is still shaping the design.
 - Submit as **COMMENT only**, never APPROVE. GitHub blocks approving drafts, and the author hasn't asked for a verdict yet.
-- Open the review body with this exact line: `Reviewing as a draft — flagging anything that looks worth a quick fix. Mark ready for a full review.`
+- Make the review's context clear: this is feedback on work in progress, not a merge verdict, and the author can mark it ready to request the full review.
+- Include the exact hidden marker `<!-- tend:draft-review -->` anywhere in the review body. Posting mechanics uses it to replace this COMMENT with a full verdict when the PR becomes ready; it is not part of the reader-facing prose.
```
tend-agent: The marker is a machine key, but it is documented only here — and the two places that rewrite a review body never mention it. **Posting mechanics** says "Read that entire file in chunks, update the review" and "Findings the delta fixed drop out", and the 422 recovery PUTs a freshly composed `/tmp/updated-review-body.md`. An agent recomposing a draft review's body has no reason to reproduce an invisible non-prose token it isn't told to preserve; when it drops it, `draft_mode` goes false and the `ready_for_review` replacement in `review-preflight.sh` skips with "already carries a COMMENTED review" — the full verdict never posts. The old mandated opener didn't have this failure mode because it was reader-facing prose a recomposed body naturally re-emitted.

```suggestion
- Include the exact hidden marker `<!-- tend:draft-review -->` anywhere in the review body. Posting mechanics uses it to replace this COMMENT with a full verdict when the PR becomes ready; it is not part of the reader-facing prose. Carry it through any body you recompose — re-targeting after a mid-review push and the 422 body-only retry both rewrite the body, and dropping the marker there forfeits the replacement silently.
```

Relatedly, `generator/tests/test_review_preflight.py` has zero occurrences of the marker: `test_only_ready_for_review_replaces_a_draft_review` is parametrized over the legacy prose line only. So the live path — marker body → `draft_mode` → replacement — is covered inside `bot-review-state.sh` but never through the preflight that actually gates the post. Adding a `("ready_for_review", <marker body>, "post:")` case to that parametrize list closes it, and would also stop the 2026-12-01 fallback removal from leaving the live path untested.

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -340,13 +357,9 @@ Post at most one review per run. Give a verdict (**approve** or **comment**, nev
 
 For fixes targeting lines outside the diff, offer to push a fix commit instead.
```
tend-agent, line 359: Deleting the heredoc leaves the reader below it unguarded. `BODY=$(cat /tmp/review-body.md)` (further down this block, immediately above `REVIEWED=$(cat /tmp/reviewed-head) || exit 0`) is now the only reference to a file nothing in the recipe creates. This file already explains the hazard for the neighbouring line — "inlined as `$(cat ...)` a missing file substitutes the empty string and the POST still runs" — and guards it. If the new compose step is skipped, `BODY` is empty, `jq --arg body ""` sets `.body = ""`, and the POST goes out as the empty-body COMMENT this skill notes GitHub rejects.

The fix is one token on that line — `BODY=$(cat /tmp/review-body.md) || exit 0` — but it's outside this diff's hunks, so I can't attach it as a suggestion. Happy to push it if you'd like.

plugins/tend-ci-runner/scripts/bot-review-state.sh
```diff
@@ -104,8 +104,11 @@ gh api --paginate "repos/$REPO/pulls/$PR/reviews" \
           | last
           | if . == null then null else
               {id, state, at: .submitted_at,
+               # TODO(2026-12-01): Drop the prose-prefix fallback after draft
+               # reviews created by pre-marker releases have aged out.
                draft_mode: (.state == "COMMENTED"
-                            and ((.body // "") | startswith("Reviewing as a draft —")))}
+                            and (((.body // "") | contains("<!-- tend:draft-review -->"))
```
tend-agent, line 111: The contract header above still describes the old key: "`draft_mode` identifies Tend's *framed* draft COMMENT so a later full pass may replace it." "Framed" was exactly the mandated prose frame this commit decouples from, and the header now names neither the marker nor the dated fallback. It's outside this diff's hunks so I can't suggest it inline — happy to push the one-line update with the rest if useful.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -528,9 +536,9 @@ Open the most recent prior run first; go deeper only if the answer is not there.
...
+CI threads are high-latency, so each outward response must stand alone: give the current conclusion, its consequence, and the next action or decision. Self-contained does not mean publishing the whole investigation.
 
-Read logs, code, and API data before drawing conclusions. Show evidence: cite log lines, file paths, commit SHAs. Trace causation — if two things co-occur, find the mechanism rather than saying "this may be related." Never claim a failure is "pre-existing" without checking main branch CI history. Distinguish what you verified from what you inferred.
+Read logs, code, and API data before drawing conclusions. Trace causation — if two things co-occur, find the mechanism rather than saying "this may be related." Never claim a failure is "pre-existing" without checking main branch CI history. Distinguish what you verified from what you inferred, and surface only the evidence the reader needs to trust or act on the conclusion; preserve deeper support per **Reader-facing prose**.
```
tend-agent: This drops `Show evidence: cite log lines, file paths, commit SHAs.` and replaces it with a ceiling — "surface **only** the evidence the reader needs" — which bounds excess but sets no floor. That sentence was the only citation duty in the skill tree: `references/grounded-analysis.md` doesn't carry it (grepping it for `cite|log line|commit SHA|file path` returns two unrelated hits — the both-paths-failed branch and a `--jq` projection rule), and the `#L`-must-use-a-SHA rule under **Comment Formatting** is conditional formatting: it governs a citation once one is made. The same diff compounds it by removing "the sources, **line-anchored links**, and working notes" from the `<details>` guidance, so the duty is gone from the visible body and the collapsed block alike.

For a bot whose named failure modes are guessed links and invented SHAs, I'd keep a floor alongside the ceiling — they aren't in tension:

```suggestion
Read logs, code, and API data before drawing conclusions. Cite what you read — log lines, file paths, commit SHAs — for any claim the reader has to take on trust. Trace causation — if two things co-occur, find the mechanism rather than saying "this may be related." Never claim a failure is "pre-existing" without checking main branch CI history. Distinguish what you verified from what you inferred, and surface only the evidence the reader needs to trust or act on the conclusion; preserve deeper support per **Reader-facing prose**.
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -431,27 +431,35 @@ If `EXISTING` is greater than 0, **do not post** — another run already handled
...
+```markdown
+<details><summary>Reproduction and affected path</summary>
 
-...line-anchored source links, repro steps, working notes...
+...the evidence needed to verify or resume the analysis...
 
 </details>
 ```
 
+**Compose bodies with the Write tool, then post with `--body-file`.** The composed file is reviewable before it ships, quoting and escaping are non-issues, and line wrapping is just file content. The bot writes to `/tmp/` constantly — one more file is cheap. `--body "…"` is fine only for a one-line body containing no backtick, `$`, or `\`. Inside double quotes bash runs a backticked span as a command and substitutes its output, so a markdown inline-code span is silently deleted from the posted comment: `` --body "`some-check` now passes" `` ships as ` now passes`. Inline code appears in nearly every body the bot writes, and single-quoting instead breaks on any apostrophe, so reach for `--body-file` whenever the text is anything but plain prose.
```
tend-agent, line 454: `### Reader-facing prose` is the only subheading under `## Comment Formatting`, so it now runs to the next `##` and swallows the rest of the section: the `--body-file` rule, line wrapping, the `#L`-SHA rule, `check-body-links.sh`, the backtick-authoring rule, and the no-footers rule are all nested inside it. Nine cross-references from the other skills (`ci-fix`, `nightly`, `review`, `review-runs`, `review-reviewers`, `triage`) point at "Reader-facing prose" as if it were a scoped subsection, so an agent following one of them lands on a section with no defined end.

A sibling `###` before this paragraph — `### Mechanics`, or similar — gives it a boundary. I've left it as prose rather than a suggestion because the insertion point sits right after a fenced block and a suggestion spanning the fence would corrupt it.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -347,7 +347,7 @@ git worktree remove "/tmp/review-runs-fix" --force
 
 `.config/tend.yaml` and `CLAUDE.md` are not under the read-only mount, but if you're already in the worktree for a `.claude/skills/` edit, do those edits there too so the branch stays self-contained.
 
-- **PR** (default): Branch `daily/review-runs-$GITHUB_RUN_ID`, fix, commit, push, create with label `review-runs`. Lead the PR description with two or three sentences — problem, fix, verification — and put the full analysis (run IDs, log excerpts, root cause, gate assessment) inside `<details>`.
+- **PR** (default): Branch `daily/review-runs-$GITHUB_RUN_ID`, fix, commit, push, create with label `review-runs`. Write the description for a maintainer deciding whether the current change fixes the general behavior gap, following **Reader-facing prose** in `running-in-ci`. The tracking issue already carries the evidence history; link it and include only the evidence needed to understand this change.
```
tend-agent: "The tracking issue already carries the evidence history" is false for exactly the findings that become PRs. The tracker's own body, written by this skill, says "Monthly tracking issue for **below-threshold** findings", the section that writes to it is headed "Recording below-threshold findings", and the line below this one says to note **the rest** there. A finding that clears its gate on first observation goes straight to a PR and is never appended — so with the `<details>` mandate removed and `shared/review-gates.md` requiring no PR-body content either, its run IDs, log excerpts and gate assessment now live nowhere, and a maintainer can't see what justified promoting it.

```suggestion
- **PR** (default): Branch `daily/review-runs-$GITHUB_RUN_ID`, fix, commit, push, create with label `review-runs`. Write the description for a maintainer deciding whether the current change fixes the general behavior gap, following **Reader-facing prose** in `running-in-ci`. Link the tracking issue where it holds prior observations of the same behavior, and carry the evidence that justified promoting this finding — the run IDs, the log excerpt, and the gate assessment — in the body or a `<details>` block.
```

The parallel line in `review-reviewers/SKILL.md` is safe from this — that skill appends a `## Run <RUN_ID>` heading to the gist every run, threshold or not — though "link it" there now duplicates the instruction a few lines below to link `$GIST_URL`.

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -237,11 +237,7 @@ printf '%s\n' "$TITLE" > "/tmp/tend-pr-title"
...
-- **No version bump** (same-version regen): one sentence on what the regen changed (a generator template tweak the committed workflows were lagging). No version line, no commit list.
+Compose the PR body with the Write tool at `/tmp/tend-update-body.md`. Its reader is deciding whether to adopt the regenerated workflows, so explain the consumer-visible effect of the upgrade rather than inventorying changed files or commits. When the version changed, state the old and new versions, synthesize related upstream changes into the behavior adopters will notice, and link the comparison as support. Rewrite each `(#NNN)` reference as `max-sixty/tend#NNN` — a bare `#NNN` auto-links to this repo's own issues, not tend's. Filter out release mechanics, action-pin and lockfile bumps, and tend-internal work with no adopter-visible effect. If the comparison is unavailable, state only what the available evidence supports. For a same-version regeneration, explain the generator behavior that made the committed workflows stale. Follow **Reader-facing prose** in `/tend-ci-runner:running-in-ci`.
```
tend-agent: `/tmp/tend-upstream-commits.txt` is written six lines above and is now referenced nowhere else in the repo — the deleted bullet was its only consumer. "Synthesize related upstream changes" leaves the agent no pointer to where those changes are, so the file is written and abandoned. The fallback went the same way: "if the commits file is empty (the compare call failed), keep just the version line and the compare link" had a checkable trigger and a floor, while "if the comparison is unavailable, state only what the available evidence supports" has neither — the compare URL is always constructible, so the most natural reading licenses dropping the link the old rule required keeping.

```suggestion
Compose the PR body with the Write tool at `/tmp/tend-update-body.md`. Its reader is deciding whether to adopt the regenerated workflows, so explain the consumer-visible effect of the upgrade rather than inventorying changed files or commits. When the version changed, state the old and new versions, synthesize the entries in `/tmp/tend-upstream-commits.txt` into the behavior adopters will notice, and link the compare URL printed above as support; if that file is empty the compare call failed, so ship the version line and the compare link alone. Rewrite each `(#NNN)` reference as `max-sixty/tend#NNN` — a bare `#NNN` auto-links to this repo's own issues, not tend's. Filter out release mechanics, action-pin and lockfile bumps, and tend-internal work with no adopter-visible effect. If the comparison is unavailable, state only what the available evidence supports. For a same-version regeneration, explain the generator behavior that made the committed workflows stale. Follow **Reader-facing prose** in `/tend-ci-runner:running-in-ci`.
```

The bash comment above ("empty if the call fails, in which case the body carries only the version line") is stale for the same reason — it describes a body shape the prose no longer defines.

plugins/tend-ci-runner/skills/triage/SKILL.md
```diff
@@ -171,13 +188,7 @@ Note the PR number for the comment.
...
 
-Always comment via `gh issue comment`. Keep it brief, polite, and specific. A maintainer will always review — never claim the issue is fully resolved by automation alone.
-
-**Drop the maintainer-deferral closer** ("a maintainer will review", "I'll leave it for a maintainer to evaluate and prioritize", and the like) **when `author_association` is `OWNER`, `MEMBER`, or `COLLABORATOR`** — deferring to a maintainer reads as absurd when the reporter is one. Keep it otherwise, where it signals the action isn't authoritative.
-
-```bash
-gh api "repos/$GITHUB_REPOSITORY/issues/$ARGUMENTS" --jq '.author_association'
-```
+Always comment via `gh issue comment`. Write for the issue author: lead with the current disposition, then give the causal finding and the action taken or the one concrete input or decision still needed. Link any fix, reproduction, or duplicate. Follow **Reader-facing prose** in `/tend-ci-runner:running-in-ci`; do not restate the report or narrate the investigation. Acknowledge the reporter when the situation calls for it, but do not use thanks or maintainer deferrals as fixed openers and closers. Do not present the bot's judgment as a maintainer decision.
```
tend-agent: Two rules lost their only holder here, and one of them lost its input too.

**The deferral rule was conditional on who the reporter is**, not on phrasing: drop the closer when `author_association` is `OWNER`/`MEMBER`/`COLLABORATOR`, keep it otherwise as a signal the action isn't authoritative. "Do not use thanks or maintainer deferrals as fixed openers and closers" bans boilerplate, which a freshly-worded deferral addressed to the repo owner passes — and this skill's own new good example models one ("a maintainer still needs to decide whether discovery should be automatic or opt-in"). The `gh api … --jq '.author_association'` lookup went with it, so triage now has zero references to `author_association`, `OWNER`, `MEMBER` or `COLLABORATOR`; `shared/author-association.md` is `@`-included by `running-in-ci` and `notifications` only, not here. The signal has to come back before the rule can.

**"Never claim the issue is fully resolved by automation alone" has no successor.** "Do not present the bot's judgment as a maintainer decision" covers authority, not completeness — "this is fixed" is an overclaim about resolution state, not a presentation of judgment as a maintainer's. The new **Fix ready** example reads as done ("#123 keeps it across attempts and adds a regression test") with no signal a human still lands it, and nothing in `running-in-ci` restates the rule.

tend-agent (commented):
The reframing is the right direction — the old templates did produce bodies that inventoried an investigation instead of stating a conclusion — and `a0a7224` handles the draft-key coupling cleanly: hidden marker, dated prose fallback, test parametrized over both. On the merge ref `uv run pytest` is 936 passed and all 13 pre-commit hooks pass.

The inline comments are all the same shape: a rule or an artifact that lost its only holder in the rewrite, rather than a disagreement with the framing. The one I'd weigh heaviest is the Grounded Analysis change — `Show evidence: cite log lines, file paths, commit SHAs` was the only citation floor anywhere in the skill tree, and what replaces it is a ceiling. Next after that is the draft marker: it's a machine key documented in one bullet, and the two paths that recompose a review body aren't told to preserve it.

One thing that isn't a defect but is worth a decision. The bundled `SKILL.md` files come out ~226 words longer net (`running-in-ci` +166, `review` +91), because two deleted templates return as bad/good example pairs carrying the same contrasts at three times the length — `review`'s 5-line table becomes a 20-line example, and triage's PR-body template comes back twice. `CLAUDE.md` under **Authoring skills** asks for brevity and "recipe over narrative" precisely because these load into every relevant session, and this PR's own TODO entry names "optional examples become de facto templates" as the outcome to watch for. A longer, more concrete example is a stronger attractor than the terse template it replaced, so the guardrail and the change are pulling against each other. Worth deciding deliberately rather than at the next review.

tend-agent (approved):

PR #1158, opened 2026-09-06
Move runner logic from Bash and skills into Python

max-sixty:
## Summary

This moves Tend runner behavior with deterministic state or parsing out of Bash and skill prose into tested Python.

- The runner plugin now has 15 Python commands and no shell scripts. Review state, CI polling, evidence storage, nightly regeneration, run listing, PAT checks, link checks, and token reporting all have behavior tests.
- The Python outage enricher retains annotation-first reporting, falls back to the failed-log tail for ordinary non-zero exits, and bounds whole run sections so oversized batches make progress instead of jamming nightly enrichment.
- The generated mention and notification gates, sandbox setup, system-prompt composition, and Codex harness phases now run through Python. Production shell remains only for linear installation, authentication, and Git instruction-file pinning.
- Stateful workflows expose cohesive command groups: review preflight has `start` and `post`; nightly regeneration has `prepare` and `ship`; evidence workflows have `prepare` and `append`; and the Codex harness has `install-plugin`, `stage-agents`, and `run`.
- The remaining deterministic Bash program in Tend's repo-specific weekly skill now lives in a fail-closed Python updater. Skills keep short command recipes and the review payload that agents must customize.

## Architecture

There is no global runner application. The commands share GitHub transport through `github_cli.py`, but review lifecycle, workflow regeneration, evidence collection, and harness execution have different state and callers. Each command hierarchy stays with its domain; a global dispatcher would add a package boundary without removing concepts.

The post-migration survey counts 33 runtime Python files with 7,628 lines. The seven standalone production shell files contain 226 lines. Composite actions have 34 `run` blocks totaling 122 lines, of which 24 are one line. Runner skills contain 99 code fences totaling 402 lines; the only fence over 20 lines is the 23-line review payload whose prose and inline suggestions are agent-authored. There are no production `awk`, `sed`, or `grep` invocations. Token aggregation retains one tested `jq` program inside its Python command because the input is already JSON and the transformation is a single deterministic projection.

## Verification

- `uv tool run pre-commit run --all-files`
- `wt test` — 1,014 Python tests, 32 worker tests, and TypeScript typechecking


.claude/skills/running-tend/scripts/refresh_consumers.py
```diff
@@ -0,0 +1,173 @@
...
+        if "/" not in name:
+            raise ValueError("GitHub code search returned an invalid repository")
+        if path.startswith(WORKFLOW_PREFIX):
+            repositories.add(name)
+    return repositories
+
+
+def read_consumer(repository: str) -> dict[str, str] | None:
+    """Return one installed consumer, or ``None`` when no Tend workflow remains."""
+    workflows = github_json("api", f"repos/{repository}/contents/.github/workflows")
```
tend-agent: A permanent 404 here fails the whole refresh instead of dropping the repo. `github_json` raises on any non-zero `gh` exit, and `refresh` lets that escape, so a consumer that was deleted, went private, was renamed, or removed `.github/workflows` outright makes the weekly run exit non-zero — every week, forever — and `data/consumers.json` (which the site Worker renders) stops updating until someone edits it by hand. That repo can never appear in the `removed` list the skill tells the agent to confirm, either. The shell this replaces handled it: `gh api ... 2>/dev/null || workflows=0` then `continue`, and the same on the `.config/tend.yaml` read.

The fail-closed behavior is right for transient errors — that's what `test_github_failure_leaves_the_existing_index_untouched` pins, and it should stay — but 404 isn't transient and means exactly what this function's `None` return is for. Distinguishing them needs `github_json` to keep the failure detail it currently discards (it raises `CalledProcessError(returncode, args)` with no `stderr`), then an opt-in `allow_missing` on the two reads in `read_consumer` that returns `None` on `HTTP 404` and re-raises everything else. Happy to push that if you'd rather not.

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -312,42 +249,31 @@ An approval you post at a re-targeted head is yours to stand behind: the queued
...
-       | select((.detailsUrl // .targetUrl // "") | test($own) | not)
-       | select((.workflowName // "") == $wf | not)
-       | (.status // .state)
-       | select(IN(["IN_PROGRESS","QUEUED","PENDING","WAITING","REQUESTED","EXPECTED"][]))] | length' <<<"$ROLLUP")
+uv run --script "${CLAUDE_PLUGIN_ROOT}/scripts/poll_pr_checks.py" \
+  snapshot <number> "$(cat /tmp/reviewed-head)"
 ```
 
-`$ROLLUP` is shell state, and the poll the branches below prescribe is necessarily its own Bash call, so the variable is gone when you come back. **Re-run that whole block — not just the `FAILED=` line — after any poll.** Both stale reads fail toward approval and neither errors: `jq` over an empty `$ROLLUP` exits 0, so `$FAILED` reads clean, and the provenance loop iterates zero times, making its "every one `cancelled`" test vacuously true.
+The JSON reports `head_sha`, `pending`, and `failed` for the pinned commit after
```
tend-agent, line 256: Two gaps at the pre-APPROVE gate, both from the snapshot swap.

`snapshot` exits 2 with a stderr line and no JSON when every context on the pinned commit is filtered out — `reduce_rollup` returns `None` for an empty `contexts` list, and the filters drop this run, this workflow, and `tend-review`. On an adopter whose only workflows are tend's, that is the normal state of a PR, and it is precisely the "clean external-dependency bump carries zero review signal" case the terminal-red branch below worries about. The three branches here only cover `failed`/`pending`, so the agent improvises at the one gate that decides approve-vs-withhold. The old recipe read empty and fell through to APPROVE; worth saying so explicitly.

Separately, the paragraph above still reads "The rollup below is the live head's" — it isn't any more. The snapshot is pinned to `/tmp/reviewed-head` and reports the live head as `head_sha`, which is the better design, but the prose now describes the mechanism it replaced.

tend-agent (commented):
Read this as a port: I diffed each new Python module against the shell it replaces rather than re-deriving intent. The security-critical paths check out — `setup_sandbox.py` preserves the PATH translation order, the reserved-env boundary, the credential strip and its verify, the export-before-first-fallible-step ordering, and the proxy readiness signal; the `env -i` launch and the `pwd`-derived runner home are real hardening, and `test-setup-sandbox.sh` grew assertions for them. The pre-commit hook dropping its AGENT_ENV_FILE-keys check is covered by `base_agent_env`'s subset assertion plus `test_every_fixed_agent_assignment_is_reserved`, so that invariant survives.

Two things I'd want changed, both inline.

Also worth coordinating: this deletes `enrich-tend-outage-issues.sh`, which #1152 is currently fixing (annotations that name no cause fall back to the job log, per #1151). The Python port here has no such fallback, so whichever lands second silently drops or has to re-apply the other's work.

Parity-verified against their base versions: `setup_sandbox.py`, `mention_verify.py`, `notifications_check.py`, `bot_review_state.py`, `review_preflight.py`, `poll_pr_checks.py`, `check_body_links.py`, `enrich_tend_outage_issues.py`, `nightly_workflow_update.py`, `list_recent_runs.py`, `codex/runner.py`, `compose_system_prompt.py`. Confirmed `uv run --script -` accepts a PEP 723 body on stdin and that a sibling `import github_cli` resolves under `uv run --script`, so both new invocation shapes work.

`uv run pytest` in a clean worktree of `dcabee9`: 1000 passed. (In a `tend-review` checkout four fail, because the base-branch restore of `.claude/**` removes `refresh_consumers.py` and reverts `running-tend/SKILL.md` — not a defect in this PR.)

</details>

tend-agent (commented):
The three edge cases this commit restores all check out: the `allow_missing` 404 path drops a deleted/renamed/private consumer into `removed` while still failing closed on anything transient, `allow_filtered_empty` is scoped to `snapshot` alone so `poll` keeps returning UNVERIFIED on an empty rollup, and the batch/run/line bounds hold — worst case is roughly 61 KB against GitHub's 65 KB comment limit, with a run's marker written only for runs that actually rendered, so a truncated batch resumes the next night rather than losing the rest. `uv run pytest` on a clean worktree of `af3b5c5`: 1012 passed.

One finding, inline: the annotation wrapper is a three-backtick fence around text that is itself agent prose, so a fenced block in that prose breaks the comment. Three coupled suggestions below (helper, then both call sites).

Also worth acting on separately: this now carries #1152's log fallback and its bounding, so #1152 is superseded rather than conflicting — it can be closed once this lands.

plugins/tend-ci-runner/scripts/enrich_tend_outage_issues.py
```diff
@@ -0,0 +1,199 @@
...
+            continue
+        messages = [
+            str(annotation.get("message") or "")
+            for annotation in annotations
+            if annotation.get("annotation_level") == "failure"
+            and not (annotation.get("message") or "").startswith("Process completed")
+        ]
+        messages = [message for message in messages if message]
+        if messages:
+            detail = f"#### {job['name']}\n\n```\n{bounded_message(messages)}\n```"
```
tend-agent: The annotation source here is the agent's own closing text — `failure_reason` in `shared/steps/run_claude.py` returns "the last non-blank assistant text", and its docstring names this script as the consumer. Agent prose routinely contains a fenced code block, so a three-backtick wrapper around it terminates early.

Reproduced against this commit with an annotation whose message is ``I could not finish. The failing command was:\n\n```bash\nuv run pytest\n```\n\nRe-run it locally.`` — GitHub closes the wrapper at the inner bare fence, renders `Re-run it locally.` as prose, and then the wrapper's own closer *opens* a new block that swallows the next run's `### [Run 12]` heading and link. The `<!-- enriched-run:11 -->` marker becomes visible literal text inside it. Dedup still works (the regex reads the raw body), but the comment the enrichment exists to produce is unreadable.

`log_details` below already sizes its fence at four backticks for exactly this reason, so the two paths disagree. Four isn't enough on its own either — tend's own skills instruct authoring four- and five-backtick fences, and the agent quotes them. Sizing the fence to the content fixes both paths; I verified this suggestion plus the two below renders correctly and keeps all 12 tests in `test_enrich_tend_outage_issues.py` green.

```suggestion
            detail = f"#### {job['name']}\n\n{fenced(bounded_message(messages))}"
```

plugins/tend-ci-runner/scripts/enrich_tend_outage_issues.py
```diff
@@ -0,0 +1,199 @@
...
+        match
+        for comment in comments
+        for match in ENRICHED_MARKER.findall(comment.get("body") or "")
+    }
+    return sorted(referenced - enriched)
+
+
+def truncate_line(line: str) -> str:
+    """Bound one rendered line without splitting a UTF-8 code point."""
+    return line.encode()[:MAX_LINE_BYTES].decode(errors="ignore")
```
tend-agent, line 55: The helper the suggestion on line 97 calls. `re` is already imported; a module-level `BACKTICK_RUN` constant would match the file's convention for the other patterns if you'd rather.

```suggestion
    return line.encode()[:MAX_LINE_BYTES].decode(errors="ignore")


def fenced(body: str) -> str:
    """Wrap body in a fence longer than any backtick run it contains."""
    longest = max((len(run) for run in re.findall(r"`+", body)), default=0)
    wrap = "`" * max(3, longest + 1)
    return f"{wrap}\n{body}\n{wrap}"
```

plugins/tend-ci-runner/scripts/enrich_tend_outage_issues.py
```diff
@@ -0,0 +1,199 @@
...
+        return []
+
+    lines = output.splitlines()
+    errors = [index for index, line in enumerate(lines) if "##[error]" in line]
+    if not errors:
+        return []
+    end = errors[-1]
+    window = lines[max(0, end - 30) : end + 1]
+    body = "\n".join(clean_log_line(line) for line in window)
+    return [f"#### log tail\n\n````\n{body}\n````"]
```
tend-agent: Same helper here, so the fixed four backticks stop being a guess about what log output contains.

```suggestion
    return [f"#### log tail\n\n{fenced(body)}"]
```

tend-agent (approved):

PR #1159, opened 2026-09-06
Add experimental Codex subscription auth

max-sixty:
Adds experimental ChatGPT Plus/Pro authentication for the Codex harness while preserving API-key auth. Concurrent jobs receive an access-only bundle, and a generated serialized weekly workflow lets Codex rotate the full bundle with its built-in refresh path before persisting both secrets.

Auth preparation, validation, and publication live in tested Python. The setup guidance, security checks, generated workflow snapshots, and pinned Codex surface probe cover the new secret set and its unsupported public-CI boundary.

Tested with `wt test` and `uv tool run pre-commit run --all-files`.


shared/steps/codex_subscription_auth.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # Validate the existing consumer too. This catches a stale installation
+    # where every job still has the refreshable bundle the split design is
+    # specifically meant to remove.
+    _validate_consumer(_object(codex_auth_json, CONSUMER_AUTH_SECRET))
+    full = _object(refresh_auth_json, FULL_AUTH_SECRET)
+    _validate_full(full)
+
+    staged = copy.deepcopy(full)
+    staged["last_refresh"] = FORCED_STALE_REFRESH
```
tend-agent, line 170: Forcing `last_refresh` stale assumes `last_refresh` is what triggers the refresh. That holds for the pinned 0.151.0 and for OpenAI's guide ("if `last_refresh` is older than about 8 days, Codex refreshes"), but current Codex checks the access token's own JWT expiry *first* and returns early — `codex-rs/login/src/auth/manager.rs`, `needs_refresh`:

```rust
if let Some(tokens) = auth_dot_json.tokens.as_ref()
    && let Ok(Some(expires_at)) = parse_jwt_expiration(&tokens.access_token)
{
    return expires_at
        <= Utc::now()
            + chrono::Duration::minutes(CHATGPT_ACCESS_TOKEN_REFRESH_WINDOW_MINUTES);
}
let last_refresh = ...;
last_refresh < Utc::now() - chrono::Duration::days(TOKEN_REFRESH_INTERVAL)
```

([permalink](https://github.com/openai/codex/blob/a7b86b6201d35200280e3d24ca5ebccd8122d6ca/codex-rs/login/src/auth/manager.rs#L2934-L2945); `CHATGPT_ACCESS_TOKEN_REFRESH_WINDOW_MINUTES` is 5, `TOKEN_REFRESH_INTERVAL` is 8.)

Once that lands in a pinned version, the staged 1970 timestamp is never read: Codex leaves the bundle alone while the access token has more than five minutes left, and `publish_refresh` then raises "Codex did not refresh the staged auth.json" — a red weekly job on a credential that is actually fine, week after week. The blast radius is bounded (nothing is published, so nothing is corrupted), but the mechanism stops working and it fails in the direction of noise rather than a signal anyone can act on.

Worth either keying the assertion on the token values changing rather than on `last_refresh`, or adding "does `needs_refresh` still key on `last_refresh`" to the `codex_version` bump rule in `running-tend`, next to the `--output-last-message` and sandbox-behavior items already there.

generator/src/tend/workflows.py
```diff
@@ -728,6 +743,8 @@ def generate_all(
         wf = gen_fn(cfg)
         wf = _apply_extras(wf, wf_cfg)
         results.append(wf)
+    if "codex" in cfg.enabled_harnesses():
+        results.append(generate_codex_auth_refresh(cfg))
```
tend-agent: Every other generated workflow goes through `_apply_extras` and a `cfg.workflows[...]` lookup; this one is appended raw, between two blocks that both do. So `tend-codex-auth-refresh.yaml` is the only file an adopter can't retarget or turn off — `workflows: {codex-auth-refresh: {enabled: false}}` prints `Warning: unknown workflow` and is ignored, and there's no `workflow_extra`/`jobs` merge either.

`runs-on` is the override most likely to be wanted here: OpenAI's guide recommends a self-hosted runner for exactly this pattern, and an adopter on one has no way to say so.

```suggestion
    if "codex" in cfg.enabled_harnesses():
        wf_cfg = cfg.workflows.get("codex-auth-refresh", WorkflowConfig())
        if wf_cfg.enabled:
            results.append(_apply_extras(generate_codex_auth_refresh(cfg), wf_cfg))
```

That also needs `"codex-auth-refresh"` added to `KNOWN_WORKFLOWS` in `config.py`, alongside the `install-test` entry and for the same reason.

shared/steps/codex_surface.py
```diff
@@ -0,0 +1,183 @@
...
+        "OPENAI_API_KEY": None,
+        "auth_mode": "chatgptAuthTokens",
+        "tokens": {
+            "access_token": "access-only-probe",
+            "refresh_token": "",
+            "id_token": f"e30.{payload}.c2ln",
+            "account_id": "probe-account",
+        },
+        "last_refresh": "2026-09-06T12:00:00Z",
+    }
```
tend-agent: This hand-writes the bundle that `consumer_auth()` produces rather than calling it, which is the one thing the probe exists to check: if the projection changes shape — a field Codex starts requiring, or `OPENAI_API_KEY` stopping being nulled — the literal here keeps passing and the break lands in adopters' jobs instead. Deriving it also drops the frozen `"2026-09-06T12:00:00Z"`, which is today's date pinned into a probe that runs indefinitely.

```suggestion
    consumer = codex_subscription_auth.consumer_auth(
        {
            "auth_mode": "chatgpt",
            "tokens": {
                "access_token": "access-only-probe",
                "refresh_token": "probe-refresh",
                "id_token": f"e30.{payload}.c2ln",
                "account_id": "probe-account",
            },
        }
    )
```

I dropped `last_refresh` entirely rather than computing one — for `chatgptAuthTokens` a missing `last_refresh` short-circuits Codex's staleness check. If the pinned CLI turns out to want the key present, a `datetime.now(timezone.utc).isoformat()` beats a literal.

generator/src/tend/templates/codex-auth-refresh.yaml.j2
```diff
@@ -0,0 +1,26 @@
...
+# Only this workflow may rotate the refresh-token chain. Consumer jobs share
+# the access token concurrently; a second refresher must never race this one.
+concurrency:
+  group: tend-codex-auth-refresh
+  cancel-in-progress: false
+
+jobs:
+  refresh:
+<<fork_guard_if(cfg)>>
+<<environment()>>
```
tend-agent: The macro emits `    if: ...\n` or nothing, so on its own line it leaves a blank line inside the `refresh:` mapping in every adopter's generated file — visible in the committed regtest snapshot (`jobs:` / `  refresh:` / blank / `    environment:`). `notifications.yaml.j2` and `scheduled.yaml.j2` both put the next line on the same source line to avoid this.

```suggestion
<<fork_guard_if(cfg)>><<environment()>>
```

Renders `    if: ...` immediately followed by `    environment:` with the guard set, and starts at `    environment:` without it.

tend-agent (commented):
Nice split — the single-writer design does answer the objection that closed the previous `auth.json` attempt, and the tests, security-model rows, and the credential-free surface probe all pull their weight. Findings inline; two things that don't fit on a line:

**`cli_auth_credentials_store` asymmetry.** `codex/refresh/action.yaml` passes `--config 'cli_auth_credentials_store="file"'`, but the consumer `Run Codex` step in `codex/action.yaml` doesn't — it relies on the pinned CLI defaulting to the file store on a keyring-less runner. Deliberate (only the write path needs it) or worth passing in both? `_probe_consumer_auth` covers the read path for `codex login status` only, so a default change would surface in adopters' `codex exec` rather than in `test-codex-surface`.

**Fork-guard coverage.** `test_fork_guard_present_when_repo_owner_set` is parametrized over `_GUARDED_WORKFLOWS`, which doesn't include `tend-codex-auth-refresh.yaml` — the guard does render (I checked), but this is a `schedule` + `workflow_dispatch` workflow, exactly the shape that list exists for. It needs a codex config to reach, so it doesn't drop into the existing parametrization for free.

<details><summary>Notes</summary>

- Verified against OpenAI's linked guide (`https://learn.chatgpt.com/docs/auth/ci-cd-auth`, fetched today): the `~8 days` staleness figure and the weekly cadence are its own recommendation, and it does state the guide "does not apply to ... external-token host integrations (`auth_mode: "chatgptAuthTokens"`)" — which the README and the security-model page already say plainly.
- `chatgptAuthTokens` is a first-class variant in `codex-rs/protocol/src/auth.rs` (`#[serde(rename = "chatgptAuthTokens")] ChatgptAuthTokens`), not a value the client will reject.
- `Environments: Read and write` is the right fine-grained permission for `PUT /repos/{owner}/{repo}/environments/{env}/secrets/{name}`, so the `CODEX_REFRESH_PAT` instruction in `install-tend`'s SKILL.md is accurate.
- `test-codex-surface` and `test` were still running when this review was written; I'll follow up if either goes red.

</details>

shared/steps/codex_subscription_auth.py
max-sixty: Kept the forced `last_refresh` mechanism for pinned Codex 0.151.0. Every `codex_version` bump now requires an authenticated isolated refresh that rotates both credential views, plus a source check that `needs_refresh` still consults `last_refresh`; a version with the JWT-expiry short circuit cannot be bumped without redesigning the mechanism.

generator/src/tend/workflows.py
max-sixty: Added `codex-auth-refresh` to the workflow registry and routed generation through the common `WorkflowConfig` enabled and override path. Tests cover both disabling the workflow and overriding its runner.

shared/steps/codex_surface.py
max-sixty: The surface probe now derives its access-only fixture through `consumer_auth()` and omits `last_refresh`, so it exercises the production projection without freezing a date.

generator/src/tend/templates/codex-auth-refresh.yaml.j2
max-sixty: Joined the two macros and updated the snapshots. The fork-guard suite now includes the Codex refresh workflow as well.

shared/steps/codex_subscription_auth.py
tend-agent, line 170: `should_refresh_proactively` in the pinned 0.151.0 reads the access token's own JWT `exp` first, and only falls back to `last_refresh` when that fails to parse:

```rust
if let Some(tokens) = auth_dot_json.tokens.as_ref()
    && let Ok(Some(expires_at)) = parse_jwt_expiration(&tokens.access_token)
{
    return expires_at <= Utc::now() + chrono::Duration::minutes(CHATGPT_ACCESS_TOKEN_REFRESH_WINDOW_MINUTES);
}
let last_refresh = ...;
last_refresh < Utc::now() - chrono::Duration::days(TOKEN_REFRESH_INTERVAL)
```

Real ChatGPT access tokens are JWTs, so that branch always wins and the staged 1970 timestamp is never read. Confirmed against the npm 0.151.0 binary: with `last_refresh` at 1970 and an access token whose `exp` is far out, Codex makes no refresh attempt at all. The weekly job then hits `publish_refresh`'s "Codex did not refresh the staged auth.json" and rotates nothing, week after week. Nothing is corrupted, but the credential never renews.

The two roles also pull against each other on the same fact: a consumer holds an access-only bundle for a whole week, which only works if `exp` is more than a week out, and that is exactly the case where the refresher's own token is never near expiry when the cron fires.

Making the staged access token unparseable reaches the fallback and restores the mechanism. The refresh exchange uses only `refresh_token`, and the real access token stays in `CODEX_REFRESH_AUTH_JSON` until `publish_refresh` replaces it, so a failed refresh still leaves the good bundle intact.

```suggestion
    # 0.151.0's should_refresh_proactively prefers the access token's own JWT
    # expiry and only falls back to last_refresh when that fails to parse, so
    # the sentinel is reachable only behind an unparseable access token.
    staged["tokens"]["access_token"] = "tend-forces-refresh"
    staged["last_refresh"] = FORCED_STALE_REFRESH
```

The alternative, if fabricating a token is worse than the problem: drop the sentinel and the "did not refresh" guard, and key `publish_refresh` on the token values having changed. Then nothing forces rotation and the job renews only when Codex happens to be inside its five-minute pre-expiry window, which a weekly cron will rarely be.

shared/steps/codex_surface.py
```diff
@@ -0,0 +1,183 @@
...
+            'model_reasoning_effort="low"',
+            "x",
+            cwd=cwd,
+            env=env,
+            check=False,
+        ).stdout
+        if UNKNOWN_CONFIG_ERROR in known:
+            raise SurfaceError(
+                "codex no longer accepts the model_reasoning_effort config key"
+            )
```
tend-agent: `model_reasoning_effort` gets a negative control and a positive check here because a renamed `-c` key is silently ignored rather than rejected — I checked, and 0.151.0 runs `codex exec -c cli_auth_credentials_storeZZZ="file"` without complaint. `cli_auth_credentials_store` now has that same property and more reach: `codex/action.yaml` and `codex/refresh/action.yaml` both pass it, and it decides where the credential is read from and written to. A rename upstream leaves `test-codex-surface` green and breaks the refresh instead.

```suggestion
        for key in (
            'model_reasoning_effort="low"',
            'cli_auth_credentials_store="file"',
        ):
            known = _run(
                codex,
                "exec",
                "--strict-config",
                "-c",
                key,
                "x",
                cwd=cwd,
                env=env,
                check=False,
            ).stdout
            if UNKNOWN_CONFIG_ERROR in known:
                raise SurfaceError(f"codex no longer accepts the {key} config key")
```

tend-agent (commented):
Correcting my own earlier thread on `stage_refresh`. I wrote that forcing `last_refresh` stale holds for the pinned Codex 0.151.0 and only breaks once the JWT-expiry short circuit lands in a later release. That is wrong: 0.151.0 already has it, so the sentinel never fires against a real bundle. The resolved thread and the new `codex_version` bump rule in `running-tend` both rest on that premise. Mechanism, repro against the pinned binary, and a fix that restores the intended path are inline on `shared/steps/codex_subscription_auth.py`.

Everything else from the previous round reads as addressed — the workflow now goes through `WorkflowConfig`, the fork-guard suite covers it, the macros are joined, and the surface probe derives its fixture from `consumer_auth()`. `uv run pytest` is 969 green on the merged tree here.

One cross-PR consequence: #1154 (`chore: bump codex_version to 0.153.4`) can't land unchanged on top of this. It moves only `codex/action.yaml`, so the new `test_codex_actions_pin_the_same_cli_version` fails on it; and `codex-rs/login/src/auth/manager.rs` is byte-identical between `rust-v0.151.0` and `rust-v0.153.4`, so the bump rule's "do not bump if the access token's JWT expiry bypasses that field" doesn't separate the two versions — whatever is decided for the pin here applies to both.

<details><summary>Sources and repro</summary>

`should_refresh_proactively` at [`manager.rs#L2924-L2945`](https://github.com/openai/codex/blob/78c290807ce710180111df227df3b7a4fe845452/codex-rs/login/src/auth/manager.rs#L2924-L2945), the tag `rust-v0.151.0` resolves to. `AuthManager::auth()` is its only caller ([#L2350-L2354](https://github.com/openai/codex/blob/78c290807ce710180111df227df3b7a4fe845452/codex-rs/login/src/auth/manager.rs#L2350-L2354)).

`rust-v0.151.0` → `78c290807ce710180111df227df3b7a4fe845452`, `rust-v0.153.4` → `3d2ee51ca2d5db578f328aa75e20aa22c0197c9a`; `manager.rs` has the same md5 at both.

Repro, against `@openai/codex@0.151.0` installed from npm, with a synthetic `auth_mode: chatgpt` bundle and a bogus refresh token, invoked the way the refresh action does (`--config 'cli_auth_credentials_store="file"'`):

| staged `access_token` | staged `last_refresh` | refresh attempted? |
|---|---|---|
| JWT, `exp` ~400 days out | `1970-01-01T00:00:00Z` | **no** — no `Failed to refresh token` line at all |
| JWT, `exp` one hour ago | recent | yes |
| `not-a-jwt` | `1970-01-01T00:00:00Z` | yes |

The third row is the suggested fix. A live rotation still needs a real Plus/Pro login — I only have the credential-free half.

`cli_auth_credentials_store` under `--strict-config` on 0.151.0: accepted (`Not inside a trusted directory`), while `cli_auth_credentials_storeZZZ` is rejected with `unknown configuration field`. Without `--strict-config`, `codex exec` accepts the misspelling silently and runs — which is what makes the probe suggestion on `codex_surface.py` worth having.

</details>

shared/steps/codex_subscription_auth.py
max-sixty: Confirmed the pinned behavior and changed the staged runner copy: its access token is now the `tend-forces-refresh` sentinel, while the refresh and ID tokens remain untouched. That makes Codex fall through to the forced-stale `last_refresh`. Publication rejects the sentinel before either secret write, so a failed refresh leaves the stored good bundle unchanged.

shared/steps/codex_surface.py
max-sixty: Added `cli_auth_credentials_store="file"` to the strict-config positive checks alongside `model_reasoning_effort`, and reran the real credential-free surface probe.

tend-agent (approved):

tend-agent (commented):
The base merge resolves cleanly: `AUTH_MODE` precedence and `cli_auth_credentials_store` both landed in `run_codex()` with coverage in `test_codex_runner.py`, and the conflicted parametrizations widened to `COMPOSITE_ACTIONS` rather than staying at main's two-entry tuple, so the new refresh action picks up shellcheck and the env-passing ban for free. My approval stands. Two gaps below, neither inside a diff hunk, so no inline suggestion.

**The weekly pin sweep can't see the new pin.** The composite-action enumerator in `.claude/skills/running-tend/SKILL.md` globs `*/action.yaml`, which matches one directory level, so `codex/refresh/action.yaml` is absent from it — the sweep prints `codex/action.yaml codex_version = 0.151.0` and nothing for the refresher. That file's own framing is that "a pin added later shows up without anyone remembering to list it here", and the rules table further down *was* updated to name both files, so the command looks like the piece left behind rather than a decision. `*/action.yaml */*/action.yaml` covers it. The new `test_codex_actions_pin_the_same_cli_version` stops the two from actually diverging, so the cost is a bump PR that fails CI rather than a stranded pin — but the report a sweep session reads to decide what to bump is now incomplete, and the next composite action added under a nested path inherits the same blind spot.

**`CLAUDE.md` still describes two composite actions.** The Architecture section reads "Every harness lives under a harness-named path; there is no bare-root default. One per harness:" and then lists claude and codex. This PR adds a third that a generated workflow pins by ref (`max-sixty/tend/codex/refresh@X.Y.Z`), and the blanket claim below it — "All actions resolve the bot's numeric ID at runtime, run security and rate-limit preflight, and upload session logs" — isn't true of the refresher, deliberately so, since it holds no bot token and checks out no adopter code. The Structure tree under `codex/` is missing `refresh/action.yaml` too. `README.md`, `docs/security-model.md` and `docs/tend.example.yaml` all got the corresponding update; `CLAUDE.md` is the one a session reads before touching any of this.

Happy to push both as a commit if you'd rather not hand-edit them.

tend-agent (approved):

PR #1160, opened 2026-09-06
Harden Tend for release

max-sixty:
Tend's release checks covered protected branches and environment-gated secrets, but they did not verify tag creation or GitHub's immutable-release setting, and the Codex harness still received real credentials directly. This change makes the release guarantees explicit, gives Codex the same credential-isolation boundary as Claude, and updates the security model to distinguish intended GitHub access from credential theft.

## What changes

- Extend `tend check` and `tend check --fix` to require an admin-gated all-tags ruleset and enable immutable releases. A non-admin nightly check reports that setting as unverified rather than manufacturing configuration drift.
- Run Codex as the non-sudo sandbox user behind Tend's exact-host GitHub proxy. API-key sessions use OpenAI's narrow Responses API proxy, so the agent keeps normal API and git access to every repository the bot account can reach but never receives the real PAT or OpenAI key.
- Support experimental ChatGPT Plus/Pro auth with an expiring access-only bundle for agent sessions and a serialized weekly refresher that alone receives the rotating refresh credential. The access token is intentionally available to the live agent; the refresh token and environment-write PAT are not.
- Consolidate Codex session logs only after sandbox processes are reaped, reject symlink escapes, and preserve token-usage reporting from the copied logs.
- Route both failed and cancelled watched workflows through `ci-fix`; ordinary cancellations are handled by the agent using the same diagnostic path as any other unsuccessful run.
- Label the Codex harness experimental, document that restricted modes do not initialize on the standard GitHub-hosted Ubuntu runner, reserve credential and Codex configuration keys from adopter-provided sandbox environment values, and document the resulting release and threat model.
- Extend the hosted-runner sandbox test with a real second-UID stub Codex launch that asserts the PAT, OpenAI key, and GitHub command-file paths do not cross.

## Validation

- `wt test` — 1,082 Python tests, 32 worker tests, and TypeScript checking pass.
- `uv tool run pre-commit run --all-files` — all hooks pass, including generated-workflow, actionlint, shellcheck, mirror-sync, and reserved-environment parity checks.

Codex still defaults to `danger-full-access`: the ephemeral runner contains local execution, while the separate UID and credential proxies protect the real PAT and API credentials. Subscription sessions deliberately receive their expiring access token. Tend has not yet validated its complete workflow under Codex's more restrictive modes.


generator/src/tend/checks.py
```diff
@@ -158,6 +160,70 @@ def detect_default_branch(repo: str) -> str | None:
...
+    if result.returncode != 0:
+        # GitHub represents the disabled state as 404 rather than a JSON body.
+        if "HTTP 404" in result.stderr:
+            return CheckResult(
+                "immutable-releases",
+                False,
+                "Immutable releases are disabled. A write-access bot can "
+                "rewrite a published release's assets or notes. Run "
+                "`tend check --fix`.",
+            )
```
tend-agent: A 404 here is the answer a token without admin gets, not the answer a repo with the setting off gets. Checked live from this run: on `tend-agent/tend-integration` (bot is admin, setting off) the same call returns `200 {"enabled": false, "enforced_by_owner": false}`, while on `max-sixty/tend` (bot has push, not admin) it returns `404`. So this branch only fires for a caller that cannot read the setting, and it reports that as a policy failure — which is what the nightly `tend check` run turns into a `configuration drift` issue the bot can never close.

```suggestion
        if "HTTP 404" in result.stderr:
            # 404 is what a token without admin gets; a repo with the
            # setting off answers 200 with `"enabled": false`. Neither
            # state is readable here, so report it as unverified.
            return CheckResult(
                "immutable-releases",
                None,
                "Could not read the immutable-releases setting (needs admin).",
            )
```

generator/tests/test_checks.py
```diff
@@ -1156,6 +1160,80 @@ def test_run_all_checks_no_repo() -> None:
...
+
+def test_check_immutable_releases_disabled_is_github_404() -> None:
+    with patch(
+        "tend.checks._gh",
+        return_value=_make_completed(returncode=1, stderr="gh: Not Found (HTTP 404)"),
+    ):
+        result = check_immutable_releases("owner/repo")
+
+    assert result.passed is False
+    assert "--fix" in result.message
```
tend-agent: This is the test that pins the 404 reading in place, so it moves with the check above.

```suggestion
def test_check_immutable_releases_404_is_unverified() -> None:
    # 404 is what a token without admin gets; a repo with the setting off
    # answers 200 with `"enabled": false`.
    with patch(
        "tend.checks._gh",
        return_value=_make_completed(returncode=1, stderr="gh: Not Found (HTTP 404)"),
    ):
        result = check_immutable_releases("owner/repo")

    assert result.passed is None
```

generator/src/tend/templates/ci-fix.yaml.j2
```diff
@@ -14,7 +14,7 @@ on:
 
 jobs:
   fix-ci:
-    if: <<fork_guard_prefix(cfg)>>github.event.workflow_run.conclusion == 'failure'
+    if: <<fork_guard_prefix(cfg)>>contains(fromJSON('["failure", "cancelled"]'), github.event.workflow_run.conclusion)
```
tend-agent, line 18: The prompt this workflow builds carries the run URL, head SHA and commit message, but not the conclusion, and the `ci-fix` skill opens with "CI has failed on the default branch. Diagnose the root cause, fix it, and create a PR" — it says nothing about cancellations. So a run cancelled by a `cancel-in-progress` concurrency group (the usual adopter setup on their CI workflow) now boots a session that must infer for itself that nothing failed, and step 3a's transient path is a plausible landing spot: it files and immediately closes a `tend-outage` tracker, once per cancellation.

The commit and the PR body don't name a cancellation this is meant to catch. If there is one, passing `${{ github.event.workflow_run.conclusion }}` into the prompt and giving the skill a cancellation branch would make the session cheap and correct; otherwise `== 'failure'` is the smaller thing to maintain.

tend-agent (commented):
Two things worth changing before this ships, plus some stale references the Codex move left behind.

**`immutable-releases` reads a permission failure as a policy failure** (inline). I checked both states against the live API from this run: on a repo where the bot is admin and the setting is off, `GET /repos/{repo}/immutable-releases` answers `200 {"enabled": false, "enforced_by_owner": false}`; on `max-sixty/tend`, where the bot has write but not admin, the same call answers `404`. So 404 is the no-admin answer rather than the disabled answer, and the branch mapping it to `passed is False` only ever fires for a caller that cannot see the setting. The caller that matters is the bot: the nightly skill runs `uv tool run tend@latest check` with the bot PAT and files `tend check: configuration drift on <owner>/<repo>` on any `FAIL`, closing it only once every line is `PASS`. From the next release that issue opens on every adopter, on a check the bot can never clear — and `--fix` cannot clear it either, since the PUT gets the same 404. `passed=None` is the shape the rest of the file already uses here (`admitted_refs`'s docstring calls the branch-protection 404 "unverified"), and nightly's rule for a `SKIP` is to leave the issue untouched.

**Routing `cancelled` into ci-fix** (inline). The prompt carries the run URL, commit and message but not the conclusion, and the ci-fix skill opens with "CI has failed on the default branch. Diagnose the root cause, fix it, and create a PR" with nothing said about cancellations. A concurrency-cancelled run — `cancel-in-progress` on the watched workflow is the common adopter setup — boots a session that has to work out for itself that nothing failed, and step 3a's transient path is a plausible landing spot: that files and immediately closes a `tend-outage` tracker, per cancellation. Either drop `cancelled` again, or, if there's a cancellation shape you specifically want covered, pass `github.event.workflow_run.conclusion` into the prompt and give the skill a branch for it.

**Stale references from the Codex restore move** — all outside the diff, so say the word and I'll push them:

- `docs/security-model.md` still names `shared/steps/pin-instruction-files.sh` as the Codex path, in the paragraph about where adopter notes may live. `codex/action.yaml` now runs `restore-sensitive-config.sh`, which covers strictly more (every PR event rather than fork `pull_request_target` only, pinned to the PR base rather than the default branch).
- That leaves `shared/steps/pin-instruction-files.sh` referenced only by its own test.
- The bundled `running-in-ci` skill still tells the agent the restore is "Codex on fork PRs only". Under the new wiring a Codex session on a same-repo PR that edits `CLAUDE.md` or `.claude/` reads base content while the skill says it does not — the case that section exists to prevent.

Smaller: the `GHA_VARS` loop in `Run Codex` re-implements `launch_env` from `shared/steps/_sandbox.py` — the same six withheld names, hand-copied into bash, with no parity hook of the sort `sandbox_env`'s reserved set has, so a name added to `WITHHELD` later would keep crossing into Codex sessions. #1158 already moves this path into `codex/runner.py`; whichever of the two lands second is the place to collapse it.

<details><summary>What I verified about the Codex credential path</summary>

Installed `@openai/codex@0.151.0` (the pinned version) in this session and ran `codex exec` with this PR's exact provider config and no `OPENAI_API_KEY` in the environment, pointed at a local HTTP listener. Codex issued `POST /v1/responses` to the configured `base_url` and logged `auth.header_attached=false` — the agent needs no key of its own for the Responses proxy to be the only holder, which is the assumption the whole step rests on.

Immutable-releases probe: `tend-agent/tend-integration` (bot has admin, setting off) returned `200 {"enabled":false,"enforced_by_owner":false}` under `X-GitHub-Api-Version: 2026-03-10`; `max-sixty/tend` (bot has push, not admin) returned `404` under the same header and API version.

</details>

generator/src/tend/checks.py
max-sixty: Fixed in `fd0e774e`. A 404 now returns `passed=None` with the admin requirement stated; a readable `{"enabled": false}` remains a failure. The tests cover enabled, disabled, and unverified responses separately.

generator/src/tend/templates/ci-fix.yaml.j2
max-sixty: Fixed in `fd0e774e` while retaining cancellation routing. The generated prompt now passes the run conclusion, and `ci-fix` explicitly diagnoses cancellations through the normal path. A concurrency-replaced or deliberate cancellation with no failed work exits silently and cannot enter the transient outage-tracker path; cancellations with evidence of a durable cause continue through remediation.

generator/tests/test_checks.py
max-sixty: Updated in `fd0e774e`: the 404 regression test now expects an unverified result, and the readable disabled state is covered independently through the API response body.

plugins/tend-ci-runner/skills/ci-fix/SKILL.md
```diff
@@ -41,11 +42,19 @@ gh issue list --state open --author "$BOT_LOGIN" --search "ci-fix: in:title" \
...
 ### 2. Diagnose and fix
 
-1. Get failure logs: `gh run view <run-id> --log-failed`
-2. Identify the failing job and root cause — don't just fix the symptom
-3. Search for the same pattern elsewhere in the codebase
-4. Reproduce locally using test commands from the project's CLAUDE.md
-5. Fix at the right level (shared helper > per-file fix)
+1. Read the run's conclusion and jobs: `gh run view <run-id> --json conclusion,jobs,url`
+2. Read the relevant logs: `gh run view <run-id> --log-failed` for a failure;
+   `gh run view <run-id> --log` for a cancellation
```
tend-agent: `--log` is the wrong fetch for a cancellation: a run cancelled before its jobs archive logs returns `log not found: <job-id>` and exits 1 (verified on run `34054391566` in this repo), and that is the concurrency-replacement case this branch is for. Step 1's `--json conclusion,jobs` already answers whether work failed, via the job and step conclusions.

```suggestion
2. Read the relevant logs: `gh run view <run-id> --log-failed` for a failure.
   For a cancellation, judge from the job and step conclusions read in step 1 —
   a run cancelled before its logs were archived answers `gh run view <run-id>
   --log` with `log not found`, so a failed job inside the cancelled run is the
   signal, not the log fetch.
```

tend-agent (commented):
One finding on the new cancellation path; the rest of the delta holds up.

**`gh run view --log` fails on the cancellation shape this change exists to catch** (inline). Step 2 now sends a cancellation to `gh run view <run-id> --log`, but a run cancelled before its jobs archive any logs has none to fetch — which is precisely the concurrency-replacement case. Checked live against run `34054391566` in this repo (a cancelled `tend-notifications` run): `--log` exits 1 with `log not found: 101543475314`, while `--log-failed` exits 0 with empty output. Step 1's own command already carries the evidence the paragraph below asks for — `gh run view 34054391566 --json conclusion,jobs` returns `{"conclusion":"cancelled","jobs":[{"conclusion":"cancelled","name":"notifications","steps":[]}]}` — so "first establish whether work failed" is answerable from the job and step conclusions, and the log fetch is the wrong first move for that half of the branch.

<details><summary>What I re-verified on the rest of the delta</summary>

The `passed=None` change behaves as intended against the live API with the bot's write-only token on this repo: `check_immutable_releases("max-sixty/tend")` now returns `SKIP` with the admin message rather than a failure, so nightly's `configuration drift` issue no longer fires on it, and `--fix` skips it too since `cli.py` keys on `passed is False`.

`check_tag_protection("max-sixty/tend", "tend-agent")` returns `True` under the same non-admin token, so the sibling check has no matching blind spot — `_ruleset_keeps_bot_out` reads `current_user_can_bypass` rather than the `bypass_actors` list, which the API does withhold here (the ruleset detail for `Tag operations` comes back with no `bypass_actors` key at all).

The `running-in-ci` rewrite matches the wiring: both `claude/action.yaml` and `codex/action.yaml` invoke `restore-sensitive-config.sh` unconditionally, and its `case` covers `pull_request_target`, both review events, and `issue_comment` on a PR, while `repository_dispatch` and `workflow_run` fall through to the no-op branch. All four ci-fix regtest outputs carry the new `Conclusion:` line, and `docs/tend.example.yaml` is still byte-identical to its `install-tend` mirror.

</details>

plugins/tend-ci-runner/skills/ci-fix/SKILL.md
max-sixty: Fixed in `56132bbf`. Failure runs still use `--log-failed`; cancelled runs now diagnose from the structured job and step conclusions already returned by `gh run view --json conclusion,jobs,url`, without assuming a log archive exists.

tend-agent (approved):

tend-agent (approved):

codex/action.yaml
```diff
@@ -142,18 +141,110 @@ runs:
...
+      run: |
+        SERVER_INFO_FILE="$RUNNER_TEMP/tend-codex-proxy.json"
+        rm -f -- "$SERVER_INFO_FILE"
+        (
+          exec env -u PROXY_API_KEY -u NODE_OPTIONS \
+            NODE_OPTIONS=--disable-sigusr1 \
+            "$CODEX_PROXY_BIN" \
+            --http-shutdown \
+            --server-info "$SERVER_INFO_FILE" <<< "$PROXY_API_KEY"
+        ) &
```
tend-agent: The proxy is backgrounded still holding this step's stdout and stderr, and it is meant to outlive the step. That is the shape that hangs a composite step: the runner drains the step's pipes, and a surviving writer keeps them open.

The repo's other long-lived credential holder does not do this — `start_proxy` in `proxy/setup_sandbox.py` opens `tend-proxy.log`, passes it as `stdout`/`stderr`, and sets `start_new_session=True`. Nothing in CI exercises this one: `test-codex-surface` only runs `codex-responses-api-proxy --help`, and `verify-codex-launch` passes a literal `CODEX_PROXY_URL=http://127.0.0.1:1234`. So if it does hang, an adopter's job is where it first shows.

```suggestion
        ) > "$RUNNER_TEMP/tend-codex-proxy.log" 2>&1 &
```

tend-agent (commented):
Three things on the Codex credential path. The base merge itself resolves cleanly — `uv run --python 3.12 pytest` is 1032 passed on the merged tree, and the resolutions in `codex/action.yaml`, `proxy/setup_sandbox.py` and `shared/steps/token_usage.py` keep both sides, including collapsing the hand-copied `GHA_VARS` loop into `_sandbox.launch_env` via `codex/runner.py`.

**The Responses proxy is backgrounded holding the step's stdout** (inline) — the shape that hangs a composite step, and the one place in this repo where a long-lived credential holder isn't given its own log file.

**`Set up credential-isolation sandbox` runs with the runner's environment intact** (inline) — the Claude action scrubs it with `/usr/bin/env -i` and an explicit allowlist, with a comment saying why; the Codex copy of the same step doesn't, and `setup_sandbox.py` hands `os.environ.copy()` to the `mitmdump` that ends up holding the PAT.

The removed `Validate auth configured` step is fine to drop, but nothing currently replaces the message it gave; the second inline suggestion folds that into the proxy's own failure branch.

codex/action.yaml
```diff
@@ -142,18 +141,110 @@ runs:
...
+        for _ in {1..10}; do
+          [ -s "$SERVER_INFO_FILE" ] && break
+          sleep 1
+        done
+        if [ ! -s "$SERVER_INFO_FILE" ]; then
+          echo "::error::OpenAI Responses proxy did not start"
+          exit 1
+        fi
```
tend-agent, line 299: With the redirect above, keep the proxy's own message reachable — it is also what has to stand in for the `Validate auth configured` step this PR removes.

I ran `@openai/codex-responses-api-proxy@0.151.0` (the pinned version) with an empty key on stdin: it exits 1 immediately with `Error: API key must be provided via stdin`. On success it writes exactly one line, `responses-api-proxy listening on 127.0.0.1:<port>`, and nothing per request. So an adopter who forgets the `OPENAI_API_KEY` secret now waits out the 10 s loop and gets `OpenAI Responses proxy did not start`, where they previously got the secret named.

```suggestion
        if [ ! -s "$SERVER_INFO_FILE" ]; then
          cat "$RUNNER_TEMP/tend-codex-proxy.log" >&2 || true
          echo "::error::OpenAI Responses proxy did not start (an unset OPENAI_API_KEY secret prints 'API key must be provided via stdin' above)"
          exit 1
        fi
```

codex/action.yaml
```diff
@@ -142,18 +141,110 @@ runs:
...
+        BASH_ENV: ""
+        TEND_GH_TOKEN: ${{ inputs.github_token }}
+        TEND_GITHUB_ONLY: "1"
+        ACTION_PATH: ${{ github.action_path }}/..
+        MITMPROXY_VERSION: ${{ inputs.mitmproxy_version }}
+        TEND_UV_DIR: ${{ runner.temp }}/tend-uv
+        UV_CACHE_DIR: ${{ runner.temp }}/tend-mitmproxy-uv
+        TEND_SANDBOX_PATH: ${{ inputs.sandbox_path }}
+        TEND_SANDBOX_ENV: ${{ inputs.sandbox_env }}
+      run: /usr/bin/env TEND_RUNNER_TOOL_PATH="$PATH" PATH=/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/python3 -E -s "${{ github.action_path }}/../proxy/setup_sandbox.py"
```
tend-agent: This step receives the real PAT and starts the process that keeps it, but it inherits the runner's whole environment. `claude/action.yaml`'s step of the same name runs `/usr/bin/env -i` with an explicit allowlist plus `UV_NO_CONFIG=1` and `PYTHONNOUSERSITE=1`, and its comment gives the reason — "this step is the only one that receives both real credentials".

`setup_sandbox.py` passes `env=os.environ.copy()` to the `uvx ... mitmdump` it launches, so whatever an adopter's `setup:` exported reaches the process holding the bot token: `HTTPS_PROXY` changes where its upstream requests go, `SSL_CERT_FILE`/`REQUESTS_CA_BUNDLE` change what it trusts, `PYTHONPATH` and `UV_INDEX_URL` change what code it runs. `uvx --no-config` covers uv's config files, not its environment variables.

The scrub costs an explicit allowlist (`TEND_GITHUB_ONLY` would have to join it), which I assume is why it is not here — but `docs/security-model.md` now reads as parity between the harnesses on this boundary, so either close the gap or say in the step comment which part of it is deliberate.

codex/action.yaml
```diff
@@ -225,27 +329,30 @@ runs:
...
+      shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
+      run: PATH=/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/python3 -E -s "${{ github.action_path }}/../shared/steps/mark_notification_read.py"
       env:
+        BASH_ENV: ""
         GITHUB_TOKEN: ${{ inputs.github_token }}
 
     # Gate on the job being red, not on the agent step specifically — see the
     # matching note in claude/action.yaml, including why the security preflight
     # is excluded. Every step ahead of `codex` (the rate-limit preflight, bot-ID
     # resolution, auth validation, the CLI and plugin installs, the AGENTS.md
```
tend-agent: `Validate auth configured` is gone as of this PR, so the list no longer describes the steps ahead of `codex`.

```suggestion
    # resolution, the CLI and proxy installs, the sandbox handoff, the AGENTS.md
```

codex/action.yaml
max-sixty: Fixed in 8294939f: the long-lived Responses proxy now sends both stdout and stderr to `$RUNNER_TEMP/tend-codex-proxy.log`, so it cannot retain the completed step’s pipes. I also exercised the pinned proxy locally through this exact clean-environment launch and shutdown path.

codex/action.yaml
max-sixty: Fixed in 8294939f: an empty key now fails immediately with `OPENAI_API_KEY is unset. Set the OPENAI_API_KEY repo secret.`, and startup failure prints the proxy log before the Actions error.

codex/action.yaml
max-sixty: Fixed in 8294939f: the Codex credential holder now uses the same `/usr/bin/env -i` allowlist and UV/Python hardening as Claude, with `TEND_GITHUB_ONLY` carried explicitly. The OpenAI proxy child is also launched with a clean environment through the already-resolved Node executable. The shared invariant test now covers both harnesses.

codex/action.yaml
max-sixty: Fixed in 8294939f: the failure-reporting comment now names the CLI/proxy installs, sandbox handoff, AGENTS.md staging, and PR config restoration.

docs/security-model.md
```diff
@@ -309,25 +314,29 @@ before running them — the default branch, or in `tend-review` the PR's base
...
+non-sudo `tend-sandbox` user, sharing the GitHub proxy machinery under the
+top-level `proxy/`. The bot PAT lives only in a local mitmproxy that the agent
+reaches over `HTTPS_PROXY`; the proxy injects it only for exact GitHub hosts
+and tunnels everything else. This authenticates API and git operations for any
+repository the bot account can access; credential isolation protects the PAT
+itself rather than restricting those operations to the triggering repository.
+Claude's Anthropic credential (OAuth token or API key) uses the same proxy and
+is injected only for `api.anthropic.com`. Codex's OpenAI key is instead read
+from stdin by OpenAI's hardened Responses API proxy, which exposes only
+`POST /v1/responses` on loopback. The agent holds only a dummy PAT and the local
```
tend-agent: The proxy's loopback surface is wider than this, and tend is what widens it: `codex/action.yaml` starts it with `--http-shutdown`, and the reap step curls `"$CODEX_PROXY_URL/shutdown"` to stop it. I ran the pinned `@openai/codex-responses-api-proxy@0.151.0` through this exact launch and `GET /shutdown` answers `200` with no auth, after which the process is gone. Nothing leaks — the sandbox UID can only end its own model access — but a reader checking the claim against the action finds a second endpoint the doc says isn't there.

```suggestion
from stdin by OpenAI's hardened Responses API proxy. It forwards only
`POST /v1/responses` upstream, and — because tend passes `--http-shutdown` for
the teardown step — also answers `GET /shutdown` on loopback. The agent holds only a dummy PAT and the local
```

The step comment above it ("accepts only `POST /v1/responses`") reads as a claim about what the key is ever attached to, which stays true either way; your call whether it wants the same qualifier.

tend-agent (commented):
One correction in the credential-isolation paragraph (inline). The base merge closes every open thread on this PR, and the delta itself reads clean.

<details><summary>What I verified about the merge's resolutions</summary>

I installed `@openai/codex-responses-api-proxy@0.151.0` (the pinned version) in this session and ran it through the exact launch shape the resolved step now uses:

- It starts fine under `/usr/bin/env -i` with no `HOME` — only `PATH` and `NODE_OPTIONS` — and logs one line, `responses-api-proxy listening on 127.0.0.1:<port>`, to the new `$PROXY_LOG_FILE` rather than the step's pipes.
- `--server-info` writes `{"port":45161,"pid":4750}` and nothing else, so `chmod 444` on it exposes no secret to the sandbox UID.
- `GET /shutdown` answers `200` and the process is gone a second later — the basis for the inline note.

The polling loop reaches its new failure branch rather than tripping `-e`: under `/usr/bin/bash --noprofile --norc -e -o pipefail`, a failing `[ -s "$F" ] && break` is exempt from errexit, so `cat "$PROXY_LOG_FILE" >&2` and `::error::OpenAI Responses proxy did not start` both run. Confirmed by running that exact fragment.

The `env -i` allowlist covers everything `setup_sandbox.py` reads — its only `os.environ` names are `TEND_RUNNER_TOOL_PATH`, `PATH`, `TEND_GH_TOKEN`, `MITMPROXY_VERSION`, `GITHUB_WORKSPACE`, `TEND_GITHUB_ONLY`, `TEND_SANDBOX_PATH`/`_ENV`, plus the `_required_env` set, all of which are passed.

`uv run --python 3.12 pytest` is 1046 passed on the merged tree.

</details>

docs/security-model.md
max-sixty: Fixed in 36cd0447. The security model now says the proxy forwards only `POST /v1/responses` upstream and answers `GET /shutdown` on loopback for Tend teardown; the action comment carries the same distinction.

tend-agent (approved):

codex/action.yaml
```diff
@@ -152,18 +179,163 @@ runs:
...
+          "$SERVER_INFO_FILE")
+        sudo chmod 444 "$SERVER_INFO_FILE"
+        sudo chown root:root "$SERVER_INFO_FILE"
+        echo "CODEX_PROXY_URL=http://127.0.0.1:$PROXY_PORT" >> "$GITHUB_ENV"
+
+    # Subscription consumers intentionally receive an expiring access token,
+    # never the rotating refresh token. Stage it only after the sandbox exists.
+    - name: Stage subscription auth (sandbox)
+      if: steps.codex_auth.outputs.mode == 'subscription'
+      shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
```
tend-agent, line 311: Every other step in this file using this shell zeroes `BASH_ENV`; this one has no `env:` block. `--noprofile --norc` doesn't cover it — non-interactive bash still sources `$BASH_ENV`, and this is the step that moves the subscription access token into the sandbox home as the runner user.

```suggestion
      shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
      env:
        BASH_ENV: ""
```

README.md
```diff
@@ -269,13 +288,12 @@ Two auth modes:
...
+experimental because it uses Codex's internal `chatgptAuthTokens` mode. The
+weekly job follows OpenAI's
 [CI/CD guidance](https://learn.chatgpt.com/docs/auth/ci-cd-auth): it runs
-Codex's built-in refresh and persists the updated full bundle. The same guide
-says not to use ChatGPT-account auth in public or open-source repositories.
-Tend pins and tests the Codex version, but an OpenAI change can still break the
-weekly refresh until Tend updates.
+Codex's built-in refresh and persists the updated full bundle. Tend pins and
+tests the Codex version, but an OpenAI change can still break the weekly refresh
+until Tend updates.
```
tend-agent: The guide this links to still says, verbatim: "Do not use this workflow for public or open-source repositories." Keeping the link but dropping the sentence leaves the paragraph citing a document whose most directly applicable line it no longer reports.

```suggestion
Codex's built-in refresh and persists the updated full bundle. That guide also
says not to use ChatGPT-account auth for public or open-source repositories.
Tend pins and tests the Codex version, but an OpenAI change can still break the
weekly refresh until Tend updates.
```

plugins/install-tend/skills/install-tend/references/security-model.md
```diff
@@ -120,13 +125,12 @@ every other runner with invalid state. Tend instead stores two projections:
...
-running Codex's built-in refresh and persisting its updated `auth.json`. The
-same guide says not to use ChatGPT-account auth for public or open-source
-repositories. Use a dedicated ChatGPT account and treat this as an experiment;
-`OPENAI_API_KEY` remains the supported path.
+The consumer path is experimental and may break when OpenAI changes Codex
+because it depends on an internal auth mode. The serialized weekly job follows
+OpenAI's [CI/CD auth guide](https://learn.chatgpt.com/docs/auth/ci-cd-auth) by
+running Codex's built-in refresh and persisting its updated `auth.json`. Use a
+dedicated ChatGPT account so the workflow's token rotation is independent of a
+maintainer's local Codex login.
```
tend-agent: Same restriction as in the README — and this is the file the install skill points the reader at for detail, so once it's gone from here the kickoff menu's pointer no longer leads anywhere that carries it.

```suggestion
running Codex's built-in refresh and persisting its updated `auth.json`. That
guide also says not to use ChatGPT-account auth for public or open-source
repositories. Use a dedicated ChatGPT account so the workflow's token rotation
is independent of a maintainer's local Codex login.
```

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -756,10 +765,10 @@ mode, keep it rather than prompting again:
...
 ```
 
-For **Plus/Pro subscription**, first repeat the unsupported-path warning from
-${CLAUDE_SKILL_DIR}/references/security-model.md. Use an isolated Codex login:
-the weekly workflow will rotate its refresh-token chain, so copying the user's
-ordinary `~/.codex/auth.json` would eventually break their local Codex login.
+For **Plus/Pro subscription**, explain that the path is experimental because it
+depends on Codex's internal auth mode. Use an isolated Codex login: the weekly
+workflow will rotate its refresh-token chain, so copying the user's ordinary
+`~/.codex/auth.json` would eventually break their local Codex login.
```
tend-agent, line 771: This is the moment the adopter commits to the subscription path, and most tend adopters are public repos. Worth stating the restriction here rather than only behind the reference link.

```suggestion
For **Plus/Pro subscription**, explain that the path is experimental because it
depends on Codex's internal auth mode, and that OpenAI's CI/CD auth guide says
not to use ChatGPT-account auth for public or open-source repositories. Use an
isolated Codex login: the weekly workflow will rotate its refresh-token chain,
so copying the user's ordinary `~/.codex/auth.json` would eventually break their
local Codex login.
```

tend-agent (commented):
Two things in the base merge's conflict resolutions.

**The docs drop OpenAI's public-repo restriction while still citing the guide that states it** (inline, three places). Resolving `README.md` and the install-tend security-model reference reworded "unsupported" to "experimental" and removed the sentence carrying upstream's actual restriction. I fetched the linked guide from this run — https://learn.chatgpt.com/docs/auth/ci-cd-auth still says verbatim: "Do not use this workflow for public or open-source repositories." Both files still link to it, so the summary now omits the line in it that most directly applies to the reader. The install skill lost it too: the kickoff menu's **Codex — Plus/Pro subscription** entry and step 7b now say only that the path depends on Codex's internal auth mode, and the reference they send the reader to no longer carries the restriction either. Public repos are tend's primary audience, so that is exactly the population being offered the choice without the caveat. Softening "unsupported" to "experimental" is a fair call about the mechanism; upstream's instruction about repo visibility is a separate fact and worth keeping alongside it.

**`Stage subscription auth (sandbox)` is the only hardened-shell step in either action without `BASH_ENV: ""`** (inline). Every other step using `/usr/bin/bash --noprofile --norc -e -o pipefail {0}` — nine in `codex/action.yaml`, nine in `claude/action.yaml` — zeroes it; this new one carries no `env:` block at all. `--noprofile --norc` does not cover `BASH_ENV`, which non-interactive bash still sources. So an adopter `setup:` that exports `BASH_ENV` into `$GITHUB_ENV` gets that file sourced in the one step that moves the subscription access token into the sandbox home, as the runner user with sudo. Same shape as the environment-inheritance thread on `Set up credential-isolation sandbox` earlier in this PR, on a step added after it.

The rest of the resolution holds up.

<details><summary>What I verified about the merge</summary>

`uv run pytest` on the merged tree: 1079 passed.

- **Auth-mode gating.** `prepare()` prefers `CODEX_AUTH_JSON` over `OPENAI_API_KEY`, matching the precedence the `openai_api_key` input description states. `Start OpenAI Responses proxy` is gated on `api-key`, `Stage subscription auth (sandbox)` on `subscription`, and `runner.py`'s `run_codex` raises on an unset or unknown `AUTH_MODE` rather than falling through to a modeless launch.
- **No token reaches the artifact.** `token_usage.py` copies only `agent_home/.codex/sessions`, and the upload path is `${{ runner.temp }}/tend-codex-logs/` — so neither `$AGENT_HOME/.codex/auth.json` nor a `$RUNNER_TEMP/tend-codex-auth.json` left behind by a failed `sudo install` is in the artifact's scope.
- **The staged auth resolves without `CODEX_HOME`.** The agent's `HOME` is `$AGENT_HOME`, so Codex finds `~/.codex/auth.json` where the step installs it, and `AGENTS.md` is staged into the same directory. `CODEX_HOME` joined `RESERVED_SANDBOX_ENV` in both `generator/src/tend/config.py` and `proxy/setup_sandbox.py` (the parity hook is satisfied), alongside `CODEX_API_KEY` and `CODEX_AUTH_JSON`, so an adopter cannot redirect it via `sandbox_env`.
- **Subscription traffic still reaches OpenAI.** `TEND_GITHUB_ONLY=1` narrows injection to GitHub hosts and leaves everything else an untouched tunnel, so a subscription session talking to ChatGPT's backend is unaffected by the absence of the Responses proxy.
- **The `ci.yaml` resolution keeps the right side.** It takes main's `shared/steps/codex_surface.py` over the branch's inline shell, and keeps the branch's `codex-responses-api-proxy` install. Nothing the deleted inline block covered was lost: exec flags matched as `--flag <`, the `Installed plugin root: ` prefix, `skills/triage/SKILL.md` and `scripts/list_recent_runs.py`, and the `--strict-config` negative control are all in the script, plus an access-only auth probe the shell version didn't have.
- **`_agent_step_inputs`** narrowing from `startswith("max-sixty/tend/")` to the exact `{claude, codex}` set is required now that `codex/refresh` exists and is not an operational agent action.
- **Doc plumbing.** No `#codex-alternative` anchor survives; the heading is `### Codex (experimental alternative)` and both links match. `codex_version` is `0.151.0` in both `codex/action.yaml` and `codex/refresh/action.yaml`.

</details>

codex/action.yaml
max-sixty: Fixed on aff78a72. The staging step now clears BASH_ENV, BASHOPTS, SHELLOPTS, and PS4 before bash starts, disables xtrace, and the static action-contract test asserts the complete scrub.

README.md
max-sixty: I am keeping the public-repository prohibition out intentionally. Tend permits this mode for public and open-source repositories and documents the actual boundary: a consumer session receives an expiring access token and can exfiltrate it, while the rotating refresh token and environment-writer PAT stay out of the session. I removed the external guide link from this paragraph as well, so the text no longer selectively cites a source whose recommendation differs from the accepted Tend design.

plugins/install-tend/skills/install-tend/references/security-model.md
max-sixty: I am keeping the public-repository prohibition out intentionally. The install security reference states the concrete exposure and mitigation instead: the live consumer can lose its expiring access token, but cannot refresh it or reach the refresh credential. I also removed the external guide link from this paragraph so the reference describes the Tend design directly.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: I am keeping that prohibition out intentionally because Tend permits subscription auth for public and open-source repositories. The kickoff text calls the path experimental, requires an isolated login, and the linked security reference now states the actual token exposure and refresh-token isolation without presenting this use as disallowed.

tend-agent (commented):
The `BASH_ENV` gap closes, and the boundary-test fix is right — `test-sandbox` was genuinely red on the previous head with `codex runner: unknown AUTH_MODE: <unset>`, and `AUTH_MODE=api-key` is what `run_codex` needs to take the proxy branch. One thing the new `env:` block doesn't cover.

**`Stage subscription auth (sandbox)` still resolves `sudo` and `rm` through the runner's PATH** (inline). It is now the only step in either action that uses `/usr/bin/bash --noprofile --norc -e -o pipefail {0}` and neither pins `PATH=/usr/sbin:/usr/bin:/sbin:/bin` nor fully-qualifies its binaries. The three siblings carrying the same four-variable block plus `set +x` — `Configure Codex auth`, `Set up credential-isolation sandbox`, `Start OpenAI Responses proxy` — all pin it, and the one other unpinned hardened step, `claude/action.yaml`'s `Remove experimental memory Gist working copy`, writes `/usr/bin/sudo /usr/bin/rm` instead. Same shape as the `BASH_ENV` thread one layer down: `--noprofile --norc` covers neither, and this is the step that moves the access token into the sandbox home.

Smaller, and not a blocker: both this and the `BASH_ENV` gap were found by reading a new step against its siblings, because `test_repo_pins.py` asserts the hardening at three named steps rather than over the set. One test iterating every step whose `shell` contains `--noprofile` and asserting the convention would have caught both, and is less to maintain than a fourth hand-written assertion.

codex/action.yaml
```diff
@@ -152,18 +179,169 @@ runs:
...
+      env:
+        BASH_ENV: ""
+        BASHOPTS: ""
+        SHELLOPTS: ""
+        PS4: ""
+      run: |
+        set +x
```
tend-agent, line 318: An adopter `setup:` step that writes to `$GITHUB_PATH` prepends to PATH for every later step in the job, composite-action steps included — so `sudo` here can resolve to an adopter-planted binary, which is then handed `install -o "$SANDBOX" -g "$SANDBOX" -m 600 "$RUNNER_TEMP/tend-codex-auth.json" ...` and runs as the runner user. (`mkdir` and `install` go through sudo's own `secure_path`, so `sudo` and the bare `rm` on the last line are the two that resolve through the inherited PATH.) The pin every other credential-handling step in this file already uses:

```suggestion
        set +x
        PATH=/usr/sbin:/usr/bin:/sbin:/bin
```

`sudo`, `mkdir`, `install` and `rm` are all under `/usr/bin`, and the test's `run.startswith("set +x\n")` assertion still holds.

max-sixty: Fixed in `ca8f4362`. The subscription-auth staging shell now pins `PATH=/usr/sbin:/usr/bin:/sbin:/bin` immediately after disabling xtrace, before it resolves `sudo` or `rm`. The regression now checks every hardened Codex shell step: startup variables are scrubbed, one-line steps invoke a fully qualified binary under the safe PATH, and multiline steps establish either `env -i` or the safe PATH before running commands.

generator/tests/test_repo_pins.py
```diff
@@ -55,6 +55,101 @@ def test_claude_transcript_summary_is_opt_in() -> None:
...
+        (REPO_ROOT / "codex" / "action.yaml").read_text()
+    )
+    safe_path = "PATH=/usr/sbin:/usr/bin:/sbin:/bin"
+
+    for step in action["runs"]["steps"]:
+        if "--noprofile" not in step.get("shell", ""):
+            continue
+
+        env = step.get("env", {})
+        assert env["BASH_ENV"] == "", step["name"]
```
tend-agent: A step arriving with no `env:` block at all is how this regression actually shows up — it is how `Stage subscription auth (sandbox)` arrived — and on that shape this line raises `KeyError: 'BASH_ENV'` before `step["name"]` is attached, so the failure never says which step. `env.get` keeps the message. The `BASHOPTS`/`SHELLOPTS`/`PS4` comprehension below has the same shape.

```suggestion
        assert env.get("BASH_ENV") == "", step["name"]
```

generator/tests/test_repo_pins.py
```diff
@@ -55,6 +55,101 @@ def test_claude_transcript_summary_is_opt_in() -> None:
...
+
+        assert {name: env[name] for name in ("BASHOPTS", "SHELLOPTS", "PS4")} == {
+            "BASHOPTS": "",
+            "SHELLOPTS": "",
+            "PS4": "",
+        }, step["name"]
+        assert lines[0] == "set +x", step["name"]
+        assert lines[1] == safe_path or lines[1].startswith("/usr/bin/env -i"), step[
+            "name"
+        ]
```
tend-agent, line 160: A hardened step added to `claude/action.yaml` without `BASH_ENV: ""` leaves the suite green today. The shape half above doesn't parametrize — claude's memory-Gist steps fully-qualify their binaries instead of pinning `PATH` — but `BASH_ENV: ""` holds for every hardened step in both files, so it can be asserted over both:

```suggestion
        assert lines[1] == safe_path or lines[1].startswith("/usr/bin/env -i"), step[
            "name"
        ]


@pytest.mark.parametrize("harness", ["claude", "codex"])
def test_hardened_shells_never_source_bash_env(harness: str) -> None:
    action = YAML(typ="safe", pure=True).load(
        (REPO_ROOT / harness / "action.yaml").read_text()
    )

    for step in action["runs"]["steps"]:
        if "--noprofile" in step.get("shell", ""):
            assert step.get("env", {}).get("BASH_ENV") == "", step["name"]
```

tend-agent (commented):
The delta does what the last thread asked, and the `PATH` pin is right — assigning to an already-exported `PATH` applies to the rest of the step, matching the three siblings. One gap in the new test.

**The set-wide guard stops at `codex/action.yaml`** (inline). `claude/action.yaml` has ten steps on the same `/usr/bin/bash --noprofile --norc -e -o pipefail {0}` shell — including the two memory-Gist steps that hold the PAT — and nothing asserts the convention over that set; only `Set up credential-isolation sandbox` is checked, by `test_privileged_sandbox_launch_scrubs_adopter_runtime_configuration`. I checked both directions from this run by injecting a hardened step carrying no `env:` block: in `codex/action.yaml` the new test fails, in `claude/action.yaml` the whole suite stays green. That is the same shape as the `BASH_ENV` gap on `Stage subscription auth (sandbox)` one thread down, in the file this test doesn't reach.

The command-resolution half is genuinely Codex-specific: claude's three memory-Gist steps satisfy it by fully-qualifying every binary (`/usr/bin/sudo /usr/bin/rm`) rather than pinning `PATH`, so parametrizing this test whole fails on them. The `BASH_ENV: ""` half is universal — it holds for all nineteen hardened steps across both files today — so it parametrizes as-is, which is what the inline suggestion adds.

**`env["BASH_ENV"]` raises before the assertion message** (inline). Neither is a release blocker.

<details><summary>What I verified</summary>

- Injected `- name: Probe unhardened step` with `shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}`, a two-line `run`, and no `env:` block. In `codex/action.yaml`: `1 failed, 20 passed`, failing as `KeyError: 'BASH_ENV'`. In `claude/action.yaml`: `21 passed`. Both files restored afterwards.
- With both suggestions applied: `uv run --python 3.12 pytest generator/tests/test_repo_pins.py` is `23 passed`, `ruff format --check` reports the file already formatted, and `ruff check` passes. Re-running the `claude/action.yaml` probe against the patched suite fails on the new test, with `Probe unhardened step` named in the output.
- The three assertions this commit deletes are all re-established by the loop: `set +x` and the `PATH` line for `Start OpenAI Responses proxy` become `lines[0]`/`lines[1]`, its four scrubbed variables become the `env` checks, and `Stage subscription auth`'s `set +x` becomes `lines[0]`. `subscription["env"]` is still asserted exactly in the named test above.

</details>

generator/tests/test_repo_pins.py
max-sixty: Fixed in `97912a7d`. `BASH_ENV` scrubbing is now a separate parametrized invariant over every hardened shell step in both `claude/action.yaml` and `codex/action.yaml`; the Codex-only test retains the stricter command-resolution check appropriate to its multiline steps.

generator/tests/test_repo_pins.py
max-sixty: Fixed in `97912a7d`. The cross-harness `BASH_ENV` assertion and the Codex startup-variable assertions now use `.get(...)`, so a missing block reports the offending step name instead of raising `KeyError`.

tend-agent (approved):

README.md
tend-agent: Understood — your call, and I'm not pressing it. The selective-citation half is resolved: at `97912a7d` the guide link is gone from this paragraph, and `learn.chatgpt.com` no longer appears in `plugins/install-tend/skills/install-tend/references/security-model.md`, `plugins/install-tend/skills/install-tend/SKILL.md`, or `docs/security-model.md` either — so nothing cites a document whose recommendation the design departs from. This covers the two sibling threads with the same reply.

PR #1161, opened 2026-09-06
Add Claude effort and harness arguments

max-sixty:
Tend exposed reasoning effort only for Codex and had no escape hatch for harness-specific CLI options. This adds one configuration surface shared by both harnesses: `effort` is validated against each CLI's supported levels, and `args` passes exact additional arguments globally or per workflow without shell evaluation.

The generator, composite actions, both Python launchers, examples, and install guidance now carry both settings. The tests cover defaults, inheritance and overrides, invalid values, Claude Haiku's unsupported effort setting, and arguments containing spaces.

Checks: `wt test`; `uv tool run pre-commit run --all-files`; local Claude and Codex `--help` parser smoke tests.


generator/src/tend/config.py
```diff
@@ -589,6 +605,20 @@ def load(cls, path: Path | None = None) -> Config:
                         f"allowlist and likely won't apply to {wf_harness}. "
                         f"Set `workflows.{name}.model:` to a valid {wf_harness} model."
                     )
+                if (
+                    wf_harness is not None
+                    or wf_model is not None
+                    or wf_effort is not None
+                ):
```
tend-agent: Two small things about this block.

When the workflow sets `effort:` itself, `wf_effort` was already validated against this same effective `(harness, model)` a few lines up, so this call repeats it verbatim. Narrowing the guard to the inherited case drops the duplicate and leaves the branch meaning one thing:

```suggestion
                if wf_effort is None and (
                    wf_harness is not None or wf_model is not None
                ):
```

That leaves the label. Both messages name `workflows.<name>.effort` even when only the top level set it — top-level `effort: max` with `workflows.nightly.harness: codex` gives `workflows.nightly.effort 'max' is not recognized for harness 'codex'`, and top-level `effort: high` with `workflows.nightly.model: haiku` gives `workflows.nightly.effort is not supported for Claude model 'haiku'; drop the key to use that model`. Neither config has that key, and on the haiku path there is nothing to drop — the fixes are removing the top-level `effort:` (which hits every other workflow) or setting `workflows.nightly.effort: ""`. Worth having `_parse_effort` say when the value was inherited; the `match=` regexes in `test_cross_harness_workflow_validates_inherited_effort` and `test_effort_rejected_for_claude_models_without_effort` would move with it.

tend-agent (commented):
One note on the effective-effort validation in `config.py`, inline.

generator/src/tend/config.py
max-sixty: Fixed in `36c6f8fa`: inherited effort is now revalidated only when a workflow changes the harness or model, and the error names both the top-level source and the consuming workflow. The Haiku error now gives the two applicable remedies.

tend-agent (approved):

PR #1165, opened 2026-09-07
chore: post-0.2.0 regeneration and release-skill fixes

max-sixty:
Post-release cleanup for
[0.2.0](https://github.com/max-sixty/tend/releases/tag/0.2.0), in three
commits.

**Regenerate the workflows.** `uvx tend@latest init` on the released
0.2.0, moving the harness pin from `max-sixty/tend/claude@0.1.24` to
`@0.2.0`. The regenerated files pick up the runtime `enabled` gate each
job reads before checkout, the `effort` and `args` inputs, and the
mention and notification gates that moved into Python. Generated output
only, no hand edits.

**Restamp `review-reviewers.yaml`.** `init` rewrites only the generated
`tend-*.yaml` files, so the hand-maintained workflow held `@0.1.24`. It
is dispatch-only and serves as the post-release spot-check for
fleet-wide behavior changes, so leaving it behind would run the
superseded action against the release it validates.

**Fix the release skill**, porting worktrunk's release step 12. The
commit range is read early but consumed at the tag, and the tag and the
PyPI version are both immutable, so a commit landing on `main` during
the PR's CI wait ships uncited with no cheap recovery. Cutting 0.2.0 hit
exactly this: a direct push landed mid-draft and never appeared in
`gh pr list`, so a PR-based cross-check would not have caught it. Step 1
now records the cut-from tip; step 9 lists `<cut-from>..origin/main`,
requires the changelog to cover it, and tags `origin/main` so the ref
checked and the ref tagged are the same. Step 11 also points at
`running-tend`'s restamp rule.

Until this merges, tend's deployed workflows run the 0.1.24 action
against tend itself.

Checks: `uv run pytest` (1082 passed) and
`uv tool run pre-commit run --all-files`, both on the merged tree.

https://claude.ai/code/session_01Lqh3N3TNWgQRqRqdAWZKBj


tend-agent (commented):
The regeneration itself checks out: I re-ran the in-tree generator against the merged tree (the base commit *is* the `0.2.0` tag, so it is the released generator) and the output is byte-identical to what's committed here — no hand edits. `max-sixty/tend/claude@0.2.0` resolves, every job that gained the `Check whether tend is enabled` step already declares `contents: read` or better for the `gh api .../contents/.config/tend.yaml` call it makes, and the step runs before checkout without needing one. I also ran the generated Ruby gate against this repo's live `.config/tend.yaml` — it exits 0 with `enabled=true`, so merging won't quietly park the workflows.

One gap: `.github/workflows/review-reviewers.yaml` still pins [`max-sixty/tend/claude@0.1.24`](https://github.com/max-sixty/tend/blob/44ab6ff5eae3c5273d158e08e5d78e6a91beed99/.github/workflows/review-reviewers.yaml#L91). `init` doesn't rewrite it, and `.claude/skills/running-tend/SKILL.md` under **Nightly: restamp the hand-maintained workflow refs** says to restamp the hand-maintained files onto the generated files' ref and fold it into the regen PR — same worktree, same commit. Every prior regen commit did exactly that (e47ccc4 `0.1.21`→`0.1.22`, ac14592 `0.1.20`→`0.1.21`). Leaving it behind is more than cosmetic here: `review-reviewers` is dispatch-only and the skill positions the dispatch as the post-release spot-check for fleet-wide behavior changes, so it would run the superseded action on precisely the release it is meant to validate.

The fix is the single line above. Happy to push it onto this branch if you'd like — say the word.

<details><summary>What the second pass cleared</summary>

- **`verify` job permissions narrowed.** `tend-mention.yaml`'s `verify` gained `permissions: contents: read` where it previously inherited the repo default. Its only other step authenticates with `TEND_BOT_TOKEN`, so nothing else in the job reads `github.token`.
- **Disabled-path fallthrough.** With `enabled != 'true'`, `verify`'s check step is skipped, `should_run` is empty, and `handle`'s `needs.verify.outputs.should_run == 'true'` is false — the job skips rather than running ungated.
- **Eyes reaction symmetry.** The `React with eyes` and `Remove the eyes reaction` steps are gated on the same expression, and a step output can't change within a job, so there's no path that adds the reaction and leaves it stuck.
- **bash → Python gate rewrites** (`tend-mention` engagement, `tend-notifications` work check) preserve the prior semantics, including the load-bearing one that a failing `gh api` stays fatal: `CalledProcessError` propagates to `SystemExit(returncode or 1)` rather than falling through to `should_run=false`.
- **`--paginate` handling.** Both new helpers decode concatenated JSON pages and extend, so the old "one count per page" hazard the bash comments warned about doesn't return. An empty result set still decodes as `[]` rather than tripping the fetch-failed path.
- **Pre-checkout `setup-uv`.** The instances that run before checkout carry `ignore-empty-workdir: true`.
- **`ci-fix` concurrency.** The new job-level group `${{ github.workflow }}-<watched workflow>-<branch>` matches the shape CLAUDE.md describes under **Concurrency and filtering**.

</details>

tend-agent (approved):

tend-agent (approved):

Issue #1164, opened 2026-09-06
Bot temporarily unavailable


tend-agent:
The bot failed to process a request. This issue tracks failures until the underlying cause is resolved.

| When | Run | Trigger |
|------|-----|---------|
| 2026-09-06T23:14:01Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/34065324135) | #1162 |

This issue was created automatically. Close it after every row above is diagnosed and the live repository has been checked for work the failed runs may have missed. The `tend-review-runs` sweep does that where it runs.

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-09-06T23:17:23Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/34066282339) | #1163 |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-09-06T23:17:43Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/34065388982) | #1160 |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-09-06T23:21:17Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/34066530023) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-09-07T01:04:05Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/34071673700) | N/A |

tend-agent:
| When | Run | Trigger |
|------|-----|---------|
| 2026-09-07T01:04:38Z | [workflow run](https://github.com/max-sixty/tend/actions/runs/34071757673) | #1154 |

tend-agent:
### [Run 34065324135](https://github.com/max-sixty/tend/actions/runs/34065324135)

No failure details could be extracted.



### [Run 34065388982](https://github.com/max-sixty/tend/actions/runs/34065388982)

No failure details could be extracted.



### [Run 34066282339](https://github.com/max-sixty/tend/actions/runs/34066282339)

No failure details could be extracted.



### [Run 34066530023](https://github.com/max-sixty/tend/actions/runs/34066530023)

No failure details could be extracted.



### [Run 34071673700](https://github.com/max-sixty/tend/actions/runs/34071673700)

#### notifications

```
claude -p exited non-zero (exit=1): You've hit your session limit · resets 1:10am (UTC) — see the session-logs artifact
[...]

PR #1166, opened 2026-09-07
Read the failing attempt when enriching outage rows, so a rerun-green run keeps its diagnosis

tend-agent:
Four of the six rows on the current outage tracker (#1164) say `No failure details could be extracted.` — and all four are runs whose failing attempt still carries the annotation naming the cause.

A `tend-outage` row is recorded the moment a run fails. If that run is then rerun and goes green, both of the enrichment's sources go blank against the latest attempt: `GET /actions/runs/{id}/jobs` defaults to `filter=latest`, whose jobs all succeeded, and `gh run view --log-failed` defaults to the same latest attempt and returns no failed log. The tracker ends up asserting nothing is knowable about precisely the runs where the diagnosis is one query parameter away.

This reads both sources at the attempt that actually failed.

**Annotations.** `filter=all` returns every attempt's jobs. Against run [34065324135](https://github.com/max-sixty/tend/actions/runs/34065324135) — one of the four stubs — the attempt-1 job yields `claude -p exited non-zero (exit=1): You've hit your session limit`, which is the whole diagnosis and, in this case, the reason the whole night's rows exist. Job headings gain an `(attempt N)` suffix only when the run has more than one attempt, so single-attempt runs render byte-identically to today.

**Log tail.** The fallback now passes the latest failing attempt to `gh run view --attempt`. This is the source that carries transient failures specifically: a `curl: (35) Recv failure` in "Run the action" annotates only `Process completed with exit code N`, which the annotation path discards as uninformative, so its diagnosis lives solely in the log — and a transient failure is exactly the kind that gets rerun to green. The attempt comes from the jobs response the annotation path already reads, so it costs no extra call. When that read is unavailable there is no attempt to name and the log source stays on the latest attempt, as before.

The four stubs already on #1164 stay stubs — their `<!-- enriched-run:ID -->` markers suppress a retry by design. Deleting those four marker comments after this merges would let the next nightly re-enrich them.

Regression tests, both failing on `main`: a run whose attempt 1 failed and attempt 2 succeeded renders the attempt-1 annotation instead of the stub; and one whose attempt-1 failure is annotation-free renders its log tail instead of the stub.

### DIFF
--- a/generator/tests/test_enrich_tend_outage_issues.py
+++ b/generator/tests/test_enrich_tend_outage_issues.py
@@ -59,7 +59,10 @@ case "$*" in
   *"/annotations"*) emit "$(cat "$ANNOTATIONS_JSON")" ;;
   "run view"*--log-failed*)
     if [ -n "${LOG_FAILS:-}" ]; then exit 1; fi
-    cat "$LOG_TXT"
+    case "$*" in
+      *--attempt*) cat "$LOG_TXT" ;;
+      *) cat "${LATEST_LOG_TXT:-$LOG_TXT}" ;;
+    esac
     ;;
   "issue comment"*)
     prev=""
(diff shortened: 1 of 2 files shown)


tend-agent (commented):
`filter=all` recovers the four stubs on #1164 — each of those runs' attempt-1 jobs carries the `claude -p exited non-zero … session limit` annotation this now renders, so the central claim holds.

The log fallback still reads only the latest attempt, so the other half of the blanking this PR describes stays open. `log_details` shells out to `gh run view <id> --log-failed` with no attempt argument, which resolves to the rerun's green attempt and yields no failed log; `failure_details` then has neither source and renders the stub. That path carries the transient failures specifically: a `curl: (35) Recv failure` in "Run the action" produces only `Process completed with exit code N`, so its diagnosis lives solely in the log — and a transient failure is exactly the kind that gets rerun to green. `gh run view` takes `-a/--attempt`, and the failing attempt is already in the jobs response this PR now reads, so passing it through closes it.

Worth either closing here or stating in the PR body that the log source remains latest-only, so the next reader of a stub row doesn't re-derive the same diagnosis.

tend-agent:
Closed it here rather than documenting it — the attempt number was already in hand, so leaving the log source on `latest` would have shipped half the fix.

`log_details` now takes the latest failing attempt and passes it to `gh run view --attempt`. The jobs response is read once in `failure_details` and shared with both sources, so this costs no extra call. Confirmed against run [34065324135](https://github.com/max-sixty/tend/actions/runs/34065324135), whose attempt 2 went green: `--log-failed` alone returns nothing, while `--log-failed --attempt 1` returns the log ending at `##[error]claude -p exited non-zero (exit=1): You've hit your session limit`.

One case worth naming, since it decides the fallback's shape: when the jobs read itself fails, there is no attempt to name, so the call omits `--attempt` and reads the latest attempt exactly as before. That keeps a broken jobs endpoint from taking the log source down with it — covered by its own test.

New regression test drives the transient path you identified: attempt 1 failed with only a `Process completed with exit code N` annotation, attempt 2 green, and the log available solely under `--attempt 1`. It renders the `curl: (35) Recv failure` tail; on the previous commit it renders the stub. PR body rewritten to describe both sources.

PR #1170, opened 2026-09-07
skills(review-runs): unscope the live-work default-branch CI check from ci-fix's watched workflows

### LINKED ISSUE #1169: review-runs: the live-work sweep's default-branch CI check inherits ci-fix's watched-workflow scope, so an unwatched failure is never seen
`review-runs`'s live-work sweep lists "failing default-branch CI with no bot fix in progress" ([`review-runs/SKILL.md:217`](https://github.com/max-sixty/tend/blob/main/plugins/tend-ci-runner/skills/review-runs/SKILL.md)) without saying how to query it. Nothing in the bundled skills defines what "default-branch CI" covers, so the session takes the definition from whatever the adopter's overlay names as the repo's CI — which is the workflow list under `workflows.ci-fix.watched_workflows`. Failures on the default branch from any *other* workflow are then invisible to the sweep, indefinitely, because no scheduled tend job looks anywhere else: `ci-fix` is gated on `workflow_run` for the watched workflows by design, and nightly's health checks don't query the default branch's run history at all.

The failures this drops are exactly the ones with no PR attached to carry them into view — Dependabot's scheduled security-update jobs, cron releases, doc builds. They surface in the Actions tab and nowhere else.

## Evidence — a moderate advisory sat undiagnosed for three days on `PRQL/prql`

`.config/tend.yaml` there sets `watched_workflows: [tests]`, and the repo's overlay opens its CI section with "Main CI workflow: `tests` (watched by tend-ci-fix)". Dependabot's security update for `time` (GHSA-r6v5-fh4h-64xc, moderate) failed on `main` three times as `security_update_not_possible`:

| run | created | conclusion |
|---|---|---|
| [33786752555](https://github.com/PRQL/prql/actions/runs/33786752555) | 2026-09-03T17:48:42Z | failure |
| [33913437130](https://github.com/PRQL/prql/actions/runs/33913437130) | 2026-09-04T19:53:09Z | failure |
| [34015531593](https://github.com/PRQL/prql/actions/runs/34015531593) | 2026-09-06T06:03:56Z | failure |

Five consecutive `review-runs` entries in that repo's tracking issue ([PRQL/prql#6266](https://github.com/PRQL/prql/issues/6266)) report the default branch as healthy in the form "`main` is green: `tests` succeeded at `<sha>`" — accurate as scoped, and blind to the rows above. The first two failures fall inside windows that reported it that way. The third was found by the 2026-09-06 `review-runs` session, on day three; diagnosing it then took that one session and produced a real result (every `time` release clearing the advisory declares `rust-version = 1.88.0` against a workspace MSRV of 1.85, so there is no lockfile-only remedy — it needs a maintainer decision).

This is delayed detection rather than a wrong outward action: no claim posted was false, because each was correctly scoped to `tests`. The cost is that the one signal the sweep exists to catch was the one it couldn't see.

## Proposed fix

One line at `review-runs/SKILL.md:217`, giving the query and saying explicitly that it is not scoped to the watched workflows:

````md
- failing default-branch CI with no bot fix in progress. This is **not** scoped to `ci-fix`'s watched workflows — Dependabot security updates, cron releases and doc builds fail on the default branch with no PR attached, and nothing else looks for them:

  ```bash
  DEFAULT_BRANCH=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')
  gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=failure&per_page=30" \
    --jq ".workflow_runs[] | select(.created_at >= \"$(cat /tmp/review-runs-since)\") | {id, name, created_at, event}"
  ```
````

Scoping the report the same way is worth a sentence too — "`main` is green" should either name the workflow it checked or cover all of them, so a future window's entry can be read at face value.

Related but distinct: [#1035](https://github.com/max-sixty/tend/issues/1035) covers a *watched* workflow whose `cancelled` conclusion drops it from `ci-fix`'s gate. This one is about workflows that were never watched.

### PR DESCRIPTION
**Pull request by tend-agent**
## Problem

`Reconcile live work` lists "failing default-branch CI with no bot fix in progress" with no query and no statement of scope, so a session infers the scope from the only definition of "the repo's CI" it has: `workflows.ci-fix.watched_workflows` and whatever the adopter's overlay calls the main CI workflow. Failures from any other workflow on the default branch are then seen by nothing — `tend-ci-fix` is `workflow_run`-gated on the watched list by design, and nightly never queries the default branch's run history.

The failures that class drops are the ones with no PR to carry them into view. On this repo right now, `main` carries recurring Dependabot security-update failures (`cryptography`, `h2`, `msgpack`, `tornado`) that no tend job has ever looked at; #1169 reports the same shape on `PRQL/prql`, where a moderate advisory sat undiagnosed for three days while five consecutive `review-runs` entries reported `main` green — each accurate as scoped to `tests`, and blind to the rows beside it. A Dependabot update run cannot be brought into scope by configuration either: its run name carries a per-update ID (`uv in /. for h2 - Update #1559215022`), so no static `workflows:` list matches it.

## Solution

The bullet now states that the check is scoped neither to `ci-fix`'s watched workflows nor to this run's window, and carries the query. Both dimensions matter, and the second is what makes the first stick: #1169's shape is a failure that runs once and then sits, so a query filtered on `/tmp/review-runs-since` would surface such a row on the single night it ran and read as an all-clear on the four nights after — the same over-broad "`main` is green" claim in the time dimension. The bullet is therefore unwindowed like the two live-state bullets above it, and a row is closed out by a fix rather than by age.

Two calls, because an unwindowed page needs both:

- **Detection.** `status=failure` returns conclusion `failure` alone, so the loop covers the three red conclusions (`failure`, `startup_failure`, `timed_out`) under their own server-side filters. A run that never started or was killed at its cap is red on the default branch with no PR attached — the same class the bullet exists to catch. `cancelled` is deliberately excluded: concurrency cancels dominate it and are not failures. Filtering server-side is what makes the query usable on a busy repo — an unfiltered page of the last 100 default-branch runs covers about 21 hours on `PRQL/prql` and never reaches the Dependabot rows, while 50 failed runs reach back a month.
- **Closure.** The cost of that reach is that already-fixed rows stay in the page, and red-only results carry no evidence that they are fixed — a session would then re-derive liveness per row or open a tracker for a month-old fixed one, which is the noisy mirror of the all-clear this bullet exists to prevent. So the block also asks for each affected workflow file's latest *green* default-branch run: it closes that path's older red rows in one call. Dependabot's security updates have no workflow file to query (`dynamic/dependabot/dependabot-updates` 404s at that endpoint) and each update's `name` carries an ID that never recurs, so those rows close only through a fix PR or a tracker.

The bullet also asks that the health report carry the scope it was checked at, in both dimensions. Its prose is ~200 words against two one-line siblings; the rationale the code block encodes lives in comments there rather than in prose, per CLAUDE.md's "Be brief" / "Prefer recipe over narrative".

## Testing

Extracted both snippets verbatim from the committed file and ran them against this repo. Detection returns 50 `failure` rows reaching back to 2026-08-05 — including all twelve Dependabot security-update failures across 09-05 and 09-06, the rows spanning more than one night that the windowed form would have shown once and then dropped — plus the one `startup_failure` row (`ci`, 2026-08-26) that a `status=failure`-only query misses entirely; `timed_out` returns none. Closure returns `ci.yaml`'s latest green at 2026-09-07T03:36:45Z, which closes the three 08-12 `ci` failures and that `startup_failure` row, and 404s on the dynamic Dependabot path as described.

`pre-commit` passes on the file, including the bang-backtick and install-tend mirror guards; `generator/tests/test_cross_file_contracts.py` passes (18 passed).

Closes #1169 — automated triage

### CODE REVIEW

File: plugins/tend-ci-runner/skills/review-runs/SKILL.md [hunk 1]
```diff
@@ -88,7 +88,15 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
-- failing default-branch CI with no bot fix in progress.
+- failing default-branch CI with no bot fix in progress. This is **not** scoped to `ci-fix`'s watched workflows — Dependabot security updates, cron releases and doc builds fail on the default branch with no PR attached, and nothing else looks for them. `status=failure` filters server-side, so the page reaches past a busy repo's green runs:
+
+  ```bash
+  DEFAULT_BRANCH=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')
+  gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=failure&per_page=50" \
+    --jq ".workflow_runs[] | select(.created_at >= \"$(cat /tmp/review-runs-since)\") | {id, name, path, event, created_at}"
+  ```
+
+  Tend's own failures are already in Step 1's census; the news here is every other row. Report the branch at the scope you checked — "`main` is green" is read later as covering every workflow, so name the workflows the claim rests on.
```
tend-agent: The empty-result case is the one this instruction has to cover, and the query can't distinguish "no workflow failed" from "no workflow ran". Naming the window alongside the workflows keeps the report honest without changing the query.

```suggestion
  Tend's own failures are already in Step 1's census; the news here is every other row. Report the branch at the scope you checked, in both dimensions — the query is windowed on the anchor, so an empty result means nothing newly failed since then, not that the branch is green. Name the workflows and the window the claim rests on.
```

tend-agent (commented):
The workflow dimension of the scope is now explicit, but the time dimension went the other way: the query is filtered on `/tmp/review-runs-since`, so it answers "what failed since the last successful `review-runs` run", while the bullet it sits under asks about a state — "failing default-branch CI with no bot fix in progress". A default branch that is red from a run older than the anchor, with nothing pushed or cron-fired since, returns zero rows. The instruction added on the last line then reads the empty result as a workflow-scope question only, so the session's honest all-clear ("nothing newly failed") is reported as "`main` is green" — the same over-broad claim #1169 is about, closed in the workflow dimension and left open in the time one.

Nothing else in **Reconcile live work** is windowed: the two sibling bullets are live-state checks with no `since` at all.

Suggestion inline to name the window alongside the workflows, which keeps the server-side `status=failure` filter and the anchor exactly as they are.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Agreed on the diagnosis, and took it a step further than the suggestion: the window is gone from the query rather than named in the report.

Naming it keeps the claim honest but leaves the branch undetected, and detection is what the bullet is for. #1169's shape is a failure that runs once and then sits — an advisory undiagnosed for three days across five consecutive reports. Anchored on `/tmp/review-runs-since`, this query catches that row on night one and returns zero on nights two through five; the caveat would make those four nights read as "nothing newly failed" instead of "`main` is green", which is more honest but still leaves the red branch to a human. Unwindowed, the row is present every night until something fixes it, which matches the two live-state bullets above it.

The trade is that already-fixed failures stay in the page, so the bullet now says a row is closed out by a fix — a later green run of the same job, a fix PR, an open tracker — not by age. Your report instruction is in, covering both dimensions.

Run verbatim against this repo, the unwindowed query returns 50 rows reaching back to 2026-08-05, including all twelve Dependabot security-update failures across 09-05 and 09-06; the windowed form reaches 09-06 08:40 and drops the 09-05 rows entirely.

Pushed as 8234a269796d760c290c1705e007d51271685e93.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -88,7 +88,15 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
+
+  ```bash
+  DEFAULT_BRANCH=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')
+  gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=failure&per_page=50" \
+    --jq '.workflow_runs[] | {id, name, path, event, created_at}'
+  ```
+
+  Deliberately unwindowed: a failure nobody fixed is still live on the nights after it ran, so anchoring this on `/tmp/review-runs-since` would surface each one on the single night it happened and report the rest as an all-clear. A row is closed out by a fix — a later green run of the same job, a fix PR, an open tracker — not by age. Tend's own failures are already in Step 1's census; the news here is every other row. Report the branch at the scope you checked: "`main` is green" is read later as covering every workflow, so name the workflows the claim rests on and how far back the page reached.
```
```suggestion
  Deliberately unwindowed: a failure nobody fixed is still live on the nights after it ran, so anchoring this on `/tmp/review-runs-since` would surface each one on the single night it happened and report the rest as an all-clear. A row is closed out by a fix — a later green run of the same `name`, a fix PR, an open tracker — not by age. Match the run `name`, not `path`: Dependabot's updates all share one path, so a green sibling update would close a live advisory, and each update's `name` carries its own ID and so never recurs — those rows close only through a fix PR or a tracker. Tend's own failures are already in Step 1's census; the news here is every other row. Report the branch at the scope you checked: "`main` is green" is read later as covering every workflow, so name the workflows the claim rests on, how far back the page reached, and that the page holds conclusion `failure` alone — a run that never started or was killed at its timeout is red and absent from it.
```

tend-agent (commented):
The unwindowed query is the right call for the detection this bullet is for, but the rule that replaces the window — "a row is closed out by a fix — a later green run of the same job" — is undefined for the exact class it was written to catch. The query returns `{id, name, path, event, created_at}`, and on this repo's twelve failing Dependabot security-update rows all twelve share `path: dynamic/dependabot/dependabot-updates` while each `name` carries a unique per-update ID (`uv in /. for h2 - Update #1559215022`). A session keying "same job" on `path` closes a live advisory as soon as any sibling update runs green; keying on `name` never closes the row at all, because that name never recurs. Both readings are available from the text, and the first is the one that reproduces #1169's failure — a green-looking all-clear over an undiagnosed advisory.

Second, smaller: the report instruction now asks for the workflow and time dimensions of the scope, but the server-side filter adds a third. `status=failure` returns conclusion `failure` alone — verified against this repo, where the 50-row page is uniformly `failure`. A run that never started (`startup_failure`, which a bad generated workflow file produces) or one killed at its timeout is red on the default branch, in the same "no PR attached and nothing else looks for it" class, and is absent from the page a "`main` is green" claim would rest on.

Suggestion inline covers both, keeping the query and the unwindowed decision as they are.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Both confirmed against this repo, and both fixed in 171d4df — the first as suggested, the second one step further than the suggestion.

**Row closure keys on `name`.** Taken as written. The twelve Dependabot rows do share `path: dynamic/dependabot/dependabot-updates` with a unique per-update `name` each, so `path` is the reading that closes a live advisory on a green sibling — #1169's failure reproduced inside the rule meant to prevent it. The bullet now names `name` in the closure clause and says those rows never recur under it, so they close only through a fix PR or a tracker.

**Missing conclusions get queried, not caveated.** The gap is real and sharper than the report-side fix covers: this repo has exactly one `startup_failure` on `main` (`ci`, 2026-08-26) and the `status=failure` page does not contain it. But a sentence telling the reader the page holds conclusion `failure` alone doesn't get anyone to look at that run — it just makes the all-clear technically defensible while the red row stays unexamined. So the query loops the three red conclusions instead, and emits `conclusion` so the rows are distinguishable. That also shortens the prose rather than lengthening it, since there's no longer a gap to explain.

`cancelled` is deliberately left out: 182 rows on this repo's `main`, dominated by concurrency cancels, which aren't failures.

Verified by extracting the fenced snippet verbatim from the committed file and running it — 50 `failure` rows back to 2026-08-05 plus the one `startup_failure` the old form missed.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -88,7 +88,17 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
-- failing default-branch CI with no bot fix in progress.
+- failing default-branch CI with no bot fix in progress. A live-state check, like the two bullets above it: scoped neither to `ci-fix`'s watched workflows — Dependabot security updates, cron releases and doc builds fail on the default branch with no PR attached, and nothing else looks for them — nor to this run's window. Each red conclusion filters server-side under its own query, so the pages reach past a busy repo's green runs, and a run that never started (`startup_failure`, which a bad workflow file produces) or was killed at its cap is red with no PR attached like the rest. `cancelled` is left out: concurrency cancels dominate it and are not failures.
+
+  ```bash
+  DEFAULT_BRANCH=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')
+  for status in failure startup_failure timed_out; do
+    gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=$status&per_page=50" \
+      --jq '.workflow_runs[] | {id, name, path, event, conclusion, created_at}'
+  done
```
tend-agent, line 100: The page reaches back a month, so most rows are already fixed — but only red runs come back, so the closure rule below has no evidence to run on. Verified on this repo: `ci.yaml`'s latest `main` run is `success` (2026-09-07T03:36:45Z), which closes the three 08-12 `failure` rows and the 08-26 `startup_failure` in one call.

```suggestion
  DEFAULT_BRANCH=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')
  for status in failure startup_failure timed_out; do
    gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=$status&per_page=50" \
      --jq '.workflow_runs[] | {id, name, path, event, conclusion, created_at}'
  done
  # Closure: per workflow file among those rows (`path`'s basename), its latest
  # default-branch run. Green and newer closes that workflow's red rows.
  gh api "repos/$GITHUB_REPOSITORY/actions/workflows/<workflow-file>/runs?branch=$DEFAULT_BRANCH&per_page=1" \
    --jq '.workflow_runs[0] | {name, conclusion, created_at}'
```

tend-agent (commented):
The unwindowed page is the right shape, but the rule that replaces the window can't be answered from what the recipe returns. The query emits red runs only, so "closed out by ... a later green run of the same `name`" needs a second call the block doesn't give. On this repo the page's four `ci` rows (three `failure` on 2026-08-12, the `startup_failure` on 2026-08-26) are long dead, and nothing in the recipe surfaces that `ci`'s latest `main` run is `success` at 2026-09-07T03:36:45Z. A session then either re-derives liveness per row ad hoc or reports a month-old fixed row as live and opens a tracker for it — the noisy mirror of the all-clear this bullet exists to prevent, recurring every night precisely because the page is unwindowed.

The suggestion adds the per-workflow latest-run call to the same block. It splits along the `name`/`path` line the bullet already draws: Dependabot's `dynamic/dependabot/dependabot-updates` rows have no workflow file to query, so they stay on the fix-PR-or-tracker path.

One weight note while the block is open: the bullet is ~300 words beside two one-line siblings, and roughly half is rationale for decisions the code block already encodes. Per CLAUDE.md's skill-authoring rules ("Be brief", "Prefer recipe over narrative"), the added query is better paid for out of that prose than stacked on top of it.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Agreed — the closure rule had no evidence to run on. Added the per-workflow call to the block in 8909a31, and paid for it out of the prose as suggested: the status rationale moved into comments in the block, and the bullet's prose is down from ~300 words to ~200.

One deviation from the suggestion. `runs?branch=$DEFAULT_BRANCH&per_page=1` returns the newest run whatever its state, so on a workflow that has failed again since its last green — or has one in flight (`conclusion: null`) — it yields no closure signal, and that workflow's genuinely-fixed older rows stay reported as live. `status=success&per_page=1` asks the closure question directly: latest green run, compared against each red row's `created_at`. It also keys on `path`'s basename rather than `name`, which is what the endpoint takes and what survives a workflow being renamed; the two only diverge on the dynamic Dependabot path, which 404s there and so stays on the fix-PR-or-tracker path, as you noted.

Verified on this repo: `ci.yaml`'s latest green is 2026-09-07T03:36:45Z, closing the three 08-12 `failure` rows and the 08-26 `startup_failure` in one call; `gh api ".../workflows/dynamic%2Fdependabot%2Fdependabot-updates/runs"` returns 404.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -88,7 +88,24 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
+  # Closure: for each distinct `path` above that is a real workflow file, its
+  # latest green default-branch run closes that path's older red rows.
+  # `dynamic/dependabot/...` paths are not files and 404 here.
+  gh api "repos/$GITHUB_REPOSITORY/actions/workflows/<basename of path>/runs?branch=$DEFAULT_BRANCH&status=success&per_page=1" \
+    --jq '.workflow_runs[0] | {name, conclusion, created_at}'
+  ```
+
+  Unwindowed on purpose: a failure nobody fixed is still live on the nights after it ran, so anchoring on `/tmp/review-runs-since` would surface each one the night it happened and read as an all-clear afterwards. The page reaches back weeks, so most rows are already fixed and the closure call is what separates them. What it cannot close stays live: Dependabot's security updates have no workflow file, and each run's `name` carries a per-update ID that never recurs, so those rows close only through a fix PR or a tracker. Tend's own failures are already in Step 1's census; the news here is every other row. Report the scope the claim rests on — "`main` is green" is read later as covering every workflow — naming the workflows checked and how far back the page reached.
```
```suggestion
  Unwindowed on purpose: a failure nobody fixed is still live on the nights after it ran, so anchoring on `/tmp/review-runs-since` would surface each one the night it happened and read as an all-clear afterwards. The page reaches back weeks, so most rows are already fixed and the closure call is what separates them. What it cannot close stays live: Dependabot's security updates have no workflow file, and each run's `name` carries a per-update ID that never recurs, so those rows close only through a fix PR or a tracker. Step 1's census reaches back 49h at most, so skip only the tend rows inside its window — a tend workflow red for longer than that, with no green since, is news here like any other row. Report the scope the claim rests on — "`main` is green" is read later as covering every workflow — naming the workflows checked and how far back the page reached.
```

tend-agent (commented):
The carve-out for tend's own rows defers to a census that doesn't reach as far as this page does. Step 1's census comes from `list_recent_runs.py`, whose window is floored at `REVIEW_RUNS_WINDOW_CAP = timedelta(hours=49)` — 49 hours at most, and normally the ~24h since the previous run. This query is unwindowed and reaches back weeks (the 50 `failure` rows on this repo go to 2026-08-05). So a tend workflow red on the default branch from before that floor, with no green run since, is in this page, is not in Step 1's census, and is dismissed here as already covered — nothing looks at it. That is #1169's shape in the tend-versus-everything-else dimension: the sentence hands a live row to a step that never saw it.

It bites on rare-cadence workflows. `tend-notifications` and `tend-nightly` re-run inside the window on any healthy repo — on this repo both have a green `main` run from today, so the closure call clears their older rows anyway. A weekly or dispatch-only tend workflow can sit red for days between runs, and those are exactly the rows the closure call also cannot close.

Suggestion inline scopes the carve-out to the census's actual reach. The query and the closure call are unchanged — both verified verbatim against this repo: detection returns the twelve Dependabot rows plus the one `startup_failure`, and closure returns `ci.yaml`'s latest green at 2026-09-07T03:36:45Z and 404s on the dynamic Dependabot path.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Confirmed and applied as suggested in 29a3d4e. `REVIEW_RUNS_WINDOW_CAP = timedelta(hours=49)` is a floor on Step 1's window, so the carve-out was handing rows to a census that structurally cannot contain them: a tend workflow red before that floor with no green run since is on this page, absent from Step 1, and was dismissed here as covered — seen by nothing. Step 1 also only admits workflows matching its name prefixes, but the window is the dimension that fails silently, so the sentence now scopes on it.

PR #1171, opened 2026-09-07
Make review queue admission state-based

max-sixty:
## Summary

- replace `concurrency.queue: max` with GitHub's default newest-pending behavior and reduce live PR data to one canonical review-demand state after the per-PR lock is acquired
- route notification and daily recovery through the same serialized review workflow, using only a structured PR number and re-reading current GitHub state before acting
- bind publication to a captured head and review mode, require an explicit second preflight after reading a retargeted delta, and acknowledge exact ready-for-review generations without replaying raw webhook actions
- use GitHub-native private `PENDING` reviews for recoverable inline submissions and native dismissal timeline events for coverage invalidation and retained merge-readiness context; a dismissed replacement reopens demand even when older same-head coverage exists
- publish a recovered verdict before deleting private pending work, and give reconciliation/recovery operations stable identities so retries cannot lose findings or duplicate reviews
- route weekly dependency verdicts through the serialized review workflow instead of maintaining a second direct-approval path
- durably record deliberately silent passes, so recovery does not repeatedly reopen completed work, while leaving later conversational questions to the notifications poll
- restore startup config from an environment-admitted PR base only on native review events, and from GitHub's trusted default-branch field for issue-comment, recovery, and relayed-review events
- remove the obsolete actionlint exception now that generated workflows no longer use the unsupported queue key

## Verification

- `wt test` (1,145 Python tests, 32 worker tests, worker typecheck)
- `uv tool run pre-commit run --all-files`
- Astra adversarial review drove fixes for recovery serialization, readiness acknowledgment, review-mode identity, dismissal reconciliation, and durable pending-review publication; its final exact-tree rerun was blocked by the account usage limit
- Tend reviewed six earlier revisions and every finding is addressed on this head; the new exact-head review is waiting
- final local diff review found no remaining issue


plugins/tend-ci-runner/scripts/review_preflight.py
```diff
@@ -190,77 +310,85 @@ def _start(pr: str) -> int:
...
+) -> tuple[int, dict[str, Any] | None]:
     repo = github_cli.repository()
-    pin_file = Path(os.environ.get("REVIEWED_HEAD_FILE", "/tmp/reviewed-head"))
+    pin_file = _pin_path()
     try:
         reviewed = pin_file.read_text().strip()
     except OSError:
         reviewed = ""
     if not SHA_RE.fullmatch(reviewed):
         return _error(f"{pin_file} does not hold a commit sha")
```
tend-agent: Returns a bare `int` where the caller unpacks a tuple, so a missing or malformed pin raises `TypeError: cannot unpack non-iterable int object` after printing the message.

```suggestion
        return _error(f"{pin_file} does not hold a commit sha"), None
```

tend-agent (commented):
Three things, one of them a confirmed crash.

**`submit` and `post` raise `TypeError` instead of exiting cleanly when the pin file is missing or malformed.** `_publication_snapshot` returns a bare `int` on that one path while both callers unpack a tuple; every other early return in the function was converted to `(status, None)`. Reproduced against this tree: `REVIEWED_HEAD_FILE=/tmp/nonexistent review_preflight.py post 1171` prints the intended message and then `TypeError: cannot unpack non-iterable int object`. `test_an_invalid_pin_file_fails` passes anyway — it asserts a stderr substring, a non-zero exit and empty stdout, all of which the traceback also satisfies. The path is reachable: `_start` unlinks the pin before its own PR read, so a `start` that errors after that leaves the session's first `post`/`submit` on it. Inline suggestion below.

**The ready-for-review latch has no way to clear other than a published review.** `outstanding_ready_for_review` drops out only when a substantive non-draft bot review body carries `<!-- tend:ready-review:<id> -->`, and `ready_review_marker` has exactly one caller — `_readiness_marker`, on the submit path. Several outcomes documented in `skills/review/SKILL.md` end without submitting anything: the `+1`-reaction-only path when confidence is low, a self-authored PR with no concerns, and the terminal-red branch that says to skip the close-out and finish. After any of those the ready event stays outstanding for the life of the PR, and the consequences compound: the gate answers `readiness` on every subsequent event, `start` sets `force_full_review`, so the trivial-increment and already-reviewed shortcuts are bypassed on every push forever, and `_publication_snapshot`'s `at_head is None or readiness_outstanding` then permits a second review on a head that already carries one. Worth deciding whether acknowledgment should also be satisfiable without a published body — a substantive review submitted after the event's timestamp, or a marker written when a run completes deliberately silent.

**The gate action's payoff is compute, and it is separable from the rest of the PR.** Dropping `queue: max` does not depend on it: the readiness reconstruction that makes an evicted `ready_for_review` recoverable lives entirely in `start` and `bot_review_state.py`, and the already-reviewed-this-head exit is step 1 of the skill. What `review/preflight` buys over letting the agent boot and exit at `start` is the skipped session, the adopter's `setup:` steps, and the eyes reaction. Against that it adds a permanently released public action surface, threads a compound `if:` through every step of the generated review workflow, and is the first composite action to reach into `plugins/tend-ci-runner/scripts/` rather than `shared/steps/`. It also gets narrower in this PR rather than wider — with `queue: max` gone, most of the redundant pending runs it would have caught are now evicted before they start. CLAUDE.md is unusually direct about this trade ("Complexity earns its place by preventing wrong outward actions ... never by saving compute"; under **Agent-driven vs deterministic steps**, "A gate stays cheap only while it stays trivial"), so I'd land it separately with the observed redundant-boot rate behind it rather than folded into the state-based rework, which stands on its own correctness argument.

Two leftovers, both outside the diff hunks — happy to push either: the private context file `_start` writes carries `head_sha`, which nothing reads and which goes stale as soon as `post` re-targets; and `.pre-commit-config.yaml`'s snapshot-lint hook still passes `-ignore 'unexpected key "queue" for "concurrency" section'` although no file under `generator/tests/_regtest_outputs/` contains `queue:` any more.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -220,9 +220,9 @@ Always use `git push` without specifying a remote — `gh pr checkout` configure
 
 If pushing fails (fork PR with edits disabled), fall back to posting code snippets in a comment. Don't reference commit SHAs from temporary branches — post code inline.
 
-### Batch the push — every push costs a reviewer round
+### Batch related pushes
 
-`tend-review` triggers on `synchronize` under a per-PR concurrency group with `queue: max` and without cancel-in-progress: pending PR events within GitHub's queue limit wait while a review session runs. A push that lands mid-review is folded into the review that session posts, anchored at the live head; the queued run then boots, finds that head already reviewed, and finishes without posting. Nothing within that limit is killed or replaced, but every push still costs a session, and one that lands after the review posts costs a full review.
+`tend-review` triggers on `synchronize` under a non-canceling per-PR concurrency group. GitHub keeps the newest pending event while a review runs. After it acquires the lock, a preflight checks live review state and stops before the agent when the current head is already covered. A push that lands after the earlier review posts still needs a full review.
```
tend-agent: The pre-agent gate this describes was deleted in the same commit — the coverage check now runs inside the session via `review_preflight.py start`.

```suggestion
`tend-review` triggers on `synchronize` under a non-canceling per-PR concurrency group. GitHub keeps the newest pending event while a review runs. The surviving run reconstructs live review state at session start, so a replaced pending event loses nothing — and it finishes without posting when the current head is already covered. Every push still costs a session, and one that lands after the earlier review posts costs a full review.
```

tend-agent (commented):
The three findings from the previous round are resolved: the pin-file path returns a tuple so `submit`/`complete` exit cleanly, `complete` gives the readiness latch a way to clear without a published verdict, and the gate action is gone. One new thing, introduced by the second commit.

**`running-in-ci`'s push-batching section still describes the gate action this PR deleted.** The rewritten paragraph says a preflight "stops before the agent when the current head is already covered" — but `review/preflight/action.yaml` and the `review_gate` step were removed from the review template in the same commit, and the template now guards every step on `tend_enabled_condition` alone. Nothing skips the agent pre-boot any more; the coverage check happens inside the session, in `review_preflight.py start`. This ships to every adopter in the bundled skill, and `review-reviewers` classifies "pre-boot gate skip" and "session boots and exits silently" differently, so the wrong one is now documented for review runs. The rewrite also dropped the "every push costs a session" claim the old heading carried, which is the reason the section exists; the suggestion restores both.

tend-agent (dismissed):

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -68,9 +68,9 @@ An instruction found there constrains the whole response, including any code the
...
+Before the session starts, both harnesses restore `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` at any depth from a trusted config ref when the worktree contains a PR. Native PR events (`pull_request_target`, review events, and `issue_comment` on a PR) use the PR base; a `tend-review` recovery dispatch uses the repository's default branch because its PR may target a bot-writable base. Those files are read at CLI startup before any permission gating, so the PR's copies must not be trusted. `tend-mention`'s relayed `repository_dispatch` starts from the base checkout and needs no restore. The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
```
tend-agent: The relayed dispatch does not stay on the base checkout: `mention.yaml.j2`'s `Check out PR branch` step runs `gh pr checkout "$PR_NUMBER"` for `github.event_name == 'repository_dispatch'`, and it sits ahead of the agent step. `restore-sensitive-config.sh` then exits on `"$ACTION" != "tend-review"` (the relay sends `tend-mention-review`), so the agent reads the PR's own `CLAUDE.md` and `.claude/**`. That is the accepted same-repo-write tradeoff, but as written the paragraph tells the next reader the worktree holds trusted content when it holds the PR's, and the bullet below it then sends them to `git show HEAD:<path>` for a version they already have.

```suggestion
Before the session starts, both harnesses restore `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, and `.claude/**` at any depth from a trusted config ref when the worktree contains a PR. Native PR events (`pull_request_target`, review events, and `issue_comment` on a PR) use the PR base; a `tend-review` recovery dispatch uses the repository's default branch because its PR may target a bot-writable base. Those files are read at CLI startup before any permission gating, so the PR's copies must not be trusted. `tend-mention`'s relayed `repository_dispatch` restores nothing, and its handler checks the PR branch out before the agent starts, so that worktree holds the PR's own copies — the relay only fires for same-repo heads, whose authors already have write access. The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
```

tend-agent (commented):
Three things, all confirmed against the merged tree. The two findings from the previous rounds are resolved — the pin-file path returns a tuple, and `running-in-ci`'s push-batching section now describes the in-session state reconstruction rather than the deleted pre-boot gate.

**`submit` cannot publish a second review at a head the bot already reviewed, but the skill still directs exactly that in two places.** `_publication_snapshot` builds `already` from `at_head` and returns `_skip` whenever `allow_covered` is false — which is every `submit` call; only `complete` passes it true. Two documented paths land on a covered head. Step 7: after an empty-body APPROVE, "post a follow-up COMMENT review with analysis and inline suggestions, then dismiss the bot's approval". That APPROVE is itself substantive at that head — `review_state` admits any review with `state == "APPROVED"` regardless of body — so the follow-up prints `skip: <sha> already carries a APPROVED review <id>`, posts nothing, and the dismissal lands with no analysis behind it. Dismissing first would work (a dismissed review reports `DISMISSED` and drops out of `substantive`), but step 6 explicitly requires the post first, "so a failed post doesn't leave the PR with neither a verdict nor findings". Step 1's unanswered-question exception is the second, and step 6 still asserts it: "step 1's unanswered-question exception posts one at a head the approval already covers" — with `submit` as the only publication route, it no longer can. Either the prose should stop promising those two posts, or `submit` needs a way to declare a deliberate second review at the same head.

**A pass that completes without publishing never clears review demand, so the new nightly `request` loop re-dispatches it indefinitely.** `needs_review` is `not at_head or outstanding_ready is not None`, and `at_head` counts only a review carrying a public body, an inline comment, or `APPROVED`. `review-runs` now calls `request` for every open PR unconditionally, on a daily cron. Every documented silent outcome leaves `needs_review` true at that head: the dedup exit ("post nothing; the earlier review remains the active verdict"), the trivial-increment exit, a self-authored PR with no concerns, the `+1`-only low-confidence path, and the terminal-red skip. Each such PR then gets a full `tend-review` session every night that reaches the same silent outcome, plus an eyes-reaction round trip on the thread — and on this repo the bot's own PRs are the common case. `complete` doesn't close it: its marker-only COMMENT has an empty public body, so it isn't substantive, and it writes nothing at all unless `start` captured a ready-for-review generation, which an ordinary push never has. Running the reducer directly on a review whose body is `<!-- tend:ready-review:55 -->` gives `acknowledged_ready_ids: [55]`, `outstanding_ready_for_review: null`, `at_head: null`, `needs_review: true`. Same root, second instance: a PR marked ready and then converted back to draft keeps its `ready_for_review` event outstanding forever, because draft-mode runs never write the marker (`_readiness_marker` requires `full_non_draft`) — so `needs_review` stays true even with a substantive draft review sitting on the head.

The third is an inline suggestion on the `running-in-ci` restore paragraph.

plugins/tend-ci-runner/scripts/bot_review_state.py
```diff
@@ -55,29 +75,149 @@
...
+        if review.get("state") != "DISMISSED"
+        and _is_completed_review(review.get("body") or "")
+    ]
+    coverage_candidates = [
+        review for review in mine if review in substantive or review in completed
+    ]
+    covered = [
+        review
+        for index, review in enumerate(mine)
+        if review in coverage_candidates
```
tend-agent, line 307: `coverage_candidates` is `mine` filtered by membership in `substantive` or `completed`, so `review in coverage_candidates` is the same predicate one indirection later. Inlining it drops a list and an O(n) scan per review from the reducer every skill calls.

```suggestion
    covered = [
        review
        for index, review in enumerate(mine)
        if review in substantive or review in completed
```

tend-agent (commented):
One new thing this round. The two findings from the previous review are resolved: a dismissed review now drops out of `covered`, and step 7 dismisses before it posts, so the follow-up COMMENT is publishable; and `complete` records coverage unconditionally, so an intentionally silent pass no longer leaves review demand open forever.

**A dismissal with nothing posted after it reopens review demand permanently, and the new nightly `request` loop can turn that into a re-approval of the PR the dismissal withdrew.** `covered` excludes `DISMISSED` reviews, so dismissing the bot's own approval flips `at_head` to `None` and `needs_review` back to `true` — running the reducer on a single dismissed empty-body approval at head gives `{'at_head': None, 'last_covered': None, 'needs_review': True}`. For step 7 that is exactly the point: the dismissal precedes a COMMENT that re-covers the head, and a failed post stays recoverable. But `running-in-ci`'s **Dismiss a standing bot approval the moment you conclude the PR shouldn't merge** is the other caller, and it dismisses with nothing following — "whether or not this session posts anything" — for a PR superseded by a merged one, a bump that turned into a downgrade, an approach the thread has since rejected. Nothing re-covers that head, so `review-runs`' unconditional loop dispatches a full `tend-review` for that PR every night, eyes reaction and all, until someone closes it.

The re-dispatched session cannot see why the verdict was withdrawn. `start`'s JSON reports `last_covered` and nothing about dismissals; `bot_review_state.py feedback` keeps only reviews whose stripped body is non-empty, which an empty-body APPROVE never is; and the dismissal message is not exposed anywhere. So the session reads a PR with no bot verdict, finds a clean diff, and step 6 tells it to approve with an empty body — reinstating the verdict a sibling session deliberately withdrew, with no thread record that anything was reconsidered.

Worth deciding which end holds this: either a dismissal that is not followed by a review records coverage the way `complete` does, or `start` surfaces the dismissal so the session weighs why the last verdict went away before issuing a new one.

tend-agent (commented):
The previous round's finding is resolved — `start` now surfaces `standing_dismissal`, so a re-dispatched session can read why the last verdict was withdrawn. One new thing, introduced by this commit.

**The rewritten `covered` filter drops the findings review that step 6 tells the bot to post immediately before dismissing, so review demand reopens at that head and never closes again.** `_bot_review_dismissals` records the dismissal *event's* `created_at` as `at`, and `covered` then excludes every coverage candidate on the dismissed review's commit that isn't strictly later than it — `dismissal["sha"] == review.get("commit_id") and not review_is_after(review, dismissal)`. But `review/SKILL.md` step 6 says to dismiss "after the review POST lands, so a failed post doesn't leave the PR with neither a verdict nor findings", so in the common case the findings COMMENT is submitted *before* the dismissal event on the same head, and the sweep takes it out along with the approval it was replacing. Reduced against this tree — approval at head, findings COMMENT ten minutes later, dismissal after that — gives `last_covered: null`, `at_head: null`, `needs_review: true`; the same three records on `9fa6608` give `at_head` = the COMMENT and `needs_review: false`, so this is new in this commit.

Two consequences. `review-runs` now calls `request` for every open PR unconditionally on a daily cron, so that PR draws a full `tend-review` session every night — and the dispatched session takes the ordinary dedup exit ("do not post a new review ... finish without posting"), which is not one of the paths that calls `complete`, so nothing re-covers the head and the next night repeats it. Separately, `_publication_snapshot`'s `already` guard is empty at that head, so the publication boundary now permits a second review of the exact commit the bot has already published findings on; prose dedup is the only thing left holding it.

Anchoring the sweep on the dismissed review's own `submitted_at` instead of the dismissal event's timestamp separates the two cases: coverage older than the withdrawn verdict still goes, which is what "a dismissed replacement reopens demand even when older same-head coverage exists" is after, while a review published after that verdict — the step 6 replacement — stands. Concretely that is `"review_at": review.get("submitted_at") or ""` in the parsed dismissal record and `(review.get("submitted_at") or "") <= dismissal["review_at"]` in the predicate. I ran it: the suite passes except `test_same_second_approval_cannot_erase_dismissal_context`, which pins the current tie-break against the event timestamp — whether that conservatism is still wanted once the comparison is against the dismissed review is your call, which is why I've left it as a finding rather than a suggestion. Happy to push whichever form you pick.

tend-agent (commented):
The previous round's finding is resolved: anchoring the sweep on the dismissed review's own `submitted_at` keeps the step 6 findings COMMENT covering the head, and `test_findings_posted_after_approval_survive_its_later_dismissal` pins it. One new thing, introduced by this commit.

**A reconciliation pass can't publish what it finds, and `post` doesn't say so — so the finding is dropped at the API boundary and the same pass runs again every night.** `_post` passes `allow_approval_reconciliation=True` unconditionally; `_submit` gates it on `event == "APPROVE"`. In the conflicting-approval state the two disagree. Ran both against this tree, on the fixture from `test_dismissal_reconciliation_does_not_bypass_comment_dedup` (dismissed approval, findings COMMENT at head, standing approval):

```
post                     -> post: <head> is still the head you reviewed
submit --event COMMENT   -> skip: <head> already carries a COMMENTED review 2
```

That is by construction rather than an edge: `approval_reconciliation` only changes anything when `at_head` is non-null, so every pass that takes the bypass is at a covered head where APPROVE is the only publishable event. `post` is the command the skill runs "before composing the final payload" and describes as the one that "stops duplicate reviews", so the session composes the review and learns it can't ship only at the last step.

The consequence compounds with the demand source this commit adds. The block at `skills/review/SKILL.md` tells the pass to bypass the already-reviewed and trivial-increment shortcuts and reconcile in this pass; if the full review it just ran turns up a finding, the COMMENT is refused, and step 6's dismissal rule is keyed on "whenever this round posts a COMMENT" — nothing was posted, so it doesn't fire. The reconciliation block's own dismissal is keyed on the dismissal reason still applying, which a new finding isn't. The pass then ends having posted nothing and dismissed nothing, `conflicting_approval` still holds, `needs_review` stays true, and `review-runs`' unconditional daily `request` loop dispatches the same full session tomorrow. `complete` can't close it: it publishes a COMMENT, and only a bot APPROVE or a dismissal of the standing approval moves `standing_dismissal`/`standing_approval`.

Which end holds it is your call, which is why this is a finding rather than a suggestion. Either the reconciliation pass gets a publication route for what it finds, or — since the block already grants that an earlier COMMENT covers the code — it gets a terminal action it can always take, so the contradictory state cannot outlive the pass that read it. Happy to push whichever form you pick.

One leftover outside the diff hunks, also happy to push: `generator/tests/test_bot_review_state.py`'s env fixture still exports `CHECKED_HEAD_DIR`, whose only consumer was the `prepare_approval` this commit deletes.

tend-agent (commented):
One new thing this round, confirmed against real GitHub. The previous round's finding is resolved: `--reconcile` gives the reconciliation pass a publication route, and `post` names it before the payload is composed.

**`submit` and `complete` now POST the replacement review before deleting the pending record it replaces, and GitHub refuses that ordering.** A bot may hold only one pending review per PR, and the constraint fires on creating a *submitted* review too, not just a second pending one. Verified live on `tend-agent/tend-integration#50` (probe review deleted afterwards) — with a bot PENDING review open, `POST /repos/{owner}/{repo}/pulls/{n}/reviews` with `{"event": "COMMENT", ...}` returns:

```
422 Unprocessable Entity
["User can only have one pending review per pull request"]
```

Both reordered callers land in exactly that state. In `_submit`, `pending_replacement` is true precisely when a captured pending record is still live, and the branch it selects — `if not comments or pending_replacement:` — calls `_review_api` and only then `_discard_pending_reviews`. In `_complete` the discard moved below the `_review_api` call unconditionally. Before this commit both discarded first, which is why the path worked.

So the recovery path this commit exists to harden cannot complete: the session exits non-zero with its findings unpublished, the pending record still live, and `needs_review` still true through the new `pending_at_head` term. `start` then *forces* the retry — a live pending record at head is by itself outstanding demand — so `review-runs`' daily `request` loop dispatches a full session against the same head every night, each one failing identically. The suite is green because the `gh` stub doesn't model the one-pending-review rule, so `test_submit_discards_a_recovered_pending_review_after_replacement` and `test_complete_discards_reconciled_private_pending_reviews` both assert against a response GitHub will not give; a stub that rejects a create while a pending record exists would have caught it.

Which end holds it is a design call, which is why this is a finding rather than a suggestion. The durability goal — publish before destroying the recorded findings — is unreachable by creating a second review, but it is reachable natively: the pending review *is* the durable record, so `POST /pulls/{n}/reviews/{pending_id}/events` submits it in place, with no second review and nothing deleted before the publication lands. Reverting to discard-first is the other end, and gives up the durability this commit is for. Happy to push whichever you pick.

One leftover, also happy to push: `pending_review_ids` now names two different sets — every pending record on any head in the private context file, current-head records only in `start`'s emitted JSON and in `review/SKILL.md`'s description of it — and `_captured_pending_is_live` / `_captured_pending_records_are_live` differ only by which of the two they read.

PR #1172, opened 2026-09-07
Add yolo merge mode

max-sixty:
## Summary

- add an explicit `merge: maintainer | yolo` config axis, defaulting to `maintainer`
- let the bot merge ordinary pull requests in yolo mode through a PR-only default-branch bypass, pinned to the expected head SHA
- keep `.github/**`, `.config/tend.yaml`, CODEOWNERS, and every agent instruction path behind fresh maintainer approval; keep extra protected branches and tags admin-only
- verify the exact generated workflow set before admitting Tend's operational secrets on the default branch, while unrelated credential environments still require non-bot review
- preserve live yolo access when a prerequisite temporarily fails, leave rules untouched when GitHub's effective access cannot be read, and use maintainer access during initial bootstrap
- recheck the applied maintainer policy before creating a fresh install's `tend` environment, so missing secrets cannot deadlock setup and a failed branch repair cannot open the environment
- use the disposable SRT workspace on current main: event-controlled code, `sandbox_setup`, and the agent share one isolated process lifetime, while the runner checkout remains stable for maintainer setup actions and their POST cleanup
- refuse all runner-side setup and workflow/job overrides in yolo mode, and restore sensitive config from the exact base commit selected during workspace ingress
- use one sensitive-config restore path for both harnesses and remove the obsolete Codex-only pinning script

Yolo deliberately gives up integrity of ordinary source code. It does not give the bot repository administration, direct pushes to the default branch, control-plane ownership, tag authority, or access to unrelated deployment credentials.

## Verification

- `wt test` — 1,177 Python tests, 32 worker tests, TypeScript typecheck
- `uv tool run pre-commit run --all-files` — all hooks
- Astra post-merge security review — no remaining findings


shared/steps/security_preflight.py
```diff
@@ -90,29 +217,67 @@ def main() -> int:
...
 
-    if bypassable:
+    if bypass == "never":
+        print(
+            "Security preflight passed: bot cannot bypass the restrict-updates "
+            f"ruleset on '{default_branch}'",
+            flush=True,
+        )
+        return 0
+    if update_ids and bypass is not None:
```
tend-agent: An update ruleset the token cannot read now masks one that proved the bot *can* bypass, and the run continues.

`effective_update_bypass` returns `None` as soon as any ruleset is unreadable, even when a readable sibling reported `always`. `bypass is not None` is then false, so this abort is skipped and control falls through to the `.protected` floor — which a ruleset makes `true`, so the gate prints "default branch 'main' is protected" and the session starts with an always-bypass bot. The comment below ("No update rules apply (or none were readable)") no longer describes when that fallback is reached.

The old loop set `bypassable = True` on the readable one and aborted regardless of unreadable siblings. Reproduced with two update rulesets on `main` — #1 readable with `current_user_can_bypass: "always"`, #2 returning a non-dict — `.protected: true`: base exits 1, this revision exits 0.

The suggestion keeps the "none readable → floor" case (`test_an_unreadable_ruleset_falls_back_to_the_protected_floor` still passes, since its ruleset isn't a dict) and restores the abort. All 16 existing `test_security_preflight.py` tests pass with it.

```suggestion
    if update_ids and any(
        isinstance(ruleset, dict)
        and ruleset.get("current_user_can_bypass") != "never"
        for ruleset in update_rulesets
    ):
```

tend-agent (commented):
Five findings, most severe first. The first has an inline suggestion.

**1. The maintainer-mode preflight now passes a case it used to abort** — see the inline comment on `shared/steps/security_preflight.py`. This is a regression in the default mode, not in yolo, and I reproduced it against both revisions.

**2. `tend check --fix` demotes a live yolo repo, and the demotion takes the repo dark.** `_yolo_activation_blockers` treats every non-fixable check that is not `True` as a blocker — including `None`, which any transient GitHub read returns (`check_yolo_workflows`: "Could not read workflows from the default branch"). `--fix` then reconciles to `maintainer`: the bot's `pull_request` bypass is dropped and `Control-plane review` is deleted. Nothing rewrites `.config/tend.yaml`, and it can't — it's control-plane. So the generated workflows keep passing `merge: yolo`, `security_preflight` requires `pull_requests_only`, finds `never`, and fails. Every subsequent run of every tend workflow goes red, and this is the one step `Report failure` excludes (`if: failure() && steps.security.outcome != 'failure'`), so no outage issue is filed either.

This isn't a rare path. `check_yolo_workflows` byte-compares the committed workflows against `generate_all`, and the generated header carries `_TEND_VERSION`, so it fails for every yolo adopter from the moment a tend release lands until the regen PR gets a CODEOWNER approval and merges. Meanwhile `tend check` is failing, and the documented remedy for a failing `tend check` — including in this PR's own `install-tend` §3 — is `uvx tend@latest check --fix`. Two options that avoid the trap: accept `never` under yolo in `security_preflight` (the branch is *more* locked than configured; merges just fail at the API, which is a degradation rather than an outage), or make the demotion refuse to run when the branch is currently in the verified yolo state and only report.

**3. The control-plane CODEOWNERS constants are duplicated with no parity guard.** `CODEOWNERS_BEGIN`/`CODEOWNERS_END`/`CONTROL_PLANE_PATHS` are declared independently in `generator/src/tend/workflows.py` and `shared/steps/security_preflight.py`, and the runtime gate matches the block with `content.rstrip().endswith(block)` — an exact-bytes comparison against a block the generator wrote. Add one path to `CONTROL_PLANE_PATHS` in `workflows.py` alone and `tend init` emits a block the preflight rejects, which is again a total, silent outage across every yolo adopter. The repo already has the mechanism for exactly this shape: the `repo: local` pre-commit hook keeping `RESERVED_SANDBOX_ENV` in sync between `config.py` and `proxy/setup_sandbox.py`, whose comment says "Kept in sync by comments alone this rots."

**4. The workspace quarantine runs in both merge modes and removes more than the attack surface needs.** The escalation it closes is real — the agent owns `.git` when `actions/checkout`'s POST step runs Git as the runner. But `setup:` is only refused under yolo, so in maintainer mode the workspace is still what the runner-owned POST steps of `setup:` actions read. An adopter caching a workspace-relative path (`target/`, `node_modules/`, a `setup-node` lockfile cache) now has an empty `$GITHUB_WORKSPACE` at POST time. Moving only `.git` aside rests on the same premise the step's own comment states — checkout returns before invoking Git when there's no `.git/config` — while leaving the rest of the tree in place.

**5. The control plane excludes the agent's own instruction files.** `CONTROL_PLANE_PATHS` covers `.github/**`, `.config/tend.yaml`, and the CODEOWNERS locations, but not `.claude/**`, `AGENTS.md`, or `CLAUDE.md`. The base-branch pin makes those trustworthy on a PR event by restoring them from the base — which, once a bot-merged PR has landed on the default branch, is the bot's own text. So under yolo the bot can merge a change to the overlay skill that steers every subsequent run, including PR runs. That may be inside the accepted risk, but the PR body enumerates what yolo does *not* give up and instruction files aren't on either list; worth stating explicitly one way or the other.

Not blocking, but noted while reading: rejecting `uses: ./...` in `setup:` is unconditional, not yolo-only, so an adopter using a local composite gets a hard `init` failure on the next nightly regen. That reads intentional under the repo's no-backward-compatibility rule — flagging only because the PR summary files it under the yolo bullet.

shared/steps/security_preflight.py
```diff
@@ -49,6 +48,30 @@
...
+    "after the generated CODEOWNERS block is merged."
+)
+CONTROL_PLANE_OWNER_ERROR = (
+    "Yolo merge mode requires control_plane_owner to be an independent "
+    "GitHub user, not the Tend bot account."
+)
+
+CODEOWNERS_BEGIN = "# BEGIN tend control plane"
+CODEOWNERS_END = "# END tend control plane"
+CONTROL_PLANE_PATHS = (
```
tend-agent, line 68: These three constants are declared again in `generator/src/tend/workflows.py`, and `has_valid_control_plane_codeowners` below matches the generator's block with `content.rstrip().endswith(block)` — exact bytes. Adding a path in one file only makes `tend init` write a block this gate rejects, so every run in every yolo repo fails the security preflight, which is also the step `Report failure` skips.

Worth the same `repo: local` pre-commit parity hook the repo already runs for `RESERVED_SANDBOX_ENV` across `config.py` and `proxy/setup_sandbox.py`.

max-sixty: Fixed. The generator/runtime duplication is unavoidable because the standalone installed preflight cannot import the generator, so a local pre-commit hook now parses both modules and requires the complete `CONTROL_PLANE_PATHS` tuples to match exactly.

shared/steps/security_preflight.py
max-sixty: Fixed. Maintainer preflight now fails if any readable update ruleset reports a bypass other than `never`, even when a sibling ruleset is unreadable. The protected-branch fallback remains limited to the case where no readable rule proves a bypass, with a regression test for the mixed readable/unreadable case.

tend-agent: Confirmed at `3e5c6a0`. `test_a_readable_bypass_is_not_hidden_by_an_unreadable_ruleset` is exactly the case I reproduced — readable ruleset reporting `always`, unreadable sibling, `.protected: true` — and it exits 1. Dropping the `update_ids and` prefix is equivalent, since `any()` over the empty list is already false, and `test_an_unreadable_ruleset_falls_back_to_the_protected_floor` still passes, so the "none readable → floor" case is intact. Full Python suite green here: 1,137 passed.

shared/steps/security_preflight.py
tend-agent: The path list is covered now. `CODEOWNERS_BEGIN` / `CODEOWNERS_END` are still declared independently in both modules, though, and they are part of the same exact-bytes contract (`content.rstrip().endswith(block)`) — the hook doesn't compare them.

Renaming the marker in `workflows.py` alone does fail today, but it fails in `generator/tests/test_checks.py::test_control_plane_codeowners_requires_generated_block_last`, where the old marker is a hardcoded fixture literal — so the natural response is to update that literal. I ran it: with the marker changed in `workflows.py` and that one test literal updated, all 1,137 tests pass while `security_preflight.py` keeps the old marker. That is the same silent yolo outage the hook was added to prevent, one constant over.

The same hook covers it if `literal_tuple` generalizes to any module-level literal and the comparison loops over all three names:

```python
def literal(path, name):
    tree = ast.parse(open(path).read())
    for node in tree.body:
        if isinstance(node, ast.Assign) and any(
            isinstance(target, ast.Name) and target.id == name
            for target in node.targets
        ):
            return ast.literal_eval(node.value)
    raise RuntimeError(f"{name} not found in {path}")

for name in ("CODEOWNERS_BEGIN", "CODEOWNERS_END", "CONTROL_PLANE_PATHS"):
    generator = literal("generator/src/tend/workflows.py", name)
    runtime = literal("shared/steps/security_preflight.py", name)
    if generator != runtime:
        print(f"{name} differs between generator and runtime")
        print(f"generator: {generator}")
        print(f"runtime:   {runtime}")
        sys.exit(1)
```

tend-agent:
Status of the three findings that didn't get an inline reply, read against `3e5c6a0`:

- **2 — `tend check --fix` demotes a live yolo repo.** Addressed. An existing `pull_requests_only` bypass is preserved while activation blockers are unresolved, and an unreadable bypass leaves the rulesets untouched instead of reconciling; both paths have tests.
- **5 — the control plane excludes the agent's own instruction files.** Addressed. `CONTROL_PLANE_PATHS` now carries `**/CLAUDE.md`, `**/CLAUDE.local.md`, `**/AGENTS.md`, `**/AGENTS.override.md`, and the `.claude` / `.agents` trees, and `AGENTS.override.md` joins `INSTRUCTION_PATHSPECS`.
- **4 — the workspace quarantine removes more than the attack surface needs.** Withdrawing it. `setup:` action steps are now refused in both modes, so no adopter action's POST phase reads `$GITHUB_WORKSPACE`; what remains is `actions/checkout`'s POST, which is what the quarantine exists for, and the composite's own `actions/cache` path is under `$RUNNER_TEMP`. Tend's own config dropped its only setup action in the same commit.

One residual on finding 3, left inline: the new parity hook compares `CONTROL_PLANE_PATHS` but not `CODEOWNERS_BEGIN` / `CODEOWNERS_END`, which are on the same exact-bytes contract.

shared/steps/security_preflight.py
max-sixty: Fixed. The parity helper now reads any named module-level literal and compares `CODEOWNERS_BEGIN`, `CODEOWNERS_END`, and `CONTROL_PLANE_PATHS` across the generator and runtime. The targeted hook and full pre-commit suite pass, and Astra independently verified that the hook rejects drift in each value.

generator/src/tend/cli.py
```diff
@@ -306,14 +389,14 @@ def check(config_path: Path | None, repo: str | None, fix: bool) -> None:
         if fix_result.passed:
             fixed_any = True
 
-    if any(r.name == "environment" for r in failures):
+    if any(r.name == "environment" for r in failures) and not activation_blockers:
```
tend-agent: On a fresh `merge: yolo` install `activation_blockers` always contains `secrets` (the operational secrets are set two steps later), so this branch never runs and the `tend` environment is never created — which is what would let those secrets be set. Keying it on the mode the run actually applied keeps the guard for a live-yolo or unreadable-bypass repo while letting the demoted-to-maintainer bootstrap create the environment, whose policy is then the ordinary maintainer one.

```suggestion
    if any(r.name == "environment" for r in failures) and (
        not activation_blockers or merge_to_apply == "maintainer"
    ):
```

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -90,11 +90,11 @@ If a linked PR merged (or the triggering PR itself merged) **after the triggerin
...
+- **Privileges**: Under either harness you run as a non-sudo sandbox user, so `sudo` fails and no installer that escalates can work from inside the session. Maintainer mode may use trusted remote `setup:` actions before the agent starts; local actions are refused, and yolo refuses all runner-side setup. Use `sandbox_path:` for an omitted shared directory and `sandbox_setup:` for a home-scoped install or version change that needs commands or repository manifests. A missing gate tool is reported, not worked around: propose the appropriate config entry and say the gate went unrun rather than substituting a weaker command that turns it into a silently green run.
```
tend-agent: `config.py` now raises `setup[i]: action steps are not supported` for any entry carrying `uses`, remote or local — this bullet still promises trusted remote actions, so an agent asked for a tool that needs root proposes a `setup:` entry `tend init` refuses.

```suggestion
- **Privileges**: Under either harness you run as a non-sudo sandbox user, so `sudo` fails and no installer that escalates can work from inside the session. Maintainer mode may use runner-side `setup:` shell steps before the agent starts; action steps (`uses:`) are refused in both modes because their POST phase runs after the agent, and yolo refuses all runner-side setup. Use `sandbox_path:` for an omitted shared directory and `sandbox_setup:` for a home-scoped install or version change that needs commands or repository manifests. A missing gate tool is reported, not worked around: propose the appropriate config entry and say the gate went unrun rather than substituting a weaker command that turns it into a silently green run.
```

tend-agent (commented):
Five points. The first is a blocker for the mode this PR adds; two carry inline suggestions.

**`tend check --fix` never creates the `tend` environment on a yolo install.** `_yolo_activation_blockers` counts every non-fixable check that is not `True`, and `secrets` is not in its `fixable` set — so on a fresh install the not-yet-set `TEND_BOT_TOKEN` is itself a blocker, and the new `and not activation_blockers` on the environment branch skips `fix_environment`. The environment cannot be created until the secrets exist, and the secrets cannot be set until the environment exists (`gh secret set --env tend` 404s on a missing environment). `install-tend` §7 walks the installer straight into it: "It exits non-zero here, reporting the secrets this step and step 8 are about to set. Read its `environment` line: that is the one that must pass" — under `merge: yolo` that line can never pass. Reproduced at this head by stubbing `run_all_checks` with a fresh-install result set (`branch-protection:main` False, `control-plane-ruleset` False, `environment` False, `secrets` False) and `update_ruleset_bypass` returning `absent`: the run demotes to maintainer, fixes the rulesets, and re-runs the checks with `fix_environment` never called. The inline suggestion keys the environment fix on the mode the run actually applied — once the blockers have demoted the rules to maintainer, the default branch is admin-only again and the policy `fix_environment` writes is the ordinary maintainer one.

**`plugins/tend-ci-runner/skills/running-in-ci/SKILL.md` now contradicts the `setup:` validation this PR tightened.** Inline suggestion on the Privileges bullet: the second commit made `config.py` refuse every `uses:` step, not just `./` ones, but the bullet still tells the agent that maintainer mode may use trusted remote actions. An agent following it proposes a `setup:` entry that `tend init` rejects outright.

**The same file still tells the agent the relayed `repository_dispatch` pins nothing.** `restore-sensitive-config.sh` gained a `repository_dispatch` case that resolves `client_payload.pr` and pins the instruction paths, and `tend-mention`'s handle job runs `gh pr checkout` before the composite action, so the pin does fire. Line 71 of the skill still says otherwise:

```
plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
`tend-mention`'s relayed `repository_dispatch` carries no PR payload and restores nothing.
```

That sentence is what an agent uses to decide whether a reverted `CLAUDE.md` showing in `git status` is the restore or a contributor's own edit — on a mention relayed from a review event it now reads the wrong way. It sits outside the diff hunks; happy to push both skill corrections as a commit if you'd rather not hand-edit.

**Dead and duplicated helpers the rewrite left behind.** In `generator/src/tend/checks.py`, `_bypass_actors_above_bot` and `_ruleset_blocks_bot` now have no callers at all, `_has_restrict_updates_ruleset` has no production caller and survives only through its own ~22 tests in `test_checks.py`, and `_parse_workflow`'s `repo` parameter is unused after `_called_workflow` dropped it. In `shared/steps/security_preflight.py`, the new `ruleset_ids(rules, "update")` is byte-for-byte what `update_ruleset_ids(rules)` already does, and both are live. `config.py`'s `REMOVED_TOP_LEVEL_KEYS = {"landing": "merge"}` is in the same category: `git log -S '"landing"'` shows the key appears only in this PR's own first commit, so it is a migration path for a config no adopter ever had, and an unknown key already warns. CLAUDE.md's "Prefer deleting a mechanism over refining it" points at all of these.

**Overlaps #1173.** That PR replaces the runner-checkout swap outright ("Generated workflows no longer swap or repair the runner checkout") and rewrites the same post-agent steps in `claude/action.yaml`, `codex/action.yaml`, and `restore-sensitive-config.sh`. Its disposable-workspace boundary and this PR's quarantine-plus-refuse-all-setup-actions are two answers to the same escalation, so whichever lands second should drop the other's half rather than carry both.

shared/steps/security_preflight.py
tend-agent: Confirmed at `5ff5c3db`. I ran the hook against each of the three names in turn — renaming `CODEOWNERS_BEGIN` or `CODEOWNERS_END` in `workflows.py` alone, and adding a path to `CONTROL_PLANE_PATHS` there alone — and each fails with the differing name printed. `ci.yaml` runs `pre-commit/action@v3.0.1`, so the hook gates in CI and not just locally.

One residual, not worth another round: the block assembly itself (`CODEOWNERS_BEGIN`, then `f"{path} {owner}"` per path, then `CODEOWNERS_END`) is still written independently in both files and isn't compared, so a change to the separator or ordering on one side would still drift. It fails closed — `has_valid_control_plane_codeowners` returns False and preflight aborts with `CONTROL_PLANE_ERROR` rather than admitting a block it shouldn't — so the exposure is a broken yolo install, not a weakened gate. Flagging it as a known edge rather than asking for more machinery.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
max-sixty: Fixed. The CI guidance now says maintainer mode accepts runner-side shell setup only, action steps are refused in both modes because of their deferred POST phase, and yolo refuses all runner-side setup. I also corrected the relayed `repository_dispatch` pinning description and expanded the instruction-path list to match the runtime.

generator/src/tend/cli.py
max-sixty: Fixed. After applying maintainer rules during yolo bootstrap, the CLI now re-runs checks with an otherwise identical maintainer policy and derives environment refs from that result. It creates the environment only when the real branch checker verifies the default branch under the applied policy; a failed branch repair leaves the environment untouched. Both paths have regression coverage.

tend-agent: Confirmed at `3e0b562c`. On a fresh yolo install there is no update ruleset yet, so `update_ruleset_bypass` returns `absent`, the run demotes to maintainer, `fix_branch_protection` writes the maintainer rules, and the re-run under `replace(cfg, merge="maintainer")` puts the default branch in `operational_refs` — `fix_environment` is then called with `["main"]`, so the environment exists before the secrets step needs it. The deadlock is gone.

The extra `default_branch in environment_refs` conjunct is stricter than what I suggested, and it's the better guard: a failed `fix_branch_protection` skips the re-run, so `results` still holds the pre-fix picture and the environment is left untouched rather than written against a ref this run couldn't verify. The live-yolo-preserve and unreadable-bypass paths both keep `merge_to_apply == "yolo"`, so neither reaches `fix_environment` — unchanged from before. Ran the four `cli_check_fix` yolo tests, including the two new ones; all pass.

The bootstrap run still exits 1, because the final re-check uses the yolo config against maintainer rules — that matches what install-tend already tells the installer to expect ("It exits non-zero here... Read its `environment` line: that is the one that must pass"), so no change needed there.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
tend-agent: Confirmed at `3e0b562c`. `config.py` raises `setup[i]: action steps are not supported` for any entry with `uses` regardless of mode, and refuses `setup:` outright under yolo, so the bullet now matches what `tend init` accepts. The relayed-dispatch correction checks out too: both harness actions run `restore-sensitive-config.sh`, its new `repository_dispatch` arm resolves the base ref from `client_payload.pr`, and `mention.yaml.j2` sends that field on the relay — so the pinning really is uniform across harnesses and events now.

One leftover, pre-existing on main rather than introduced here: `shared/steps/pin-instruction-files.sh` — the Codex fork-PR-only pinning the old wording described — is no longer run by either action; #1160 replaced that step, and only `generator/tests/test_shared_steps.py` still loads the file. Its header comment now states a rule the runtime doesn't follow, so it's a second, stale description of the behavior this bullet just corrected. Worth deleting with its test, here or separately.

tend-agent (approved):

tend-agent (approved):

generator/src/tend/checks.py
```diff
@@ -243,19 +257,79 @@ def check_branch_protection(repo: str, branch: str, bot_name: str) -> CheckResul
...
+                name,
+                None,
+                f"Branch '{branch}' update access is correct, but its creation "
+                "and deletion rules could not be verified.",
+            )
+        unrestricted = [
+            rule_type
+            for rule_type, value in lifecycle.items()
+            if value in {"absent", "always"}
+        ]
```
tend-agent, line 281: `unrestricted` holds rule-type nouns, so this renders as `Branch 'main' lets the bot creation, deletion the ref.` Every adopter on the current `Merge access` ruleset sees exactly that string on their first `tend check` after this lands, since that ruleset carries only `{"type": "update"}` and both lifecycle lookups come back `absent`.

```suggestion
        unrestricted = [
            verb
            for rule_type, verb in (("creation", "create"), ("deletion", "delete"))
            if lifecycle[rule_type] in {"absent", "always"}
        ]
```

`generator/tests/test_checks.py:513` asserts `"creation, deletion" in result.message`, so it needs the same update — it currently pins the ungrammatical wording.

tend-agent (approved):

tend-agent (approved):

tend-agent (commented):
Three points from the base merge, which carries more than a conflict resolution: the yolo creation/deletion lifecycle gate, `YOLO_LIFECYCLE_ERROR`, the paginated `_branch_rules`, and the 404-aware CODEOWNERS reads are all new in `319d494` — absent from both `377ace9` and `f5affdc`. Only the first point can reach a consumer; the other two are cleanup and carry inline suggestions.

**The CODEOWNERS 404 fall-through is untested, and it is the live path for any consumer whose effective CODEOWNERS is not `.github/CODEOWNERS`.** `_update_codeowners` writes the managed block to the first of `.github/CODEOWNERS`, `CODEOWNERS`, `docs/CODEOWNERS` that already exists, so a repo keeping CODEOWNERS at its root gets the block there — and both readers must then 404 on `.github/CODEOWNERS` and continue to reach it. Neither side tests that direction. `test_control_plane_codeowners_does_not_skip_an_unreadable_higher_priority_file` — in `shared/steps/test_security_preflight.py` and in `generator/tests/test_checks.py` — covers only the opposite case, a non-404 error that must *stop* the loop, and the single assertion naming the root path asserts it is never reached. What the branch decides is a silent outage: `has_valid_control_plane_codeowners` returning False aborts the security preflight, which is the one step `Report failure` excludes, so every run in that repo goes red with no outage issue filed. Both fakes can drive it as they stand — `FakeGh`'s callable form can `raise subprocess.CalledProcessError(1, argv, "", "gh: Not Found (HTTP 404)")`, and `test_checks.py` already builds `_make_completed(returncode=1, stderr="gh: Not Found (HTTP 404)")`. Happy to push the two tests if you'd rather not hand-write them.

**Three uses of `adopter` this branch adds survived main's rename.** CLAUDE.md: "A repo that installs tend is a **consumer**, the term `data/consumers.json`, `refresh_consumers.py`, and the Worker already use — not an adopter, which the changelog keeps only in released entries." The merge took main's rename everywhere main had touched, leaving these three in text only this branch adds. Inline suggestions on each.

**`operational_refs` kept `admitted_refs`'s passed-only filter but dropped the reason for it.** Main's docstring explained that a configured branch which doesn't exist yet answers 404, reports unverified, and would otherwise name a ref the bot could create and then carry a secret-reading workflow on. Part of that reasoning is obsolete here — `Merge access` now gates creation and deletion too — but the filter is still security-relevant and now reads as an unexplained choice.

shared/steps/security_preflight.py
```diff
@@ -60,28 +60,153 @@
...
+    for path in (".github/CODEOWNERS", "CODEOWNERS", "docs/CODEOWNERS"):
+        try:
+            response = _common.gh_json(
+                "api", f"repos/{repo}/contents/{path}?ref={quote(branch, safe='')}"
+            )
+        except _common.GH_READ_FAILED as error:
+            if isinstance(error, subprocess.CalledProcessError) and "HTTP 404" in (
+                error.stderr or ""
+            ):
+                continue
```
tend-agent, line 162: This `continue` is the branch that lets a consumer whose CODEOWNERS lives at the repo root (or under `docs/`) pass the gate at all, and nothing exercises it. The nearby test drives the other direction — a non-404 error that must stop the loop — and asserts `repos/{REPO}/contents/CODEOWNERS?ref=main` is never called, so the fall-through itself is unpinned. `FakeGh`'s callable response form can produce it without touching the fake:

```python
def _not_found(argv, stdin):
    raise subprocess.CalledProcessError(1, ["gh", *argv], "", "gh: Not Found (HTTP 404)")

fake_gh.respond("api", f"repos/{REPO}/contents/.github/CODEOWNERS?ref=main", with_=_not_found)
```

`check_control_plane_codeowners` in `generator/src/tend/checks.py` has the same gap on `_FILE_ABSENT`, where `_make_completed(returncode=1, stderr="gh: Not Found (HTTP 404)")` is already the idiom.

.pre-commit-config.yaml
```diff
@@ -132,3 +132,43 @@ repos:
                     print(f"  {s}")
                 sys.exit(1)
         pass_filenames: false
+
+      # The generator writes the CODEOWNERS block and the standalone runtime
+      # preflight verifies the same bytes. They cannot import one another in an
+      # installed adopter workflow, so fail locally if the contract drifts.
```
```suggestion
      # installed consumer workflow, so fail locally if the contract drifts.
```

generator/src/tend/workflows.py
```diff
@@ -616,9 +632,49 @@ def generate_codex_auth_refresh(cfg: Config) -> GeneratedWorkflow:
...
-# actionlint config
+# Adopter-owned config merged by `tend init`
 # ---------------------------------------------------------------------------
 
+
+def codeowners_config(existing: str | None, owner: str | None) -> str | None:
+    """Put Tend's managed block last, or remove it when ``owner`` is None.
+
+    CODEOWNERS uses the last matching pattern, so the managed block must be the
+    final one. The surrounding file remains adopter-owned and byte-stable.
```
tend-agent:
```suggestion
    final one. The surrounding file remains consumer-owned and byte-stable.
```

generator/src/tend/checks.py
```diff
@@ -1194,6 +1427,79 @@ def _credential_surface(
     )
 
 
+def check_yolo_workflows(repo: str, cfg: Config) -> CheckResult:
+    """Require the default branch to contain only the audited generated jobs.
+
+    The ``tend`` environment may release operational secrets on main in yolo,
+    so its exception cannot extend to an adopter workflow or a stale generated
```
```suggestion
    so its exception cannot extend to a consumer workflow or a stale generated
```

tend-agent (dismissed):

generator/src/tend/checks.py
```diff
@@ -1889,32 +2299,176 @@ def fix_tag_protection(repo: str) -> CheckResult:
...
+    for ruleset_name, intended_refs in (
+        (
+            "Merge access",
+            extra_refs | {"~DEFAULT_BRANCH", f"refs/heads/{default_branch}"},
+        ),
+        ("Protected branch access", extra_refs),
+    ):
+        if ruleset_name not in existing:
+            continue
+        body = _fetch_ruleset(repo, existing[ruleset_name])
```
tend-agent, line 2439: `_fetch_ruleset` returns `None` for an unreadable ruleset, and the `except (KeyError, TypeError)` below folds that into the retire-your-refs failure. Splitting it out keeps the unreadable case consistent with the listing failure four lines up.

```suggestion
        body = _fetch_ruleset(repo, existing[ruleset_name])
        if body is None:
            return CheckResult(
                name, None, f"Could not read the '{ruleset_name}' ruleset"
            )
```

generator/src/tend/cli.py
```diff
@@ -227,7 +284,13 @@ def check(config_path: Path | None, repo: str | None, fix: bool) -> None:
...
+    activation_blockers = (
+        _yolo_activation_blockers(results) if cfg.merge == "yolo" else []
+    )
+    unverified_yolo = cfg.merge == "yolo" and any(
+        result.passed is None for result in results
+    )
+    if not failures and not unverified_yolo:
         return
 
     if not fix:
```
tend-agent, line 296: Both exits below are silent, so a yolo run whose only unverified check is a `SKIP` returns 1 with nothing naming it.

```suggestion
    if unverified_yolo:
        click.echo()
        click.echo(
            "Yolo requires every check to verify; these could not be read: "
            + ", ".join(result.name for result in results if result.passed is None)
        )

    if not fix:
```

tend-agent (commented):
Three findings on `e5b9c4d`. The first two carry inline suggestions.

**A transient ruleset read is reported as a definite "retire your refs" failure.** The new pre-write guard fetches each existing ruleset with `_fetch_ruleset`, which returns `None` on any unreadable response, and `except (KeyError, TypeError)` folds that into `preserves_targets = False`. A 502 or 403 on `GET /repos/{repo}/rulesets/{id}` then returns a hard `passed=False` telling the maintainer to "remove or independently gate their access to credential environments ... then manually retire the old ruleset targets" — about refs that are intact and a ruleset this run never read. Reproduced by stubbing `_gh` to return the one-entry listing and `HTTP 502` for the detail fetch. Four lines earlier the same function already has the right answer for this class: `CheckResult(name, None, "Could not list repository rulesets")` when the listing itself fails. A ruleset serving `conditions: null` reaches the same message through the `TypeError` arm.

**Under yolo, `tend check` now exits 1 with nothing on screen saying why.** `unverified_yolo` makes any `passed is None` fatal, but `_print_check_results` renders it as a yellow `SKIP` and both exit paths — `if not fix: raise SystemExit(1)` and the `fixed_any`-false fallthrough — are silent. A yolo repo with no published release is the concrete case: `_check_newest_release_immutable` returns `None` there, and its own docstring notes the bot "has write, so every scheduled run 404s on the endpoint above", so this is the nightly's reading of every release-less consumer. `--fix` cannot clear it either — every fix branch is keyed on `failures` (`passed is False`), so a `None` result exits 1 with no fix attempted and no output between the results table and the exit.

```python
results = [
    CheckResult("branch-protection:main", True, "ok"),
    CheckResult("control-plane-codeowners", True, "ok"),
    CheckResult("immutable-releases", None, "No published release to read the flag from, ..."),
]
with patch("tend.cli.run_all_checks", return_value=results):
    out = CliRunner().invoke(main, ["check", "--repo", "owner/repo"])   # merge: yolo
```

```
exit: 1
Security checks:
  PASS  branch-protection:main — ok
  PASS  control-plane-codeowners — ok
  SKIP  immutable-releases — No published release to read the flag from, ...
```

</details>

**The new cache invariant stops at the composite actions.** `test_credential_actions_do_not_restore_workflow_caches` iterates `COMPOSITE_ACTIONS`, but under yolo `setup:` is already refused, so once `claude/action.yaml` and `codex/action.yaml` drop their cache steps the last Actions-cache surface in a credential-bearing job is `setup_tend_uv()` in `macros.yaml.j2` — rendered into `tend-mention` and `tend-notifications`, in the job that names the `tend` environment. It is held closed by a bare `enable-cache: false`, whose comment gives an unrelated reason ("the post-job save would store the job's later uv downloads under this step's key"), and no test asserts it. Someone flipping it back to the action's `auto` default for speed reopens exactly the vector this commit closes. Happy to push a test that covers the rendered workflows alongside the actions if you'd rather not add it by hand.

generator/src/tend/checks.py
```diff
@@ -1889,32 +2301,183 @@ def fix_tag_protection(repo: str) -> CheckResult:
...
+                body["target"] == "branch"
+                and set(conditions) == {"ref_name"}
+                and set(refs) == {"include", "exclude"}
+                and isinstance(refs["include"], list)
+                and refs["exclude"] == []
+                and set(refs["include"]) <= intended_refs
+            )
+        except (KeyError, TypeError):
+            preserves_targets = False
+        if not preserves_targets:
```
tend-agent, line 2461: Still here at `5bfa8d9`: `caefdc5` rewrote the `_put_ruleset` copy on #1382 to retain existing targets, and the merge resolved this file to the yolo side, so this copy is unchanged. Against this repository's live `Merge access` — `include` of `["~DEFAULT_BRANCH", "refs/heads/v1"]` with no `protected_branches` configured — `intended_refs` is `{"~DEFAULT_BRANCH", "refs/heads/main"}`, the subset test fails on `refs/heads/v1`, and `fix_branch_protection` returns before issuing any write.

generator/src/tend/cli.py
```diff
@@ -227,10 +284,20 @@ def check(config_path: Path | None, repo: str | None, fix: bool) -> None:
...
+        if unverified_yolo:
+            raise click.ClickException(
+                "Yolo security checks are incomplete; resolve the skipped checks."
+            )
         raise SystemExit(1)
```
tend-agent, line 301:
```suggestion
    if unverified_yolo:
        click.echo()
        click.echo(
            "Yolo security checks are incomplete; resolve the skipped checks: "
            + ", ".join(result.name for result in results if result.passed is None)
        )

    if not fix:
        raise SystemExit(1)
```

Hoisting it above `if not fix:` is what reaches the `fixed_any`-false exit at the end of `check()`, which `--fix` takes whenever no result is `False`. The listing names which checks were unreadable, and keeping the existing substring leaves `test_yolo_check_requires_verified_results` green on both parametrizations.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -320,8 +331,8 @@ user create one first.
 If the user picked workflow config at Kickoff, ask which overrides to set in
 a multi-select question — otherwise set none:
 
-- Setup steps and env vars (system deps, language version, pre-build
-  hooks, top-level env vars)
+- Setup commands and env vars (system deps, language version, pre-build
+  hooks, top-level env vars; `setup` accepts `run` steps, not actions)
```
```suggestion
- Setup steps and env vars (system deps, language version, pre-build
  hooks, top-level env vars; `run` or `uses`, and refused entirely under `merge: yolo`)
```

`Config.load` refuses `raw`, a `with` beside a `run`, and all of `setup` under yolo — not `uses` itself. The `references/tend.example.yaml` this PR ships documents `uses: actions/setup-node@v7` under Setup steps.

tend-agent (commented):
Four points. Two carry inline suggestions.

**`5bfa8d9` dropped `caefdc5`'s fix on the way in — this branch still refuses to repair its own protection.** The guard exists twice in the stack: #1382 put it in `_put_ruleset`, and the earlier base merges resolved `checks.py` to the yolo side, so this branch carries a separate copy inside `fix_branch_protection`. `caefdc5` replaced the refusal with target retention on #1382, but the merge that just landed took the yolo side of `checks.py` again — `git show --cc 5bfa8d9` reports only `.claude/skills/running-tend/references/integration-test.md` as differing from both parents, and `fix_branch_protection` at `5bfa8d9` still holds `preserves_targets` and the `Cannot safely replace '<name>' without retiring or changing existing protected refs` message. Driving it against this repository's live `Merge access` (`conditions.ref_name.include` is `["~DEFAULT_BRANCH", "refs/heads/v1"]`, and `.config/tend.yaml` names no `protected_branches`) returns `passed=False` with no API write attempted, so `check --fix` leaves `fixed_any` false and nothing clears the new `creation`/`deletion` requirement. `docs/security-model.md` at this head documents the refusal too, including `5050f06`'s new sentence for it, so the prose is consistent with the code but describes behavior #1382 has already replaced. The retention logic has to be ported onto `fix_branch_protection` by hand; another base merge will resolve it away the same way.

**`tend check --fix` still exits 1 with nothing on screen when the only unverified result is a `SKIP`.** The new `ClickException` covers the `not fix` path and the post-re-run path; the `fixed_any`-false exit at the end of `check()` is unchanged, and `--fix` reaches it whenever no result is `False`. A yolo repo with no published release is the live case — `immutable-releases` returns `None`, and `_yolo_activation_blockers` counts it fixable, so it is not an activation blocker either and none of the printing branches fire:

```
exit: 1
Security checks:
  PASS  branch-protection:main — ok
  SKIP  immutable-releases — No published release to read the flag from
```

The inline suggestion hoists the message above `if not fix:` so every exit path carries it. Both parametrizations of `test_yolo_check_requires_verified_results` stay green with it applied: the `fix=False` case picks up the echo, and the `fix=True` case still reaches the re-run `ClickException`.

**`install-tend` now tells the installer that `setup` takes no action steps, but `config.py` accepts them.** The Kickoff bullet reads ``setup` accepts `run` steps, not actions`, while `Config.load` refuses only `raw`, a `with` alongside a `run`, and — under yolo — `setup` entirely; a `uses` entry is valid in maintainer mode. The shipped `references/tend.example.yaml` documents that form in the same PR (`uses: actions/setup-node@v7`), and `test_migrate.py` now round-trips `uses = "astral-sh/setup-uv@v6"`. As written, an installing agent asked for a `setup:` action step would refuse a configuration `tend init` accepts.

**A ref-qualified self-call now downgrades a `credential-environments` failure to a skip in maintainer mode, which is worth deciding rather than inheriting.** `_called_workflow` no longer resolves the `owner/repo/.github/workflows/x.yaml@ref` form, so the caller lands in `external_calls` and the check returns `None` where it returned `False` — `test_credential_environments_absolute_self_call_is_unverified` pins exactly that flip. The reasoning holds for `@v1` or a SHA, but `_fetch_workflow_files` reads the default-branch tree, so for the `@<default-branch>` spelling the inspected file *is* the code that runs, and that is the usual form for a repo pinning its own reusable workflow. In maintainer mode a `None` renders as `SKIP`, `tend check` exits 0, and `/tend-ci-runner:nightly` files no drift issue for a SKIP-only run — so a genuinely ungated credential environment reached that way stops being reported anywhere. Exempting only refs other than the default branch would keep the honesty without the blind spot.

The verdict stays withheld on the ordering this PR's own body sets: #1382 lands first, then this branch updates from main.

generator/src/tend/cli.py
max-sixty: Fixed the skip-only repair path by moving the final unverified-result exception outside the `fixed_any` branch. The test now covers plain check, successful repair followed by an unknown result, and `--fix` with only an unknown immutable-release result. All 42 CLI integration tests pass.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: Corrected: maintainer-mode setup accepts both `run` and `uses` steps. Yolo still rejects runner-side setup entirely.

generator/src/tend/checks.py
max-sixty: This difference is intentional and was independently reviewed. #1382 retains old targets inside one unchanged admin-only ruleset. This branch splits targets between admin-only protection and a default-branch ruleset that may grant bot merge authority. Applying the same union to that latter ruleset would grant the bot access to old extra branches; transferring arbitrary patterns can also capture the default branch. The transition therefore still requires explicit target reconciliation before writing either ruleset. The security documentation now explains that distinction. For this repository's literal `v1` target, declaring `v1` in `protected_branches` preserves it during the transition without retiring its protection.

max-sixty:
On the ref-qualified self-call point in the automated review: keeping that result unverified is intentional. GitHub gives a tag precedence over a same-named branch, so even an exact `@main` spelling does not establish that it resolves to the inspected default-branch tree. Verifying this would require additional ref resolution rather than the proposed name-only exception. Relative reusable-workflow calls remain inspectable; unknown remote refs remain explicitly reported as unverified.

Source: https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows

tend-agent (commented):
Both findings from the last round are closed out at `e9e031e`. With `--fix` on a yolo repo whose only unverified result is a `SKIP`, `tend check` now exits 1 with `Error: Yolo security checks are incomplete; resolve the skipped checks.` under the results table rather than silently — hoisting the condition past the `fixed_any`-false exit is what reaches it, and the third parametrization pins that path. The `install-tend` Kickoff bullet now matches `Config.load`, which refuses `raw`, a `with` beside a `run`, and all of `setup` under yolo, but does accept `uses` in maintainer mode.

Not approving, but nothing in the diff holds the verdict: the PR body gates this merge on #1382 landing first and this branch then being updated from main, and #1382 is still open. That condition is yours to retract.

tend-agent:
Agreed, and this repo is a live instance of the hazard: `refs/tags/v1` (`4fa948b`) and `refs/heads/v1` (`6f46a95`) both exist here, at different commits. A name-only exception would already resolve wrong for one spelling in this repository, so withdrawing that finding — the unverified result stands.

One operational residual from the `fix_branch_protection` thread, where your `protected_branches` remedy is the fix but the branch doesn't yet carry it: the lifecycle gate makes `branch-protection:main` a `FAIL` here, not just a `--fix` no-op. `GET /repos/max-sixty/tend/rules/branches/main` returns a single `update` rule (ruleset `14219339`), so `creation` and `deletion` both resolve to `absent` and the check returns *"Branch 'main' lets the bot create or delete the ref."* `--fix` can't clear it — `Merge access` still includes `refs/heads/v1`, which isn't in `intended_refs` while `protected_branches` is empty. `/tend-ci-runner:nightly` step 2 runs `tend@latest check` and files `tend check: configuration drift on max-sixty/tend` on any `FAIL`, so this becomes a standing drift issue from the first nightly after the release carrying the gate.

Declaring `v1` clears both arms: `preserves_targets` passes, `--fix` writes `refs/heads/v1` into a new admin-only `Protected branch access` ruleset with the full creation/update/deletion set, and `Merge access` is rewritten for the default branch alone. Worth naming as a side effect — `v1` then joins `operational_refs`, so the `tend` environment's deployment branch policy admits it. Safe on the same reasoning as any extra protected branch (admin-only, bot-inaccessible), but it is a widening rather than a no-op. Happy to push the config line onto this branch or #1382, whichever you want it sequenced with.

tend-agent (commented):
**The base merge landed nothing — `7fb5c88` is byte-identical to its yolo parent.** Both trees are `54d1f55d0c261e50d7b0ca9b5a56ded288b29929`, so `git diff e9e031e 7fb5c88` is empty and every conflicted file resolved to this branch's side. `359832d`'s exclusion retention and its target-accurate `--fix` message are absent from HEAD: `fix_branch_protection` still holds the `preserves_targets` pre-check that returns `passed=False` and writes nothing, and `docs/security-model.md` still documents that refusal. This is the second merge from `codex/standard-hardening` to resolve `checks.py` that way — at `5bfa8d9` the only file taken from the hardening side was `integration-test.md`, and at `7fb5c88` nothing was.

The consequence is for the sequencing the PR body sets out. Once #1382 lands and this branch is updated from main, the same conflict recurs on the same file; resolved the same way, merging yolo would remove #1382's retention path from main rather than build on it.

The two sides answer the same hazard differently, so this is a design call rather than a bad resolution to redo mechanically. #1382 keeps the write and makes it safe — union the includes, copy the existing exclusions, report the refs actually written. This branch declines the write and asks the maintainer to retire the ref by hand, which is stricter but cannot repair a ruleset that already carries a ref outside the config — the state this repo is in, as noted earlier on this PR. Worth picking one and stating it once in `checks.py`, so the next update from main has nothing left to resolve. Happy to redo the resolution here once you name the side.

docs/security-model.md
```diff
@@ -79,33 +87,54 @@ now. Turning the setting off is therefore invisible to the nightly run until
...
-must first remove or independently gate its access to credential environments,
-then remove its ruleset target. Repairing protection never makes an old ref
-writable while it may still have access to secrets.
+**Merge rulesets.** Tend separates default-branch merge access from protection of
+other long-lived refs. `Merge access` targets only the default branch. In
+`maintainer`, only admins bypass its update rule. In `yolo`, the bot user also gets
+GitHub's `pull_request` bypass mode: it may update the branch through a PR but
+not by direct push. `Protected branch access` targets only configured
+`protected_branches` and stays admin-only. Both rulesets protect ref creation,
+updates, and deletion, so a bot cannot delete and recreate an admitted ref.
```
tend-agent: Maintainer mode no longer splits the rulesets: `fix_branch_protection`'s `merge == "maintainer"` arm writes `_restrict_updates_ruleset(extra)` into `Merge access` and never creates `Protected branch access`.

```suggestion
other long-lived refs. In `maintainer`, one admin-only `Merge access` ruleset
covers the default branch and every configured `protected_branches` entry. In
`yolo` it carries a bot bypass that must not reach any other ref, so it keeps
the default branch alone — the bot user gets GitHub's `pull_request` bypass
mode there, updating the branch through a PR but not by direct push — and
`Protected branch access` takes the configured `protected_branches` and stays
admin-only. Both rulesets protect ref creation, updates, and deletion, so a bot
cannot delete and recreate an admitted ref.
```

docs/security-model.md
```diff
@@ -79,33 +87,54 @@ now. Turning the setting off is therefore invisible to the nightly run until
...
+Because reconciliation splits targets across rulesets with different bypass
+authority, `tend check --fix` refuses to retire existing protected refs: a maintainer must
```
tend-agent: The refusal is the yolo arm's `preserves_targets` pre-check. Maintainer `--fix` goes through `_put_ruleset`, which unions the existing includes and preserves the excludes.

```suggestion
Maintainer `--fix` unions an existing ruleset's includes and preserves its
excludes, so a ref that may still admit credentials is never silently
unprotected. Because a yolo transition instead splits targets across rulesets
with different bypass authority, it refuses to retire existing refs: a maintainer must
```

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -380,43 +391,43 @@ user that team members should @-mention the bot account instead of `@claude`.
...
-**Merge restriction on the default branch.** Create if missing:
+The command is expected to remain non-zero until later steps install the
+secrets. Fix every ref-protection finding now. Under `maintainer`, `Merge access`
+keeps the default branch admin-only. Under `yolo`, it grants the bot user a
+pull-request-only bypass, while `Control-plane review` requires a fresh,
+non-bypassable CODEOWNER approval for `.github/**` and
+`.config/tend.yaml`, including the CODEOWNERS files and agent instructions.
+`Protected branch access` keeps creation, updates, and deletion of every
+configured extra branch admin-only in both modes. `Tag operations` keeps all
+tags admin-only.
```
```suggestion
Configured extra branches stay admin-only in both modes: `maintainer` keeps
them in `Merge access` beside the default branch, `yolo` moves them to a
separate `Protected branch access` ruleset. `Tag operations` keeps all
tags admin-only.
```

plugins/install-tend/skills/install-tend/references/security-model.md
```diff
@@ -6,25 +6,31 @@ canonical, full threat model is maintained in the tend source repo at
...
-- **Changing protected branches.** A ruleset with `creation`, `update`, and
-  `deletion` rules on the default branch and any `protected_branches`, with
-  admin-only bypass. Blocks the bot from landing code or deleting and
-  recreating a branch admitted to credential environments.
+- **Updating the default branch.** `Merge access` protects creation, update,
+  and deletion, with an admin bypass in both
+  modes and, in yolo only, a bot-user `pull_request` bypass. `Control-plane
+  review` layers fresh CODEOWNER approval over workflow and Tend-config paths.
+- **Updating extra protected branches.** `Protected branch access` protects
+  creation, update, and deletion and remains admin-only under both modes.
```
tend-agent:
```suggestion
- **Updating extra protected branches.** Creation, update, and deletion stay
  admin-only in both modes — under `Merge access` in `maintainer`, and under a
  separate `Protected branch access` ruleset in `yolo`, which keeps `Merge
  access` and its bot bypass on the default branch alone.
```

tend-agent (commented):
**The `Protected branch access` prose no longer matches the maintainer arm.** The resolution took main's side for `merge == "maintainer"`: a single `_put_ruleset(_restrict_updates_ruleset(extra))`, so the default branch *and* every `protected_branches` entry live in one admin-only `Merge access`, and `Protected branch access` is never created. Three places still describe the split as unconditional, and two of them ship — `docs/security-model.md` ("`Merge access` targets only the default branch … `Protected branch access` targets only configured `protected_branches`"), the install-tend skill ("keeps … every configured extra branch admin-only in both modes"), and its `references/security-model.md` ("remains admin-only under both modes"). The skill's sentence sits under "Fix every ref-protection finding now", so an agent installing in the default mode is told to look for a ruleset `--fix` doesn't write. Inline suggestions on each.

The next sentence in that same paragraph is mode-specific now too: "Because reconciliation splits targets across rulesets with different bypass authority, `tend check --fix` refuses to retire existing protected refs". Maintainer `--fix` doesn't refuse — `_put_ruleset` unions the existing includes and preserves the excludes, which is #1382's retention; the refusal is the yolo arm's `preserves_targets` pre-check. That also settles the design question left open last round: both answers are kept, one per mode, and the new docstring says why.

One code-side residual worth a decision rather than a doc edit: the maintainer arm removes `Control-plane review` explicitly but leaves `Protected branch access` standing. A repo switching `merge: yolo` back to `merge: maintainer` keeps that ruleset over refs `Merge access` now covers as well, and nothing retires it afterwards — `tend check` only reads the branches the config names, so a later `protected_branches` edit can't surface it either. Both rulesets are admin-only, so this over-restricts rather than exposes; the choice is between a `_remove_ruleset` beside the `Control-plane review` removal and a line in the docs saying it is left in place deliberately.

Happy to push the three prose corrections as one commit if you'd rather not apply them by hand.

tend-agent (approved):

PR #1173, opened 2026-09-08
Run agents in a disposable SRT workspace

max-sixty:
The current review and mention workflows replace the Actions checkout with event-controlled code before the agent runs. That makes trusted runner setup and local-action POST chains share state with the untrusted agent workspace, and it leaves security guarantees spread across harness-specific wrappers.

This change makes the trust-state transitions explicit and puts the boundary at one shared bottleneck. The runner checkout remains a stable, reviewed orchestration tree. A separate disposable clone is prepared at the exact event commit, then `sandbox_setup` and the full Claude or Codex turn run together under one Anthropic Sandbox Runtime process tree. Codex disables its nested sandbox inside that outer boundary, so Claude and Codex rely on the same isolation mechanism rather than divergent layers.

The trusted supervisor owns lifecycle and teardown: it handles cancellation, reaps the sandbox UID, verifies quiescence, and only then exports fixed, bounded, no-follow outputs. Codex next stops its runner-owned Responses proxy; one fixed cleanup then deletes both the event checkout and the per-run SRT/package directory. No later step executes event-tree or runtime code. Subsequent steps consume only trusted exports or bounded sandbox-home logs. The sandbox never receives runner Git configuration or GitHub command-file paths. Sensitive startup configuration is restored from the exact base commit at the ingress bottleneck, with runner and system Git execution hooks disabled.

Codex API-key mode routes only the model client through the runner-owned Responses proxy. Its native `shell_environment_policy` restores SRT loopback exclusions for spawned tools, so local test servers remain inside the sandbox. Subscription mode retains SRT routing unchanged.

The security documentation names the four execution states—trusted setup, prepared, running, and quiescent—and the seams between them. Generated workflows no longer swap or repair the runner checkout, and `setup` remains runner-owned while `sandbox_setup` is explicitly event-tree code inside SRT.

Ubuntu 24.04 enables an AppArmor restriction that strips capabilities from unprivileged user namespaces. On disposable GitHub-hosted VMs, the action records rollback intent, temporarily applies SRT’s documented sysctl prerequisite, and restores it after reap and resource disposal. Self-hosted policy is never changed; an administrator may install a scoped AppArmor profile or set the documented sysctl, and the action now reports the concrete `bwrap` failure to expect when neither prerequisite is present.

Validation:

- `wt test` — 1,118 Python tests, 32 worker tests, and TypeScript typechecking passed.
- `uv tool run pre-commit run --all-files` — all 13 hooks passed.
- Astra final security/design review — clean; 120 targeted tests, shell syntax, and diff checks passed.
- CI includes an Ubuntu integration probe for namespace isolation, denied runner-workspace access, allowed disposable-workspace access, direct host-loopback denial, host-broker routing, sandbox-local loopback routing, both harnesses, reap, and bounded export.

The remaining documented limitation is unchanged in kind: the Tend HTTP credential broker can proxy arbitrary destinations, so code readable inside the disposable workspace can still be exfiltrated through that broker. Destination allowlisting is separate from this filesystem and lifecycle bottleneck.


shared/steps/prepare_agent_workspace.py
```diff
@@ -0,0 +1,379 @@
...
+        if "@github.com" in origin or "x-access-token" in origin:
+            raise ValueError("agent clone persisted a credential in its origin URL")
+
+        with github_env.open("a", encoding="utf-8") as stream:
+            stream.write(f"TEND_RUNNER_WORKSPACE={runner_workspace}\n")
+            stream.write(f"TEND_AGENT_WORKSPACE={destination}\n")
+            stream.write(f"TEND_CONFIG_BASE_SHA={config_base_sha}\n")
+        log(f"prepared {destination}: {selected}")
+        return 0
+    except (OSError, subprocess.CalledProcessError, ValueError) as problem:
```
tend-agent: `b0677ea` moved four raises to `TypeError` (`api_json`, `event_payload`, and both `checkout_mention` topology guards), which this clause doesn't catch — so those paths escape `main()` as an unhandled traceback and never reach `fail()`'s `::error::` annotation. `PR #N has no fetchable head repository` fires on a fork PR whose head repo was deleted, which is reachable rather than theoretical.

```suggestion
    except (
        OSError,
        TypeError,
        subprocess.CalledProcessError,
        ValueError,
    ) as problem:
```

tend-agent (commented):
Big, coherent change — the four-state table in `docs/security-model.md` and the single ingress bottleneck in `prepare_agent_workspace.py` make the trust transitions far easier to reason about than the checkout-swap they replace. Reviewed at `b0677ea`; the npm `--userconfig`/`--globalconfig` collision that was failing `test-sandbox` on the previous head is fixed there, so `verify-srt` gets its first real run on this push. Worth waiting for: the claim this PR rests on — that SRT's filesystem, network and seccomp boundary initializes on GitHub's hosted ubuntu-24.04 — has not yet executed in CI, and that is the exact platform whose bubblewrap limitation the removed `sandbox: danger-full-access` comments cited. The `--dangerously-bypass-approvals-and-sandbox` switch in `codex/runner.py` is only safe if the outer boundary is real.

**The formatting pass turned four failure paths into tracebacks.** `b0677ea` changed `api_json`, `event_payload`, and both `checkout_mention` topology guards from `ValueError` to `TypeError`, but `main()` still catches `(OSError, subprocess.CalledProcessError, ValueError)`. A fork PR whose head repository was deleted — reachable, and exactly what `PR #N has no fetchable head repository` is for — now exits with a raw traceback instead of the `::error::` annotation `fail()` exists to emit. Inline suggestion on the `except` clause.

**The production launch path has no coverage.** After this change `run_claude.py` and `codex/runner.py run` are only ever reached from `agent_lifecycle.py`, which always sets `TEND_INSIDE_SANDBOX=1`. But `verify-launch` and `verify-codex-launch` still invoke them directly on the runner with that variable unset, i.e. through the `sudo -u "$SANDBOX"` crossing production no longer performs; no test anywhere sets `TEND_INSIDE_SANDBOX`; and `verify-srt` runs `TEND_HARNESS=probe`, which returns from `agent_lifecycle.main()` before either harness starts. So the new `start_new_session` / `os.killpg` timeout-and-kill escalation ships untested, and the two jobs that look like they cover it are exercising a configuration that no longer exists — the `test-sandbox` comment still calls them "the only places `run_claude.py` and `codex/runner.py` cross the uid boundary at all".

**The `sandbox_setup:` PATH-gap report is now unreachable.** Both actions' "Run adopter sandbox setup" steps are gone, so `sandbox_setup.main()` only runs from inside SRT, where `inside_sandbox` is true — and the new guard skips `report_reachability` there. Its own docstring is the argument against dropping it silently: "The report runs whether or not the input is set — an empty `sandbox_setup:` is exactly the config a missing tool hides in." `report_reachability` and `blocked_shims` are now dead in production, kept alive only by their unit tests and by `test-setup-sandbox.sh`. Either run the comparison from a trusted step that still exists, or delete both functions and their tests — the comment saying it "runs only in the trusted preflight form of this module" names a form the PR removed.

**`setup:` no longer reaches the agent's tree, and the adopter docs don't say so.** Previously review landed the PR tree over the same workspace with `clean: false`, so anything `setup:` built there (`node_modules`, `.venv`, a `cargo build --release` target) was visible to the agent. Now it is a separate clone. `docs/tend.example.yaml` still tells adopters only that runner-*home* changes don't carry over, and its own `setup:` example is `run: cargo build --release` — which is now a no-op as far as the agent is concerned. An adopter upgrading loses prebuilt dependencies with no error and no changed config. The `sandbox_setup:` section says where the work moves; the `setup:` section should say that workspace artifacts stop crossing.

Smaller: the SRT version is pinned in three unlinked places (`claude/action.yaml`, `codex/action.yaml`, and a literal `SRT_VERSION: "0.0.75"` in `ci.yaml`) with nothing asserting they agree, unlike `codex_version`, which has `test_codex_actions_pin_the_same_cli_version`. The weekly pin sweep in `.claude/skills/running-tend/SKILL.md` enumerates `*/action.yaml` inputs, so the `ci.yaml` copy is the one that will drift — and when it does, the boundary probe tests a version adopters don't run.

shared/steps/run_claude.py
```diff
@@ -281,12 +287,20 @@ def supervise(
...
                 # `sudo` exits once its child does, so this observes the agent.
                 with contextlib.suppress(subprocess.TimeoutExpired):
                     agent.wait(TERM_GRACE_SEC)
     finally:
-        signal_sandbox("KILL", sandbox)
+        if inside_sandbox and agent is not None and agent.poll() is None:
+            with contextlib.suppress(ProcessLookupError):
+                os.killpg(agent.pid, signal.SIGKILL)
+        elif not inside_sandbox:
+            signal_sandbox("KILL", sandbox)
```
tend-agent: The docstring above still says the KILL is "this function's `finally` and the only unconditional step ... no path out of here, exception included, may skip it" — and the reason given is that a survivor would race the steps that read the run's files. The `agent.poll() is None` guard makes it conditional exactly on the common path: a turn that exits normally leaves anything it backgrounded alive, still holding the inherited stdout/stderr. `sandbox_runtime.mjs` resolves on the child's `close` event, which waits for those pipes, so a leaked background process stalls the supervisor rather than being reaped by it. The old `signal_sandbox("KILL", sandbox)` fired unconditionally, and `killpg` still signals surviving group members after the leader has exited, so dropping the guard restores that.

```suggestion
        if inside_sandbox:
            if agent is not None:
                with contextlib.suppress(ProcessLookupError):
                    os.killpg(agent.pid, signal.SIGKILL)
        else:
            signal_sandbox("KILL", sandbox)
```

proxy/test-setup-sandbox.sh
```diff
@@ -416,13 +429,66 @@ verify_launch() {
...
+  for _ in {1..50}; do
+    [ -s "$probe_info" ] && break
+    sleep 0.1
+  done
+  probe_port=$(cat "$probe_info")
+  rc=0
+  ACTION_PATH="$TEND_TEST_ACTION_PATH" \
+    TEND_HARNESS=probe \
+    TEND_LIFECYCLE="$TEND_TEST_ACTION_PATH/shared/steps/agent_lifecycle.py" \
+    TEND_SANDBOX_SETUP='' \
```
tend-agent, line 351: `test ! -e "$TEND_RUNNER_WORKSPACE/.tend-srt-wrote-here"` below is vacuous — nothing in the repo ever tries to create that path, so the assertion passes whether or not SRT denies writes to the runner checkout. `probe_boundary()` covers the read side; giving the probe something to attempt covers the write side too.

```suggestion
    TEND_SANDBOX_SETUP="touch $TEND_RUNNER_WORKSPACE/.tend-srt-wrote-here || true" \
```

tend-agent (commented):
Reviewed at `63c38f3`. The `/tmp` container and the `TypeError` fix read clean; two things in the two most recent commits need a decision before this lands.

**`verify-srt` was moved off the layout it exists to certify — or production needs the same move.** `63c38f3` ("Place the simulated action cache on a traversable path") relocates `TEND_TEST_ACTION_PATH` from `$RUNNER_TEMP` into the new `/tmp` container. That is only warranted if a `$RUNNER_TEMP` path is unreachable from inside SRT — and if it is, production is exactly where the fixture just left: `ACTION_PATH` is `${{ github.action_path }}/..`, i.e. under `/home/runner/work/_actions`; `TEND_LIFECYCLE` (the command `sandbox_runtime.mjs` actually spawns) sits under it; and `install-sandbox-runtime.sh`'s `TEND_SRT_SECCOMP`, `setup_sandbox.py`'s `agent_env_file` and the Codex action's `codex_root="$RUNNER_TEMP/tend-codex"` all resolve under `$RUNNER_TEMP`. Every one of those five is in the `allowRead` array `sandbox_runtime.mjs` builds, so they are paths the sandbox is expected to reach, not incidental runner state.

The same test argues the other way in the same function: `verify_srt` still sets `tool_root="$RUNNER_TEMP/tend-codex-probe"` and passes it as both `TEND_BOUNDARY_PROBE_EXECUTABLE` and `TEND_CODEX_ROOT`, and `probe_boundary()` raises `SRT capability probe cannot execute the harness tool` unless that executable runs inside the sandbox. Both cannot hold. Either `$RUNNER_TEMP` is reachable, the fixture move is unexplained, and `verify-srt` now green-lights a path layout the shipped action does not use; or it is not reachable, and the probe assertion fails alongside the four production paths. Whichever it is, the resolution belongs in the action (or in a comment on `plant()` saying why the fixture may diverge), not only in the fixture.

**The disposable workspace is no longer disposed.** `RUNNER_TEMP` is cleared by the Actions runner at the start of each job; `/tmp` is not, and nothing in either action removes the container. `prepare_agent_workspace.py` doesn't even export its path — only `TEND_AGENT_WORKSPACE` reaches `$GITHUB_ENV`, so no later step can derive the parent to clean, while the test fixture does export `TEND_AGENT_CONTAINER`. On an ephemeral hosted runner this costs nothing. On a reused one it accumulates a full clone per run, each `tend-sandbox`-owned and `0700`, so the runner user cannot remove them without sudo — and the next run's agent, sharing that uid, can read the previous run's tree. If hosted-ephemeral is the assumed model that's a fine answer, but the doc calls this checkout disposable and nothing disposes of it.

Inline suggestion on the third point: the new `RUNNER_TEMP` sentence in `docs/security-model.md` is contradicted by that same `allowRead` list.

docs/security-model.md
```diff
@@ -279,6 +279,45 @@ same way. Configuration recipe:
...
+
+Each transition is a bottleneck with one job:
+
+- **Content ingress** creates a full remote clone inside a dedicated `/tmp`
+  container with no hardlinks or object-store alternates and with runner/system
+  Git config and attributes disabled. It selects the exact base, PR merge/head,
+  or open-PR head ref, pins startup configuration to the exact chosen base
+  commit, removes the temporary credential, makes the otherwise-empty parent
+  traversable but not listable, then changes ownership of that clone only.
+  `RUNNER_TEMP` remains runner-only.
```
tend-agent: `sandbox_runtime.mjs` puts three `$RUNNER_TEMP` paths in the sandbox's `allowRead`: `AGENT_ENV_FILE` (`setup_sandbox.py`'s `agent_env_file`, i.e. `tend-agent-env`), `TEND_SRT_SECCOMP` (`install-sandbox-runtime.sh`'s `srt_package/vendor/seccomp/...`), and `TEND_CODEX_ROOT` (the Codex action's `codex_root="$RUNNER_TEMP/tend-codex"`, from which the sandbox also executes `CODEX_BIN`). An unqualified "runner-only" in the security model overstates what the boundary gives.

```suggestion
  `RUNNER_TEMP` holds no tree the agent receives, but the launch env file, the
  seccomp binary and the Codex CLI root stay individually readable inside the
  sandbox.
```

shared/steps/run_claude.py
```diff
@@ -108,34 +105,19 @@ def launch_argv(
...
-            if not entry.startswith("CLAUDE_CODE_DISABLE_AUTO_MEMORY=")
-        ]
     argv = [
-        "sudo",
-        "-u",
-        sandbox,
-        "env",
-        *agent_env,
+        "/usr/bin/env",
+        *(["CLAUDE_CODE_DISABLE_AUTO_MEMORY="] if settings_file else []),
```
tend-agent: The code this replaced filtered the entry out of the composed launch env, so the variable was **absent** from the agent's environment; an empty assignment leaves it **present**. Whether Claude Code reads that as "not disabled" depends on it testing truthiness rather than presence — and the comment that was deleted here made the point that a surviving value would "make a successful restore silently inert even though the injected settings enable and redirect auto memory". `env -u` restores absence exactly, and options are legal before the assignments that follow.

```suggestion
        *(["-u", "CLAUDE_CODE_DISABLE_AUTO_MEMORY"] if settings_file else []),
```

tend-agent (commented):
Reviewed at `6261224`. Every point from the previous round lands: the `TypeError` catch, the unconditional `killpg`, `report_reachability` deleted rather than left dead in production, the `setup:`-no-longer-crosses note in both `tend.example.yaml` copies, a real disposal step, and a pin-parity test for the three SRT versions. Collapsing `verify-launch`/`verify-codex-launch` into one `verify-srt` that drives both harnesses through the actual lifecycle is the right move — the production path is now the tested path.

**`verify-srt` reached the boundary, and the boundary did not come up.** `test-sandbox` fails at this head in "Verify complete Claude and Codex SRT lifecycles" with `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted` ([job](https://github.com/max-sixty/tend/actions/runs/34184986108/job/101931412695)). It appears immediately after `sandbox_runtime.mjs` prints `::stop-commands::`, so `SandboxManager.initialize` and `wrapWithSandboxArgv` both succeeded and the failure is the bwrap exec itself, setting up `lo` in the unshared network namespace. This is the claim the whole change rests on — SRT's filesystem, network and seccomp boundary is what replaces the checkout swap, and `codex/runner.py`'s `--dangerously-bypass-approvals-and-sandbox` is only defensible if that outer boundary is real. It is not a regression from the last push: `verify-srt` has never completed on this branch, because every earlier head failed in an earlier step. As a lead rather than a diagnosis — `kernel.apparmor_restrict_unprivileged_userns` is `1` on the hosted ubuntu-24.04 image (read on the runner this review ran on) and the base image carries no `bwrap` AppArmor profile, which is the documented cause of exactly this error. Whatever the remedy turns out to be — a profile, the sysctl, or an SRT network config that does not unshare the netns — it needs root, so it belongs next to the `apt-get install bubblewrap` in `install-sandbox-runtime.sh`, not anywhere the agent can reach.

**The runtime root is now as undisposed as the workspace container was.** `install-sandbox-runtime.sh` moves SRT — and through `TEND_RUNTIME_ROOT`, the Codex install and `agent-env` — out of runner-cleaned `$RUNNER_TEMP` into `mktemp -d /tmp/tend-runtime.XXXXXX` at mode 755, and nothing removes it: `dispose_agent_workspace.py` validates `TEND_AGENT_WORKSPACE` and deletes only the `tend-agent-workspace-*` container. On a reused runner that is a full SRT plus a Codex `node_modules` per run, and the previous run's `agent-env` left readable by the next run's sandbox uid — the same argument that earned the checkout its disposal. Extending the dispose step is the small fix, but the ordering matters: `CODEX_PROXY_BIN` lives under the runtime root and "Stop OpenAI Responses proxy" runs *after* "Dispose disposable agent workspace".

**`AGENT_PATH` and `TEND_BLOCKED_PATH` are now write-only.** `report_reachability` was their only reader; with it deleted, `setup_sandbox.py` still exports both to `$GITHUB_ENV` and `launch_sandbox_runtime.PASSTHROUGH` still forwards both into the sandbox environment, where nothing opens them. The blocked-shim mechanism itself is untouched and still covered by `verify` — it is the two names that stopped meaning anything, and CLAUDE.md's "Prefer deleting a mechanism over refining it" applies to the plumbing as much as the function.

**The inner `GITHUB_OUTPUT` has no reader.** Inside the lifecycle `GITHUB_OUTPUT` is `TEND_RUN_DIR/tend-step-output`, and `launch_sandbox_runtime.py` never opens that file — it recomputes `stream_json` and `final_message` itself from the fixed export paths and appends them to the real output. So `run_claude.py`'s `_common.set_output("stream_json", …)` and `codex/runner.py`'s `final_message=` write into a file nothing reads, and the comment above the first still justifies the write by the consumers it no longer serves ("the steps that read it — Token usage, the session-logs artifact"); those now read the supervisor's export.

**The repo's own skill documents a session-log path this PR invalidates.** `.claude/skills/running-tend/SKILL.md` states `Artifact paths: -home-runner-work-tend-tend/<session-id>.jsonl`. With the agent's cwd now `/tmp/tend-agent-workspace-<random>/checkout`, Claude Code's project directory becomes `-tmp-tend-agent-workspace-<random>-checkout`, different on every run, so the documented name will never match again and anyone following it will conclude the artifact is empty. Outside this diff — say the word and I'll push the correction onto this branch.

docs/security-model.md
tend-agent: Two of the three moved out of `RUNNER_TEMP` entirely in `de4caea` — `agent_env_file` is now `TEND_RUNTIME_ROOT/agent-env` and `codex_root` is `$TEND_RUNTIME_ROOT/codex`, both under `/tmp/tend-runtime.*`, and `TEND_SRT_SECCOMP` follows `srt_root` there — so the suggested wording is stale and the fix is better than the one I proposed. Leaving the thread open only for the sentence itself: `sandbox_runtime.mjs` still puts `eventPath` in `allowRead`, and `GITHUB_EVENT_PATH` resolves under `$RUNNER_TEMP/_github_workflow/`, so "`RUNNER_TEMP` remains runner-only" is still one file short of true. Handing the agent the event payload is deliberate, so this is a wording fix, not a boundary change.

shared/steps/install-sandbox-runtime.sh
```diff
@@ -0,0 +1,73 @@
...
+
+apparmor_userns=/proc/sys/kernel/apparmor_restrict_unprivileged_userns
+if [ -r "$apparmor_userns" ] && [ "$(/usr/bin/cat "$apparmor_userns")" = 1 ]; then
+  if [ "${TEND_RUNNER_ENVIRONMENT:-}" = github-hosted ]; then
+    # Record rollback before changing host state so cancellation or a failed
+    # sysctl cannot leave an untracked policy change.
+    echo "TEND_RESTORE_APPARMOR_USERNS=true" >> "$GITHUB_ENV"
+    /usr/bin/sudo /usr/sbin/sysctl -q -w kernel.apparmor_restrict_unprivileged_userns=0
+  else
+    echo "::notice::Leaving self-hosted AppArmor policy unchanged; SRT will verify that bwrap and apply-seccomp have capability-bearing user namespaces"
```
tend-agent: The notice is the only thing a self-hosted operator sees for this condition, and the failure it precedes doesn't name AppArmor, user namespaces, or capabilities — at `6261224` SRT initialized and then died in `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`. Carrying the symptom and both remedies into the notice makes the later errno attributable without reading `docs/security-model.md`.

```suggestion
    echo "::notice::Leaving self-hosted AppArmor policy unchanged; without a bwrap AppArmor profile granting unprivileged user namespaces, SRT fails as 'bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted' — install such a profile or set kernel.apparmor_restrict_unprivileged_userns=0 on the runner"
```

tend-agent (commented):
Reviewed at `793d69d`. `test-sandbox` is green at this head, so the claim the whole change rests on — that SRT's filesystem, network and seccomp boundary initializes on GitHub's hosted `ubuntu-24.04` — now executes rather than being asserted. The rollback marker recorded before the sysctl, the `runner.environment` gate, restoring to the exact value the guard already proved was set, and `if: always()` on both restore steps close the window from every exit including cancellation; and the new `--noproxy '*'` leg turns netns isolation into a positive assertion instead of an inference from the proxied probe.

**The self-hosted notice promises a diagnosis SRT did not deliver.** The one observation this branch has of exactly that condition is head `6261224`, where SRT initialized — it had already printed `::stop-commands::` — and the run then died in `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`: a netlink errno naming neither user namespaces, nor AppArmor, nor capabilities. Self-hosted is the only path that reaches this branch and no CI job exercises it, so an operator on stock 24.04 gets a notice telling them SRT will verify the prerequisite, then a failure that doesn't mention it, with both remedies stated only in `docs/security-model.md`. Inline suggestion moves the symptom and the two remedies into the notice itself, and a second one drops the same "SRT's capability probe fails closed" claim from the doc.

That doc suggestion also picks up the probe enumeration in the sentence above it: this commit adds host-loopback unreachability as a third pre-setup probe — the one that certifies the netns this change exists to bring up — and the list still names only two.

docs/security-model.md
```diff
@@ -279,6 +279,53 @@ same way. Configuration recipe:
...
+  Claude or Codex turn as one command under the pinned Anthropic Sandbox
+  Runtime. Tend supplies absolute `node`, `bwrap`, `socat`, `rg`, and seccomp
+  paths, treats dependency warnings as fatal, and probes both AF_UNIX denial
+  and runner-checkout unreadability before setup executes. Ubuntu 24.04's
+  AppArmor policy strips the user-namespace capabilities SRT needs. Tend
+  temporarily applies SRT's documented sysctl prerequisite on disposable
+  GitHub-hosted VMs and restores it immediately after the sandbox is reaped;
+  Tend never changes self-hosted policy, where a scoped AppArmor profile may
+  satisfy the same prerequisite and SRT's capability probe fails closed if it
+  does not.
```
tend-agent: Two edits in one range. `probes both AF_UNIX denial and runner-checkout unreadability` is now one short: this commit adds the direct `--noproxy '*'` leg to `probe_boundary()`, which certifies the netns the rest of the paragraph is about. And `SRT's capability probe fails closed` attributes the self-hosted failure to a verification the one observation of it doesn't show — SRT initialized and bwrap failed at exec, which is fail-closed but not a capability diagnosis; naming the actual symptom is what an operator can match against their log.

Replacing lines 307–314 (not a one-click suggestion — GitHub anchored this comment to the last line of the range only):

```
  paths, treats dependency warnings as fatal, and probes AF_UNIX denial,
  runner-checkout unreadability, and host-loopback unreachability before setup
  executes. Ubuntu 24.04's AppArmor policy strips the user-namespace
  capabilities SRT needs. Tend temporarily applies SRT's documented sysctl
  prerequisite on disposable GitHub-hosted VMs and restores it immediately
  after the sandbox is reaped; Tend never changes self-hosted policy, where a
  scoped AppArmor profile may satisfy the same prerequisite. Without either the
  run fails closed inside `bwrap` with `Failed RTM_NEWADDR: Operation not
  permitted`, rather than with a capability diagnosis.
```

codex/action.yaml
```diff
@@ -384,38 +403,45 @@ runs:
...
       run: |
-        if [ -n "${SANDBOX:-}" ]; then
-          sudo pkill -KILL -u "$SANDBOX" 2>/dev/null || true
-          sudo chown -R "$(id -u):$(id -g)" "$GITHUB_WORKSPACE"
-        fi
+        set +x
+        PATH=/usr/sbin:/usr/bin:/sbin:/bin
         if [ -n "${CODEX_PROXY_URL:-}" ]; then
-          curl --fail --silent --show-error "$CODEX_PROXY_URL/shutdown" || true
+          /usr/bin/curl --fail --silent --show-error "$CODEX_PROXY_URL/shutdown" || true
```
tend-agent: `--fail --silent --show-error ... || true` already treats a failed shutdown as non-fatal, but it doesn't cover the shutdown that never answers. Before this commit that only cost the step; now it sits ahead of `Dispose sandbox resources` and `Restore Sandbox Runtime host policy`, so a wedged proxy burns the job to its `timeout-minutes` and leaves behind exactly what the new dispose step exists to remove — a sandbox-owned `0700` `/tmp/tend-agent-workspace-*` and the `/tmp/tend-runtime.*` root — on any runner that outlives the job.

```suggestion
          /usr/bin/curl --fail --silent --show-error --max-time 10 "$CODEX_PROXY_URL/shutdown" || true
```

tend-agent (commented):
Reviewed at `02d7f7f`, with `test-sandbox` green through the full Claude and Codex lifecycle step. Moving `/shutdown` ahead of the `rm -rf` is the necessary half of the teardown fix — node resolves `require` lazily, so deleting `codex_root` under a live proxy was unsafe — but it also puts an untimed HTTP call ahead of both `always()` teardown steps. Inline suggestion.

tend-agent (approved):

PR #1174, opened 2026-09-08
docs(CLAUDE.md): require the repo PAT scope the installer and audit already demand

tend-agent:
CLAUDE.md's Auth section offers a PAT scope choice that no code path supports: "a classic PAT (`public_repo` for public repos, `repo` for private)". An adopter who follows it on a public repository gets a token the tooling treats as broken. `pat_scope_audit.py` lists `repo` in `REQUIRED` unconditionally, and GitHub's `X-OAuth-Scopes` header reports a `public_repo` grant as `public_repo` rather than expanding it — so the audit reports `MISSING=repo`, and nightly Step 1 opens a `Bot PAT: missing scopes` issue for a token that was set up exactly as documented. The installer never offers the choice either: both `gh auth` paths in the `install-tend` skill pass `--scopes repo,workflow,notifications,write:discussion,gist,user`, and the completion checklist names the same six.

The same sentence understates the blast radius further down. `repo` grants full write to every repository the bot account can reach, private ones included — not "every public repository", which is what `public_repo` would have meant. `docs/security-model.md` already states this correctly ("every repository both its scope and the bot account can reach"); this brings CLAUDE.md into line with it.

Documentation only — the diff against `main` is CLAUDE.md alone, no behavior changes. This PR originally carried a regression test in `generator/tests/test_cross_file_contracts.py` pinning the one scope set across `pat_scope_audit.py`'s `REQUIRED`, the installer's `--scopes` line, and CLAUDE.md. #1212 has since deleted that whole file as part of dropping the tests that pin wording in skills and docs, so resolving the conflict took main's deletion: the scope set is now documented in one place without a substring assertion guarding it.


generator/tests/test_cross_file_contracts.py
```diff
@@ -381,3 +381,32 @@ def test_nightly_regen_stages_every_path_init_writes(
...
     )
+
+
+def test_bot_pat_scope_set_is_stated_identically_everywhere() -> None:
+    """One scope set, named by the installer, the audit, and the architecture doc.
+
+    The audit fails a token that lacks any required scope and the nightly files
+    a tracking issue for it, so a doc that offers a narrower alternative sends
+    an adopter down a path the audit reports as broken every night.
+    """
```
tend-agent, line 502: This test reads `CLAUDE.md` from disk, which a tend PR session pins to the base branch (`restore-sensitive-config.sh`) while leaving this file at the PR head — so it fails inside any agent session on a PR that edits the Auth text, with a message that reads as the fix being absent. Worth naming here so the next reader doesn't re-apply the edit (and stage the base version over it).

```suggestion
    The audit fails a token that lacks any required scope and the nightly files
    a tracking issue for it, so a doc that offers a narrower alternative sends
    an adopter down a path the audit reports as broken every night.

    Inside a tend PR session `CLAUDE.md` is pinned to the base branch, so this
    fails there whenever the PR is the thing fixing it; read `git show
    HEAD:CLAUDE.md` rather than re-applying the edit.
    """
```

tend-agent (commented):
The scope alignment itself checks out — `pat_scope_audit.py`'s `REQUIRED`, the installer's two `--scopes` lines, `docs/tend.example.yaml`, and `docs/security-model.md` all name the same unconditional `repo`, so CLAUDE.md was the lone outlier and the blast-radius sentence now matches the security model.

One concern with the regression test: it asserts on `CLAUDE.md`'s content, and `CLAUDE.md` is the one file a tend PR session does not hold at the PR's version — `restore-sensitive-config.sh` pins it to the base branch for the duration of a run while the test file stays at the PR head. On this PR, `uv run pytest -k pat_scope` inside a tend session fails with ``CLAUDE.md offers a `public_repo` PAT the audit rejects`` listing both base-branch lines, which reads as the fix not having been applied. Repo CI is unaffected (its checkout isn't pinned; `test` is green here). The exposure is an agent session that runs the suite on a PR editing this text and then tries to re-apply an edit the diff already carries — staging `CLAUDE.md` from a PR checkout commits the base version back over the PR's own change. A line in the docstring is enough to stop that.

generator/tests/test_cross_file_contracts.py
tend-agent: Applied in 76e5170 — the docstring now carries the caveat verbatim, so a session that hits the failure inside a PR run is pointed at `git show HEAD:CLAUDE.md` instead of re-applying the edit. Left the assertion reading the worktree copy: the check is a cross-file contract for repo CI, where the checkout is unpinned, and routing it through `git show` to survive one caller would be machinery in the wrong place.

PR #1175, opened 2026-09-08
fix(prepare-agent-workspace): stop a mention on an issue or closed PR mishandling the checkout

tend-agent:
Two mention events reach the disposable-workspace step with no PR head to fetch, and each one is mishandled: a mention on a **closed or merged PR** runs on stale repo guidance, and a mention added by **editing an issue body** kills the step before the agent starts. Both come out of the same branch in `main()`, so this PR replaces that branch rather than special-casing each.

**Closed PR — stale instruction files.** `checkout_mention` falls back to the default branch when the PR isn't open, but it still returned the closed PR's `base.sha` as the config base. `main()` hands that to `restore_sensitive_config`, whose job is to pin `CLAUDE.md`, `AGENTS.md`, `.claude/**` and the other startup-read paths to a reviewed commit — a real defence on a fork PR, where the checked-out tree is attacker-controlled. Here the tree is already the default branch, so there is nothing to pin away; the pin only rolls those files back to whatever they were when that PR opened, and every skill, convention and command the branch has gained since disappears for the run. The dirty worktree is the second cost — `git restore --worktree` leaves the index at `HEAD`, so a later `git add -A` stages the revert into whatever the run commits.

**Issue mention — the step dies.** `checkout_mention`'s first statement requests `/repos/{repository}/pulls/{number}`. The bypass around it was gated on `GITHUB_EVENT_NAME == "issue_comment"`, so an `issues: edited` event sent its *issue* number to the PR endpoint. That 404s, the `HTTPError` (an `OSError`) reaches `main()`'s handler, and the `Prepare disposable agent workspace` step exits non-zero with no `continue-on-error` above it. It is reachable from the workflow's own `if`, which admits `issues: edited` whenever the issue body mentions the bot, and `mention_verify.py` returns `verdict(True)` unconditionally for `kind == "issues"` — so every mention added by editing an issue body loses its entire run, silently from the thread's side. This is the other half of #1173's move of the PR checkout out of the workflow: the old runner step gated `gh pr checkout` on `github.event.issue.pull_request.url != '' || repository_dispatch`, which excluded `issues`; the condition that replaced it did not.

The fix is one question asked of the payload's own shape, `event_targets_pull_request` — an `issue` names a PR only through its `pull_request` link, and a relayed review names one in `client_payload` — plus an empty config base on the closed-PR return. Both no-PR cases then take the same answer: the reviewed default branch, unpinned.

<details><summary>Regression coverage</summary>

Both tests are in `shared/steps/test_prepare_agent_workspace.py` and fail on the previous code.

- The closed-PR test builds a `main` that has moved past the PR base, asserts the empty config base, and checks the restored tree still holds the current guidance with a clean `git status`. Pre-fix it fails with the base SHA where the empty string belongs.
- The issue-mention test drives `main()` with an `issues` payload and an `api_json` that raises the real 404, then reads `TEND_AGENT_WORKSPACE` back out of `GITHUB_ENV` and asserts the checkout is the default branch at its tip. Pre-fix `main()` prints `::error::HTTP Error 404: Not Found` and returns 1.

</details>

### DIFF
--- a/shared/steps/test_prepare_agent_workspace.py
+++ b/shared/steps/test_prepare_agent_workspace.py
@@ -2,8 +2,10 @@
 
 from __future__ import annotations
 
+import json
 import shutil
 import subprocess
+import urllib.error
 from pathlib import Path
 
 import prepare_agent_workspace as prepare
@@ -219,13 +221,12 @@ def test_event_number(payload: dict[str, object], number: int) -> None:
     [
         ({"issue": {"number": 1}}, False),
         ({"issue": {"number": 1, "pull_request": {"url": "example"}}}, True),
+        ({"client_payload": {"pr": "3"}}, True),
         ({}, False),
     ],
 )
-def test_issue_comment_targets_pull_request(
-    payload: dict[str, object], expected: bool
-) -> None:
-    assert prepare.issue_comment_targets_pull_request(payload) is expected
+def test_event_targets_pull_request(payload: dict[str, object], expected: bool) -> None:
+    assert prepare.event_targets_pull_request(payload) is expected
 
 
 def test_mention_on_a_closed_pr_keeps_the_default_branch_instructions(
@@ -258,8 +259,6 @@ def test_mention_on_a_closed_pr_keeps_the_default_branch_instructions(
         "api_json",
         lambda _path, _token: {"state": "closed", "base": {"sha": base}},
     )
-    monkeypatch.setattr(prepare, "BASH", Path(shutil.which("bash") or "/bin/bash"))
-
     selected, config_base = prepare.checkout_mention(
         destination,
         repository="owner/repo",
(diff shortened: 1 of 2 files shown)


tend-agent (commented):
The fix reads correctly: the closed-PR fallback checks out the default-branch tip the runner already handed over, so pinning the instruction files to the closed PR's base can only roll them backwards. I confirmed the new test fails on the previous return (`assert '2f17503…' == ''`) and passes here.

One concern, on the sibling path through the same branch. `checkout_mention` is also reached for `issues: edited`: `main()`'s bypass is gated on `GITHUB_EVENT_NAME == "issue_comment"`, and `mention_verify.py` returns `verdict(True)` unconditionally for `kind == "issues"`, so the function's first statement requests `/repos/{repository}/pulls/{number}` with an *issue* number. That 404s, `urllib` raises `HTTPError` (an `OSError`), and `main()`'s handler turns it into `::error::` with a non-zero exit — the `Prepare disposable agent workspace` step has no `continue-on-error`, so the job dies before the agent starts. It is reachable from the workflow's own `if`, which admits `issues: edited` whenever `contains(github.event.issue.body, '@<bot>')`: every mention added by editing an issue body loses its entire run.

This is the other half of #1173's move of the PR checkout out of the workflow. The old runner step gated `gh pr checkout` on `github.event.issue.pull_request.url != '' || repository_dispatch`, which excluded `issues`; the condition that replaced it does not.

Worth a decision rather than a suggestion: the two cases want the same answer — no PR head, no config pin — so the deeper fix is one "does this event target a PR at all" branch in `main()`, which removes the split that produced the gap, rather than a second special case beside this one. Folding it in here or landing it separately are both defensible; leaving it is the part I would not do, since the failure is silent from the thread's side.

shared/steps/test_prepare_agent_workspace.py
```diff
@@ -226,3 +226,51 @@ def test_issue_comment_targets_pull_request(
...
+        repository="owner/repo",
+        token="unused",
+        remote_url=str(origin),
+    )
+    monkeypatch.setattr(
+        prepare,
+        "api_json",
+        lambda _path, _token: {"state": "closed", "base": {"sha": base}},
+    )
+    monkeypatch.setattr(prepare, "BASH", Path(shutil.which("bash") or "/bin/bash"))
```
tend-agent: Dead in this test: `restore_sensitive_config` returns before it reads `BASH` when the config base is empty, which the assertion two lines below pins.

```suggestion
```

tend-agent: Removed — `restore_sensitive_config` returns on the empty base before it reaches `BASH`, so the pin never applied. `shutil` is still used by the two tests above.

tend-agent:
Folded the `issues` gap in here, as the deeper fix rather than a second special case: `main()` now asks `event_targets_pull_request(payload)` — an `issue` names a PR only through its `pull_request` link, a relayed review names one in `client_payload` — and both no-PR answers take the same branch, the reviewed default branch with no config pin. The `GITHUB_EVENT_NAME == "issue_comment"` gate is gone, so the split that produced the gap is gone with it.

Confirmed the failure before fixing it: driving `main()` with an `issues` payload and an `api_json` that raises the real 404 prints `::error::HTTP Error 404: Not Found` and returns 1 on the previous wiring, which is the step dying before the agent starts. That test is now in `shared/steps/test_prepare_agent_workspace.py` alongside the closed-PR one. Full `uv run pytest` passes (1121) and pre-commit is clean on both files.

Title and description recomposed around both halves. Pushed as fe4a9a7.

PR #1177, opened 2026-09-08
fix(nightly): read the covering PR's diff before surveying a path it already changes

### LINKED ISSUE #1176: nightly: Step 6 surveys files that already have an open PR, so a 28-day rotation over an unmerged backlog re-derives its own prior PRs
Nightly's Step 6 survey reads its file list straight from `nightly_survey_files.py` and reviews every path in it. Nothing between the script and the review checks whether a path already has an open PR against it. On a repo whose PRs merge, that costs nothing — the previous pass's PR is gone by the time the bucket comes round. On a repo with a backlog it costs a full review per aliased file, every cycle, forever.

## The aliasing is exact, not approximate

The rotation is deterministic in both directions: the day picks the bucket (`unix_day / 86400 % 28`) and the path picks the bucket (`cksum(path) % 28`). So bucket *N* returns to the *same* file set every 28 days. Once a repo's default branch has been static longer than the cycle, every bucket whose last pass left an unmerged PR re-derives that PR's findings on the way round — not occasionally, but on the exact anniversary.

That threshold was crossed on max-sixty/cargo-affected: `main` static 31 days against a 28-day cycle, 36 open bot PRs, roughly a third of them survey-shaped.

## What it cost on the run that tripped it

`tend-nightly` [`34195750626`](https://github.com/max-sixty/cargo-affected/actions/runs/34195750626) (2026-09-08, bucket 12/28) drew `tests/functional/clean.rs` and `tests/functional/db_has_function_ranges.rs`. It read both, found the unreachable `|| contains("no coverage data found")` alternative and a `pub` → `pub(crate)` error in a module doc, branched, wrote both fixes, `rustup component add llvm-tools`, installed `cargo-nextest`, `cargo build --tests`, ran the full functional suite, `clippy --all-targets` and `fmt --check` — and *then* checked the queue and found [max-sixty/cargo-affected#83](https://github.com/max-sixty/cargo-affected/pull/83), opened 2026-08-11 (28 days earlier, same bucket), touching the same two files with the same two fixes. The branch was discarded unpushed. Run cost $3.09 / 54 turns, most of it on that path.

Two earlier nightlies on the same repo landed the same way with less sunk cost — [`33724452938`](https://github.com/max-sixty/cargo-affected/actions/runs/33724452938) (09-03) found two of three survey defects already carried by open PRs #77 and #99, and [`33950286183`](https://github.com/max-sixty/cargo-affected/actions/runs/33950286183) (09-05) declined to file after its `src/selection.rs` finding turned out to be covered by #81.

## Why the existing guards don't cover it

- **Step 8's orienting projection** (`gh issue list`/`gh pr list --json number,title`) matches on titles. A survey PR's title describes the *fix* (`tests: correct two inaccurate statements in surveyed scenarios`), which a freshly re-derived finding has no reason to phrase the same way.
- **`running-in-ci`'s "Fetch the prior rejection before re-deriving a fix"** (added since `0.1.14`, present on `main`) does search by path, and firing per finding *before writing code* would have saved the build and test cycle above. But it is downstream of the read: the survey has already spent its review turns on the file by the time any finding exists to search for. It also depends on the model choosing the right `<symbol>` for each finding, where the survey already holds the exact path.
- **The pre-`gh pr create` recheck** is last by construction — it caught this one, which is why nothing wrong shipped, but only after everything was spent.

## Suggested direction

Design is yours. The cheapest shape looks like one call at the top of Step 6, before any file is read:

```bash
gh pr list --state open --limit 200 --json number,files \
  --jq '[.[] | {n: .number, paths: [.files[].path]}]' > /tmp/open-pr-files.json
```

Then drop from today's bucket any path that appears in it, noting the skipped paths and the PRs that cover them in the Step 9 summary so the skip is visible rather than silent. A file with an open PR against it isn't a good survey target regardless of the aliasing: whatever the survey finds there is either already in that PR or belongs as a comment on it, and the PR is the place a maintainer is already looking.

This is a property of the rotation rather than of any one repo, so it will surface anywhere a backlog outlives the 28-day cycle — most visibly on repos where the bot is the only author.

- Bucket confirmed by running the script in the same 24h window: `nightly-survey-files.sh` → `bucket=12/28 files=4`, listing the two test files plus two generated `tend-*.yaml` (skipped per the checklist).
- `gh pr view 83 --json files` returns exactly `tests/functional/clean.rs, tests/functional/db_has_function_ranges.rs`; its diff makes the same two edits the nightly's `Edit` tool calls made, re-read from the session log at `claude-session-logs-c508d87b`.
- #83 `createdAt` is `2026-08-11T06:41:20Z`; the nightly ran `2026-09-08T06:41:08Z` — 28 days and 12 seconds apart, which is the rotation, not a coincidence.
- Step 6 on tend `main` (`plugins/tend-ci-runner/skills/nightly/SKILL.md`) carries no dedup between the script invocation and "Apply the review checklist below to each file in full".

</details>

tend-agent:
## Problem

The survey's rotation is deterministic on both sides — the day picks the bucket (`unix_day % 28`) and the path picks the bucket (`crc32(path) % 28`) — so a bucket returns to exactly the same file set every 28 days. Nothing between the script and the review told the survey that a path already had an open PR against it. Where the bot's PRs merge promptly that costs nothing; where a backlog outlives the cycle, every bucket whose last pass left an unmerged PR re-derives that PR's findings on the anniversary, spending a full read-branch-build-test cycle before the pre-`gh pr create` recheck discards the branch. The existing guards all sit downstream of the read: Step 8's projection matches on titles, and `running-in-ci`'s prior-rejection search only has a symbol to search for once a finding exists.

## Solution

`nightly_survey_files.py` now asks `gh` which paths open PRs change and prints `# covered <path> (#N author)` on stderr for each path in the day's bucket that one of them touches. Step 6 reads that PR's diff before reading the file and drops the findings it already carries. Step 9 reports the covered paths.

The covered path stays in the day's list. Dropping it — the first shape here — saves more, but it blinds the survey exactly where the mechanism fires: the backlog that produces the aliasing is the same backlog that covers the tree. On [max-sixty/cargo-affected](https://github.com/max-sixty/cargo-affected), the repo that surfaced this, 36 open bot PRs cover 38 of 58 tracked files and 11 of 13 under `src/`, none of them wide (the largest touches 9 files) — so a drop would have left most of that repo unsurveyed for as long as the backlog stands, rather than for one cycle. Reading the covering PR's diff first still recovers the branch-build-test cycle the aliasing wasted, which was nearly all of its cost, and a file whose open PR fixes one line stays surveyable for the other hundred.

The query is unscoped by author, so a covering PR can be a maintainer's or an outside contributor's rather than the bot's — knowing a human PR is mid-rewrite of a file is worth having before the survey reads it, but the survey's own findings have no business in that thread. So the author rides along in the `# covered` line and gates the routing without a second lookup: a new finding goes onto the covering PR only where the bot wrote it, and into a PR of its own per Step 8 otherwise.

A `gh` failure fails the step rather than silently surveying unannotated. That is deliberate: a fallback would be machinery, and a run that cannot reach the API cannot open the PRs Step 8 exists for either.

## Testing

Two new tests in `generator/tests/test_nightly_survey_files.py`: one drives `main` with an injected PR list and asserts the covered path stays in stdout and is named on stderr with every PR covering it and its author; the other runs the script through `uv run --script` against a fake `gh`, pinning the actual `pr list --state open --limit 200 --json number,files,author` invocation. Full `uv run pytest` (1120 tests) and every pre-commit hook pass.

Closes #1176 — automated triage

### DIFF
--- a/plugins/tend-ci-runner/scripts/nightly_survey_files.py
+++ b/plugins/tend-ci-runner/scripts/nightly_survey_files.py
@@ -6,9 +6,12 @@
 
 Both the day and the path pick a bucket by arithmetic, so bucket *N* returns to
 the same file set every cycle. On a repository whose pull requests sit unmerged
-longer than that, a path whose last pass left an open PR would be re-surveyed on
-the anniversary and re-derive that PR's findings, so paths an open pull request
-already changes are dropped before the survey reads anything.
+longer than that, a path whose last pass left an open PR comes round again and
+its findings get re-derived on the anniversary. So each path an open pull
+request already changes is reported with the pulls covering it, for the survey
+to read before it reads the file. The path stays in the list: on the repositories
+where this fires the backlog can cover most of the tree, and dropping those paths
+would leave them unsurveyed for as long as it stands.
 """
 
 from __future__ import annotations
@@ -63,18 +66,17 @@ def main(
             "pr", "list", "--state", "open", "--limit", "200", "--json", "number,files"
         )
     covering = covering_pulls(pull_requests)
-    skipped = [path for path in selected if path in covering]
-    selected = [path for path in selected if path not in covering]
+    covered = [path for path in selected if path in covering]
     if selected:
         print(*selected, sep="\n")
     print(
         f"# bucket={bucket}/{CYCLE_LENGTH} files={len(selected)}"
-        f" skipped={len(skipped)}",
+        f" covered={len(covered)}",
         file=sys.stderr,
     )
-    for path in skipped:
+    for path in covered:
         pulls = ", ".join(f"#{number}" for number in covering[path])
-        print(f"# skipped {path} ({pulls})", file=sys.stderr)
+        print(f"# covered {path} ({pulls})", file=sys.stderr)
     return 0
 
 
--- a/plugins/tend-ci-runner/scripts/nightly_survey_files.py
+++ b/plugins/tend-ci-runner/scripts/nightly_survey_files.py
@@ -2,7 +2,14 @@
 # requires-python = ">=3.12"
 # dependencies = []
 # ///
-"""Select today's deterministic slice of tracked files for the nightly survey."""
+"""Select today's deterministic slice of tracked files for the nightly survey.
+
+Both the day and the path pick a bucket by arithmetic, so bucket *N* returns to
+the same file set every cycle. On a repository whose pull requests sit unmerged
+longer than that, a path whose last pass left an open PR would be re-surveyed on
+the anniversary and re-derive that PR's findings, so paths an open pull request
+already changes are dropped before the survey reads anything.
+"""
 
 from __future__ import annotations
 
@@ -10,6 +17,9 @@ import subprocess
 import sys
 import time
 import zlib
+from typing import Any
+
+import github_cli
 
 CYCLE_LENGTH = 28
 
@@ -23,7 +33,21 @@ def survey_files(files: list[str], *, unix_day: int) -> tuple[int, list[str]]:
     return bucket, selected
 
 
-def main(argv: list[str] | None = None, *, now: float | None = None) -> int:
+def covering_pulls(pull_requests: list[dict[str, Any]]) -> dict[str, list[int]]:
+    """Map each path an open pull request changes to the pulls changing it."""
+    covering: dict[str, list[int]] = {}
+    for pull in pull_requests:
+        for changed in pull["files"]:
+            covering.setdefault(changed["path"], []).append(pull["number"])
+    return covering
+
+
+def main(
+    argv: list[str] | None = None,
+    *,
+    now: float | None = None,
+    pull_requests: list[dict[str, Any]] | None = None,
+) -> int:
     args = sys.argv[1:] if argv is None else argv
     if args:
         print(f"usage: {sys.argv[0]}", file=sys.stderr)
@@ -34,9 +58,23 @@ def main(argv: list[str] | None = None, *, now: float | None = None) -> int:
     bucket, selected = survey_files(
         tracked, unix_day=int(time.time() if now is None else now) // 86400
     )
+    if pull_requests is None:
+        pull_requests = github_cli.json_call(
+            "pr", "list", "--state", "open", "--limit", "200", "--json", "number,files"
+        )
+    covering = covering_pulls(pull_requests)
+    skipped = [path for path in selected if path in covering]
+    selected = [path for path in selected if path not in covering]
     if selected:
         print(*selected, sep="\n")
-    print(f"# bucket={bucket}/{CYCLE_LENGTH} files={len(selected)}", file=sys.stderr)
+    print(
+        f"# bucket={bucket}/{CYCLE_LENGTH} files={len(selected)}"
+        f" skipped={len(skipped)}",
+        file=sys.stderr,
+    )
+    for path in skipped:
+        pulls = ", ".join(f"#{number}" for number in covering[path])
+        print(f"# skipped {path} ({pulls})", file=sys.stderr)
     return 0
 
 
@@ -44,4 +82,4 @@ if __name__ == "__main__":
     try:
         raise SystemExit(main())
     except subprocess.CalledProcessError as error:
-        raise SystemExit(error.returncode or 1) from None
+        raise SystemExit(github_cli.exit_code(error))
--- a/plugins/tend-ci-runner/scripts/nightly_survey_files.py
+++ b/plugins/tend-ci-runner/scripts/nightly_survey_files.py
@@ -8,10 +8,12 @@ Both the day and the path pick a bucket by arithmetic, so bucket *N* returns to
 the same file set every cycle. On a repository whose pull requests sit unmerged
 longer than that, a path whose last pass left an open PR comes round again and
 its findings get re-derived on the anniversary. So each path an open pull
-request already changes is reported with the pulls covering it, for the survey
-to read before it reads the file. The path stays in the list: on the repositories
-where this fires the backlog can cover most of the tree, and dropping those paths
-would leave them unsurveyed for as long as it stands.
+request already changes is reported with the pulls covering it and who wrote
+them, for the survey to read before it reads the file. The author decides where
+a new finding goes: onto the covering pull request where the bot wrote it, into
+a pull request of its own where anyone else did. The path stays in the list: on
+the repositories where this fires the backlog can cover most of the tree, and
+dropping those paths would leave them unsurveyed for as long as it stands.
 """
 
 from __future__ import annotations
@@ -36,12 +38,13 @@ def survey_files(files: list[str], *, unix_day: int) -> tuple[int, list[str]]:
     return bucket, selected
 
 
-def covering_pulls(pull_requests: list[dict[str, Any]]) -> dict[str, list[int]]:
+def covering_pulls(pull_requests: list[dict[str, Any]]) -> dict[str, list[str]]:
     """Map each path an open pull request changes to the pulls changing it."""
-    covering: dict[str, list[int]] = {}
+    covering: dict[str, list[str]] = {}
     for pull in pull_requests:
+        label = f"#{pull['number']} {pull['author']['login']}"
         for changed in pull["files"]:
-            covering.setdefault(changed["path"], []).append(pull["number"])
+            covering.setdefault(changed["path"], []).append(label)
     return covering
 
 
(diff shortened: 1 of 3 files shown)


plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -129,6 +129,8 @@ uv run --script \
   "${CLAUDE_PLUGIN_ROOT}/scripts/nightly_survey_files.py"
 ```
 
+The rotation is deterministic in both directions, so a bucket returns to the same files every cycle. The script therefore drops any path an open PR already changes, printing `# skipped <path> (#N)` on stderr for each: whatever the survey would find there is either already in that PR or belongs as a comment on it, and the PR is where a maintainer is already looking. Carry the skipped paths into the Step 9 summary.
```
tend-agent:
```suggestion
The script drops any path an open PR already changes — a file with a pending PR is where a maintainer is already looking — printing `# skipped <path> (#N)` on stderr for each. Carry the skipped paths into the Step 9 summary.
```

tend-agent (commented):
One point, on the skill text rather than the script.

`plugins/tend-ci-runner/skills/nightly/SKILL.md` loads into every nightly session, and CLAUDE.md's *Authoring skills* rule is "Be brief. … Lead with the rule or recipe; cut motivation, anecdotes, and historical context unless required to apply the rule." The new paragraph opens with the aliasing mechanism and then restates the justification from the PR body — neither of which the model acts on, since the script now does the drop itself. What Step 6 actually needs from the prose is that skips happen, that they surface on stderr, and that they carry into Step 9; the inline suggestion tightens it to that, keeping just enough rationale that a session doesn't re-add a skipped path by hand.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Applied in 28d1708 — the paragraph is now the suggested two sentences. Agreed on the routing: the rotation mechanism and the "already in that PR" justification explain why the script drops the path, which the session no longer decides, so they belong in the PR body rather than in text every nightly run loads. What stays is the rule (skips happen, they print on stderr, they carry into Step 9) plus the one clause that keeps a session from re-adding a skipped path by hand.

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -129,6 +129,8 @@ uv run --script \
   "${CLAUDE_PLUGIN_ROOT}/scripts/nightly_survey_files.py"
 ```
 
+The script prints `# covered <path> (#N)` on stderr for each path an open PR already changes. Read that PR's diff before reading the file: drop the findings it already carries, and raise anything new as a comment on that PR rather than a second PR against the same file. Carry the covered paths into the Step 9 summary.
```
tend-agent:
```suggestion
The script prints `# covered <path> (#N)` on stderr for each path an open PR already changes. Read that PR's diff before reading the file and drop the findings it already carries. Raise anything new as a comment on the covering PR only where the bot authored it — otherwise open a PR per Step 8 rather than commenting on someone else's work. Carry the covered paths into the Step 9 summary.
```

tend-agent (commented):
One finding on the incremental, from the design change in adb01ea.

`gh pr list --state open --limit 200 --json number,files` is unscoped by author, but Step 6's new instruction routes findings by coverage alone: "raise anything new as a comment on that PR rather than a second PR against the same file". So when a maintainer's or an outside contributor's open PR happens to touch a path in today's bucket, the survey's findings — unrelated to what that PR changes — land as a comment on their in-flight work. This repo has that shape right now: #1171 and #1172 are open and human-authored, and `generator/` paths they touch will come round in some bucket. `running-in-ci`'s Conduct section names exactly this ("Avoid causing disruption — unnecessary comments"), and the finding is also worse off there than in a PR of its own, buried in a feature thread nobody is tracking it in.

The repo already has the scoped form of this query, for the same "is this path already covered" question — `plugins/tend-ci-runner/skills/running-in-ci/references/skill-pr-workflow.md`, under **Mechanics** step 2, filters with `--author "$BOT_LOGIN"` and comments only where the bot is the author.

Two ways to close it, both one edit. The inline suggestion takes the prose route — keep the query unscoped, since knowing a human PR is mid-rewrite of a file is itself worth having in the Step 9 summary, and gate only the comment on authorship. The alternative is to scope the query (`"--author", "@me"` in the `json_call` argument list), which matches the mechanism's stated purpose — the bot's own unmerged backlog aliasing against the 28-day cycle — and leaves the skill sentence as written, at the cost of the pinned-invocation test and the informational half.

plugins/tend-ci-runner/skills/nightly/SKILL.md
tend-agent: Fixed in f80c33e, taking the prose route but making the gate mechanical rather than leaving it to a lookup. The query stays unscoped — a human PR mid-rewrite of a file is worth knowing before the survey reads it, which the scoped alternative would have thrown away — and `--json number,files,author` puts the author into the `# covered <path> (#N author)` line, so Step 6 can route on it without a second `gh` call. The sentence is close to your suggestion: comment on the covering PR only where the bot wrote it, otherwise open a PR per Step 8 rather than commenting on someone else's in-flight work.

The pinned-invocation test moves with the field, and the injected-PR test now covers a bot PR and a human PR on the same path so the label order and both logins are asserted. Full `uv run pytest` (1120) and every pre-commit hook pass.

plugins/tend-ci-runner/scripts/nightly_survey_files.py
```diff
@@ -23,7 +38,22 @@ def survey_files(files: list[str], *, unix_day: int) -> tuple[int, list[str]]:
     return bucket, selected
 
 
-def main(argv: list[str] | None = None, *, now: float | None = None) -> int:
+def covering_pulls(pull_requests: list[dict[str, Any]]) -> dict[str, list[str]]:
+    """Map each path an open pull request changes to the pulls changing it."""
+    covering: dict[str, list[str]] = {}
+    for pull in pull_requests:
+        label = f"#{pull['number']} {pull['author']['login']}"
```
tend-agent:
```suggestion
        label = f"#{pull['number']} {github_cli.actor_login(pull['author'])}"
```

tend-agent (commented):
One point on the incremental, on the new `author` field.

`covering_pulls` reads `pull['author']['login']` directly, and it is the only actor read in `plugins/tend-ci-runner/scripts/` that doesn't go through `github_cli.actor_login` — the helper whose docstring is "Return a GitHub actor login, including for deleted-account records", used at ten sites across `bot_review_state.py`, `review_runs.py`, and `review_runs_corrections.py`. `gh pr list --json author` emits every field of the object whether or not it is populated (a live PR here returns `"name": ""` alongside a real login), so a deleted account gives `login: ""` rather than a missing key: the label degrades to `#7 ` and Step 6's authorship gate reads a blank author. That fails to the safe side — a blank login isn't the bot's, so the finding opens its own PR — so the cost is the degraded line plus the outlier itself, which is the part worth closing: this is the line a later script will be copied from.

plugins/tend-ci-runner/scripts/nightly_survey_files.py
tend-agent: Applied in 6514597, with `pull.get("author")` rather than `pull["author"]` so the argument shape matches the other ten call sites and the helper's missing-record tolerance covers an absent key too. Behaviour is unchanged for every populated record; a deleted account now degrades to the blank label instead of raising, and the authorship gate keeps failing to the safe side.

max-sixty:
@tend-agent This conflicts with the root `CLAUDE.md` rule "Simplicity outranks efficiency". That rule accepts complexity only when it prevents a wrong outward action, never when it only saves compute, and it names "a duplicated survey" as waste to leave alone unless the fix is a simple knob: a cadence value, a deleted step, or a one-line condition.

What #1176 and this PR's body describe is compute. The check before `gh pr create` already caught the duplicate, and nothing wrong shipped. This fix is not a knob: it adds a `gh pr list` call that can now fail the nightly step, a `# covered` line format on stderr, a new rule for posting findings onto covering PRs, and about 100 lines of tests to maintain.

Please close this PR and reply on #1176 that the rotation's duplicate cost stays as it is under that rule. If a genuine one-line knob removes the duplicated cycle, that would fit instead.

tend-agent:
Closing per your call. The duplicated survey cycle is compute, the pre-`gh pr create` recheck already caught it, and this change was a mechanism rather than a knob — so it fails the `CLAUDE.md` rule on both counts.

Recorded the outcome on #1176, including why the one-line shape available here (dropping covered paths in the script) isn't the knob it looks like. Leaving the branch in place.

PR #1179, opened 2026-09-08
Add Codex subscription authentication

max-sixty:
Add first-class Codex subscription authentication for Tend while keeping the raw action model-neutral. Generated Codex workflows pin the current Sol-tier model and weekly guidance advances that pin only within the same capability and price tier.

The install skill now drives an isolated Codex device login, stores access-only and refreshable bundles in the gated `tend` environment, and opens a pre-scoped fine-grained PAT form before securely accepting the token from stdin. A serialized weekly workflow rotates the refresh chain. This repo trials the Codex harness on `review-runs`.

The helper and generator behavior are covered by integration-style Python tests; the full Python and worker suites, typecheck, and all pre-commit hooks pass locally.


plugins/install-tend/skills/install-tend/scripts/install_codex_subscription_auth.py
```diff
@@ -0,0 +1,216 @@
...
+    if bundle.get("auth_mode") != "chatgpt":
+        raise ProvisionError("Codex auth.json must have auth_mode 'chatgpt'")
+    tokens = _tokens(bundle, "Codex auth.json")
+    for field in ("access_token", "refresh_token", "id_token", "account_id"):
+        value = tokens.get(field)
+        if not isinstance(value, str) or not value:
+            raise ProvisionError(f"Codex auth.json is missing tokens.{field}")
+
+
+def consumer_auth(full_bundle: Mapping[str, Any]) -> dict[str, Any]:
```
tend-agent, line 54: This reproduces the projection `shared/steps/codex_subscription_auth.py` already owns — a `consumer_auth` of the same name, same signature, same three assignments over a deep copy — minus the `_validate_consumer(consumer)` round-trip the shared copy ends on. So nothing checks that the bundle the installer stores is one the action's `prepare` will accept, and nothing in the repo ties the two definitions together. If the consumer contract changes on the action side, the installer keeps writing the old shape and the adopter learns about it when the first scheduled Codex job fails at `Configure Codex auth`, not during install.

`validate_full`, `_tokens`, `CONSUMER_AUTH_MODE`, and `_set_secret` are copies of the same module's `_validate_full`, `_tokens`, `CONSUMER_AUTH_MODE`, and `_set_environment_secret` too.

The marketplace ships this plugin as `./plugins/install-tend`, so importing the shared module at runtime isn't on the table — but a contract assertion is: the root pytest run collects both trees, so `test_install_codex_subscription_auth.py` can push `consumer_auth(FULL_AUTH)` through the shared `_validate_consumer` and fail when the two shapes diverge. Happy to push that test if you want it.

plugins/tend-ci-runner/skills/nightly/SKILL.md
```diff
@@ -165,6 +165,17 @@ Otherwise its JSON supplies the title, old and new versions, compare URL, and
 upstream commit subjects. It leaves the prepared worktree and state on disk for
 the shipping command; do not `cd` into it.
 
+Before composing the PR body, inspect the upstream changes for a new harness
+model default. Update each explicit model pin in the prepared `.config/tend.yaml`
+only when the replacement is a newer model in the same capability and price
+tier, verified from the provider's model and pricing docs; preserve cross-tier
+pins as product choices. After an edit, rerun generation without moving this
+session's cwd:
```
tend-agent: "Verified from the provider's model and pricing docs" states the bar but not what to do when it can't be met, and the fallback an agent reaches for when a docs fetch fails or the tier is ambiguous is exactly the name-similarity guess the sentence rules out — on a bundled skill, in every adopter's nightly, proposing a change to the model their bot runs on. Naming the source and the default keeps it from turning into a guess:

```suggestion
Before composing the PR body, read `compare_url`'s diff for a change to the
harness model default. Update each explicit model pin in the prepared
`.config/tend.yaml` only when the replacement is a newer model in the same
capability and price tier, confirmed from the provider's own model and pricing
docs; leave the pin unchanged when those docs are unreachable or the tier is
unclear, and preserve cross-tier pins as product choices. After an edit, rerun
generation without moving this session's cwd:
```

tend-agent (commented):
Two points, both about install-time and run-time drifting apart; details inline.

plugins/install-tend/skills/install-tend/scripts/install_codex_subscription_auth.py
max-sixty: Added an integration test that sends the installer-produced consumer bundle through the action-side `prepare` command and verifies the staged bundle. This now fails during the repository test run if either side's contract changes incompatibly.

plugins/tend-ci-runner/skills/nightly/SKILL.md
max-sixty: Updated the nightly guidance to inspect `compare_url`, retain the current pin when provider docs are unreachable or the tier is unclear, and preserve cross-tier pins as product choices.

tend-agent: Confirmed — the landed text names both the source and the default, so the ambiguous case now has a defined outcome instead of falling through to a name-similarity guess. Nothing further from me on this one.

plugins/install-tend/skills/install-tend/scripts/install_codex_subscription_auth.py
tend-agent: Confirmed, and the test is stronger than what I offered — driving the installer's bundle through the action's real `prepare` covers the whole acceptance path rather than just `_validate_consumer`.

I checked it actually binds the contract rather than merely passing: adding a `workspace_id` requirement to the action-side `_validate_consumer` makes `test_consumer_auth_is_accepted_by_action` fail with `CODEX_AUTH_JSON is missing tokens.workspace_id`, and the suite is green with that mutation reverted (14 passed). So an action-side tightening now surfaces in the repo test run instead of at an adopter's first scheduled Codex job.

One residual, not a blocker: `validate_full` is still an unbound copy of the shared `_validate_full`. Drift there wouldn't fail install or this test — it would surface when the weekly refresh workflow calls `stage_refresh` on a bundle the installer considered valid. Lower stakes than the consumer path, and reasonable to leave if you'd rather not grow the test surface.

tend-agent (approved):

PR #1180, opened 2026-09-08
fix(codex): parse cumulative rollout token usage

max-sixty:
The live Codex subscription test completed successfully, but its token-usage summary reported zeros because Codex now records cumulative counts inside nested `event_msg` events.

Parse each rollout through its final readable cumulative token event, sum those independent rollout totals, count assistant response messages as turns, and recover the model from turn context when the action did not supply one. The parser retains the last complete count if a rollout has a truncated tail. The Codex summary now labels cached input as a subset of total input.

The updated parser reports 2,570,502 input tokens, 2,464,384 cached input tokens, 7,558 output tokens, and 7 turns against the captured subscription-backed rollout. The full Python and worker test suites, typecheck, pre-commit hooks, and diff check pass.

### DIFF
--- a/shared/steps/test_token_usage.py
+++ b/shared/steps/test_token_usage.py
@@ -75,6 +75,23 @@ def _assistant(msg_id: str, usage: dict[str, int], *, final: bool) -> dict[str, 
     }
 
 
+def _codex_token_count(**usage: int) -> dict[str, object]:
+    return {
+        "type": "event_msg",
+        "payload": {
+            "type": "token_count",
+            "info": {"total_token_usage": usage},
+        },
+    }
+
+
+def _codex_message() -> dict[str, object]:
+    return {
+        "type": "response_item",
+        "payload": {"type": "message", "role": "assistant"},
+    }
+
+
 def _ndjson(path: Path, lines: list[dict[str, object]]) -> Path:
     path.parent.mkdir(parents=True, exist_ok=True)
     path.write_text("".join(json.dumps(line) + "\n" for line in lines))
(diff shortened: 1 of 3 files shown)


shared/steps/test_token_usage.py
```diff
@@ -377,18 +389,19 @@ def test_codex_counts_an_absent_token_count_as_zero() -> None:
     """
     usage = token_usage.codex_usage(
         [
-            {"type": "agent_message"},
-            {"type": "event_msg", "token_count": None},
-            {"type": "turn_context", "token_count": {"input_tokens": None}},
-            {"type": "token_count", "token_count": {"input_tokens": 42}},
+            [
+                {"type": "response_item", "payload": {"type": "reasoning"}},
+                {"type": "event_msg", "payload": {"type": "token_count"}},
```
tend-agent, line 394: The docstring promises a null field *inside* the usage dict; this restores it, and it's the only place the Codex path exercises that branch of `number()`.

```suggestion
                {"type": "event_msg", "payload": {"type": "token_count"}},
                {
                    "type": "event_msg",
                    "payload": {
                        "type": "token_count",
                        "info": {"total_token_usage": {"input_tokens": None}},
                    },
                },
```

tend-agent (commented):
The parser change reads correctly to me — last cumulative count per rollout, summed across rollouts that are genuinely independent sessions, with a truncated tail costing only its own line. Three things left over.

`codex/action.yaml`'s "Token usage" step still documents the premise this PR removes. The comment block above `- name: Token usage` (immediately after the "Codex writes rollouts under the sandbox user's `~/.codex/sessions`" lines) reads `Each event includes a token_count field (input/output/cached). Sum them.` — that is exactly the guess that produced the zeros, and it now describes neither the schema nor the parser. It ships to adopters in the pinned action and it is the copy a reader hits before `token_usage.py`, so it is worth moving with the module docstring. Say the word and I'll push the correction rather than have you respin.

The docstring on `test_codex_counts_an_absent_token_count_as_zero` still claims a case the rewritten body dropped: "An event without `token_count` — or with a null field inside it — must contribute 0". The new events cover the absent-`info` shape but nothing exercises a null *inside* a usage dict, and that branch of `number()` has no coverage anywhere else in the file — a grep for `None` across the tests turns up only `stop_reason` and unrelated fixtures. Since `number()`'s own docstring promises "absent, null, or the wrong type", restoring the case looks better than narrowing the claim; suggestion inline (verified: `ruff format --check` clean, 20 passed).

Last one is a question rather than a finding. If Codex's `total_token_usage.input_tokens` is inclusive of `cached_input_tokens`, then `CODEX_TABLE`'s `Input` and `Cached input` rows are not additive — while `CLAUDE_TABLE`'s `Input` excludes both cache rows, because Anthropic reports it that way. Same column heading, two meanings, and this PR is what turns those cells from zero into numbers people will read. The description's own figures point at inclusive: 2,464,384 cached against 2,570,502 input over 7 turns is a 96% cache rate, whereas the exclusive reading puts 367k of freshly-uncached prompt in every turn. You have the captured rollout — if its `total_tokens` lands near input + output rather than input + cached + output, the rows want relabeling (or the render wants `input - cached`) before the first adopter compares a Codex summary against a Claude one.

shared/steps/test_token_usage.py
max-sixty: Added the null-valued `input_tokens` case to preserve coverage of `number()`'s promised behavior.

max-sixty:
Addressed all three points in `dd4b4b4`: the action comment now describes final cumulative totals, the null-field test is restored, and the Codex summary labels cached input as included in total input. The captured rollout confirms that `total_tokens == input_tokens + output_tokens`, so subtracting cached tokens would change the stored OpenAI total rather than clarify it.

tend-agent (approved):

PR #1181, opened 2026-09-08
chore: release 0.2.1

max-sixty:
Releases Tend 0.2.1 with disposable Sandbox Runtime workspaces, end-to-end Codex subscription setup, release-integrity checks, and fixes for review-run coverage, outage diagnostics, and Codex token accounting. It also corrects the bundled CI restrictions to describe both harnesses' current credential and privilege boundary. The new changelog section is published verbatim as the GitHub Release notes.


CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ published verbatim as that version's GitHub Release notes
...
+- **Codex subscription authentication has an end-to-end install flow.** An isolated device login writes separate refreshable and access-only bundles to the gated environment without printing them, and a pre-scoped fine-grained PAT form supplies the weekly refresh workflow. Generated Codex installs pin `gpt-5.6-sol`; a per-workflow harness change now selects that harness's default model unless the workflow names one. ([#1179](https://github.com/max-sixty/tend/pull/1179))
+- **`tend check` verifies release integrity.** It requires an admin-gated all-tags ruleset and GitHub's immutable-release setting, and `tend check --fix` configures both. Codex agents now run as the sandbox user with the bot PAT outside their environment; API-key model requests use a narrow Responses API proxy, while subscription sessions receive only the expiring access token. ([#1160](https://github.com/max-sixty/tend/pull/1160))
```
tend-agent:
```suggestion
- **`tend check` verifies release integrity.** It requires an admin-gated all-tags ruleset and GitHub's immutable-release setting, and `tend check --fix` configures both. ([#1160](https://github.com/max-sixty/tend/pull/1160))
- **Codex agents run as the sandbox user, with the bot PAT outside their environment.** API-key model requests use a narrow Responses API proxy, while subscription sessions receive only the expiring access token. ([#1160](https://github.com/max-sixty/tend/pull/1160))
```

tend-agent (commented):
The version bump, lockfile, and section header all line up with what `pypi-release.yaml` extracts, and each bullet's claim matches the code it describes. Two things on the notes themselves.

The bullet leads are the index a reader scans on the Release page, and the largest adopter-facing security change in this release — Codex agents moving behind the credential-isolation sandbox — is a trailing clause under a `tend check` lead. #1160 is already split by topic across two bullets here, so a third for the credential change fits the pattern; suggestion inline.

Separately, `plugins/tend-ci-runner/skills/running-in-ci/SKILL.md` still tells agents the opposite of what these notes publish. Its **Restrictions** bullet on secrets reads "Under `codex` the agent runs as the runner user with the real PAT and model key in its own environment," while `codex/action.yaml` now runs `Set up credential-isolation sandbox` unconditionally and launches the agent through `launch_sandbox_runtime.py`. That line ships to every adopter session in 0.2.1. It's outside this diff and doesn't belong in the release commit — happy to open a follow-up PR correcting it once this lands.

CHANGELOG.md
max-sixty: Split release integrity and credential isolation into separate top-level bullets in `a88564e8`. The same commit updates the bundled restrictions to describe both harnesses' current sandbox boundary.

tend-agent (approved):

CHANGELOG.md
tend-agent: Both points check out at `a88564e85547c5a8379789dc8e346aedc425a88b`. The split gives credential isolation its own lead, and the restrictions bullet now matches what the actions do: `codex/action.yaml` runs `Set up credential-isolation sandbox` unconditionally and launches through `sudo -u "$SANDBOX" env`, the Responses proxy holds the API key behind a loopback endpoint, and subscription mode installs only the expiring `auth.json` into the sandbox home.

Worth noting the one deletion that isn't a rewording: the old text warned that `actions/checkout`'s persisted credential stays live in the runner workspace's git config. No checkout in the repo sets `persist-credentials: false`, so it is still there — but it is no longer reachable from the agent, since the runner checkout is unreadable to the sandbox for the whole running state and launch probes that before `sandbox_setup:` executes. Dropping the clause is right, and the general "never print a credential from anywhere else" covers the residue. `docs/security-model.md` and `CLAUDE.md` carry no leftover copy of the old Codex-runs-as-runner claim, so the skill was the last one.

No follow-up PR needed from me then.

PR #1182, opened 2026-09-08
chore: release 0.2.2

max-sixty:
Releases Tend 0.2.2 with the tested mention-workspace fix from [#1175](https://github.com/max-sixty/tend/pull/1175). Issue-body mentions now stay on the current default branch, and closed-PR mentions no longer pin instructions to an old base commit. The changelog section is published verbatim as the GitHub Release notes.


tend-agent (commented):
`checkout_mode: mention` moves this repo's own mention workflow onto the 0.2.1 workspace path, where a mention added by **editing an issue body** loses the entire run. `tend-mention` subscribes to `issues: [edited]`, and at `0.2.1` the mention branch in `prepare_agent_workspace.py`'s `main()` is gated on `GITHUB_EVENT_NAME == "issue_comment"`, so an `issues` payload falls through to `checkout_mention`, which requests `/repos/{repo}/pulls/{issue-number}` and 404s. The `HTTPError` reaches `main()`'s handler and `Prepare disposable agent workspace` exits non-zero with no `continue-on-error` above it. It is silent from the thread's side as well: an `issues` event has no `github.event.comment`, so neither the eyes reaction nor its removal fires and the issue shows no sign the bot was ever summoned.

The `Check out PR branch` step this PR deletes is what had been keeping that path alive — its condition (`github.event.issue.pull_request.url != '' || repository_dispatch`) excluded `issues` outright, so the run simply continued on the default branch. #1175 replaces the gate with a payload-shape check and is the fix, but it isn't released, so the call is whether to accept the gap (nightly's `uvx tend@latest init` regenerates to `0.2.1` regardless, so it is roughly a day either way) or land #1175 and cut `0.2.2` before deploying the new topology here.

tend-agent (approved):

PR #1188, opened 2026-09-09
fix(skills): make Tend guidance portable across harnesses

max-sixty:
Codex CI installed only `tend-ci-runner`, even though runner skills load the diagnostic skill from `install-tend`, and the Codex log recipes still described the previous rollout schema. Install both plugins in the Codex harness and document current `custom_tool_call`, `custom_tool_call_output`, and `event_msg` / `AgentMessage` records.

Installation now links `CLAUDE.md` and `AGENTS.md` when only one exists, so Claude and Codex consume one project-guidance source. Generic Tend skills use harness-neutral tool language and delegate token-heavy exploration to a smaller, cheaper model without naming provider-specific models.

Cross-file contracts pin plugin availability, the Codex log schema, instruction-file aliases, and neutral vocabulary. The full Python and worker suites, TypeScript typecheck, pre-commit hooks, a real Codex 0.153.4 plugin-install probe, the jq recipes, and both symlink directions pass.


plugins/install-tend/skills/debug-tend-run/references/codex-logs.md
```diff
@@ -13,33 +13,36 @@ Each JSONL line has a top-level `type` of `session_meta`, `turn_context`,
   `.payload.content[] | select(.type == "input_text") | .text`
 - `agent_message` — text emitted by the model during the turn
   (`.payload.message`)
```
tend-agent: `agent_message` doesn't appear in current rollouts at all, so this bullet still documents the old schema. The model's visible text is an `event_msg` entry rather than a `response_item` variant, so it doesn't belong in this list — the next suggestion re-homes it in the lifecycle paragraph.

```suggestion
```

plugins/install-tend/skills/debug-tend-run/references/codex-logs.md
```diff
@@ -13,33 +13,36 @@ Each JSONL line has a top-level `type` of `session_meta`, `turn_context`,
...
-  `task_complete.last_agent_message` carries the final reply
 
-The bot drives shell through one tool, `exec_command`, whose arguments
-parse to `{cmd, workdir, yield_time_ms, max_output_tokens}`. Long-running
-commands also produce `write_stdin` calls. Codex has no dedicated
-Read/Write/Edit tool; file I/O appears as `cat`, `sed`, `tee`,
-`apply_patch`, etc. inside `exec_command`.
+Lifecycle metadata such as `task_started`, `task_complete`, and `token_count`
+sits under `event_msg.payload`. `task_complete.last_agent_message` carries the
+final reply.
```
tend-agent: Re-homes the model narrative where it actually lives.

```suggestion
Lifecycle metadata such as `task_started`, `task_complete`, and `token_count`
sits under `event_msg.payload`, as does the model's visible narrative —
`item_completed` entries whose `.payload.item.type` is `AgentMessage`, with the
text at `.payload.item.content[].text`. `task_complete.last_agent_message`
carries the final reply.
```

plugins/install-tend/skills/debug-tend-run/references/codex-logs.md
```diff
@@ -13,33 +13,36 @@ Each JSONL line has a top-level `type` of `session_meta`, `turn_context`,
...
 jq -r 'select(.payload.type == "task_complete") | .payload.last_agent_message' "$FILE"
 
 # Tool calls in order
-jq -r 'select(.payload.type == "function_call") |
-  "\(.payload.name): \(.payload.arguments | fromjson | (.cmd // tostring) | .[0:160])"' "$FILE"
+jq -r 'select(.payload.type == "custom_tool_call") |
+  "\(.payload.name): \(.payload.input | .[0:160])"' "$FILE"
 
 # Interim model narrative (its visible reasoning; the encrypted blob is opaque)
 jq -r 'select(.payload.type == "agent_message") | .payload.message' "$FILE"
```
tend-agent: Returns 15 and 14 narrative lines on the two artifacts I checked, against 0 for the current form.

```suggestion
# Interim model narrative (its visible commentary; the reasoning blob is opaque)
jq -r 'select(.payload.type == "item_completed" and .payload.item.type == "AgentMessage") |
  .payload.item.content[]?.text' "$FILE"
```

tend-agent (commented):
`agent_message` is not a payload type in current Codex rollouts, so the two recipes that select on it return nothing — the same schema drift this PR fixes for tool calls, left in place for the model's narrative. Neither unexpired `codex-session-logs-*` artifact on this repo (runs 34283240365 and 34276722915) contains the string anywhere; the visible narrative is an `event_msg` line whose `item_completed` payload has `.payload.item.type == "AgentMessage"`, with text at `.payload.item.content[].text`.

The inline suggestions cover the bullet, the lifecycle paragraph, and the overview recipe. The same fix applies to the keyword recipe under "Searching for specific behavior", which falls outside a diff hunk:

```bash
# Model text mentioning a keyword
jq -r 'select(.payload.type == "item_completed" and .payload.item.type == "AgentMessage") |
  .payload.item.content[]?.text | select(test("KEYWORD"; "i"))' "$FILE"
```

Happy to push all four as a commit if that's easier than applying them.

tend-agent (approved):

PR #1195, opened 2026-09-09
fix(running-tend): make the restamp check see nested action paths

tend-agent:
The nightly restamp check in the `running-tend` overlay told every session to restamp workflows that were already correct. Its rule — "one line means every workflow agrees" — was written when one action path was pinned; three are pinned now (`claude`, `codex`, `codex/refresh`), so on today's `main` a repo in full agreement at `0.2.5` prints two lines and the recipe reads that as drift. The check is on the versions, so the section says so and gives the `sed 's/.*@//' | sort -u` pipe. Harness drift — a hand-maintained file naming a harness the config no longer selects — is called out as a separate question the listing cannot answer, because `.config/tend.yaml` overrides `review-runs` to `codex` and a listing naming two harnesses is this repo's normal state.

The listing regex was blind to nested action paths in the other direction: `max-sixty/tend/[a-z-]+@[0-9.]+` has no `/` in its character class, so `max-sixty/tend/codex/refresh@X.Y.Z` never appeared in the output at all. That one is latent rather than live — `codex/refresh` is pinned only in the generated `tend-codex-auth-refresh.yaml`, which `init` restamps with every other generated file — but it would hide a stale pin the day a hand-maintained workflow names a nested path. The class now admits `/`, and the listing prints all three pinned paths.

`test_every_workflow_pins_the_same_tend_release` asserts the property the restamp step exists to detect: every `max-sixty/tend/...` ref committed under `.github/workflows/` carries one version. A pin left behind at an older release now fails CI on every run rather than waiting for a nightly session to run the prose recipe, and the ref pattern accepts any pin, so a hand-edited `@main` is caught too. It globs `*.y*ml` to match the recipe's directory scan, since Actions loads `.yml` as readily as `.yaml`. Verified both failure modes locally by re-pinning `review-reviewers.yaml` to `0.2.4` and by dropping a `.yml` at an older pin into the directory: `assert 2 == 1 … workflows pin more than one tend release`.

The invariant holds on `main` continuously because the release skill already requires the hand-maintained restamp in the same commit as the regeneration (step 11). The harness half of the recipe is prose with no code path to pin, so it stays uncovered.

Found while running the restamp step of a nightly sweep.

### DIFF
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -114,17 +114,16 @@ Run this after the regen step, whether or not it produced a PR:
 rg -o --no-filename 'max-sixty/tend/[a-z/-]+@[0-9.]+' .github/workflows/ | sort -u
 ```
 
-The character class must admit the `/` in a nested action path, or a stale
-`codex/refresh` pin never appears in the output at all.
-
 Several action paths are pinned at once (`claude`, `codex`, `codex/refresh`),
-so several lines is the normal case. Compare the versions instead — pipe the
-same output through `sed 's/.*@//' | sort -u` and expect exactly one. Two or
-more, restamp the hand-maintained files onto the generated files' ref and fold
-it into the regen PR — same worktree, same commit. A hand-maintained file
-naming a different *harness* from the generated ones is the worse case: a
-config change reached the generated workflows and stopped there, so check what
-else that change was supposed to carry.
+so several lines is the normal case — as are two harnesses, while
+`.config/tend.yaml` overrides one workflow to `codex`. The check is on the
+versions: pipe the same output through `sed 's/.*@//' | sort -u` and expect
+exactly one. Two or more, restamp the hand-maintained files onto the generated
+files' ref and fold it into the regen PR — same worktree, same commit. Whether
+a hand-maintained file names the right *harness* is a separate question the
+listing cannot answer; compare it against the harness `.config/tend.yaml`
+configures, and on a mismatch check what else that config change was supposed
+to carry.
 
 ## Weekly: refresh `data/consumers.json`
 
--- a/.claude/skills/running-tend/SKILL.md
+++ b/.claude/skills/running-tend/SKILL.md
@@ -111,14 +111,20 @@ regenerating skips them entirely.
 Run this after the regen step, whether or not it produced a PR:
 
 ```bash
-rg -o --no-filename 'max-sixty/tend/[a-z-]+@[0-9.]+' .github/workflows/ | sort -u
+rg -o --no-filename 'max-sixty/tend/[a-z/-]+@[0-9.]+' .github/workflows/ | sort -u
 ```
 
-One line means every workflow agrees. Two or more, restamp the hand-maintained
-files onto the generated files' ref and fold it into the regen PR — same
-worktree, same commit. A differing *harness* rather than a differing version is
-the worse case: a config change reached the generated workflows and stopped
-there, so check what else that change was supposed to carry.
+The character class must admit the `/` in a nested action path, or a stale
+`codex/refresh` pin never appears in the output at all.
+
+Several action paths are pinned at once (`claude`, `codex`, `codex/refresh`),
+so several lines is the normal case. Compare the versions instead — pipe the
+same output through `sed 's/.*@//' | sort -u` and expect exactly one. Two or
+more, restamp the hand-maintained files onto the generated files' ref and fold
+it into the regen PR — same worktree, same commit. A hand-maintained file
+naming a different *harness* from the generated ones is the worse case: a
+config change reached the generated workflows and stopped there, so check what
+else that change was supposed to carry.
 
 ## Weekly: refresh `data/consumers.json`
 
--- a/generator/tests/test_cross_file_contracts.py
+++ b/generator/tests/test_cross_file_contracts.py
@@ -458,28 +458,22 @@ def test_nightly_regen_stages_every_path_init_writes(
     )
 
 
-def test_restamp_check_sees_every_action_ref_the_workflows_pin() -> None:
-    """The overlay's drift check must match every `max-sixty/tend/...` pin.
+def test_every_workflow_pins_the_same_tend_release() -> None:
+    """`init` rewrites only the generated `tend-*.yaml` files.
 
-    Its regex is what decides whether a hand-maintained workflow's ref is
-    reported as stale, so a nested action path (`codex/refresh`) that the
-    character class cannot spell is a pin the nightly never restamps.
+    Every other workflow keeps whatever `max-sixty/tend/...` ref it was last
+    given by hand, so each release leaves it a version behind until someone
+    restamps it. The nightly sweep does the restamping; this is what decides
+    whether it is needed.
     """
-    skill = _read(".claude", "skills", "running-tend", "SKILL.md")
-    block = next(
-        block
-        for block in _bash_blocks(skill)
-        if "max-sixty/tend/" in block and ".github/workflows/" in block
-    )
-    match = re.search(r"'(max-sixty/tend/[^']+)'", block)
-    assert match
-    pattern = match.group(1)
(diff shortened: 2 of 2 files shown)


.claude/skills/running-tend/SKILL.md
```diff
@@ -111,14 +111,20 @@ regenerating skips them entirely.
...
+`codex/refresh` pin never appears in the output at all.
+
+Several action paths are pinned at once (`claude`, `codex`, `codex/refresh`),
+so several lines is the normal case. Compare the versions instead — pipe the
+same output through `sed 's/.*@//' | sort -u` and expect exactly one. Two or
+more, restamp the hand-maintained files onto the generated files' ref and fold
+it into the regen PR — same worktree, same commit. A hand-maintained file
+naming a different *harness* from the generated ones is the worse case: a
+config change reached the generated workflows and stopped there, so check what
+else that change was supposed to carry.
```
tend-agent: The check can no longer surface the harness case this paragraph calls the worse one: `.config/tend.yaml` overrides `review-runs` to `codex`, so a listing naming two harnesses is the normal state, and the version pipe reports agreement whichever harness each file names. Give the harness half something to compare against, and drop the sentence explaining the character class — the new test enforces it, and CLAUDE.md's **Authoring skills** asks for motivation to be cut from text loaded into every session.

```suggestion
Several action paths are pinned at once (`claude`, `codex`, `codex/refresh`),
so several lines is the normal case — as are two harnesses, while
`.config/tend.yaml` overrides one workflow to `codex`. The check is on the
versions: pipe the same output through `sed 's/.*@//' | sort -u` and expect
exactly one. Two or more, restamp the hand-maintained files onto the generated
files' ref and fold it into the regen PR — same worktree, same commit. Whether
a hand-maintained file names the right *harness* is a separate question the
listing cannot answer; compare it against the harness `.config/tend.yaml`
configures, and on a mismatch check what else that config change was supposed
to carry.
```

tend-agent (commented):
Two points, both on the shape of the fix rather than the regex itself — the character-class change is right and the ref inventory confirms `codex/refresh` was the ref it was missing.

`test_restamp_check_sees_every_action_ref_the_workflows_pin` pins the recipe's regex text rather than the property the recipe exists to detect. It already builds the full `refs` set, so `assert len({ref.split("@")[1] for ref in refs}) == 1` would go red on the drift itself — a `codex/refresh` pin left behind at an older version — instead of on one of the several ways the prose check can under-report it, and it would catch that drift on every CI run rather than when a nightly session happens to run the recipe. It also drops machinery: the markdown-block extraction, the single-quoted-string search, `re.fullmatch` standing in for `rg`'s search semantics, and a second ref regex that has to stay in sync with the first while sharing its blind spots (both require a numeric version, so a hand-edited `@main` is invisible to either). `test_review_reviewers_matrix_covers_consumers` in the same file is that shape already. The sweep step still does the restamping; CI would just decide whether it's needed. Worth settling before this lands, since the regex test is what becomes redundant.

Separately, the new pass/fail rule is version-only, which leaves the paragraph's stated "worse case" with no procedure behind it. `.config/tend.yaml` overrides `review-runs` to `codex`, so a listing naming two harnesses is normal here, and `sed 's/.*@//' | sort -u` collapses a harness mismatch to agreement. Wording inline.

.claude/skills/running-tend/SKILL.md
tend-agent: Applied as suggested. The character-class sentence is gone, and the harness question now points at `.config/tend.yaml` as the thing to compare against rather than at the listing, which can't answer it.

tend-agent:
Took the test-shape point — `test_restamp_check_sees_every_action_ref_the_workflows_pin` is now `test_every_workflow_pins_the_same_tend_release`, asserting one version across every `max-sixty/tend/...` ref committed under `.github/workflows/`. That drops the markdown-block extraction, the single-quoted-string search, and the second ref regex, and it turns the drift itself red on every CI run instead of on one of the ways the prose check can under-report it. Taking the blind spot with it: the surviving pattern accepts any pin (`@[^\s"']+`), so a hand-edited `@main` fails too, where both numeric-only regexes were blind to it.

Checked the one thing that assertion needs to be safe as a gate: it must hold on `main` continuously, not just after a nightly sweep. The release skill's step 11 already regenerates and follows **Nightly: restamp the hand-maintained workflow refs** in the same commit, so a release never leaves the hand-maintained files behind across a merge. Verified the failure mode by re-pinning `review-reviewers.yaml` to `0.2.4` locally — `assert 2 == 1 … workflows pin more than one tend release: [...codex/refresh@0.2.5]` — and reverting.

PR description recomposed around the current diff; it still claimed `0.2.4` from before the 0.2.5 regeneration landed on `main`.

generator/tests/test_cross_file_contracts.py
```diff
@@ -456,3 +456,24 @@ def test_nightly_regen_stages_every_path_init_writes(
...
+    """`init` rewrites only the generated `tend-*.yaml` files.
+
+    Every other workflow keeps whatever `max-sixty/tend/...` ref it was last
+    given by hand, so each release leaves it a version behind until someone
+    restamps it. The nightly sweep does the restamping; this is what decides
+    whether it is needed.
+    """
+    refs = {
+        ref
+        for path in (REPO_ROOT / ".github" / "workflows").glob("*.yaml")
```
tend-agent: This is the one place the test is narrower than the recipe it backs: the `rg` listing scans the directory, and Actions loads `.yml` as readily as `.yaml`. A hand-maintained `foo.yml` left at an older pin passes CI here and only shows up when a nightly session happens to run the prose check — the relationship this test exists to invert.

```suggestion
        for path in (REPO_ROOT / ".github" / "workflows").glob("*.y*ml")
```

tend-agent (commented):
The `/` in the character class and the version-only rule both check out — I re-ran the old and new listings against the committed workflows and confirmed the before/after. Two things on the current commit.

**The description's lead attributes the fix to a state that can't occur.** `codex/refresh` appears in exactly one workflow, `tend-codex-auth-refresh.yaml`, and that file carries the `Generated by tend X.Y.Z` header: `init` stamps it with `ACTION_VERSION` alongside every other generated file, and the cleanup loop in `cli.py` (`for path in sorted(outdir.glob("tend-*.yaml"))` … `path.unlink()`, under the comment "Remove stale tend-*.yaml files the generator didn't produce this run") deletes it once the config stops selecting codex. Either way it moves with the other generated pins, so "a stale `codex/refresh` pin read as agreement" names nothing reachable — the nested-path blind spot is real but latent, waiting on a *hand-maintained* workflow that pins a nested path, of which there are none. What was live is the defect paragraph 2 describes: on today's `main` the old regex prints `max-sixty/tend/claude@0.2.5` and `max-sixty/tend/codex@0.2.5`, two lines from a repo in full agreement, so "one line means every workflow agrees" sent every nightly to restamp something already correct. Leading with the false positive would tell a maintainer what this actually fixes.

Inline on the test's glob.

generator/tests/test_cross_file_contracts.py
tend-agent: Applied as suggested — the glob is `*.y*ml`. Confirmed it inverts the relationship the way you describe: dropping a `zz-probe.yml` pinning `max-sixty/tend/claude@0.2.4` into `.github/workflows/` turns the test red (`assert 2 == 1`), where the `*.yaml` glob passed it.

tend-agent:
Recomposed the description around the false positive. You're right that the old lead named an unreachable state: `codex/refresh` is pinned only in `tend-codex-auth-refresh.yaml`, which `init` stamps with `ACTION_VERSION` like every other generated file, so it can't fall behind the pins the listing does see. The nested-path blind spot is now stated as latent — it hides a stale pin only once a hand-maintained workflow names a nested path — and the lead is the defect that was live: three pinned paths against a rule written for one, so a repo in full agreement at `0.2.5` printed two lines and every nightly was sent to restamp correct files.

PR #1199, opened 2026-09-09
fix(sandbox): provide writable agent temp directory

### LINKED ISSUE #1192: skills: bundled scripts and recipes default to /tmp, which the SRT sandbox mounts read-only
Since [#1173](https://github.com/max-sixty/tend/pull/1173) the agent runs inside SRT with an allowlisted write set — the agent workspace, the agent home, `TEND_STEP_SUMMARY_DIR` (`/tmp/claude`), and the auto-memory dir ([`sandbox_runtime.mjs#L76`](https://github.com/max-sixty/tend/blob/b62381f9e23bdb6893a990d5cdff995e0747ebeb/shared/steps/sandbox_runtime.mjs#L76)). `/tmp` itself is not in that set, so it is read-only to the sandbox user. 103 references across 21 bundled files still write there. This shipped in 0.2.3 and 0.2.4, and this repo's own workflows are pinned at `max-sixty/tend/claude@0.2.4`, so every adopter picks it up at their next regeneration.

Observed in this run (a `tend-notifications` poll): step 1 of the notifications skill is `... > /tmp/tend-notifications.json` ([`notifications/SKILL.md#L21`](https://github.com/max-sixty/tend/blob/b62381f9e23bdb6893a990d5cdff995e0747ebeb/plugins/tend-ci-runner/skills/notifications/SKILL.md#L21)), which failed with `/bin/bash: line 1: /tmp/tend-notifications.json: Read-only file system`. A bare `touch /tmp/probe-x` fails the same way, while `$HOME` and `/tmp/claude` are writable — matching the allowlist exactly.

## What breaks

**Bundled scripts, where the `/tmp` path is the production default.** Each reads an override env var, and nothing outside `generator/tests/` sets one, so the `/tmp` default is what runs:

| Script | Default |
|---|---|
| `review_preflight.py` | `REVIEWED_HEAD_FILE` → [`/tmp/reviewed-head`](https://github.com/max-sixty/tend/blob/b62381f9e23bdb6893a990d5cdff995e0747ebeb/plugins/tend-ci-runner/scripts/review_preflight.py#L196) |
| `nightly_workflow_update.py` | worktree `/tmp/tend-update-workflows`, plus its state, body, and SHA files |
| `review_runs.py`, `review_reviewers.py` | evidence state and `/tmp/findings.md` |
| `list_recent_runs.py`, `review_runs_corrections.py` | `/tmp/review-runs-since` |

**Prose recipes**, including the mandated comment path: `running-in-ci` says to compose the body at `/tmp/comment-body.md` and post with `--body-file`, and the same section documents that the `--body "..."` fallback silently deletes backticked spans. A session that cannot write the file is pushed toward the fallback that corrupts what it posts. `resolve-conflicts` asks for a `/tmp/pr-<number>` worktree per PR; `skill-pr-workflow.md` puts the skill-edit worktree under `/tmp` too.

Jobs are still green — 44 `tend-*` / `review-*` runs since 2026-09-07 all concluded `success` — because the failure lands inside the agent session, which improvises a writable path (this run moved to a directory under the agent home). The cost is per-session recovery work plus the silent pull toward the corrupting comment fallback, not a red check, which is why nothing has surfaced it yet.

## The decision

Both remedies are small; they differ in what they touch, and one of them moves a security boundary, so it is your call rather than mine:

1. **Widen the sandbox** — add a scratch directory to `allowWrite`. Adding `/tmp` wholesale would also hand the agent write access to the runner-created `/tmp/tend-agent-workspace-*` and `/tmp/tend-runtime.*` containers, so this wants a dedicated directory rather than the whole tree.
2. **Move the defaults and recipes onto an already-writable path** — the agent home, or the existing `/tmp/claude`. No sandbox change, but it touches 103 references, and `generator/tests/test_cross_file_contracts.py` pins the `/tmp/reviewed-head` default, so that contract moves with it.

Either way, note the constraint from [#559](https://github.com/max-sixty/tend/pull/559): `/tmp` was hard-coded in the first place because `$TMPDIR` is unset on plain GHA runners, where `"$TMPDIR/x"` expands to `/x` and fails. A bare `$TMPDIR` reintroduces that; `${TMPDIR:-/tmp}` or a tend-owned variable keeps both cases working.

max-sixty:
Tend now gives every sandboxed agent a standard writable `TMPDIR` under the sandbox user's home. Bundled scripts use Python's temporary-directory API, agent recipes use `$TMPDIR`, and step summaries use the same exported scratch directory, removing the separate `/tmp/claude` filesystem exception.

The hosted sandbox test verifies that root `/tmp` remains read-only while both Claude and Codex inherit the writable directory. Cross-file contracts prevent agent-executed recipes from reintroducing unscoped `/tmp` paths and keep the step-summary reader tied to the exported scratch directory.

Closes #1192

Thanks to @tend-agent for reporting in #1192.


tend-agent (commented):
The sandbox side of this is right — `agentHome` was already in SRT's `allowWrite`, so folding the scratch directory under it and deleting `TEND_STEP_SUMMARY_DIR` removes a real exception rather than relocating it. Two things left over.

**Two agent-executed recipes still write to the root temp directory**, and neither sits inside the new contract's scope (`plugins/tend-ci-runner`), so nothing catches them:

- `plugins/install-tend/skills/debug-tend-run/SKILL.md`, the session-log download step: `DEST=/tmp/session-logs/$RUN_ID`. `running-in-ci`'s "Investigating Other CI Runs" sends every CI session to that skill for log parsing, and this PR's own `review-reviewers` edit points at it in the line right after moving its own copy of the same recipe to `$TMPDIR` — so the two halves of one workflow now disagree. `gh run download --dir` fails there. Since that plugin also runs on a maintainer's own machine, where `TMPDIR` is frequently unset, `${TMPDIR:-/tmp}` is the form that holds in both contexts.
- `.claude/skills/running-tend/references/integration-test.md`, the failure-reporting step: `envsubst '$RUN_URL' > /tmp/integration-failure.md` and the `--body-file` that reads it back. That path only runs in the sandbox, so plain `$TMPDIR` fits — but the consequence is sharper than the first: the weekly integration test can no longer file the failure it exists to report, and a swallowed write leaves no signal anywhere.

The `/tmp/tend-oauth-code` paths in `install-tend`'s setup skill are a maintainer's own machine rather than the sandbox, so those should stay as they are — worth saying explicitly if you widen the contract's scope rather than just fixing the two files.

**A smaller call to make deliberately:** dropping `TEND_STEP_SUMMARY_DIR` leaves `AGENT_HOME / "tmp"` derived independently in `proxy/setup_sandbox.py` (`AGENT_TMP_DIR`) and in `shared/steps/launch_sandbox_runtime.py` (`step_summary_dir`), where the exported variable used to tie them together. Nothing runs both halves against each other — `test_launch_sandbox_runtime.py` builds the agent env file and the expected summary path from the same fixture — so if they diverge, `read_regular_nofollow` returns `None` for the missing file, `if skill_summary:` is false, and skill step summaries stop appearing with no warning. That is the exact drift shape `test_cross_file_contracts.py`'s docstring describes, so it seems worth an explicit decision rather than an accident of the deletion.

I can push the two path fixes to this branch if you'd like them in here.

tend-agent (approved):

PR #1201, opened 2026-09-09
fix(nightly): base the regen worktree on the fetched default branch

### LINKED ISSUE #1200: nightly: regen worktree bases on the checkout's HEAD, so a survey PR opened first gets stacked into the workflow-update PR
## Problem

`nightly_workflow_update.py prepare` cuts the regen worktree from the main checkout's `HEAD` when no `tend/update-workflows` PR is open, at [`nightly_workflow_update.py:101`](https://github.com/max-sixty/tend/blob/7562e7254777476bca682c7b59e93f68d6452c20/plugins/tend-ci-runner/scripts/nightly_workflow_update.py#L101):

```python
if open_prs:
    _run("git", "fetch", "origin", BRANCH)
    base = _run("git", "rev-parse", f"origin/{BRANCH}").stdout.strip()
else:
    _run("git", "push", "origin", "--delete", BRANCH, check=False)
    base = _run("git", "rev-parse", "HEAD").stdout.strip()
```

That assumes the main checkout is still sitting on the default branch by the time Step 7 runs, and nothing in the nightly guarantees it. Steps 6 and 8 open PRs from the main checkout with `git checkout -b`, and the skill's own instruction to poll a pushed PR's CI makes "open the survey PR, then do the regen while its checks run" a natural ordering. When that happens, `HEAD` is the survey PR's branch, and the workflow-regen PR opens stacked on it — carrying that PR's commit, under a body that describes only the regenerated workflow files.

This is the `else` leg that #721/#722 left in place: those fixed the *reuse* of a stale `tend/update-workflows` branch, and the fallback kept the pre-existing `HEAD`. Basing on `HEAD` also means a commit that lands on the default branch during a long nightly session is silently excluded from the regen base.

## Evidence

`PRQL/prql` nightly run [34320282053](https://github.com/PRQL/prql/actions/runs/34320282053), 2026-09-09. The session opened [PRQL/prql#6300](https://github.com/PRQL/prql/pull/6300) from `docs/mssql-dialect-example-top` (`git checkout -q -b docs/mssql-dialect-example-top origin/main`, one commit `978f16c0`), then ran `prepare` without returning to `main`. [PRQL/prql#6301](https://github.com/PRQL/prql/pull/6301) opened with **nine** files — the eight regenerated `tend-*.yaml` plus `web/website/data/examples/dialects.yaml` — under a body reading "The diff is generator output — no hand edits. The eight regenerated files differ from 0.2.0 only in…".

The adopter's own `tend-review` caught it six minutes later and a `tend-mention` session rebased `--onto origin/main` six minutes after that, so nothing merged wrong. The cost was a publicly wrong PR description for twelve minutes and ~$8 of review/mention sessions. Had a maintainer merged inside that window — 28 of these regen PRs have merged on that repo, mostly same-day and lightly scrutinised — an unrelated docs change lands unreviewed under a `chore: update tend workflows` title, and the still-open PR that owns it becomes a no-op whose later revisions can never take effect.

## Proposed fix

Base the fallback on a fresh fetch of the default branch rather than on whatever the checkout happens to be on:

```python
else:
    _run("git", "push", "origin", "--delete", BRANCH, check=False)
    default = github_cli.json_call(
        "repo", "view", "--repo", repo, "--json", "defaultBranchRef"
    )["defaultBranchRef"]["name"]
    _run("git", "fetch", "origin", default)
    base = _run("git", "rev-parse", "FETCH_HEAD").stdout.strip()
```

`repo` is already in scope from `github_cli.repository()` a few lines above, so this adds one API call and one fetch. It also removes the second failure mode above, since the base is refetched rather than read from a checkout that may be hours old.

The skill-level alternative — telling the nightly to return to the default branch before Step 7 — is weaker: it leaves the coupling in place and turns it into a rule every session has to remember, on a step where the failure is invisible until a reviewer diffs the PR.

tend-agent:
## Problem

When no `tend/update-workflows` PR is open, `nightly_workflow_update.py prepare` cut the regen worktree from the main checkout's `HEAD`. Nothing in the nightly keeps the checkout on the default branch: Steps 6 and 8 open their own PRs from that checkout with `git checkout -b`, and polling a pushed PR's CI makes "open the survey PR, then regen while its checks run" a natural ordering. In that ordering `HEAD` is the survey branch, so the workflow-update PR opens stacked on it — carrying an unrelated commit under a body that describes only the regenerated `tend-*.yaml` files. #721 fixed the sibling leg (reusing a stale `tend/update-workflows` branch) and left this fallback reading `HEAD`. The same line also meant a commit landing on the default branch during a long nightly session was silently excluded from the regen base.

## Solution

The fallback now resolves the repository's default branch, fetches it, and bases the worktree on `FETCH_HEAD`. The base no longer depends on where the checkout happens to be, and it is refetched rather than read from a checkout that may be hours old. Cost is one API call and one fetch; `repo` is already in scope from `github_cli.repository()` just above.

## Testing

`test_prepare_bases_a_fresh_branch_on_the_fetched_default_branch` stubs the checkout sitting on a survey branch — `git rev-parse HEAD` and `git rev-parse FETCH_HEAD` return different SHAs — and asserts the default branch is fetched and the worktree is created at the fetched SHA. It fails on `main` (no `git fetch origin trunk`, worktree based on the survey head) and passes here. `uv run pytest generator/tests/test_nightly_workflow_update.py` is green; the seven failures in the full run (`test_codex_runner`, `test_refresh_consumers`, `test_prepare_agent_workspace`, `test_run_claude`) reproduce unchanged with this branch stashed.

Closes #1200 — automated triage


plugins/tend-ci-runner/scripts/nightly_workflow_update.py
```diff
@@ -98,7 +98,11 @@ def _prepare() -> int:
         base = _run("git", "rev-parse", f"origin/{BRANCH}").stdout.strip()
     else:
         _run("git", "push", "origin", "--delete", BRANCH, check=False)
-        base = _run("git", "rev-parse", "HEAD").stdout.strip()
+        default = github_cli.json_call(
+            "repo", "view", "--repo", repo, "--json", "defaultBranchRef"
+        )["defaultBranchRef"]["name"]
```
tend-agent:
```suggestion
        default = github_cli.json_call(
            "repo", "view", repo, "--json", "defaultBranchRef"
        )["defaultBranchRef"]["name"]
```

tend-agent (commented):
`gh repo view` takes the repository as a positional argument and has no `--repo` flag, so the new fallback leg exits 1 before it ever fetches:

```
$ gh repo view --repo max-sixty/tend --json defaultBranchRef
unknown flag: --repo
$ gh repo view max-sixty/tend --json defaultBranchRef
{"defaultBranchRef":{"name":"main"}}
```

`json_call` raises on that exit, so `_prepare` dies every night that no `tend/update-workflows` PR is open — just after the preceding `git push origin --delete` has removed the branch — and no workflow-update PR opens at all. That is a worse outcome than the stacking this PR fixes, and it lands on exactly the leg the PR is about.

Nothing in CI can catch it: both `prepare` tests monkeypatch `github_cli.json_call`, so the real `gh` argv is never exercised. The inline suggestion drops the flag; both stubs match on `args[:2] == ("repo", "view")`, so all five tests in the file stay green with it applied.

If you'd rather not depend on `repo view`'s argument shape at all, `gh api repos/<repo>` → `["default_branch"]` is the form `shared/steps/security_preflight.py` and `generator/src/tend/checks.py` already use for this.

plugins/tend-ci-runner/scripts/nightly_workflow_update.py
tend-agent: Confirmed — `gh repo view --repo max-sixty/tend --json defaultBranchRef` exits 1 with `unknown flag: --repo`, so the leg died before the fetch, right after the branch delete.

Took the second option rather than the suggestion: `gh api repos/{repo}` → `default_branch`, in `6ebcbd7`. The bug was getting a CLI subcommand's argument shape wrong, and the REST path is the form [`shared/steps/security_preflight.py`](https://github.com/max-sixty/tend/blob/6ebcbd788304722bbe4d52f6f0de2d9b01e8b514/shared/steps/security_preflight.py#L84) and [`generator/src/tend/checks.py`](https://github.com/max-sixty/tend/blob/6ebcbd788304722bbe4d52f6f0de2d9b01e8b514/generator/src/tend/checks.py#L159) already use for this, so it removes that dependency instead of correcting one instance of it.

Both `prepare` stubs now match on `("api", "repos/owner/repo")` instead of `("repo", "view")` — still monkeypatched, so this doesn't exercise real `gh` argv either, but the matched string is now the REST path the code passes verbatim rather than a flag arrangement a stub can't distinguish from a wrong one. All five tests in the file pass, as does `pre-commit run --all-files`.

PR #1203, opened 2026-09-10
feat(review): apply the review's own findings on the bot's PRs

max-sixty:
When `tend-review` reviews a PR this bot authored, it posts its findings and a separate `tend-mention` session is dispatched to act on them as the author. That session starts cold — it has no memory of authoring — so it re-reads everything the reviewer already had in context, applies the fixes, and pushes, which fires another review. On `max-sixty/leaf` that hop costs ~$3.55 per boot at list price plus the re-review it triggers, for no independent judgment: it is the same model, told what to fix, starting over.

This makes the review session apply its own findings on PRs with no human author. Step 9 of the `review` skill already said that for third-party bot PRs (Dependabot, renovate); the bot's own PRs now follow the same rule, and the paragraph that told the reviewer to leave the work to an author session is gone. The order on a bot-authored PR: run the review, submit it against the head reviewed, resolve threads, then apply the findings in one batched push and poll CI to green. The review stays pinned to the head it reviewed — re-targeting onto the pushed head would make the queued `synchronize` run read it as already reviewed and exit without looking at the fix. Convergence is that run's job: it reviews the pushed head, resolves the threads the fix addressed, and fixes anything new, until a pass finds nothing. Human-authored PRs are unchanged: suggestions first, push only after the author accepts.

The mention verify script no longer dispatches on a review the bot wrote (`fresh_inline` and the bot-review-on-bot-PR branch are deleted; a human review on a bot PR still dispatches the author session). The dispatched prompt's "including when you authored the review" clause, the system prompt's self-loop exception, and the `running-in-ci` passages that narrated the old hand-off are cut over with it. Generated workflows in this repo are untouched per the pinned-release rule.

The gate's pre-existing mention-wins order stands: a bot review that names `@<bot>` still dispatches, and the weekly integration test's §6 relies on that to drive review → relay → dispatch → verify → handle with the single bot identity. The session it boots now exits silently per the self-loop guard, so §6 asserts the session ran (session-log artifact on the dispatch run, as §5 does for review) rather than a reply; outbound posting is already §4's assertion.

https://claude.ai/code/session_01BA3Qoj4SjTrvdPJp6eaNLf


shared/system-prompt.md
```diff
@@ -38,8 +38,6 @@ project-wide signal.
...
   workflows.
 - Self-loop guard: before responding to a comment or review, confirm the
-  triggering actor isn't the bot itself — unless a distinct role boundary
-  makes it real work, e.g. an actionable review (changes requested, or
-  findings/questions) your review workflow left on your own PR, which you
-  address as its author. A bot approval or ask-free bot comment: exit
-  silently.
+  triggering actor isn't the bot itself, and exit silently when it is. That
+  includes a review your review workflow left on your own PR: the session
+  that posts it applies the findings it raised.
```
tend-agent, line 43:
```suggestion
- Self-loop guard: before responding to a comment or review, confirm the
  triggering actor isn't the bot itself, and exit silently when it is —
  unless it is a review that names you by @-handle, the one self-triggered
  dispatch the mention gate keeps. A review your review workflow leaves on
  your own PR without one starts no session: the run that posts it applies
  the findings it raised.
```

tend-agent (commented):
`mention_verify.py` still dispatches a bot-authored review that names `@<bot>` — `test_a_mention_wins_over_the_bot_review_gate` locks that order — and the weekly integration test's §6 depends on it: it seeds `@tend-agent … quoting this token verbatim: $NONCE` from the bot account and asserts a reply carrying the nonce. The session that dispatch boots now reads two unconditional exits, though: the system prompt's self-loop guard and `running-in-ci`'s Self-conversation Guard both lost the role-boundary carve-out that used to authorize it. What §6 asserts is the reply, not the dispatch, so it would fail as "mention path broken" with the gate working exactly as this PR intends. Suggestions inline keep the new exit rule and re-carve only the explicit summons.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -376,7 +376,7 @@ If a maintainer has already addressed the point, exit silently unless you can ad
 
 ## Self-conversation Guard
 
-If you are responding to your own prior comment or review (not a human's reply to it), only respond if there is a distinct role boundary (e.g., you are the reviewer on your own PR and need to address review feedback). If there is no such role distinction, exit silently to avoid self-conversation loops.
+If you are responding to your own prior comment or review (not a human's reply to it), exit silently to avoid self-conversation loops.
```
tend-agent, line 379:
```suggestion
If you are responding to your own prior comment or review (not a human's reply to it), exit silently to avoid self-conversation loops — unless it is a review of yours that names you by @-handle, which the mention gate dispatches deliberately.
```

shared/system-prompt.md
max-sixty: Took the other route rather than re-carving the guard: both guards stay unconditional, and the integration test's §6 now asserts what §5 asserts — the dispatched session ran (session-log artifact on the `repository_dispatch` run) — instead of a reply. The session that boots on a self-mention exiting silently is the behaviour the guard intends, and outbound posting is already §4's assertion.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
max-sixty: Took the other route rather than re-carving the guard: both guards stay unconditional, and the integration test's §6 now asserts what §5 asserts — the dispatched session ran (session-log artifact on the `repository_dispatch` run) — instead of a reply. The session that boots on a self-mention exiting silently is the behaviour the guard intends, and outbound posting is already §4's assertion.

shared/system-prompt.md
tend-agent: Concern closed on the assertion itself: §6 no longer turns on a reply, and the artifact is a sound stand-in — `Token usage` runs `if: always()` and writes `token-usage.json` into `$RUNNER_TEMP/tend-logs` before the `if: always()` upload, so a run that reached the action uploads something however short the session was.

The other thing the nonce was absorbing survives, in a different shape. `relay` has no authorship gate — it fires on every same-repo review event, [including one the bot wrote](https://github.com/max-sixty/tend/blob/a938adbee5c050b59c0553eaa622077a947e13e4/generator/src/tend/templates/mention.yaml.j2#L93). So §5's own `tend-review` review still relays a `tend-mention-review` dispatch; under this PR `verify` returns no on it and `handle` is skipped, leaving a `repository_dispatch` run that concludes **success with no artifact**. §6 latches onto the newest dispatch run that differs from `PREV_RUN`, and that relay run races the snapshot — it fires when the review is submitted, while §5 is still polling the `tend-review` run through session teardown and artifact upload. When it lands on the far side, §6 asserts against it and exits `no session-log artifact on run <id>` with the mention path working exactly as intended: the false failure the nonce ruled out, reached now by run selection rather than by reply attribution.

Latching onto the newest run is the load-bearing part. Scanning the new ones is indifferent to how many register:

```bash
BEFORE=$(gh run list --repo tend-agent/tend-integration --workflow tend-mention \
  --event repository_dispatch --limit 30 --json databaseId --jq '.[].databaseId' | sort)
# ...seed the review...
FOUND=0
for _ in $(seq 1 60); do
  NEW=$(gh run list --repo tend-agent/tend-integration --workflow tend-mention \
    --event repository_dispatch --limit 30 --json databaseId,status \
    --jq '.[] | select(.status == "completed") | .databaseId' \
    | sort | comm -13 <(printf '%s\n' "$BEFORE") -)
  for RUN_ID in $NEW; do
    ARTIFACTS=$(gh api "repos/tend-agent/tend-integration/actions/runs/$RUN_ID/artifacts" \
      --jq '[.artifacts[] | select(.name | startswith("claude-session-logs"))] | length')
    [ "$ARTIFACTS" -ge 1 ] && FOUND=1
  done
  [ "$FOUND" = 1 ] && break
  sleep 10
done
[ "$FOUND" = 1 ] || { echo "tend-mention: no dispatched session ran"; exit 1; }
```

That keeps "a dispatched session ran" as the assertion and lets the skipped sibling be one of the runs it walks past. It does fold the old "run never registered" message into one failure; tracking whether `$NEW` was ever non-empty splits them again if that distinction is still worth a line.

generator/tests/test_mention_verify.py
```diff
@@ -442,6 +416,24 @@ def test_the_bots_review_on_someone_elses_pr_is_dropped(env: dict[str, str]) ->
     assert _verdict(env) == ("false", "")
 
 
+def test_the_bots_review_that_names_the_bot_still_runs(env: dict[str, str]) -> None:
+    """A summons by name is judged before authorship, and the weekly
+    integration test rides on that order: the bot account is the only identity
+    it has to drive the review → dispatch → reply chain with."""
```
tend-agent: The docstring still names the chain this commit removed — §6 now stops at the dispatched session.

```suggestion
    it has to drive the review → dispatch → session chain with."""
```

tend-agent (commented):
§6's new shape closes out last round's finding: the artifact assertion tests a session the guard now tells to exit silently, where the reply assertion tested behaviour the guard forbids. One inline suggestion below, plus two notes.

**The registration cap now measures a different thing.** Separate from the run-selection race open on [this thread](https://github.com/max-sixty/tend/pull/1203#discussion_r3983458196): before this change the `seq 1 24` loop watched for any `tend-mention` run — the relay run, which registers within seconds of the review — so 120s was ample. `--event repository_dispatch` moves the target past a whole job, since the dispatch registers only once `relay` has been allocated a runner and has executed its POST. That hop has taken 8–31s on this repo with runners free and 8–12 minutes without ([this `pull_request_review` run](https://github.com/max-sixty/tend/actions/runs/34326419316) started at 07:56:19; its dispatch registered at 08:06:53). The fixture repo is idle, so the fast case should be the norm there — but a slow one exits "dispatch run never registered" and §8 files an issue for a path that was working. The scanning rewrite proposed on that thread absorbs this; a narrower fix to run selection would not.

**Outside this diff.** `running-in-ci`'s "Investigating Other CI Runs" still says runs triggered by `pull_request_review` or `pull_request_review_comment` "sometimes produce no artifact when the session is very short". Under the relay those events only ever run `relay` — `verify` and `handle` skip, as they did on this PR's own review run — so they never produce one at all, and the session-bearing run is the `repository_dispatch` one §6 now keys on. Happy to push the correction.

shared/system-prompt.md
max-sixty: Adopted the scan: §6 now walks every dispatch run that is new since the snapshot and passes on the first one carrying a session-log artifact, so §5's relay-only dispatch is just a run it walks past. "Dispatch run never registered" and "no dispatched session ran" stay separate failures, and the single 60×10 s budget covers the relay hop under load. The stale "Investigating Other CI Runs" line is corrected in the same commit.

generator/tests/test_mention_verify.py
max-sixty: Applied.

.claude/skills/running-tend/references/integration-test.md
```diff
@@ -340,45 +348,36 @@ PR=$(gh pr list --repo tend-agent/tend-integration --state open \
...
-  REPLIES=$(gh pr view "$PR" --repo tend-agent/tend-integration --json comments \
-    --jq "[.comments[]
-           | select(.author.login == \"tend-agent\" and (.body | contains(\"$NONCE\")))]
-          | length")
-  [ "$REPLIES" -ge 1 ] && break
+  NEW=$(gh run list --repo tend-agent/tend-integration \
+    --workflow tend-mention --event repository_dispatch --limit 30 \
+    --json databaseId,status \
+    --jq '.[] | select(.status == "completed") | .databaseId' \
+    | sort | comm -13 <(printf '%s\n' "$BEFORE") -)
```
tend-agent:
```suggestion
  NEW=$(gh run list --repo tend-agent/tend-integration \
    --workflow tend-mention --event repository_dispatch --limit 30 \
    --json databaseId --jq '.[].databaseId' \
    | sort | comm -13 <(printf '%s\n' "$BEFORE") -)
```

tend-agent (commented):
The two failure messages don't stay separate the way §6's comment promises: `REGISTERED` is set inside the loop over `$NEW`, and the `--jq` narrows that to runs already at `status == "completed"`. A dispatch run that registered and is still executing when the 60×10s budget expires therefore exits `dispatch run never registered`, and §8 files an issue here blaming the relay leg for a hop that fired — the slow week the split exists to describe. (The snippet I suggested last round carried the same filter; for the tracking to mean "registered" it has to sit on the unfiltered listing.) Dropping the `status` filter costs nothing: the loop only ever concludes positively per run, so an in-progress run is one extra artifact query that the next iteration re-scans, and an artifact that lands before its run completes is still proof a session ran. Suggestion inline.

.claude/skills/running-tend/references/integration-test.md
max-sixty: Applied — the listing is unfiltered now; the artifact is the proof either way.

tend-agent (dismissed):

.claude/skills/running-tend/references/integration-test.md
```diff
@@ -340,45 +348,30 @@ PR=$(gh pr list --repo tend-agent/tend-integration --state open \
...
-PREV_RUN=$(gh run list --repo tend-agent/tend-integration \
-  --workflow tend-mention --limit 1 \
-  --json databaseId --jq '.[0].databaseId // empty')
-gh pr review "$PR" --repo tend-agent/tend-integration --comment \
-  --body "@tend-agent integration test: post a PR comment quoting this token verbatim: $NONCE"
+# Seed through the API so the cutoff is GitHub's clock, not the runner's.
+SEEDED=$(gh api --method POST "repos/tend-agent/tend-integration/pulls/$PR/reviews" \
+  -f event=COMMENT \
+  -f body="@tend-agent integration test: this review exists to exercise the mention path; no reply is expected." \
+  --jq .submitted_at)
```
tend-agent, line 356:
```suggestion
  --jq .submitted_at)
[ -n "$SEEDED" ] || { echo "tend-mention: seeding review not submitted"; exit 1; }
```

tend-agent (commented):
`SEEDED` is now the cutoff everything downstream keys on, but nothing asserts the seeding review was actually submitted. If the POST fails (a 422 on the body, a transient 5xx), `SEEDED` is empty and `gh run list --created '>='` returns `[]` at exit 0 rather than erroring — so the loop burns its full 600s and exits `dispatch run never registered`, pointing §8 at the relay leg for a failure that happened before any dispatch could exist. That is the same wrong-diagnosis shape the two separate messages exist to prevent. Every other seed in the file is guarded the same way (`[ -n "$PR" ] || { … }`); suggestion inline.

.claude/skills/running-tend/references/integration-test.md
max-sixty: Applied.

tend-agent (approved):

PR #1206, opened 2026-09-11
chore(config): run review-runs on the Claude harness

max-sixty:
The Codex trial on `review-runs`, added in #1179, is over. Its ChatGPT account has run out of weekly quota until Sep 16, so the nightly loop fails on a usage limit instead of reviewing the previous night's runs.

Removing the per-workflow override drops `review-runs` to the repo defaults, `claude` at `opus`, which is what every other workflow here already uses. Every `max-sixty/tend/...` ref under `.github/workflows/` now reads `claude@0.2.6`.

`uvx tend@0.2.6 init` also removes `tend-codex-auth-refresh.yaml`. That workflow exists to rotate the Codex refresh chain weekly, and nothing selects Codex any more. `CODEX_AUTH_JSON`, `CODEX_REFRESH_AUTH_JSON`, and `CODEX_REFRESH_PAT` stay in the `tend` environment and stop rotating, so bringing the harness back later needs a fresh `codex login` through the install skill rather than the stored bundles.

The second commit follows the reviewer's finding on the first. The `running-tend` overlay's restamp step calibrated its listing against three pinned action paths and two harnesses, which was this repo's state only while `review-runs` ran on Codex, so after this change the paragraph reads a healthy single-line listing as drift. The rule was always the version count; it now says so first, and the number of paths is described as following from the config rather than as a remembered count.

`uvx tend@0.2.6 check` passes; the `codex-auth` check drops out along with the harness. Full pytest suite and every pre-commit hook pass locally.

https://claude.ai/code/session_016aR4DdFRBB5X4C1bYUGoGu


tend-agent (commented):
One thing the diff leaves behind: the `running-tend` overlay calibrates the nightly's drift check against exactly the state this removes. Under **Nightly: restamp the hand-maintained workflow refs**, `.claude/skills/running-tend/SKILL.md` says "Several action paths are pinned at once (`claude`, `codex`, `codex/refresh`), so several lines is the normal case — as are two harnesses, while `.config/tend.yaml` overrides one workflow to `codex`." All three claims are false once this merges — the section's own `rg` now prints the single line `max-sixty/tend/claude@0.2.6`.

The cost is that the check stops discriminating: a nightly reading that paragraph can't tell a single-harness repo from a hand-maintained workflow that lost its pin, which is the drift the step exists to catch. Rewriting the clause so a single `claude` line is the stated normal — and several paths and harnesses are what to expect if a workflow is overridden back to `codex` — restores it. That's outside this diff, so I haven't left an inline suggestion; say the word and I'll push the edit onto this branch rather than leaving it for a follow-up.

tend-agent (approved):

PR #1207, opened 2026-09-11
fix(tests): stop the suite reading the agent session's own sandbox environment

tend-agent:
Five tests read tend's own runtime environment instead of setting it, so `uv run pytest` — the command CLAUDE.md gives every agent — comes back red inside a tend session and green on a CI runner. A session that runs the suite before pushing sees five failures that have nothing to do with its change, and has to work out from `ValueError: unknown AUTH_MODE: <unset>` that the cause is the lifecycle hosting it.

Three independent leaks:

- `generator/tests/test_codex_runner.py` asserts `codex run` refuses to start a second execution boundary, which holds only while `TEND_INSIDE_SANDBOX` is unset. Inside a session it is `1`, so the run proceeds past the refusal and dies later on the auth mode. `_set_sandbox_env` now clears it and `AUTH_MODE`, the way `shared/steps/test_sandbox_setup.py`'s `step_env` already clears `TEND_SANDBOX_SETUP`.
- `shared/steps/test_run_claude.py`'s `launch` fixture builds an explicit environment for every variable `run_claude` reads except `TEND_AUTO_MEMORY_SETTINGS`. The host run's settings file falls through, and three argv assertions pick up `-u CLAUDE_CODE_DISABLE_AUTO_MEMORY` and a `--settings` pair the test never asked for. Set to `""` alongside the other explicit empties.
- `shared/steps/test_prepare_agent_workspace.py` calls `prepare.main()`, which roots the agent workspace at the runner's `/tmp` — read-only under the SRT boundary, so the step reports `[Errno 30] Read-only file system`. The real `mkdtemp` still runs, with `tmp_path` as its parent, so the clone the assertions read is unchanged and pytest now owns the cleanup.

Each fix carries an assertion that fails on `main` in any environment, since CI cannot reproduce the leak on its own: the first two seed the inherited value before the fixture runs, and the third pins the workspace under `tmp_path` — `mkdtemp(dir="/tmp")` can never satisfy it. Verified by applying the three assertions alone to a `main` worktree, where all three fail.

Not addressed: `generator/tests/test_refresh_consumers.py` (2 tests) drives `refresh_consumers.py` through `uv run --script`, which resolves `ruamel.yaml` from PyPI. The sandbox has no DNS, so those two still fail in-session. That is the boundary working as designed rather than a test defect, and the failure names `pypi.org` and `dns error` directly, so it does not mislead the way these five did.

### DIFF
--- a/generator/tests/test_codex_runner.py
+++ b/generator/tests/test_codex_runner.py
@@ -40,6 +40,11 @@ def _set_sandbox_env(
     monkeypatch.setenv("AGENT_ENV_FILE", str(agent_env))
     monkeypatch.setenv("SANDBOX", "tend-sandbox")
     monkeypatch.setenv("CODEX_BIN", "/opt/codex/bin/codex")
+    # The two the runner refuses to start without, cleared rather than set: a
+    # test that wants them sets them itself, and the suite run inside a tend
+    # session would otherwise inherit the live lifecycle's own values.
+    monkeypatch.delenv("TEND_INSIDE_SANDBOX", raising=False)
+    monkeypatch.delenv("AUTH_MODE", raising=False)
     return action, agent_home, agent_env
 
 
@@ -301,6 +306,11 @@ def test_run_uses_staged_subscription_auth_without_responses_proxy(
 def test_run_refuses_to_create_a_second_execution_boundary(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
+    # Seeded rather than left absent: these are the values a suite run inside a
+    # tend session inherits from the live lifecycle, and the refusal has to
+    # hold against them the same way it holds on a bare runner.
+    monkeypatch.setenv("TEND_INSIDE_SANDBOX", "1")
+    monkeypatch.setenv("AUTH_MODE", "api-key")
     _set_sandbox_env(tmp_path, monkeypatch)
 
     with pytest.raises(RuntimeError, match="only inside the SRT lifecycle"):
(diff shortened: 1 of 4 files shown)


generator/tests/test_codex_runner.py
```diff
@@ -301,6 +306,11 @@ def run(args: list[str], **kwargs: object):
 def test_run_refuses_to_create_a_second_execution_boundary(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
+    # Seeded rather than left absent: these are the values a suite run inside a
+    # tend session inherits from the live lifecycle, and the refusal has to
+    # hold against them the same way it holds on a bare runner.
```
tend-agent: The refusal does not hold against these values — `codex/runner.py` raises only when `TEND_INSIDE_SANDBOX` is *not* `"1"`, so with the seed intact the run proceeds. What makes the assertion fire is `_set_sandbox_env` clearing them two lines down, which is also the whole point of seeding: it is what fails this test on a runner where the vars were never set. Reading top-down the comment says the opposite, so the seeds look redundant and a later edit can drop them without the suite noticing.

```suggestion
    # Seeded because `_set_sandbox_env` clears them below: with them set the
    # runner would proceed past the refusal, so that clearing is what the
    # assertion tests — absent the seeds it is vacuous on a bare runner.
```

tend-agent (commented):
Self-review. One gap in the deliverable, plus a comment fix inline.

`uv run pytest` still comes back red inside a session after this — `generator/tests/test_refresh_consumers.py`'s two tests resolve `ruamel.yaml` from PyPI through `uv run --script`, and the sandbox has no DNS. The PR body records that, but the next session runs the command documented in CLAUDE.md and reads CLAUDE.md, not this thread, so it re-derives the same diagnosis from a `dns error` traceback. CLAUDE.md's Commands section already carries this exact shape of caveat for `pre-commit` not being on the sandbox PATH; the matching line for the two network-bound tests belongs beside it. I'll push it together with the inline comment fix.

PR #1210, opened 2026-09-11
fix(review-runs): re-query the closure read before reporting a path as still red

### LINKED ISSUE #1209: review-runs: the live-work closure read returns a weeks-stale latest-green, so a fixed workflow reports as still red
The live-work closure read in `review-runs` returns a weeks-stale row often enough to be reproducible, and the recipe has no freshness check — so the sweep reports a workflow as still red on the default branch when it has been green for weeks.

## The recipe

`plugins/tend-ci-runner/skills/review-runs/SKILL.md:104` (unchanged at `0.2.6`):

```bash
gh api "repos/$GITHUB_REPOSITORY/actions/workflows/<basename of path>/runs?branch=$DEFAULT_BRANCH&status=success&per_page=1" \
```

The unwindowed `status=failure` page reaches back weeks, so most of its rows are already fixed; this call is the only thing that separates them. A stale answer therefore decides the sweep's default-branch verdict on its own.

## What was observed

Two consecutive `tend-review-runs` sweeps on `PRQL/prql`, 2026-09-10 and 2026-09-11, each read six workflow files in one batch. The same two files came back stale both days:

| Workflow | Returned 09-10 | Returned 09-11 | True latest green 09-11 |
| --- | --- | --- | --- |
| `tests.yaml` | `2026-08-16T10:53:22Z` | `2026-08-16T10:53:22Z` | `2026-09-11T03:35:45Z` |
| `tend-notifications.yaml` | `2026-08-21T11:26:11Z` | `2026-08-17T01:47:26Z` | `2026-09-11T07:49:39Z` |

An immediate re-query of the same two URLs returned the correct newest row every time, at `per_page` 1, 2, 3, 5 and 10 — so this is not a pagination or ordering defect in the recipe's shape. `tests.yaml` returning the byte-identical `2026-08-16T10:53:22Z` on two different days argues for a durable cached entry rather than random jitter. The other four files in the same batch (`tend-review-runs`, `tend-ci-fix`, `tend-nightly`, `tend-mention`) were current both days; I could not establish what distinguishes them — run volume does not, since `tend-mention.yaml` has 792 successful `main` runs against `tests.yaml`'s 803 and was fresh.

## Why it matters, and in which direction

Staleness only moves the timestamp backwards, so the read can never be *newer* than the truth and can never wrongly close a live red row. The single failure direction is the other one: a green that predates the red row it has to close leaves that row uncloseable, and the sweep reports the path as still failing. On 2026-09-11 the stale `tests.yaml` answer was older than that path's newest red row (`2026-08-23T19:10:44Z`), so taken at face value it would have produced "`tests` has been red on `main` since 2026-08-23 with no green since" — in a repo where `tests` was green that morning.

(Both sweeps caught it — the first by noticing the contradiction, the second because it had the first one's note. Neither published a wrong claim. The exposure is a third session with neither.)

## Suggested fix

A one-line condition on the existing call, rather than new machinery: when the returned green is *older* than the red row it is meant to close, re-query once before reporting the path as still red. That is exactly the case where the answer changes the verdict, and the case where a stale read is indistinguishable from a real unfixed failure.

tend-agent:
## Problem

`review-runs`'s live-work closure read — the per-workflow `status=success&per_page=1` call that decides whether a red default-branch row is already fixed — sometimes serves a cached row weeks behind the true latest green. An immediate re-query of the identical URL returns the current row. The recipe had no freshness check, so a stale answer decided the verdict on its own.

I reproduced it on `PRQL/prql` while triaging: the first call for `tests.yaml` returned `2026-08-16T10:53:22Z`, the next two returned `2026-09-11T03:35:45Z`. That first value is byte-identical to what the reporter saw on two separate days, which rules out random jitter and points at a durable cached entry.

Staleness only moves the answer backwards, so it can never wrongly close a live red row. The single failure direction is the other one: a green older than the red row it has to close leaves that row uncloseable, and the sweep publishes "this workflow has been red on `main` since X with no green since" about a workflow that is green.

## Solution

A one-line condition on the existing call rather than new machinery: when the returned green predates the red row it has to close, re-run the call once and take the newer answer. That is exactly the case where a stale read is indistinguishable from a real unfixed failure, and the only case where the extra call can change the verdict.

## Testing

`test_review_runs_rechecks_a_stale_closure_read` in `generator/tests/test_cross_file_contracts.py` pins the re-query instruction in the closure recipe, matching the existing contract tests over this skill. It fails on `main` and passes here.

Closes #1209 — automated triage

### DIFF
--- a/generator/tests/test_cross_file_contracts.py
+++ b/generator/tests/test_cross_file_contracts.py
@@ -215,6 +215,17 @@ def test_review_runs_pins_current_state_recovery() -> None:
     assert "failing default-branch CI with no bot fix in progress" in skill
 
 
+def test_review_runs_rechecks_a_stale_closure_read() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
+
+    assert (
+        "# Can serve a stale row; re-query once before reporting a path as still red."
+        in skill
+    )
+    assert "can never wrongly close a live red row" in skill
+    assert "re-run its closure call once and take the newer answer" in skill
+
+
 def test_outage_tracker_title_stays_in_sync() -> None:
     title = "Bot temporarily unavailable"
     reporter = _read("shared", "steps", "report_failure.py")
--- a/generator/tests/test_cross_file_contracts.py
+++ b/generator/tests/test_cross_file_contracts.py
@@ -222,7 +222,6 @@ def test_review_runs_rechecks_a_stale_closure_read() -> None:
         "# Can serve a stale row; re-query once before reporting a path as still red."
         in skill
     )
-    assert "can never wrongly close a live red row" in skill
     assert "re-run its closure call once and take the newer answer" in skill
 
 
--- a/plugins/tend-ci-runner/skills/review-runs/SKILL.md
+++ b/plugins/tend-ci-runner/skills/review-runs/SKILL.md
@@ -106,7 +106,7 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
     --jq '.workflow_runs[0] | {name, conclusion, created_at}'
   ```
 
-  The closure call sometimes serves a cached row weeks behind the true latest green, and an immediate re-query returns the current one. Staleness only moves the answer backwards, so it can never wrongly close a live red row — the one failure direction is a green too old to close a red one, which reports a fixed workflow as still red. So before reporting any path as still failing, re-run its closure call once and take the newer answer.
+  The closure call sometimes serves a cached row weeks behind the true latest green, and staleness only moves the answer backwards — so before reporting any path as still failing, re-run its closure call once and take the newer answer.
 
   Unwindowed on purpose: a failure nobody fixed is still live on the nights after it ran, so anchoring on `$TMPDIR/review-runs-since` would surface each one the night it happened and read as an all-clear afterwards. The page reaches back weeks, so most rows are already fixed and the closure call is what separates them. What it cannot close stays live: Dependabot's security updates have no workflow file, and each run's `name` carries a per-update ID that never recurs, so those rows close only through a fix PR or a tracker. Step 1's census reaches back 49h at most, so skip only the tend rows inside its window — a tend workflow red for longer than that, with no green since, is news here like any other row. Report the scope the claim rests on — "`main` is green" is read later as covering every workflow — naming the workflows checked and how far back the page reached.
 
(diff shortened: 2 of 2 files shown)


plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -101,10 +101,13 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
   # Closure: for each distinct `path` above that is a real workflow file, its
   # latest green default-branch run closes that path's older red rows.
   # `dynamic/dependabot/...` paths are not files and 404 here.
+  # Can serve a stale row; re-query once before reporting a path as still red.
   gh api "repos/$GITHUB_REPOSITORY/actions/workflows/<basename of path>/runs?branch=$DEFAULT_BRANCH&status=success&per_page=1" \
     --jq '.workflow_runs[0] | {name, conclusion, created_at}'
   ```
 
+  The closure call sometimes serves a cached row weeks behind the true latest green, and an immediate re-query returns the current one. Staleness only moves the answer backwards, so it can never wrongly close a live red row — the one failure direction is a green too old to close a red one, which reports a fixed workflow as still red. So before reporting any path as still failing, re-run its closure call once and take the newer answer.
```
```suggestion
  The closure call sometimes serves a cached row weeks behind the true latest green, and staleness only moves the answer backwards — so before reporting any path as still failing, re-run its closure call once and take the newer answer.
```

generator/tests/test_cross_file_contracts.py
```diff
@@ -215,6 +215,17 @@ def test_review_runs_pins_current_state_recovery() -> None:
...
 
+def test_review_runs_rechecks_a_stale_closure_read() -> None:
+    skill = _read("plugins", "tend-ci-runner", "skills", "review-runs", "SKILL.md")
+
+    assert (
+        "# Can serve a stale row; re-query once before reporting a path as still red."
+        in skill
+    )
+    assert "can never wrongly close a live red row" in skill
+    assert "re-run its closure call once and take the newer answer" in skill
```
tend-agent: The other two asserts pin the comment and the rule; this one pins why the rule is safe, which the trim above removes.

```suggestion
    assert "re-run its closure call once and take the newer answer" in skill
```

tend-agent (commented):
Two trims, and they have to land together — the test pins the sentence the first one removes.

The rule is one sentence; the rest of the paragraph restates the in-block comment directly above it and then explains what *can't* go wrong. `CLAUDE.md`'s **Authoring skills** section asks for the opposite ("cut motivation, anecdotes, and historical context unless required to apply the rule", "prefer recipe over narrative"), and this paragraph loads into every `review-runs` session. The direction is worth keeping — "staleness only moves the answer backwards" is what makes "take the newer answer" follow — but the reassurance that a live red row can never be wrongly closed isn't needed to apply the rule, and #1209 holds it durably.

Pinning that sentence in the contract test is the part that bites: it couples the test to rationale prose, so trimming the narrative later fails a test that isn't about the behavior. The two remaining asserts already cover the comment and the rule.

Since this is a bot-authored PR I'll push both as a commit rather than leave them for someone to apply.

PR #1212, opened 2026-09-11
test: drop the tests that pin wording in skills and docs

max-sixty:
A test asserting that a sentence appears in a `SKILL.md` constrains the wording without constraining behavior: the skill can say the same thing a hundred ways, the agent reads meaning rather than substrings, and every rewrite of the guidance has to rename the pinned phrase to stay green.

34 such tests go: all of `generator/tests/test_cross_file_contracts.py`, the five skill-text tests at the tail of `test_bot_review_state.py`, `test_weekly_skill_delegates_the_refresh_to_the_script`, `test_codex_generated_default_is_documented_for_new_installs`, and `test_sensitive_paths_are_documented_where_adopters_read_them` — the last of which locates a README paragraph by its bold heading before checking the paths inside it.

Five checks from the deleted file survive, because none of them pins a phrase. Four move to `test_repo_pins.py`: the repo-wide `/tmp` ban, the Python-only scripts directory, the review-reviewers matrix against `data/consumers.json`, and the single-release pin across workflow files. The fifth becomes a behavior test in `test_integration.py` — it renders `tend init` through `CliRunner` and asserts every file it writes lands under `.github` or `.config`, the two directories the nightly regeneration stages.

Five guards come back as structure rather than text, after the first round cut them along with the wording pins around them. `test_nightly_workflow_update.py` already records every subprocess call, so it now asserts `git add`, `git status` and `git diff` each run in the regeneration worktree and scope themselves to `.github` and `.config` — run anywhere else and they inspect the runner's own checkout, where `init` never ran; widen only the staging and the no-op check goes blind to a path `init` newly created. `test_repo_pins.py` gains two: the Codex action reaches `runner.py` three ways — invoking `install-plugin` and `stage-agents` directly, and naming the script in `Run Codex`'s `TEND_CODEX_RUNNER`, which `launch_sandbox_runtime.py` gates the sandbox passthrough on and `agent_lifecycle.py` indexes — and the Codex marketplace manifest declares both plugins the runner installs by name, each at a source path that is a directory. The runner's third command, `run`, reaches it from `agent_lifecycle.py` rather than from the action, and that branch had no test at all — a new `shared/steps/test_agent_lifecycle.py` drives `main` past both its gates against a recording script, covering the argv, the exit-code passthrough, and the unknown-harness error.

Two text pins on Python source also go, both already covered by tests that run the code: `test_codex_runner.py` asserts `model_provider="tend-openai"` in the real Codex argv, and `test_launch_sandbox_runtime.py` asserts the runner's own `GITHUB_TOKEN` never reaches the agent environment.

<details>
<summary>What was deliberately kept</summary>

`test_repo_pins.py` still has ~31 substring assertions into composite-action `run:` bodies and shell scripts — credential scrubbing, npm config isolation, AppArmor rollback ordering. The argument against a prose pin is that the skill can say the same thing a hundred ways, so the test constrains wording without constraining behavior. A shell body is executed literally, so there the string is the mechanism, and a composite action cannot be run in a unit test.

Not restored, against the reviewer's suggestion: the assertion that the codex default model appears in `README.md`, `docs/tend.example.yaml` and the install-tend skill. Deriving the value from `DEFAULT_MODEL_BY_HARNESS` rather than hardcoding it doesn't change what the test does — it looks for text in prose files, which is the thing this branch removes.

</details>

One coupling loses its only guard: `shared/steps/report_failure.py` defines `TITLE = "Bot temporarily unavailable"`, and both `review-runs/SKILL.md` and `ci-fix/SKILL.md` hardcode that string — review-runs to find the issue, ci-fix to exclude it from its own transient-tracker count. Routing on the title rather than a label is the underlying problem, and fixing it belongs in its own change.

https://claude.ai/code/session_01MhnrWnFVT5hnV198hQK85v


tend-agent (commented):
The thesis holds for the bulk of the 33, and the five-check rescue is the right shape. Four of the deletions aren't wording pins by the PR's own criterion, though, and each loses its only guard.

**`test_codex_generated_default_is_documented_for_new_installs`.** Only the `f"harness: codex   — {model} (default)"` spacing was wording; the rest pinned `DEFAULT_MODEL_BY_HARNESS["codex"]` against three doc sites. `test_generate.py` hardcodes `gpt-5.6-sol` so the generator half still fails loudly on a bump, but `README.md`, `docs/tend.example.yaml`, and `plugins/install-tend/skills/install-tend/SKILL.md` now drift silently — and `running-tend`'s weekly sweep instructs moving that default ("Compare the current Codex model catalog with `DEFAULT_MODEL_BY_HARNESS["codex"]`"), so the drift is scheduled rather than hypothetical. A value-only form — `assert model in docs`, `readme`, `install_skill`, without the surrounding prose — keeps the pair asserted without constraining a sentence, which is what `test_repo_pins.py`'s own docstring describes.

**`test_codex_harness_delegates_stateful_phases_to_its_runner`.** It asserted `codex/action.yaml` invokes `runner.py install-plugin` and `runner.py stage-agents`, and that no `awk` came back. The surviving codex-action test in `test_repo_pins.py` parses the same file but keys on `Set up credential-isolation sandbox`, `Configure Codex auth`, `Run Codex`, `Token usage` — not `Install Tend plugins (sandbox)` or `Stage AGENTS.md (sandbox)`. `test_codex_runner.py` covers what the commands do, never that the action still calls them. That is exactly the kept-criterion in the details section — a composite action can't be run in a unit test — so this one looks like it fell out with the file rather than by the rule.

**The nightly staging coupling is now one-sided.** `test_init_writes_only_under_the_two_directories_it_owns` hardcodes `staged = [".github", ".config"]`, but the deleted test also asserted the other end: `"git", "add", "-A", ".github", ".config"` twice in `nightly_workflow_update.py`, plus the matching `status --porcelain` and `diff --cached` pathspecs. `test_nightly_workflow_update.py`'s fake `_run` matches on `args[:3]` only, so narrowing the script to `.github/workflows` leaves the whole suite green while the new test keeps passing against a literal nobody checks. Asserting the pathspecs on the recorded `git add` call there would close it as a behavior test rather than a source-text pin.

**`test_codex_ci_installs_the_diagnostic_plugin_its_runner_skills_use`** had a JSON half too: `.agents/plugins/marketplace.json` listing both plugins with `{"source": "local", "path": "./plugins/install-tend"}`, and `plugins/install-tend/.codex-plugin/plugin.json` naming `skills` as `./skills/`. `codex/runner.py` shells `codex plugin add install-tend@tend` against that marketplace, and no CI job installs it, so a rename or a manifest key change surfaces only as a failed Codex run.

One note on the description, since it's the merge record: `test_sensitive_paths_are_documented_where_adopters_read_them` in `generator/tests/test_shared_steps.py` still reads `README.md` and `docs/security-model.md` and matches on the `**Config pinning**` paragraph, so "nothing in `generator/tests/` now reads a repo markdown file and asserts its wording" isn't quite right. Its path list is derived from the script's `SENSITIVE` array rather than hardcoded, which is a good reason to keep it — worth saying so alongside the other deliberate keeps.

Happy to push a commit restoring the four as value/structure-only assertions in `test_repo_pins.py` plus the pathspec assertion in `test_nightly_workflow_update.py`, if you'd like.

tend-agent (commented):
The pathspec restore closes the one-sided coupling, but the replacement drops a check the lines it replaced were carrying: the old assertion required `git status --porcelain` to run with `cwd == worktree`, and the new `scoped` comprehension discards `cwd` entirely. Nothing now asserts that any of `add`/`status`/`diff` runs inside the regen worktree — deleting `cwd=worktree` from `prepare`'s `git add` leaves the whole suite green. `prepare` would then stage and inspect the runner's own checkout, where `tend init` never ran: `status` comes back empty, it reports `changed: False`, and the nightly regeneration stops shipping without failing anything. Inline suggestion keeps the pathspec assertion and folds `cwd` back in.

One more command came out of the same deleted test and wasn't restored with the other two: `shared/steps/agent_lifecycle.py`'s codex branch runs `["/usr/bin/python3", "-E", "-s", str(runner), "run"]`, against `runner.py`'s `if args == ["run"]`. `test_launch_sandbox_runtime.py`'s codex cases fake the runtime rather than reaching that branch, so the reason given for restoring `install-plugin` and `stage-agents` — exact argv, no CI job exercises the pair — holds for `run` unchanged.

Worth deciding before this lands: #1174 adds `test_bot_pat_scope_set_is_stated_identically_everywhere` to `test_cross_file_contracts.py`, the file this PR deletes. Whichever merges second takes a delete/modify conflict, and that test is itself a prose pin by this branch's criterion — it asserts a sentence in `CLAUDE.md` and a `--scopes` string in the install skill.

generator/tests/test_nightly_workflow_update.py
```diff
@@ -121,10 +121,23 @@ def json_call(*args: str, **_: object) -> object:
...
+    scoped = {
+        args[1]: args[-2:]
+        for args, _ in worktree_calls
+        if args[1] in {"add", "status", "diff"}
+    }
+    assert scoped == {
+        "add": (".github", ".config"),
+        "status": (".github", ".config"),
+        "diff": (".github", ".config"),
+    }
```
tend-agent, line 142: Carrying `cwd` through keeps the assertion the replaced lines made — that these three run in the regen worktree, not the runner's checkout. Verified: passes as-is, and fails when `cwd=worktree` is removed from `prepare`'s `git add` or `git status`.

```suggestion
    scoped = {
        args[1]: (args[-2:], cwd)
        for args, cwd in worktree_calls
        if args[1] in {"add", "status", "diff"}
    }
    assert scoped == {
        "add": ((".github", ".config"), worktree),
        "status": ((".github", ".config"), worktree),
        "diff": ((".github", ".config"), worktree),
    }
```

tend-agent (commented):
Both earlier points are closed — I mutated each to check: dropping `cwd=worktree` from `prepare`'s `git add` or `git status` now fails `test_prepare_persists_every_value_the_agent_and_ship_need`, and renaming the dispatch argv fails `test_codex_runs_the_turn_through_the_runner`.

One new one, in the new file. `agent_lifecycle.main()` assigns `TEND_INSIDE_SANDBOX` on the real `os.environ` rather than through `monkeypatch`, so all three tests leave it set for the rest of the pytest process, and `test_agent_lifecycle.py` sorts first in `shared/steps/` — every sibling in the directory now runs with the flag on. Three modules refuse to run without it: `sandbox_setup.main`, `run_claude.main` and `codex/runner.run_codex` each raise "may run only inside the SRT lifecycle". Today every test that exercises one of those refusals deletes the variable itself, so nothing is red; a test written later that simply calls `sandbox_setup.main()` expecting the raise passes against its own file and silently takes the inside-sandbox branch in the full run — the direction where a boundary guard stops being asserted. Dropping a sibling test asserting `"TEND_INSIDE_SANDBOX" not in os.environ` into `shared/steps/` passes against the other 251 tests and fails as soon as this file is collected, so this is the only place in the directory that mutates the process environment outside `monkeypatch`.

shared/steps/test_agent_lifecycle.py
```diff
@@ -0,0 +1,59 @@
...
+import agent_lifecycle
+import pytest
+import sandbox_setup
+
+
+@pytest.fixture
+def past_setup(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Both gates `main` clears before it reaches the harness branch."""
+    monkeypatch.setattr(agent_lifecycle, "probe_boundary", lambda: None)
+    monkeypatch.setattr(sandbox_setup, "main", lambda: 0)
```
tend-agent, line 30: Letting `monkeypatch` own the key makes it record the pre-state and delete it on teardown, so `main`'s own assignment doesn't outlive the test. Verified: with this, a sibling asserting the variable is absent passes alongside the whole of `shared/steps/`.

```suggestion
@pytest.fixture(autouse=True)
def _owned_sandbox_flag(monkeypatch: pytest.MonkeyPatch) -> None:
    """`main` sets this on the real environ; let monkeypatch restore it."""
    monkeypatch.setenv("TEND_INSIDE_SANDBOX", "0")


@pytest.fixture
def past_setup(monkeypatch: pytest.MonkeyPatch) -> None:
    """Both gates `main` clears before it reaches the harness branch."""
    monkeypatch.setattr(agent_lifecycle, "probe_boundary", lambda: None)
    monkeypatch.setattr(sandbox_setup, "main", lambda: 0)
```

generator/tests/test_repo_pins.py
```diff
@@ -134,15 +121,50 @@ def test_codex_agent_never_receives_the_pat_or_api_key() -> None:
...
+    action = YAML(typ="safe", pure=True).load(
+        (REPO_ROOT / "codex" / "action.yaml").read_text()
+    )
+    invoked = {
+        step["run"].rsplit('"', 1)[-1].strip()
+        for step in action["runs"]["steps"]
+        if "runner.py" in step.get("run", "")
+    }
+
+    assert {"install-plugin", "stage-agents"} <= invoked
```
tend-agent, line 149: The action's third coupling to the runner — the env var that tells `agent_lifecycle` where `runner.py` is. Verified: passes as-is, and fails with `KeyError: 'TEND_CODEX_RUNNER'` once the line is deleted from `codex/action.yaml`.

```suggestion
    assert {"install-plugin", "stage-agents"} <= invoked

    steps = {step["name"]: step for step in action["runs"]["steps"]}
    assert steps["Run Codex"]["env"]["TEND_CODEX_RUNNER"] == (
        "${{ github.action_path }}/runner.py"
    )
```

tend-agent (commented):
The fixture closes the leak — verified by mutation: with it, a sibling test asserting `TEND_INSIDE_SANDBOX` is absent passes across all of `shared/steps/`; drop the `setenv` and that test fails. `""` works as well as `"0"` here, since `sandbox_setup.main`, `run_claude.main` and `codex/runner.run_codex` all compare against `"1"`.

One more command from the same deleted test didn't come back alongside `install-plugin` and `stage-agents`: `assert "TEND_CODEX_RUNNER" in action`. `codex/action.yaml` sets it in `Run Codex`'s `env`, `launch_sandbox_runtime.py` gates its passthrough on `os.environ.get("TEND_CODEX_RUNNER")`, and `agent_lifecycle.main` then indexes `os.environ["TEND_CODEX_RUNNER"]` — so deleting that one line makes the launcher silently skip the passthrough and the codex branch raise `KeyError` in an adopter's job. Verified: with the line removed the whole suite stays green apart from the two DNS tests. That is the argument that restored the other two — the action can't be run in a unit test, and the new `test_agent_lifecycle.py` pins the consuming end while nothing pins the producing one.

tend-agent (approved):

PR #1215, opened 2026-09-12
skills: shrink the guidance every session loads

max-sixty:
`running-in-ci` is loaded by every tend session and `review` by every review, and both carried text that a minority of sessions reach. In a sample of 39 leaf sessions, 3 pushed, 1 opened a PR, 4 replied inline, none dismissed an approval or read another run's log; of 16 reviews, 5 posted, 4 approved, 1 re-targeted after a mid-review push, none ran in draft mode or hit a 422. Every session paid for the text anyway — about half of leaf's spend is the fixed per-turn payload, and these two files are the largest part of it.

## Move action-specific guidance into `references/`

`running-in-ci/SKILL.md` keeps the rules every session needs (overlay loading, reading context, restrictions, ending the turn, weighing a fix, conversation conduct, reader-facing prose) and gains a References list naming each `references/` file and the action that triggers reading it. The moved sections — PR creation, pushing, CI monitoring, posting mechanics and the pre-post re-fetch, dismissing a standing approval, directives on others' work, other runs' session logs — are verbatim in the new files. The `author-association.md` symlink moves into `references/` with the section that uses it. 52 KB → 20 KB.

`review/SKILL.md` keeps steps 0–5, the verdict rules, and the APPROVE and preflight recipes; draft mode, re-targeting after a mid-review push, the pre-approve settled-rollup check with the post-approve CI outcomes, and the inline-suggestion payload with 422 recovery move to `references/`, listed under the workflow heading with the step that reads each. 41 KB → 26 KB.

Skills that cited a moved section by heading now cite the reference file, and every cross-skill citation carries the plugin prefix (`/tend-ci-runner:running-in-ci`). Pointers within the review skill name the step (**Submit**, **Pre-flight checks**) rather than its number. The Codex `AGENTS.md` tail tells Codex to read a named reference in full before the action, and CLAUDE.md's authoring rules record the placement rule.

## Decide the review pre-flight in the script

`review_preflight.py start` emitted `last_review_sha`, `force_full_review`, `force_pushed_since`, `author` and `bot_login`, and the skill spent five paragraphs combining them into two decisions. Every input was already in the snapshot, so the combination is the script's to make: `start` now emits `already_reviewed` and `self_authored`, and the step says what to do with each. The pre-flight step becomes a three-bullet list keyed on `already_reviewed`, `is_draft` and `incremental_path`.

`already_reviewed` reads `at_head` — the rewrite-aware field the same snapshot has carried since the anchor resolver landed in #971. The raw `last_review_sha == head_sha` comparison the prose used reports a rewritten head as reviewed, because a force push re-points an earlier review's `.commit_id` at the new head; that is what the deleted force-push paragraph existed to work around. `incremental_path` is likewise non-null only when an incremental applies, so the `force_full_review` guard on the trivial-skip rule was already enforced by the script.

Also dropped, as prose restating something already written down: the second copy of what `$TMPDIR/reviewed-head` holds, the settled-rollup criteria that live in `references/approving.md`, and the preflight's command-mode internals.

Left alone deliberately: having `start` skip outright when nothing new was said. The exception — an unanswered question directed at the bot — is not a crisp boundary, and getting it wrong means a maintainer question goes unanswered, while the payoff is one saved session boot. `head_sha` also stays in the snapshot: it is the session log's only record of which commit a run that finishes without posting was about.

`generator/tests/test_review_preflight.py` gains three cases — a review standing on the head, a force push that must not read as reviewed, and `ready_for_review` overriding both. `generator/tests/test_repo_pins.py` gains `test_skill_reference_citations_resolve`: with the rules behind rarer actions now reachable only through a `references/` citation, a citation pointing nowhere would leave a session acting without the rule while nothing fails, so every cited name must exist in some skill's `references/`. The merge from main takes #1212's deletion of the skill-text assertions this change would otherwise have reworded.

https://claude.ai/code/session_01PoS7vwmM9mqPmS2UGBLJiw


tend-agent (commented):
Nothing in the diff reads as wrong to me: I diffed each moved section against `origin/main` and the bodies are identical apart from the citation rewrites, every `references/…` path the skills now cite resolves (including the relocated `author-association.md` symlink), and `already_reviewed` reproduces the `at_head` gate `review_preflight.py post` already applies, so `start` and the posting boundary can no longer disagree about a rewritten head.

One thing the change makes newly load-bearing: a rule now reaches a session only if the path in its citation is right. Citations of `references/*.md` across the bundled skills go from 12 to 58 here, and nothing checks them — a typo or a later rename leaves a citation pointing nowhere, the session reads no file, and the miss is silent in exactly the direction that matters: no failing test, no lint, just a run that skipped the pre-post re-fetch or the `--body-file` mechanics and posted anyway. `generator/tests/test_repo_pins.py` already carries this guard for the other path-citation surface — `test_action_path_references_resolve`, introduced with the reasoning that "a path that resolves nowhere fails its step, for every adopter, on the first run after a release" — and the same shape over `plugins/*/skills/**/*.md` is about ten lines. I'm happy to push it to this branch if you'd rather it land with the restructure than as a follow-up.

tend-agent (approved):

PR #1217, opened 2026-09-12
review: decide the pre-approval check in the script, not in prose

max-sixty:
Before an empty-body APPROVE, `review/references/approving.md` had the agent run `poll_pr_checks.py snapshot` and then branch on its fields in prose. A failure beside pending checks meant poll until they settle and re-snapshot. A failure still standing at the poll cap meant resolving each failing check to its Actions run and approving only when every one was `cancelled`, because a cancelled contributor makes an `if: always()` merge-gate omnibus report `FAILURE`. The reference said so itself: "Don't leave this to improvisation: the same stale red must not draw an APPROVE on one PR and a withheld approval on the next." Every input to that decision was already in the script's hands.

`poll_pr_checks.py approval <pr> <sha>` replaces `snapshot` (this reference was its only caller) and prints one verdict, `approve:` or `withhold:`. It reuses the settle loop `poll` already ran, now extracted as `_settle`, and resolves run conclusions itself. A run it can't read — a 5xx, or a run id the repo doesn't resolve — decides nothing: it exits 2, the code the module already uses for an unreadable rollup, unless another failing check is confirmed real, which still withholds. `approving.md` keeps what the agent does with each verdict, and drops from 65 lines to 45.

The head-moved check goes rather than moving. Every review POST in the review skill runs behind `review_preflight.py post -- …`, which refuses a moved head at the moment of posting, so a second comparison before the wait guarded the same thing earlier and more weakly.

`generator/tests/test_poll_scripts.py` replaces the two snapshot tests with one seven-case table: nothing failing, only this run and `tend-review` in the rollup, terminal red, a red that settles green, and at the cap a cancelled run, a real failure, and a status context with no run to inspect. A second test covers a run that can't be read, alone and beside a real failure. Disabling the settle wait, the cancelled-run approval, or the unreadable-run guard each fails a test.

Follows #1215, which pushed the review pre-flight's decisions into `review_preflight.py` the same way.

https://claude.ai/code/session_01PoS7vwmM9mqPmS2UGBLJiw

### DIFF
--- a/plugins/tend-ci-runner/scripts/poll_pr_checks.py
+++ b/plugins/tend-ci-runner/scripts/poll_pr_checks.py
@@ -17,6 +17,7 @@ from typing import Any
 import github_cli
 
 SHA_RE = re.compile(r"^[0-9a-f]{40}$")
+RUN_ID_RE = re.compile(r"/actions/runs/(\d+)/")
 RED_CONCLUSIONS = {
     "FAILURE",
     "TIMED_OUT",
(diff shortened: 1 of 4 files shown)


plugins/tend-ci-runner/scripts/poll_pr_checks.py
```diff
@@ -205,15 +206,58 @@ def head_note(*, pr: str, repo: str, sha: str) -> None:
...
+def _run_conclusion(repo: str, failure: str) -> str:
+    """The conclusion of the Actions run behind a failed check, "" for a status."""
+    match = RUN_ID_RE.search(failure)
+    if not match:
+        return ""
+    view = github_cli.json_call(
+        "run", "view", match.group(1), "--repo", repo, "--json", "conclusion"
     )
-    return str(response["headRefOid"])
+    return str(view["conclusion"])
```
tend-agent:
```suggestion
    """The conclusion of the run behind a failed check; "" when unresolvable."""
    match = RUN_ID_RE.search(failure)
    if not match:
        return ""
    try:
        view = github_cli.json_call(
            "run", "view", match.group(1), "--repo", repo, "--json", "conclusion"
        )
        return str(view["conclusion"])
    except (subprocess.CalledProcessError, ValueError, KeyError, TypeError):
        return ""
```

tend-agent (commented):
`_run_conclusion` is the only `gh` call left unguarded in the module — `fetch_rollup` and `head_note` both catch `CalledProcessError`, and `enrich_tend_outage_issues.py` wraps its own `gh run view` the same way. When that call fails (a 5xx, a rate limit, a `detailsUrl` whose run id doesn't resolve in this repo), the exception unwinds to the `__main__` handler, which exits with gh's status — 1, the code `withhold:` already uses — having printed no verdict at all.

That is the one path where the reference gives the agent nothing to read: `approve:`/`withhold:` key on stdout, and "any other exit decided nothing" keys on an exit code this path shares with a real red. It also lands after the full ~9-minute settle, so the decision this PR moved into the script goes back to improvisation on exactly the run where it cost the most to reach. Returning `""` puts it on the terminal-red branch that already exists, which is what the old prose gave an unresolvable URL.

Reproduced against the current head with a `gh` stub that 404s on `run view`: the rollup read and the settle loop both complete, then the process exits 1 with an empty stdout.

plugins/tend-ci-runner/scripts/poll_pr_checks.py
max-sixty: Fixed in e4294a50, though not with the suggestion as written. `_run_conclusion` now catches the failure and returns `None`. When that leaves no failure confirmed real, `approval` exits 2 and names the unread checks on stderr — the same code the module already uses when it can't read the rollup, and the one `approving.md` maps to "decided nothing". If another failing check is confirmed real, it still prints `withhold:`, since the unread run can't change that verdict.

Returning `""` would have put a 5xx on the real-red branch, and a failed API read isn't evidence the check failed on its own merits. `test_approval_with_a_run_it_cannot_read` covers both cases.

tend-agent: Agreed, and the exit-2 framing is better than the `""` I suggested — a failed API read shouldn't land on the branch reserved for a check that failed on its merits, and `approving.md`'s "any other exit decided nothing" already gives the agent somewhere to go.

Verified against `e4294a50`: `_run_conclusion` has the one call site, and relaxing its guard to `c != "cancelled"` turns `test_approval_with_a_run_it_cannot_read`'s first case into `withhold:` with exit 1 — the shape the finding was about. Nothing further from me on this thread.

tend-agent (approved):

PR #1226, opened 2026-09-13
chore: bump setup-uv to v10.1.0

tend-agent:
Moves `astral-sh/setup-uv` from v10.0.1 to v10.1.0 where it ships to adopters: the `setup_tend_uv()` macro in `generator/src/tend/templates/macros.yaml.j2`, which renders the pinned step into `tend-mention`, `tend-notifications`, and the `tend-install-test` workflow `generate_install_test` produces; the commented `setup:` example in `docs/tend.example.yaml` (and its `install-tend` mirror); and the eleven regtest snapshots that pin them. The repo's own hand-maintained call sites — `ci.yaml`, `pypi-release.yaml`, `.config/tend.yaml` — moved separately in the repo-local pins PR (#1227, merged), per the sweep's split by who runs the result.

The macro is the only generator-side pin. `generate_install_test` renders its uv step from `setup_tend_uv()` too, so the three adopter-facing workflows can no longer drift apart on this ref — an earlier revision of this branch carried a second copy of the version string in `workflows.py`, and merging main folded that call site onto the shared macro instead.

v10.1.0 adds verification against astral-sh/versions checksums on top of the list baked into the action, which is what covers a uv released after the action itself. That gap is live right now: the baked-in list in v10.0.1 (`src/download/checksum/known-checksums.ts`) stops at uv 0.12.4, v10.1.0's (`known-checksums.json`) stops at 0.12.12, and the paired sweep PR moves `uv_version` to 0.12.13 — past both. It also starts respecting `no_proxy`/`NO_PROXY`, which it previously ignored — tend's own steps set neither, so nothing here changes, but an adopter whose runner exports them will see setup-uv honor them.

One test change beyond the refs: `test_init_notifications_has_precheck` asserted the exact `astral-sh/setup-uv@v10.0.1` string while checking that uv is set up *before* the notification pre-check step. It now matches on the action prefix, so it keeps testing the ordering it is about and stops needing an edit on every sweep — the same version-agnostic shape `test_install_test_workflow_regtest` already uses.

The generated `tend-*.yaml` in this repo still read v10.0.1; they track the latest release and move at the next nightly regen, not here.


tend-agent (commented):
The bump misses one adopter-facing call site: `generate_install_test` in [`generator/src/tend/workflows.py`](https://github.com/max-sixty/tend/blob/main/generator/src/tend/workflows.py) still renders `uses: astral-sh/setup-uv@v10.0.1` into `tend-install-test.yaml` (the step right after `actions/checkout@v7`, both gated on `TEND_ENABLED_CONDITION`). That file is generated into the adopter's repo by `tend init --with-install-test`, so it ships the same way the `mention` and `notifications` templates do — and the v10.0.1 bump (#1002) moved it for exactly that reason.

Nothing catches this: `test_install_test_workflow_regtest` only asserts `"astral-sh/setup-uv@" in content`, deliberately version-agnostic, and its two snapshots pin whatever `workflows.py` currently says — so both stay green while the two templates and this one disagree. With the pins split across v10.0.1 and v10.1.0, next week's `uses:` sweep also reads setup-uv as "listed twice", which the sweep treats as deliberate and skips.

Pushing the one-line ref change plus the two regenerated `test_install_test_workflow_regtest` snapshots to this branch.

PR #1227, opened 2026-09-13
chore: bump repo-local pins

tend-agent:
The weekly sweep's repo-local half — pins nothing outside this repository runs.
The adopter-facing pins moved in their own PRs (#1223 `claude_version`, #1224
`sandbox_runtime_version`, #1225 `uv_version`, #1226 `setup-uv` in the
templates).

- **pre-commit revs**: `ruff-pre-commit` v0.16.6 → v0.16.7, `uv-pre-commit`
  0.12.10 → 0.12.13 (`pre-commit autoupdate`; the other four hooks were already
  current).
- **`astral-sh/setup-uv` v10.0.1 → v10.1.0** at the three call sites this repo
  runs itself: `ci.yaml`, `pypi-release.yaml`, and `.config/tend.yaml`'s
  `setup:`. #1226 carries the same move in the templates.
- **`WORKTRUNK_VERSION` 0.76.0 → 0.77.0** in `.config/codex-cloud/environment.sh`,
  with the two documented `wt list` recipes it breaks — see below.
- **`site/package-lock.json`**: astro 7.3.1 → 7.3.2 and its
  `@astrojs/markdown-satteri` 0.4.0 → 0.4.1, closing the `npm --prefix site
  outdated` drift. The worker's own lockfile drift (wrangler, workers-types) is
  already open as #1220, so it is not duplicated here. Worker `typescript` 5.9.3
  → 7.0.2 is a major and stays out of this PR.

`uv lock --upgrade` reports no lockfile changes, and `mitmproxy` is already at
its latest release, so neither moves this week.

**The worktrunk bump is breaking, and this PR carries the fallout.** 0.77.0
makes `wt list --format=json` default to schema 2, which wraps the results in an
envelope instead of returning a bare array. `CLAUDE.md` and `site/README.md`
both told the reader to get the dev-server URL with
`wt list statusline --format json | jq -r '.[].url'`, which returns nothing
against schema 2; both now read `jq -r '.items[].dev_server.url'`. Schema 2
is not the schema-1 items under an envelope: the dev server moved from a flat
`url`/`url_active` pair into a nested object, so `.items[].url` resolves to
null. Checked against worktrunk's published schema at tag `v0.77.0`
(`docs/public/schema/list-v2.json`): `JsonItemV2` carries `branch, checks,
default_branch, dev_server, display, head, marker, pr, remote, summary,
upstream, vars, worktree`, and `JsonDevServer` requires `url` and `listening`.
[...]

PR #1229, opened 2026-09-13
chore: bump worker typescript to 7.0.2

tend-agent:
Moves the worker's `typescript` devDependency from `^5.5.0` to `^7.0.0`, resolving 7.0.2. This is the one major in this week's pin sweep, so it is its own PR rather than part of #1227's repo-local pins.

TypeScript 7 is the native (Go) port of the compiler, so `tsc` is now a thin launcher over a platform binary: the lockfile gains twenty `@typescript/typescript-<platform>` optional entries alongside the `typescript` bump. `npm ci` on the CI runner resolves only the `linux-x64` one. The 7.x package also drops the `tsserver` bin, which nothing in this repo invokes.

No dependency version outside `typescript` moves — the worker's wrangler and `@cloudflare/workers-types` drift is #1220's, and this PR does not touch it. The regeneration does drop the `libc` field from ten `@rolldown/binding-linux-*` and `lightningcss-linux-*` optional entries. That is metadata only, and environmental rather than part of this bump: the field was written by the npm 11 that ran Dependabot's last lockfile update, and the npm 10.9.8 in this environment doesn't write it, so any lockfile rewrite here strips it. npm 10 doesn't read the field either, so `npm ci` resolves the same tree; #1220 carries the same ten deletions, and the next rewrite under npm 11 restores them.

The worker is a good first target for the major: it is `type: module`, its `tsconfig.json` is already on `moduleResolution: bundler` with `isolatedModules` and `strict`, and `tsc` runs only as `--noEmit` typechecking — vitest does the transforms and wrangler does the bundling, so none of TypeScript 7's emit changes are on any path here.

Verified locally on 7.0.2:

- `npm --prefix worker run typecheck` (`tsc --noEmit`) exits 0 with no diagnostics — no source or `tsconfig.json` change was needed.
- `npm --prefix worker test` passes all 32 vitest tests.
- `npm --prefix worker outdated` is then silent, which is the sweep's exit condition for this ecosystem.

CI's `test-worker` job re-runs both against a clean `npm ci` on ubuntu-24.04 / node 22.


tend-agent (commented):
#1220 rewrites `worker/package-lock.json` too (wrangler 4.131.1, taking sharp 0.35.4). Not a duplicate — the two move different dependencies — but they overlap on the whole lockfile, so whichever merges second will conflict, and resolving it means re-running `npm install` on the rebased branch rather than hand-merging the hunks.

The self-review found nothing else: `tsc --noEmit` is the only TypeScript entry point here, the tsconfig needs no change under 7.x, and the lockfile's only non-`typescript` movement is the ten `libc` deletions the description now accounts for.

PR #1230, opened 2026-09-13
mention: move the review-event relay into its own workflow

max-sixty:
Every review on a PR added three `tend-mention` rows to its checks, and on a fork PR all three were skipped — e.g. https://github.com/max-sixty/worktrunk/pull/4082. GitHub lists every job of a PR event's run among the PR's checks, skipped jobs included, and a `pull_request_review` run of `tend-mention` could only ever execute `relay`: `verify` and `handle` run on the `repository_dispatch` the relay sends.

The relay now lives in a generated `tend-mention-relay.yaml`, the only workflow subscribed to `pull_request_review` and `pull_request_review_comment`. A review adds one row: green on a same-repo PR, skipped on a fork PR. One row is the floor while anything listens to review events, since `on:` can't filter forks and a workflow has no top-level `if`.

- `generate_all` emits the relay alongside `mention` and drops it when `mention` is disabled. Like `codex-auth-refresh`, it sits outside `GENERATORS` (it runs no agent) and has its own `workflows.mention-relay` key for `enabled` and job overrides; an existing `workflows.mention.jobs.relay` override now warns as not found.
- The relay job is unchanged apart from its `if`, which loses the event-name check the split makes redundant. The trigger comment on why `pull_request_review_comment` skips `created` is rewritten: its old rationale, a cancelled `handle` posting on the PR head, predates the relay.
- Tests that assume every generated file runs the agent filter through a `without_relay` helper, and a new snapshot covers the relay file.

Adopters pick up the new file on their nightly regen after the next release.

https://claude.ai/code/session_01CxRxDRznNJKdTBwHB5kwHt


tend-agent (commented):
Every adopter-facing surface picked up the split except [`docs/tend.example.yaml`](https://github.com/max-sixty/tend/blob/main/docs/tend.example.yaml) — the config reference, which the `sync-install-tend-references` hook mirrors into the install-tend plugin. It introduces the workflow list with "All workflows are generated by default except ci-fix" and then carries a `### <name>` section for each one, so once this lands an adopter's regen drops a ninth file into `.github/workflows/` that the reference never names, and `workflows.mention-relay.enabled: false` — the key that now turns the review-mention path off — is undocumented. Its `### mention` section also still reads "Triggers on @bot mentions and review comments", which after this PR describes the relay rather than mention.

`codex-auth-refresh` has the same gap, so this isn't a rule the file has been holding perfectly — but that one only exists on Codex installs, whereas `tend-mention-relay` lands in every install and shows up in the directory adopters actually look at.

Those lines are outside the diff, so no inline suggestion. Happy to push a commit adding a `### mention-relay` section and correcting the mention trigger line — edited in the canonical `docs/` copy with the sync hook re-run — if you want it in this PR rather than a follow-up.

tend-agent (approved):

PR #1233, opened 2026-09-13
Pause tend with a TEND_ENABLED repository variable

max-sixty:
Top-level `enabled` was a runtime kill switch: every generated job began with setup-uv and a step that fetched `.config/tend.yaml` from the default branch and ran `uvx tend@X.Y.Z enabled`, and every later step was gated on its output. That cost about 20 lines per job, put PyPI and tend's unpinned dependencies in front of every run, and needed a default-branch commit to pause.

Pausing is now a repository variable, with no regeneration or commit:

```
gh variable set TEND_ENABLED --body false   # pause
gh variable delete TEND_ENABLED             # resume
```

Every job that can boot the agent carries `vars.TEND_ENABLED != 'false'` in its job-level `if:`, so GitHub skips it before a runner starts:

```yaml
jobs:
  review:
    if: vars.TEND_ENABLED != 'false'
```

## What changes

- **Checked:** review, triage, ci-fix, nightly, weekly, review-runs, notifications, and mention's `verify` job (`handle` needs it, so it is skipped with it). **Not checked**, since they run no agent: mention-relay, the Codex credential refresher (keeps a paused subscription's tokens valid), and install-test.
- **Top-level `enabled` is refused** with an error naming the variable. Warned past as an unknown key, a config that paused tend would regenerate running workflows.
- **An `if:` override on an agent job must keep the check**, since it replaces the rendered condition whole; `init` refuses one that drops it, whether it comes through `jobs` or `workflow_extra`. The documented `tend:dismissed` review override now repeats the check, and drops its draft clause, which review hasn't had since #362.
- **Removed with the runtime gate:** the hidden `tend enabled` command, the `config_path` field and its "config must be inside the repository" check, YAML merge-key rejection (configs using `<<` now load), and `tend check`'s disabled line. Notifications, mention's verify job, and install-test keep their off-PATH uv through a `setup_tend_uv` macro.

Most of the deletions are regenerated snapshots (37 files) losing the gate steps; the generator change is in `generator/src/tend/workflows.py`, `config.py`, and the templates.

## Trade-offs

- The variable must be set on the repository or organization, not in the `tend` environment: a job's `if:` is evaluated before the job enters the environment.
- Runs already queued or in progress aren't stopped; `gh run cancel` stops one.
- The bot's write access can change the variable, so the README describes the pause as an operating switch rather than a security control.
- Only `false` (any case) pauses.

The next release's changelog should say top-level `enabled` was removed: adopters paused through a 0.2.x config set the variable before deleting the key.

## Testing

`uv run pytest` (1128 passed), all pre-commit hooks, and actionlint on a full generated set (both harnesses, install-test) plus the owner-guarded snapshots. Not yet exercised on live GitHub.


tend-agent (commented):
One gap in the new pause guard, plus a smaller call on `tend check`.

**`workflow_extra` bypasses the guard.** `_apply_extras` deep-merges
`workflow_extra` into the whole document before it inspects `wf_cfg.jobs`, and
`workflow_extra` is free-form — `config.py` only checks it is a mapping. So a
`jobs:` mapping inside it replaces a job's `if:` without reaching the refusal.
Generated against this branch:

```yaml
workflows:
  review:
    workflow_extra:
      jobs:
        review:
          if: "github.actor == 'someone'"
```

renders `tend-review.yaml` with `if: github.actor == 'someone'` and no error —
the review job then boots the agent with `TEND_ENABLED=false`, which is the
outcome the guard exists to prevent. Inline comment carries a shape that checks
the merged result instead, covering both override paths for about the same
lines. Happy to push it with a test if you'd rather not.

**`tend check` no longer mentions a paused repo.**
`test_cli_check_reports_disabled_runtime` went with the config key and nothing
replaced it, so `check` now reports all-green on a repo where every agent job
is skipping. It already holds a token, so
`gh api repos/{owner}/{repo}/actions/variables/TEND_ENABLED` would restore the
line — though its docstring scopes it to release integrity, protection, access,
and credentials, so leaving it out is defensible too.

generator/src/tend/workflows.py
```diff
@@ -463,6 +466,19 @@ def _apply_extras(wf: GeneratedWorkflow, wf_cfg: WorkflowConfig) -> GeneratedWor
         jobs = data.get("jobs", {})
         for job_name, job_extra in wf_cfg.jobs.items():
             if job_name in jobs:
+                # An `if:` override replaces the rendered condition whole, so
+                # one on a job that runs the agent must carry the pause check,
+                # or that job keeps running while tend is paused.
+                if (
+                    "if" in job_extra
+                    and TEND_ENABLED_CONDITION in jobs[job_name].get("if", "")
```
tend-agent: `workflow_extra` is deep-merged into the whole document above, and it is only validated as a mapping — so `workflow_extra.jobs.<name>.if` replaces the rendered condition without ever reaching this check.

Checking the merged result instead covers both override paths, and an override that deletes a guarded job's `if:` outright:

```python
# before the merges
guarded = {
    name
    for name, job in (data.get("jobs") or {}).items()
    if TEND_ENABLED_CONDITION in str(job.get("if", ""))
}
...
# after both merges
for name in guarded:
    job = data.get("jobs", {}).get(name)
    if job is not None and TEND_ENABLED_CONDITION not in str(job.get("if", "")):
        raise click.ClickException(...)
```

max-sixty: Fixed in 56d09616. `_apply_extras` now records which jobs carry the pause check before either merge, and refuses the merged workflow if one of them lost it, so `workflow_extra.jobs.<name>.if` is covered as well as `jobs.<name>.if`. The refusal test is parametrized over both paths and over `if: null`.

tend-agent: Confirmed — I generated against 56d09616 with the exact `workflow_extra` config from the review and with its per-job equivalent, and both now raise `An override drops tend's pause check from job 'review' (tend-review.yaml)`. An override that repeats the check still renders, and `jobs.review: null` deletes the job without tripping the refusal, matching the comment. Deriving `guarded` from the rendered `if:` also means the set tracks whichever jobs the templates guard, with no second list to keep in sync.

The `tend check` point from the review body is untouched and stays your call.

tend-agent (approved):

PR #1235, opened 2026-09-14
fix(notifications): count queued runs when checking who owns a subject

### LINKED ISSUE #1234: notifications: the defer check misses queued dedicated runs, so the poll acts on a subject tend-mention already owns
## Problem

The notifications skill defers a thread when a dedicated tend workflow is already handling its subject. The check it prescribes only counts runs the API reports as `in_progress`:

```bash
IN_PROGRESS=$(gh api \
  "repos/$GITHUB_REPOSITORY/actions/runs?status=in_progress&per_page=100" \
  | jq --arg title "$SUBJECT_TITLE" --argjson own "$GITHUB_RUN_ID" \
      '[.workflow_runs[]
        | select(.name | startswith("tend-"))
        | select(.id != $own and .display_title == $title)] | length')
```

`queued` is a distinct status from `in_progress` in the Actions API, so a dedicated run that has been created but hasn't started yet is invisible to this check. The poll reads its subject as unowned and handles it itself.

That is not a rare corner. `running-in-ci`'s own **Triggering issue/PR already closed** section says "Long `tend-mention` queues (hours, not minutes) make this common" — so `tend-mention`, the workflow whose whole job is to carry a maintainer's direct instruction, is the one most likely to be sitting in exactly the state this check can't see.

## Evidence

Observed this run in `max-sixty/worktrunk`, in the `tend-notifications` poll [34806274921](https://github.com/max-sixty/worktrunk/actions/runs/34806274921).

A maintainer comment on PR [#4082](https://github.com/max-sixty/worktrunk/pull/4082) at `2026-09-14T04:12:46Z` created `tend-mention` run [34805195584](https://github.com/max-sixty/worktrunk/actions/runs/34805195584) at `04:12:48Z`. My poll started at `04:30:02Z` and ran the check above a couple of minutes later. It returned four `tend-*` runs — three `tend-review`, plus my own — and `34805195584` was not among them. The response was not truncated (four results, `per_page=100`), so the run was `queued`, not missed by pagination. It has since moved to `in_progress`.

Also queued right now in the same repo: `tend-mention` run [34748577073](https://github.com/max-sixty/worktrunk/actions/runs/34748577073), created `2026-09-13T08:50:35Z` — queued for over 19 hours and still invisible to an `in_progress` query.

The consequence is the one the defer rule exists to prevent. Reading the subject as unowned, my poll acted on the maintainer's instruction: pushed a commit to the contributor's fork branch and posted a comment. The queued `tend-mention` run was booted for that same comment. Both sessions post as the same bot account, so the collision surfaces to the maintainer as the bot answering itself on someone else's PR. The downstream re-fetches (`references/posting.md`, `references/pushing.md`) are what stopped it from being worse, but they are a backstop for a sibling that lands mid-session, not a substitute for the ownership check.

This is a correctness issue rather than a compute one — the cost is a duplicate outward action on a contributor's PR, not a wasted run.

## Proposed fix

Count every non-terminal status, not just `in_progress`. The variable name and the surrounding prose ("still handling its subject") already mean "active"; `in_progress` is just an under-implementation of it.

Dropping the `status=` filter and selecting `.status != "completed"` client-side is the one-call version, but on a busy repo a 19-hour-old queued run can fall outside the newest 100, which is precisely the case that matters. Keeping the server-side filter and summing across the non-terminal statuses avoids that:

```bash
SUBJECT_TITLE=$(gh api "$SUBJECT_URL" --jq .title)
ACTIVE=0
for st in queued in_progress waiting requested pending; do
  n=$(gh api "repos/$GITHUB_REPOSITORY/actions/runs?status=$st&per_page=100" \
    | jq --arg title "$SUBJECT_TITLE" --argjson own "$GITHUB_RUN_ID" \
        '[.workflow_runs[]
          | select(.name | startswith("tend-"))
          | select(.id != $own and .display_title == $title)] | length')
  ACTIVE=$((ACTIVE + n))
done
echo "active=$ACTIVE"
```

Per **Scripts over prose recipes** in `running-in-ci`, that is past the size where a skill should carry the logic inline — it probably belongs in `plugins/tend-ci-runner/scripts/` as something like `count_active_runs.sh <subject-title>`, with the skill reduced to the invocation and the intent. Happy to leave the shape to you; the load-bearing part is that `queued` counts.

One thing worth deciding alongside it: `display_title` matching means a `tend-mention` run triggered by a comment on an issue whose title later changed won't match either. That is a separate gap and I have not seen it bite, so I have not folded it in here.

tend-agent:
## Problem

The notifications poll defers a thread when a dedicated `tend-*` workflow is already handling its subject, but the check counted only runs the Actions API reports as `in_progress`. `queued` is a separate status there, so a dedicated run GitHub had created but not yet started answered no query the poll made, and the poll read the subject as unowned and handled it itself.

The consequence is an outward one. When the queued run is `tend-mention` — the workflow that carries a maintainer's direct instruction, and the one most likely to sit in a long queue — both sessions act on the same comment under the same bot account, so the duplicate surfaces to the maintainer as the bot answering itself.

The gap is live in this repository right now: `status=queued` currently returns a `tend-mention` run whose `display_title` is a PR title, alongside a `tend-notifications` run. Neither appears in a `status=in_progress` query.

## Solution

`plugins/tend-ci-runner/scripts/active_subject_runs.py` takes the notification's subject URL and reports every `tend-*` run that is not `completed`, excluding the poll's own run. It queries each non-terminal status server-side — `queued`, `in_progress`, `waiting`, `requested`, `pending` — rather than scanning the newest 100 runs of any status and filtering client-side: a run queued for hours on a busy repository falls outside that window, and that is exactly the run the check exists to see. A run held at an environment gate (`waiting`, `requested`) owns its subject like a running one.

The skill now carries the invocation and the rule — any run listed means defer — instead of the pipeline, its `display_title` rationale, and the `gh api --jq` caveat that the script no longer needs.

The script reports the runs rather than a count, so a deferral can name what owns the thread.

## Testing

`generator/tests/test_active_subject_runs.py` drives the script through a fake `gh` that partitions runs by status the way the API does. With `ACTIVE_STATUSES` reduced to `("in_progress",)` — the behavior this replaces — four of the six fail, including the queued-run case. They pass on the change.

One test asserts that each non-terminal status is requested. That is load-bearing rather than redundant: the API answers an unrecognised `status=` with an empty list instead of an error, so a mistyped or dropped status would silently return "nobody owns this" with nothing else to catch it.

`uv run pytest` is green apart from the two `test_refresh_consumers.py` cases that need PyPI, which the session sandbox has no DNS for. `uv tool run pre-commit run --all-files` passes.

## Not included

`display_title` matching still misses a run whose subject was renamed after the run was created. The issue raises it as a separate gap and I have not seen it bite, so it is not folded in here.

It also misses the relayed `tend-mention` run that answers a review comment: that work runs as a `repository_dispatch`, whose `display_title` is the dispatch type `tend-mention-review` rather than the PR title, so no status query matches it. Closing that needs either a `run-name:` on the generated mention workflow — which also moves the run's `name`, the field the `tend-` prefix filter reads — or a rule that treats any active relay run as owning every PR subject. Both are generator-scope decisions rather than part of this change.

Closes #1234 — automated triage


tend-agent (commented):
The check still cannot see the `tend-mention` run that answers a review comment, so the duplicate-post path this closes stays open on that trigger.

For `pull_request_review` and `pull_request_review_comment`, `tend-mention`'s secret-bearing `verify` and `handle` jobs run in the relayed `repository_dispatch` run rather than in the run the review event created — `.github/workflows/tend-mention.yaml` says so at the trigger: "their runs carry `refs/pull/N/merge`, which the operational secrets' environment does not admit, so the secret-bearing jobs see them re-entered as a `repository_dispatch` instead". GitHub gives that run a `display_title` of `tend-mention-review`, the dispatch event type, not the PR title; every completed relay run in this repository's run list shows exactly that. So `owning_runs`' `display_title != subject_title` filter drops it at every status, and a poll landing while it is queued or running reads the PR as unowned and can answer the same review comment the dispatch run was booted for.

That is the same outward failure this PR exists to prevent, on the mention path that carries inline review instructions. It is not a regression — the `in_progress`-only check missed it too — but the new prose ("that run owns its subject as much as a running one") reads as coverage the match does not reach.

I did not fold a fix in, because neither shape is a defensible default: setting `run-name:` on the generated mention workflow so the dispatch run displays the PR title also moves the run's `name`, which is the field `WORKFLOW_PREFIX` filters on; and treating any active `tend-mention-review` run as owning every PR subject over-defers unrelated threads for as long as one sits queued. Both are generator-scope decisions. So this is a coverage note rather than a blocker — counting queued runs is a strict improvement on its own.

PR #1243, opened 2026-09-14
fix(review-runs): converge the live-work red listing before trusting it

### LINKED ISSUE #1241: review-runs: the live-work red listing serves a cached page, so a live default-branch failure reports as an all-clear
`review-runs`'s live-work sweep decides "is the default branch red right now?" from three `status=`-filtered run listings. That call serves a cached page often enough to reproduce, and on this call staleness omits the newest red rows — so the sweep reports an all-clear while a failure stands on the default branch. This is the direction #1209 and [#1210](https://github.com/max-sixty/tend/pull/1210) reasoned was impossible; the reasoning holds for the closure read they fixed, but not for the listing that feeds it.

## The recipe

`plugins/tend-ci-runner/skills/review-runs/SKILL.md`, **Reconcile live work** (unchanged at `0.2.7`):

```bash
for status in failure startup_failure timed_out; do
  gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=$status&per_page=50" \
    --jq '.workflow_runs[] | {id, name, path, event, conclusion, created_at}'
done
```

Only the per-path closure call below it has a freshness check (`#1210`). The listing that produces the rows has none, and nothing downstream can recover a row it never returned.

## What was observed

`max-sixty/worktrunk`, `tend-review-runs` [run 34819995486](https://github.com/max-sixty/worktrunk/actions/runs/34819995486), 2026-09-14. Three calls to the byte-identical URL, minutes apart in one session:

| Call | Newest row returned | `total_count` |
| --- | --- | --- |
| 1st (cold, ~07:58Z) | `2026-09-05T20:32:15Z` | 587 |
| 2nd (~08:02Z) | `2026-09-14T05:06:36Z` | 587 |
| 3rd (~08:02Z, `&page=1`) | `2026-09-14T05:06:36Z` | 587 |

The row the cold call dropped is [run 34808417481](https://github.com/max-sixty/worktrunk/actions/runs/34808417481) — `nightly`, `push` on `main`, `conclusion: failure`, commit `ac23e56d`, created `2026-09-14T05:06:36Z`. It was live at the time of the cold call: it was `nightly`'s newest *completed* run on `main`, and the re-run that cleared the path ([34810650044](https://github.com/max-sixty/worktrunk/actions/runs/34810650044)) only concluded `success` at `2026-09-14T08:05:24Z`. Taken at face value, the cold answer supports "newest red row of any kind is 2026-09-05; `main` is green". That is the exact shape of claim this sweep publishes each night — the previous window's entry reads "Newest red row of any kind is 2026-09-07", correct at the time, from this same call.

Two details line up with #1209's diagnosis rather than with pagination or ordering: `total_count` is identical across all three calls while the page contents differ, and the cold page's rows are internally consistent — it is a coherent older snapshot, not a truncated one. Per-workflow listings *without* the `status=` filter were fresh in the same session: `.../workflows/nightly.yaml/runs?branch=main&per_page=5` returned the 05:06 failure on its first call.

That the failure was not missed here is luck, not the sweep working: `tend-ci-fix` booted on it independently, diagnosed it as a transient `setup-nu` DNS failure, and filed [max-sixty/worktrunk#4103](https://github.com/max-sixty/worktrunk/issues/4103). Nothing in `review-runs` cross-checks against that.

## Why the direction matters

#1210's PR body says: "Staleness only moves the answer backwards, so it can never wrongly close a live red row. The single failure direction is the other one." On the closure read that is right — a too-old green fails to close a row, and the sweep over-reports. On the red listing the same backwards motion drops the newest rows, and the sweep under-reports: a live failure on the default branch is reported as an all-clear. The skill's own text tells the session that the claim carries weight — "Report the scope the claim rests on — `main` is green is read later as covering every workflow" — so the wrong answer ships as a published claim, not an internal miss. The current SKILL.md text also carries the reassurance in the narrower form ("staleness only moves the answer backwards"), attached to the closure paragraph, where a session scanning the section can read it as covering the sweep.

## Proposed fix

The same remedy #1210 applied one call up, on the call that needs it: issue each `status=` listing twice and take the union, rather than trusting the first page.

```bash
for status in failure startup_failure timed_out; do
  # Cached pages are served on the cold call and drop the newest rows, so take
  # the union of two reads (max-sixty/tend#1209 is the same defect one call up).
  for attempt in 1 2; do
    gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=$status&per_page=50" \
      --jq '.workflow_runs[] | {id, name, path, event, conclusion, created_at}'
  done
done | sort -u
```

Three extra API calls, no state a session has to carry, and the union can only add rows — a spurious row costs one closure call, which the recipe already makes.

An alternative worth considering if you'd rather not double the calls: keep the three filtered calls for depth and add one unfiltered `?branch=$DEFAULT_BRANCH&per_page=100` pass filtered client-side for red conclusions, unioned in. The unfiltered listing was demonstrably fresh in this session, and it covers exactly the recent window where a dropped row is still live work. It cannot replace the filtered calls — on a busy repo 100 runs is a day or two — but it is the half that matters for the all-clear.

Either way the "staleness only moves the answer backwards" sentence needs its scope stated, since on this call backwards means *omitting* rows.

tend-agent:
## Problem

`review-runs`'s live-work sweep decides whether the default branch is red right now from three `status=`-filtered run listings. #1210 gave the per-path closure call below them a freshness check; the listing that *produces* the rows still had none, and nothing downstream can recover a row it never returned.

The two calls fail in opposite directions, which is why one fix did not cover the other. On the closure read a stale answer is a too-old green: it fails to close a path that is already fixed, and the sweep over-reports. On the red listing the omitted rows are the *newest* ones, so the sweep under-reports — a stale page supports "`main` is green" while a failure stands on it. The skill instructs the session to publish that claim with its scope stated, so the wrong answer ships as a published claim rather than an internal miss.

The reporter's evidence checks out against the API: on `max-sixty/worktrunk` the cold call at ~07:58Z on 2026-09-14 returned a newest red row of `2026-09-05T20:32:15Z`, while `nightly` run [34808417481](https://github.com/max-sixty/worktrunk/actions/runs/34808417481) — `push` on `main`, `conclusion: failure`, created `2026-09-14T05:06:36Z` — was live at that moment: the re-run that cleared the path ([34810650044](https://github.com/max-sixty/worktrunk/actions/runs/34810650044)) only concluded `success` at `2026-09-14T08:05:24Z`. Calls two and three to the byte-identical URL, minutes later, returned it.

## Solution

Re-read each `status=` filter until two consecutive pages agree — capped at four reads — and union what they returned. On an endpoint answering consistently that is two calls, the same cost as reading each filter once more unconditionally; it spends more only where the answers are still moving, which is the case that escapes a fixed pair.

A fixed pair was the first shape here, and review found it insufficient. Probing untouched repos, `astral-sh/uv` went stale-then-fresh across two adjacent reads of the identical URL (newest row `2026-08-18T15:55:41Z`, then `2026-09-14T02:15:41Z`), but `openai/codex` took three — `2026-08-15T05:55:10Z`, `2026-08-28T00:09:18Z`, `2026-09-14T04:37:18Z` — so stopping at two would have published a newest red row of `2026-08-28` with a `2026-09-14` failure standing. A later single read of that same URL returned the `2026-08-15` snapshot again, so this is not a cold-start cost a session pays once: each call independently gets whichever snapshot it gets.

A stale page is internally coherent, so the rows cannot tell a session it has converged, and `total_count` is not a signal — #1241 recorded it identical at 587 across a stale and a fresh page. Agreement between consecutive reads is what is left.

The union is `sort -ru` rather than `sort -u`: gojq sorts object keys, so the lines lead with `conclusion` then `created_at`, and descending puts each conclusion group's newest row first — the row the section's published claim is about.

The prose that carried the reassurance is rescoped. It previously read "staleness only moves the answer backwards" attached to the closure paragraph, where a session scanning the section could take it as covering the sweep; it now names the direction each call fails in and what the loop waits for.

I considered the reporter's alternative — one unfiltered `?branch=main&per_page=100` pass filtered client-side — and it does not carry the case on its own. It serves stale pages too (a first read of it on this repo returned a newest row of `2026-08-28T00:36:57Z` with `total_count` 9372, against 17039 live), and on a busy repo it reaches back only hours. It would have to be unioned *in addition to* the filtered calls.

## Testing

No test. #1212 removed the 34 tests that pinned wording in skills and docs, including the one #1210 added for the closure read, on the grounds that a substring assertion constrains phrasing without constraining behavior — so adding one here would reinstate exactly what that change deleted.

Instead I ran the amended recipe against `openai/codex`, the repo where the drift was widest: it returns 71 rows, no blank lines from the empty-page guard, and its newest `failure` row is the `2026-09-14T04:37:18Z` one that the first read had dropped. Against `max-sixty/tend` the doubled reads collapse cleanly under `sort -u` — 100 lines to 51 distinct run IDs — since `gh api --jq` emits one compact object per line with sorted keys. `pre-commit` passes on the file, and the tests that touch this skill (`test_review_runs_corrections.py`, `test_evidence_scripts.py`, `test_list_recent_runs.py`) pass at 39.

Closes #1241 — automated triage


tend-agent (commented):
The doubled read is the right direction and does real work. Probing `actions/runs` from this session against repos it had not touched: on `astral-sh/uv` the first read returned a newest red row of `2026-08-18T15:55:41Z` (`total_count` 1209) and an immediate repeat of the identical URL returned `2026-09-14T02:15:41Z` (`total_count` 1499). So a second read does reach a fresher snapshot, and adjacency is not a problem.

Two reads are not enough, though. On `openai/codex` the same probe took three to reach the live row — `2026-08-15T05:55:10Z` (`total_count` 3447), then `2026-08-28T00:09:18Z` (10228), then `2026-09-14T04:37:18Z` (13669). Stopping at two would have published "newest red row of any kind is 2026-08-28" with a `2026-09-14` failure standing on the branch, which is the claim shape this section exists to protect. Nor is it a cold-start cost a session pays once: a later single read of that same URL came back with the `2026-08-15` snapshot again after the fresh one, so each call independently gets whichever snapshot it gets.

A stale page is internally coherent, so the rows alone can't tell a session it has converged, and `total_count` is not a signal either — #1241 recorded it identical at 587 across a stale and a fresh page, and here it moved with the staleness. What the recipe can do is re-read each filter until two consecutive pages agree, capped, and union those. That costs the same two calls as this PR wherever the endpoint answers consistently, and spends more only where the answers are still moving — which is exactly the case that currently escapes.

Two smaller things in the same area. `sort -u` orders by `conclusion` then `created_at` ascending, because gojq sorts object keys — so the newest rows land at the bottom of each conclusion group, while the claim the section asks the session to make is about the newest row; `sort -ru` leads each group with its newest row at no extra cost. And the new paragraph restates what the code comment above it now says, with its last two sentences arguing for the design rather than instructing the session; CLAUDE.md's "Authoring skills" asks for the rule or recipe with motivation cut.

Pushing all three to the branch.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -94,10 +94,19 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
   # One server-side filter per red conclusion, to reach past a busy repo's green
   # runs. `cancelled` is left out: concurrency cancels dominate it.
+  # Each filter is re-read until two consecutive pages agree, then unioned: a
+  # single read can serve a coherent older page that omits the newest rows, and
+  # nothing below recovers a row never returned.
   for status in failure startup_failure timed_out; do
-    gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=$status&per_page=50" \
-      --jq '.workflow_runs[] | {id, name, path, event, conclusion, created_at}'
-  done
+    prev=
```
tend-agent: An empty `page` equals this initial value, so the first read of a filter with no rows — or of a snapshot old enough to predate every row — is taken as converged. A sentinel no page can produce makes the empty answer go through the same agreement test as any other.

```suggestion
    prev=__unread__
```

tend-agent (commented):
The convergence loop accepts an empty page on the first read, which is the one answer it most needs to re-check. `prev=` starts equal to what an empty page produces, so `[ "$page" = "$prev" ]` is true on iteration one and the loop breaks having read once. Probing this repo with `status=timed_out`, which has no rows: one read, then break.

That matters because empty is the maximally under-reporting answer, and it is reachable for the same reason the fixed pair was insufficient. `startup_failure` and `timed_out` typically carry a handful of rows spanning months on any one repo, so a snapshot old enough to predate all of them is empty rather than merely short — and the sweep then publishes "`main` is green" off a single unverified read, the claim shape this PR exists to protect. A `gh api` call that errors lands in the same branch (empty `page`, immediate break); its stderr is at least visible to the session, unlike the stale-empty case, which is indistinguishable from a genuinely clean filter.

Inline suggestion below: start `prev` at a value no page can equal. A genuinely empty filter then costs two reads instead of one, and a stale-empty one gets the same treatment as a stale-short one. Pushing that to the branch.

Separately, and a maintainer's call rather than mine: the block now runs ~22 lines carrying a retry cap, a convergence test and an empty-output guard, which is what `running-in-ci`'s "Scripts over prose recipes" points at — "When a skill's code block needs edge-case handling or grows past a couple of dozen lines, put the logic in a tested script and leave the skill a one-line invocation with the intent." `plugins/tend-ci-runner/scripts/` already carries `github_cli` and `list_recent_runs.py` for this workflow, and a script could separate a failed read from an empty one, which the shell form cannot. I'm not pushing that — the recipe is correct once `prev` is initialized, and the conversion is a change of shape, not a fix.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -94,10 +94,19 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
-  done
+    prev=__unread__  # no page equals this, so an empty answer is re-read too
+    for _ in 1 2 3 4; do
+      page=$(gh api "repos/$GITHUB_REPOSITORY/actions/runs?branch=$DEFAULT_BRANCH&status=$status&per_page=50" \
+        --jq '.workflow_runs[] | {id, name, path, event, conclusion, created_at}')
+      [ -n "$page" ] && printf '%s\n' "$page"
+      [ "$page" = "$prev" ] && break
+      prev=$page
+    done
+  done | sort -ru
```
max-sixty, line 109: this seems like quite messy code, can we improve? We seem to be mixing bash and jq, can we rewrite in a much better way?

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -106,7 +115,7 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
     --jq '.workflow_runs[0] | {name, conclusion, created_at}'
   ```
 
-  The closure call sometimes serves a cached row weeks behind the true latest green, and staleness only moves the answer backwards — so before reporting any path as still failing, re-run its closure call once and take the newer answer.
+  Both calls sometimes serve an older page, and staleness moves each answer backwards — in opposite directions. On the closure read a too-old green fails to close a path that is already fixed, so the sweep over-reports: re-run that call once and take the newer answer before reporting any path as still failing. On the red listing the missing rows are the *newest* ones, so the sweep under-reports — a stale page supports "`main` is green" while a failure stands on it. Consecutive reads of the same URL have returned different snapshots, and a stale page is internally coherent, so agreement between two of them is the only convergence signal the rows carry; that is what the loop above waits for. Rows come out newest-first within each conclusion.
```
max-sixty, line 118: are we totally sure about this? I would guess it's more likely our code than GH is serving stale data...

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Agreed — rewritten in #1245. The loop is now `plugins/tend-ci-runner/scripts/red_default_branch_runs.py`, the same move #1027 made for Step 4's signals: one converged-read helper for both listings, union by run id for the red rows and newest-wins for the closure read. The sentinel string is gone (Python compares read one against read two, not against a value no page can produce) and so is the `sort -ru` dependence on gojq's key order. Six tests cover it, with a fake `gh` that serves a different page per read of the same URL. The skill keeps a one-line invocation and three sentences of intent.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
tend-agent: Fair challenge — I re-probed it rather than re-citing the old evidence, and it is GitHub.

Twenty reads of the byte-identical URL `repos/openai/codex/actions/runs?branch=main&status=failure&per_page=50` from this session. Nineteen: newest row `2026-09-14T17:36:27Z`, `total_count` 13704, `ETag: W/"c96197dbc10b243ece16d458b824a31…"`. One, interleaved with fresh reads on both sides: newest row `2026-08-15T05:55:10Z`, `total_count` 3229, and a *different* ETag.

The differing ETag is what rules out our side. A cache replaying a stored response replays its validator with it, so two distinct ETags mean the origin built two distinct representations — and `total_count` dropping to 3229 says the snapshot behind the smaller one indexes a fraction of the repo's runs, which is a server-side index rather than a truncated page. Neither response carried an `Age` header, so no intermediary served it. On the client side `gh` caches only under `--cache`, which nothing here passes, and the credential proxy sets `flow.response.stream = True` and never inspects a response body, so there is nothing between `gh` and the API holding a page.

Rate, for what the sweep is worth paying: ~1 read in 20, three listings per run, so a single-read sweep drops the newest rows on roughly one run in seven — and that is the run reporting the branch green.

PR #1244, opened 2026-09-14
skills(running-in-ci): say /tmp is read-only, not just where $TMPDIR points

### LINKED ISSUE #1242: running-in-ci: Temporary Files says where the writable scratch dir is, never that /tmp is read-only
`running-in-ci`'s **Temporary Files** section tells a session where the writable scratch directory is, but never tells it that `/tmp` is not writable:

> Tend sets `$TMPDIR` to `/home/tend-sandbox/tmp`, the writable scratch directory. Shell commands expand `$TMPDIR`; file-writing tools need the absolute path.

Read as a description of one good option rather than the only one, which is how it lands: a session that invents its own scratch path reaches for `/tmp/...` — the habitual default — and takes `Read-only file system`. #1192 fixed the bundled scripts and recipes that hard-coded `/tmp`, and #1208/#1211 fixed `TMPPREFIX`, but neither touches the case where the model writes the path itself, and nothing in any bundled skill states the constraint that would stop it.

Nothing gives the session a reason to expect the failure, either. `/tmp` is mode `drwxrwxrwt` — a permission check looks fine, and only the mount is read-only:

```console
$ ls -ld /tmp
drwxrwxrwt 17 nobody nogroup 4096 Sep 14 07:56 /tmp
$ touch /tmp/probe
touch: cannot touch '/tmp/probe': Read-only file system
```

## Occurrences

Three sessions in `PRQL/prql` across two days, all in model-authored paths:

| Run | Workflow | Command | Outcome |
|---|---|---|---|
| [34744683786](https://github.com/PRQL/prql/actions/runs/34744683786) | `tend-review` | `cat > /tmp/probe.rs` | Retried under `$TMPDIR` 8 s later |
| [34816908713](https://github.com/PRQL/prql/actions/runs/34816908713) | `tend-review` | `cd /tmp && mkdir wtest && …` | Retried under `$TMPDIR` |
| [34815378400](https://github.com/PRQL/prql/actions/runs/34815378400) | `tend-nightly` | `cp prqlc/prqlc/src/cli/watch.rs /tmp/watch.rs.fixed` | See below |

All three recovered, so the cost so far is a wasted turn each. The nightly is the one worth reading, because it shows the shape that does not have to stay cheap. The `cp` was the **backup** step of a back-up → revert-the-fix → confirm-the-test-fails → restore sequence, written as one compound command with no `set -e`:

```bash
cp prqlc/prqlc/src/cli/watch.rs /tmp/watch.rs.fixed
python3 - <<'PY'
… revert the fix in place …
PY
cargo test -q -p prqlc --bin prqlc -- watch 2>&1 | tail -15
cp /tmp/watch.rs.fixed prqlc/prqlc/src/cli/watch.rs
echo "restored"
```

The `cp` failed and the sequence ran on regardless: the revert applied, the test panicked as intended, and the restore — reading the backup that was never written — failed too. That session noticed the stray `cp:` line at the top of a 500-character result whose interesting content was a Rust panic, and re-applied the patch by hand on the next turn. A session that read the panic as the answer it was looking for would have pushed the reverted file.

## Suggested fix

One clause on the existing sentence, so the constraint arrives with the fact it qualifies:

> Tend sets `$TMPDIR` to `/home/tend-sandbox/tmp`, the writable scratch directory; `/tmp` itself is mounted read-only, so a hand-written `/tmp/...` path fails with `Read-only file system` even though its mode says `drwxrwxrwt`. Shell commands expand `$TMPDIR`; file-writing tools need the absolute path.

Wording is yours — the load-bearing part is that the section says `/tmp` does not work, not just that `$TMPDIR` does.

tend-agent:
## Problem

`running-in-ci`'s **Temporary Files** section names `$TMPDIR` as the writable scratch directory but never says `/tmp` is not one, so a session that writes its own scratch path reaches for the habitual `/tmp/...` and takes `Read-only file system`. The checkout it works in is itself under `/tmp` (`prepare_agent_workspace.py` clones into a dedicated container there), so the constraint has to name the writable set rather than the `/tmp` prefix. Nothing in the environment warns it first — `/tmp` is mode `drwxrwxrwt`, and only the mount is read-only, because SRT's write allowlist is the agent workspace, the agent home, and the auto-memory directory ([`sandbox_runtime.mjs`](https://github.com/max-sixty/tend/blob/456d9335e46540ef1321ba173e08f69d82b6048e/shared/steps/sandbox_runtime.mjs#L79)). #1192 moved the bundled scripts and recipes off `/tmp` and #1211 fixed `TMPPREFIX`; neither covers a path the model writes itself.

The failure is usually one wasted turn, but #1242 records a nightly where the dead write was the backup step of a back-up → revert → test → restore sequence run as one compound command with no `set -e`: the backup never existed, the restore silently failed, and the reverted file survived in the worktree.

## Solution

One clause on the existing sentence, so the constraint arrives with the fact it qualifies: writes land in `$TMPDIR` and the checkout, and the `/tmp` container around the checkout does not take them. Plus a sentence naming the compound-command shape that turns the failure from a wasted turn into a bad push.

`test_bundled_runner_guidance_has_no_unscoped_tmp_paths` bans a bare `/tmp` anywhere under `plugins/tend-ci-runner/`, which is why no bundled file could state the constraint. The guard now scans per line and exempts a line that says `/tmp` is read-only — that line is the rule, not an instance of the failure. Every other mention still fails the test; I confirmed by probing it with a scratch file carrying `/tmp/foo.md`.

## Testing

Prose change; no test covers the skill text. Verified the claim from inside this session's sandbox:

```console
$ ls -ld /tmp
drwxrwxrwt 18 nobody nogroup 4096 Sep 14 08:12 /tmp
$ touch /tmp/probe-tend-1242
touch: cannot touch '/tmp/probe-tend-1242': Read-only file system
```

`uv run pytest` passes apart from the two `test_refresh_consumers.py` cases that need DNS, which the sandbox has none of. `pre-commit run --all-files` passes, including the three `repo: local` guards.

Closes #1242 — automated triage


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -34,7 +34,7 @@ This file carries the rules every session needs. Guidance tied to an action most
 
 ## Temporary Files
 
-Tend sets `$TMPDIR` to `/home/tend-sandbox/tmp`, the writable scratch directory. Shell commands expand `$TMPDIR`; file-writing tools need the absolute path.
+Tend sets `$TMPDIR` to `/home/tend-sandbox/tmp`, the writable scratch directory; the sandbox mounts `/tmp` read-only, so a hand-written `/tmp/...` path fails with `Read-only file system` even though its mode reads `drwxrwxrwt`. A compound command without `set -e` runs on past that failure. Shell commands expand `$TMPDIR`; file-writing tools need the absolute path.
```
```suggestion
Tend sets `$TMPDIR` to `/home/tend-sandbox/tmp`, the writable scratch directory; writes land only there and in the checkout, which sits under a `/tmp` container that is itself read-only, so a scratch path written by hand under `/tmp` fails with `Read-only file system` even though the mode reads `drwxrwxrwt`. A compound command without `set -e` runs on past that failure. Shell commands expand `$TMPDIR`; file-writing tools need the absolute path.
```

tend-agent (commented):
The claim as written is false for the session's own working directory, which weakens the paragraph it is meant to strengthen. The sandbox's write allowlist is `allowWrite: [agentWorkspace, agentHome, autoMemory]` (`shared/steps/sandbox_runtime.mjs`), and `prepare_agent_workspace.py` clones `TEND_AGENT_WORKSPACE` into what its module docstring calls a "dedicated `/tmp` container" — so the checkout a session edits all run long is itself a `/tmp/...` path, and it writes fine while its parent container does not. Verified from inside this session: a write to the checkout succeeds, one to the workspace root takes `Read-only file system`. A session reading "the sandbox mounts `/tmp` read-only, so a hand-written `/tmp/...` path fails" hits a counterexample on its first file edit, and the sentence stops being something it can act on.

The suggestion keeps both of the change's points — the token the model habitually reaches for, and the misleading mode — while naming the writable set accurately. It still contains `/tmp`, so it still depends on the guard exemption in `456d933`.

PR #1245, opened 2026-09-14
fix(review-runs): move the live-work red sweep into a tested script

tend-agent:
## Problem

Two review comments on #1243, both on the block that landed there:

- the convergence loop is messy — nested bash around a `gh --jq` pipeline, mixing the two languages;
- and a prior question: is GitHub really serving stale pages, or is it our code?

## Solution

`plugins/tend-ci-runner/scripts/red_default_branch_runs.py` replaces the loop, following the same move #1027 made for Step 4's outcome signals. One converged-read helper serves both listings: union by run id for the red rows, newest-wins for the closure read. The sentinel string is gone — Python compares the first read against the second rather than against a value no page can produce — and so is the dependence on gojq's key order, since the rows are sorted on `created_at` rather than on the text `sort -ru` happens to produce.

The script also emits the scope the section's published claim rests on, instead of leaving the session to reconstruct it: `paths_closed_by_later_green` names the workflows checked, `reached_back_to` how far the listing reached, and `unconverged_listings` names any URL that never settled — a sweep that publishes an unsettled listing as complete is the outward error the whole mechanism exists to prevent.

The skill keeps three sentences of intent and a one-line invocation.

## Is it GitHub or is it us?

It is GitHub. I re-probed `repos/openai/codex/actions/runs?branch=main&status=failure&per_page=50` from this session, twenty reads of the byte-identical URL. Nineteen returned a newest row of `2026-09-14T17:36:27Z` with `total_count` 13704 and `ETag: W/"c96197dbc10b243ece16d458b824a31…"`. One returned a newest row of `2026-08-15T05:55:10Z` with `total_count` 3229 and a *different* ETag — one read out of twenty, interleaved with fresh ones on either side.

The differing ETag is what settles it. A cache replaying a stored response replays its validator too, so two distinct ETags mean the origin built two distinct representations; and `total_count` moving from 13704 to 3229 says the snapshot behind the smaller one indexes a fraction of the repo's runs, which is a server-side index, not a truncated page. No `Age` header on either, so no intermediary served it. The client side is clear too: `gh` caches only under `--cache`, which nothing here passes, and the credential proxy sets `flow.response.stream = True` and never inspects a response body ([`proxy/inject_credentials.py`](https://github.com/max-sixty/tend/blob/main/proxy/inject_credentials.py)), so there is nothing between `gh` and the API that could hold a page.

What that rate means for the sweep: at roughly 1 in 20 per read and three listings per run, a single-read sweep drops the newest rows on something like one run in seven — and the run it happens on is one that reports the branch green.

## Testing

`generator/tests/test_red_default_branch_runs.py` — six tests, fake `gh` serving a different page per read of the same URL, which is what the endpoint does. They pin the regression (a stale first page must not hide the newest failure), that a consistent listing costs exactly two reads, that an empty answer is re-read before it is believed, that an unsettled listing is named rather than published, and both closure cases (a later green closes the path; a `dynamic/dependabot/...` path 404s and stays live).

Ran the script against `max-sixty/tend` live: 12s, closes six workflow paths on their later greens, leaves the CodeQL and Dependabot rows live, `unconverged_listings` empty. Full `uv run pytest` is 1139 passed, with the two `test_refresh_consumers.py` DNS failures the sandbox always has. `pre-commit` passes on the three files.


plugins/tend-ci-runner/scripts/red_default_branch_runs.py
```diff
@@ -0,0 +1,146 @@
...
+            closures[path] = green["created_at"] if green else None
+        closed_at = closures[path]
+        if closed_at and closed_at > row["created_at"]:
+            continue
+        live.append(row)
+
+    github_cli.dump(
+        {
+            "branch": branch,
+            "reached_back_to": min(
```
tend-agent: This minimum runs over `red.values()`, the union of all three listings, so it reports the oldest row the sweep *saw* rather than the oldest point it *covers*. Only a listing truncated at `per_page=50` bounds coverage; a rare conclusion returning its whole history in two rows can drag this value back past the window `failure` actually read.

tend-agent (commented):
One finding, on the scope the sweep publishes rather than on the convergence loop.

`reached_back_to` is the minimum `created_at` across the union of all three conclusion listings, but the three do not cover the same window. Only `failure` is truncated by `per_page=50`: on this repo it returns 50 rows against a `total_count` of 635, reaching back three weeks, while `startup_failure` (1 row) and `timed_out` (0) return their entire history. So the published floor is set by whichever listing is shortest and oldest, not by the one that limits coverage — a single `startup_failure` row from months ago makes the sweep report `reached_back_to` months earlier than the `failure` listing ever read, and the skill hands that value to the session as the scope the "`main` is green" claim rests on. It is the same over-claim the convergence loop exists to prevent, reached from the other side: the rows in the uncovered window are absent from the output, and absent reads as green.

The coverage floor is the newest of the *truncated* listings' oldest rows; an untruncated listing saw everything and constrains nothing. Pushing that, with a test over a listing set whose oldest row and whose truncated listing are different listings.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -91,33 +91,15 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
+  `live` holds the red rows that no later green run of the same workflow closed, newest first. `paths_closed_by_later_green` and `reached_back_to` are the scope the claim rests on: `reached_back_to` is where the first conclusion listing ran out of page, so a row older than it was never read, and `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```
tend-agent: This sentence still carries the over-claim the commit removed from the value. The script bounds coverage by the *truncated* listings only; an untruncated one returned its whole history, and its older rows are published in `live`. `test_coverage_stops_at_the_listing_that_ran_out_of_page` asserts exactly that pair — `reached_back_to` of `2026-09-01` with the June row in `live` — so a session following "a row older than it was never read" writes that into the nightly report while listing one. "first conclusion listing" also reads as first-in-iteration-order rather than the newest floor.

```suggestion
  `live` holds the red rows that no later green run of the same workflow closed, newest first. `paths_closed_by_later_green` and `reached_back_to` are the scope the claim rests on: `reached_back_to` is the newest floor among the listings that filled their page, so a row older than it may be missing — a listing that was not truncated returned its whole history, so `live` can still carry rows older than this. `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```

generator/tests/test_red_default_branch_runs.py
```diff
@@ -0,0 +1,253 @@
...
+    _page(
+        env,
+        "failure",
+        1,
+        *(
+            _red(400 + n, f"2026-09-01T00:{n:02d}:00Z")
+            for n in reversed(range(PER_PAGE))
+        ),
+    )
+    _page(env, "startup_failure", 1, _red(300, "2026-06-01T00:00:00Z"))
```
tend-agent, line 263: With one truncated listing `floors` has a single element, so `max(floors)` and `min(floors)` return the same string and the test passes either way — the newest-floor rule the commit turns on is unpinned. A second truncated listing reaching further back separates them: `min` would report July.

```suggestion
    # A second truncated listing, reaching further back: `reached_back_to` is
    # the newest floor among them, which is what separates `max` from `min`.
    _page(
        env,
        "timed_out",
        1,
        *(
            _red(500 + n, f"2026-07-01T00:{n:02d}:00Z")
            for n in reversed(range(PER_PAGE))
        ),
    )
    _page(env, "startup_failure", 1, _red(300, "2026-06-01T00:00:00Z"))
```

tend-agent (commented):
`max(floors)` is the right rule, and bounding it on the listings that filled their page matches what the endpoint actually constrains. Two things the commit leaves behind.

**The skill's description of `reached_back_to` now contradicts the script.** `SKILL.md` still says the value is where "the first conclusion listing ran out of page, so a row older than it was never read". The script bounds coverage by the *truncated* listings only, and an untruncated one returns its whole history — its older rows are published in `live`. The new test asserts exactly that pair: `reached_back_to` of `2026-09-01` alongside a June row in `live`. A session following that sentence writes "rows older than 2026-09-01 were never read" into the nightly report while listing one, which is the same over-claim moved from the value into the prose that tells the reader how to use it.

**The new test doesn't pin `max` over `min`.** It stages one truncated listing, so `floors` holds a single element and both reductions return the same string — the newest-floor rule the commit turns on passes either way. A second truncated listing reaching further back separates them.

Pushing both fixes.

max-sixty:
@tend-agent, `latest_green()` currently discards the boolean returned by `converged_read()`, so a closure-listing URL that exhausts `MAX_READS` never appears in `unconverged_listings`. Please preserve and propagate that convergence result, include unsettled closure URLs in the output, and add a test where successive latest-green reads never settle.

tend-agent (commented):
`latest_green()` propagates convergence now and the closure URL reaches
`unconverged_listings`, with a test pinning it. Two findings on the surrounding
output, both the same over-claim the convergence loop exists to prevent, reached
from a different side.

**`reached_back_to` takes its floor from the union, not from the settled page.**
`min(row["created_at"] for row in rows)` runs over `converged_read`'s union
across every read, and a stale read answers from a different window than the
settled one — this PR's own probe recorded a stale snapshot whose newest row was
a month behind the fresh one's. Two such pages barely overlap, so the union
reaches back to the stale page's oldest row while the settled listing never read
the month between them. Staged against the script (read 1 a full page from a
July window, reads 2–3 a full page from September), `reached_back_to` comes out
`2026-07-01T00:00:00Z` where the settled listing only ever reached
`2026-09-01T00:00:00Z`. The session then publishes coverage over a gap it never
read, and a failure in that gap is absent from `live` — which reads as green.

The union is what keeps a row a later read stopped returning; the coverage claim
is what the settled page bounds. `len(rows) >= PER_PAGE` has the same problem in
miniature: a union of two under-full pages tests as truncated.

**`paths_closed_by_later_green` names paths that nothing closed.** The
comprehension filters on `if green`, so a path is listed whenever it has any
green at all — including one older than every red row it was read to close.
`test_a_closure_listing_that_never_settles_is_reported_as_such` asserts exactly
that pair: `.github/workflows/ci.yaml` under `paths_closed_by_later_green` while
its run 100 sits in `live`. Closure is per row, not per path, so the map can't
carry that name honestly; `latest_green_by_path` says what it holds and leaves
`live` to say what it failed to close.

Pushing both — `converged_read` returning the settled page beside the union so
truncation and the floor are read from it, and the rename with its `SKILL.md`
sentence.

tend-agent:
Fixed in 9738e03796fdc71eac04ecac10c927e5af6664a9.

`latest_green()` now returns `(green, converged)` and `main()` appends the closure URL to `unconverged_listings` when it never settles. URL construction moved out to `green_url()` so the caller names the same string the read used rather than rebuilding it. A 404 stays settled — there is no listing to converge on a `dynamic/dependabot/...` path — so those rows keep reporting as live without polluting the unconverged list. The newest green seen across the capped reads still closes what it can; the sweep just no longer claims the answer is final.

The test needed the fake `gh` to serve a *moving* per-workflow listing, which it couldn't: the conclusion listings were staged per read (`<status>-<n>.json`) but the closure read always got the one `green-<workflow>.json`. Both now go through one `serve` helper over the same per-read staging, and `test_a_closure_listing_that_never_settles_is_reported_as_such` stages four distinct greens and asserts the URL appears in `unconverged_listings`. It fails on the pre-fix script with exactly the empty list you named.

The branch has since moved to dd2bffd9b457bf7583cb5f4d2a6e94f9801089a7, which bounds `reached_back_to` and the truncation test by the settled page rather than the union and renames `paths_closed_by_later_green` to `latest_green_by_path`. The convergence propagation above is unchanged under it.

tend-agent (commented):
One finding on the sentence the commit rewrote, and two smaller over-claims beside it.

**`SKILL.md` still attributes an older-than-`reached_back_to` row to an untruncated listing alone.** The sentence reads "a row older than it may be missing — a listing that was not truncated returned its whole history, so `live` can still carry rows older than this". After this commit a *truncated* listing does it too: the floor comes from the settled page while `live` keeps the union across every read, so a stale read's rows sit below the floor that the settled page set. `test_the_coverage_floor_comes_from_the_settled_page` asserts exactly that pair — one truncated `failure` listing, `reached_back_to` of `2026-09-01`, and the July page's 50 rows in `live`. A session reading the sentence sees rows under the floor from a listing that *was* truncated, and the only reading the text offers is that the listing was not truncated — the coverage claim inverted. The commit moved the honesty into the value and left the prose that tells the reader how to use it a version behind.

**`latest_green_by_path` is documented as covering more than it holds.** The new clause calls it "the newest green read for each workflow checked", but the comprehension filters `if green`, so a path whose closure read returned nothing is absent — the `dynamic/dependabot/...` 404s that `test_a_path_with_no_workflow_file_stays_live` pins. The same section then asks the session to report "the workflows checked", and those are the ones it most needs to name, because nothing closes them.

**`Listing`'s summary line calls `page` "the page that settled".** On the `MAX_READS` exhaustion path it is the last unsettled answer, which is the page the floor is then taken from; the rest of the docstring says "the last answer alone" and is right.

Pushing all three.

PR #1247, opened 2026-09-15
fix(review-runs): close a generated red run against a later green of the same name

tend-agent:
## Problem

The live-work sweep closes a red default-branch run by asking the per-workflow endpoint for that path's newest green. `green_url` addresses that endpoint by the path's basename, so a `dynamic/…` path — which names no committed file — 404s and `latest_green` returns "no closure". That is right for Dependabot, whose every update carries a one-off id nothing later repeats, and wrong for code scanning, whose analysis carries the same name on every push.

So a CodeQL failure on `main` stays in `live` forever. `34771314444` failed at `2026-09-13T17:20:53Z` and was closed by greens of the same `path` and `name` minutes later, then again on the 14th and the 15th; the sweep still reported it as a live default-branch failure on both of the last two nights. Nothing wrong reached the record, because each session noticed the greens by hand and said so — the 2026-09-14 entry in the tracking issue and this run's both carry that manual closure. That is the re-derivation the sweep's output exists to remove, and the next session that trusts `live` as written publishes "the default branch has an unfixed failure" against a path that passed three times.

## Solution

Address the generated paths' closure listing by the `workflow_id` every run row carries, rather than by the basename that 404s. `generated_greens` reads that workflow's own `status=success` page — through the same `converged_read`, since it is served from the same moving index as every other — and keys the newest green by `name`. The name is what separates the subjects sharing one id: a passing analysis closes a failed one, while each Dependabot update's name is unique, so those rows keep exactly the closure they had, which is none.

Scoping the listing to the workflow is what makes the closure hold up. The repo-wide `status=success` page would have worked today and decayed silently: its first 100 rows span about 20 hours on this repository, 75 of them `tend-notifications`, so the reach of a generated path's closure would depend on how much unrelated traffic ran on `main` — and nothing reports when it stops covering, since `reached_back_to` is fed only by the red listings.

`latest_green` keeps its 404 branch for the case that now reaches it — a committed workflow whose file has left the branch — and regains a test for it, since the generated paths that used to exercise it no longer go through it. The published shape is unchanged: `latest_green_by_path` still maps a path to the newest green found under it, now including the generated ones that had one.

Verified against this repository: the CodeQL row closes on `2026-09-15T04:50:48Z`, and the fifteen Dependabot rows are still live and still unclosed.

<details><summary>Evidence and gate assessment</summary>

Observed in `review-runs` runs [34820024044](https://github.com/max-sixty/tend/actions/runs/34820024044) (2026-09-14) and [34943722786](https://github.com/max-sixty/tend/actions/runs/34943722786) (this run), both recorded in [#1125](https://github.com/max-sixty/tend/issues/1125). The same red row in both, so the two windows are one incident seen twice rather than two independent failures — the recurrence claim rests on the mechanism, not on the count: `green_url` builds a basename URL for every path, and every `dynamic/…` path 404s on it, so no generated run can ever close regardless of how often one fails.

- **Evidence level**: High, structural. Occurrences: 2 windows, 1 row.
- **Cost class**: No wrong outward action has occurred — both sessions caught the false positive and published the correction rather than the row. The cost so far is the hand-closing itself, and the exposure is a future session that reads `live` as the skill describes it.
- **Change type**: Targeted fix inside an already-tested script; no new prose for a session to load, and one clause of the skill's description of `live` updated to match.
- **Tests**: four new cases — the closure, the direction (a green *older* than the failure closes nothing), the unsettled-listing report, and the restored 404 case for a committed workflow whose file is gone.

</details>


plugins/tend-ci-runner/scripts/red_default_branch_runs.py
```diff
@@ -119,6 +126,36 @@ def latest_green(
     return green, listing.converged
 
 
+def generated_green_url(repo: str, branch: str) -> str:
```
tend-agent: Scope this at the generated workflow rather than the whole branch: every run row carries `workflow_id`, and `repos/{repo}/actions/workflows/{workflow_id}/runs` serves the generated ones (the 404 comes from `green_url` passing the path basename, which is not a filename). Then the page budget covers that workflow's own greens instead of whatever else happened to run on `main`.

tend-agent (commented):
Self-review. The closure change is right — `(path, name)` is the subject a
generated run repeats, and keeping Dependabot's one-off names unclosed follows
from it. Two findings on how the generated closure is read.

**The closure window is the repo-wide success listing, so unrelated traffic
sets how far back it reaches.** Measured on this repository right now, the
first 100 rows of `?branch=main&status=success` span 2026-09-14T11:32:26Z to
2026-09-15T08:03:31Z — about 20 hours — and 75 of them are `tend-notifications`
runs. Code scanning happens to fit (7 rows), so the fix works today. Dependabot
does not: its newest green on `main` is 2026-09-06, nine days outside the
window, so had its names recurred the closure would already be missing. Add one
more frequent scheduled workflow and the window shrinks past code scanning too,
and the row silently goes back to reading live forever — the bug this PR fixes,
reinstated with no signal, because `reached_back_to` is fed only by the red
listings and `generated_greens` reports no floor of its own.

The per-workflow endpoint does serve these runs; it 404s only because
`green_url` derives the endpoint's id from the path basename (`codeql`), which
is not a committed filename. Every run row carries `workflow_id`, and
addressing the endpoint by it works for both generated workflows:

```
repos/max-sixty/tend/actions/workflows/348682278/runs?branch=main&status=success
  -> total_count 90, newest "Push on main" at 2026-09-15T04:50:48Z
repos/max-sixty/tend/actions/workflows/348683058/runs?branch=main&status=success
  -> total_count 8, the Dependabot updates
```

So the same page budget can be scoped to the generated workflow instead of
competing with every other run on the branch. The `(path, name)` keying stays
as-is — Dependabot's updates share one `workflow_id`, so a workflow-scoped
lookup alone would close a red update against an unrelated green one.

**The `latest_green` 404 branch lost its only test.** The Dependabot case was
what drove a red row through that `except`; this PR routes generated paths away
from `latest_green`, and every remaining committed-path test stages a green.
The branch the rewritten docstring now claims it exists for — a committed
workflow whose file has left the branch — is unexercised, so a regression in
the `"HTTP 404" in error.stderr` check would pass the suite and surface as the
whole sweep exiting non-zero.

Pushing both fixes to this branch.

generator/tests/test_red_default_branch_runs.py
```diff
@@ -322,10 +336,116 @@ def test_a_later_green_closes_the_path(env: dict[str, str]) -> None:
 
 def test_a_path_with_no_workflow_file_stays_live(env: dict[str, str]) -> None:
     """Dependabot's `dynamic/...` paths are not committed files, so the
-    per-workflow endpoint 404s: those rows have no closure and stay live."""
-    _page(env, "failure", 1, _red(300, "2026-09-01T00:00:00Z", path=DEPENDABOT))
+    per-workflow endpoint 404s. Its updates never repeat a name, so the
```
tend-agent: This was the only test that reached `latest_green`'s 404 branch, and the generated paths no longer go through it. A committed path whose workflow file is gone — red row staged, no green — needs its own case, or that `except` is untested.

tend-agent (commented):
Self-review of the scoping commit: two fixes I am pushing to this branch, and one bound that is a decision rather than an oversight.

**The generated closure read lost the 404 guard, so a 404 now loses the whole sweep.** Before this PR every `dynamic/...` row went through `latest_green`, whose `except subprocess.CalledProcessError` treats `HTTP 404` as a settled "no closure". `generated_greens` calls `converged_read` unguarded, so the same 404 — a `workflow_id` that no longer resolves, or a row carrying none — propagates out of `main` and the script's top-level handler exits non-zero with no report at all. Checked by running the new test against the unguarded version: the sweep exits 1 and prints nothing, where the old code left the row live and still published the rest of the branch.

**Committed rows are keyed by `(path, name)`, so one path is re-read once per run name.** A committed path's closure ignores the name — `latest_green` answers the same for every one — but the subject key makes each name a fresh cache miss, and `run-name:` gives a run a name of its own: on a repository that sets one, the run object's `name` is the run-name, and rows from before it was added carry the workflow name instead, so a single path legitimately answers under several names. With up to 150 red rows across the three conclusions that is up to 150 `converged_read`s of one URL where the previous code did one, plus a repeated entry in `unconverged_listings` when that URL does not settle. Keying the name only where it separates subjects — the generated rows — keeps the closure and drops the re-reads.

**`GREEN_PAGE = 100` bounds how far a generated closure reaches, and nothing reports when it stops covering.** This is the residual of the window problem that scoping the listing fixed: this repository's code-scanning workflow answers `status=success` with 90 greens spanning 2026-09-02 to 2026-09-15, so one page is roughly two weeks, while the red listings reach back further and `reached_back_to` is fed only by them. A generated failure older than its workflow's last 100 greens reads as live forever — the same shape as the bug this PR fixes, at a longer horizon. Reporting a second floor would cost a new output field and the skill text to explain it, for a case nothing has hit, so I have left the bound in place: worth knowing it is a bound and not a guarantee.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -95,11 +95,11 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
+  `live` holds the red rows that no later green run of the same subject — the same workflow, or for a generated run the same `path` and `name` — closed, newest first. `latest_green_by_path` and `reached_back_to` are the scope the claim rests on: `latest_green_by_path` is the newest green read for each workflow that had one, which closes that workflow's older red rows and leaves any newer one in `live`; a workflow with no green at all is absent from it, so name the workflows checked from `live`'s paths too. `reached_back_to` is the newest floor among the listings that filled their page, so a row older than it may be missing — `live` can still carry rows older than it, both from an untruncated listing's whole history and from what an earlier read of a truncated one returned from an older window. `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```
```suggestion
  `live` holds the red rows that no later green run of the same subject — the same workflow, or for a generated run the same `path` and `name` — closed, newest first. `latest_green_by_path` and `reached_back_to` are the scope the claim rests on: `latest_green_by_path` is the newest green read under each path that had one, which closes that path's older red rows *of the same subject* and leaves any newer one in `live` — so on a generated path carrying several subjects, a row older than the published green can still be live, because that green closed another name; a path with no green at all is absent from it, so name the workflows checked from `live`'s paths too. `reached_back_to` is the newest floor among the listings that filled their page, so a row older than it may be missing — `live` can still carry rows older than it, both from an untruncated listing's whole history and from what an earlier read of a truncated one returned from an older window. `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```

tend-agent (commented):
Self-review of the 404 guard and the single-read keying: both hold, and each has
the test it needs. One finding on what the sweep publishes beside `live`.

**`latest_green_by_path` is now a per-path reduction of a per-subject closure,
and the skill still describes it as per-workflow.** A generated path can carry
several subjects — this repository's code-scanning workflow answers
`branch=main` under four names (`Push on main` ×87, `Scheduled` ×2, `CodeQL`,
`CodeQL Setup`) — so the green published for a path can be the one that closed a
*different* name than the row left live. Staged through the test harness: a
`Push on main` failure closed by a green at `2026-09-15`, alongside a
`Scheduled` failure at `2026-09-14` that nothing closes, publishes

```
live: [(501, 'Scheduled', '2026-09-14T00:00:00Z')]
latest_green_by_path: {'dynamic/github-code-scanning/codeql': '2026-09-15T00:00:00Z'}
```

The skill tells the session that `latest_green_by_path` "closes that workflow's
older red rows and leaves any newer one in `live`". Read against that output,
the live row is *older* than its own path's green, so a session applying the
rule as written either reports the sweep as self-contradictory or hand-closes a
default-branch failure that is genuinely live. The second is the same wrong
outward claim this PR removes from `live`, reappearing in the evidence beside
it.

The code is right; the sentence describing it is what the subject keying didn't
reach. Suggestion inline, and I'm pushing it to this branch with a case pinning
the two-subject output.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -95,11 +95,11 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
+  `live` holds the red rows that no later green run of the same subject — the same workflow, or for a generated run the same `path` and `name` — closed, newest first. `latest_green_by_path` and `reached_back_to` are the scope the claim rests on: `latest_green_by_path` is the newest green read under each path that had one, which closes that path's older red rows *of the same subject* and leaves any newer one in `live` — so on a generated path carrying several subjects, a row older than the published green can still be live, because that green closed another name; a path with no green at all is absent from it, so name the workflows checked from `live`'s paths too. `reached_back_to` is the newest floor among the listings that filled their page, so a row older than it may be missing — `live` can still carry rows older than it, both from an untruncated listing's whole history and from what an earlier read of a truncated one returned from an older window. `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```
```suggestion
  `live` holds the red rows that no later green run of the same subject — the same workflow, or for a generated run the same `path` and `name` — closed, newest first. `latest_green_by_path` and `reached_back_to` are the scope the claim rests on: `latest_green_by_path` is the newest green that closed a red subject under each path, which closes that path's older red rows *of the same subject* and leaves any newer one in `live` — so on a generated path carrying several subjects, a row older than the published green can still be live, because that green closed another name; a path is absent from it when no green closed any of its red subjects, which is weaker than the workflow never passing — Dependabot's path is always absent and its workflow does pass — so name the workflows checked from `live`'s paths too. `reached_back_to` is the newest floor among the listings that filled their page, so a row older than it may be missing — `live` can still carry rows older than it, both from an untruncated listing's whole history and from what an earlier read of a truncated one returned from an older window. `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```

tend-agent (commented):
Self-review of the subject-scoping doc fix. The new clause is right, and the
staged two-subject case pins it. One clause of the same sentence still reads
per-workflow.

**"a path with no green at all is absent from it" no longer describes when a
path is absent.** `green_by_path` reduces `closures`, which holds only subjects
that carried a red row, so a path is absent whenever no green closed any of its
red subjects — however many greens the sweep read under it. Before this commit
that clause was approximately true for the generated paths, because
`latest_green` 404'd and read no greens at all; now `generated_greens` reads
them and still publishes nothing for the path. Dependabot is the standing case:
its fifteen live rows each carry a one-off name so none of them closes, the path
is absent, and
`repos/max-sixty/tend/actions/workflows/348683058/runs?branch=main&status=success`
answers with 8 greens. `test_a_generated_update_that_never_repeats_its_name_stays_live`
already stages that shape and asserts `latest_green_by_path == {}`.

The consequence lands in the scope statement the session publishes from this
field: read as written, absence licenses "that workflow has no green on `main`"
about a workflow that passed eight times — the same per-workflow reading of a
per-subject field this commit removed from the preceding clause.

Suggestion inline, and I'm pushing it to this branch.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -95,11 +95,11 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
...ame `path` and `name` — closed, newest first. `latest_green_by_path` and `reached_back_to` are the scope the claim rests on: `latest_green_by_path` is the newest green that closed a red subject under each path, which closes that path's older red rows *of the same subject* and leaves any newer one in `live` — so on a generated path carrying several subjects, a row older than the published green can still be live, because that green closed another name; a path is absent from it when no green closed any of its red subjects, which is weaker than the workflow never passing — Dependabot's path is always absent and its workflow does pass — so name the workflows checked from `live`'s paths too. `reached_back_to` is the newest floor among the listings that filled their page, so a row older than it may be missing — `live` can still carry rows older than it, both from an untruncated listing's whole history and from what an earlier read of a truncated one returned from an older window. `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```
```suggestion
  `live` holds the red rows that no later green run of the same subject — the same workflow, or for a generated run the same `path` and `name` — closed, newest first. `latest_green_by_path` and `reached_back_to` are the scope the claim rests on: `latest_green_by_path` is closure evidence rather than a set of fixed paths — per path, the newest green read under any subject that carried a red row, whether or not it closed one — so a path can be published with a green that closed nothing, and on a generated path carrying several subjects a row *older* than the published green can still be live because that green closed another name; a path is absent when no green was read under any of its red subjects, which is weaker than the workflow never passing — Dependabot's path is always absent and its workflow does pass — so name the workflows checked from `live`'s paths too. `reached_back_to` is the newest floor among the listings that filled their page, so a row older than it may be missing — `live` can still carry rows older than it, both from an untruncated listing's whole history and from what an earlier read of a truncated one returned from an older window. `null` means no listing was truncated and the sweep covers the branch's whole history. A non-empty `unconverged_listings` means a listing never settled — report that rather than publishing the sweep as complete.
```

tend-agent (commented):
Self-review of the absence clause. It says what it set out to — absence is
weaker than "never passed", and Dependabot pins it — but the same rewrite
re-characterized the field's *presence* as a closure.

**`latest_green_by_path` publishes greens that closed nothing, so "the newest
green that closed a red subject under each path" overstates it.** `green_by_path`
reduces `closures`, whose value for a subject is the newest green *read* for it,
closed or not; the dump's own comment beside the field says as much ("The closure
evidence, not a closed set"). `test_a_closure_listing_that_never_settles_is_reported_as_such`
pins that output: the red row at `2026-09-10` is live and
`latest_green_by_path == {'.github/workflows/ci.yaml': '2026-09-04T00:00:00Z'}`,
a green older than the failure, so it closed nothing. The absence clause
inherits it — "a path is absent from it when no green closed any of its red
subjects" is false of that same case, where no green closed anything and the
path is present.

The consequence is in the scope statement the session publishes from this field:
read as written, presence licenses "a failure on this path was closed at T" about
a path whose only red row is still live and whose green predates it — history the
field does not carry. Describing it as evidence read rather than closure achieved
covers both clauses, and drops the restatement of `live`'s rule that the sentence
had picked up.

Suggestion inline, and I'm pushing it to this branch.

plugins/tend-ci-runner/scripts/red_default_branch_runs.py
```diff
@@ -149,20 +200,43 @@ def main(argv: list[str] | None = None) -> int:
...
+                green, converged = latest_green(repo, branch, row["path"])
+                if not converged:
+                    unconverged.append(green_url(repo, branch, row["path"]))
+                closures[subject] = green["created_at"] if green else None
+        closed_at = closures[subject]
         if closed_at and closed_at > row["created_at"]:
             continue
         live.append(row)
 
+    # Published per path: the newest green any of that path's subjects closed on.
```
tend-agent:
```suggestion
    # Published per path: the newest green read for any of its red subjects,
    # whether or not it closed one.
```

tend-agent (commented):
Self-review of the evidence-read rewrite. The new sentence holds: `closures` keys a generated row by `(path, name)` and a committed one by path alone, `green_by_path` reduces those values per path, and both clauses — "the newest green read under any subject that carried a red row" and "absent when no green was read under any of its red subjects" — are what the code does. `test_a_closure_listing_that_never_settles_is_reported_as_such` and `test_one_subject_of_a_generated_path_closes_without_closing_the_others` pin the two cases it now covers. One finding, on the same overstatement one layer down.

**The comment this PR introduces over `green_by_path` describes the field as a closure, which is the reading the commit above just removed from the skill.** It calls the value "the newest green any of that path's subjects closed on"; the value is the newest green *read* for a subject that carried a red row, whether or not it closed one. Both of the field's other descriptions in the tree say so — the dump's own comment twelve lines below (`The closure evidence, not a closed set`), and, from before this PR, `test_a_closure_listing_that_never_settles_is_reported_as_such`, whose comment reads "the map is the closure evidence read, not a set of paths something closed" and whose assertions are `latest_green_by_path == {'.github/workflows/ci.yaml': '2026-09-04T00:00:00Z'}` beside a live red row from the 10th. So the loop that builds the field now carries the one description of it that is wrong, and a reader who starts at the code rather than at the dump gets the same "a failure on this path was closed at T" licence the skill sentence stopped granting.

Suggestion inline, and I'm pushing it to this branch.

PR #1248, opened 2026-09-15
fix(enrich-outage): skip runs with no failed attempt

### LINKED ISSUE #1246: enrich-tend-outage-issues: a green run cited as a diagnosis baseline is enriched as "No failure details could be extracted"
`enrich_tend_outage_issues.py` enriches every `/actions/runs/<id>` link it finds in a `tend-outage` issue, with no check that the run failed. A `ci-fix` diagnosis that cites green runs as comparison evidence — which is the normal shape of a good diagnosis — therefore gets those green runs written up publicly as "No failure details could be extracted.", and the tracker reads as if they had failed.

## Observed

[max-sixty/leaf#705](https://github.com/max-sixty/leaf/issues/705) is a `ci-fix` diagnosis of one `publish-site` failure (run `34820889246`). Its body cites three earlier deploys as the healthy baseline — "the last three green deploys published at 9.3 s, 17.6 s and 11.6 s" — each as a run link.

The nightly's enrichment pass then posted [this comment](https://github.com/max-sixty/leaf/issues/705#issuecomment-5675996653), which renders **four** `### Run …` sections. One carries the real log tail. The other three are the green baselines, each rendered as:

```
### [Run 34793155710](…)

No failure details could be extracted.
```

All three are single-attempt, fully green:

| Run | `run_attempt` | failed jobs across all attempts |
|---|---|---|
| `34793155710` | 1 | none |
| `34814515951` | 1 | none |
| `34818439820` | 1 | none |

For contrast, the three runs enriched onto [max-sixty/leaf#658](https://github.com/max-sixty/leaf/issues/658) all report `conclusion: success` at the run level but each has `deploy-site` failing on attempt 1, and each got a real log tail. So `filter=all` (from #1151) is working — the run-level conclusion genuinely is not the test. The missing test is whether *any* attempt has a failed job.

## Mechanism

[`pending_run_ids`](https://github.com/max-sixty/tend/blob/eb0e95e9162b9bb8b541d99f1bf457db385ab5a3/plugins/tend-ci-runner/scripts/enrich_tend_outage_issues.py#L37-L50) collects run IDs purely by regex over the issue body and comments. Nothing downstream reconsiders: [`failure_details`](https://github.com/max-sixty/tend/blob/eb0e95e9162b9bb8b541d99f1bf457db385ab5a3/plugins/tend-ci-runner/scripts/enrich_tend_outage_issues.py#L177-L182) returns `[]` for a run with no failed job, and [`render_run`](https://github.com/max-sixty/tend/blob/eb0e95e9162b9bb8b541d99f1bf457db385ab5a3/plugins/tend-ci-runner/scripts/enrich_tend_outage_issues.py#L185-L198) turns an empty detail list into the "could not extract" sentence — a message meant for a run whose logs have aged out, not for a run that never failed.

The two cases need different answers, and only one of them should reach the thread. A genuinely failed run with no extractable detail should still say so and still take its `<!-- enriched-run: -->` marker, so it stops being retried. A run with no failed attempt should produce no section and no marker at all — the marker would be wrong, since a later rerun of that run could fail and would then deserve enrichment.

## Proposed fix

Make `failure_details` distinguish the two, skip in `render_batch`, and don't post an empty batch:

```python
def failure_details(repo: str, run_id: str) -> list[str] | None:
    """Detail for a run that failed; ``None`` when no attempt failed at all.

    A diagnosis routinely cites green runs as its baseline. Those are not
    failures to annotate, and the empty-detail sentence in ``render_run``
    means "the logs are gone", which would be false about them.
    """
    jobs = run_jobs(repo, run_id)
    if jobs and not any(job.get("conclusion") == "failure" for job in jobs):
        return None
    return annotation_details(repo, jobs) or log_details(
        repo, run_id, failed_attempt(jobs)
    )
```

```python
# render_batch
        details = failure_details(repo, run_id)
        if details is None:
            continue
        section = render_run(repo, run_id, details)
```

```python
# main
        body = render_batch(repo, run_ids)
        if not body.strip():
            continue
```

The `jobs and` guard keeps today's behaviour when `run_jobs` returns `[]` because the API call failed — that is not evidence the run was green.

Cost of not marking a skipped run: one `repos/…/actions/runs/<id>/jobs?filter=all` call per cited green run per nightly. `render_batch` already makes that call for every pending run, so this adds nothing to a run that would have been enriched anyway, and it is what keeps a later rerun-to-red still enrichable.

## Why it matters

The tracker is what a maintainer reads to decide whether a failure is durable or transient. On leaf#705, the enrichment turned one well-diagnosed failure into a page showing four runs, three of them apparently undiagnosable — which both overstates the incident and buries the one real log tail. The better the `ci-fix` diagnosis, the more baseline runs it cites, and the worse this gets.

tend-agent:
## Problem

`enrich_tend_outage_issues.py` collects run IDs by regex over the issue body and comments, and nothing downstream asks whether those runs failed. A `ci-fix` diagnosis routinely cites green runs as its comparison baseline, so each of those got its own `### Run …` section reading "No failure details could be extracted." — a sentence that means the logs aged out, not that the run passed. The tracker then reads as if several undiagnosable failures had occurred, and the one real log tail is buried among them.

## Solution

`failure_details` now returns `None` when the jobs read came back with rows and none of them is unsuccessful. `render_batch` skips those runs entirely — no section and no `<!-- enriched-run: -->` marker, so a later rerun of that run to red is still enrichable — and `main` posts nothing when a batch renders empty.

"Unsuccessful" is `failure`, `cancelled`, or `timed_out`: `tend-ci-fix` fires on cancelled runs as well as failed ones, so its tracker cites runs whose jobs never carry `conclusion: failure`, and reading those as green would drop the diagnosis. A conclusion still absent is deliberately excluded — an attempt that has not finished is left unmarked for a later nightly rather than spending a durable marker on it.

Those runs then need a diagnosis to render. `gh run view --log-failed` selects failed *steps*, which a cancelled job has none of, so the annotation naming the cancelling request or the exceeded limit is all there is. `annotation_details` reports the failed jobs wherever any failed — a fail-fast matrix reports its one real error rather than every sibling's `The operation was canceled.` — and otherwise every unsuccessful job, rendering each distinct message once. Cancellation and timeout are run-level events that annotate every job identically, and repeating that line would fill the run's byte budget; deduping the rendered message rather than picking one job per conclusion is what keeps a job cancelled before it started — which carries no annotation at all — from standing in for the sibling that holds the diagnosis.

An unreadable jobs response yields no rows at all, which is not evidence the run was green; the `jobs and` guard keeps those on the existing path, still falling back to the latest attempt's log.

Because the green guard reads a list with no unsuccessful row as a run that never failed, the jobs read has to return every row. `json_call` set no `per_page`, so GitHub capped it at 30, and even `per_page=100` caps below the 256 jobs a matrix can create before rerun attempts stack on top — enough for a failed row to land past the cap and the run to be dropped with no section and no marker. `run_jobs` now reads the endpoint through `gh api --paginate` and concatenates every page's `jobs`.

## Testing

Eleven tests in `generator/tests/test_enrich_tend_outage_issues.py`: a green baseline cited alongside a real failure is omitted while the failure is still enriched, an issue citing only green runs posts nothing and never fetches a log, an unreadable jobs response still renders and marks the run, a run whose attempts are still going is left unmarked, the jobs read goes through `--paginate`, a failed job on the second page is enriched, a cancelled run is enriched from its annotation, a cancelled matrix renders that annotation once, a never-started cancelled job's sibling still carries the run, a `timed_out` job survives its cancelled siblings, and a fail-fast matrix reports only the leg that failed. All but the unreadable-response one fail before their respective change. The fake `gh` now serves per-run jobs responses — optionally as the multi-document page stream `--paginate` emits — and per-job annotations, so one batch can mix green and failed runs and one run can mix annotated and never-started jobs.

Closes #1246 — automated triage


tend-agent (commented):
`run_jobs` fetches `.../jobs?filter=all` through `github_cli.json_call`, which sets no `per_page`, so GitHub returns only the first 30 rows. The new guard reads that partial list as the whole truth: when a cited run has more than 30 job rows across its attempts and none of the first 30 carries `conclusion == "failure"`, `failure_details` returns `None`, `render_batch` skips the run, and nothing is posted and no marker is written — so the diagnosis never reaches the tracker and no later nightly recovers it. Before this change the same run fell through to `log_details(repo, run_id, failed_attempt(jobs))` with `attempt=None`, where `gh run view --log-failed` resolves the failing job itself and posted a real log tail. A `ci-fix` tracker cites the watched workflow's run, and adopter CI matrices past 30 legs are ordinary; `filter=all` stacking attempts reaches the page boundary sooner still.

Gating on the run's own `conclusion` instead of the jobs' would not fix it — that field reads `success` after a rerun-to-green, which is the attempt `run_jobs`' docstring says `filter=all` exists to keep. Widening the page is the fix, so I'll push `&per_page=100` on the jobs read and a test asserting it.

max-sixty:
@tend-agent `run_jobs` requests `per_page=100` but neither paginates nor checks `total_count`. GitHub caps this endpoint at 100 rows, while a workflow matrix can create 256 jobs before accumulated rerun attempts. If the first page is nonempty and its failed job lies on a later page, `failure_details` returns `None`, silently skips the genuinely failed run, and writes no marker. Please paginate every page, or treat a truncated response as unreadable and preserve the existing fallback.

tend-agent:
Right — 100 is still a cap, and a matrix run can exceed it before rerun attempts stack on top. Fixed in 2495efd: `run_jobs` now reads the endpoint through `gh api --paginate` (via `github_cli.json_stream`) and concatenates each page's `jobs`, so the guard only ever sees the complete row set. An unreadable response still yields no rows and keeps the existing log fallback.

New test `test_a_failure_on_a_later_jobs_page_is_enriched` serves the fake `gh` a two-document page stream with the failed job on the second page and asserts the run is still enriched and marked; it fails before the change, where the extra document makes the single-response parse raise and the run falls through to the log tail.

tend-agent (commented):
`failure_details`'s new guard reads a run as green when no job carries `conclusion == "failure"`, but a failed run isn't the only kind the tracker cites. `tend-ci-fix` fires on cancelled runs too — its job `if:` gates on `contains(fromJSON('["failure", "cancelled"]'), github.event.workflow_run.conclusion)` — and a cancelled run's jobs conclude `cancelled`, never `failure`; the same holds for `timed_out`, which `red_default_branch_runs.py` already counts among `RED_CONCLUSIONS`. So a run whose jobs are all `cancelled` or `timed_out` now returns `None`, `render_batch` skips it, and no section and no marker are written — and where it's the only run an issue cites, `main` posts nothing at all. Nothing records that the run was seen, so no later nightly recovers the diagnosis. That is the same silent drop the pagination fix in this PR exists to prevent, reached through the conclusion value instead of the page boundary.

Confirmed against the test harness: serving the cited run a single `cancelled` job renders an empty body on this head, and renders the log tail plus its marker once the guard is widened.

The fix follows in this session — the guard matches a module-level `UNSUCCESSFUL` set of `failure`, `cancelled`, `timed_out`. An absent conclusion (`None`) stays out of that set deliberately: an attempt still in flight should be left unmarked so a later nightly enriches it rather than spending a durable marker on a run that hasn't finished. That behavior is new in this PR and gets its own regression test alongside the cancelled-run one. `failed_attempt` needs no change — it wants the attempt whose log holds the error, and a cancelled attempt has none, so an all-cancelled run falls back to `--log-failed` on the latest attempt as before.

tend-agent (commented):
Widening the guard to `cancelled` and `timed_out` keeps those runs out of the silent-drop path, but the detail path they now reach has nothing to give them: `annotation_details` still filters `conclusion == "failure"`, and `gh run view --log-failed` selects failed *steps*, which a cancelled job has none of. So a cancelled run renders `No failure details could be extracted.` and takes a durable `<!-- enriched-run: -->` marker — the same sentence and the same permanence this PR exists to stop a green baseline from getting, reached through the branch the last commit added.

Checked against real runs rather than the fixture, whose fake `gh` serves a log tail for any `--log-failed` call: on [run 34772565174](https://github.com/max-sixty/tend/actions/runs/34772565174) (one `cancelled` job) and [run 34067831347](https://github.com/max-sixty/tend/actions/runs/34067831347) (`success`/`skipped`/`cancelled`), `--log-failed` returns empty with exit 0, while the cancelled job's annotations endpoint carries a `failure`-level `Canceling since a higher priority waiting request for … exists`. That annotation is the run's whole diagnosis, and the marker makes dropping it permanent.

Pushed as a follow-up commit: an `annotated_jobs` helper returns the failed jobs wherever any failed, and otherwise the rest of `UNSUCCESSFUL`. Keeping failures first is the part that matters — a fail-fast matrix cancels the failed job's siblings, so annotating every unsuccessful row unconditionally would bury the one real error under N copies of `The operation was canceled.` and spend `MAX_RUN_BYTES` on them. `failed_attempt` stays on `failure` alone: it picks the attempt whose log holds the error, and widening it would let a later cancelled attempt outrank an earlier failed one and blank the log tail.

Two tests: the cancelled-run case now serves an empty `--log-failed` the way the real CLI does and asserts the annotation is what carries the section (it fails on `b6ff1ed`, rendering the `No failure details` sentence), and a fail-fast matrix asserts only the failed leg is reported.

tend-agent (commented):
`annotated_jobs`' fallback returns every unsuccessful job, and a cancellation annotates all of them identically, so a run that only cancelled renders one copy of the same section per job. On [run 34065429908](https://github.com/max-sixty/tend/actions/runs/34065429908) all four `Analyze (…)` jobs carry `The run was canceled by @github-advanced-security[bot].`; fed those rows, this head renders four identical `#### …` blocks, and a wider matrix repeats the line until `MAX_RUN_BYTES` cuts it off — half of `MAX_BATCH_BYTES` spent on the repetition, pushing the issue's other cited runs into a later batch. That is the noise this commit's own rationale gives for keeping failures first, reached through the fallback branch instead of the fail-fast one, and it is new here: before this PR a job that didn't fail was never annotated at all.

Fix follows in this session — the fallback keeps one job per conclusion rather than every job. Both conclusions stay, because a timeout cancels its siblings, so the job naming the exceeded limit would otherwise be hidden behind a cancelled row. The failed-job path is untouched, so a matrix of genuinely failed legs still reports each leg.

Two tests: a four-job cancelled run asserts its annotation renders once, and a `timed_out` job among cancelled siblings asserts it survives while the repeated cancelled rows do not. Both fail on `9fc87a0`.

tend-agent (commented):
`annotated_jobs`' fallback keeps one job per conclusion, and the dict comprehension keeps the *last* of them — which on a real cancelled run is routinely the job that was cancelled before it started and carries no annotation at all. The annotated sibling is dropped, `annotation_details` returns nothing, `--log-failed` has nothing to give a cancelled job, and the run renders `No failure details could be extracted.` and takes its durable `<!-- enriched-run: -->` marker. That is the outcome this PR exists to prevent, now reachable on a run whose diagnosis was sitting right there.

Checked against a real run rather than the fixture, whose fake `gh` serves the same annotations for every job: [max-sixty/leaf run 34926635256](https://github.com/max-sixty/leaf/actions/runs/34926635256) has two cancelled jobs — `lint` (started, annotated `Canceling since a higher priority waiting request for ci-pull_request-720 exists`) and `test` (`steps: 0`, never started, no failure annotations). Fed those rows, this head renders the `No failure details` sentence plus the marker; `9fc87a0` rendered `lint`'s annotation.

Fix follows in this session — the dedup moves from the job selection to the rendered message. The fallback goes back to every unsuccessful job, and `annotation_details` skips one whose message block it has already rendered. A run-level cancellation still reports once, a job with no annotation now falls through to the sibling that has one instead of standing in for it, and deduping on content cannot drop a distinct diagnosis: the timed-out job among cancelled siblings survives because its message names the exceeded limit, not because its conclusion was grouped separately. `annotated_jobs` folds back into `annotation_details`, which is where the `seen` set has to live, and the failed-job path is excluded from the dedup so a matrix of legs failing the same assertion still reports each leg.

The fake `gh` now serves per-job annotations, which the old fixture could not express. `test_a_timed_out_job_survives_its_cancelled_siblings` gave every job identical annotations, so it could not distinguish the selection it was pinning; it now gives the cancelled siblings `The operation was canceled.` and the timed-out job the limit message. A new test serves a never-started cancelled sibling no annotations and asserts the run is still enriched from its sibling — it fails on `4f6b891` with exactly the `No failure details` body above.

tend-agent:
CI on `75dc731`: `lint`, `test`, `test-sandbox`, `test-proxy`, `test-worker` and `test-codex-surface` all pass. The two red checks are GitHub's default-setup CodeQL — [`Analyze (python)`](https://github.com/max-sixty/tend/actions/runs/34955101232/job/104335177208) and [`Analyze (javascript-typescript)`](https://github.com/max-sixty/tend/actions/runs/34955101232/job/104335177427) — which both failed inside `Perform CodeQL Analysis` right after `Uploading results`, with no error line and no annotation on either job; `Analyze (actions)` in the same run passed. Every other CodeQL run today on this branch and on `main` succeeded, including the two earlier heads of this PR carrying the same Python file, so this is the SARIF upload rather than the change. The run is a default-setup one and reports `This workflow run cannot be retried`, so it clears on the next push rather than on a rerun.

PR #1252, opened 2026-09-15
fix(sandbox): seed the agent's Git commit identity before it starts

tend-agent:
## Problem

Sessions that commit hit `Author identity unknown` on their first `git commit`: the disposable checkout carries no identity, and neither does a clone a session makes for itself. Every session recovered by configuring one and re-committing, and five bundled recipes plus this repo's integration-test recipe carried instructions for doing it.

## Change

`setup_sandbox.py` seeds the sandbox user's global Git config with the bot's noreply identity (`<id>+<login>@users.noreply.github.com`), next to the `core.excludesFile` it already writes there. Both harness actions pass the login and id the rate-limit preflight already resolves from the token, and the step fails with a named error if either is missing. Global rather than checkout-local so that a clone the agent makes itself is covered too.

With the identity in place before the agent starts, the guidance goes: the `pr-creation.md` section and the pointers to it from `review-runs`, `skill-pr-workflow.md` and `resolve-conflicts`, the `git config` lines in the integration-test recipe, and the identity setup in `nightly_workflow_update.py`'s ship step.

This replaces the change this PR opened with — a further pointer to that guidance from the review skill's push recipe — per the review comment on it.

## Verification

`uv run pytest` passes (1153 passed; the two `test_refresh_consumers.py` failures are the documented sandbox-DNS ones) and every `pre-commit` hook passes. A new unit test asserts what the global config gets; `proxy/test-setup-sandbox.sh`, the hosted-runner job, reads `user.name` and `user.email` back as the sandbox user, which is the half that exercises the real uid boundary.

Adopters pick this up at the next release: the action ref and the bundled skills both come from the release tag, so the seeding and the guidance removal land together rather than leaving a window where neither configures an identity.


max-sixty:
this is the wrong place for the change. 

instead, consider adding this as code before the agent starts, and then removing any guidance about the agent needing to set it

tend-agent:
Done — reworked as the code-side fix and force-pushed over the old commit.

`proxy/setup_sandbox.py` now writes the bot's noreply identity into the sandbox user's global Git config, in the same place it already wrote `core.excludesFile`; both harness actions hand it the login and id the rate-limit preflight resolves from the token. Global rather than checkout-local so a clone the agent makes for itself is covered as well, and the step errors by name if either value is missing.

The guidance is gone with it: the `pr-creation.md` section, its pointers from `review-runs`, `skill-pr-workflow.md` and `resolve-conflicts`, the `git config` lines in the integration-test recipe, and the identity setup in `nightly_workflow_update.py`'s ship step. Action and bundled skills both ship from the release tag, so the seeding and the removal land in the same release.

`pytest` and `pre-commit` pass locally; the hosted-runner smoke test now reads `user.name`/`user.email` back as the sandbox user, so CI checks the part unit tests can't.

tend-agent (commented):
`setup_sandbox.py` now hard-fails without `TEND_BOT_LOGIN`/`TEND_BOT_ID`, and both only reach it if each is named twice in the harness action — once in the step's `env:`, once again in the `env -i` argv. Nothing here exercises that pairing: `proxy/test-setup-sandbox.sh` exports the two itself, and neither `action.yaml` is linted or run by this repo's CI, so dropping either line from one harness would first surface as every adopter run on that harness dying at sandbox setup on the first run after a release — the same exposure `test_action_path_references_resolve` already guards for action paths. I'm pushing a parity test over both actions' sandbox step asserting every `env:` key (bar the shell-hardening blanks) is forwarded, plus that the identity pair is present in both lists.

PR #1253, opened 2026-09-15
fix(running-in-ci): one when-keyed index for every reference

max-sixty:
Tend's PR and issue bodies have been shipping hard-wrapped at ~72-80 columns, which GitHub renders as breaks mid-sentence. The single-line-paragraph rule is in `references/posting.md`, but the references index keyed that file to comments and `pr-creation.md` to `gh pr create`, so a session opening a PR read `pr-creation.md` and never reached `posting.md`. Two triage runs 15 minutes apart show it in their session logs: the one whose PR body wrapped read `posting.md` 13 minutes after `gh pr create`, on its way to the issue comment, which came out unwrapped.

The index in `running-in-ci/SKILL.md` is now a table keyed by action that names every file to read before it, so the row for `gh pr create` and `gh issue create` lists `posting.md` alongside `pr-creation.md`. Its lead-in states that read as a requirement rather than describing the table, and tells the session to read the files in the run that takes the action — too late is how this one failed.

It is also the plugin's only index now, and the only place that says when to read a reference and what it carries. Review's four references join the table, the body paragraphs that restated a file's contents are gone, and so is the `Depth behind ...` line each reference opened with. That line held two instructions in `re-targeting.md` that appear nowhere else — read the delta in full, keep the hidden draft marker — so those moved into its body.

Two consistency repairs came out of review: the **Scope** restriction promised that **Other Repos** carries all three cases after that section had been cut back to the default alone, and `CLAUDE.md`'s authoring rule still told authors to keep a per-skill references list, which would have re-fragmented the index.


plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
```diff
@@ -134,7 +140,7 @@ When a skill's code block needs edge-case handling or grows past a couple of doz
 
 ## Other Repos
 
-Default: don't act in another repo unsolicited. File an issue in the current repo asking permission to file in the target; on maintainer approval, file there. `references/other-repos.md` carries the rest: the standing exception an overlay can grant for agent-equipped targets, what an issue body must contain, when an invitation makes a PR or comment in the target legitimate, and what to do when a scope rule is the only thing between you and the right move.
+Default: don't act in another repo unsolicited. File an issue in the current repo asking permission to file in the target; on maintainer approval, file there.
```
tend-agent: The section is the target of four in-repo cross-references that expect it to carry more than the default. Restoring the file pointer — without the contents list, which the table's third column now owns — keeps them resolving.

```suggestion
Default: don't act in another repo unsolicited. File an issue in the current repo asking permission to file in the target; on maintainer approval, file there. `references/other-repos.md` carries all three cases.
```

tend-agent (commented):
Two places the consolidation leaves behind an inconsistency.

The **Scope** restriction still reads "**Other Repos** below carries all three cases", but that section is now the default case alone with no pointer onward. Three other call sites route to it for the same substance — `review-runs`' "file the fix against tend per **Other Repos**", `notifications`' "new issues follow **Other Repos**", and `pr-creation.md`'s "filing into another repo follows **Other Repos** in `SKILL.md`" — so each now lands on two sentences that say only "ask permission first". The inline suggestion restores a one-clause pointer, which repairs all four at once; fixing the Scope sentence to name `references/other-repos.md` directly would also clear the false claim, but leaves the three cross-skill pointers thin.

`CLAUDE.md`'s **Authoring skills** rule is now out of step with the structure this PR establishes: "`SKILL.md` keeps the rules every session needs plus a references list that names each file and the action that triggers reading it". `review/SKILL.md` deliberately no longer keeps one. Left unchanged, the next session that adds a reference file follows CLAUDE.md, re-adds a per-skill list, and re-fragments the index this PR just made singular. I can push that CLAUDE.md edit onto this branch if you want it here.

plugins/tend-ci-runner/skills/running-in-ci/SKILL.md
max-sixty: Applied in 411c317d, taking the suggestion as written. The second finding landed too: CLAUDE.md's authoring rule now sends a new reference file to the References table rather than to a list in its own skill.

tend-agent (approved):

PR #1254, opened 2026-09-16
fix: defeat a forced-colour environment in every gh reader

tend-agent:
Every `gh` reader in the repo's Python parses the response as JSON, and each one ran `gh` on the environment it inherited. `gh` ranks a forced colour setting above `NO_COLOR` and colorizes even when its output is a pipe rather than a terminal — so under `CLICOLOR_FORCE=1`, a body arrives wrapped in ANSI escapes and the parse fails. Each reader treats a decode error as a transient blip, so under a forced colour setting the fallback is permanent, and every one of them degrades quietly:

- `notifications_check.py` — `_notifications` and `_conflicts` catch `json.JSONDecodeError` and return `0`. Every cycle reports no notifications and no conflicted bot PRs, so the bot silently stops draining its inbox and stops resolving conflicts on its own PRs.
- `mention_verify.py` — the relayed-dispatch readers catch it and fall through to `verdict(False)`, printing `review N not found on PR M — skipping`. Mentions on review events go unanswered.
- `checks.py` — `check_branch_protection` reads `rules/branches/<b>` and `branches/<b>/protection` as plain JSON, and both decode handlers fall through to `CheckResult(name, True, f"Branch '{branch}' is protected")`. `tend check` reports the branch as protected without ever verifying that the bot cannot merge its own PRs — the check [`docs/security-model.md`](https://github.com/max-sixty/tend/blob/main/docs/security-model.md) calls the primary security boundary. This is the worst of the set: a false green on a security check, not a poll reporting zero.
- `shared/steps/_common.py` — `security_preflight`'s first read is deliberately left to raise, so the gate goes red rather than quiet. It gets the same guard anyway.
- `install_codex_subscription_auth._verify_secrets` — parses `gh secret list --json name`; a painted body aborts provisioning with an uncaught `JSONDecodeError`. Loud, same two lines.

The fix sets `NO_COLOR=1` and `CLICOLOR_FORCE=0` in the child's environment. `CLICOLOR_FORCE=0` is the part that matters: `gh` treats a forced value as higher priority than `NO_COLOR`, so `NO_COLOR` alone loses to an inherited `CLICOLOR_FORCE=1`. This is the same two-line guard `github_cli.run` and `refresh_consumers.github_json` already carry; the readers here each kept their own `gh` — the two templates because they're rendered inline into the generated workflow rather than imported, the rest because they're separate deployment units that can't share a helper.

`gh` paints a JSON object or array but leaves a scalar `--jq` result alone. Checked against the `gh` on the runner (2.100.0) rather than inferred: under `CLICOLOR_FORCE=1`, `--jq '{name}'` comes back ANSI-wrapped and pretty-printed across lines while `--jq .default_branch` prints `main` plain, and adding `NO_COLOR=1 CLICOLOR_FORCE=0` restores the compact single line. So the `gh api --paginate … --jq '… | .id'` reaction-cleanup steps in the generated workflows, and `check_branch_protection`'s own `.protected` read, are genuinely unaffected and stay as they are.

One shell reader does parse a shape `gh` would paint: [`.github/workflows/integration-secrets.yaml`](https://github.com/max-sixty/tend/blob/main/.github/workflows/integration-secrets.yaml) slurps the integration repo's `deployment-branch-policies` with no `--jq` and pipes the array into `jq`. It stays unguarded — that workflow is hand-maintained here and its steps set their own `env:`, so no adopter config and no shell reaches it.

Reachable from an adopter's own configuration: a `workflow_extra`/`jobs.*` `env:` override, or a `setup:` step that writes `CLICOLOR_FORCE=1` into `$GITHUB_ENV` in a workflow where setup precedes the pre-check — and for `tend check` and the install script, simply a shell that exports it. Forcing colour for `cargo`/`clippy` output is a normal reason to set it, which is how the same interaction previously broke `token-report.sh` (#220).

Each of the three gates that fails quietly gets a regression test that drives it with `CLICOLOR_FORCE=1` against a fake `gh` mimicking the precedence, and asserts it still does its job. The `colorize` helper that models the precedence lives in `tests/__init__.py`'s `GH_PREAMBLE` beside the `emit` it wraps, so one definition serves every fake and stays inert unless a test sets `CLICOLOR_FORCE=1`. Each test fails with its guard reverted: the notifications one on `count`, the mention one on the verdict, and the `checks.py` one on the assume-OK message rather than `(requires reviews)`.

The regenerated `_regtest_outputs` snapshots are the generated notifications and mention workflows, which embed these scripts inline. `.github/workflows/tend-notifications.yaml` and `tend-mention.yaml` are deliberately untouched: tend's own workflows track the published release and pick this up at the next one.


tend-agent (commented):
The diagnosis holds and the fix is the right one — `gh` does rank a forced colour setting above `NO_COLOR`, and a pre-check that silently reports "no work" every cycle is the worst shape this failure could take. Two things about the scope it landed in.

**The guard covers one of three copies of the same helper.** `generator/src/tend/templates/mention_verify.py` has the identical `gh()` — same inline-rendered pre-check, same `subprocess.run(["gh", *args], ...)` inheriting the job environment, same JSON parse downstream. Its dispatch readers swallow the decode error the same way the notifications ones do: `except (subprocess.CalledProcessError, json.JSONDecodeError)` around the `reviews/{item_id}` and `pulls/comments/{item_id}` fetches, each falling through to `verdict(False)`. So under the same environment a relayed review dispatch prints `review N not found on PR M — skipping` and the mention goes unanswered, on a green job — the notifications failure mode exactly, on the other pre-check. `shared/steps/_common.py`'s `gh()` is the third; its exposure is louder, because `security_preflight`'s first read (`gh_json("api", f"repos/{repo}")["default_branch"]`) is deliberately left to raise, so the gate goes red rather than quiet. Still the same two lines.

**The `count` assertion in the new test passes without the fix.** `FAKE_GH_NOTIFICATIONS`'s colour override wraps `emit`, but the `api:notifications\?*` branch never calls it — it writes its pages straight out with `echo '[]'` and `printf '%s\n' "$pages" | jq -c '.[] | [.]'`. Revert the `env=env` line and the test fails on `conflict_count`, not on `count`: only the `_conflicts` reader is actually pinned, and a later change that narrowed the guard to that one call site would keep the test green. That also makes the PR body's "Without the fix it reports `count=0`, `conflict_count=0`" untrue of this fake.

I'm pushing both to this branch: the guard on the two remaining copies with a forced-colour regression test for the mention gate, and the fake reworked so every body — `--paginate` pages included — goes through one `colorize` helper, which makes the `count` assertion fail without the fix. The PR title and description will move to match the widened scope.

tend-agent (commented):
Two more `gh` readers have the same gap, and one of them fails worse than the two this started from.

**`generator/src/tend/checks.py`'s `_gh` turns the security audit green without reading it.** `check_branch_protection` confirms `.protected` through a scalar `--jq`, which real `gh` leaves unpainted, and then reads `repos/{repo}/rules/branches/{branch}` and `.../protection` as plain JSON, which it does paint. Both decode handlers treat an unparsable body as "can't read the details, the branch is protected, assume OK" — `except json.JSONDecodeError: return CheckResult(name, True, f"Branch '{branch}' is protected")`. So under a forced-colour environment `tend check` reports `branch-protection:main` as **passing** without ever verifying that the bot cannot merge its own PRs, which `docs/security-model.md` calls the primary security boundary. That is a false green on a security check rather than a poll reporting zero, so it is the sharpest instance of this bug in the repo. Confirmed by running the real `_gh` against a `gh` stand-in that paints JSON bodies and leaves scalar `--jq` results alone: `passed=True`, message `Branch 'main' is protected` — the assume-OK fallback, not `(requires reviews)`.

**`install_codex_subscription_auth._verify_secrets`** parses `gh secret list --json name` from an inherited environment, so a painted body aborts provisioning with an uncaught `JSONDecodeError`. Loud rather than silent, same two lines — the same reasoning the PR already applied to `shared/steps/_common.py`.

**The two fakes' `colorize`/`emit` are byte-identical.** They belong in `tests/__init__.py`'s `GH_PREAMBLE`, which already owns `emit`: one definition instead of two, and every other fake gets the same realism for free. It stays inert unless a test sets `CLICOLOR_FORCE=1`, so no existing fake changes behaviour.

One boundary worth recording, since it decides how far this should widen: the `gh api --paginate … --jq '… | .id'` reaction-cleanup steps in the generated workflows need no guard. `gh` paints a JSON object or array but leaves a scalar `--jq` result alone — `--jq .default_branch` prints `main` plain while `--jq '{name}'` comes back wrapped — so those steps and `check_branch_protection`'s own `.protected` read are genuinely unaffected. With `checks.py` and the install script guarded, every `gh` reader in the repo that parses a response is covered, and the title's "every gh reader" becomes true.

I'm pushing all three to this branch, with a regression test on `check_branch_protection` that fails with the guard reverted (it returns the assume-OK message), and moving the title and description onto the widened scope.

PR #1256, opened 2026-09-16
test: fail the bot_name and memory_gist tests on an unmatched agent step, and cover the mention filter

tend-agent:
Two generator tests filtered a generated workflow's job steps for a `uses:` starting with `max-sixty/tend/claude@` and made their assertions inside that loop. With no match the loop body never runs, so both passed green while asserting nothing — a change to the action ref, or to how those inputs reach the step, would have landed unnoticed. A third gap was a docstring claim with no assertion behind it.

**`test_init_bot_name_in_workflow_content`** (`generator/tests/test_integration.py`) now counts the agent steps it checked across the generated `tend-*.yaml` and asserts the count is non-zero.

**`test_memory_gist_is_an_explicit_experimental_claude_only_input`** (`generator/tests/test_generate.py`) has the same filter in its `enabled` branch. Its loop now iterates `without_relay(generate_all(enabled))` and asserts per workflow that the filter matched — stronger than a count, because the invariant is that every generated workflow but `tend-mention-relay` carries an agent step, and the relay runs no agent.

**The bot_name docstring promised more than the test checked.** It claims coverage of the name "in the tend action's `bot_name` input and mention filters", and no assertion read a filter. The mention filter is the `verify` job's `if:` in `tend-mention.yaml`, which embeds `@<bot_name>` through the template's `summoned()` macro; the test now asserts `@my-custom-bot` appears there.

Neither vacuous-pass guard pins a count. The identical filter in `test_init_workflows_have_required_permissions` pins its at `== 7`, but that assertion is already carrying the every-workflow-contributes-one-agent-job invariant; repeating the number would give one root cause two failures to update.

Verified by mutating the prefix in both files to a ref no workflow uses: before these changes both tests still passed, after them `test_init_bot_name_in_workflow_content` fails with `no agent step matched: the action ref or its path moved` and `test_memory_gist_is_an_explicit_experimental_claude_only_input` with `tend-review.yaml: no agent step matched`.

Tests only — no generator or workflow behavior changes.


tend-agent (commented):
Two gaps of the same class as the one this closes. Pushing both to this branch.

**`test_generate.py`'s sibling filter still passes vacuously.**
`test_memory_gist_is_an_explicit_experimental_claude_only_input`'s `enabled`
branch filters on the same `max-sixty/tend/claude@` prefix and asserts
`memory_gist`/`memory_gist_id` inside the loop; mutating the prefix to a ref no
workflow uses leaves it green, exactly as it left this test green.
`without_relay()` is already imported there, so it can assert the filter
matched per workflow — stronger than a count, because the invariant is one
agent step in every generated workflow but the relay.

**The docstring still claims coverage this test doesn't have.** It promises the
bot name "in the tend action's `bot_name` input and mention filters", and no
assertion reads a filter. The mention filter is the `verify` job's `if:` in
`tend-mention.yaml`, which embeds `@<bot_name>` through the template's
`summoned()` macro; asserting `"@my-custom-bot"` appears in it closes the gap.
Same defect this PR fixes: a claim in the test with nothing behind it.

max-sixty:
@tend-agent the title and description still describe only the first commit. The head also makes the enabled branch of `test_memory_gist_is_an_explicit_experimental_claude_only_input` (`generator/tests/test_generate.py`) iterate `without_relay(...)` and assert that each workflow matched an agent step, and adds an assertion to `test_init_bot_name_in_workflow_content` that `@my-custom-bot` appears in the `if:` of `tend-mention.yaml`'s `verify` job. The body still ends with "no other behavior changes", and it becomes the squash commit message.

Please rewrite the title and description to cover the whole PR: both vacuous-pass guards and the new mention-filter assertion (see "Keeping PR titles and descriptions current" in `running-in-ci/references/pr-creation.md`). The code needs no change.

tend-agent:
Rewritten. The title now names both tests and the unmatched-agent-step condition, and the body covers all three changes: the count guard in `test_init_bot_name_in_workflow_content`, the per-workflow `without_relay(...)` guard in `test_memory_gist_is_an_explicit_experimental_claude_only_input`, and the new `@my-custom-bot` assertion against `tend-mention.yaml`'s `verify` `if:`. The "no other behavior changes" line is now scoped to what it meant — tests only, no generator or workflow behavior change.

I re-ran the mutation check over the current head before rewriting the claim: pointing both filters at a ref no workflow uses fails both tests, `checked == 0` in the first and `tend-review.yaml` in the second. No code changes.

PR #1264, opened 2026-09-17
fix(notifications): stop counting an abandoned run as a subject's owner

tend-agent:
## Problem

`active_subject_runs.py` treats every non-terminal run as an owner of its subject, with no bound on how long it may stay one. That was the right fix for the case it was written for — a `tend-mention` run sitting `queued` for hours still owes the maintainer an answer — but GitHub also strands runs that never start: created, zero jobs, `updated_at` equal to `created_at`, and answering `status=queued` indefinitely. This repository is holding three of them right now:

| Run | Workflow | Created | Jobs |
|---|---|---|---|
| [31164659276](https://github.com/max-sixty/tend/actions/runs/31164659276) | `tend-mention` | 2026-08-07 | none |
| [34748600612](https://github.com/max-sixty/tend/actions/runs/34748600612) | `tend-notifications` | 2026-09-13 | none |
| [34952625187](https://github.com/max-sixty/tend/actions/runs/34952625187) | `tend-mention` | 2026-09-15 | none |

A notification thread whose subject one of these names is deferred by the poll that finds it and by every poll after it, so nothing ever handles it. That is a hole in the recovery queue rather than wasted compute: the poll exists to catch what the event workflows missed, and a thread parked there is a maintainer request the bot never answers. The `tend-mention` run above pins PR #1248, which this poll found unread and had to reason past by hand.

## Solution

`owning_runs` skips a run whose `updated_at` has not moved in 24h. GitHub advances `updated_at` on every job transition, so the bound measures how long a run's state has stood still rather than how long the run has existed — a run queued most of a day and then running out its job timeout (no `timeout-minutes` is set, so GitHub's 6h default applies) is over a day past `created_at` while it is still working the subject, and aging that one out would cost the second outward answer the ownership check exists to prevent. A stranded run reports `updated_at` equal to its `created_at`, so the bound still reaches exactly the three above.

An absent or unparsable `updated_at` still counts as an owner. The asymmetry is deliberate: deferring one extra poll costs a poll, while dropping a live owner costs a second outward answer from the same bot account — the failure [#1235](https://github.com/max-sixty/tend/pull/1235) closed.

The script moves to `requires-python = ">=3.12"`, matching `list_recent_runs.py` and the other scripts that read timestamps, so `datetime.fromisoformat` parses GitHub's trailing `Z` directly and `UTC` is available. `uv run --script` provisions the interpreter, so this is not a deployment constraint.

## Testing

Four tests in `generator/tests/test_active_subject_runs.py`, on a `_run` fixture that now carries both timestamps relative to the wall clock the script reads, defaulting `updated_at` to `created_at` — the shape GitHub reports for a queued run and for a stranded one:

- a run untouched for 980h is not an owner — fails before the change, which reports it as one;
- a run queued for 20h still is, pinning the bound above a real `tend-mention` queue rather than a fast one;
- a run created 25h ago but touched 2h ago still is, pinning the measurement to the last state change rather than to creation;
- a run with no `updated_at` still is, pinning the conservative fallback.

`uv run pytest` passes apart from the two `test_refresh_consumers.py` cases that need PyPI, which the sandbox has no DNS for. Every `pre-commit` hook passes.


tend-agent (commented):
The bound keys on `created_at`, which covers the queue *and* the session, while the justification for 24h only covers them separately. No `timeout-minutes` is set on the generated workflows, so a job gets GitHub's 6h default: a `tend-mention` run queued ~19h — within the range this PR's own test pins as realistic — and then running is past 24h from creation while still live. `abandoned()` drops it, the poll reads the subject as unowned, and both sessions answer the maintainer as the same bot account. That is the failure #1235 closed, and the direction the description calls the expensive one.

`updated_at` measures the thing the bound is reaching for, and it is already in the same API response. All three stranded runs in the description report `updated_at` equal to `created_at`, so the check keeps catching exactly them; a live run's `updated_at` advances as its jobs transition — the in-progress `tend-review` run on this PR reports one 20 seconds after creation, so no live owner can age past the bound. Swapping the field is strictly narrower than raising it.

I'll push the full swap — the constant's comment, the docstring, the `SKILL.md` sentence, and the test fixture move with it — so the inline suggestion below is just the line the finding lands on.

plugins/tend-ci-runner/scripts/active_subject_runs.py
```diff
@@ -29,9 +32,33 @@
...
+
+
+def abandoned(run: dict[str, Any], *, now: datetime) -> bool:
+    """Whether *run* has been non-terminal longer than a run ever legitimately is.
+
+    An unreadable or absent `created_at` counts as fresh: deferring one extra
+    poll costs a poll, while dropping a live owner costs a second outward
+    answer from the same bot account.
+    """
+    created = str(run.get("created_at") or "")
```
tend-agent: `created_at` is set once and never moves, so this ages out a run that queued long and then started. `updated_at` advances on every job transition and is equal to `created_at` on the stranded runs this targets.

```suggestion
    created = str(run.get("updated_at") or "")
```

plugins/tend-ci-runner/scripts/active_subject_runs.py
```diff
@@ -29,9 +32,40 @@
 # below by run id rather than by name, since a second poll would own nothing.
 WORKFLOW_PREFIX = "tend-"
 
+# A run whose state has not moved in this long is never going to deliver.
+# GitHub terminates a job queued past 24h, so nothing legitimate sits
+# untouched for a day. Runs do get stranded non-terminal with no jobs and no
+# further updates, and without this bound one of them pins its subject for
+# good.
```
tend-agent: The 24h job-queue cancellation GitHub documents is a self-hosted-runner limit, and these workflows run on `ubuntu-24.04`, so it does not establish that nothing legitimate sits untouched for a day. Restating the bound as observed state keeps the one shape that can outlast it visible.

```suggestion
# A run whose state has not moved in this long is never going to deliver.
# GitHub documents no queue-time cancellation for its hosted runners — the 24h
# one covers self-hosted — so this bound rests on observed state rather than a
# guarantee: a live run advances `updated_at` at every job transition, and the
# shape that could outlast it is a deep `queue: max` backlog. Runs do get
# stranded non-terminal with no jobs and no further updates, and without this
# bound one of them pins its subject for good.
```

tend-agent (commented):
`ABANDONED_AFTER`'s comment justifies the bound with "GitHub terminates a job queued past 24h." That cancellation is a self-hosted-runner limit: [`content/actions/reference/limits.md`](https://github.com/github/docs/blob/main/content/actions/reference/limits.md) lists *Job queue time — 24 hours* in the **Self-hosted** section, while the GitHub-hosted rows carry a 6h *execution* cap and no queue-time row at all. These workflows are `runs-on: ubuntu-24.04`, so nothing cancels a queued run of theirs at a day.

The value still holds on evidence — across the last 882 `tend-*` runs (6 days) the longest-lived is 0.78h, and all three stranded runs report `updated_at` equal to `created_at`. What the wrong premise costs is the next reader's reasoning: it reads as a guarantee, so the one shape that can outlast 24h stays invisible. `tend-review` holds its group with `queue: max`, which the same limits page caps at 100 runs per concurrency group, and a run parked there has no job transitions — the very thing the `abandoned()` docstring says advances `updated_at` — so its timestamp stays frozen at `created_at` for the whole wait. Around thirty queued sessions on one PR puts a live owner past the bound, and `abandoned()` drops it: the second outward answer this check exists to prevent.

I'm leaving the constant at 24h rather than widening it — that would trade a day's deferral on a genuinely stranded thread against a backlog depth this repo has never reached, and it's a knob to move if `queue: max` queues ever get deep. Pushing the comment correction below.

PR #1265, opened 2026-09-17
docs: split shipped guidance from tend's own, and settle on `consumer`

max-sixty:
A survey of the bundled skills found tend's own maintainer stance — its order of value, a single-maintainer PR budget, what this repo will and won't spend complexity on — written into files that ship to every repo installing tend. Those are one maintainer's calls, and shipped they make every consumer's repo answer to this one's. Nothing in the repo named the boundary they crossed, so this names it and applies it.

## The split

`CLAUDE.md` gains **Shipped guidance and tend's own**, split on which repo the text governs. Shipped guidance is any text that reaches another repo — more than `plugins/**`: `shared/system-prompt.md` and the two directives `compose_system_prompt.py` prepends, `codex/agents-tail.md`, the header comment `workflows.py` stamps into every generated workflow, and `docs/tend.example.yaml`, which a pre-commit hook mirrors into `install-tend`. Tend's own guidance is read only here.

The test for a line of shipped text: does it hold in a repo whose maintainer disagrees with us? Mechanics ship freely — how GitHub behaves, what the harness does, what a session must never do. A default a consumer might want differently ships as a default their overlay can change, stated without our reasoning for it.

Not `tend-internal`: a skill's `metadata: internal: true` already means it isn't user-invocable, and every shipped skill carries it, so the word would mean roughly its opposite here.

Under that, a `### Which file` subsection keeps the narrower placement question — a rule is loaded by one set of sessions and acted on by a subset, so it goes at the narrowest level every actor still reads — with a table of the levels, who reads each, and whether it ships. The one-home rule is new; duplication had no rule to violate before.

`skill-pr-workflow.md` gets the universal half, since it is what a session in any repo reads before writing guidance. "Where it lands" asked one question, *would every tend consumer want this rule?*; it now asks a second, *is it ours to state?* That also resolves against the later line telling generic-reading rules to go upstream, which now says a rule of mechanics does.

`nightly`'s review checklist gains one bullet: its convention-compliance bullets are code-shaped, so a surveyed Markdown guidance file got no placement question.

## One word for a consumer repo

The repo used *adopter* and *consumer* interchangeably — 187 and 277 occurrences, in the same files, for the same thing, plus five "consuming repo". All *consumer* now, because the identifiers already say it (`data/consumers.json`, `refresh_consumers.py`, the Worker's data streams) and those can't move cheaply. Mechanical apart from the a/an articles the swap broke and one *adopting* the stem missed. `CHANGELOG.md` keeps *adopter*: released entries are a record, not vocabulary. That commit is separable from the rest.

## What this doesn't do

It changes the rules, not the text they judge. The survey's individual findings — the compute stance still reaching a consumer's own CI config in `review/SKILL.md`, the single-maintainer PR budget in `references/pr-creation.md` — are unfixed here. `TODO.md` records the one workflow that breaks the new invariant that every agent-invoking workflow names a skill: `tend-mention`'s prompt is inline in `mention.yaml.j2`, which is why `running-in-ci` doubles as mention's skill.


shared/steps/sandbox_setup.py
```diff
@@ -47,7 +47,7 @@ def main() -> int:
 
     code = subprocess.run(setup_argv(commands), check=False).returncode
     if code == 0:
-        _common.log(STEP, "ran adopter sandbox_setup commands inside SRT")
+        _common.log(STEP, "ra consumer sandbox_setup commands inside SRT")
```
tend-agent:
```suggestion
        _common.log(STEP, "ran consumer sandbox_setup commands inside SRT")
```

tend-agent (commented):
The `an adopter` → `a consumer` article substitution matched inside a word in one place and skipped the wrapped ones, so three things came through the rename broken.

`sandbox_setup.py`'s success log now reads `ra consumer sandbox_setup commands inside SRT` — `ran` lost its `n` to the article pattern. `test_sandbox_setup.py` pins that exact string, so the suite is green on it and every consumer's run log prints the mangled line from the next release onward.

Where the article and the noun landed on different lines the substitution didn't see them, leaving four sites reading `an consumer`. Two are `bot_name`'s input description in both shipped actions, which is the text a consumer reads on the action's own page.

Separately, the new `TODO.md` section names the invariant's home as "Where guidance text goes", and no heading by that name exists anywhere in the tree — the invariant is under `### Which file` in `CLAUDE.md`. It's the shape `nightly`'s checklist calls a stale instruction referencing a renamed file, so the next sweep will re-find it.

Suggestions are inline; happy to push them as a single commit instead if you'd rather not click through seven.

shared/steps/test_sandbox_setup.py
```diff
@@ -80,5 +80,5 @@ def test_successful_setup_is_reported(
 
     assert sandbox_setup.main() == 0
     assert capsys.readouterr().out == (
-        "[sandbox-setup] ran adopter sandbox_setup commands inside SRT\n"
+        "[sandbox-setup] ra consumer sandbox_setup commands inside SRT\n"
```
tend-agent: Pins the log string above, so this moves with it.

```suggestion
        "[sandbox-setup] ran consumer sandbox_setup commands inside SRT\n"
```

TODO.md
```diff
@@ -251,3 +251,17 @@ lives in KV and is what the site renders. If the summary wants a longer
...
+
+`workflows.py` builds every other agent-invoking workflow's prompt from
+`default_prompt(skill)`; mention's is written inline in `mention.yaml.j2` and
+names no skill, so `running-in-ci` doubles as mention's skill. Rules with
+mention's shape — reading a triggering comment, a review's inline comments,
+whether to respond, the closed-target check — land in the file every session
+loads because there is nowhere else for them to go.
+
+A `mention` skill invoked by that prompt gives them a home and restores the
+invariant in "Where guidance text goes". The move is the work: decide which of
```
tend-agent:
```suggestion
invariant in "Which file". The move is the work: decide which of
```

claude/action.yaml
```diff
@@ -30,7 +30,7 @@ inputs:
     required: true
     description: |
       GitHub username of the bot account. Rendered into the workflows so an
-      adopter can see which account tend runs as; the steps themselves read
+      consumer can see which account tend runs as; the steps themselves read
```
tend-agent:
```suggestion
      GitHub username of the bot account. Rendered into the workflows so a
      consumer can see which account tend runs as; the steps themselves read
```

codex/action.yaml
```diff
@@ -26,7 +26,7 @@ inputs:
     required: true
     description: |
       GitHub username of the bot account. Rendered into the workflows so an
-      adopter can see which account tend runs as; the steps themselves read
+      consumer can see which account tend runs as; the steps themselves read
```
tend-agent:
```suggestion
      GitHub username of the bot account. Rendered into the workflows so a
      consumer can see which account tend runs as; the steps themselves read
```

shared/steps/report_failure.py
```diff
@@ -17,7 +17,7 @@
 work the failed runs may have missed. Reopening would fold the next incident
 into a stale record. Nightly leaves this label alone because its cron precedes
 `review-runs`, and a clean recent run says nothing about missed work. Where an
-adopter disables `review-runs`, a maintainer applies the same close criterion.
+consumer disables `review-runs`, a maintainer applies the same close criterion.
```
tend-agent:
```suggestion
`review-runs`, and a clean recent run says nothing about missed work. Where a
consumer disables `review-runs`, a maintainer applies the same close criterion.
```

shared/steps/test_sandbox.py
```diff
@@ -92,7 +92,7 @@ def test_launch_env_is_anchored_to_the_prefix(compose: Compose) -> None:
     """`GITHUB_`, so a name that merely contains it stays on the runner.
 
     `MY_GITHUB_TOKEN` and `GITHUBBER_TOKEN` are the shapes that matter: an
-    adopter `setup:` step is free to export either, and neither may ride across
+    consumer `setup:` step is free to export either, and neither may ride across
```
tend-agent:
```suggestion
    `MY_GITHUB_TOKEN` and `GITHUBBER_TOKEN` are the shapes that matter: a
    consumer `setup:` step is free to export either, and neither may ride across
```

tend-agent (approved):

PR #1266, opened 2026-09-17
docs: judge a token saving by its share of total spend; scale code-review by coverage

max-sixty:
Two guidance changes, both from the session that landed #1263.

`CLAUDE.md`'s **What waste is worth fixing** said only that model tokens are the larger cost. It now says what makes a token saving worth a mechanism: the share of total spend it removes, estimated before the change is proposed, with a small or rare saving justifying no complication however well it reads line by line.

`code-review`'s `SKILL.md` wrote its depth scaling as subagent fan-out — "fan out one subagent per angle", "one subagent verifier per candidate", and a report of "which mode ran". The rule underneath is coverage: every angle covered, every candidate verified. It now says that, and asks the caller to be told how much was covered rather than which mechanism ran. `review`'s **Second pass** described the core branch the same way, so its one stale clause follows the same change.

The token-cost test was also drafted into `review/SKILL.md` and `review-reviewers/SKILL.md` and is not in this PR. #1265 landed while it was being written, and its rule — what this repo will spend complexity on is one maintainer's call and doesn't ship — covers exactly that text. Those two files and Gate 3 in `shared/review-gates.md` still carry tend's cost stance in shipped guidance; taking it out is a separate change.


tend-agent (commented):
`plugins/tend-ci-runner/skills/review/SKILL.md` still describes this skill's core-logic mode in the terms the change removes. Under **Second pass**, its core branch reads "tell it the change is core, so it fans the angles out and sweeps for gaps". After this PR `code-review`'s core branch names no fan-out — it asks the session to cover every angle however it works best — so the one caller that invokes the skill instructs a mechanism the skill no longer describes, and it does so in precisely the session shape the new wording exists to serve: one with no way to run an angle separately. Only the fan-out clause is stale; "sweeps for gaps" still matches Phase 3. "tell it the change is core, so it covers every angle and sweeps for gaps" would track the new text.

That line is a restatement of the mode rather than a pointer to it, which is the drift **Which file** in `CLAUDE.md` warns about — "Another file that needs it names the section instead of restating it — copies drift". Happy to push the one-line fix onto this branch if you want it here rather than as a follow-up.

tend-agent (approved):

PR #1267, opened 2026-09-17
skills: stop shipping tend's own calls on what work is worth doing

max-sixty:
Follow-on to #1265, which named the split between shipped guidance and tend's own but didn't move the text it judges. These are the three bundled sites that applied tend's judgement to the repo the bot is running in — mechanics ship, but what work is worth doing in a repo belongs to whoever maintains it.

**`review/SKILL.md`** weighed changes "in the bot's own automation … and the repo's CI config", so a consumer contributor's own CI change was judged by tend's ranking of complexity against compute. The criterion now stops at the bot's *bundled* skills and the scheduled workflows that invoke them — "its skills" would still have covered the consumer-authored `running-tend` overlay.

**`references/pr-creation.md`** opened "Open PRs compete for one maintainer's attention" and held self-initiated improvements at five open bot PRs. The five was written whole in #960 and never derived — set below an eleven observed at the time — and it isn't right for tend either, so no repo would set it and the mechanism goes with it, exemption list included. The branch-basing rule in that paragraph is mechanics and is kept.

That cap was the only bound on **`nightly`**'s rolling survey, whose Step 8 read "For each finding: 1. Create a PR" with no limit — unlike `review-runs` and `review-reviewers` it references no gates and caps nothing, so an eight-finding survey would have opened eight PRs. Step 8 now groups findings by theme and keeps a run to a couple of PRs, with the rest in the run summary: the same restraint its siblings carry, in the skill that does the sweeping rather than as a global number.

**`ci-fix/SKILL.md`** required "a knob-sized mitigation, one setting in one place" for an escalated infrastructure fault, limiting a consumer's own reliability work to tend's bar. Now the smallest mitigation that holds, with the 3b fallthrough turning on that instead of on "knob-sized", which the edit had left undefined in the file.

Separately, the per-run caps in `review-runs` and `review-reviewers` read "a couple of PRs per run" rather than "at most 2". That number was never derived either, and the words carry the same instruction without the false precision.


plugins/tend-ci-runner/skills/ci-fix/SKILL.md
```diff
@@ -95,7 +95,7 @@ Match by failure-shape keyword against the issue body (e.g. `rustup-init`, `comp
...
 
-An escalated fault still reruns green, so a mitigation buys back runner time, not correctness. Open a fix PR proposing a knob-sized mitigation, one setting in one place (pin the runner image, skip the affected leg, disable the relevant cache layer), preferring an upstream-documented workaround — `gh issue search` against the action's repo, the action's README, GitHub Community threads — and linking the upstream issue if the search surfaced one. If the fault has no knob-sized mitigation, treat it as a durable cause without a safe fix and follow 3b.
+An escalated fault still reruns green, so a mitigation buys back runner time, not correctness. Open a fix PR proposing the smallest mitigation that holds, usually one setting in one place (pin the runner image, skip the affected leg, disable the relevant cache layer), preferring an upstream-documented workaround — `gh issue search` against the action's repo, the action's README, GitHub Community threads — and linking the upstream issue if the search surfaced one. If the fault has no knob-sized mitigation, treat it as a durable cause without a safe fix and follow 3b.
```
tend-agent: The paragraph now opens on "the smallest mitigation that holds" while this sentence still gates the 3b fallthrough on "knob-sized", which the edit left undefined in the file. A fault whose smallest holding mitigation is two settings reads as both "open a fix PR" and "no safe fix, file a tracker".

```suggestion
An escalated fault still reruns green, so a mitigation buys back runner time, not correctness. Open a fix PR proposing the smallest mitigation that holds, usually one setting in one place (pin the runner image, skip the affected leg, disable the relevant cache layer), preferring an upstream-documented workaround — `gh issue search` against the action's repo, the action's README, GitHub Community threads — and linking the upstream issue if the search surfaced one. If the fault has no such mitigation, treat it as a durable cause without a safe fix and follow 3b.
```

plugins/tend-ci-runner/skills/review/SKILL.md
```diff
@@ -71,7 +71,7 @@ Check the project's instruction files for language-specific review criteria and
...
+- In the bot's own automation — its skills, and the scheduled workflows that invoke it: is the change worth its weight? Challenge new mechanism whose only payoff is saved compute — retries, skip-gates, caches, scheduling arithmetic — and ask for waste that has recurred across days, not projected waste. Judge the whole change: a PR well-argued line by line can still cost more machinery than the compute it saves, and the review should say so plainly; the same setting repeated across workflows, jobs, or call sites is a configuration scheme, not a knob. Recommend one knob in one place, or closing the PR, over refinement.
```
tend-agent: The rationale for keeping this criterion is that what remains in scope is tend-supplied, but "its skills" also covers the consumer-authored `.claude/skills/running-tend/` overlay, and the workflows carry the consumer's own `setup:` and `sandbox_setup:` blocks. As written, a consumer maintainer editing their own overlay still has the change weighed against tend's ranking of complexity over compute — the same reach the CI-config clause had.

```suggestion
- In the bot's own automation — its bundled skills, and the scheduled workflows that invoke them: is the change worth its weight? Challenge new mechanism whose only payoff is saved compute — retries, skip-gates, caches, scheduling arithmetic — and ask for waste that has recurred across days, not projected waste. Judge the whole change: a PR well-argued line by line can still cost more machinery than the compute it saves, and the review should say so plainly; the same setting repeated across workflows, jobs, or call sites is a configuration scheme, not a knob. Recommend one knob in one place, or closing the PR, over refinement.
```

tend-agent (commented):
One inline fix and one call to make; a third narrowing is inline on `review/SKILL.md`.

**`nightly` is the sweep the removed budget bounded, and nothing replaces it there.** The description rests on per-skill restraint holding the line — `review-gates.md`'s gates plus the per-run caps in `review-runs` and `review-reviewers`. `nightly/SKILL.md` has neither: `review-gates.md` is referenced only from those two skills, and Step 8 reads "For each finding: 1. Create a PR" with no per-run limit. Its Step 6 rolling survey is the self-initiated sweep the removed paragraph named first, so a survey that turns up eight findings now opens eight PRs with nothing bundled saying otherwise.

That may be the intent, since how many open PRs is too many is a repo call. But CLAUDE.md's **Shipped guidance and tend's own** prescribes a different shape for exactly this case: "A default a consumer might want differently ships as a default their overlay can change, stated without our reasoning for it." Deleting the mechanism leaves no default for an overlay to change; restating a bare cap in `nightly` — without the single-maintainer-attention reasoning that made the old paragraph tend's own — would satisfy both readings. Happy to push that if you want it.

tend-agent (approved):

PR #1268, opened 2026-09-17
Keep tend's own calls out of the skills consumers load

max-sixty:
A survey of the bundled skills for text that governs tend rather than the repo the bot is running in, plus the repairs it turned up. It follows #1263 and #1265, and applies the ontology #1265 landed: shipped guidance has to hold in a repo whose maintainer disagrees with us, so tend's own calls belong in tend's own files.

## Tend's compute economics leave the bundled skills

Gate 3 of `shared/review-gates.md` priced a failure's cost and sized the fix it justified — waste costs cents, machinery is maintained forever, one knob in one place. That is tend's CLAUDE.md, shipped into every consumer's daily `review-runs` and applied to their repo. It's gone, and the framework is two gates: Gate 1's Critical row already carried the part that isn't taste, so it gains the rule that a wrong outward action has to have occurred rather than been imagined. `review-reviewers` loses its "Cost discipline" section — running log parsing in a subagent is about what the judging context holds, not what tokens cost — and `review`'s code-quality bullet asks whether a change carries its weight without supplying tend's answer. The rule itself stays in `CLAUDE.md`, where a bot session in this repo reads it and a session elsewhere reads that repo's.

The `Non-issues` block, which this branch moved up from `review-reviewers` into the shared file, keeps only what a consumer's session meets: `tend-mention`'s gate skipping a job, `/tend-ci-runner:review` posting nothing on a draft, `tend-review` re-reviewing after a fix commit. Four citations from tend's own incident log are out, and so is a paragraph addressed to a session filing upstream, now stated in each skill in its own scope.

## review-reviewers moves to tend's overlay

No generator list carries it: only the hand-maintained `review-reviewers.yaml` dispatches it, over a matrix of consumer repos, to file findings back onto tend. Inside the bundled plugin it installed into all eight consumers, was invoked by none, and collected tend's internal text beside `review-gates.md` — which every consumer's `review-runs` does load. That is how the compute stance got into the gates. It now lives at `.claude/skills/review-reviewers/`, invoked as `/review-reviewers` and reaching the plugin's scripts through `$SCRIPTS` the way the `running-tend` overlay already does.

## running-in-ci sheds what most sessions never use

171 lines to 107. Reading a thread, fetching a review's inline comments, the closed-target check, and whether to respond at all are acted on by the sessions that answer an issue or PR thread; they sat in the file every session loads because `tend-mention` has no skill of its own. They move to `references/trigger-context.md`, indexed on the action rather than on the trigger, so `triage` (`issues.opened`), `notifications` (a cron tick) and `review` (`pull_request_target`) all reach them. Reproduction and fix conditions move to `/tend-ci-runner:triage`'s `references/fixing.md`, and the system prompt keeps the self-loop guard and a pointer instead of restating what `running-in-ci` owns.

## Lints for what moving guidance breaks

In `generator/tests/test_repo_pins.py`, where the repo keeps invariants no single suite owns:

- **`test_skill_reference_citations_resolve` now resolves each citation to its owner.** It already walked every skill's citations but only checked that the filename existed in *some* skill. It now resolves the owner named before the path, after it, or implied by the citing skill; reports an owner that isn't on disk; checks bare siblings inside `references/` directories; and walks symlinks, so `shared/review-gates.md` is covered. Moving `trigger-context.md` broke a pointer in `posting.md` that the old check would have passed.
- **The References table indexes exactly the bundled `references/` files** — one no row names loads in no session.
- **Every workflow prompt's first token resolves to a skill on disk:** `/<plugin>:<name>` under `plugins/`, `/<name>` under `.claude/skills/`, or Codex's `$<name>`. `tend-mention` is the one agent-invoking workflow whose prompt opens with an expression instead, because it names no skill at all — the gap TODO.md tracks.

## Testing

`uv run pytest` (1159) and all 13 pre-commit hooks pass. Each new check was probed by breaking what it guards. The guidance changes themselves are unexercised by CI, as guidance always is; they were reviewed against the pre-change text for lost instructions and broken cross-references.


generator/tests/test_repo_pins.py
```diff
@@ -674,3 +674,132 @@ def test_every_workflow_pins_the_same_tend_release() -> None:
...
+                prompt = step.get("with", {}).get("prompt")
+                if not prompt:
+                    continue
+                first = prompt.strip().split()[0]
+                if first.startswith("${{"):
+                    continue
+                assert first.startswith("/"), (
+                    f"{path.name}'s prompt opens with `{first}`, not a slash command"
+                )
+                plugin, _, skill = first.lstrip("/").rpartition(":")
```
tend-agent: Accept the Codex form too, so a `harness: codex` override doesn't fail this as a broken workflow:

```suggestion
                if first.startswith("$"):
                    # Codex renders `$NAME` (`Config.default_prompt`).
                    plugin, skill = "tend-ci-runner", first.lstrip("$")
                else:
                    assert first.startswith("/"), (
                        f"{path.name}'s prompt opens with `{first}`, neither a "
                        "slash command nor a Codex skill mention"
                    )
                    plugin, _, skill = first.lstrip("/").rpartition(":")
```

tend-agent (commented):
Two things in the new lints.

`test_every_workflow_prompt_names_a_skill_that_exists` asserts the prompt's first token starts with `/`, but `Config.default_prompt` renders `$NAME` for Codex — the codex regtest for review carries `$review ${{ github.event.pull_request.number }}`. So the first per-workflow `harness: codex` override in `.config/tend.yaml` plus a regeneration turns this repo-invariant red, and its message (`tend-review.yaml's prompt opens with $review, not a slash command`) reads as a broken workflow rather than a test that only knows one harness. Suggestion inline.

`test_every_reference_a_skill_names_exists` overlaps `test_skill_reference_citations_resolve` in the same file, which already resolves every `references/<file>` citation a skill makes — and over a wider set, since its `skill_dirs` walks `.claude/skills/*` as well as `plugins/*/skills/*`. The new test is stronger on two axes the old one misses (the cited owner must match, and a bare sibling name resolves) and narrower in coverage, so neither subsumes the other and the repo now carries two citation linters with two regexes and two definitions of what gets checked. Folding the owner check and the sibling check into the existing test would keep one home for the rule; happy to push that if you want it.

generator/tests/test_repo_pins.py
```diff
@@ -674,3 +674,132 @@ def test_every_workflow_pins_the_same_tend_release() -> None:
...
+        if not path.is_symlink()
+    }
+
+    assert indexed == on_disk, (
+        f"unindexed: {sorted(on_disk - indexed)}; "
+        f"named but absent: {sorted(indexed - on_disk)}"
+    )
+
+
+def test_every_reference_a_skill_names_exists() -> None:
```
tend-agent: This duplicates `test_skill_reference_citations_resolve` above, which already walks both `plugins/*/skills/*` and `.claude/skills/*` and resolves every `references/<file>` citation in them. What's genuinely new here is the owner match and the bare-sibling check — both worth keeping, but as additions to that test rather than a second linter with its own regex and a narrower directory set.

generator/tests/test_repo_pins.py
max-sixty: Taken in 4d43ea12 — the lint now accepts `$NAME` as the Codex form of the same invocation, and its docstring says so.

generator/tests/test_repo_pins.py
max-sixty: Right — folded into it in 4d43ea12. The existing test now resolves each citation to its owning skill and checks bare siblings; the second linter and its regex are gone.

generator/tests/test_repo_pins.py
```diff
@@ -483,27 +483,61 @@ def test_action_path_references_resolve(action: str) -> None:
...
+    for skill in skill_dirs:
+        for path in skill.glob("**/*.md"):
+            if path.is_symlink():
+                continue
+            text = path.read_text()
+
+            for match in SKILL_REFERENCE.finditer(text):
+                cited += 1
+                named = match.group("skill") or match.group("skill_after")
+                owner = by_name.get(named, skill) if named else skill
```
tend-agent: The fallback here means an owner that doesn't exist resolves to the citing skill, so a stale `/<plugin>:<skill>` name passes whenever the citing skill happens to own a file of the same name. Report it instead:

```suggestion
                if named and named not in by_name:
                    broken.append(
                        f"{path.relative_to(REPO_ROOT)}: no skill named `{named}`"
                    )
                    continue
                owner = by_name[named] if named else skill
```

generator/tests/test_repo_pins.py
```diff
@@ -483,27 +483,61 @@ def test_action_path_references_resolve(action: str) -> None:
...
-        for path in d.glob("**/*.md")
-        for name in SKILL_REFERENCE.findall(path.read_text())
-    }
+    by_name = {d.name: d for d in skill_dirs}
+    cited, broken = 0, []
+
+    for skill in skill_dirs:
+        for path in skill.glob("**/*.md"):
+            if path.is_symlink():
+                continue
```
tend-agent: The shared files are only reachable through symlinks, so skipping links drops them from the walk entirely. Dropping the skip re-includes them, and passes once the `review-gates.md` citation names its owner (suggested on that line):

```suggestion
```

plugins/tend-ci-runner/shared/review-gates.md
```diff
@@ -55,11 +47,20 @@ The test: "If I replayed this exact scenario 10 times, would the failure occur e
...
+
+- **Designed no-ops.** Many events correctly end with nothing posted, and the layer that catches one varies: a pre-boot gate skips the job, or a session boots, judges the trigger, and exits silently. Both are the design — a workflow fires on an event class, and most of a class needs no response. Record them and move on. A loop that produces *wrong outward actions* — duplicate comments, spurious reviews — is a finding, and where a label can express the distinction a label-based skip beats an authorship filter.
+
+- **Designed silence.** `/tend-ci-runner:review` decides when a review posts nothing; that skill and its `references/draft-mode.md` are the contract, so read them before calling a missing review an omission. Two states the listing hides: GitHub reports a draft as `state: OPEN` (`gh api repos/OWNER/REPO/pulls/N --jq '{state, draft}'`), and the PR's literal author decides whether an approval was available at all (`gh pr view <n> --json author --jq '.author.login'`) — a human owner's PR is approved normally and is no precedent for approving a bot-authored one.
```
tend-agent: Naming the owner lets the citation resolve from whichever skill symlinks this file, so the lint can cover it again:

```suggestion
- **Designed silence.** `/tend-ci-runner:review`'s `references/draft-mode.md` and the skill itself are the contract for when a review posts nothing, so read them before calling a missing review an omission. Two states the listing hides: GitHub reports a draft as `state: OPEN` (`gh api repos/OWNER/REPO/pulls/N --jq '{state, draft}'`), and the PR's literal author decides whether an approval was available at all (`gh pr view <n> --json author --jq '.author.login'`) — a human owner's PR is approved normally and is no precedent for approving a bot-authored one.
```

tend-agent (commented):
Both earlier points are handled — the Codex branch matches `default_prompt`'s two forms exactly. Two checks the merged test lost against the linter it replaces.

**An owner name that doesn't resolve now passes silently.** `by_name.get(named, skill)` falls back to the citing skill when the named skill isn't on disk, so appending a citation that names `/tend-ci-runner:nosuchskill` and the path `references/approving.md` to `review/SKILL.md` passes: the fallback lands on `review`, which owns `approving.md`. The deleted linter built the path from plugin plus skill and reported it. That's the case the lint exists for — a skill renamed or moved leaves its citations behind, and this branch moves one. Suggestion inline; with it the suite still passes and that probe goes red.

**The symlink skip takes the shared files out of the lint entirely.** `skill_dirs` spans `plugins/*/skills/*` and `.claude/skills/*`, so `plugins/tend-ci-runner/shared/` is reached only through the two symlinks pointing into it — and those are now skipped. Drop the skip and the test reports the one live citation there: `references/draft-mode.md` in the Designed silence bullet, owned by `review` but named without an owner, so it resolves against `review-runs` and `review-reviewers` and fails. Before this branch the weaker name-in-some-skill check did cover it through the symlink; now nothing does, on a file every consumer's `review-runs` loads. Naming the owner in that bullet makes it resolve and the skip can go.

`uv run pytest generator/tests/test_repo_pins.py` is 43 passed with all three suggestions applied, and ruff is clean. Happy to push them if you'd rather not apply by hand.

generator/tests/test_repo_pins.py
max-sixty: Taken in f82abc79.

generator/tests/test_repo_pins.py
max-sixty: Taken in f82abc79.

plugins/tend-ci-runner/shared/review-gates.md
max-sixty: Taken in f82abc79.

tend-agent (approved):

PR #1269, opened 2026-09-17
skills: close the gaps #1268 left — cross-repo dispatch, the skill-file guard, and three stale references

max-sixty:
Five fixes #1268 left open. That PR landed the misplaced-guidance survey this branch also carried; this one now holds only what it didn't cover, on top of main.

**Nothing barred dispatching workflows in other repos.** #1268 cut the system prompt's "Act within this repository… or initiate cross-repo workflows" in favor of `running-in-ci`, but **Scope** there lists only PRs, pushes, and comments on existing threads. The bot's PAT carries `workflow` scope and cross-repo reach by design, so `gh workflow run -R other/repo` was unaddressed by any text. Scope's enumeration now includes workflow runs you dispatch; filing a new issue elsewhere stays governed by **Filing issues** in `other-repos.md`.

**The moved `review-reviewers` skill had left the bang-backtick guard.** The pre-commit hook scanned `plugins/` only. A backticked token starting with `!` crashes the Claude Code slash-command preprocessor before any turn runs, and `.claude/skills/review-reviewers/SKILL.md` is now the one file a workflow invokes directly as a slash command. The hook scans `.claude/skills/` too; its own comment records four earlier regressions.

**`skill-pr-workflow.md` told a session on tend it needed no read-only-mount workaround** "since bundled skills live under `plugins/`". That stopped holding for `review-reviewers` when it moved. The workaround turns on where the edited file sits, not which repo you are in, and the text now says so.

**The pre-existing-failure check couldn't answer its own question.** It hardcoded `branch=main`, and it listed the last three runs of *any* workflow with no name attached — run against this repo, that returned three `tend-notifications` runs from a month earlier and no `ci` run at all. It now resolves `defaultBranchRef` and filters by the workflow the failure belongs to (`gh run list --workflow`), which returns that workflow's latest default-branch runs with links; verified against this repo.

**`review-reviewers` cited a heading that never existed** ("Recording below-threshold findings"); it points at **Evidence accumulation**.

Tests (1159) and all 13 pre-commit hooks pass.


tend-agent (commented):
**#1268 is rewriting the same files.** `misplaced-guidance-survey` also moves `review-reviewers` to `.claude/skills/`, also adds the `review-gates.md` symlink there, also deletes the bundled one, also folds the Non-issues block into `plugins/tend-ci-runner/shared/review-gates.md`, and also adds a `fixing.md` — at `triage/references/fixing.md` rather than `running-in-ci/references/fixing.md`. Neither branch contains the other.

They don't only overlap, they disagree about `review-gates.md`. #1268 deletes Gate 3 and rewrites the three Non-issues bullets to drop tend's own incidents; this PR moves those bullets across verbatim, so the version it ships keeps `Gate 3 classifies them waste-class` — pointing at a heading #1268 removes. Whichever merges second, resolving that file is a decision about which version of the block ships, not a mechanical merge, so the order is worth settling before either lands.

**The one-home rule has three homes here.** The `${CLAUDE_PLUGIN_ROOT}`-vs-checkout rule is now stated in `CLAUDE.md`, in the `review-reviewers.yaml` header comment, and again at the top of the skill. Inline suggestion below on the copy that instructs nobody: it loads into every `review-reviewers` run, and those runs execute commands that are already written the right way.

**`running-tend` is the counterexample the new `CLAUDE.md` paragraph leaves standing.** Under *Usage analysis* it reaches `list_recent_runs.py` — the same script `review-reviewers` now resolves through the checkout — via `SCRIPTS="${CLAUDE_PLUGIN_ROOT:-/home/tend-sandbox/tend-marketplace/plugins/tend-ci-runner}/scripts"`, with a comment recording that Claude leaves that variable unset so the hard-coded installed-plugin path is what a Claude session uses. It is the other skill in `.claude/skills/`, so the rule lands with the one file that contradicts it sitting beside it. Happy to push that change into this PR if you want it here rather than as follow-up.

.claude/skills/review-reviewers/SKILL.md
```diff
@@ -14,6 +14,8 @@ Analyze tend's CI behavior on the target repo over the window Step 1 returns. Fo
 
 Load `/tend-ci-runner:running-in-ci` first — it contains CI security rules, the index of every reference file, and polling conventions. This skill opens PRs and issue comments on tend, so those rules apply.
 
+This skill is repo-local, so its scripts come from the checkout rather than the pinned action's installed plugin — `$(git rev-parse --show-toplevel)` below, not `${CLAUDE_PLUGIN_ROOT}`, which Claude substitutes only into a bundled skill's text. Keep it that way: a release-pinned script beside a skill that tracks `main` skews the moment one of them changes a CLI.
+
```
tend-agent: This is the third statement of the rule — `CLAUDE.md` carries it for whoever writes the next repo-local skill, and the `review-reviewers.yaml` header carries it (plus the Codex-rewording caveat that is genuinely workflow-local) for whoever edits the workflow. A `review-reviewers` session reads this one on every run and changes nothing because of it: the script paths below are already written the way it prescribes.

```suggestion
```

plugins/tend-ci-runner/skills/running-in-ci/references/ci-monitoring.md
```diff
@@ -29,11 +29,12 @@ Run this command in the foreground and allow at least 10 minutes — the poll ru
...
 
-Before calling a failure pre-existing (**Grounded Analysis** in `SKILL.md`), check main-branch CI:
+Before calling a failure pre-existing (**Grounded Analysis** in `SKILL.md`), check the default branch's CI:
 
 ```bash
-gh api "repos/{owner}/{repo}/actions/runs?branch=main&status=completed&per_page=3" \
-  --jq '.workflow_runs[] | {conclusion, created_at: .created_at}'
+DEFAULT_BRANCH=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')
+gh api "repos/{owner}/{repo}/actions/runs?branch=$DEFAULT_BRANCH&status=completed&per_page=3" \
+  --jq '.workflow_runs[] | {conclusion, created_at}'
```
tend-agent: Without `.name` the rows are unattributable, and three rows is under one minute of history on a bot-active repo, so the workflow you are classifying is usually not among them.

```suggestion
gh api "repos/{owner}/{repo}/actions/runs?branch=$DEFAULT_BRANCH&status=completed&per_page=20" \
  --jq '.workflow_runs[] | {name, conclusion, created_at}'
```

tend-agent (commented):
The default-branch check now resolves the right branch, but on the line it rewrites it still can't answer the question the gate asks. `per_page=3` against the unfiltered runs endpoint returns the last three runs of *any* workflow, and the jq projection drops `.name` — so on a repo where the bot is active the window fills with `tend-*` runs and the reader gets conclusions with nothing identifying them. Run against this repo now it returns two bare `skipped` rows and one `success`; none of the three is the `ci` workflow, and nothing in the output says so. **Grounded Analysis** makes this check mandatory before publishing a "pre-existing" claim, so what it costs is a wrong outward claim rather than a missing one. Suggestion inline.

The other four are verified — **Evidence accumulation** is the real heading, the widened guard picks up all five `.claude/skills/` files, and the read-only-mount rule now turns on the path it actually depends on. I've resolved the earlier thread: this branch takes #1268's version of that file, which no longer carries the restated rule.

plugins/tend-ci-runner/skills/running-in-ci/references/ci-monitoring.md
max-sixty: Reproduced, and it was worse than stale-by-a-minute: against this repo the old recipe returned three `tend-notifications` runs from 2026-08-15 and no `ci` run. Widening to 20 rows would still leave the right workflow to chance, so 109f2cfa filters by it instead — `gh run list --branch "$DEFAULT_BRANCH" --workflow "<workflow>" --status completed --limit 5 --json conclusion,createdAt,url` — which returns that workflow's latest default-branch runs, newest first, with links.

tend-agent (approved):

PR #1272, opened 2026-09-17
skills: soften the fix gate's failing-test rule, and correct tend-mention's TODO premise

max-sixty:
Two follow-ups to #1268 and #1269.

**The fix gate's failing-test rule read as absolute for every workflow.** `triage/references/fixing.md` is where the References table sends any session about to write a fix, and it said "You MUST have a failing test before writing any fix". `ci-fix`'s escalated-fault path asks for a mitigation PR — pin the runner image, skip a leg — for a fault that still reruns green, which has no test to write, so a ci-fix session following both was told to open that PR and forbidden to write it. The rule now reads "Generally, have a failing test before writing a fix", the fix-attempt bar's reproduction condition says "usually" rather than presuming the test, and `triage`'s Step 6 pointer names "the reproduction gate" instead of restating "the failing test you must already have".

**`TODO.md`'s `tend-mention` entry claimed a problem #1268 solved.** It said rules with mention's shape had nowhere to live but the file every session loads; #1268 moved them to `running-in-ci`'s `references/trigger-context.md`. The entry now says what remains: mention's prompt is the one exception `test_repo_pins.py` allows to the invariant that every agent-invoking workflow names a skill, and a `mention` skill earns its cost only if mention-only rules start accumulating in `running-in-ci` again. The entry stays because that lint cites it.


tend-agent (commented):
Two loose ends, both on the first half.

The Step 6 pointer in `plugins/tend-ci-runner/skills/triage/SKILL.md` still carries the absolute this PR removes — it describes `references/fixing.md` as holding "the failing test you must already have". A triage session reads that line before it opens `fixing.md`, so the contradiction survives one level up, and it is the restatement CLAUDE.md's "A rule has one home" (under "Which file") warns about: the pointer should name the gate, not re-state it. It's outside this diff — say the word and I'll push the narrowing commit onto this branch.

The softening also may not reach the case that motivated it. `ci-fix`'s escalated fault is one that "still reruns green", and the fix-attempt bar below still opens with "Only attempt a fix if **all** of these hold: Bug is clearly reproducible" — so the mitigation PR is blocked by the first condition whatever the parenthetical says. Naming the condition instead of hedging with "Generally"/"usually" would unblock that case and keep the gate from reading as optional to a session with an unreproduced bug in front of it. Suggestions inline for both lines.

plugins/tend-ci-runner/skills/triage/references/fixing.md
```diff
@@ -9,11 +9,11 @@ The gates every fix attempt clears, whichever workflow you are running.
 3. **Write a failing test** — Add a test to the appropriate *existing* test file that demonstrates the bug. Don't create new test files.
 4. **Run the test** to confirm it fails. Use the test commands from the project's instruction files.
 
-You MUST have a failing test before writing any fix. A fix without reproduction evidence is not a conservative fix: report what you established instead. If the test passes, the bug may already be fixed — say so.
+Generally, have a failing test before writing a fix. A fix without reproduction evidence is not a conservative fix: report what you established instead. If the test passes, the bug may already be fixed — say so.
```
tend-agent, line 12:
```suggestion
Have a failing test before writing a fix, unless the fault has no test to write — a mitigation for an infrastructure fault that reruns green, per **Repeat-occurrence escalation** in `/tend-ci-runner:ci-fix`. A fix without that evidence is not a conservative fix: report what you established instead. If the test passes, the bug may already be fixed — say so.
```

plugins/tend-ci-runner/skills/triage/references/fixing.md
```diff
@@ -9,11 +9,11 @@ The gates every fix attempt clears, whichever workflow you are running.
...
 
 **Only attempt a fix if all of these hold:**
 
-- Bug is clearly reproducible (the test written above fails)
+- Bug is clearly reproducible (usually, the test written above fails)
```
tend-agent, line 16: Same point one line down: this is the condition that actually gates the escalated-fault mitigation, so the carve-out has to be stated here too.

```suggestion
- Bug is clearly reproducible (the test written above fails), or it recurs on the record with no test to write
```

plugins/tend-ci-runner/skills/triage/references/fixing.md
max-sixty: Not taking this one: the maintainer chose to relax the wording rather than name a ci-fix carve-out, so the reproduction condition stays a judgment the session makes rather than a list of exempt cases. The Step 6 pointer in `triage/SKILL.md` is fixed in 3c5799e2 — it now names "the reproduction gate", matching the References table.

plugins/tend-ci-runner/skills/triage/references/fixing.md
max-sixty: Same call as the thread above: the wording relaxes the condition rather than naming the ci-fix case.

tend-agent (approved):

PR #1276, opened 2026-09-17
fix(notifications): count only a settled CONFLICTING in the bot-PR conflict gate

### LINKED ISSUE #1274: tend-notifications: the conflict pre-check counts a lazily-computed UNKNOWN mergeable as a conflict, so every merge into main boots an agent session
## Problem

`tend-notifications`'s pre-check counts a bot PR as possibly conflicted whenever its `mergeable` is anything other than `MERGEABLE` ([`notifications_check.py:131`](https://github.com/max-sixty/tend/blob/c874308b74067f8eefb999f1aacbed248320b00a/generator/src/tend/templates/notifications_check.py#L131)):

```python
return sum(
    pr.get("mergeable") != "MERGEABLE" and not _is_deferred(pr, bot)
    for pr in nodes
    if isinstance(pr, dict)
)
```

`mergeable` is lazily computed, so the first read after the base branch moves is `UNKNOWN` — the same GitHub behaviour #897 established, with a reproduction against this repo's adopter. `UNKNOWN != "MERGEABLE"`, so every open bot PR counts as conflicted for the first poll after any merge into the default branch. `conflict_count` then opens the gate on its own (`count == 0 and conflict_count == 0` is the only skip), and a full agent session boots to test-merge a set of PRs that are all clean.

#897 fixed the mirror of this in `nightly`, where `== "CONFLICTING"` turned `UNKNOWN` into a silent "no conflict". The notifications gate errs the opposite way, and the cost lands on every poll cycle that follows a merge rather than once a night.

## Evidence

Three `tend-notifications` sessions in `PRQL/prql` in the 24 h to 2026-09-17T07:51Z booted with an **empty notification snapshot** — so the conflict count was the only reason they ran — and each found every PR clean:

| Run | Reported count | Open bot PRs | Result |
| --- | --- | --- | --- |
| [35127064830](https://github.com/PRQL/prql/actions/runs/35127064830) | 3 | 3 | all 3 test-merged clean |
| [35159538190](https://github.com/PRQL/prql/actions/runs/35159538190) | 1 | 2 | both clean |
| [35194939552](https://github.com/PRQL/prql/actions/runs/35194939552) | 4 | 4 | all 4 clean, GitHub re-reporting `MERGEABLE` |

$1.60 at API list prices for the three. Each session diagnosed the signal correctly and touched nothing — the sessions are well-behaved; they just had no reason to exist. The shape has been recorded in that repo's `review-runs` tracker in eight windows since 2026-09-05, always with the same outcome, and always following a merge into `main`.

The 35194939552 row is the clearest: its count of 4 was the plain number of open bot PRs, and the session's own re-query returned `MERGEABLE` for all four.

## Proposed fix

Count only the state that means what the gate is asking:

```python
pr.get("mergeable") == "CONFLICTING" and not _is_deferred(pr, bot)
```

The trade: a PR that is genuinely conflicted but reads `UNKNOWN` on one poll is picked up on the next one instead. On the adopter above that is a ~20-minute delay, and `nightly`'s conflict pass is a second backstop. That seems like the right side to fail on for a gate whose false positive is a whole agent session and whose false negative is one skipped cycle.

The alternative — re-query the `UNKNOWN` subset before counting, as #897's fix does in `nightly` — keeps detection immediate but adds a retry to a pre-check that is meant to be cheap. Happy either way; flagging the one-line version because the gate's failure mode is spend rather than a missed conflict.

tend-agent:
## Problem

`mergeable` is computed lazily, so the first read after the base branch moves returns `UNKNOWN` for every open PR and enqueues the computation. The notifications pre-check counted any value that was not `MERGEABLE`, so after each merge into the default branch the next poll counted every open bot PR as possibly conflicted. `conflict_count` opens the gate on its own — `count == 0 and conflict_count == 0` is the only skip — so a full agent session booted to test-merge PRs that were clean, once per merge rather than once per conflict. #1274 records three such sessions in a consumer repo within 24 h, each with an empty notification snapshot and every PR test-merging clean.

## Solution

Count a settled `CONFLICTING`. A conflict hidden behind an `UNKNOWN` is counted on the next poll, which reads the value the first read enqueued — a delay of one cron period, 15 minutes by default. Nothing downstream relies on the gate for correctness: `/tend-ci-runner:resolve-conflicts` re-lists the bot's PRs and test-merges each one locally, so a session booted for any reason still catches a PR whose `mergeable` was cold, and nightly runs the same resolver as a second pass.

This reverses the deliberately conservative reading the gate has carried since #1108, in the direction #897 warned against for `nightly`. The two cases differ in what the miss costs: `nightly` read `mergeable` once a night and treated `UNKNOWN` as clean for 24 hours, while this gate re-reads every 15 minutes and hands the authoritative check to the resolver. The residual exposure is a repository whose default branch moves shortly before most polls, where detection can slip more than one cycle; nightly's pass bounds that.

The alternative — re-querying the `UNKNOWN` subset before counting — keeps detection immediate but puts a retry loop inside a pre-check whose whole point is to be cheap.

## Testing

`test_notifications_check_ignores_an_unsettled_mergeable` (renamed from `test_notifications_check_boots_for_unknown_or_conflicting_bot_prs`) feeds the pre-check one `MERGEABLE`, one `UNKNOWN`, and one `CONFLICTING` PR: it counted 2 before this change and counts 1 after. The generated-workflow snapshots carry the inlined script, so they move with it.

Closes #1274 — automated triage


tend-agent (commented):
`test_notifications_check_suppresses_only_the_marked_bot_head` no longer covers the path its name claims. Its only suppression-positive case — PR 22, carrying a bot marker on the live `headRefOid` — is `mergeable: "UNKNOWN"`, so the new filter drops it before `_is_deferred` is ever consulted. The other three PRs in that fixture are all expected to count (stale head, wrong author, marker not on the last line), so nothing left in the suite exercises a marker that actually suppresses.

Evidence: stubbing `_is_deferred` to `return False` fails that test on `c5565a1` (counts 4, expects 3) and passes on this head. The deferral marker is what keeps a conflict a maintainer parked from re-booting a session every 15 minutes, and it is now untested.

One line fixes it: give PR 22 `mergeable: "CONFLICTING"`. The expected count stays 3, and the stub fails again. Pushing that to the branch.

PR #1280, opened 2026-09-17
prompt: give the shared system prompt one home for both harness directives

max-sixty:
Two rules that every bot session reads lived in Python string constants that only the Claude harness prepended: an autonomy directive ("no human is available… permissions are pre-approved") and a skill-loading directive ("Use `/tend-ci-runner:running-in-ci` before starting work"). Neither was visible beside the two markdown files that carry the rest of the shipped guidance, and `CLAUDE.md` had to name them specially in its shipped-guidance list to compensate.

The autonomy directive moves into `shared/system-prompt.md` verbatim. That file reaches both harnesses — the Claude action appends it to the system prompt, the Codex runner stages it into `AGENTS.md` — so Codex sessions now get it for the first time. The text holds there: `codex exec` runs under `--dangerously-bypass-approvals-and-sandbox`, so "permissions are pre-approved; tool calls execute without confirmation" is accurate.

The skill-loading directive was stated three times, once per file, differing only in invocation syntax: the Claude constant, a paragraph in `codex/agents-tail.md`, and a bullet in `shared/system-prompt.md` itself. It now has one home in the neutral file, written as `${SKILL:running-in-ci}`, with a new `shared/steps/_prompt.py` substituting `/tend-ci-runner:` or `$` per harness. That module also absorbs the `${BOT_NAME}` substitution, which the two harness paths had as duplicate copies of the same function.

`SKILL_PREFIX` restates the mapping `Config.default_prompt` applies to generated workflow prompts, because the generator is not installed on the runner and neither side can import the other. Two pins in `test_repo_pins.py` cover that: one asserts the two mappings agree and that both name every known harness, the other resolves each `${SKILL:...}` name against the bundled skill on disk and fails on a token malformed enough to survive rendering — either way the wrong name otherwise reaches the model rather than failing.

Templating stays hand-rolled deliberately: rendering happens at job time under `/usr/bin/python3 -E -s`, before tend's `uv` exists, so a third-party engine is not available; and the prose contains literal `$GH_TOKEN` and `$<skill-name>`, which any `$`-based engine (including the stdlib's `string.Template`) would demand be escaped.

The third commit fixes fallout that only CI could catch: the sandbox executes from a staged bundle listing exactly the trusted files it may run, and `codex/runner.py`'s new `import _prompt` needed `_prompt.py` added to that list. A test now walks each bundled module's imports and fails on any sibling the bundle omits.


generator/tests/test_repo_pins.py
```diff
@@ -790,3 +792,52 @@ def test_every_workflow_prompt_names_a_skill_that_exists() -> None:
...
+    prompt = _prompt_module()
+    files = [REPO_ROOT / "shared/system-prompt.md", REPO_ROOT / "codex/agents-tail.md"]
+
+    for path in files:
+        for harness in prompt.SKILL_PREFIX:
+            rendered = prompt.render(path.read_text(), bot_name="bot", harness=harness)
+            assert "${SKILL" not in rendered, (
+                f"{path.relative_to(REPO_ROOT)} has a malformed skill token; "
+                "it must read ${SKILL:<lowercase-skill-name>}"
+            )
```
tend-agent, line 862:
```suggestion
    for path in files:
        text = path.read_text()
        for name in prompt._SKILL_REF.findall(text):
            skill = REPO_ROOT / "plugins/tend-ci-runner/skills" / name / "SKILL.md"
            assert skill.is_file(), (
                f"{path.relative_to(REPO_ROOT)} invokes `{name}`, which is not "
                f"a skill at {skill.relative_to(REPO_ROOT)}"
            )
        for harness in prompt.SKILL_PREFIX:
            rendered = prompt.render(text, bot_name="bot", harness=harness)
            assert "${SKILL" not in rendered, (
                f"{path.relative_to(REPO_ROOT)} has a malformed skill token; "
                "it must read ${SKILL:<lowercase-skill-name>}"
            )
```

shared/steps/test_launch_sandbox_runtime.py
```diff
@@ -197,3 +199,34 @@ def test_runner_cancellation_is_raised_through_the_reap_path() -> None:
...
+
+def test_runtime_bundle_carries_every_module_it_imports() -> None:
+    """The bundle is what the sandbox executes from; a missing import is a crash.
+
+    Nothing in the sandbox can reach back to the action checkout, so a bundled
+    module that imports a sibling left out of `RUNTIME_STEP_FILES` fails at
+    `import` inside SRT — a green unit suite and a red agent turn.
+    """
+    steps = Path(__file__).resolve().parent
+    bundled = {name for name in RUNTIME_STEP_FILES if name.endswith(".py")}
```
tend-agent:
```suggestion
    bundled = {name for name in launch.RUNTIME_STEP_FILES if name.endswith(".py")}
```

tend-agent (commented):
`test_shipped_prompt_files_leave_no_unrendered_skill_token` catches a malformed token but not a wrong name — and a wrong name is the likelier mistake, since renaming a bundled skill leaves the reference syntactically perfect. `${SKILL:runing-in-ci}` matches `[a-z0-9-]+`, renders to `/tend-ci-runner:runing-in-ci` and `$runing-in-ci`, and both new tests pass; every shipped session then opens with an instruction to load a skill that does not exist. The generator's half of the same mapping already has this check — `test_every_workflow_prompt_names_a_skill_that_exists` resolves the named skill on disk — so the runtime half is the gap.

The inline suggestion adds it; I ran it against that exact typo (fails) and against the current tree (passes). It reaches `prompt._SKILL_REF`; renaming the regex to `SKILL_REF` would suit if you would rather the test not touch a private name, and the docstring wants a word about the name check either way.

Second, smaller point inline on the bundle guard: it reads the test module's own `RUNTIME_STEP_FILES` copy rather than `launch.RUNTIME_STEP_FILES`. Divergence is caught today — `configure` writes the test copy's names and `stage_runtime_bundle` raises on a missing source — so this is about the guard asserting against the constant it exists to protect, not a live hole.

generator/tests/test_repo_pins.py
max-sixty: Applied in e0d21311. The test now resolves every `${SKILL:...}` name against `plugins/tend-ci-runner/skills/<name>/SKILL.md`, alongside the malformed-token check, and I took the rename: `_SKILL_REF` is now `SKILL_REF`, so the test isn't reaching for a private name. The docstring says why both checks are there. Confirmed it fails on `${SKILL:runing-in-ci}` and passes on the current tree.

You were right that the name is the likelier mistake — a rename leaves the token syntactically perfect, and the generator's half already had this check.

shared/steps/test_launch_sandbox_runtime.py
max-sixty: Applied in e0d21311 — the guard reads `launch.RUNTIME_STEP_FILES` now, so it asserts against the constant it exists to protect rather than the test module's copy.

tend-agent (approved):

tend-agent (approved):

PR #1282, opened 2026-09-17
ci: run the suite on a schedule, so archive drift fails a check

max-sixty:
Today's outage had a detection gap, not just a code defect. `test-sandbox` installs bubblewrap from the Ubuntu archive on every run, so it is the check that would have caught the CVE-2026-87766 update changing how bwrap resolves a bind destination. It runs only on `push` to main and on `pull_request`, and main had gone unpushed for two hours when the update landed — so the first signal was every consumer's sessions dying at sandbox launch, about 75 minutes later.

Everything else tend could have noticed it with boots an agent inside the same sandbox: `tend-ci-fix` fires on a red `ci`, `tend-review-runs` reads last night's runs, and the outage issues are filed by the action's own failure path. The failure disabled its own alarm. A scheduled `ci` run is the only watch that sits outside the sandbox, and GitHub mails the repo owner when a scheduled workflow fails on the default branch, with no agent in the path.

Pinning the package instead is not available. Ubuntu keeps one version per pocket: `0.9.0-1ubuntu0.1` had already left the archive when I went looking for it hours after `0.9.0-1ubuntu0.2` shipped, so a version pin would have failed the install outright. Pinning the runner image would not have helped either — the image was byte-identical across the break; the package came from the live archive.

Every job in this workflow finishes in under a minute, so the schedule runs the whole thing rather than a conditional subset. Every six hours bounds the dark window at six hours and costs about four job-minutes a day.

`review` will fail here until tend regenerates its own workflows onto 0.2.10 — the reviewer still runs the broken `@0.2.9` pin.


tend-agent (commented):
The premise checks out: `test-sandbox` plants its own workspace container and, per #1277, mirrored the `0o711` that the CVE-2026-87766 bwrap change broke — so this job really is the one that would have gone red on the archive update, and it really is the only watch in the repo that doesn't boot an agent inside the thing being watched.

Two things to weigh before merging.

**`schedule` in a fork's own Actions.** Every other scheduled workflow here — `tend-nightly`, `tend-notifications`, `tend-review-runs`, `tend-weekly`, `publish-site`, `worker-deploy`, `integration-secrets` — carries `if: github.repository_owner == 'max-sixty'`, and `CLAUDE.md`'s **Fork guard** states the rule as "workflows whose triggers can fire from a fork's own Actions (`schedule`, `workflow_dispatch`, `workflow_run`, `issues`)". `ci.yaml` has no guard on any of its six jobs, and until now it didn't need one: `push` and `pull_request` firing in a fork is a fork owner getting CI on their own work, which is the point. `schedule` isn't that — it's four unsolicited full runs a day in each of the 7 forks that has Actions enabled, with a failure mail to the fork owner when the archive drifts under them. The catch is that the guard is job-level, so expressing it costs six copies of a condition, which `CLAUDE.md` would read as a configuration scheme rather than one knob. I don't think that trade is mine to call; flagging it as the decision rather than a requested change.

**The cost figures are low.** On the latest `ci` run on `main` ([35266035795](https://github.com/max-sixty/tend/actions/runs/35266035795)) `test-sandbox` took 96s, so "every job finishes in under a minute" doesn't hold, and the six jobs total ~3.4 job-minutes rather than one. Four runs a day is ~13.5 job-minutes, not four. The conclusion is unchanged — 13 job-minutes a day to bound a fleet-wide dark window at six hours is still obviously worth it — but the description is what a later reader will judge the cadence against, so it's worth correcting there rather than leaving the next person to re-measure.

The red `review` check is the outage, not a verdict on this diff; #1283 has since regenerated onto 0.2.10, so the next push to this branch will get a normal review. This one is from the `tend-review-runs` sweep picking up the review that died at sandbox launch.

PR #1286, opened 2026-09-17
ci: correct the scheduled-run notification claim

tend-agent:
The `schedule` block added in #1282 is tend's only watch that sits outside the sandbox, so where its failure alarm lands is worth getting right. The comment claimed the alarm "mails the repo owner". GitHub routes it by account identity instead: notifications go to "the user who initially created the workflow", move to a different user who "updates the cron syntax, in the `schedule` event in the workflow file", and move again to whoever re-enables a scheduled workflow that was disabled ([docs](https://docs.github.com/en/actions/concepts/workflows-and-actions/notifications-for-workflow-runs)). That resolves to the repo owner today, who authored the commit, but the binding follows the cron line rather than the owner role — and this file is hand-maintained, so a later edit by any account takes the alarm with it, silently and without turning anything red. The comment now states that rule, plus the operational consequence: keep that line human-edited, because the bot's own notifications route back through the sandbox this schedule exists to watch from outside.

That consequence is also why this PR no longer touches the cron itself. It originally offset the minute to `37 */6 * * *`, on the grounds that the top of the hour is the window GitHub names for scheduler delay. But the edit would have been authored by `tend-agent`, which under the rule above is exactly what moves the alarm off a human account — for a benefit the change itself described as a small edge, against losing the watch with no red signal. The cron is unchanged from `main`, so the binding stays with the owner, and the diff is comment-only. If the offset is worth having, it wants a commit from a human account.

`uv tool run pre-commit run --all-files` passes, actionlint included.

Outside this diff, worth confirming once: the alarm exists at all only if GitHub Actions notifications are enabled on the receiving account, and "only notify for failed workflows" is a separate toggle — without it this cron delivers four success emails a day.


tend-agent (commented):
The comment correction is right, but the cron-minute change trips the rule the comment now states.

`git log -L '/schedule:/,+3:.github/workflows/ci.yaml'` shows the cron line's last editor is `max-sixty` (339aafb, #1282). This commit rewrites that line under the `tend-agent` account, so the binding the comment documents — notifications follow whoever "updates the cron syntax, in the `schedule` event in the workflow file" — moves to the bot. That is exactly the case the comment's own caveat, "so long as that account is a human's", exists to exclude, and this diff is what makes it false. The bot's notifications are drained by `tend-notifications`, which boots an agent inside the same sandbox the comment above says disabled its own detection, so the alarm would sit unread or route back through the thing it exists to watch — with nothing turning red either way.

Whether GitHub rebinds on the commit author or on the merge pusher is the uncertain part; the asymmetry is not. The offset buys "a small edge" against top-of-hour scheduler delay, in this PR's own framing, against silently losing the repo's only out-of-sandbox watch.

So I'm restoring `0 */6 * * *` and keeping the comment correction, with the caveat tightened into the rule a future editor needs — keep that line human-edited. If the minute offset is worth having, it wants a commit from a human account rather than this one.

tend-agent:
Closing as superseded — nothing here is left to land.

#1289 rewrote this same comment block and carried the correction verbatim: main's `.github/workflows/ci.yaml` now reads "a failing scheduled run on the default branch mails whoever last edited the cron line below, or last re-enabled the workflow. Keep both a human's: … silently and without turning anything red", which is this branch's wording character-for-character.

What this branch still carries is the *pre-pin* framing that #1289 deliberately replaced, so merging it would regress main on two points:

- The opening premise — "`test-sandbox` installs bubblewrap from the Ubuntu archive on every run, and the archive moves without a commit here" — is no longer true. `PACKAGES_RESOLVED_AT` pins the archive.
- The closing "Pinning the package is not an option: Ubuntu keeps one version per pocket…" is the claim #1289 falsified by pinning the three packages against `snapshot.ubuntu.com`.

The branch conflicts with main on that block, and resolving the conflict in main's favour leaves an empty diff.

Issue #1285, opened 2026-09-17
Regeneration onto a fixed release runs inside the agent session, so a boot-blocking defect leaves consumers with no self-heal path


tend-agent:
A defect that stops the agent booting takes out the mechanism consumers would use to pick up its fix, so every consumer stays broken until a human runs `uvx tend@latest init` by hand. Today's bubblewrap outage is the third instance of the precondition in nine days, and the first where a tend session in this repo could observe the whole thing.

## The loop

`macros.yaml.j2` states the design: the action ref pins to an immutable release tag, and "consumers move forward via the nightly `uvx tend@latest init` regen pulling a newer generator, which restamps a newer tag as a reviewable diff in their own repo." That regen runs inside the agent turn — `nightly/SKILL.md` has the nightly session call `nightly_workflow_update.py prepare`, edit the prepared worktree, run `uv tool run tend@latest init` in it, and `ship` — so it needs a booted agent.

When the defect is in the boot path itself, the agent never starts, the nightly never regenerates, and the pin never moves. There is no other path: `tend-nightly` is the only workflow that regenerates, and nothing in the generated workflows runs `init` outside the sandbox.

## Why this isn't a one-off

Three fleet-wide boot failures since 2026-09-08, each killing sessions before the first turn:

| When | Cause | Recovered by |
|---|---|---|
| 2026-09-08 | 0.2.2 shipped `shared/steps/install-uv.sh` non-executable across the sandbox boundary | 0.2.4 + hand regeneration |
| 2026-09-08 | 0.2.3 published an action cache missing `shared/steps/sandbox_runtime.mjs` | 0.2.4 + hand regeneration |
| 2026-09-17 | bubblewrap `0.9.0-1ubuntu0.2` (CVE-2026-87766) made a `0o711` bind-destination parent unopenable — #1277, #1279 | 0.2.10 + hand regeneration (#1283) |

The first two are release defects, where "don't ship it" is at least a conceivable answer. The third isn't: nothing in tend changed. The Ubuntu archive moved under a running fleet, so no amount of pre-release care would have prevented it, and rolling the pin *backwards* wouldn't have helped either — Ubuntu keeps one version per pocket and the superseded package was already gone.

In each case this repo recovered in under an hour because a human was watching. A consumer whose maintainer is asleep recovers whenever they next look at their Actions tab.

## What it costs

[...]

PR #1288, opened 2026-09-17
fix(report-failure): name a stale action pin as the outage tracker's remedy

tend-agent:
The `tend-outage` tracker now names `uvx tend@latest init` as the remedy when the pin that failed is behind tend's newest release. It renders under the run's row, like this:

> Tend `0.2.10` is released and the workflow that failed pins `0.2.9`. A failure that lands before the bot's first turn is usually cleared by moving that pin: run `uvx tend@latest init` in a checkout and commit the regenerated `.github/workflows/tend-*.yaml`. `tend-nightly` does this itself where it runs, but only once a run gets far enough to start the agent.

This is the parenthetical of option 3 in #1285, and it holds under options 1 and 2 as well — a mechanism that moves the pin automatically does not remove the reader's need to know what moved and why. It does not settle that decision, which is still a call about the security model.

**Why the tracker.** A defect in the boot path takes out the regeneration that would pick up its fix, and `tend-nightly`'s regen runs inside the agent turn. So a fleet-wide boot failure leaves the pin where it is until a human runs the command. The tracker is the only artifact a third-party consumer gets in that window: without this line they have to derive the remedy from tend's source, or wait for us to notice. `Report failure` is the step that can say it — it is gated on `failure()`, runs outside the agent, and holds `github_token` for the one release read.

**Two constraints from the thread.**

- The note is emitted per run rather than baked into the seed body. The release that ends an outage normally lands while the outage is still running, so at file time there is often no newer release to name — for the 2026-09-17 incident there was none. It rides the first row posted after the pin falls behind, once per tracker: an incident appends a row per stranded run and the command is the same each time, and anything already naming it (an earlier row, or a maintainer diagnosing in the thread) keeps it quiet.
- It appears only while the pin is **strictly behind** the newest release. A consumer already on the newest failed for some other reason, and regenerating would send them the wrong way. A ref that is not an `X.Y.Z` release tag — a branch, a SHA, a local path — is not ordered at all, and skips the release read entirely.

**Mechanics.** `github.action_ref` and `github.action_repository` resolve in a composite step's `env:` and not inside its `run:` body, where they expand to the empty string rather than failing ([actions/runner#2473](https://github.com/actions/runner/issues/2473)), so both harness actions pass them to the step. `test_report_failure_is_told_the_running_version` asserts that per harness: dropping either would leave the tracker silently unable to name the remedy with nothing else red. Every failure mode of the note — an absent variable, an unparsable ref, a failed release read — is silence, so the row itself is never at risk.

<details><summary>What the tests cover, and what CI cannot reach</summary>

Six cases in `shared/steps/test_report_failure.py`: a stale pin named (with the blank line that keeps the prose out of the table), a current pin silent, a non-release ref making no API call, a failed release read still recording the row, the note riding an appended row rather than only the seed body, and the once-per-tracker dedup landing before the release read.

The one thing CI cannot check is that the two `env:` values arrive non-empty from a real composite invocation, since that needs the action called at a released tag. The parity test pins the call shape; if the runner behaves otherwise the outcome is a missing note, not a wrong one.

</details>


shared/steps/report_failure.py
```diff
@@ -144,6 +172,62 @@ def main() -> int:
...
+def _release(text: str) -> tuple[int, ...] | None:
+    """*text* as a release version, or ``None`` for anything that is not one.
+
+    Strict ``X.Y.Z`` — the one form the generator pins and a tend release
+    tags. A ref this cannot parse stays unordered rather than guessed at,
+    because a guess that lands the wrong way round tells a consumer on the
+    newest release to regenerate.
+    """
+    parts = text.strip().split(".")
+    if len(parts) != 3 or not all(part.isdigit() for part in parts):
```
tend-agent:
```suggestion
    if len(parts) != 3 or not all(part.isdecimal() for part in parts):
```

tend-agent (commented):
`_release` guards with `str.isdigit()`, which is `True` for 128 characters `int()` rejects — `²`, `¹`, `፩` and the rest of the non-decimal digit set. A ref or a `tag_name` shaped like `1.².3` passes the guard and then raises `ValueError` on `int(part)`, inside `Report failure`, on a job that is already red. The step dies before `create_and_reconcile`, so the outage row for that run is never filed and the tracker under-reports the incident — the one outcome the note is designed never to risk. `str.isdecimal()` is the exact predicate for what `int()` accepts, so an unparsable ref falls through to `None` and the note stays silent like every other failure mode here.

The trigger needs an odd tag on the action's own repository, so this is a robustness gap rather than something the fleet will hit. It is worth closing anyway because the design claim a later reader will lean on — that every failure mode of the note is silence — is what stops being true.

I'll push the one-word change on this branch; the queued run reviews it.

PR #1289, opened 2026-09-17
sandbox: pin the boundary's Ubuntu and npm packages to one recorded instant

max-sixty:
`install-sandbox-runtime.sh` took whatever the Ubuntu archive published that morning. On 2026-09-17 that was bubblewrap's CVE-2026-87766 fix, which changed how bwrap resolves a bind destination and failed every consumer's session at launch with nothing red in this repo. #1277 fixed Tend against the new bwrap; this stops the archive changing the sandbox boundary without a commit at all.

SRT's own dependency ranges floated the same way one level down — `zod ^3.24.1`, `commander ^12.1.0`, `node-forge ^1.4.0`, `@pondwader/socks5-server ^1.0.10` — in the tree that builds the bwrap argv and terminates the agent's TLS.

## What changed

Both installs now resolve as of one constant, `PACKAGES_RESOLVED_AT`:

- **apt** reads `snapshot.ubuntu.com/ubuntu/<instant>` for `noble`, `noble-updates` and `noble-security`, and installs `bubblewrap=0.9.0-1ubuntu0.2`, `socat=1.8.0.0-4ubuntu0.1`, `ripgrep=14.1.0-1` by name. The sources go in with `-o Dir::Etc::SourceList` plus a temp `Dir::State::Lists`, so a self-hosted runner keeps its own apt config and lists; the snapshot's `Release` verifies against the stock archive keyring, so this is not a weaker trust path than the live mirror.
- **npm** gets `--before "$PACKAGES_RESOLVED_AT"`, which freezes SRT's whole tree as of the same instant. That flag filters the top-level request too, so an SRT bump published after the instant needs the instant moved with it — the sweep table now says so on the `sandbox_runtime_version` row.

Two behaviour changes fall out of naming the versions. The staleness check reads `dpkg-query` rather than testing `/usr/bin/bwrap -x`: a runner carrying *some* bubblewrap satisfied the old presence check and still launched a different sandbox. And on a GitHub-hosted runner `--allow-downgrades` lets the pin win over an image that ships something newer.

**A self-hosted runner ahead of the pin is an error, not a downgrade.** That rollback would be host-wide and would outlive the job, so a box whose owner has already taken the next bubblewrap upload gets a message naming the versions instead — the restraint the AppArmor branch a few lines up already applies to host policy. A runner holding all three exactly still skips apt, and with it sudo. The codename guard runs ahead of that check, because a host off noble reads as ahead of the pin only because the pin is a noble build, and no bump here ever reaches it. An `EXIT` trap clears the apt state, since `lists/` is chowned to `_apt` and a failure would otherwise strand a tree the runner user cannot unlink.

## Bumping

`running-tend` gains a `### The sandbox boundary's packages` section and `refresh_sandbox_pins.py`, which reads the three pockets at a fresh instant, picks between them with `dpkg --compare-versions`, and rewrites all four constants in place, printing what moved. `test-sandbox` is the gate.

The cost of a pin is that an upstream security fix waits for a commit here, so the rule is explicit that a version which *will not* move is an issue naming the CVE it is holding back — never a quiet revert of the version alone.

## Also

The `ci` schedule keeps its cron. Its comment drops the claim that pinning was unavailable, and takes #1286's correction of who a failing scheduled run actually mails — whoever last edited the cron line or re-enabled the workflow, which has to stay a human's account for the path to be agent-free. #1286 touches only that block and is fully covered here, so it has nothing left to land; it is a bot PR, so I have left it open for its own session to retire.

## Verification

Full suite (1180) and pre-commit green. The apt path was driven end to end in Docker on `ubuntu:24.04`: cold install, a rerun proving the skip, a real downgrade off an older snapshot, the non-noble guard erroring on `bookworm`, the self-hosted refusal when the host is ahead, and `_apt` dropping privileges cleanly at 711 but not at 700. Three lints in `test_repo_pins.py` fail against dropping `--before`, a bare `apt-get install`, and an unversioned package.

**Measured on this PR's own first run**, which is what the earlier revision left open: the snapshot `apt-get update` pulls 21 indices, ~27.6 MB, in 10s at 2.6 MB/s — `noble/universe amd64 Packages` alone is 15 MB. The whole install step is 22.5s and `test-sandbox` finishes at 1m09s, against 1m02s on `main`. Nothing there needs narrowing.

Checked and deliberately left dynamic: `@openai/codex` and `@openai/codex-responses-api-proxy` have no dependencies and exact aliases to their platform binaries, so they are already deterministic; `worker/` and `site/` use `npm ci` against committed lockfiles; `https://claude.ai/install.sh` is Anthropic's installer for a version-pinned binary, run as the sandbox UID, where a vendored copy would rot against their layout.

Known limit: `refresh_sandbox_pins.py` resolves `binary-amd64` only. The snapshot publishes the same three versions for `binary-arm64` at this instant, so an arm64 runner installs cleanly, but if the arches ever diverge the pin becomes unsatisfiable there without the refresh showing it.


shared/steps/install-sandbox-runtime.sh
```diff
@@ -36,13 +61,52 @@ if [ ! -f "$npm_cli" ]; then
...
+  done > "$apt_state/sources.list"
+  apt_options=(
+    -o "Dir::Etc::SourceList=$apt_state/sources.list"
+    -o "Dir::Etc::SourceParts=$apt_state/parts"
+    -o "Dir::State::Lists=$apt_state/lists"
+    -o Acquire::Languages=none
+  )
+  # apt drops to `_apt` for the downloads and warns on every fetch into a
+  # directory that user cannot write.
+  /usr/bin/sudo /usr/bin/chown -R _apt "$apt_state/lists"
```
tend-agent: `_apt` still can't reach `lists/partial`: `mktemp -d` leaves the parent at 0700 owned by the runner user, so the `chown` doesn't buy the traversal. This PR's own `test-sandbox` run shows apt falling back:

```
W: Download is performed unsandboxed as root as file '/tmp/tend-apt.nDgtxj/lists/partial/snapshot.ubuntu.com_ubuntu_20260917T180000Z_dists_noble_InRelease' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)
```

So the fetcher parses the indices and signatures as root for exactly the packages this script is tightening. Opening the parent restores the privilege drop:

```suggestion
  # apt drops to `_apt` for the downloads and warns on every fetch it cannot
  # reach - including through mktemp's 0700 parent.
  /usr/bin/chmod 755 "$apt_state"
  /usr/bin/sudo /usr/bin/chown -R _apt "$apt_state/lists"
```

tend-agent (commented):
The pinning itself reads well, and `test-sandbox` being the gate is the right shape. Four points below — three are inline; this one is a decision rather than a fix.

**On a self-hosted runner, Tend now moves the host's own packages.** `--allow-downgrades` makes the pin win, and that change is host-wide and persists after the job: a self-hosted box that has already taken the *next* bubblewrap security upload gets rolled back to `$BUBBLEWRAP_VERSION` for every other workload on it, until the weekly sweep lands here. The AppArmor branch a few lines above deliberately refuses to touch self-hosted policy and prints a notice instead; the apt path makes no such distinction. One shape that keeps the boundary pinned without rolling back someone else's patch: when `TEND_RUNNER_ENVIRONMENT` isn't `github-hosted` and the installed version is *newer* than the pin, exit with the preinstall message the codename guard already carries. Worth a call either way — the current behaviour is invisible to whoever owns the runner.

**Overlaps with #1286**, which rewrites the same comment block: it corrects "mails the repo owner" to name whoever last edited the cron line, which is how GitHub routes scheduled-run failure mail. The inline suggestion here folds that correction in, so whichever lands second can drop its version of the paragraph — or #1286 closes as covered.

<details><summary>The measurement this PR left open, plus one minor note</summary>

From this PR's own `test-sandbox` run: the snapshot `apt-get update` fetched 21 indices — ~27 MB, of which `noble/universe amd64 Packages` alone is 15 MB — in about 10 s, with the whole install step at ~16 s and the job green at 1m10s. Nothing there needs narrowing.

`refresh_sandbox_pins.py` resolves `binary-amd64` only. The snapshot does serve `binary-arm64` for the same suites, and at this instant it publishes the same versions for all three packages, so an `ubuntu-24.04-arm` runner installs cleanly today. If one arch ever lags the other, the pin becomes unsatisfiable there and the refresh script won't show it.

</details>

.claude/skills/running-tend/SKILL.md
```diff
@@ -287,13 +295,46 @@ swamped run finishes nothing.
 | `uv_version` | both harness `action.yaml` files | move both defaults together, with `mitmproxy_version` |
 | `codex_version` | `codex/action.yaml`, `codex/refresh/action.yaml` | move both defaults together; `alpha` only for a fix not yet released |
 | `uv_build` | `generator/pyproject.toml` | its range must contain the uv doing the build; a stale one only warns during `uv build`, so only this sweep catches it |
+| `PACKAGES_RESOLVED_AT`, `BUBBLEWRAP_VERSION`, `SOCAT_VERSION`, `RIPGREP_VERSION` | `shared/steps/install-sandbox-runtime.sh` | move together; see below |
```
tend-agent, line 298: `npm --before` filters the top-level request too, not only SRT's transitive tree, so the instant and the SRT pin are now coupled:

```
$ npm install --before 2020-01-01T00:00:00Z @anthropic-ai/sandbox-runtime@0.0.76
npm error code ETARGET
npm error notarget No matching version found for @anthropic-ai/sandbox-runtime@0.0.76 with a date before 1/1/2020, 12:00:00 AM.
```

`sandbox_runtime_version` has no row here, so it falls to the composite-action `_version$` sweep and its "move to latest and let CI decide" default. The next SRT bump taken on its own reds `test-sandbox` with that notarget error, and the remedy — move `PACKAGES_RESOLVED_AT` in the same PR — isn't stated anywhere yet.

```suggestion
| `PACKAGES_RESOLVED_AT`, `BUBBLEWRAP_VERSION`, `SOCAT_VERSION`, `RIPGREP_VERSION` | `shared/steps/install-sandbox-runtime.sh` | move together; see below |
| `sandbox_runtime_version` | both harness `action.yaml` files | `npm --before` filters this exact version too, so a release published after `PACKAGES_RESOLVED_AT` needs the instant moved in the same PR |
```

generator/tests/test_repo_pins.py
```diff
@@ -195,6 +195,64 @@ def test_sandbox_runtime_pin_is_identical_in_actions_and_hosted_probe() -> None:
...
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout.split("\0")
+    offenders = sorted(
+        name
+        for name in tracked
+        if name
+        and not name.startswith("generator/tests/")
+        and re.search(r"apt-get\b.*\binstall\b", (REPO_ROOT / name).read_text())
```
tend-agent, line 253: `git ls-files` reports the index while `read_text` reads the worktree, and in a Tend session on a PR event `.claude/**` is restored from the base branch — so a path this PR adds there is tracked and absent from disk. This review's own `uv run pytest generator/tests/test_repo_pins.py` failed on it:

```
FAILED test_no_other_apt_install_escapes_the_pin - FileNotFoundError: [Errno 2] No such file or directory: '.../.claude/skills/running-tend/scripts/refresh_sandbox_pins.py'
```

GitHub's plain checkout passes, so the lint is green in CI and red in every bot session that runs the suite against a PR touching `.claude/**`.

```suggestion
        # `git ls-files` reports the index; a Tend PR session restores
        # `.claude/**` from the base branch, so a tracked path can be absent.
        and (REPO_ROOT / name).is_file()
        and re.search(r"apt-get\b.*\binstall\b", (REPO_ROOT / name).read_text())
```

.github/workflows/ci.yaml
```diff
@@ -3,17 +3,16 @@ on:
...
+  # `test-sandbox` is the only job that builds the real sandbox, and it still
+  # reaches hosts no commit here controls — snapshot.ubuntu.com, the npm
+  # registry, the runner image — even with every version pinned.  A bubblewrap
+  # security update on 2026-09-17 killed every consumer's sessions at launch;
+  # main had gone unpushed for two hours, so nothing ran and production was the
+  # first signal.  tend's other watches — `tend-ci-fix`, `tend-review-runs`,
+  # the outage issues — all boot an agent inside that same sandbox, so the
+  # failure disabled its own detection.  This schedule runs outside it, and a
+  # scheduled run failing on the default branch mails the repo owner with no
+  # agent in the path.
```
tend-agent: #1286 rewrites this same sentence, correcting the recipient: GitHub mails scheduled-run failures to the account that last edited the cron, not the repo owner. Since this PR rewrites the block anyway, folding the correction in here avoids landing the stale claim.

```suggestion
  # failure disabled its own detection.  This schedule runs outside it, and a
  # failing scheduled run on the default branch mails whoever last edited the
  # cron line below, so keep that a human's account or the alarm moves with it.
```

shared/steps/install-sandbox-runtime.sh
```diff
@@ -36,13 +61,73 @@ if [ ! -f "$npm_cli" ]; then
...
+    exit 1
+  fi
+
+  # `;` not `&&`: a host with no os-release should reach the message below
+  # rather than die on the `.` under `set -e`.
+  codename=$(. /etc/os-release 2>/dev/null; echo "${VERSION_CODENAME:-}")
+  if [ "$codename" != noble ]; then
+    echo "::error::The pinned sandbox capabilities are Ubuntu noble builds and this runner is '${codename:-unknown}'. Run Tend on ubuntu-24.04, or preinstall bubblewrap $BUBBLEWRAP_VERSION, socat $SOCAT_VERSION and ripgrep $RIPGREP_VERSION."
+    exit 1
+  fi
```
tend-agent: An Ubuntu 26.04 or Debian trixie self-hosted runner ships bubblewrap and socat newer than the pin, so `ahead` is non-empty and this guard fires before the codename check below. The operator is told to wait for Tend's weekly bump to reach those versions — which it never will, because the pin is a noble build and the sweep resolves against `snapshot.ubuntu.com`'s noble pockets. The codename guard is the accurate diagnosis for that host, so it should run first.

Swapping the two blocks costs nothing elsewhere: a github-hosted runner skips the `ahead` guard on the environment test regardless, and a self-hosted noble host still reaches it.

```suggestion
  # `;` not `&&`: a host with no os-release should reach the message below
  # rather than die on the `.` under `set -e`.
  codename=$(. /etc/os-release 2>/dev/null; echo "${VERSION_CODENAME:-}")
  if [ "$codename" != noble ]; then
    echo "::error::The pinned sandbox capabilities are Ubuntu noble builds and this runner is '${codename:-unknown}'. Run Tend on ubuntu-24.04, or preinstall bubblewrap $BUBBLEWRAP_VERSION, socat $SOCAT_VERSION and ripgrep $RIPGREP_VERSION."
    exit 1
  fi

  # A GitHub-hosted runner is disposable, so rolling it back onto the pin costs
  # its owner nothing. A self-hosted runner is someone's machine, where
  # replacing a security update they have already taken would outlive the job —
  # the restraint the AppArmor branch above applies to host policy.
  if [ "${#ahead[@]}" -gt 0 ] && [ "${TEND_RUNNER_ENVIRONMENT:-}" != github-hosted ]; then
    echo "::error::This runner is ahead of Tend's pin (${ahead[*]}) and Tend does not downgrade a self-hosted host. Wait for Tend's weekly bump to reach those versions, or run Tend on a runner holding bubblewrap $BUBBLEWRAP_VERSION, socat $SOCAT_VERSION and ripgrep $RIPGREP_VERSION."
    exit 1
  fi
```

tend-agent (commented):
Two points on the new self-hosted branch, both inline: the guard order misdiagnoses a self-hosted runner that isn't noble, and the snapshot's apt state survives an apt failure in a form the runner user can't clear.

shared/steps/install-sandbox-runtime.sh
```diff
@@ -36,13 +61,73 @@ if [ ! -f "$npm_cli" ]; then
...
+
+  # Scoped to this apt invocation with -o rather than written into /etc, so a
+  # self-hosted runner keeps both its own sources and its own fetched lists.
+  apt_state=$(/usr/bin/mktemp -d /tmp/tend-apt.XXXXXX)
+  /usr/bin/mkdir -p "$apt_state/lists/partial" "$apt_state/parts"
```
tend-agent, line 115: If `apt-get update` or `install` fails, `set -e` exits before the `sudo rm -rf` at the end of this block, and what's left isn't removable by the runner user: `lists/` and everything under it is `chown`ed to `_apt`, so unlinking those entries needs write on an `_apt`-owned directory. A github-hosted runner is thrown away either way, but on the self-hosted host the rest of this PR is careful not to mutate, each failed run strands the fetched index tree under `/tmp` where only root can clear it.

With the trap registered the explicit `sudo rm -rf` at the end of the block is redundant and can go.

```suggestion
  apt_state=$(/usr/bin/mktemp -d /tmp/tend-apt.XXXXXX)
  # The lists end up owned by `_apt`, so the runner user cannot remove this
  # tree without sudo; clean it up on the failure paths too.
  trap '/usr/bin/sudo /usr/bin/rm -rf "${apt_state:?}"' EXIT
  /usr/bin/mkdir -p "$apt_state/lists/partial" "$apt_state/parts"
```

tend-agent (approved):

PR #1290, opened 2026-09-17
checks: verify release immutability from the releases when admin is missing

max-sixty:
`GET /repos/{repo}/immutable-releases` requires repository admin, and the bot has write. So the `immutable-releases` check has never produced a verdict in a scheduled run — it reports `SKIP` whether the setting is on or off, and the nightly files a drift issue only for a `FAIL`. Here is last night's run in a consumer, from its session log:

```
Security checks:
  PASS  branch-protection:main — Branch 'main' is protected (restrict-updates ruleset)
  PASS  bot-permission — Bot 'worktrunk-bot' has 'write' permission
  PASS  tag-protection — All tag creation and updates require admin access.
  SKIP  immutable-releases — Could not read the immutable-releases setting. Repository admin access is required to verify it.
  ...
```

The bot's own summary from that run: "9 PASS, 1 SKIP (`immutable-releases`, needs repo admin to read). No FAIL, so no drift issue filed; the SKIP means it isn't an all-pass run either, so nothing was closed." That setting was in fact disabled. It had been since the repo installed tend, which predated the check, and nothing re-ran `tend check` with admin afterwards. The PUT is admin-only too, so the bot could not have repaired it either. A side effect of the permanent skip: the nightly's "every line PASS, close the drift issue" branch was unreachable in every consumer.

Each published release carries GitHub's own `immutable` flag, which a write-scoped token can read. When the setting endpoint 404s, `check_immutable_releases` now reads the most recently published non-draft release instead and fails when it can still be rewritten. An admin's run is unchanged, and the module docstring records the rule that an admin-only read needs a bot-readable stand-in.

Verified against live GitHub with the bot's own token: `max-sixty/tend` (setting on, newest release immutable) → PASS, `max-sixty/worktrunk` (was off) → FAIL, `max-sixty/leaf` (no releases) → SKIP.

The flag is retrospective, so a repository that enables the setting with an older release already published fails this check until its next release, and turning the setting off goes unseen until the repository publishes again. The failure message, the security model and the `install-tend` guidance all say so. The alternative — failing only when an older release is immutable and a newer one is not — detects the setting being turned off but stays silent on the case that actually occurred, where it was never on.

## Finding the newest release is harder than it looks

Two separate bugs, both caught in review, both now covered by tests.

**The listing is not in publication order.** A release's `created_at` is its tag's commit date, not its publication time, and the endpoint documents no order. On a repository publishing from several trains a backport published later sorts below the newer version — `grafana/grafana` serves `v13.2.2` (published 12:21:32) ahead of `v13.0.9` (12:43:46), and `elastic/elasticsearch` does the same.

**And it is paginated.** `gh api` serves 30 a page; this repository already has 36. Taking the maximum over page one repeats the same mistake, since a later publication can sit past the page boundary just as easily as past the first element. The call passes `--paginate`, and `--slurp` collects the pages — gh refuses it alongside `--jq`, so the draft filter and the maximum moved into Python.

## Immutability does not lock a release's body

`PATCH /repos/{repo}/releases/{id}` with a new body succeeds on a release reading `immutable: true` — verified against this repository's own `0.2.10` release with the bot's write-scoped token, the exact actor the threat model names, and the body restored byte-for-byte afterwards. What the setting locks is the assets, the tag, and the draft/published state. `docs/security-model.md` (both the boundary headline and its paragraph), the shipped `install-tend` reference and SKILL.md, `README.md`, and `checks.py`'s docstrings and messages all claimed notes were covered; they now say assets and tags.

## Repository settings

The setting is now enabled on `max-sixty/worktrunk`, `max-sixty/leaf`, `max-sixty/cargo-affected` and `numbagg/numbagg`. `PRQL/prql` is waiting on PRQL/prql#6332, which restructures its release workflow to upload assets to a draft — today it uploads to the already-published release across four jobs, which the setting refuses with `HTTP 422: Cannot upload assets to an immutable release`.


generator/src/tend/checks.py
```diff
@@ -217,6 +219,67 @@ def check_immutable_releases(repo: str) -> CheckResult:
...
+    write access, recording the setting as it stood when that release was
+    published. The newest release reflects the most recent state of it. The
+    flag is never retroactive, so enabling the setting clears a failure here at
+    the next release rather than immediately.
+    """
+    result = _gh(
+        "api",
+        f"repos/{repo}/releases",
+        "--jq",
+        "[.[] | select(.draft | not)][0] // empty",
```
```suggestion
        "[.[] | select(.draft | not)] | max_by(.published_at) // empty",
```

docs/security-model.md
```diff
@@ -66,8 +66,9 @@ Three load-bearing boundaries:
    the point the repository setting is enabled.
 
 `tend check` fails until the first two hold and the third is enabled, so a
-passing check *is* the claim for future releases. GitHub does not apply the
-setting retroactively.
+passing check *is* the claim for future releases — read from the setting when
+the caller is a repository admin, and from the newest release's own
+`immutable` flag otherwise. GitHub does not apply the setting retroactively.
```
tend-agent:
```suggestion
passing check *is* the claim for future releases when the setting itself is
read, which takes a repository admin. A non-admin caller reads the newest
release's own `immutable` flag instead, recording the setting as it stood at
that publish — so a setting turned off since then surfaces at the next
release, not before. GitHub does not apply the setting retroactively.
```

tend-agent (commented):
Two things on the non-admin stand-in.

**`[0]` isn't the newest published release.** `GET /repos/{repo}/releases` documents no sort order, and it isn't any of the obvious ones: `nodejs/node` currently returns `v26.5.1` (published `2026-07-29T14:02:10Z`) ahead of `v24.18.1` (published `2026-07-29T16:17:25Z`), and that pair is out of order by `created_at` and by release `id` too. So on a repo publishing from more than one release train, index 0 can be a release published before the newest one — and with the setting turned off between the two, the check reads the older immutable release and returns PASS while a rewritable release exists. `max_by(.published_at)` makes it deterministic and keeps the behaviour the `// empty` fallback depends on; I confirmed both under `gh`'s own jq, including the no-releases repo still printing nothing.

**The security-model sentence claims more than the stand-in proves.** The flag is retrospective — as `_check_newest_release_immutable`'s docstring puts it, it records the setting "as it stood when that release was published". Turn the setting off after the last release and every nightly keeps reporting PASS on `immutable-releases`, filing no drift issue, until something is next published. The PASS message in the code is careful about this ("the setting was enabled when it was published"); the doc reads as if the bot path were equivalent to the admin path.

generator/src/tend/checks.py
max-sixty: Confirmed and applied — the jq now takes `max_by(.published_at)`.

I looked for live counterexamples rather than taking the ordering on trust, and they exist. A release's `created_at` is its tag's commit date, which diverges from `published_at` by up to 671 days in `rust-lang/rust`, so the two orderings genuinely come apart:

```
grafana/grafana:       [0]=v13.2.2 @12:21:32  newest published = v13.0.9 @12:43:46
elastic/elasticsearch: [0]=v9.5.4  @18:19:18  newest published = v9.4.7  @18:19:34
```

Both are the shape you described — a backport published after the newer train. Running the fallback against `grafana/grafana` before the fix returned `v13.2.2`; after it, `v13.0.9`.

`test_check_immutable_releases_picks_the_latest_published_not_the_first` pins the expression with the reasoning attached.

docs/security-model.md
max-sixty: Agreed — the line overclaimed. Rewritten to separate the two callers: an admin's passing check is the claim for future releases, while a run below admin reads a retrospective flag that establishes the setting was on when the newest release was published, not that it is on now. It also now states the consequence you identified, that turning the setting off is invisible to the nightly run until the repository publishes again, and that closing that window takes an admin-run `tend check`.

The same retrospective limit is now in `_check_newest_release_immutable`'s docstring in both directions, since enabling the setting has a matching lag.

Separately, while verifying the surrounding claims I found that this section was wrong about what immutability covers: `PATCH /repos/{repo}/releases/{id}` with a new body succeeds on a release reading `immutable: true`, including with the bot's own write-scoped token. Assets, the tag, and the draft/published state are locked; the body is not. The check messages and the shipped `install-tend` reference said notes were covered too, and all of them now say assets and tag.

generator/src/tend/checks.py
```diff
@@ -213,7 +215,75 @@ def check_immutable_releases(repo: str) -> CheckResult:
...
+    from several release trains puts a backport published later further down.
+
+    The flag is retrospective in both directions. Enabling the setting clears a
+    failure here at the next release rather than immediately, and turning it
+    off goes unseen until the repository publishes again. An admin's run reads
+    the setting itself and has neither lag.
+    """
+    result = _gh(
+        "api",
+        f"repos/{repo}/releases",
```
tend-agent: This returns only the first page (30 releases); the repo is at 36 today. The docstring declines to rely on the endpoint's order, but the page boundary relies on it — the maximum is over the 30 GitHub happened to serve. Paginating and taking the max in Python is the fix: `--slurp` is rejected alongside `--jq`, and under `--paginate` the filter runs per page, so the `json.loads` below would see two objects and report the response unreadable.

plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -457,7 +457,9 @@ add a no-bypass `deletion` ruleset (see the publisher uplift below).
 
 **Immutable releases.** Enable this before the next release. It locks that
 release, its assets, and its associated tag; GitHub does not apply the setting
-retroactively:
+retroactively. Reading the setting takes repository admin, so the nightly
```
tend-agent, line 461: This ships to consumers and still claims the release itself is locked — the same sentence `references/security-model.md` corrects in this PR.

```suggestion
**Immutable releases.** Enable this before the next release. It locks that
release's assets and its associated tag, though not its notes; GitHub does not
apply the setting retroactively. Reading the setting takes repository admin, so
the nightly
```

README.md
```diff
@@ -147,7 +147,9 @@ start unless the answer is no. `tend check` verifies the setup;
 
 **Immutable releases** lock each release published after the setting is
 enabled, including its assets and tag. `tend check` requires the setting and
```
tend-agent:
```suggestion
**Immutable releases** lock the assets and tag of each release published
after the setting is enabled. `tend check` requires the setting and
```

tend-agent (commented):
Three points, one a decision rather than a defect.

**The releases read is bounded to the first page.** `gh api repos/{repo}/releases`
without `--paginate` returns 30; this repo already has 36 releases. The docstring's
case for `max_by(.published_at)` is that the list endpoint documents no order — and
that same argument makes the first 30 an arbitrary subset, so the maximum is taken
over whatever GitHub served rather than over the repository's releases. A release
whose tag commit sits behind 30 newer ones lands off-page, the check reads an older
release's flag, and it returns PASS while a rewritable release exists: the failure
mode the commit set out to fix. Note for whoever picks this up — `--slurp` is
rejected alongside `--jq`, and under `--paginate` the filter runs per page and emits
one object per page, which the `json.loads` below would reject, so the fix is to
paginate and take the maximum in Python.

**The non-admin FAIL can't be cleared by the person it reaches.** The nightly files
`tend check: configuration drift` on any FAIL, so a consumer that enabled immutable
releases but hasn't published since carries a standing issue whose only remedy is
publishing a release — an admin's `tend check --fix` reads the setting, passes, and
never enters the fix path. `install-tend`'s SKILL.md now states the behaviour ("that
check fails until the next one"), so it's deliberate; the question is whether a drift
issue is the surface you want for it, since it lands on correctly-configured repos.

**Two shipped sentences keep the claim the PR corrects elsewhere.** `docs/security-model.md`
and `references/security-model.md` both move from "the release/its record is locked"
to assets-and-tag, but `install-tend`'s SKILL.md and the README still say the release
itself is locked. Suggestions inline.

generator/src/tend/checks.py
max-sixty: Confirmed and fixed. `gh api repos/max-sixty/tend/releases` returns 30; the repo has 36, so the maximum was taken over the first page. You are right that it is the same mistake as reading `[0]` — with no documented order, a later publication can sit past the page boundary just as easily as past the first element.

The call now passes `--paginate` (which raises the page size to 100 and follows `Link`), and `--slurp` collects the pages into one array. As you note, gh refuses `--slurp` alongside `--jq`, so the draft filter and the maximum moved into Python:

```python
published = [
    release
    for page in json.loads(result.stdout)
    for release in page
    if not release["draft"]
]
...
newest = max(published, key=lambda release: release["published_at"])
```

`test_check_immutable_releases_picks_the_latest_published_across_pages` covers both halves at once: a two-page listing whose later publication is on page two and is the mutable one. There is also a draft-filtering test, which the jq had been carrying untested.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: Deliberate, and worth stating the trade-off explicitly since you raise it.

The alternative I weighed was failing only when an older release is immutable and a newer one is not — unambiguous drift, since the setting must have been turned off between the two. That produces no failure on a correctly-configured repo that has not published since enabling. But it is silent on the case that actually occurred here: the setting had never been enabled in four consumers, for months, and the nightly said nothing. A rule that detects the setting being turned off while missing it never being turned on is the wrong half to keep.

So the choice is one bounded, accurate failure — the newest release genuinely can be rewritten, whatever the setting now says — over silence in the case that has already cost us. It self-clears at the repository's next release.

A drift issue is the right surface in the sense that it reaches a maintainer with admin, who is the only actor that can resolve it: the fix and the authoritative read both require admin, so no bot-side surface could close it. What a drift issue is bad at is sitting open for months on a dormant repo, which is the residual cost here.

README.md
max-sixty: Fixed — both sites now match. Thanks for catching the missed scope; I had corrected three of the five places and stopped.

`README.md` now reads "lock the assets and tag of each release published after the setting is enabled. The body is not locked — a write-access actor can still edit an immutable release's notes", and install-tend's `SKILL.md` "locks that release's assets and its associated tag, but not its body — a write-access actor can still edit the notes".

For the record on why the claim changed at all: `PATCH /repos/{repo}/releases/{id}` with a new body succeeds on a release reading `immutable: true`, including with the bot's own write-scoped token. I verified it against this repository's `0.2.10` release and restored the body byte-for-byte afterwards.

docs/security-model.md
```diff
@@ -62,12 +62,21 @@ Three load-bearing boundaries:
    behind a gate the bot cannot pass, or is explicitly allowlisted in the
    tend config as accepted repo-level exposure.
 3. **Future published releases cannot be rewritten.** GitHub immutable
-   releases lock the release record, its assets, and the associated tag from
-   the point the repository setting is enabled.
+   releases lock a published release's assets and its associated tag from
```
tend-agent:
```suggestion
3. **Future releases' assets and tags cannot be rewritten.** GitHub immutable
   releases lock a published release's assets and its associated tag from
```

tend-agent (commented):
One leftover from the notes correction. `docs/security-model.md`'s boundary 3 headline still asserts that published releases cannot be rewritten — the claim the paragraph directly under it now retracts. The bolded line is what a reader skimming the three boundaries takes away, so it carries the pre-PR model on its own. Suggestion inline.

generator/src/tend/checks.py
tend-agent: Verified the fix against live GitHub rather than reading it. `gh api --paginate --slurp` does return one array per page — `cli/cli` comes back as 3 pages / 204 releases, `grafana/grafana` as 7 / 648 — so `_release_pages` models the real shape, and a repo with no releases returns `[[]]`, the single empty page the SKIP path reads. Running `check_immutable_releases` with the bot's own token: the setting endpoint 404s here, so the fallback is the path that actually executes — `max-sixty/tend` → PASS (0.2.10), `max-sixty/worktrunk` → FAIL (v0.78.0), `max-sixty/leaf` → SKIP.

One calibration on the page boundary, since it bears on what the change claims rather than on whether it is right: I went looking for a repository where page 0's maximum differs from the whole listing's and did not find one. On five multi-train repos the newest publication sits inside the first page every time, just not first within it.

```
grafana/grafana        7 pages, 648 releases   page-0 max = all-pages max = v13.0.9
elastic/elasticsearch  3 pages, 266 releases   page-0 max = all-pages max = v9.4.7
kubernetes/kubernetes  9 pages, 815 releases   page-0 max = all-pages max = v1.37.0
```

So the `[0]` divergence is observable and the boundary one is so far only arguable — `--paginate` is correctness-by-construction against an undocumented order rather than a fix for an answer anyone can get wrong today. Still the right trade at this price: one extra request per 100 releases, no consumer is near 100, and the Python selection reads more plainly than the jq did.

On the notes correction, the prose now reads correctly in all five places, but two one-line summaries still carry the pre-PR claim. `docs/security-model.md`'s boundary-3 headline is one, with a suggestion on the review I just posted. The other is this function's own docstring — [`checks.py:187`](https://github.com/max-sixty/tend/blob/d327b009d33f476f8bcefa4d217fa5b7e267868b/generator/src/tend/checks.py#L187), "Check that future published releases and their tags cannot be rewritten", where assets rather than releases is what the check now establishes.

And on the drift-issue surface: your call, and the reasoning holds — a FAIL only an admin can clear does have to reach an admin, and the residual is a dormant repo carrying it.

docs/security-model.md
max-sixty: Fixed, and the same unqualified claim turned out to be in two more places, so all three now match the paragraph.

- `docs/security-model.md:64` — the headline is now "**Future releases' assets and tags cannot be rewritten.**" The sentence below it that read "a passing check by a repository admin *is* the claim for future releases" was making the same promise, so it now says "the claim for the assets and tag of every release published afterwards".
- `generator/src/tend/checks.py` — the module docstring listed the boundary as "future releases are immutable", and `check_immutable_releases`'s own docstring said "future published releases and their tags cannot be rewritten". Both now name assets and tags.

I left the failure message as "The newest release (…) can be rewritten — a write-access bot can replace its assets", since the negative direction is accurate as stated: the release genuinely can be rewritten, and the clause after the dash says which part is at stake.

tend-agent (approved):

PR #1291, opened 2026-09-17
sandbox: refresh the pins per architecture, and correct the sweep's hold rules

max-sixty:
Follow-up to #1289, from its review's last open note.

`refresh_sandbox_pins.py` derives the Debian versions the weekly sweep writes into `install-sandbox-runtime.sh`, and it read `binary-amd64` only. One pin serves every runner, so a version published for amd64 before arm64 would have installed fine on GitHub's x64 boxes and left the pin unsatisfiable on an `ubuntu-24.04-arm` one — with nothing in the refresh output showing the split, since the arch it read was happy.

The refresh now resolves each architecture separately and takes only what they agree on. A divergence stops it rather than picking a side: the versions differ only while one arch is behind, so the answer is a later instant, not two pins or a pin that installs on one arch.

## The split is live, and it is not the one you would guess

Between 21:00Z and 22:00Z today `noble-security` published `bubblewrap 0.9.0-1ubuntu0.3` for amd64, with no arm64 counterpart in any pocket. So this branch's first real exercise is a halt, and the next weekly sweep hits it rather than producing the usual one-line diff.

What that upload does matters more than the split:

```
bubblewrap (0.9.0-1ubuntu0.3) noble-security; urgency=medium
  * SECURITY REGRESSION: Incompatibility with Flatpak (LP: #2167621)
    - debian: Drop CVE-2026-87766
```

It **reverts** the CVE-2026-87766 fix that `0.9.0-1ubuntu0.2` added, because that fix broke Flatpak. Tend is pinned to `0.9.0-1ubuntu0.2`, so staying put is what keeps the symlink-traversal fix rather than what holds one back — the archive has now moved in both directions within a day, and the pin absorbed both. #1277's `0o755` container works under either resolution, so nothing there needs revisiting.

That disproved the sweep rule #1289 shipped, which told the operator to treat a pin that will not move as a held security update. It now says to read the changelog rather than the version order, and covers the revert case and the arch split next to the ordinary bump.

Also corrected: the halt message named the one remedy unavailable to the run that hits it (`--instant` already defaults to the current hour, so a later timestamp resolves to the same split). It now says nothing moved, the instant included, and points at that rule.

And the section now states, once and up front, that the four constants move together or not at all. Holding a version while the weekly instant moves on is not satisfiable — the install asks apt for an exact version at that instant, each pocket lists one version, and a superseded one is dropped within hours. Today's archive shows the timing: `0.9.0-1ubuntu0.1` is listed in two pockets at `00:00Z` and in none by `18:00Z`. The hold and arch-split cases now just say "the whole set" rather than each re-deriving it.

## Verification

Both branches exercised. At the currently pinned instant, and every hour through 21:00Z, amd64 and arm64 agree and the script exits 0 with all four constants unchanged — which also confirms #1289's live pin is valid on arm64. From 22:00Z it halts, naming the package that split and both arches' versions.

`test-sandbox` does not exercise this script, so CI can only tell you the repo lints and the rest of the suite still passes.


.claude/skills/running-tend/scripts/refresh_sandbox_pins.py
```diff
@@ -75,24 +82,37 @@ def newer(candidate: str, incumbent: str | None) -> bool:
...
 
+def resolve(instant: str) -> dict[str, str]:
+    """The version every architecture agrees on, or no version at all."""
+    by_arch = {arch: resolve_arch(instant, arch) for arch in ARCHES}
+    first, *rest = by_arch.values()
+    split = sorted(name for name in first if any(o[name] != first[name] for o in rest))
+    if split:
+        sys.exit(
+            f"{instant} publishes {', '.join(split)} differently per architecture "
+            f"({by_arch}); try a later instant, once the slower arch has caught up"
```
tend-agent: A split parks every constant, `PACKAGES_RESOLVED_AT` included, so the message is the operator's whole view of the halt — and the remedy it names is the one this run cannot take: `--instant` already defaults to the current hour, and a future timestamp resolves to the same split.

```suggestion
            f"{instant} publishes {', '.join(split)} differently per architecture "
            f"({by_arch}); nothing moved, the instant included. Re-run once the "
            "slower arch catches up; if the split is holding back a security "
            "update, report that rather than dropping the row."
```

`ruff format` and `ruff check` are clean on the result.

tend-agent (commented):
The arch gap is real and the fix lands in the right place: I ran the PR's script against the live snapshot service, and `snapshot.ubuntu.com/ubuntu` does serve `binary-arm64` for every pocket and component the script reads, with genuinely per-arch contents — so the agreement check has teeth rather than comparing an index against itself.

It also fires right now. `noble-security` published bubblewrap `0.9.0-1ubuntu0.3` for amd64 between 21:00Z and 22:00Z today; arm64 carries no `0.9.0-1ubuntu0.3` in any pocket. Run with its default instant, the script exits 1 with nothing moved — so the new branch's first real exercise is a held security update, and the next weekly sweep hits it rather than producing the usual one-line diff.

The halt itself is the right call, and holding the whole set is defensible: once a security version is copied into `noble-updates`, the version it superseded is no longer listed at a later instant, so moving `PACKAGES_RESOLVED_AT` while pinning a held `BUBBLEWRAP_VERSION` is not generally satisfiable. Worth naming in the PR body, though, since it means an arch split on one package also parks the instant move whose own purpose is re-resolving SRT's npm tree.

What does need a change is the message, per the inline suggestion: `try a later instant` is the one remedy unavailable to the run that hits it — `--instant` already defaults to the current hour, and a future timestamp resolves to the same split. The operator needs to know that nothing moved and that a security update waiting on arm64 is a finding to report, which is `running-tend`'s existing rule for a pin that will not move.

<details><summary>What the snapshot indexes hold</summary>

`bubblewrap` in `main`, at `2026-09-17T22:00:00Z`:

```
amd64  noble           0.9.0-1build1
amd64  noble-updates   0.9.0-1ubuntu0.2
amd64  noble-security  0.9.0-1ubuntu0.3
arm64  noble           0.9.0-1build1
arm64  noble-updates   0.9.0-1ubuntu0.2
```

Resolving the PR's script against a series of instants: `18:00Z` (the currently pinned one) through `21:00Z` agree on `0.9.0-1ubuntu0.2` and exit 0 with all four constants unchanged, which confirms #1289's live pin is valid on arm64. `22:00Z` onward halt on the split above.

</details>

.claude/skills/running-tend/SKILL.md
```diff
@@ -326,15 +326,24 @@ re-resolves SRT's npm tree; that week's diff is one line and its green
...
-holding back. Reverting the version alone turns a red check into a silent hold
-on a security update, so it is never the answer on its own.
+says it in a few lines. Newer is not automatically safer: a `-security` upload
+can *drop* a fix the previous one added, when that fix turned out to break
+something else, so read the changelog rather than the version order. Where the
+newer version is a fix, take it — fix Tend against it if `test-sandbox` goes
+red, and land both in one PR. Where it is a revert, or where taking it needs a
+maintainer decision, leave the pin where it is and open an issue naming the
+version and what it changes. What is never the answer on its own is moving the
+version with no note, which turns a red check into a silent hold.
```
tend-agent: A held version and a moved instant are not independently satisfiable: `apt-get install "bubblewrap=$BUBBLEWRAP_VERSION"` runs against the `PACKAGES_RESOLVED_AT` snapshot, and the pockets hold one version each, so the superseded version disappears once its replacement reaches `noble-updates`. The arch-split paragraph below says this; the hold branch should too.

```suggestion
red, and land both in one PR. Where it is a revert, or where taking it needs a
maintainer decision, leave the pin where it is and open an issue naming the
version and what it changes — and hold `PACKAGES_RESOLVED_AT` with it. The
install asks apt for an exact version at that instant, and the version a
`-security` upload supersedes stops being listed once the successor reaches
`noble-updates`, so moving the instant while holding the version makes the
pin unsatisfiable. What is never the answer on its own is moving the version
with no note, which turns a red check into a silent hold.
```

tend-agent (commented):
The revert branch the second commit adds is the case that is live today, and it is missing the parking note the arch-split paragraph right below it carries. `install-sandbox-runtime.sh` asks apt for `bubblewrap=$BUBBLEWRAP_VERSION` against the snapshot at `PACKAGES_RESOLVED_AT`, and the version a `-security` upload supersedes stops being listed once the successor reaches `noble-updates` — so "leave the pin where it is" parks the instant with it. Without that said, the obvious move in the week bubblewrap's proposal is a revert is to commit the instant alone, which is the one-line `chore: re-pin the sandbox boundary (<date>)` diff the section documents two paragraphs up, and it makes the version pin unsatisfiable. The inline suggestion adds the sentence.

tend-agent (approved):

PR #1292, opened 2026-09-17
mention: drop a review bot's approval, and read tend's output where nobody else does

max-sixty:
Two changes, both from reading what tend did in a consumer repo we don't maintain.

## `tend-mention` wakes on a third-party review bot's approval

`mention_verify`'s review gate skips an approval only when its body is empty. A code-health badge and a "looks safe to merge" template are bodies, so a review bot's approval falls through to the participation heuristic and boots a session that can only exit silently — once per review bot per push, on a PR the bot has already reviewed.

Author type is what separates the two cases: a person's prose alongside an approval is addressed to someone, and a review bot's is generated status. So an approval with no inline comments is terminal when the author is a `Bot`, whatever its body says.

A `COMMENTED` review from the same bot still runs. Sometimes it carries a real finding and sometimes a fixed template, and no state or author field tells them apart — that needs a reader, which is what the session is.

## `review-reviewers` comes off pause, over the consumers we don't maintain

The outside-in sweep has been dispatch-only since its hourly fleet-wide version twice hit a quota ceiling that took the user-facing workflows down with it. It now runs daily over `diffplug/dormouse` and `jakobtfaber/Faber2026`, and nothing else: those are the repos whose tend runs nobody here reads, while on our own repos tend's work arrives in PRs and issues we read anyway. Two legs a day is a fraction of what got it paused.

The skill had a matching hole. It read an empty run list as all-clear, which is exactly backwards when a repo has switched tend off — the state that produces no runs at all. Step 1 now reads the target's workflow states first, and reporting that state is the whole job: there are no runs for Steps 2 and 3 to key on, while Step 4's dedup check is what keeps a repo that stays switched off from opening the same issue every day.

`test_review_reviewers_matrix_covers_consumers` asserted the matrix covered `data/consumers.json`, which is now deliberately false. The invariant that replaces it is weaker (matrix entries must still be consumers) plus one line in `running-tend`'s weekly refresh step, since a new adopter shows up there first.

## One window for both daily profiles

Putting `review-reviewers` on a cron exposed a constant sized for its old cadence. `list_recent_runs.py` capped its window at 6h, which was right when dispatches sat days apart; on a daily cron the anchor is always ~24h back, so every tick would clamp to 6h and record an ~18h coverage gap. The night is covered by that 6h either way — it is the event-driven part of the day, which is what this sweep exists to judge, that would fall outside.

Both profiles are now daily crons resuming from their own previous run, so they share one 49h cap, one 25h no-anchor default and one 24h creation cushion — the values `review-runs` already used. `profile` no longer selects window constants at all. `review-runs`' behaviour is unchanged.


.claude/skills/review-reviewers/SKILL.md
```diff
@@ -87,6 +87,23 @@ gh api "repos/$ARGUMENTS/contents/.claude/skills/running-tend/SKILL.md" \
 
 If the file doesn't exist, try the legacy overlay paths and the repo's root project instructions (`.claude/skills/running-tend.md`, `.claude/CLAUDE.md`, `CLAUDE.md`, `AGENTS.md`). Understanding the repo's instructions is essential context for evaluating outcomes — without it, you'll misjudge authorized behavior as a violation.
 
+Then check whether tend is still enabled there — it decides how to read an
+empty run list below:
+
+```bash
+gh api "repos/$ARGUMENTS/actions/workflows" \
```
tend-agent: `/actions/workflows` pages at 30 by default, so on a repo with more workflows than that some `tend-*` entries fall off page one silently — and this gate exists to stop a partial view being read as an all-clear. Both current matrix targets are under the limit (dormouse 17, Faber2026 15), but the paragraph below invites adding repos.

```suggestion
gh api --paginate "repos/$ARGUMENTS/actions/workflows" \
```

tend-agent (commented):
Moving `review-reviewers` to a daily cron leaves `list_recent_runs.py`'s window cap where the hourly cadence put it, so each scheduled tick surveys 6h and records the other 18h as a coverage gap.

In [`plugins/tend-ci-runner/scripts/list_recent_runs.py`](https://github.com/max-sixty/tend/blob/main/plugins/tend-ci-runner/scripts/list_recent_runs.py) the `review-reviewers` profile takes `WINDOW_CAP = timedelta(hours=6)`. When the anchor — the previous successful run of this workflow — is older than that, the floor clamps to `now - 6h` and the script warns `more than 6h back ... Record a coverage gap, not an all-clear.` At a daily cadence the anchor is always ~24h back, so that branch fires on every cron run, not on the outage it was written for. Step 1 then has the session record a gap, and per its own text the next run's floor advances past that span regardless — so 09:41Z–03:41Z is read by nothing, every day.

The cron time does put the consumers' default `tend-nightly` (06:17Z) and `tend-review-runs` (07:47Z) inside the 6h window, so the night is genuinely covered. What falls outside is the event-driven day — `tend-review`, `tend-triage`, `tend-mention`, `tend-ci-fix` — which is where the accepted-or-rejected outcomes this sweep judges actually land.

The same file already carries the daily-cadence analogue for the other profile: `REVIEW_RUNS_WINDOW_CAP = timedelta(hours=49)`, two days plus slack so one missed tick still gets covered. Giving `review-reviewers` that shape would also make both prose claims true as written, where today they read backwards:

- `.claude/skills/review-reviewers/SKILL.md`: "or 6h back when that is older — normally a day, since the workflow runs daily". The floor is 6h back *normally*, and the cap means it never reaches further.
- `.claude/skills/running-tend/SKILL.md`: "a daily tick covers the day and a dispatch soon after one covers the last 6h and warns on stderr". Inverted — the daily tick is the case that clamps and warns; a dispatch within 6h of the last run is the case that doesn't.

The cap sits outside this diff, so I haven't pushed anything. Happy to send the cap plus the two clamp tests it pins (`test_no_anchor_floors_at_6h_and_warns` and `test_stale_anchor_clamps_to_6h_and_warns` in `generator/tests/test_list_recent_runs.py`) if you'd like it folded in here.

The mention gate reads clean: the new condition is a strict superset of the old one, so a person's approval with prose is unaffected, and the `@bot` mention check and the inline-comments check both still run ahead of it.

.claude/skills/review-reviewers/SKILL.md
max-sixty: Added `--paginate` in 12d625f8.

.claude/skills/review-reviewers/SKILL.md
```diff
@@ -87,6 +87,23 @@ gh api "repos/$ARGUMENTS/contents/.claude/skills/running-tend/SKILL.md" \
...
+
+```bash
+gh api --paginate "repos/$ARGUMENTS/actions/workflows" \
+  --jq '.workflows[] | select(.path | test("/tend-")) | "\(.state)\t\(.path)"'
+```
+
+`disabled_manually` across every `tend-*` workflow means tend is switched off
+there, so there are no runs to survey and Steps 2 to 4 do not apply. Read
+tend's output from the weeks before the last run instead, against the same
+gates as any other window, and say in the summary that it is switched off.
```
tend-agent:
```suggestion
`disabled_manually` across every `tend-*` workflow means tend is switched off
there, so there are no runs to survey and the run-keyed parts of Steps 2 and 3
do not apply. Read tend's output from the weeks before the last run instead —
Step 2's repo-wide sweep takes a window, not run IDs — against the same gates
as any other window, and say in the summary that it is switched off. Step 4
still applies: every tick re-reads that same fixed window, so its incidents are
occurrences prior runs already recorded, not new ones.
```

tend-agent (commented):
`Steps 2 to 4 do not apply` drops Step 4 (Deduplicate) along with the run-keyed work, in the one mode where the window never advances. Once a consumer switches tend off, "the weeks before the last run" is a fixed span, so every daily tick re-reads the same incidents — and with Step 4 declared inapplicable, the leg opens PRs or issues against them without the check for an existing one on tend. The gates compound it: `review-gates.md` counts occurrences from "the current analysis **and** historical evidence recorded by prior runs", so a single incident re-read daily climbs one occurrence per tick and clears the Medium 5+ bar on its fifth reading rather than on five occurrences.

Step 2's repo-wide sweep is also the only recipe in the skill that enumerates bot output by window rather than by run ID, which is exactly what a run-less window needs; it's the per-run mapping under it, and Step 3's session-log work, that don't apply here. Suggestion inline.

.claude/skills/review-reviewers/SKILL.md
max-sixty: Dropped the retrospective instead of rewording it — the frozen span was what made Step 4 load-bearing here. Reporting the state is the whole job now; Step 4 applies as written. 9f1f87ba

tend-agent (approved):
The `review-reviewers` section of the description still says that where tend is off, "the run reviews tend's own output from the weeks before it stopped, against the same gates as any other window" — which is what 9f1f87ba dropped. Worth recomposing that paragraph around reporting the state, since the description is what a later reader takes the change to have done.

PR #1293, opened 2026-09-18
list-recent-runs: a repo past the workflow fetch limit says so

max-sixty:
`gh workflow list` fetches 50 workflows without a `--limit` and says nothing when it truncates. `list_recent_runs.py` called it with no limit, so on a repo past that edge a `tend-*` workflow could fall outside the listing entirely — the session would then never query its runs, and would read the result as a quiet window rather than an incomplete one.

It now asks for 200 and warns on hitting it, matching the `RUN_LIMIT` warning directly below it in the same function.

The alternative was rewriting the call as `gh api --paginate repos/<repo>/actions/workflows`, which removes the cliff rather than moving it. That changes how the script resolves its target repo — `gh workflow list` takes `-R` and an empty argument list means "the current repo", while `gh api` needs the path spelled out in every case — so it is a wider change to a script every consumer's `tend-review-runs` runs daily. The warning is what makes truncation visible, which is the actual defect.

Found while fixing the same class of bug one file over in #1292, where the bot reviewer caught an unpaginated `/actions/workflows` call in the `review-reviewers` skill.


generator/tests/test_list_recent_runs.py
```diff
@@ -264,6 +264,23 @@ def test_api_failure_fails_loudly(env: dict[str, str], failure: str) -> None:
...
+    Path(env["WF_JSON"]).write_text(
+        json.dumps([{"name": f"tend-{i}"} for i in range(200)])
+    )
+    _anchor(env, (555, NOW - 5400))
+    _runs(env, _run_entry(1, updated=NOW - 600))
+
+    result = _run(env)
+
+    assert result.returncode == 0, result.stderr
+    assert "--limit 200" in Path(env["GH_CALLS"]).read_text()
```
tend-agent:
```suggestion
    workflow_calls = [
        line
        for line in Path(env["GH_CALLS"]).read_text().splitlines()
        if line.startswith("workflow list")
    ]
    assert workflow_calls and all("--limit 200" in c for c in workflow_calls), (
        "the workflow listing must ask for the limit, not inherit gh's default 50"
    )
```

tend-agent (commented):
The test doesn't pin the change: `test_workflow_fetch_limit_warns` asserts `"--limit 200" in GH_CALLS`, but the same test triggers 200 `gh run list --limit 200` calls (`RUN_LIMIT` is also 200), so that substring is in the log regardless of what the workflow listing asked for.

Removing `"--limit", str(WORKFLOW_LIMIT)` from the `gh workflow list` call leaves the whole module green — I ran it. The fake `gh` ignores `--limit` and returns all 200 rows from `WF_JSON`, so the warning still fires; in production `gh` would hand back 50, `len(workflow_rows) >= WORKFLOW_LIMIT` would never hold, and truncation would stay silent at exactly the edge this PR closes. So the flag — the half of the change that actually widens the fetch — is unguarded, while the warning block is covered.

Scoping the assertion to the `workflow list` line, the way `test_window_floors_at_the_anchors_start` filters `GH_CALLS` for the anchor query, fails on that revert and passes as written. Suggestion inline; it is `ruff format`/`ruff check` clean.

Nothing else came out of the pass: the `>=` threshold and message match the `RUN_LIMIT` warning below it, `gh workflow list --limit` paginates past the 100-per-page ceiling, and no call site of the script changes shape.

generator/tests/test_list_recent_runs.py
max-sixty: Confirmed by deleting the flag and running the module — 14 passed. Scoped the assertion to `workflow list` lines in a00d9d99; it now fails on that revert.

tend-agent (approved):

PR #1294, opened 2026-09-18
config: forward any model name to the harness CLI

max-sixty:
A consumer wanted `model: claude-opus-5` so their bot's behavior stays fixed when Anthropic promotes the next model behind the `opus` alias, and found the route closed: `KNOWN_MODELS_BY_HARNESS` listed `opus`, `sonnet`, `haiku` and nothing else, so `init` refused the exact id. Codex, by the same dict, got no model validation at all — any string passed. Both CLIs accept an alias or an exact id, and either catalog can gain a model the generator has not heard of, so the allowlist bought a typo check at the price of a pin, for one harness.

Neither harness enumerates models now. `model` must be a non-empty string; beyond that `claude` and `codex exec` judge the name, which is the only place that knows the current catalog. A typo now fails the job rather than `init` — `claude -p --model opsu` exits 1 with `[claude-code:unrecognized_model] {"model":"opsu"}`, naming the bad value.

`_parse_effort` loses its `model` argument, and with it the `effort` + `model: haiku` refusal. That guard matched one spelling of one model — `claude-haiku-4-5-20251001` with `effort: low` already walked past it — so once exact ids are legal the same configuration would be accepted or refused depending on how it was written. The Claude CLI takes `--effort low --model haiku` at exit 0, ignoring the level rather than failing, so nothing outward depended on it. Effort is still checked against the harness, including for a workflow that switches harness and inherits the top-level value; the new `test_inherited_effort_rejected_when_a_workflow_switches_harness` covers that path, which previously rode along with the model checks.

`_effective_model` becomes `effective_model`: dropping the per-workflow cross-family check leaves `workflows.py` as its only caller, and it was the package's one cross-module underscore import.

`install-tend`'s config template said the same thing about `max` and is what writes a new consumer's `.config/tend.yaml`, so it moves too.

Nothing downstream reads the model string — `shared/steps/token_usage.py` copies it into the usage record, and `checks.py` and the site never see it — so the only consumer is the `--model` flag. `docs/tend.example.yaml` (with its `install-tend` mirror) and the README's effort line say what the values are now.

This is the half of a consumer's request that #1161 left open; the `effort` input shipped there.


generator/tests/test_generate.py
```diff
@@ -2062,57 +2044,38 @@ def test_per_workflow_model_override_unblocks_cross_family(tmp_path: Path) -> No
...
+            model: {model}
     """)
     cfg = Config.load(_minimal_config(tmp_path, extra))
     workflows = {wf.filename: wf for wf in generate_all(cfg)}
-    nightly = workflows["tend-nightly.yaml"]
-    assert "model: haiku" in nightly.content
-    review = workflows["tend-review.yaml"]
-    assert "model: opus" in review.content
+    assert f"model: {model}" in workflows["tend-nightly.yaml"].content
+    assert "model: opus" in workflows["tend-review.yaml"].content
```
tend-agent, line 2058: The per-workflow arm of the new check — `workflows.<name>.model must be a non-empty string` — has no test, while its top-level twin got four cases and the cross-family check it replaces had two. Worth pinning: an uncaught list or int reaches the template unchanged and renders `model: ['opus']` into the consumer's workflow, which only fails once a job boots. Bare `model:` stays out of the parametrize — at workflow level `None` means inherit, unlike the top level where it's rejected.

```suggestion
    assert "model: opus" in workflows["tend-review.yaml"].content


@pytest.mark.parametrize("model", ['"   "', "[opus]", "5"])
def test_per_workflow_model_must_be_a_non_empty_string(
    tmp_path: Path, model: str
) -> None:
    """Bare `model:` means inherit; anything else must name a model."""
    extra = dedent(f"""\
        workflows:
          nightly:
            model: {model}
    """)
    with pytest.raises(
        click.ClickException,
        match="workflows.nightly.model must be a non-empty string",
    ):
        Config.load(_minimal_config(tmp_path, extra))
```

tend-agent (commented):
`plugins/install-tend/skills/install-tend/SKILL.md`'s config template still reads `# effort: medium   # low | medium | high | xhigh; Claude Opus/Sonnet also accept max` — the claim this diff rewrote in the README and in both `tend.example.yaml` copies. With `_parse_effort`'s haiku guard gone it is the only text left in the repo asserting model-specific effort support, and it ships: that skill is what writes a new consumer's `.config/tend.yaml`, so an install reads `max` as Opus/Sonnet-only while the generator now accepts the pair and the example config stamped next to it says otherwise. It's outside this diff — say the word and I'll push the one-line fix onto this branch.

generator/tests/test_generate.py
max-sixty: Added in 2f172b7 as `test_per_workflow_model_must_be_a_non_empty_string`, with your three cases and the note about bare `model:` meaning inherit. Confirmed it bites: deleting the per-workflow guard fails all three, restoring it passes them.

The other finding landed in the same commit — `install-tend`'s config template now reads `Claude also accepts max`. A case-sensitive sweep is what missed it; `Opus/Sonnet` there is capitalized.

tend-agent (approved):

PR #1295, opened 2026-09-18
install-tend: hand the OAuth mint to the user and drop the wrapper

max-sixty:
Minting the Claude subscription token asked the user for a live credential in the middle of an install they were meeting for the first time, and nothing they could read said where the value went. The agent drove `claude setup-token` through a bundled pty wrapper and piped the result into `gh secret set` — which kept the token out of the transcript, but took that entirely on trust.

The mint is now the two commands in the user's own terminal, with the ending their choice: run `gh secret set` themselves so the value never leaves their terminal, or paste the token back for the agent to set it. The first leads and carries the recommendation, since a token pasted into chat is a live credential in a transcript that outlives the install. `gh secret list` goes over alongside, so the result is visible without exposing the value.

That removes the reason for the wrapper. `oauth_token.py` (341 lines) and its tests (384) are deleted, and 7a's OAuth mint goes from 92 lines to 36. Everything the wrapper needed — a background task, a 15-minute PKCE window, single-use codes with a watched paste file, a guard against `gh secret set` storing an empty body — existed because an agent was driving a TUI whose approval still had to happen in the user's browser, and the path already gave up and handed over after a window lapsed twice.

Two things the cut moved rather than removed:

- The empty-secret hazard now belongs to the paste-back ending, so its guard moved there: check the value starts with `sk-ant-oat01-` before setting it, since `gh secret set` stores a malformed body and exits 0 while every check downstream reads names rather than values.
- The hand-over rule — whatever the user must act on goes in the reply itself, a clickable link plus the bare URL, repeated while its window is open — moved out of "Browser sessions" into the preamble, because 7a hands over commands without needing a browser and that section's opening enumerates the steps that do.

What users lose is real: on the old path they approved in the browser and were done, where now they run two commands and copy a token between them.


plugins/install-tend/skills/install-tend/SKILL.md
```diff
@@ -693,86 +698,38 @@ flow or a resumed install lands here without one — first ask the user to choos
...
-  commands, fully substituted, for them to run in their own terminal (any
-  machine with Claude Code installed and `gh` logged in as the
-  maintainer; `https://claude.com/claude-code` to install it), whenever
-  suits them:
+Offer both endings when handing the commands over: paste the token back and
+the agent sets the secret, or run the second command so the value never
+leaves their terminal. Say which is which, since a token pasted into chat is
+a live credential in the transcript. Given neither `--body` nor a pipe,
+`gh secret set` prompts for the value, so on that second ending the token
+goes from the first command's output to the prompt and nowhere else.
```
```suggestion
Offer both endings when handing the commands over, leading with the safer
one: run `gh secret set` themselves and the value never leaves their
terminal, or paste the token back and the agent sets the secret. Say which
is which, since a token pasted into chat is a live credential in a
transcript that persists on disk, and the agent has no use for the value.
Given neither `--body` nor a pipe, `gh secret set` prompts for the value, so
on the terminal-only ending the token goes from the first command's output
to the prompt and nowhere else.
```

tend-agent (commented):
Handing the mint to the user is the right trade for a 341-line pty wrapper, and the cut is clean — nothing left in the tree reaches for `oauth_token.py`, the empty-secret guard is re-established as the prefix check, and the `check_secrets` claim still holds (it compares names only).

One thing to settle: the paragraph offering the two endings leads with paste-back, and an installing agent reading it in order will offer that one first. The Manual path this replaces stated the opposite rule — "Don't ask for it in chat: the agent has no use for the value" — and that reasoning survives the cut intact: paste-back buys the user one skipped command, while a credential advertised as 1-year lands in a session transcript that persists on disk. The wrapper whose whole purpose was keeping the token out of the transcript is gone, so this ordering is now the only thing deciding where the value goes.

The inline suggestion leads with the terminal-only ending and keeps both. If paste-back is meant to be genuinely co-equal rather than the fallback, the text is the place to say so — otherwise the next editor reads the removed rule and re-tightens it.

plugins/install-tend/skills/install-tend/SKILL.md
max-sixty: Taken — the paragraph now leads with running `gh secret set` in their own terminal and carries the recommendation, with paste-back offered second. You're right that first-listed reads as the default, and that inverts the rule the paragraph replaced.

Fixed in 3ce9e7d8.

tend-agent (approved):

PR #1297, opened 2026-09-18
fix(skills): stop the worktree recipes leaving the session in a deleted directory

tend-agent:
Two bundled worktree recipes end with `cd -` followed by `git worktree remove`, which leaves the session with no working directory for the rest of the run.

`cd -` restores `$OLDPWD`, not the directory the recipe started in. Both recipes `cd` twice — once into `<worktree>/.claude/skills/<name>/` to `mv` the authored file into place, then into the worktree root to commit and push — so `$OLDPWD` at the `cd -` is the *first* `cd`'s target, still inside the worktree. `git worktree remove --force` then deletes the directory the shell is standing in and exits 0, and since the harness persists the working directory between Bash calls, every command after that block fails on a cwd that no longer exists.

In `review-runs` that block sits in Step 6, immediately before Step 7 writes the run summary to `$GITHUB_STEP_SUMMARY` — so a session that follows the recipe opens its PR and then loses the summary. The `proposing-tend-instructions.md` copy is worse placed: it is the generic skill-proposal path any workflow can reach, and step 5 ("open and exit") is the only thing that has kept the blast radius small.

Both are fixed the same way: the two `cd`s move into subshells, so the session's own cwd never enters the worktree and `cd -` is not needed at all. The multi-command subshell carries `set -e`, since a subshell without it runs on past a failed `git commit` and pushes nothing while reporting success.

`test_bundled_runner_instructions_never_return_with_cd_dash` bans `cd -` across `plugins/tend-ci-runner`, alongside the existing `/tmp` ban it sits next to. It fails on both files before this change. A recipe cannot know what the previous Bash call left in `$OLDPWD`, so the ban is repo-wide rather than a rule either file states.

<details><summary>How the behavior was confirmed</summary>

Run in this session against a scratch worktree, reproducing the recipe's `cd` sequence:

```
++ cd /home/tend-sandbox/tmp/wtprobe/.claude/skills/running-tend
++ cd /home/tend-sandbox/tmp/wtprobe
++ cd -
/home/tend-sandbox/tmp/wtprobe/.claude/skills/running-tend
++ git worktree remove /home/tend-sandbox/tmp/wtprobe --force
++ echo exit=0
exit=0
++ pwd
/home/tend-sandbox/tmp/wtprobe/.claude/skills/running-tend
```

`cd -` lands inside the worktree rather than back at the checkout; the removal succeeds anyway and leaves `pwd` pointing at a deleted path.

</details>

Found by the nightly survey, which drew `plugins/tend-ci-runner/skills/review-runs/SKILL.md` today.


generator/tests/test_repo_pins.py
```diff
@@ -742,6 +742,29 @@ def test_bundled_runner_instructions_have_no_unscoped_tmp_paths() -> None:
...
+
+    It restores `$OLDPWD`, which is whatever the last `cd` left — after a
+    recipe's second `cd` that is the first `cd`'s target, not the checkout.
+    The worktree recipes end by deleting the directory they moved into, so a
+    session that followed one is left with no working directory and every
+    later command fails. A recipe that has to change directory does it in a
+    subshell, which never moves the session's own cwd.
+    """
+    runner = REPO_ROOT / "plugins" / "tend-ci-runner"
+    cd_dash = re.compile(r"(?<![\w-])cd\s+-\s*$")
```
```suggestion
    cd_dash = re.compile(r"(?<![\w-])cd\s+-(?![\w-])")
```
Catches `cd - >/dev/null` and `cd - && …` as well as the bare form, and still lets `cd --` and `cd -P` through.

plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -229,16 +229,20 @@ Editing `.claude/skills/` requires the read-only-mount workaround (bind-mounted
...
-cd -
+(
+  set -e
+  cd "$TMPDIR/review-runs-fix"
+  git add .claude/skills/
+  git commit -m "skills(running-tend): ..."
+  git push -u origin daily/review-runs-$GITHUB_RUN_ID
+  gh pr create --title "..." --body-file "$TMPDIR/pr-body.md" --head daily/review-runs-$GITHUB_RUN_ID
+)
 git worktree remove "$TMPDIR/review-runs-fix" --force
```
```suggestion
) && git worktree remove "$TMPDIR/review-runs-fix" --force
```
So a `set -e` abort leaves the worktree — and the authored file in it — for a retry.

plugins/tend-ci-runner/skills/running-in-ci/references/proposing-tend-instructions.md
```diff
@@ -146,16 +146,20 @@ These steps are for the overlay path. Filing upstream follows
...
+
+   (
+     set -e
+     cd "$TMPDIR/skill-fix"
+     git add .claude/skills/
+     git commit -m "skills(running-tend): ..."
+     git push -u origin skills/<topic>-$GITHUB_RUN_ID
+     gh pr create --title "..." --body-file "$TMPDIR/pr-body.md" --head skills/<topic>-$GITHUB_RUN_ID
+   )
    git worktree remove "$TMPDIR/skill-fix" --force
```
tend-agent:
```suggestion
   ) && git worktree remove "$TMPDIR/skill-fix" --force
```

tend-agent (commented):
The diagnosis and the fix are right, and the subshell form matches what `nightly/SKILL.md` already does (`( cd "<worktree from prepare output>" && uv tool run tend@latest init )`), so the two recipes now agree with the third. Two points on the guard and the cleanup line.

**The guard misses the spelling this repo actually writes.** `cd\s+-\s*$` only fires when `cd -` is the entire tail of a line, so `cd - >/dev/null`, `cd - && …`, and `cd -  # back` all pass it. That gap isn't hypothetical: every `cd -` in the repo today is written `cd - >/dev/null` — three of them in `.claude/skills/running-tend/references/integration-test.md`, outside the scanned tree — so a reintroduction into `plugins/tend-ci-runner` in the house spelling lands green and the recipe regresses to exactly the failure this PR is fixing. Replacing the `$` anchor with a trailing `(?![\w-])` catches all three forms while still excluding `cd -- "$X"` and `cd -P /x`; run over the plugin tree it reports nothing after this change. The sibling `/tmp` test shows the other way to handle a prose mention of the banned token, if a shipped file ever needs to name this one: a content exemption (`"read-only" not in line`) rather than an anchor.

**The cleanup line runs even when the subshell aborted.** With `set -e`, a failing `git commit` — a rejecting pre-commit hook, or nothing staged because the `mv` landed somewhere unexpected — exits the subshell before the push, and the very next line force-removes the worktree. Since the commit never happened, no branch ref survives it, so the authored skill file is gone and the session has to regenerate the text to retry. Gating the removal on the subshell keeps the worktree around for that retry, and it lives in `$TMPDIR`, which is per-run scratch, so leaving it behind on the failure path costs nothing. Suggested on both copies below.

PR #1301, opened 2026-09-18
fix(token-report): price a run on when it finished, not when it started

### LINKED ISSUE #1300: review-runs: Step 2 prices on run creation while Step 1 censuses on completion, so a straddling run is counted and never priced
`review-runs` Step 1 and Step 2 window on different timestamps, so a run that starts before the sweep's anchor and finishes inside the window is counted in the census and priced by nobody. The runs this drops are by construction the *longest* ones, so the spend line is blind to exactly the sessions it exists to surface. On `max-sixty/worktrunk` that is the nightly, which has gone unpriced four times in sixteen days for **$74.73** of real work — once amounting to 44% of the window's published spend.

## The asymmetry

`list_recent_runs.py` is written for the straddle. It fetches from `anchor - CREATION_CUSHION` (24 h) and then admits a row on **completion**:

```python
CREATION_CUSHION = timedelta(hours=24)
created_since = (completed_after - CREATION_CUSHION).strftime("%Y-%m-%dT%H:%M:%S")
...
if conclusion and _parse_time(row["updatedAt"]) >= completed_after:
```

[`list_recent_runs.py#L27-L149`](https://github.com/max-sixty/tend/blob/ae17ee159b8a330221004384a3e4d1df6aba9eb2/plugins/tend-ci-runner/scripts/list_recent_runs.py#L27-L149)

Step 2's recipe has no such cushion. It converts the anchor to whole hours and hands that to `token_report.py`, which filters on **creation** (`gh run list --created ">=$since"`):

```bash
SINCE=$(cat "$TMPDIR/review-runs-since")
HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 ))
```

[`SKILL.md#L151-L155`](https://github.com/max-sixty/tend/blob/ae17ee159b8a330221004384a3e4d1df6aba9eb2/plugins/tend-ci-runner/skills/review-runs/SKILL.md#L151-L155)

The `+ 3599` rounding buys 0–60 minutes of lead, and the comment above it reads "rounded up so the whole band is priced" — which is the intent it misses. Anything that started earlier is censused and unpriced.

Each such run falls through a two-sweep gap, which is why no entry catches it:

- The sweep it **overlaps** cannot price it. The run is still in flight, so it has uploaded no artifact and carries no conclusion, and the census excludes it for want of one.
- The **next** sweep censuses it — `updatedAt` is after the anchor — but its `created_at` is already outside the 25 h creation window, so `token_report.py` never sees it.

## Four measured occurrences on `max-sixty/worktrunk`

Every one is `tend-nightly`; every cost is read from the run's own `token-usage.json`.

| Nightly run | Started → finished | Turns | Cost | Sweep that censused it | That sweep's published spend | Share |
|---|---|---|---|---|---|---|
| [33724232227](https://github.com/max-sixty/worktrunk/actions/runs/33724232227) | 09-03T06:39:21Z → 08:17:04Z (97.7 min) | 182 | **$21.08** | [33850413217](https://github.com/max-sixty/worktrunk/actions/runs/33850413217) (09-04) | $47.68 | **44%** |
| [34092448184](https://github.com/max-sixty/worktrunk/actions/runs/34092448184) | 09-07T06:48:09Z → 08:39:33Z (111.4 min) | 122 | **$12.87** | [34201181992](https://github.com/max-sixty/worktrunk/actions/runs/34201181992) (09-08) | $152.47 | 8% |
| [34743264540](https://github.com/max-sixty/worktrunk/actions/runs/34743264540) | 09-13T06:39:02Z → 08:12:38Z (93.6 min) | 146 | **$18.04** | [34819995486](https://github.com/max-sixty/worktrunk/actions/runs/34819995486) (09-14) | $167.03 | 11% |
| [35190842380](https://github.com/max-sixty/worktrunk/actions/runs/35190842380) | 09-17T06:40:05Z → 09:04:44Z (144.6 min) | 167 | **$22.74** | [35321221567](https://github.com/max-sixty/worktrunk/actions/runs/35321221567) (this sweep) | $182.35 | 12% |

The mismatch is already legible in the tracking log and no entry flagged it: each of those four censuses reads `2 nightly` while its spend line reads `nightly $7.73 (1)`, `$10.48 (1)`, `$9.71 (1)`, `$8.54 (1)`. The censused-but-unpriced row is the straddler in every case.

The 09-04 entry is the one worth reading twice. Its headline is "spend fell 74% to $47.68, breaking a three-window rise" — drawn on a figure that omits the window's single most expensive run. The true total is $68.76, and the fall is 62%.

## Why it matters beyond the total

The entries carry a window-over-window spend series, and escalation terms are written against it — this repo's log currently holds a "$150+ self-review spend in a window merging fewer than three bot PRs" term and a "$40 cumulative" term on no-op `ci-fix` boots, among others. Those are being evaluated against a number with a systematic notch wherever the nightly ran long, in the direction that suppresses escalation.

## Proposed fix

Give the cost window the same shape as the census window: fetch with a cushion, admit on completion. `token_report.py` already lists the runs itself, so this is a filter change rather than new plumbing — fetch from `since - CREATION_CUSHION` and keep a run when its `updatedAt` is at or after `since`. A `--since <iso>` argument would let Step 2 pass `$TMPDIR/review-runs-since` straight through and drop the hours arithmetic from the recipe, which also retires the `date -d ""` footgun its comment currently has to warn about.

If the filter stays creation-based, the minimal version is to give the recipe the census's own cushion:

```bash
# Match list_recent_runs.py's CREATION_CUSHION: a run that began before the
# anchor and finished inside the window is censused, so it must be priced too.
HOURS=$(( ( $(date -u +%s) - $(date -u -d "$SINCE" +%s) + 3599 ) / 3600 + 24 ))
```

That costs a wider fetch and leaves the session hand-subtracting a day of pre-anchor runs instead of an hour's, which is why the completion filter is the better of the two.

Either way, Step 2 is worth one line telling the session to reconcile the two: where the census counts *n* runs of a workflow and the spend line prices fewer, the difference is unpriced work, not an absence of it. That check is what would have caught all four of these from the entries alone.

tend-agent:
`token_report.py` filtered its run list on **creation** (`gh run list --created
">=$since"`) while `list_recent_runs.py` fetches from `since - CREATION_CUSHION`
and admits a run on **completion** (`updatedAt >= since`). A run that started
before the sweep's anchor and finished inside the window was therefore censused
by Step 1 and priced by neither sweep: the overlapping one has no artifact yet,
and the next one's creation window has already moved past it. Those runs are by
construction the longest in the window, so the published spend was short by
exactly the sessions the figure exists to surface — #1300 measures four such
nightlies on `max-sixty/worktrunk`, one of them 44% of its window's reported
total.

The two windows now have the same shape. `token_report.py` takes `--since <ISO>`
(or `--hours N` for an ad-hoc look), fetches with the same 24 h
`CREATION_CUSHION` the census uses, and keeps a run when its `updatedAt` is at
or after the anchor. review-runs Step 2 passes `$TMPDIR/review-runs-since`
straight through, which also retires the hours arithmetic and the `date -d ""`
footgun its comment had to warn about. The cushion widens the fetch only — the
admission filter still cuts at the anchor, so consecutive sweeps tile rather
than double-count.

Four regression tests cover the band: a run straddling the anchor is priced, a
run that finished before it stays out, the fetch reaches a day behind the
anchor, and the report header still names the anchor rather than the cushioned
fetch. The test harness's fake `gh` now honours `--created`, so the cushion is
load-bearing in the suite rather than assumed; a stale call site passing an hour
count in the prefix slot errors instead of silently reporting an empty fleet.

<details><summary>What this deliberately leaves out</summary>

The issue also proposes a line in Step 2 telling the session to reconcile the
census count against the spend count. With the windows aligned, the only
remaining divergence is a run that uploaded no readable artifact, which the
report already names in its own footnote (`N run(s) uploaded no readable
claude-session-logs artifact...`) — a second instruction for it would duplicate
guidance that already has a home.

`review-runs`' own cost gate is not in tension here: the defect is a wrong
published figure (an entry reported "spend fell 74%" on a total missing the
window's costliest run), not duplicated compute, and the change removes
scheduling arithmetic from the recipe rather than adding it.

</details>

Closes #1300 — automated triage


plugins/tend-ci-runner/scripts/token_report.py
```diff
@@ -237,15 +243,47 @@ def _report(jobs: list[dict[str, Any]], *, skipped: int, since: str) -> dict[str
...
+    parser.add_argument(
+        "prefixes",
+        nargs="*",
+        metavar="PREFIX",
+        help="extra workflow-name prefixes beyond 'tend-'",
+    )
+    options = parser.parse_args(args)
+    if hours := next((p for p in options.prefixes if p.isdigit()), None):
+        parser.error(f"PREFIX {hours!r} is an hour count — pass --hours {hours}")
+    if options.since:
```
tend-agent: An empty `--since` is falsy, so it silently falls through to the 168-hour default instead of erroring. `is not None` sends `""` into `_parse_time`, which raises and trips the `parser.error` below.

```suggestion
    if options.since is not None:
```

tend-agent (commented):
`--since ""` is accepted and silently prices a 168-hour window, so the recipe this PR writes into review-runs Step 2 fails open. `--since "$(cat "$TMPDIR/review-runs-since")"` collapses to `--since ""` whenever the anchor file is missing — `cat` writes to stderr and the substitution yields the empty string — and `if options.since:` then takes the `--hours 168` branch. Driving the new parser with an empty value through the test harness exits 0 and prints `1 runs since 2026-09-11T08:30:28Z`: a week's spend published as the night's sweep, under a header that looks well-formed.

That is the failure mode the prose this PR deletes was guarding (``The `cat` isn't optional: an unset `$SINCE` makes `date -d ""` today's midnight, not an error``), and the old positional caught it too, since `int("")` raised and the usage error fired. `review_runs_corrections.py`, reading the same anchor file, still rejects it explicitly (`if len(args) != 1 or not args[0]`). The inline suggestion restores the rejection: with `is not None`, `""` reaches `_parse_time` and errors out the way a whitespace-only value already does.

Separately, `CREATION_CUSHION`, `_stamp`, and `_parse_time` now exist in both `token_report.py` and `list_recent_runs.py`, tied only by a comment asserting the values match — and matching is the premise of the change, with nothing failing if one moves. Worth deciding whether they get a single home (both scripts already import `github_cli`) or whether the comment is the intended link. The copies have already diverged: only the new `_parse_time` defaults a naive timestamp to UTC.

PR #1306, opened 2026-09-18
skills: promote the cross-cutting references to skills, and rename running-in-ci to run-tend

max-sixty:
The References tables indexed sixteen `references/` files, and sessions were reading 26% of the ones their actions required — 11 of 42 gated actions across 32 sessions. Compliance tracked where the pointer sat rather than what it said: a reference named inline by the workflow skill got read, one named only by the central table did not. Per workflow it followed the inline-citation count almost exactly — review 10 citations/67%, triage 5/60%, nightly 3/19%, notifications 1/0%, weekly 0/0%.

The harness already carries an index that fires on the action rather than on choosing a task: the skill listing, whose name and description every session holds without reading anything. So the twelve cross-cutting references become skills in their own right.

| was | is |
|---|---|
| `trigger-context.md` | `/tend-ci-runner:respond-on-thread` |
| `posting.md` | `/tend-ci-runner:post-to-github` |
| `pr-creation.md` | `/tend-ci-runner:open-pr` |
| `pushing.md` | `/tend-ci-runner:push-commits` |
| `ci-monitoring.md` | `/tend-ci-runner:monitor-ci` |
| `grounded-analysis.md` | `/tend-ci-runner:ground-claims` |
| `directives.md` | `/tend-ci-runner:check-requester-access` |
| `dismissing-approval.md` | `/tend-ci-runner:dismiss-approval` |
| `other-repos.md` | `/tend-ci-runner:act-in-other-repos` |
| `session-logs.md` | `/tend-ci-runner:read-session-logs` |
| `proposing-tend-instructions.md` | `/tend-ci-runner:propose-instructions` |
| `triage`'s `fixing.md` | `/tend-ci-runner:fix-a-bug` |

`review`'s four stay references. A pointer gets read where the session already is when it acts — the listing, or a step in a skill it has loaded — and every session that needs those four arrives through a skill whose own step names them, including the `weekly` and `review-runs` sessions that reach two of them. `CLAUDE.md` states that criterion, rather than the reader-set one it had, which this PR's own `weekly` citations contradicted. Both References tables go with the promotion — the listing is the one index now, and `run-tend` names only the handful nearly every task needs. `nightly`, `weekly`, `notifications`, and `ci-fix` cite the skill at the step that takes the action, which is the placement the compliance data says gets read.

`running-in-ci` becomes `run-tend`, in the imperative like the rest.

## The description is the trigger, and Codex truncates it

Each description names a situation a session can recognise at the moment it is about to act, and fits inside Codex's per-skill share of the skills context budget. That share is 130 characters, measured against the real plugin install (both plugins, 22 runner skills) on Codex 0.155.0 rather than assumed: over it, Codex cuts the description mid-sentence and the trigger stops partway, which is the one thing this design cannot afford. Re-verified as untruncated after the change.

`test_skill_frontmatter_is_loadable` holds that ceiling and also catches an unquoted `: ` inside a description, which makes the whole frontmatter block invalid YAML — one of these descriptions had that, and nothing in the suite would have caught it before it shipped. It also pins the skill count the 130 was measured at — 24, spanning both plugins, since the install carries both and they draw on the one budget. The share falls as skills are added and four descriptions sit within four characters of the ceiling, so the next skill fails with a message to re-measure rather than silently truncating them. `test_plugin_skill_citations_resolve` replaces the table-index test's forward direction — a rename now fails here rather than leaving slash commands that resolve to nothing — and `test_skill_reference_citations_resolve` keeps the reverse one, failing on a `references/` file no skill cites, which loads in no session.

<details><summary>Where the 26% came from</summary>

Session-log artifacts from 32 recent runs across `max-sixty/worktrunk` and `max-sixty/tend`, spanning review, triage, nightly, weekly, notifications, and mention. For each gated Bash action taken (`gh pr create` / `gh issue create`, a PR body edit, a comment or review POST, `git push`), the check was whether the reference that action required had been read at an earlier tool call in the same session. 11 of 42 passed; of the 31 misses, 25 never read the file at all and 6 read it after the action.

The decisive evidence for *why* is a within-session control: weekly runs reliably read `integration-test.md`, the file their own workflow skill names, and zero files from the central table; nightly runs always read `nightly-cleaner.md`. Progressive disclosure works when the trigger is choosing a task and fails when the trigger is an action inside a plan already formed.

</details>

Follow-up, not in this PR: five consumer overlays still name `running-in-ci` in prose, two of them pointing at sections that have moved out of it.


generator/tests/test_repo_pins.py
```diff
@@ -812,31 +812,88 @@ def test_every_workflow_pins_the_same_tend_release() -> None:
...
+    in the listing — either way the skill stops being findable at the moment it
+    is needed, with nothing failing. `install-tend` is out of scope: a person
+    reads its two descriptions and invokes them by name, so neither the budget
+    nor `internal` applies.
+    """
+    yaml = YAML(typ="safe", pure=True)
+    broken = []
+    runner = REPO_ROOT / "plugins" / "tend-ci-runner" / "skills"
+
+    for path in sorted(runner.glob("*/SKILL.md")):
```
tend-agent: `DESCRIPTION_BUDGET` is a per-skill share of a listing budget that the comment above says shrinks as skills are added, but nothing ties the constant to the count it was measured at. Four descriptions already sit within four characters of 130 — `check-requester-access` and `propose-instructions` at 128, `push-commits` at 127, `ground-claims` at 126 — so the 23rd skill truncates their triggers mid-sentence with this test still green. That is the one failure the whole design can't absorb.

```suggestion
    paths = sorted(runner.glob("*/SKILL.md"))
    assert len(paths) == 22, (
        f"{len(paths)} skills, not the 22 DESCRIPTION_BUDGET was measured at — "
        "re-measure the share against the installed plugin before adding one"
    )

    for path in paths:
```

tend-agent (commented):
Three findings inline, all on what the new guards hold rather than on the citation rewrites.

Two are gaps in the guards this PR adds: the description ceiling is only valid at today's skill count, and orphan `references/` files stopped being detected when the table test went. The third is the `CLAUDE.md` criterion for what stays a reference, which this diff's own `weekly` citations contradict — that one needs a call rather than an edit.

I'm happy to push the first two as a commit if you want them.

generator/tests/test_repo_pins.py
```diff
@@ -812,31 +812,88 @@ def test_every_workflow_pins_the_same_tend_release() -> None:
...
+# A bundled skill invoked as a slash command. `<name>` and `NAME` placeholders
+# don't match, so prose about the citation form isn't read as a citation.
+PLUGIN_SKILL = re.compile(r"/tend-ci-runner:(?P<skill>[a-z0-9-]+)")
 
-    A `references/` file no row names loads in no session, and a row naming a
-    file that moved sends a session to a path that isn't there. The one symlink
-    into `shared/` is excluded: it is reached by `@` embedding in
-    `notifications` and by a pointer from `directives.md`, not by a row here.
+
+def test_plugin_skill_citations_resolve() -> None:
```
tend-agent, line 839: The test this replaces asserted `indexed == on_disk`, so it failed in both directions. This one only checks that a cited skill exists; the direction that caught a `references/` file nothing points at is gone. I added an empty `plugins/tend-ci-runner/skills/review/references/orphan.md` on this branch and the whole of `test_repo_pins.py` stayed green — a reference that loads in no session now ships silently, which is exactly the failure the removed docstring named.

`test_skill_reference_citations_resolve` already resolves the `(owner, file)` pairs it validates, so collecting them and comparing against `skills/*/references/*.md` restores the direction in about four lines.

CLAUDE.md
```diff
@@ -424,27 +425,26 @@ When adding to or editing files in `plugins/tend-ci-runner/skills/` or
...
-  `` `references/<file>.md` ``, with the owning skill in front where the file
-  belongs to another skill
-  (`` `/tend-ci-runner:running-in-ci`'s `references/posting.md` ``). That holds
+- **An action more than one workflow takes gets its own skill**, named for
+  the action and described by the situation that calls for it
+  (`post-to-github`, `push-commits`, `open-pr`). The skill listing every
+  session already carries is the index, so `run-tend/SKILL.md` names only the
+  handful nearly every task needs and sends the session to the listing for the
+  rest — don't rebuild a table of them. A `references/` file is for text one
+  skill owns and no other skill's session reaches (`review`'s four); a repo
```
tend-agent: This clause and the bullet's first sentence disagree about `approving.md`. "An action more than one workflow takes gets its own skill" selects it: `weekly` approves dependency PRs, and this PR adds two citations routing weekly sessions into `/tend-ci-runner:review`'s `references/approving.md` (Step 0, and Step 2 item 3). So naming `review`'s four as files "no other skill's session reaches" is false as of this diff — and `shared/review-gates.md`, which `review-runs` sessions read, already reaches `references/draft-mode.md` the same way.

The consequence is that the one approval gate two workflows share stays out of the listing that is now the index. Either promote it, or scope the rule to the skill that owns the *text* rather than to the sessions that reach it; as written, a later session applying the criterion gets a different answer than this PR did.

max-sixty:
All three are in `73ca958a`.

The budget now pins the count it was measured at (`SKILLS_MEASURED_AT = 22`), so a 23rd skill fails the test with a message saying to re-measure rather than raise it. `test_skill_reference_citations_resolve` collects the `(owner, file)` pairs it already resolves and compares them against `skills/*/references/*.md`, restoring the direction the table test covered — verified by adding your `orphan.md` and watching it fail.

On the third: the criterion was wrong, and promoting `approving.md` would be the wrong repair, since `weekly` and `review-runs` both reach the files they need from a step in a skill they have already loaded — the placement the compliance data says gets read. So the rule states what both cases follow: a pointer gets read where the session already is when it acts, which is the listing or a step in a loaded skill; text keeps its `references/` file while the skills whose steps cite it can be named, and becomes a skill where the need arises from an action any workflow might take.

tend-agent:
All three hold up. Verified against `73ca958a` in a clean worktree — this session's checkout has `CLAUDE.md` and `.claude/**` pinned to base, so both citation tests fail there on the old `running-in-ci` name and pass on the PR's actual tree.

- **Description budget** — `SKILLS_MEASURED_AT` fires as intended: a 23rd skill fails `test_skill_frontmatter_is_loadable` with the re-measure message rather than silently truncating the four descriptions sitting near the ceiling. The assert is symmetric, so *removing* a skill fails too; that direction is safe for the budget (the share grows), and re-measuring is still the right response, so nothing to change.
- **Orphan references** — restored. Adding back `plugins/tend-ci-runner/skills/review/references/orphan.md` now fails with `reference files no skill cites, so no session loads them`.
- **The criterion** — the rewrite resolves it, and better than promoting `approving.md` would have. "The skills whose steps cite it can be named" is a test a later session can apply and get this PR's answer, and it matches what's on disk: `weekly/SKILL.md` names `approving.md` at Step 0 and Step 2, `shared/review-gates.md` names `draft-mode.md`, both from a step in a skill those sessions have already loaded.

No remaining concerns from the review.

tend-agent (commented):
Three points on `73ca958`, all inline. Both guards it adds fire as intended; each has a gap at its edge, and the criterion rewrite left two places still carrying the claim it replaced.

- `test_plugin_skill_citations_resolve` doesn't match `${SKILL:<name>}` — the citation every session on both harnesses follows, and the one this PR's rename had to update by hand.
- `CLAUDE.md`'s "Which file" row and a sentence in `codex/agents-tail.md` still say a `references/` file is read only by its own skill's sessions.
- `SKILLS_MEASURED_AT` pins a count narrower than the basis the share was measured against.

Happy to push all three as one commit.

generator/tests/test_repo_pins.py
```diff
@@ -812,32 +829,98 @@ def test_every_workflow_pins_the_same_tend_release() -> None:
     )
 
 
-def test_references_table_indexes_every_bundled_reference() -> None:
-    """`running-in-ci`'s References table is the plugin's one index.
+# A bundled skill invoked as a slash command. `<name>` and `NAME` placeholders
+# don't match, so prose about the citation form isn't read as a citation.
+PLUGIN_SKILL = re.compile(r"/tend-ci-runner:(?P<skill>[a-z0-9-]+)")
```
tend-agent: `_prompt.py`'s `SKILL_REF.sub` substitutes the name blind, so a rename that misses `${SKILL:run-tend}` in `shared/system-prompt.md` renders `/tend-ci-runner:<stale>` into both harnesses' prompts and every session starts with none of the base rules — the failure this test's docstring names, in the one citation form it can't see. This PR's own rename had to update that placeholder by hand. `shared/system-prompt.md` is already in the `rglob("*.md")` walk, so the alternation is the whole fix; with it the scan resolves 136 citations and reports none broken.

```suggestion
# A bundled skill invoked as a slash command, or named by the system prompt's
# `${SKILL:<name>}` placeholder, which `_prompt.py` renders into that same
# invocation. `<name>` and `NAME` placeholders don't match, so prose about the
# citation form isn't read as a citation.
PLUGIN_SKILL = re.compile(r"(?:/tend-ci-runner:|\$\{SKILL:)(?P<skill>[a-z0-9-]+)")
```

CLAUDE.md
```diff
@@ -397,9 +397,10 @@ sessions in the gap each pay attention for text that changes nothing they do.
 | File | Read by | Ships |
 |---|---|---|
 | `shared/system-prompt.md` | every session, both harnesses | yes |
-| `running-in-ci/SKILL.md` | every session | yes |
+| `run-tend/SKILL.md` | every session | yes |
 | a workflow's `SKILL.md` | that workflow's sessions | yes |
-| a skill's `references/` | only the sessions taking that action | yes |
+| a per-action `SKILL.md` | only the sessions taking that action | yes |
+| a skill's `references/` | that skill's sessions, when the case arises | yes |
```
tend-agent: The bullet below now says two of `review`'s four are also reached by `weekly` and `review-runs` from their own steps, so this row still states the criterion it replaced. Read off the table, a rule both `review` and `weekly` sessions need can't live in `references/approving.md` — where it does live — and the author's options become a second copy or the promotion the bullet argues against. `codex/agents-tail.md` carries the same claim in shipped text ("`review` keeps `references/` files for the cases only it hits").

```suggestion
| a skill's `references/` | sessions of the skills whose steps cite it | yes |
```

generator/tests/test_repo_pins.py
```diff
@@ -812,32 +829,98 @@ def test_every_workflow_pins_the_same_tend_release() -> None:
...
+    is needed, with nothing failing. `install-tend` is out of scope: a person
+    reads its two descriptions and invokes them by name, so neither the budget
+    nor `internal` applies.
+    """
+    yaml = YAML(typ="safe", pure=True)
+    broken = []
+    runner = REPO_ROOT / "plugins" / "tend-ci-runner" / "skills"
+
+    paths = sorted(runner.glob("*/SKILL.md"))
+    assert len(paths) == SKILLS_MEASURED_AT, (
```
tend-agent: The share was measured with both plugins installed, but the pin counts only `tend-ci-runner`'s skills, so a skill added under `plugins/install-tend/skills/` shrinks the same listing budget and leaves this green — with four descriptions within four characters of 130, that is the case the pin exists for. Counting `(REPO_ROOT / "plugins").glob("*/skills/*/SKILL.md")` here (24 today) closes it while `paths` stays runner-only for the loop below. No suggestion inline because `SKILLS_MEASURED_AT` has to move in the same commit.

max-sixty:
All three are in `bc0590aa`.

`PLUGIN_SKILL` takes the alternation, so the scan covers `${SKILL:<name>}` as well and resolves 136 citations; breaking the placeholder fails the test. `SKILLS_MEASURED_AT` counts `plugins/*/skills/*/SKILL.md` — 24 today — since the share was measured with both plugins installed, while `paths` stays runner-only for the loop. The `Which file` row and `codex/agents-tail.md` now state the criterion the bullet does.

tend-agent:
Verified against `bc0590aa` in a clean worktree — this session's checkout has `CLAUDE.md` and `.claude/**` pinned to base. `test_repo_pins.py` is green there, 51 passed.

- **The budget's basis** — holds. Dropping a `plugins/install-tend/skills/probe/SKILL.md` in fails `test_skill_frontmatter_is_loadable` with `25 skills across both plugins, not the 24 the budget was measured at`, so a skill added under either plugin now moves the pin, and `paths` staying runner-only leaves the per-description loop where it was.
- **The criterion** — the `Which file` row and the `codex/agents-tail.md` sentence now say what the bullet says, and an `rg` for the replaced phrasings turns up nothing else carrying it.
- **The placeholder** — my finding was wrong; the alternation isn't buying anything. `test_shipped_prompt_skill_tokens_resolve_to_a_bundled_skill` has been on `main` since #1280 and already covers it: it runs `_prompt.SKILL_REF` over `shared/system-prompt.md` and `codex/agents-tail.md`, asserts each name resolves to a skill on disk, then renders both under every harness prefix and fails if `${SKILL` survives. Those two files are the only ones `_prompt.render` is ever called on (`compose_system_prompt.py` and `codex/runner.py`), and the repo's single `${SKILL:` occurrence is in `system-prompt.md` — so the widened `PLUGIN_SKILL` reaches no file the older test doesn't, and it's the weaker of the two on a malformed token, since it matches without requiring the closing brace. It's harmless as it stands. The one cost is that its new comment reads as if this test were the placeholder's guard, when the fuller one is elsewhere.

Nothing else outstanding from the review.

generator/tests/test_repo_pins.py
```diff
@@ -812,32 +829,103 @@ def test_every_workflow_pins_the_same_tend_release() -> None:
...
+        try:
+            front = yaml.load(head)
+        except YAMLError as error:
+            broken.append(
+                f"{name}: frontmatter is not YAML ({error.__class__.__name__})"
+            )
+            continue
+        if front.get("name") != path.parent.name:
+            broken.append(f"{name}: `name: {front.get('name')}` isn't the directory")
+        if not front.get("metadata", {}).get("internal"):
```
tend-agent: `yaml.load` raises `YAMLError` only for malformed YAML; valid YAML that isn't a mapping falls straight through to `front.get`. Probed at this head: a `SKILL.md` with no `---` block gives `AttributeError: 'str' object has no attribute 'get'`, and a bare `metadata:` key gives `AttributeError: 'NoneType' object has no attribute 'get'` — neither traceback carries `name`, so a contributor sees a line number in this file and has to bisect 22 skills to find theirs.

```suggestion
        if not isinstance(front, dict):
            broken.append(f"{name}: no `---` frontmatter block above the body")
            continue
        if front.get("name") != path.parent.name:
            broken.append(f"{name}: `name: {front.get('name')}` isn't the directory")
        if not (front.get("metadata") or {}).get("internal"):
```

tend-agent (commented):
One finding, on the frontmatter guard this branch adds rather than on `bc0590a`'s own hunks.

`test_skill_frontmatter_is_loadable` reaches `front.get(...)` on whatever `yaml.load` returned, and the two authoring slips it most needs to name both route around the curated message. A `SKILL.md` that forgets the `---` block leaves `head` as the whole file, which loads as a `str`; a `metadata:` key with nothing under it makes `front.get("metadata", {})` return `None`. Either way CI fails on a bare `AttributeError` naming neither the skill nor the problem, and the `broken` list is the reason this test is worth having.

Suggestion inline — happy to push it.

max-sixty:
Fixed in `efd30831` — the guard now names the skill and the problem for both cases. Probed at that head: a `SKILL.md` with no `---` block reports `no `---` frontmatter block above the body` rather than the `AttributeError`.

tend-agent (approved):

PR #1309, opened 2026-09-18
tests: record how the skill description budget is measured

max-sixty:
`test_skill_frontmatter_is_loadable` pins the skill count the 130-character description budget was measured at, and fails with "re-measure the share against the install" when a skill is added. Nothing said how, and the setup existed only in the session that measured it.

The comment above `DESCRIPTION_BUDGET` now carries it: install both plugins from the checkout into a throwaway `CODEX_HOME` with its own `codex login` (copying `~/.codex/auth.json` in would let a refresh rotate the chain your own login depends on), run any `codex exec`, and read Codex's "Skill descriptions were shortened to fit the skills context budget" warning as the signal. A description it cut, quoted back by the model, shows the length each share gets.

Comment-only; no behavior change.


tend-agent (commented):
The recipe copies the live `~/.codex/auth.json` into the throwaway home, which is the one step that can cost something outside the measurement. If Codex refreshes that credential during the `codex exec`, the refresh-token chain rotates inside `$CODEX_HOME` and the original is left stale — the hazard `install-tend`'s `SKILL.md` names when it rules the same copy out for the consumer install ("copying the user's ordinary `~/.codex/auth.json` would eventually break their local Codex login"). It only fires when the access token happens to be near expiry, so it is exactly the kind of thing the recorded recipe should carry rather than the next maintainer rediscovering it. Suggested caveat inline; it also asks for the temp home to be removed, since it holds a copy of the full bundle.

generator/tests/test_repo_pins.py
```diff
@@ -873,6 +873,19 @@ def test_plugin_skill_citations_resolve() -> None:
...
+#
+#     export CODEX_HOME=$(mktemp -d); cp ~/.codex/auth.json "$CODEX_HOME"
+#     codex plugin marketplace add "$PWD"
+#     codex plugin add tend-ci-runner@tend; codex plugin add install-tend@tend
+#     codex exec --skip-git-repo-check -s read-only "Quote your skills listing."
+#
+# Codex prints "Skill descriptions were shortened to fit the skills context
+# budget" when any description is cut, and a description it cut, quoted back by
+# the model, shows the length each share gets. Tighten the descriptions to fit
+# and re-run until the warning is gone.
```
tend-agent, line 894:
```suggestion
# and re-run until the warning is gone. Copying `~/.codex/auth.json` carries the
# rotation hazard `install-tend`'s SKILL.md names: a refresh inside the throwaway
# home rotates the chain there and leaves the local login to be redone with
# `codex login`. Remove `$CODEX_HOME` when done — it holds the full bundle.
```

max-sixty:
Taken, differently: the recipe now logs in afresh inside the throwaway home (`codex login`) instead of copying `~/.codex/auth.json` in, so the measurement gets its own refresh-token chain and the user's is never at risk — removing the hazard rather than warning about it. It also ends with `rm -rf "$CODEX_HOME"`. In `93eac449`.

tend-agent (approved):

PR #1311, opened 2026-09-19
fix(skills): correct run-tend's list of restored instruction paths and events

tend-agent:
`run-tend`'s **Instruction paths read as the base version on a PR** section is wrong in two ways about what the pre-session restore actually does, and both errors point the same direction: they tell a session a path or an event is *unpinned* when it is pinned. That is the direction that loses work — the section's own rule is "never stage one of these paths from the PR checkout", because `git add` copies the restored base content over the PR's edit, and a path the section doesn't list is a path the rule doesn't reach.

**`.agents/` is missing from the list.** [`pin-instruction-paths.sh`](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/shared/steps/lib/pin-instruction-paths.sh#L14) pins `**/.agents` and `**/.agents/**` alongside the `.claude` pair — Codex discovers repo-local skills there — and both `README.md` and `docs/security-model.md` name it. Only the skill omits it. A PR that edits `.agents/` therefore reads as unpinned to the session handling it.

**A relayed review event does restore.** The section says `tend-mention`'s relayed `repository_dispatch` "carries no PR payload and restores nothing". It carries `client_payload[pr]`, which is a PR payload for the purpose that matters: [`event_targets_pull_request`](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/shared/steps/prepare_agent_workspace.py#L133) returns true for any `client_payload` dict, so the run takes the `checkout_mention` path, checks out the PR head, and [returns the PR's base SHA](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/shared/steps/prepare_agent_workspace.py#L250) as `config_base_sha` — which is exactly what makes `restore_sensitive_config` run. The sentence was accurate when it was written; the mention path grew its PR-head checkout afterwards and the prose stayed put.

So a session woken by a review on a PR that edits `CLAUDE.md` or a skill currently reads the restored worktree as the PR's own version, reports the PR's additions as absent, and reads `git status`'s phantom modification as a contributor mistake — the three failures the section exists to prevent.

The rewritten paragraph states the condition structurally (restore happens where the checkout is a PR head) and names the cases that genuinely restore nothing: a schedule, an `issues` event, a mention on an issue thread, and a mention on a [closed PR](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/shared/steps/prepare_agent_workspace.py#L225), where the fallback is the default branch and there is nothing to pin away.

`test_run_tend_names_every_pinned_instruction_path` covers the path half: it reads `INSTRUCTION_PATHSPECS` out of the shell and asserts the section names every root. It fails on `['.agents']` against the current text. The event half isn't mechanically checkable — the restore's trigger is a Python branch, not a list — so it stays prose, and the paragraph now phrases it as a condition on the checkout rather than an enumeration of event names that a later topology change can silently invalidate.


generator/tests/test_repo_pins.py
```diff
@@ -1079,3 +1079,32 @@ def test_report_failure_is_told_the_running_version(harness: str) -> None:
...
+        spec.strip("'").removeprefix(":(glob)**/").removesuffix("/**")
+        for spec in declaration.group(1).split()
+    }
+
+    skill = (REPO_ROOT / "plugins/tend-ci-runner/skills/run-tend/SKILL.md").read_text()
+    heading = "## Instruction paths read as the base version on a PR"
+    assert heading in skill, f"{heading!r} was renamed — repoint this test"
+    section = skill.split(heading, 1)[1].split("\n## ", 1)[0]
+
+    missing = sorted(name for name in pinned if f"`{name}" not in section)
```
```suggestion
    missing = sorted(
        name
        for name in pinned
        if f"`{name}`" not in section and f"`{name}/`" not in section
    )
```

plugins/tend-ci-runner/skills/run-tend/SKILL.md
```diff
@@ -35,7 +35,7 @@ Anyone can ask for help with a problem they raise. A request that directs you at
...
+Before the session starts, both harnesses restore `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, `.claude/`, and `.agents/` at any depth from the PR's base commit. That runs on every event whose checkout is a PR head: `pull_request_target`, `issue_comment` on an open PR, and the relayed `repository_dispatch` carrying a review event, which names its PR and is checked out the same way. Those files are read at CLI startup before any permission gating, so the PR's copies must not be trusted. A session checked out on the default branch instead — a schedule, an `issues` event, a mention on an issue thread or on a closed PR — restores nothing, since that tree is reviewed code already. The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
```
tend-agent:
```suggestion
Before the session starts, both harnesses restore `CLAUDE.md`, `CLAUDE.local.md`, `AGENTS.md`, `.claude/`, and `.agents/` at any depth from the PR's base commit. That runs on every event whose checkout is taken from a PR rather than the default branch: `pull_request_target` (the merge ref), `issue_comment` on an open PR, and the relayed `repository_dispatch` carrying a review event, which names its PR and is checked out the same way. Those files are read at CLI startup before any permission gating, so the PR's copies must not be trusted. A session checked out on the default branch instead — a schedule, an `issues` event, a mention on an issue thread or on a closed PR — restores nothing, since that tree is reviewed code already. The restore touches the worktree only; the index and `HEAD` keep the PR's version. So on a PR that legitimately edits these paths:
```

tend-agent (commented):
Two accuracy points on the new text and its guard, plus one adjacent observation. Pushing the first two as a fix commit.

The guard's membership test is a prefix match — a backtick followed by the name — so `.claude` is satisfied by any backticked token that merely starts with those characters, `` `.claude.json` `` among them. That one sits in `restore-sensitive-config.sh`'s own `SENSITIVE` list, so it is a plausible thing for this paragraph to grow. Replacing `` `.claude/` `` with `` `.claude.json` `` in the section leaves the test green while a pinned path goes unlisted, which is precisely the failure it exists to catch. Anchoring on the closing backtick, accepting either the bare name or the name plus a slash, keeps `` `.claude/` `` matching and closes it.

"every event whose checkout is a PR head" doesn't describe the main case. `checkout_review` fetches `refs/pull/N/merge` first and only falls back to `refs/pull/N/head`, so a `pull_request_target` session sits on the merge commit, not the head. A session that tests the stated condition against its own detached checkout can conclude the restore didn't run — the same direction the section is being corrected away from.

Adjacent, outside the diff: `shared/steps/pin-instruction-files.sh` is no longer reachable. Both `claude/action.yaml` and `codex/action.yaml` run `prepare_agent_workspace.py`, and nothing else in the tree names the script outside `generator/tests/test_shared_steps.py` and the changelog. Its header still describes a Codex-only fork pin against the default branch, which is the split this PR's "both harnesses" sentence corrects — so the stale second story survives the fix. Deleting it along with its two parametrized test cases is a maintainer call rather than this PR's scope, so it is not in the fix commit.

PR #1313, opened 2026-09-19
docs(config): state the sandbox write set and read-only /tmp in tend.example.yaml

tend-agent:
`.config/tend.yaml`'s sandbox section says where files come from and how to add tools, but never says what the sandbox may write or what it refuses. This adds both constraints as comments, because each one lets a tool look available and then fail in a way that doesn't point at the sandbox.

**`/tmp` is read-only.** The write set is the disposable checkout and the sandbox home, where `$TMPDIR` points; no config key widens it. The .NET SDK is the sharp case and is now the example. The abort happens in `DotnetFirstTimeUseConfigurer.Configure()`, which runs ahead of every managed SDK command, so `build`, `restore`, `test`, `run`, `new` and `nuget` each fail identically — `dotnet test` included, which is the command a consumer configuring tend for a .NET repo reaches for first. Only the host-level queries (`--info`, `--version`, `--list-sdks`) skip the configurer, so the SDK looks usable. `DOTNET_SKIP_FIRST_TIME_EXPERIENCE`, `DOTNET_CLI_HOME`, `TMPDIR` and `NUGET_PACKAGES` do not avoid it, so there is no env workaround the text leaves out. The text does name the one route that works — a private mount namespace with a tmpfs over `/tmp`, which adds an empty filesystem rather than relaxing this one — because the paragraph otherwise reads as "a .NET repo cannot be maintained here", which is not true.

**`AF_UNIX` sockets are refused.** `socket(AF_UNIX, …)` returns `EPERM` — creation, not bind — while `socketpair` still works, so ordinary parent/child pipes are unaffected and only cross-process Unix-socket IPC is lost. This one is worth documenting rather than filing, because `probe_boundary()` in `shared/steps/agent_lifecycle.py` asserts it: the boundary fails closed if an `AF_UNIX` socket can be created, so it is a property consumers can rely on rather than one a future change might relax. MSBuild is the loud case and it is silent about it — a multi-project `dotnet build` whose worker nodes fail the handshake prints `Build FAILED.` with `0 Error(s)` and no diagnostic, which reads as a broken checkout. `-m:1` keeps MSBuild in one process.

Comment-only change to `docs/tend.example.yaml`, with the byte-identical mirror under `plugins/install-tend/skills/install-tend/references/` re-synced. Both additions sit above the `sandbox_env` list, so the reserved-set parity hook that parses that section is unaffected.

This documents the constraints; it does not decide whether the sandbox should gain a writable `/tmp` or a narrower writable path, which is what #1312 asks. Refs #1312.

<details><summary>Reproduction — the `/tmp` abort</summary>

Run as the sandbox user from a writable cwd, .NET SDK 10.0.400:

```
dotnet new console -o app  -> exit=1  System.IO.IOException ... : 'NuGet-Migrations'
dotnet build               -> exit=1  (same)
dotnet restore             -> exit=1  (same)
dotnet test                -> exit=1  (same)
dotnet run                 -> exit=1  (same)
dotnet tool list --global  -> exit=1  (same)
dotnet nuget locals --list -> exit=1  (same)
dotnet --list-sdks         -> exit=0
```

```
System.IO.IOException: The system cannot open the device or file specified. : 'NuGet-Migrations'.
One or more system calls failed: mkdtemp("/tmp/.dotnet.V921dk") == nullptr; errno == EROFS;
   at System.Threading.Mutex.CreateMutexCore(...)
   at NuGet.Common.Migrations.MigrationRunner.Run(...)
   at Microsoft.DotNet.Configurer.DotnetFirstTimeUseConfigurer.Configure()
   at Microsoft.DotNet.Cli.Program.ConfigureDotNetForFirstTimeUse(...)
```

</details>

<details><summary>Reproduction — the socket refusal and MSBuild's silence</summary>

Socket surface, as the sandbox user:

```
socket(AF_UNIX, SOCK_STREAM)  -> EPERM
socket(AF_UNIX, SOCK_DGRAM)   -> EPERM
socketpair(AF_UNIX)           -> OK
socket(AF_INET, SOCK_STREAM)  -> OK
```

Same result inside a private mount namespace, which is how the `/tmp` half can be worked around without touching this one.

A three-project reference graph, built with `/tmp` already made writable so the only remaining constraint is the socket:

```
$ dotnet build app/app.csproj
  Determining projects to restore...

Build FAILED.
    0 Warning(s)
    0 Error(s)

$ dotnet build app/app.csproj -m:1
  lib1 -> .../lib1.dll
  lib2 -> .../lib2.dll
  app  -> .../app.dll

Build succeeded.
```

A single-project build succeeds either way — MSBuild only reaches for worker nodes once there is more than one project to schedule.

</details>


docs/tend.example.yaml
```diff
@@ -203,6 +203,14 @@ bot_name: my-project-bot
 # home-scoped installs the agent needs, such as `cargo install`,
 # `pip install --user`, or `uv tool install`.
 #
+# The sandbox can write to the disposable checkout and to the sandbox home,
+# which holds `$TMPDIR`. `/tmp` itself is read-only there and no config key
+# makes it writable, so a tool that hard-codes a path under `/tmp` fails
+# inside the sandbox even where it works in a `setup:` step. The .NET SDK is
+# one: its first-run configurer builds a named `Mutex` under `/tmp`, so
+# `dotnet build`, `restore` and `new` abort with `EROFS` even though
+# `dotnet --info` succeeds.
```
tend-agent: The configurer runs ahead of every managed SDK command, not just these three — `dotnet test` and `dotnet run` abort identically. Naming the mechanism's reach instead of a three-command list keeps a reader from concluding `dotnet test` is unaffected.

```suggestion
# one: its first-run configurer builds a named `Mutex` under `/tmp`, so every
# SDK command that runs it — `build`, `restore`, `test`, `run`, `new`, `nuget`
# — aborts with `EROFS`. Only the host-level queries (`--info`, `--version`,
# `--list-sdks`) skip it, so the SDK looks usable.
```

tend-agent (commented):
The .NET example under-states the reach of the failure. It names `build`, `restore` and `new`, but the abort happens in `DotnetFirstTimeUseConfigurer.Configure()`, which runs ahead of every managed SDK command — on SDK 10.0.400 in this session's sandbox, `dotnet test`, `dotnet run`, `dotnet tool list --global` and `dotnet nuget locals all --list` each exit 1 with the same `mkdtemp("/tmp/.dotnet.*") == nullptr; errno == EROFS`. Only the host-level queries (`--info`, `--version`, `--list-sdks`) skip the configurer. `dotnet test` is the command a consumer configuring tend for a .NET repo reaches for first, so a reader who takes the three-command list as the affected set concludes that the case mattering most to them is fine — which is the conclusion this paragraph exists to prevent.

`DOTNET_SKIP_FIRST_TIME_EXPERIENCE`, `DOTNET_CLI_HOME` and `NUGET_PACKAGES` don't avoid it, so there is no env workaround the text is leaving out.

Suggestion inline; I'm pushing the same wording to both copies of the file.

Run as the sandbox user from a writable cwd, SDK 10.0.400:

```
dotnet new console -o app  -> exit=1  System.IO.IOException ... : 'NuGet-Migrations'
dotnet build               -> exit=1  (same)
dotnet restore             -> exit=1  (same)
dotnet test                -> exit=1  (same)
dotnet run                 -> exit=1  (same)
dotnet tool list --global  -> exit=1  (same)
dotnet nuget locals --list -> exit=1  (same)
dotnet --list-sdks         -> exit=0
```

```
System.IO.IOException: The system cannot open the device or file specified. : 'NuGet-Migrations'.
One or more system calls failed: mkdtemp("/tmp/.dotnet.V921dk") == nullptr; errno == EROFS;
   at System.Threading.Mutex.CreateMutexCore(...)
   at NuGet.Common.Migrations.MigrationRunner.Run(...)
   at Microsoft.DotNet.Configurer.DotnetFirstTimeUseConfigurer.Configure()
   at Microsoft.DotNet.Cli.Program.ConfigureDotNetForFirstTimeUse(...)
```

</details>

docs/tend.example.yaml
```diff
@@ -203,6 +203,29 @@ bot_name: my-project-bot
...
+# one: its first-run configurer builds a named `Mutex` under `/tmp`, so every
+# SDK command that runs it — `build`, `restore`, `test`, `run`, `new`, `nuget`
+# — aborts with `EROFS`. Only the host-level queries (`--info`, `--version`,
+# `--list-sdks`) skip it, so the SDK looks usable.
+#
+# A single command can still get a writable `/tmp` for itself by running in a
+# private mount namespace — `unshare --map-root-user --mount`, then
+# `mount -t tmpfs tmpfs /tmp` — which adds an empty filesystem rather than
+# relaxing this one. It applies only to the command it wraps, so each such
+# command carries it.
```
```suggestion
# A single command can still get a writable `/tmp` for itself by running in a
# private mount namespace, which adds an empty filesystem rather than relaxing
# this one. The checkout is under `/tmp` too, so the new filesystem hides it
# and it has to be carried across; without that the build fails a second time,
# now on a project file it can no longer resolve. The namespace lasts only as
# long as the command it wraps, so each such command carries the whole thing:
#
#     unshare --map-root-user --mount sh -c '
#       mkdir -p "$TMPDIR/ws" && mount --bind "$GITHUB_WORKSPACE" "$TMPDIR/ws"
#       mount -t tmpfs tmpfs /tmp
#       mkdir -p "$GITHUB_WORKSPACE"
#       mount --bind "$TMPDIR/ws" "$GITHUB_WORKSPACE"
#       exec dotnet test -m:1'
```

tend-agent (commented):
The mount-namespace route masks the checkout, so it fails at the case the paragraph offers it for. The agent workspace is always under `/tmp` — `prepare_agent_workspace.py` creates it with `tempfile.mkdtemp(prefix="tend-agent-workspace-", dir="/tmp")` — so `mount -t tmpfs tmpfs /tmp` hides it. The wrapped command keeps its inherited working directory, which is why `ls` still lists the tree, but nothing reaches the checkout by absolute path, and MSBuild resolves the project path against the absolute working directory before opening it, so a relative `app/app.csproj` fails too. A `dotnet build` through the recipe as written stops at `MSBUILD : error MSB1009: Project file does not exist.` — a second failure, with no more relation to the real problem than the first one had. Binding the checkout back over the tmpfs fixes it: suggestion inline, and I'm pushing the same wording to both copies of the file.

Separately, and left out of the suggestion: the route needs unprivileged user namespaces, which Tend only arranges on GitHub-hosted runners. `install-sandbox-runtime.sh` sets `kernel.apparmor_restrict_unprivileged_userns=0` there and deliberately leaves self-hosted policy alone, where the same prerequisite may be met by a bwrap-scoped AppArmor profile instead — and a profile scoped to `bwrap` wouldn't extend to an `unshare` the agent runs inside it. I have no self-hosted runner to test that on, so I'd rather not qualify the text on a guess; it's worth a clause if you know which way it goes.

Run as the sandbox user in this session's checkout (`/tmp/tend-agent-workspace-*/checkout`), .NET SDK 10.0.400.

The masking, with the recipe exactly as the diff writes it:

```
$ unshare --map-root-user --mount bash -c 'mount -t tmpfs tmpfs /tmp; ls "$GITHUB_WORKSPACE/CLAUDE.md"'
ls: cannot access '/tmp/tend-agent-workspace-.../checkout/CLAUDE.md': No such file or directory
```

A three-project build with the checkout masked versus carried across, otherwise identical:

```
masked       -> MSBUILD : error MSB1009: Project file does not exist.
carried over -> Build succeeded.  0 Warning(s)  0 Error(s)
```

The suggested preamble, checked under `set -e` in this sandbox: `mount --bind` and the tmpfs mount both return 0, and `$TMPDIR` is outside `/tmp` so the staging bind survives the mask. (`mount --move` also works but exits non-zero on the read-only `/etc/mtab`, which would abort a `set -e` script — hence the second `--bind`.)

The two claims the diff already made hold as written: `/tmp` is read-only, and `socket(AF_UNIX, …)` returns `EPERM` for both `SOCK_STREAM` and `SOCK_DGRAM` while `socketpair` and `AF_INET` succeed — matching the `probe_boundary()` assertion in `shared/steps/agent_lifecycle.py`. The multi-project `Build FAILED.` with `0 Error(s)`, and `-m:1` clearing it, reproduce too.

</details>

docs/tend.example.yaml
```diff
@@ -203,6 +203,37 @@ bot_name: my-project-bot
...
+# `--list-sdks`) skip it, so the SDK looks usable.
+#
+# A single command can still get a writable `/tmp` for itself by running in a
+# private mount namespace, which adds an empty filesystem rather than relaxing
+# this one. The checkout is under `/tmp` too, so the new filesystem hides it
+# and it has to be carried across; without that the build fails a second time,
+# now on a project file it can no longer resolve. The namespace lasts only as
+# long as the command it wraps, so each such command carries the whole thing:
+#
+#     unshare --map-root-user --mount sh -c '
```
tend-agent: Without `set -e` the preamble's failure is not the command's failure: a mount that doesn't take still leaves the wrapped command running against a masked checkout.

```suggestion
#     unshare --map-root-user --mount sh -c '
#       set -e
```

tend-agent (commented):
The recipe runs the wrapped command even when its preamble fails. `sh -c` without `set -e` continues past a failed `mount`, so if the staging bind doesn't take, the tmpfs still masks the checkout and the build reaches exactly the "project file does not exist" failure the paragraph above spends four lines warning about — the mount's own error is on stderr, but a build's output scrolls past it and the exit code comes from the build. `set -e` stops at the mount, where the message names the real cause. Suggestion inline; I'm pushing the same line to both copies of the file.

Run as the sandbox user in this session's checkout, with the staging path forced to fail (`$NOPE` unset in place of `$TMPDIR`) and everything else exactly as the diff writes it:

```
as written   mount: /tmp/.../checkout: special device /ws does not exist.
             ls: cannot access '/tmp/.../checkout/CLAUDE.md': No such file or directory
             exit=2   (the wrapped command ran, against a masked checkout)

set -e       mount: /tmp/.../checkout: special device /ws does not exist.
             exit=32  (the wrapped command never ran)
```

The recipe's success path is unaffected — as written it carries the checkout across, `/tmp` is writable inside, and both absolute and relative paths into the workspace resolve.

</details>

docs/tend.example.yaml
```diff
@@ -203,6 +203,14 @@ bot_name: my-project-bot
 # home-scoped installs the agent needs, such as `cargo install`,
 # `pip install --user`, or `uv tool install`.
 #
+# The sandbox refuses `AF_UNIX` sockets: creating one returns `EPERM`,
+# and the boundary asserts that at startup, so it is not something a config
+# key relaxes. Pipes between a process and its own children are unaffected.
+# A build tool that drives worker processes over a Unix socket loses them,
+# and MSBuild loses them silently — a multi-project `dotnet build` prints
+# `Build FAILED.` with `0 Error(s)` and no diagnostic. `-m:1` keeps MSBuild
+# in one process and makes real errors visible again.
```
tend-agent: The exempt set is `pipe()` and `socketpair()`, not the parent-child relationship — and .NET's named pipes, which is how MSBuild reaches its child nodes, are `AF_UNIX` sockets and fail.

```suggestion
# key relaxes. Anonymous pipes and `socketpair()` are unaffected; anything
# that opens a named socket is not, however closely its two ends are
# related. So a build tool that drives worker processes over one loses them
# — and .NET's "named pipes" are `AF_UNIX` sockets, so MSBuild loses its
# worker nodes silently: a multi-project `dotnet build` prints
# `Build FAILED.` with `0 Error(s)` and no diagnostic. `-m:1` keeps MSBuild
# in one process and makes real errors visible again.
```

tend-agent (commented):
`Pipes between a process and its own children are unaffected.` splits the sandbox's behaviour on the wrong axis, and the next two lines are the counterexample: MSBuild's worker nodes *are* its own children, and it talks to them over what .NET calls named pipes — which on Unix are `AF_UNIX` sockets and so fail. A reader who files MSBuild's parent-to-child pipes under "unaffected" gets the opposite of the paragraph's own conclusion. What actually survives the boundary is the mechanism, not the relationship: `pipe()` and `socketpair()` work however the ends are related, and anything that opens a named socket does not.

In this session's sandbox, .NET splits its own "pipes" exactly where the boundary does — `NamedPipeServerStream` throws while `AnonymousPipeServerStream` succeeds — so the word is doing real damage in a .NET-facing paragraph.

Suggestion inline; I'm pushing the same wording to both copies of the file. The rest of the paragraph holds as written: `socket(AF_UNIX, …)` returns `EPERM` for both `SOCK_STREAM` and `SOCK_DGRAM`, `probe_boundary()` runs ahead of `sandbox_setup:` and fails the run if one can be created, and the multi-project `Build FAILED.` with `0 Error(s)` reproduces with `-m:1` clearing it.

Run as the sandbox user, .NET SDK 10.0.400. `/tmp` is read-only in this session (the action runs the released pin, which predates #1324), so the dotnet probes ran inside the private mount namespace this PR used to document.

```
NamedPipeServerStream:     SocketException: Permission denied
AnonymousPipeServerStream: created OK
```

```
socket(AF_UNIX, SOCK_STREAM) -> EPERM
socket(AF_UNIX, SOCK_DGRAM)  -> EPERM
socketpair()                 -> OK
pipe()                       -> OK
```

Three-project solution, otherwise identical invocations:

```
dotnet build s.slnx        -> Build FAILED.   0 Warning(s)  0 Error(s)
dotnet build s.slnx -m:1   -> Build succeeded. 0 Warning(s)  0 Error(s)
```

</details>

PR #1319, opened 2026-09-19
skills(ground-claims): read a past run's code at that run's commit

### LINKED ISSUE #1315: ground-claims: a diagnosis reads the checkout, so a claim about what a past run executed cites code that did not exist then
A session diagnosing a past CI failure reads the repository from its checkout, which is at the current default branch. When the answer is a claim about *what that run executed*, that is the wrong tree — and nothing in the bundled skills says so. The one instruction that does tell a session which SHA to use points at the wrong one: `post-to-github` says "Get the SHA with `git rev-parse HEAD` before composing the link", and `check_body_links.py` then validates the result, because the SHA is real and the link resolves. The claim built on top of it is false and everything downstream reports green.

## What happened

[max-sixty/leaf#658](https://github.com/max-sixty/leaf/issues/658) is a `ci-fix` tracker for `publish-site` failures. A maintainer asked the bot to re-examine the cause; the [answer posted at 2026-09-18T19:09:44Z](https://github.com/max-sixty/leaf/issues/658#issuecomment-5734921407) explained three failing runs like this:

> `reader_session` already treats a rollout 503 as a wait, but only when it carries `Retry-After`. These reach `check(activated.ok, …)` instead, so they carry no `Retry-After`.

Every code link in that paragraph is pinned to `dd8e0de56754b8d2efc533eb89f52841af11a3b5` — the session's own `HEAD`. The three runs it explains ran at `c9a5f207`, `de1f4a4e` and `5311ed03`. At all three, `scripts/verify_site.py` contains **no** `rolling_out` and **no** `Retry-After` at all:

```console
$ for s in c9a5f207 de1f4a4e 5311ed03 dd8e0de5; do
    printf '%s ' "$s"; git show "$s:scripts/verify_site.py" | grep -c 'rolling_out\|Retry-After'
  done
c9a5f207 0
de1f4a4e 0
5311ed03 0
dd8e0de5 3
```

The branch the comment says those runs took *instead of* the `Retry-After` test did not exist when they ran. It arrived in `2684c18d` at `2026-09-18T09:01:50Z`; the runs are from `07:42`, `08:27` and `08:35Z`. The maintainer corrected it the next morning — "reads current `main` against runs that predate it" — and the thread cost three comments to land a cause that one would have carried.

## The mechanism is structural, and the session log shows it cleanly

From [run 35383483872](https://github.com/max-sixty/leaf/actions/runs/35383483872)'s session log, in order:

```
grep -n "Retry-After\|retry_after\|still starting\|503" scripts/verify_site.py
sed -n '350,420p' scripts/verify_site.py
git rev-parse HEAD
...compose the comment, pinning every link to that HEAD...
for s in 0309bd07 c9a5f207 de1f4a4e 5311ed03; do gh api "repos/max-sixty/leaf/commits/$s"; done
```

The session had the failing runs' SHAs — it resolved them explicitly — and still read the file from the working tree. That is the default path, not a lapse: `grep`, `sed` and `Read` all answer from the checkout, and the checkout is current `main` by construction on a `workflow_run` or scheduled trigger. Replayed ten times, ten sessions read the same wrong tree unless one independently thinks to reach for `git show <run-sha>:<path>`.

There is precedent for treating this class as structural rather than as a stochastic accuracy lapse. `run-tend` already carries the same shape for a different cause — "**Instruction paths read as the base version on a PR** … Read the PR's version with `git show HEAD:<path>` before making any claim about what these files contain." The rule is identical; only the direction differs.

## Proposed fix

A short rule in `ground-claims` under **Recurring hallucination shapes**, along these lines:

> **A claim about what a past run executed is read at that run's own commit.** The checkout is the current default branch; a run from yesterday is not. Before explaining a failure in terms of a branch, guard, or helper, read it at the run's `head_sha` — `gh api repos/{owner}/{repo}/actions/runs/<id> --jq .head_sha`, then `git show <sha>:<path>` — and pin the comment's links to that SHA, not to `HEAD`. A fix that landed between the run and the session inverts the reading: the branch you are citing may be the one that was added *because* of the failure you are explaining.

`ci-fix` step 2 is the other natural home, or in addition, since that is where the reader already has a run id in hand. `post-to-github`'s `git rev-parse HEAD` line could also name the exception directly, since that line is what supplied the wrong SHA here.

This is not the general "be accurate" ask — tend's consumers already carry plenty of that. It is one named default that silently answers a different question than the one being asked, with a one-command remedy.

tend-agent:
A session diagnosing a past CI failure reads the repository from its checkout, which is at the current default branch — so an explanation of why a run failed can cite a branch, guard, or helper that did not exist when that run executed. Nothing downstream catches it: the SHA is real and the permalink resolves, so `check_body_links.py` passes and the wrong cause ships as a confident diagnosis.

`ground-claims` now carries the rule under **Recurring hallucination shapes**: resolve the run's `head_sha`, read the file at that SHA, and pin the body's links to it. The read goes through the contents API rather than `git show <sha>:<path>`, because the checkout holds only what a ref reaches — the head of a squash-merged or deleted PR branch is absent from it, which is most of the runs a session gets asked to explain, and a run in another repository is never there at all. `git show` also fails into the rule's own failure mode rather than out of it: the `fatal:` goes to stderr while a `| grep -c` downstream prints `0`, indistinguishable from a genuine no-match, which reads as "the guard was absent at that run". The contents call is chained off the run lookup's exit status for the same reason: an empty `$SHA` makes `ref=` serve the default branch, with a clean exit and plausible content.

`post-to-github`'s `git rev-parse HEAD` line is the instruction that supplied the wrong SHA, so it names the exception and points at that section rather than restating the rule.

Text only. `uv run pytest -k skill` passes, including `test_skill_reference_citations_resolve`, which is what validates the new cross-reference, and the recipe was run end-to-end against a run in this repo before it shipped.

Closes #1315


tend-agent (commented):
The `git show <sha>:<path>` half of the recipe is unavailable in the sessions that most need it, and where it fails it fails into the same wrong answer the rule exists to prevent.

`fetch-depth: 0` fetches everything reachable from a ref, and nothing else. This repo squash-merges and deletes the branch, so a merged PR's head commit is reachable from no ref and is simply absent from the checkout — of the eight most recently merged PRs here, `git cat-file -e <headRefOid>` finds none of them. Default-branch runs are fine, which is `ci-fix`'s case; a run on a PR branch is not, which is what `review-runs`, the notification drain, and a review session explaining an earlier run are usually looking at. `review-reviewers` never has the object at all — its checkout is a different repository from the runs it reads.

The failure mode is the part that matters. The absent object does not announce itself as absent:

```console
$ git show dfdb610ef17e80e2c9839e7b244cc06e38f950ef:README.md
fatal: path 'README.md' exists on disk, but not in 'dfdb610ef17e80e2c9839e7b244cc06e38f950ef'
```

That message reads as "the file was not there at that commit". Piped — which is the idiom [#1315](https://github.com/max-sixty/tend/issues/1315) itself uses as its proof, `git show "$s:scripts/verify_site.py" | grep -c 'rolling_out\|Retry-After'` — the fatal goes to stderr and stdout carries `0`, exit `0`. A session reads `0` and concludes the guard did not exist when that run executed. That is the hallucination this rule was written to stop, now produced by the rule's own recipe, and with the same false confidence: a clean exit and a plausible number.

Fetching at the SHA has none of these cases, and covers the local-commit one too:

```bash
gh api "repos/{owner}/{repo}/contents/<path>?ref=$SHA" -H 'Accept: application/vnd.github.raw'
```

I'm pushing that as the recipe. `post-to-github`'s cross-reference and the choice of `ground-claims` over `ci-fix` as the home both look right to me and I haven't touched them.

<details><summary>How the object absence was checked</summary>

```console
$ for n in $(gh pr list --state merged --limit 8 --json number --jq '.[].number'); do
    sha=$(gh pr view $n --json headRefOid --jq .headRefOid)
    git cat-file -e "$sha" 2>/dev/null && echo "$n local" || echo "$n absent"
  done
1310 absent
1309 absent
1308 absent
1307 absent
1306 absent
1305 absent
1303 absent
1301 absent
```

The same SHA fetches fine over the API, which is why the replacement recipe holds where `git show` does not:

```console
$ gh api "repos/{owner}/{repo}/contents/README.md?ref=dfdb610ef17e80e2c9839e7b244cc06e38f950ef" \
    -H 'Accept: application/vnd.github.raw' | head -1
<h1><img src="assets/logo-512.png" alt="tend logo" width="50" align="absmiddle">Tend</h1>
```

`check_body_links.py` is unaffected either way — it resolves a SHA through `repos/{slug}/commits/{sha}`, and a run head resolves there whether or not the branch still exists.

</details>

tend-agent (commented):
Two things in the new recipe, both the shape the last commit set out to fix — a path that returns a plausible answer rather than failing.

**An empty `$SHA` reads the default branch.** If the run lookup fails and leaves `$SHA` empty, `ref=` with no value serves the repository's default branch and exits 0:

```console
$ gh api "repos/max-sixty/tend/contents/README.md?ref=" -H 'Accept: application/vnd.github.raw' | head -1
<h1><img src="assets/logo-512.png" alt="tend logo" width="50" align="absmiddle">Tend</h1>
```

That is the reading the rule exists to prevent, arriving with a clean exit and content that looks right. `/tend-ci-runner:review`'s posting recipe already guards the same shape — "Read the sha first and bail if it isn't there" — and chaining the contents call off the run lookup's exit status covers it here.

**`| grep -c` exits 1, not 0.** `grep -c` returns 1 when it selects no lines, so the pipeline's status is 1 whether the object was missing or the guard was genuinely absent:

```console
$ git show <absent-sha>:README.md 2>/dev/null | grep -c Tend
0
$ echo $?
1
```

The argument survives — `0` on stdout is indistinguishable from a real no-match — but the exit status as stated is wrong, in the section telling the reader not to assert unrun tool behavior. Pushing both fixes.

One judgment call left alone: `head_sha` is the PR's head commit, not the merge commit a `pull_request` run checks out (`refs/pull/N/merge`), so for a PR-triggered run the file at that SHA can differ from the tree that actually executed when base moved after the branch point. Saying so costs a sentence in a skill that is meant to be brief, and for the code a diagnosis usually asks about the two read the same — so it seems a maintainer's call rather than mine.

PR #1320, opened 2026-09-19
skills(review-runs): find a workflow wedged behind a run parked in `waiting`

### LINKED ISSUE #1314: review-runs: a run parked in GitHub's pre-job `waiting` state holds its concurrency group forever, and nothing in tend can see it
A scheduled run parked in GitHub's pre-job `waiting` state holds its concurrency group forever. With `cancel-in-progress: false`, every later tick takes the single pending slot and is cancelled by its successor, so the workflow stops running entirely and nothing in tend notices. On `max-sixty/leaf` this took `tend-notifications` down for **6h40m** and burned 23 scheduled ticks. It is the second occurrence there; the first, on 2026-09-01, ran 71 minutes.

Nothing self-heals it. Today's `review-runs` sweep found it only because the census showed 23 consecutive `cancelled` rows and the session went looking for what they were queued behind.

## The wedge

[Run 35411904142](https://github.com/max-sixty/leaf/actions/runs/35411904142) (`tend-notifications`, `schedule`) was created at `2026-09-19T01:12:14Z` and was still `status: waiting` at `07:52Z`, its single job `waiting` with `runner_id: null`. Its environment gate had nothing to release it:

```console
$ gh api repos/max-sixty/leaf/actions/runs/35411904142/pending_deployments
[{"environment":{"name":"tend"},"wait_timer":0,"wait_timer_started_at":null,
  "current_user_can_approve":false,"reviewers":[]}]
```

No timer, no reviewer, nothing to approve. Meanwhile every 15-minute tick from `01:32:46Z` to `07:45:44Z` — 23 runs — was created, sat pending, and was cancelled at the exact moment the next one was created. The last drain to actually run was [35410956291](https://github.com/max-sixty/leaf/actions/runs/35410956291) at `00:55Z`.

`gh run cancel 35411904142` cleared it: the pending tick [35430199636](https://github.com/max-sixty/leaf/actions/runs/35430199636) moved to `in_progress` within ten seconds.

## Why nothing in tend sees it

Four mechanisms that look like they should catch this each miss it for a different reason.

- **The Step 1 census cannot see the wedged run.** `list_recent_runs.py` passes `--status completed` and admits a row only on a non-null `conclusion`. A run in `waiting` has neither, so it is absent from the census by construction — and the 23 cancellations it causes *are* in the census, reading as ordinary concurrency replacement.
- **No outage row is written.** The run never fails, so `report-failure` never fires and the `tend-outage` drain has nothing to read. This is adjacent to #1067 but not the same: there the job got no runner and the run still concluded; here the run never concludes at all.
- **`red_default_branch_runs.py` does not apply.** A wedge is not a red row on the default branch.
- **`timeout-minutes` cannot bound it, and is not the scheme anyway.** The job never starts, so no job-level cap applies. That is measurable rather than inferred: this run sat for **400 minutes** against GitHub's 360-minute default job cap and was not killed. Separately, #1023 was closed and #1024 classified distributed timeout settings as a configuration scheme, so the cap is not the lever here in any case.

## Proposed fix

One diagnostic in `review-runs` Step 1's **Reconcile live work**, beside the red-default-branch and Dependabot checks — a workflow whose queue is dead is live work in exactly the same sense:

```bash
gh api "repos/$GITHUB_REPOSITORY/actions/runs?status=waiting&per_page=50" \
  --jq '.workflow_runs[] | {id, name, created_at, html_url}'
```

A tend run older than a few of its own scheduling intervals and still `waiting` is wedged. Confirm with `pending_deployments` — `wait_timer: 0` with an empty `reviewers` is a gate that will never open — then `gh run cancel <id>`, and the successor starts immediately.

Worth distinguishing from the `queued` shape, which needs different handling: `max-sixty/leaf` also carries [34749167699](https://github.com/max-sixty/leaf/actions/runs/34749167699), `queued` since 2026-09-13, where `gh run cancel` answers `Cannot cancel a workflow run that is completed` while the run API still reports it `queued`. That one is a GitHub bookkeeping inconsistency holding nothing live, and the guidance should say so rather than sending a session to fight it.

## Scope

Every consumer is exposed: the generated workflows all declare `environment: name: tend`, and the scheduled ones pair it with `cancel-in-progress: false`. The consequence scales with the cadence — `tend-notifications` at `*/15` loses four drains an hour.

On `leaf` the cost stayed at waste: `tend-mention` carried the live threads through the blackout, and the live scan at the end of the window found every open issue and PR already answered on its latest activity. The backstop being down for a third of a day with no signal is the part worth fixing.

tend-agent:
A scheduled run parked in GitHub's pre-job `waiting` state never concludes, so it holds its concurrency group indefinitely. Under `cancel-in-progress: false` — which every generated scheduled workflow carries, paired with `environment: tend` — each later tick takes the one pending slot and is replaced by its successor, and the workflow stops running with nothing red to show for it. Step 1's census admits a row only on a non-null `conclusion`, so the parked run is absent by construction and the replacements it causes read as ordinary concurrency.

`review-runs`'s **Reconcile live work** now lists it beside the red-default-branch and Dependabot checks, since a workflow whose queue is dead is live work in the same sense. The remedy is `gh run cancel` on the parked run, after `pending_deployments` confirms the gate has no timer and no reviewers to release it.

The wedge test covers event-driven runs too, not just cron ones: `environment: tend` comes from a shared template that every generated workflow but the secretless `tend-mention-relay` uses, and `tend-review`, `tend-mention`'s handle job and `tend-ci-fix` pair it with `cancel-in-progress: false`, so a parked run there strands a PR's review queue the same way. Those have no scheduling interval to measure staleness against, so the bullet names the event that created the run as the bound instead.

The bullet also separates `queued`, which looks similar and needs the opposite handling. I reproduced that half here rather than taking it from the report: `tend-notifications` run `34748600612` has been `queued` since 2026-09-13, and `gh run cancel` on it answers `Cannot cancel a workflow run that is completed` while the runs API still reports `status: queued, conclusion: null`. Nothing is held — this repo's `tend-notifications` has run throughout — so the text says to leave it rather than fight it.

Text only, and deliberately one bullet: #1069 established that the recovery half of this class of finding belongs in a small report-or-remedy step, not a branching instruction table. `uv run pytest -k skill` passes.

Closes #1314


plugins/tend-ci-runner/skills/review-runs/SKILL.md
```diff
@@ -101,6 +101,15 @@ As a daily backstop for delayed notifications, retention, edited activity, and r
...
+
+  ```bash
+  gh api "repos/$GITHUB_REPOSITORY/actions/runs?status=waiting&per_page=50" \
+    --jq '.workflow_runs[] | {id, name, created_at, html_url}'
+  ```
+
+  A tend run still `waiting` after several of its own scheduling intervals — or, on an event-driven workflow, long after the event that created it — is wedged. Every generated workflow carries the `tend` environment, so any of them can park. `pending_deployments` on it confirms which kind: `wait_timer: 0` with an empty `reviewers` is an environment gate with nothing left to release it, so `gh run cancel <id>` is the remedy and the pending successor starts. `queued` is a different state and not this shape — `gh run cancel` there answers `Cannot cancel a workflow run that is completed` while the runs API still reports the run `queued`. That is GitHub bookkeeping holding nothing live; leave it rather than fighting it.
```
```suggestion
  A tend run still `waiting` after several of its own scheduling intervals — or, on an event-driven workflow, long after the event that created it — is wedged. Every generated workflow but the secretless `tend-mention-relay` carries the `tend` environment, so any of the rest can park. `pending_deployments` on it confirms which kind: `wait_timer: 0` with an empty `reviewers` is an environment gate with nothing left to release it, so `gh run cancel <id>` is the remedy and the pending successor starts. `queued` is a different state and not this shape — `gh run cancel` there answers `Cannot cancel a workflow run that is completed` while the runs API still reports the run `queued`. That is GitHub bookkeeping holding nothing live; leave it rather than fighting it.
```

tend-agent (commented):
`tend-mention-relay` is the exception to "every generated workflow carries the `tend` environment" — `mention-relay.yaml.j2` calls no `environment()` macro, and its header explains why: the review events it carries run on `refs/pull/N/merge`, which the operational secrets' environment does not admit, so the workflow holds no secret. A run with no environment has no deployment gate and cannot enter `waiting` at all, so the universal form is both false and slightly wider than the point it is making. The clause is load-bearing only to stop the check being read as cron-only, and the narrower form still does that — suggestion inline.

Issue #1318, opened 2026-09-19
sandbox_env is the only route for a third-party credential, so a contributor's code can read it


max-sixty:
Tend's proxy injects the bot PAT and the Anthropic credential into requests, so neither value ever enters the agent's environment. A third-party credential has no such route: `sandbox_env` is the only way to give the agent one, and `action.yaml` documents that as the launch environment, which `sandbox_setup` and every command the agent runs inherit.

That leaves the value readable by whatever code the session runs, and some of that code is a contributor's. On `review`, and on a mention on a pull request, `prepare_agent_workspace.py` checks the pull request out, and `sandbox_setup` builds that tree before the agent starts — in max-sixty/leaf's case `uv sync --frozen` against the pull request's own lock and `pyproject.toml`. The notifications poll starts on `main`, but `tend-mention-relay.yaml` leaves fork pull requests to it, and reviewing one means fetching its code.

The case that prompted this is max-sixty/leaf#825, which gives leaf's tend agent a read-only Cloudflare token so it can read the deployed site's logs when an issue or a red `publish-site` concerns them. The safety rule ended up as a GitHub expression in `.config/tend.yaml` that empties the value except on `issues`, `issue_comment` on an issue, and `ci-fix`'s `workflow_run` from this repository, plus an instruction in the repo's own skill to run a pull request's code with the variable unset. Both live in the consumer's config and restate things tend already knows: which checkout mode a workflow uses, and whose code the event carries. Every repo that wants an agent to read a third-party service will write that rule again, and a repo that doesn't think of it gets no warning.

The proposal is to let `.config/tend.yaml` declare a credential against a host — the shape `sandbox_env` has now, plus the host it belongs to — and have the proxy inject it into requests to that host the way it already does for GitHub and Anthropic. Code in the sandbox can still use the credential while the session runs, which is the point of giving it one, but it cannot carry the value out. Tend would also be the one deciding whether a session handling untrusted code gets the injection at all, since it knows the checkout mode and the event's provenance, and the consumer's config would carry no rule about it.

> _This was written by Claude Code on behalf of max-sixty_

tend-agent:
Nothing in the proxy is host-generic today, so this is a gap rather than a missing knob. Three sites hold the assumption that there are exactly two credentials:

[...]

PR #1321, opened 2026-09-19
docs(config): warn that a sandbox_env secret is readable by a PR's own code

tend-agent:
`sandbox_env` is the only way to give the agent a third-party credential, and the sandbox launch environment is inherited by every command the sandbox runs. On `tend-review` and on a mention on a pull request, those commands include the pull request's own code — `sandbox_setup:` builds that tree before the agent starts — so a secret placed there is readable by anyone who can open a pull request. The key's documentation said nothing about this, so a repo that didn't think of it got no warning.

This adds a paragraph to the `sandbox_env` entry in `docs/tend.example.yaml` (and its `install-tend` mirror) stating the exposure and what a consumer can do about it today:

- **The contrast with tend's own credentials**, which a runner-side proxy injects into the requests that need them, with the one exception named — Codex under subscription auth stages an expiring access-only `auth.json` inside the sandbox, as `docs/security-model.md` records.
- **Least privilege on the credential**, called out as the mitigation that holds everywhere: `sandbox_env` is a single top-level block rendered into every workflow's action call, and sandbox egress is not destination-allowlisted, so a workflow that only *reads* attacker-supplied text still holds the value and can be steered into moving it.
- **Withholding the value from the events that carry unreviewed code**, which closes the largest path. The value renders into the workflow verbatim, so a GitHub Actions expression works; the example names the events that get the secret rather than the ones that don't, because an empty string is falsy in Actions expressions and the obvious `<cond> && '' || secrets.X` form hands the secret over on every event.

It documents current behavior only. #1318 proposes removing the exposure by letting `.config/tend.yaml` declare a credential against a host and having the proxy inject it; if that lands, this paragraph changes with it.

No code paths change. `pre-commit run --all-files` passes, including the `sync install-tend references/ from canonical sources` hook that enforces the mirror.

Refs #1318


tend-agent (commented):
Three accuracy problems in the new paragraph. Fix pushed to this branch in the same session.

**The credential-isolation claim is too absolute.** "Tend's GitHub and model credentials never enter the sandbox" doesn't hold for Codex under subscription auth: the `Stage subscription auth (sandbox)` step in [`codex/action.yaml`](https://github.com/max-sixty/tend/blob/main/codex/action.yaml) copies the access-only `auth.json` into `$AGENT_HOME/.codex/auth.json`, and [`docs/security-model.md`](https://github.com/max-sixty/tend/blob/main/docs/security-model.md) states it — "Under subscription auth, it additionally receives an expiring access-only `auth.json`, but not the rotating refresh token." As written, a Codex-subscription consumer reads a guarantee this repo's own threat model contradicts. Naming the exception costs a clause and keeps the contrast the paragraph is drawing.

**The `github.event_name` mitigation is narrower than the exposure it's offered against.** `sandbox_env` is one top-level block rendered into every workflow's action call — there is no per-workflow form — so the value sits in the agent's env on `tend-triage` and the rest, not only on the two events that check out PR code. Sandbox egress is not destination-allowlisted (`docs/security-model.md`: "Tend's HTTP broker currently tunnels arbitrary destinations"), so an issue body that steers the agent moves the value out of a run the event gate cleared. Gating still closes the largest path; offering it as a peer of least privilege overstates what it buys.

**The obvious form of the suggested expression fails open.** An empty string is falsy in Actions expressions, so the natural `${{ github.event_name == 'pull_request_target' && '' || secrets.MY_TOKEN }}` hands over the secret on every event, including the one it was meant to exclude. The paragraph points at the technique without the ordering it depends on. The revision names the events that *get* the value rather than the ones that don't — allowlist-shaped, and it puts the secret on the `&&` side — and shows the expression.

`pre-commit run --all-files` passes on the revision, including the `sync install-tend references/` mirror hook.

tend-agent (commented):
One accuracy problem left in the new paragraph. Fix pushed to this branch in the same session.

**The clause sorting workflows into "runs attacker-supplied code" and "only reads attacker-supplied text" doesn't hold, and it sits immediately before the gating recommendation.** The workspace is the only thing an event decides: `checkout_review` and `checkout_mention` in `shared/steps/prepare_agent_workspace.py` are the sole modes that fetch a PR ref, everything else gets `checkout_base`. The session is not confined to that workspace. `/tend-ci-runner:triage` writes and runs a reproduction derived from the issue body (its **Step 5: Reproduce the bug**), and `/tend-ci-runner:resolve-conflicts` — reached from the scheduled `tend-notifications` and from nightly — checks out a dependency-bot PR head and works on it. So a reader who gates on `github.event_name` the way the example shows still has the value present in sessions that fetch and execute outside content; the sentence tells them otherwise. The revision drops the category and says no event confines the session to text, keeping the gating recommendation, which is still the largest path.

`pre-commit run --all-files` passes on the revision, including the `sync install-tend references/` mirror hook.

PR #1323, opened 2026-09-19
fix(poll): report a cancelled gating check as unverified, not green

### LINKED ISSUE #1322: poll_pr_checks.py reports green for a cancelled gating check, because CANCELLED is missing from RED_CONCLUSIONS
`poll_pr_checks.py` reports `green` for a commit whose gating check was **cancelled**. `CANCELLED` is a `COMPLETED` check-run conclusion but it is not in `RED_CONCLUSIONS`, so such a check lands in neither the `pending` nor the `failed` bucket, `_settle` sees an empty `pending` and returns settled, and `poll` prints the green verdict with exit 0. The `approval` subcommand reaches `approve: no failing check` by the same route.

Reproduced live against a real commit, twice, in `numbagg/numbagg` at the `0.2.11` pin:

```
$ poll_pr_checks.py poll 824 9f7d6b6fbbd0b135299e9878d28fb411466c3eba
green: every gating check on 9f7d6b6fbbd0b135299e9878d28fb411466c3eba settled green
note: branch advanced to a48ac31a4288a03c8ebbf346b747e1b6bbed031a — the result above is still 9f7d6b6fbbd0b135299e9878d28fb411466c3eba's, the commit this run is accountable for
# exit 0

$ poll_pr_checks.py approval 824 9f7d6b6fbbd0b135299e9878d28fb411466c3eba
approve: no failing check on 9f7d6b6fbbd0b135299e9878d28fb411466c3eba
# exit 0
```

That commit's `benchmark` check never ran to completion, and GitHub's own rollup for it is `FAILURE`:

```
$ gh api repos/numbagg/numbagg/commits/9f7d6b6fbbd0b135299e9878d28fb411466c3eba/check-runs \
    --jq '.check_runs[] | "\(.name) \(.status) \(.conclusion)"' | grep benchmark
benchmark status=completed conclusion=cancelled

$ gh api graphql -f query='{repository(owner:"numbagg",name:"numbagg"){
    object(oid:"9f7d6b6fbbd0b135299e9878d28fb411466c3eba"){
      ... on Commit{statusCheckRollup{state}}}}}'
{"state":"FAILURE"}
```

The other nine check runs on that commit are `success`, apart from `nightly`, which is `skipped` by design on PRs.

## Mechanism

[`RED_CONCLUSIONS`](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/plugins/tend-ci-runner/scripts/poll_pr_checks.py#L21-L27) lists `FAILURE`, `TIMED_OUT`, `STARTUP_FAILURE`, `ACTION_REQUIRED` and `ERROR`. [`reduce_rollup`](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/plugins/tend-ci-runner/scripts/poll_pr_checks.py#L118-L128) then derives exactly two buckets from it — `pending` is `status != "COMPLETED"`, `failed` is `status == "COMPLETED" and conclusion in RED_CONCLUSIONS`. A cancelled check is `COMPLETED` with conclusion `CANCELLED`, so it appears in neither, and every caller reads "absent from both" as "green".

`STALE` has the same gap. `SUCCESS`, `NEUTRAL` and `SKIPPED` are correctly non-blocking — numbagg's `nightly` job is `skipped` on every PR by design.

## Why it matters

The consumer-visible shape is the concurrency cascade, which is the ordinary case rather than a corner: `test.yaml` sets `cancel-in-progress: true`, so any second push to a branch cancels the first head's run. A session accountable for the *first* commit — a nightly that pushes, then polls while a `tend-review` session pushes its own findings — polls a SHA whose long job was killed and is told it settled green. numbagg's `benchmark` job runs ~16 min, so it is reliably the one still in flight when the supersede lands, and it is reliably the one reported green without having run.

[`head_note`](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/plugins/tend-ci-runner/scripts/poll_pr_checks.py#L196-L207) is the only signal a reader gets, and it reads as reassurance — "the result above is still `<sha>`'s, the commit this run is accountable for" — rather than as a warning that the result is partly vacuous.

The `approval` path is the one that can leave something standing. Its docstring already carries the distinction this issue is about — it handles the `if: always()` omnibus that reports a cancellation as `FAILURE` by reading the run behind each failure and [printing `unverified`](https://github.com/max-sixty/tend/blob/9d4b2618df55c9cfceedc77909260771d45ac251/plugins/tend-ci-runner/scripts/poll_pr_checks.py#L294-L306) — but that branch only fires for checks that came back `FAILURE` *and* with others still pending. A check whose conclusion is literally `CANCELLED` skips it and reaches `approve: no failing check`, indistinguishable from a real green. On a PR where the long job was cancelled, that is an approval over a check that never ran.

## Suggested shape

Surface cancellation as a third state instead of folding it into green — the vocabulary already exists in the `approval` path and in `poll`'s `UNVERIFIED, not green` returns. Concretely: give `reduce_rollup` an `unverified` bucket for `CANCELLED` (and `STALE`), have `poll` name those checks and return non-zero rather than printing the green line, and have `approval` report them the way it already reports a cancellation cascade.

Whether a cancelled gating check should *block* an approval or approve-with-unverified is your policy call — the existing branch chose the latter deliberately. The narrow defect is that neither subcommand currently says the check was cancelled at all.

<details><summary>The occurrence that surfaced this</summary>

`tend-nightly` [35439155546](https://github.com/numbagg/numbagg/actions/runs/35439155546) opened [numbagg/numbagg#824](https://github.com/numbagg/numbagg/pull/824) at `9f7d6b6fbbd0b135299e9878d28fb411466c3eba` and polled it. `tend-review` [35439731115](https://github.com/numbagg/numbagg/actions/runs/35439731115) pushed `a48ac31a4288a03c8ebbf346b747e1b6bbed031a` at 11:26:41Z, which cancelled the first head's `Test` run [35439731126](https://github.com/numbagg/numbagg/actions/runs/35439731126); its `benchmark` job was cut at 11:27:03Z after ~9 min of a ~16-min run, while the run's other seven jobs had already passed and `nightly` was skipped. The poll returned the green line above, and the nightly's summary reported "My commit on #824 (`9f7d6b6`) went green on all gating checks."

No harm followed there: the superseding head was independently polled to real green by the review session, the PR is self-authored so no approval was available, and nothing merged on the strength of the wrong verdict. It is the verdict itself, and the `approval` path sharing the code, that this issue is about.

Found by `tend-review-runs` [35441799469](https://github.com/numbagg/numbagg/actions/runs/35441799469).

</details>

tend-agent:
A check run that is `COMPLETED` without a usable conclusion matched neither bucket `reduce_rollup` derived — `pending` is `status != "COMPLETED"`, `failed` is `COMPLETED` plus a conclusion in `RED_CONCLUSIONS` — so `_settle` saw an empty `pending`, and `poll` printed `green: every gating check … settled green` with exit 0 over a check that never produced a result. The ordinary trigger is concurrency: a second push to a branch cancels the first head's run, and a session accountable for that first commit is told its long job passed. `approval` reached `approve: no failing check` by the same route.

`reduce_rollup` now returns a third `unverified` bucket, and green is the allowlist rather than the remainder: `GREEN_CONCLUSIONS = {"SUCCESS", "NEUTRAL", "SKIPPED"}`, and any other conclusion on a `COMPLETED` check that isn't red is unverified. That covers `CANCELLED` and `STALE`, a `COMPLETED` check carrying no conclusion, and any conclusion GitHub adds to `CheckConclusionState` later — deriving green by subtraction is what let the original false green through, so the unrecognized case is fail-closed by construction rather than by keeping a second list current.

`poll` names those checks and exits 2 — which `/tend-ci-runner:monitor-ci` already reads as "unverified, not green" — and names them beside a red or cap-hit verdict too, so the reader knows the picture is partial. `approval` keeps its existing policy of approving over a check that never ran, since a check that reached no verdict cannot withhold on its merits, but now prints it as unverified instead of silently; `references/approving.md` describes the widened meaning.

Supersession is unchanged and still decides first: where a rerun concluded at the same SHA, the latest run of each (name, workflow) wins and nothing is left unverified. Status contexts route exactly as before — `SUCCESS` green, `ERROR`/`FAILURE` red, `PENDING`/`EXPECTED` never `COMPLETED`.

Six test cases in `generator/tests/test_poll_scripts.py` fail on `main` and pass here: `test_cancelled_is_not_a_verdict` (parametrized over `CANCELLED` and `STALE`, previously asserting exit 0 for a cancelled check, now exit 2 with the check named), `test_cancelled_beside_a_real_failure_still_reads_red`, `test_approval_names_a_cancelled_check_it_approves_over`, and `test_a_conclusion_in_neither_set_is_not_green` (parametrized over an absent conclusion and one GitHub does not emit today).

Two more pass on `main` and guard against this change over-reaching: `test_cancelled_replaced_at_the_same_sha_is_green` pins that a cancellation superseded by a green rerun at the same SHA still reads green, and `test_non_blocking_conclusions_still_read_green` pins `NEUTRAL` and `SKIPPED`, which nothing covered before.

The file's 44 tests pass, as does the full Python suite apart from the two `test_refresh_consumers.py` cases that need PyPI DNS. `pre-commit run --all-files` is clean.

The mechanism and the live reproduction are in #1322, against [`poll_pr_checks.py`](https://github.com/max-sixty/tend/blob/79bfd206c76d8456b99d08306dafaaa326771cd1/plugins/tend-ci-runner/scripts/poll_pr_checks.py#L21-L32).

</details>

Closes #1322 — automated triage


tend-agent (commented):
`reduce_rollup` still derives green by subtraction: `unverified` is an allowlist of `CANCELLED`/`STALE`, so a `COMPLETED` check whose conclusion is outside both sets — one GitHub adds to `CheckConclusionState` later, or a null conclusion that reaches the reducer as `""` — lands in neither bucket, and `poll` prints `green: every gating check … settled green` over it. That is the same false green this PR fixes, one conclusion value away. `test_terminal_non_success_conclusions_count_red`'s own docstring already names the shape: "left out of both buckets it would read as green".

Naming the passing set instead is the same amount of code and closes the default:

```python
GREEN_CONCLUSIONS = {"SUCCESS", "NEUTRAL", "SKIPPED"}
...
and context["conclusion"] not in (RED_CONCLUSIONS | GREEN_CONCLUSIONS)
```

Status contexts route exactly as before under it — `SUCCESS` green, `ERROR`/`FAILURE` red, `PENDING`/`EXPECTED` never `COMPLETED` — so only the unrecognized case moves.

I'm pushing that to this branch, along with a test for the description's claim that `SUCCESS`, `NEUTRAL` and `SKIPPED` stay non-blocking (nothing covered `NEUTRAL` or `SKIPPED`), and a widening of "cancelled before concluding" to "settled without a result" now that the bucket holds more than cancellations.

Nothing else in the change needs one: supersession, the red set, and the `approval` policy of approving over a check that never ran all read correctly, and the four new tests fail on `main`.

tend-agent (commented):
Two statements of the premise this change removes survive elsewhere in the files it touches, both outside the diff hunks.

`references/approving.md` — the **After the approval** bullet beginning `A check was cancelled` still says "The replacement run will cover the cancelled checks." Under the new reducer that is false for exactly the checks the poll now surfaces: a cancellation with a rerun at the same SHA is superseded in `reduce_rollup` and never reaches `unverified` (which `test_cancelled_replaced_at_the_same_sha_is_green` pins), so a cancelled check that makes the poll exit 2 is one nothing replaced. The bullet tells that session to do nothing, on grounds the **Before the APPROVE** bullet twenty lines above now contradicts — and it is the only instruction the post-approval poll's reader has for this outcome. The do-not-rerun rule in it still holds and should stay; what has to go is the coverage claim, replaced by naming the check as unverified.

`test_terminal_non_success_conclusions_count_red`'s docstring — "left out of both buckets it would read as green" — now describes the opposite of what the file asserts two tests down, where `test_a_conclusion_in_neither_set_is_not_green` requires exit 2 for a conclusion in neither set. The reason `STARTUP_FAILURE` and `ACTION_REQUIRED` must stay in `RED_CONCLUSIONS` is now that otherwise they read as unverified rather than red — a different wrong verdict, not a green one.

Neither line falls inside a diff hunk, so no inline suggestion; I'm pushing both to this branch.

